PR #627, opened 2022-01-24
move device_code_flow example

cataggar:
It is unfortunate that dev_dependencies are checked when publishing. They shouldn't be.
https://github.com/rust-lang/cargo/issues/4242

This moves the examples that uses azure_storage_blobs to azure_storage_blobs to eliminate the cyclic dependency and allow publishing to azure_identity.


bmc-msft:
While I agree this specific example should be in the storage SDKs, it would be great to have a simple example that uses this SDK.

thovoll:
Could you just do what the other two identity tests do that use the blob API?
client_credentials_flow_blob.rs
code_flow_blob.rs

Moving just device_code_flow.rs to the blob crate doesn't make that much sense.

thovoll:
There's an independent case to be made for having examples in the blob crate that use different auth methods.

cataggar:
Today's goal is to get azure_identity published. The dependency on azure_storage & azure_blob_storage must be removed.

bmc-msft:
I understand the goal.  Please create a device_code_flow example that doesn't rely on the azure storage crate, as this is a common pattern that we should include as part of this crate.

cataggar:
I do not have time for that today. I don't think adding that example should block publishing.

bmc-msft (approved):

bmc-msft:
Please add a work item so adding it doesn't get forgotten.

PR #630, opened 2022-01-25
revert back to "azure_core"

ctaggart:
- This changes the core crate back to `azure_core`.
- Bumps `azure_core` & `azure_identity` to 0.1.1.
- Relaxes the service crates to depend on `0.1` instead of `0.1.0`.

### DIFF
--- a/sdk/iot_hub/Cargo.toml
+++ b/sdk/iot_hub/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Microsoft Corp."]
 edition = "2018"
 
 [dependencies]
-azure_core = { package = "azure_base", path = "../core", version = "0.1.0" }
+azure_core = { path = "../core", version = "0.1.0" }
 base64 = "0.13"
 bytes = "1.0"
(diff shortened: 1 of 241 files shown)


sdk/iot_hub/Cargo.toml
```diff
@@ -5,7 +5,7 @@ authors = ["Microsoft Corp."]
 edition = "2018"
 
 [dependencies]
-azure_core = { package = "azure_base", path = "../core", version = "0.1.0" }
+azure_core = { path = "../core", version = "0.1.0" }
```
ctaggart, line 8: This should be `0.1` as well.

yoshuawuyts (approved):

yoshuawuyts:
> Relaxes the service crates to depend on 0.1 instead of 0.1.0.

Note though that `0.1.0` and `0.1` resolve to the same version on Crates.io. If someone takes a dependency on `0.1.0` and a `0.1.1` is published, that is considered "semver compatible" and the latest version will always be used.

PR #647, opened 2022-01-29
 Add partition key support to DocumentClient.get_document 

bngo92:
All item operations should act on partition keys.
The implementation is copied from delete_document.


rylev (approved):
This seems like a strict improvement. However, is the partition key *always* necessary. it's my understanding that the partition key is not included when the document in question was [not created with a partition key](https://docs.microsoft.com/en-us/rest/api/cosmos-db/get-a-document#headers).

PR #656, opened 2022-02-12
list_crates

ctaggart:
This script is useful to figure out which service crates still need to be published for a version.


bmc-msft:
In a future PR, it might be useful to run `git clone --depth 1 https://github.com/rust-lang/crates.io-index` within an ephemeral [tempdir](https://docs.rs/tempfile/3.3.0/tempfile/fn.tempdir.html).

bmc-msft (approved):

Issue #652, opened 2022-02-04
Return an error instead of panicking when finalizing builders


dbanty:
Specifically, I followed [this example](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/storage_blobs/examples/shared_access_signature.rs) in order to create an Azure Blob SAS in tests, but didn't bother adding an expiration because I didn't realize it was required. Calling `finalize()` resulted in a panic due to [this line](https://github.com/Azure/azure-sdk-for-rust/blob/967b9bdc3388c6f305836abd3836b4430b7f4fc4/sdk/storage/src/core/shared_access_signature/account_sas.rs#L278). It would be a better experience to return a `Result` from `finalize` so it's clear that it's fallible and that some steps may be required. 

Alternatively, it could fail during compilation either by using types to track which of the required steps have been completed or by putting the required parameters in the initial constructor function. That may be an even easier experience for consumers but I suspect is more work to add to the API.

dbanty:
Just upgraded to the newer interfaces, thanks for fixing this!

PR #661, opened 2022-02-22
replace `unwrap` calls in azure_core

bmc-msft:
In multiple places, the SDK panics at runtime.  This changes the unwrap
calls to handle the error in a non-panic fashion.

For places that have predicates that validate the invariant, unwrap has
been replaced with `expect` that describes the justification.

This is related to, but does not address, #652

### DIFF
--- a/sdk/core/src/error/http_error.rs
+++ b/sdk/core/src/error/http_error.rs
@@ -20,10 +20,8 @@ impl HttpError {
         let mut headers = HashMap::new();
 
         for (name, value) in response.headers() {
-            // TODO: non-UTF8 header values will not be included
-            if let Ok(value) = value.to_str() {
-                headers.insert(name.to_string(), value.to_string());
-            }
+            let value = String::from_utf8_lossy(value.as_bytes()).to_string();
+            headers.insert(name.to_string(), value);
         }
 
         let body = response.into_body_string().await;
--- a/sdk/core/src/error/http_error.rs
+++ b/sdk/core/src/error/http_error.rs
@@ -1,4 +1,5 @@
 use crate::Response;
+use std::collections::HashMap;
 
 /// An unsuccessful HTTP response
 #[derive(Debug)]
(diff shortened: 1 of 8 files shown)


sdk/core/src/error/http_error.rs
```diff
@@ -16,13 +17,15 @@ impl HttpError {
...
-            .headers()
-            .into_iter()
-            // TODO: the following will panic if a non-UTF8 header value is sent back
-            // We should not panic but instead handle this gracefully
-            .map(|(n, v)| (n.as_str().to_owned(), v.to_str().unwrap().to_owned()))
-            .collect();
+        let mut headers = HashMap::new();
+
+        for (name, value) in response.headers() {
+            // TODO: non-UTF8 header values will not be included
```
rylev, line 23: I suppose if we cannot convert to a string, we could convert to bytes and then use `String::from_utf8_lossy`.

sdk/core/src/policies/retry_policies/retry_policy.rs
```diff
@@ -61,29 +61,33 @@ where
...
                         ),
-                        error,
+                        http_error,
                         "server returned error status which will not be retried",
                     );
 
-                    if !RETRY_STATUSES.contains(&status) {
-                        log::error!(
-                            "server returned error status which will not be retried: {}",
+                    if let Ok(status) = status {
```
rylev, line 77: Hmm I'm not sure this is better. If the status getting returned is not able to be turned into a `Status` type, that's a bug on our part. I think panicking here is actually the right thing to do.

rylev (changes requested):
Looks good, but I do have one thing that I think we should not change.

sdk/core/src/error/http_error.rs
bmc-msft: addressed

sdk/core/src/policies/retry_policies/retry_policy.rs
bmc-msft: The response parsing context here is handling a pre-defined set of status codes that should require retrying the request.

My thought is that there are valid HTTP status codes that `http::StatusCode` does not support, including 425 Early Data.   Instead of panicking in that case, this instead does not retry.

Of note, we're not actually returning the StatusCode, just using it as a method to let users set how we respond to specific status codes.

https://www.rfc-editor.org/rfc/rfc8470.html

rylev: > My thought is that there are valid HTTP status codes that http::StatusCode does not support, including 425 Early Data. Instead of panicking in that case, this instead does not retry.

Hmm... that's a real shame and seems like a bug in the `http` crate. I wonder if it would be better to just get rid of the use of the `StatusCode` type completely...

bmc-msft: Revisiting this today, I think this is a non-issue.

The underlying libraries used to parse the response use `http::StatusCode`, which means the value has to be valid in this context anyways.

Also relevant, this isn't a user-facing feature.  The SDK does not expose specifying a custom set of status codes to retry.  In this context, we only use `http:;StatusCode` as a convenience mechanism.  It's easier to read remember `StatusCode::REQUEST_TIMEOUT` than the number 408 and a comment.

I'll update this code to use expect, such that it's clear we this should not ever fail.

rylev (approved):
Looks great! Just needs a rebase

Issue #657, opened 2022-02-14
Some crates unconditionally depend on ring


Skepfyr:
Four of the SDK crates unconditionally depend on *ring*: [storage](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/storage/Cargo.toml#L18), [data_cosmos](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/data_cosmos/Cargo.toml#L20), [storage_datalake](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/storage_datalake/Cargo.toml#L31), and [messaging_servicebus](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/messaging_servicebus/Cargo.toml#L19). In order to comply with the Microsoft security policy we must use openssl instead of ring. This can probably be configured using features [like in core](https://github.com/Azure/azure-sdk-for-rust/blob/b27e4bf79455aa8cc25c3779c11232252a8011b6/sdk/core/Cargo.toml#L48-L51). However it's not obvious to me what the code should do if ring and openssl are present.

The current use of ring appears to be entirely confined to hmac:
* https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/storage_datalake/src/shared_key_authorization_policy.rs#L228
* https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/messaging_servicebus/src/event_hub/client.rs#L32
* https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/data_cosmos/src/authorization_policy.rs#L233
* https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/storage/src/core/shared_access_signature/mod.rs#L14
* https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/storage/src/core/clients/storage_account_client.rs#L460

PR #662, opened 2022-02-22
move to hmac instead of ring for hmac calculation

bmc-msft:
As #657 identifies, multiple crates depend on ring.  This PR replaces the use of `ring` with the `hmac` and `sha256`, as we implicitly use `sha256` from the `oauth2` crates in `core`, `identity`, `storage_blobs`, and `security_keyvault`.


cataggar (approved):
This looks like the right way to go.

rylev (approved):

PR #663, opened 2022-02-24
More `into_future` usage

rylev:
Worked on with @yoshuawuyts 

In #510 we decided on the `into_future` pattern for operations. These PR takes that pattern and simply applies it to more operations in the cosmos crate. 

We also introduced a new feature which allows the hopefully soon to be stabilized [`IntoFuture`](https://doc.rust-lang.org/stable/std/future/trait.IntoFuture.html) from the standard library. 

We were able to get through all operations in cosmos up to `delete_user`. This PR might be easier to review commit by commit as opposed to all at once.

PR #665, opened 2022-02-25
More `into_future` implementations

rylev:
A simple continuation of #663 with three more operations


yoshuawuyts (approved):
LGTM; this makes me happy!

PR #668, opened 2022-03-01
Remove `Box::pin` requirement in usage of `Pageable`

yoshuawuyts:
Need to wait on CI to pass. But if things work out this removes the need to use `Box::pin` when iterating over a `Pageable` type, simplifying the use of our streaming APIs. Thanks!


rylev (approved):
🎉 Amazing!

PR #672, opened 2022-03-02
update services with latest specs

cataggar:
I also added a CHANGELOG.md for the generated services.

I'm checking all the features in https://github.com/cataggar/azure-sdk-for-rust/actions/runs/1924895767 . If it runs for 6 hours , we are good.


services/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# 0.2 (2022-03)
+
+- generated from azure-rest-api-specs [commit from 2022-03-02](https://github.com/Azure/azure-rest-api-specs/commit/c56b44b794ff17caad4d31c6ab64b07e28253504)
```
cataggar, line 3: This is the commit used. It is now in the `CHANGELOG.md`.

services/mgmt/appplatform/Cargo.toml
```diff
@@ -33,4 +33,3 @@ no-default-version = []
 "package-preview-2021-06" = []
 "package-preview-2020-11" = []
 "package-2020-07" = []
-"package-2019-05-01-preview" = []
```
heaths, line 36: Interesting to see a version removed. What prompted this?

heaths (approved):
Shared files and a sampling of generated files LGTM.

bmc-msft (commented):
I thought we were going to bump the revs of all of the crates when regenerating with updated specs.

Even the crates with changes, I don't see version bumps.

services/mgmt/appplatform/Cargo.toml
cataggar: The spec was removed. The easiest way to find the changes is to go to its readme and then look at history. https://github.com/Azure/azure-rest-api-specs/blob/main/specification/appplatform/resource-manager/readme.md . That will take you to https://github.com/Azure/azure-rest-api-specs/pull/17506 where it was removed.

PR #673, opened 2022-03-02
[datalake] More Authorization policies

roeap:
This PR adds more authorization options to the datalake crate. Specifically using a token credential.

In many scenarios having auto refresh available for tokens is quite handy. Thus I added an auto refreshing token credential, which wraps around token credentials from the core crate. To get the refresh to work, I had to use the tokio Mutex - I believe this is required due to the refresh being an async operation and the regular mutex cannot be used across an await statement. But I may also just have missed the simpler solution. Due to that I put the implementation behind a feature gate, since I seem to remember reading pulling in a specific runtime as dependency was not desired.

I still need to add tests and clean up a bit, but would appreciate some feedback on the general approach - spcifically the use of tokio.

@thovoll @rylev @ctaggart 

Update: usage of `futures::lock::Mutext` solves the Tokio issue (thanks @bmc-msft)


bmc-msft:
Are you able to use futures::lock::Mutex for a runtime agnostic Mutex implementation?

roeap:
Indeed, that works :) Thanks for the advice!

of course the next question follows suit... when returning an error inside the "refresh loop" I could not get it to work to return a Box'ed error due to lifetime constraints. The new errors using an `ErrorKind` and message would work fine, but this would require adjusting the return type of the `TokenCredential.get_token` function. Also I was hesitant to introduce a new variant to the "old" errors, and changing the signature would have a much bigger impact. So again asking for some guidance. 

Right now an `HeaderNotFound` erro is returned since it only needs a string as input, but that obviously is more misleading then helpful :).

yoshuawuyts:
> Also I was hesitant to introduce a new variant to the "old" errors

I discussed this with @thovoll today, and we think this would be the best option. There is no "right" answer here, but this would allow us to start incrementally migrating, the way we've been doing with other types as well.

Also we on reading the code we realized using `Mutex` might cause contention between readers; instead using an [async `RwLock`](https://docs.rs/async-lock/2.5.0/async_lock/struct.RwLock.html) might make reads cheaper.

Hope that's helpful!

roeap:
Thanks @yoshuawuyts, @thovoll for the support. The code now uses RwLock from the newly introduced `async_lock` crate (did not find one in `futures`), and also added a new variant to the error enum.

The last remaining concerns was testing - locally I gave the credential a try and it seemed to work. Looked a bit around the identity crate and found that specific credentials are not tested there as well - not sure how to proceed with this.

Otherwise I hope  we are approaching a merge-able state :).

yoshuawuyts:
I'm not super familiar with how we're testing things; perhaps @cataggar can answer that question?

thovoll (approved):

sdk/storage_datalake/src/authorization_policies/auto_refreshing_token.rs
```diff
@@ -0,0 +1,65 @@
...
+
+impl std::fmt::Debug for AutoRefreshingTokenCredential {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        f.debug_struct("AutoRefreshingTokenCredential")
+            .field("credential", &"TokenCredential")
+            .finish()
+    }
+}
+
+impl AutoRefreshingTokenCredential {
```
cataggar, line 27: Shouldn't this go in `azure_core`?

thovoll: It will eventually. I know @roeap wanted to do it in stages.

roeap: I completely agree :). If we can move this there now, great. Similarly, the pipeline authorization policies should likely end up at least in `azure_storage`, or if they are useful in other crates as well maybe also core?

If this is OK, I am happy to move this to core in this PR.

thovoll:
> I'm not super familiar with how we're testing things; perhaps @cataggar can answer that question?

@roeap and I discussed the testing plan and came up with these incremental steps:
1) Add an example (this PR)
2) Add an e2e record/playback test (maybe another PR)
3) Add a deterministic test for the loop with the RwLock (definitely another PR)

sdk/storage_datalake/src/authorization_policies/auto_refreshing_token.rs
roeap: Moved the `AutoRefreshingCredential` into the identity crate to the other `TokenCredentials`. Was considering to also move the policies to the `storage_core` crate, but thought it better to do this when migrating blob to the pipeline architecture.

roeap:
Moved the refreshing credential in the identity crate and updated the client constructors to also offer an option for using a token credential. With respect to client initialization I tried to keep changes minimal. Main reason being that from what I gather discussion is still ongoing on what the specifics of client configuration should look like.

sdk/identity/src/token_credentials/auto_refreshing_credentials.rs
```diff
@@ -0,0 +1,66 @@
...
+        }
+        loop {
+            let mut guard = self.current_token.write().await;
+            match guard.as_ref() {
+                None => {
+                    let res = self.credential.get_token(resource).await;
+                    *guard = Some(res);
+                }
+                Some(Err(err)) => {
+                    return Err(Error::AuthorizationPolicy(err.to_string()));
```
cataggar, line 54: `err.to_string` does not feel right

sdk/core/src/errors.rs
```diff
@@ -42,6 +42,8 @@ pub enum Error {
     Json(#[source] serde_json::Error),
     #[error("Other error")]
     Other(#[source] Box<dyn std::error::Error + Send + Sync + 'static>),
+    #[error("Other error")]
```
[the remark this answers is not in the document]
cataggar: Better description need here

cataggar (changes requested):

sdk/identity/src/token_credentials/auto_refreshing_credentials.rs
cataggar: I'm hoping @rylev has better advice for a solution.

sdk/core/src/errors.rs
```diff
@@ -42,6 +42,8 @@ pub enum Error {
     Json(#[source] serde_json::Error),
     #[error("Other error")]
     Other(#[source] Box<dyn std::error::Error + Send + Sync + 'static>),
+    #[error("Authorization policy error")]
+    AuthorizationPolicy(String),
```
rylev, line 46:
```suggestion
    #[error("authorization policy error: {0}")]
    AuthorizationPolicy(Box<dyn std::error::Error + Send + Sync + 'static>),
```

Nit: can you also move this above the `Other` variant so that `Other` is last?

sdk/core/src/error/azure_core_errors.rs
```diff
@@ -19,6 +19,10 @@ impl From<crate::errors::Error> for Error {
             crate::errors::Error::GetToken(e) => Error::new(ErrorKind::Credential, e),
             crate::errors::Error::HttpPrepare(e) => e.into(),
             crate::errors::Error::Other(e) => Error::new(ErrorKind::Other, e),
+            crate::errors::Error::AuthorizationPolicy(msg) => Error::with_message(
```
rylev, line 22: If `AuthorizationPolicy` takes a boxed error trait object, you can just use `Error::new` instead of `Error::with_message` which doesn't really add much value.

sdk/identity/src/token_credentials/auto_refreshing_credentials.rs
rylev: I suggested above changing the error to take a boxed error trait object which would clean this up a bit.

sdk/identity/src/token_credentials/auto_refreshing_credentials.rs
```diff
@@ -0,0 +1,66 @@
+use async_lock::RwLock;
+use azure_core::{
+    auth::{TokenCredential, TokenResponse},
+    Error,
+};
+use chrono::{Duration, Utc};
+use std::sync::Arc;
+
+fn is_expired(token: TokenResponse) -> bool {
```
rylev, line 9: Can this just take a `&TokenResponse` so you don't have to clone it all the time?

sdk/storage_datalake/src/lib.rs
```diff
@@ -16,14 +16,14 @@ extern crate serde_derive;
 #[macro_use]
 extern crate azure_core;
 
+pub mod authorization_policies;
```
rylev, line 19: Can this be `pub(crate)`?

rylev (changes requested):
Just a few comments but largely this looks good.

sdk/core/src/error/azure_core_errors.rs
roeap: It seems that when working with the guard from RwLock, we need to get a reference (`guard.as_ref()`) , which in turn ties the lifetime of the error to the TokenCredential itself, which then would need to be constrained by the 'static lifetime.

At least that's more or less what i gathered from the compiler errors, I get when trying to pass a box'ed error into the variant. Tried around different variations of de-referencing the error, but with no success. Somehow these lifetimes still manage to escape my grasp every so often :).

sdk/core/src/errors.rs
roeap: Switched the variants, but was not able to work with the boxed error. See comment below.

PR #510, opened 2021-11-11
Operation Builder Future

rylev:
*coauthored with @yoshuawuyts*

This is a proposal for a new one style of API for operations that eliminates some of the awkwardness of usage without reducing the functionality.

### Summary

Operations now return "builders" which implement an `into_future` method (more of this later) that turns the builder into a future which can be awaited with `await` and yields a response. These builders provide setter methods for any options as well as a way to set a `Context` object. 

**Currently**

```rust
collection_client
            .create_document(
                my_context,
                &document_to_insert,
                CreateDocumentOptions::new().is_upsert(true),
            )
            .await?;
```

**After this PR**

```rust
collection_client
            .create_document(&document_to_insert)
            .is_upsert(true) // if the default is desired this line can be removed
            .context(my_context) // if no context is needed, this line can be removed
            .into_future() // more of this below
            .await?;
```

## `into_future`

Currently, when the code `some_value.await` is called, Rust desugars this to a direct call to the `poll` function on `Future`. However, there is an accepted RFC and [implementation](https://github.com/rust-lang/rust/pull/90737) that changes this desugaring to call the currently unstable `IntoFuture::into_future`. When this lands on stable Rust, we will be able to remove the need for calling `into_future()` as the call to `await` will call `into_future` automatically. 

## The builder pattern

The builder pattern presented here is very common (and is in fact what the Azure SDK for Rust used before the pipeline architecture). Here are examples of it being used in the most popular HTTP crates for Rust:
* [hyper](https://docs.rs/http/0.2.5/http/request/struct.Builder.html)
* [reqwest](https://docs.rs/reqwest/0.11.6/reqwest/blocking/struct.RequestBuilder.html)
* [surf](https://docs.rs/surf/2.3.2/surf/struct.RequestBuilder.html)

Thoughts?

cc @JeffreyRichter @heaths 

I also just saw @JeffreyRichter's [comment](https://github.com/Azure/azure-sdk-for-rust/pull/504#issuecomment-965809346) on requiring `Context`. That doesn't change this proposal too much, it would just mean keeping context as the first argument to every operation.

Issue #290, opened 2021-06-11
[Tracking Issue] Convert Cosmos to Pipeline Architecture


rylev:
This is a tracking issue for converting operations over to the pipeline architecture. There are some [sub](https://github.com/Azure/azure-sdk-for-rust/issues/254) [issues](https://github.com/Azure/azure-sdk-for-rust/issues/255) for converting individual SDK crates over, but this issue acts as a guide for how to switch to the new architecture. 

### Overview 

Previous to the pipeline architecture, each operation was independently controlled meaning there was no central way to control the way operations were performed. 

The pipeline architecture on the other hand allows users of the SDK to create pipelines through which outgoing requests and incoming responses are processed. The pipeline are composed of "policies" which can both change the outgoing request and the incoming response. Policies are usually either on a "per-request" basis (meaning they are called only once per operation) or "per-retry" (meaning they are applied every time an operation is attempted). 

Examples of policies are the `TelemetryPolicy` which adds various telemetry headers to the request and the retry policy which checks whether an incoming response was successful and if it detects a transient failure, retries the request. 

### Converting to the Architecture

For an example of converting one operation (Cosmos's `get_database` operation) over to the pipeline architecture, take a look at https://github.com/Azure/azure-sdk-for-rust/pull/286.

The basic steps are:
* Create a new submodule of the `operations` module where the operation will live. This submodule should be named after the operation (e.g., for `get_database` it's called `get_database`). 
* Move the "request builder" (e.g., for `get_database` the `GetDatabaseBuilder`) to new operation submodule and rename it from `Builder` to `Options` (e.g., `GetDatabaseBuilder => `GetDatabaseOptions`). This now represents the options that can be supplied when performing the operation. Note that many options that were part of the builder are no longer valid options because there will be individual policies that take their place. For instance, previously many operations took an option to set the user agent. This is now done by a policy. 
* Change the `execute` method to a `decorate_request` method that instead of performing the request, simply mutates a request parameter with any of the options supplied. 
* Move the associated response type (e.g., for `get_database` the `GetDatabaseResponse` type from the `responses` submodule to the operation submodule. Previously a `std::convert::TryFrom` implementation was used to convert from an `http::Response<Bytes>` object. This should now be a plain associated async function called `try_from` that converts from the new `Response` type. See the example for how this should be done. 
[...]

PR #677, opened 2022-03-04
Last `into_future`

rylev:
Finishes the work started in https://github.com/Azure/azure-sdk-for-rust/pull/510 moving to a builder style `into_future` pattern.

This also completes https://github.com/Azure/azure-sdk-for-rust/issues/290 moving all operations in Cosmos over to the pipeline architecture.


sdk/core/src/mock/mock_request.rs
```diff
@@ -8,17 +8,6 @@ use std::str::FromStr;
...
-    fn new(uri: Uri, method: Method, headers: HeaderMap, body: Body) -> Self {
-        Self {
-            uri,
-            method,
-            headers,
-            body,
-        }
-    }
-}
-
```
yoshuawuyts, line 21: Oh, it took me a sec to realize this had just been moved to `core/src/request.rs`. That seems good.

sdk/data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -18,13 +13,10 @@ use std::sync::Arc;
 pub const EMULATOR_ACCOUNT_KEY: &str =
     "C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==";
 
-const AZURE_VERSION: &str = "2018-12-31";
-
```
yoshuawuyts, line 22: Where did this go? Has this been replaced by something else?

yoshuawuyts (approved):
Had one question, but other than that this looks good!

sdk/data_cosmos/src/clients/cosmos_client.rs
rylev: This is done in the `telemetry_policy` now (not a part of this PR). This can be removed since now everything goes through the pipeline and the telemetry policy will be used.

Issue #634, opened 2022-01-25
hitting docs.rs sandbox limits


cataggar:
For some of the generated service crates, we are hitting sandbox limits. We want documentation for all features/tags/versions, which is a factor.

azure_svc_synapse
https://docs.rs/crate/azure_svc_synapse/0.1.0/builds
https://docs.rs/crate/azure_svc_synapse/0.1.0/builds/498821
memory limit

cataggar:
The page to monitor for failures during a release is:
https://docs.rs/releases/recent-failures

The failures in this release:
- [ ] azure_mgmt_apimanagement
- [ ] azure_mgmt_compute, 8 GB failed
- [ ] azure_mgmt_containerregistry
- [ ] azure_mgmt_containerservice
- [x] azure_mgmt_cosmosdb, 8 GB
- [ ] azure_mgmt_monitor
- [ ] azure_mgmt_network
- [ ] azure_mgmt_recoveryservicesbackup
- [ ] azure_mgmt_recoveryservicessiterecovery
- [ ] azure_mgmt_sql
- [x] azure_mgmt_synapse, 8 GB
- [ ] azure_mgmt_web

cataggar:
Hopefully fixed by #679.

cataggar:
Only 1 recent failure:
https://docs.rs/releases/recent-failures
- https://docs.rs/crate/azure_svc_servicefabric/0.2.0/builds


125 are stuck in the queue as priority `-1` 😞 
https://docs.rs/releases/queue

cataggar:
I requested 6 GB for azur_svc_servicefabric.
https://github.com/rust-lang/docs.rs/issues/1709

cataggar:
Reducing the number of versions solved the problem for most. May reopen next time a publish is done.

cataggar:
The 2022 August Release had these fail:
- https://docs.rs/crate/azure_svc_servicefabric/0.5.0

PR #679, opened 2022-03-04
limit docs.rs to 5 API tags

cataggar:
For #634, this limits the published API tags to 5, the default and 4 more.


services/svc/schemaregistry/Cargo.toml
```diff
@@ -21,7 +21,7 @@ azure_identity = { path = "../../../sdk/identity", version = "0.1" }
 tokio = { version = "1.0", features = ["macros"] }
 
 [package.metadata.docs.rs]
-all-features = true
+features = []
```
heaths, line 24: Doesn't this mean nothing gets documented? I'd expect this to be a subset of `features.default` <= 5 features. This might be a symptom of no default with which I was concerned above.

heaths (changes requested):

services/autorust/codegen/src/cargo_toml.rs
```diff
@@ -77,3 +83,22 @@ fn get_default_feature(tags: &[&Tag], default_tag: Option<&str>) -> String {
         None => tags[0].rust_feature_name(),
     }
 }
+
+const MAX_DOCS_RS_FEATURES: usize = 4;
+
+/// Get a list of features to document at docs.rs in addition the default
+pub fn docs_rs_features(tags: &[&Tag], default_feature: &str) -> Vec<String> {
```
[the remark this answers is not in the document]
ctaggart: There may not be a default tag. There default feature takes that into account, see [fn get_default_feature](https://github.com/Azure/azure-sdk-for-rust/blob/6945d239adfa3557912361333557639ad3c29608/services/autorust/codegen/src/cargo_toml.rs#L65-L79).

services/svc/schemaregistry/Cargo.toml
ctaggart: The `default` feature gets documented.

ctaggart: This can be tested with:
```
cargo doc -p azure_svc_schemaregistry --no-deps --open --features "no-default-version"
```
image

ctaggart:
I've tested by passing the list of features to `cargo doc`. All these work:

cargo doc -p azure_svc_schemaregistry --no-deps --open --features "no-default-version"

cargo doc -p azure_mgmt_app --no-deps --open --features "no-default-version"

cargo doc -p azure_mgmt_vmware --no-deps --open --features "no-default-version package-2021-06-01 package-2021-01-01-preview package-2020-07-17-preview package-2020-03-20"

heaths (approved):

PR #681, opened 2022-03-09
Downgrade log line from debug to trace

adeschamps:
This statement logs the entire request body, which can be fairly large. I think it should at least be moved down to trace level -- or, if it's not useful to end users, removed.


heaths:
Of more concern is that we do this without any sanitization like we do in other languages. But I don't see that as blocking for this PR, which I think makes sense. I'll open a separate issue to address sanitization by default.

heaths (approved):

PR #683, opened 2022-03-10
Turn on 'test_e2e' feature for rust-analyzer

rylev:
For rust-analyzer users this will show enable rust-analyzer features in e2e tests

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,6 @@
 {
-    "editor.formatOnSave": true
+    "editor.formatOnSave": true,
+    "rust-analyzer.cargo.features": [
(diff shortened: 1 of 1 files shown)


heaths:
What's the impact to normal dev workflow? Does this effectively mean e2e tests will run during integrated tests in VSCode?

rylev:
No, this just means that rust-analyzer will be active (i.e., showing red squiggles, auto-completion, go to definition, etc.) in code that has `cfg(feature = "test_e2e")` set on it (meaning it only compiles when the "test_e2e" feature is on). This doesn't impact anything outside of rust-analyzer features.

heaths (approved):

Issue #164, opened 2021-01-29
Storage: Stream is not Sync


thomastaylor312:
This is a follow on to #133 in which I admit I didn't read my own notes. The returned stream from the blob client does is not `Sync` (and not just Unpin), which limits some of its downstream use (as it can often get passed down the stack). The specific error in my case when I try to return an `impl Stream + Send + Sync`:

```
error[E0277]: `dyn Future<Output = std::result::Result<http::response::Response<bytes::Bytes>, Box<dyn StdError + Send + Sync>>> + Send` cannot be shared between threads safely
   --> src/provider/mod.rs:315:12
    |
315 |         Ok(Box::new(resp))
    |            ^^^^^^^^^^^^^^ `dyn Future<Output = std::result::Result<http::response::Response<bytes::Bytes>, Box<dyn StdError + Send + Sync>>> + Send` cannot be shared between threads safely
    |
    = help: the trait `Sync` is not implemented for `dyn Future<Output = std::result::Result<http::response::Response<bytes::Bytes>, Box<dyn StdError + Send + Sync>>> + Send`
    = note: required because of the requirements on the impl of `Sync` for `Unique<dyn Future<Output = std::result::Result<http::response::Response<bytes::Bytes>, Box<dyn StdError + Send + Sync>>> + Send>`
    = note: required because it appears within the type `Box<dyn Future<Output = std::result::Result<http::response::Response<bytes::Bytes>, Box<dyn StdError + Send + Sync>>> + Send>`
    = note: required because it appears within the type `Pin<Box<dyn Future<Output = std::result::Result<http::response::Response<bytes::Bytes>, Box<dyn StdError + Send + Sync>>> + Send>>`
    = note: required because it appears within the type `for<'r, 's, 't0, 't1, 't2, 't3, 't4, 't5> {ResumeTy, &'r GetBlobBuilder<'s>, reqwest::Url, http::request::Request<bytes::Bytes>, StatusCode, GetBlobBuilder<'t0>, &'t1 BlobClient, &'t2 (dyn azure_core::HttpClient + 't3), Pin<Box<(dyn Future<Output = std::result::Result<http::response::Response<bytes::Bytes>, Box<(dyn StdError + Send + Sync + 't4)>>> + Send + 't5)>>, ()}`
[...]

PR #685, opened 2022-03-12
Add reqest properties for `azure_kusto_data` crate

roeap:
With Kusto being my favourite cloud analytics engine, i have been working with a rust implementation to interact with the service. Recently I updated the code to conform with what I hope is the current best practices in this repository. Talking with some members of the Kusto product group, they mentioned they would welcome a contribution towards a rust SDK for their service (@adieldar @cosh). MY naive hope is, that is would be OK to contribute the code to this repo, such that it can evolve alongside the best practices developed in this repo.

So far the client only supports executing queries, but i this hope that it could be useful to a wider audience.

Any feedback on whether such a contribution would be welcome is greatly appreciated.

@yoshuawuyts @ctaggart @rylev @thovoll


AsafMah:
In general, with the naming convention in other kusto sdks the crate should be called "azure-kusto-data"

cataggar:
Hey @AsafMah, we followed https://azure.github.io/azure-sdk/general_design.html#namespaces when naming the other crates. In the guidelines it says `data` is for "Dealing with structured data stores like databases". We followed those guidelines for naming both:

- azure_data_cosmos
- azure_data_tables

I think that the `azure_data_kusto` matches the guidelines and our other crates the best. Please ping me on Teams if you want to discuss. Good to have you on here.

sdk/data_kusto/src/operations/query.rs
```diff
@@ -0,0 +1,174 @@
...
+        Box::pin(async move {
+            let url = this.client.query_url();
+            let mut request = this.client.prepare_request(url, http::Method::POST);
+
+            add_mandatory_header2(
+                &ContentType::new("application/json; charset=utf-8"),
+                &mut request,
+            )
+            .unwrap();
+            add_mandatory_header2(&Accept::new("application/json"), &mut request).unwrap();
```
AsafMah, line 66: Any reason for all the unwraps in this file?
Using `?` seems to compile without error, and of course won't trigger a panic.

sdk/data_kusto/src/operations/query.rs
```diff
@@ -0,0 +1,174 @@
...
+use azure_core::prelude::*;
+use azure_core::setters;
+use azure_core::{
+    collect_pinned_stream,
+    headers::{add_mandatory_header2, add_optional_header2},
+    Response as HttpResponse,
+};
+use futures::future::BoxFuture;
+
+/// A future of a delete file response
```
AsafMah, line 12: I don't get it, when are you deleting a file?

sdk/data_kusto/src/request_options.rs
```diff
@@ -0,0 +1,67 @@
+use azure_core::{AddAsHeader, HttpHeaderError, Request};
+use http::request::Builder;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Accept<'a>(&'a str);
```
AsafMah, line 5: This file seems a bit sus,
Accept and AcceptEncoding are really common headers, it's a bit weird that we need to define them like that in the kusto package specfically.
It seems like something that should be defined in azure_core.

sdk/data_kusto/src/connection_string.rs
```diff
@@ -0,0 +1,336 @@
...
+                APPLICATION_KEY_NAME => application_key = Some(v),
+                APPLICATION_TOKEN_NAME => application_token = Some(v),
+                APPLICATION_CERTIFICATE_NAME => application_certificate = Some(v),
+                APPLICATION_CERTIFICATE_THUMBPRINT_NAME => {
+                    application_certificate_thumbprint = Some(v)
+                }
+                AUTHORITY_ID_NAME => authority_id = Some(v),
+                MSI_PARAMS_NAME => msi_params = Some(v),
+                FEDERATED_SECURITY_NAME => match v {
+                    "true" => federated_security = Some(true),
```
AsafMah, line 228: DRY - you're doing this exact comparison for multiple props

sdk/data_kusto/src/connection_string.rs
```diff
@@ -0,0 +1,336 @@
...
+            let v = match kv.next() {
+                Some(v) if v.chars().all(char::is_whitespace) => {
+                    return Err(ConnectionStringError::MissingValue { key: k.to_owned() })
+                }
+                None => return Err(ConnectionStringError::MissingValue { key: k.to_owned() }),
+                Some(v) => v,
+            };
+
+            match k {
+                DATA_SOURCE_NAME => data_source = Some(v),
```
AsafMah, line 214: Parsing connection strings are a bit more complex than that - the names aren't case sensitive, and some of them have a aliases.
Take a look at https://github.com/Azure/azure-kusto-node/blob/master/azure-kusto-data/source/connectionBuilder.ts

sdk/data_kusto/src/authorization_policy.rs
```diff
@@ -0,0 +1,55 @@
...
+
+#[async_trait::async_trait]
+impl Policy for AuthorizationPolicy {
+    async fn send(
+        &self,
+        ctx: &Context,
+        request: &mut Request,
+        next: &[Arc<dyn Policy>],
+    ) -> PolicyResult {
+        assert!(
```
AsafMah, line 41: Why assert over an error?

AsafMah (commented):
Some comments from going over the code

sdk/data_kusto/src/operations/query.rs
roeap: this is an artifact from migration from where i developed the code ... the plan is to remove all unwraps throughout the code before merging!

sdk/data_kusto/src/operations/query.rs
roeap: copy and paste error - will clean this up :)

sdk/data_kusto/src/authorization_policy.rs
roeap: This was copied from other pipelines. My understanding would be that this should either never or always happen at runtime, since creating the pipeline is crate internal. Thus - i think - the decision was made to just panic. @cataggar - is that correct?

sdk/data_kusto/src/operations/query.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    setters! {
+        client_request_id: ClientRequestId => Some(client_request_id),
+        query: String => query,
+        database: String => database,
+        context: Context => context,
+    }
+
+    pub fn into_future(self) -> ExecuteQuery {
```
AsafMah, line 53: I think we can completely avoid Box pin in this method and also simplify it:


```rust
pub async fn into_future(self) -> crate::error::Result<KustoResponseDataSetV2> {
        let url = self.client.query_url();
        let mut request = self.client.prepare_request(url, http::Method::POST);

        add_mandatory_header2(
            &ContentType::new("application/json; charset=utf-8"),
            &mut request,
        )
            .unwrap();
        add_mandatory_header2(&Accept::new("application/json"), &mut request).unwrap();
        //add_mandatory_header2(&AcceptEncoding::new("gzip,deflate"), &mut request).unwrap();
        add_optional_header2(&self.client_request_id, &mut request).unwrap();

        let body = QueryBody {
            db: self.database,
            csl: self.query,
        };
        request.set_body(bytes::Bytes::from(serde_json::to_string(&body)?).into());

        let mut context = self.context;

        let response = self
            .client
            .pipeline()
            .send(&mut context, &mut request)
            .await?;

        KustoResponseDataSetV2::try_from(response).await
    }
```

sdk/data_kusto/src/operations/query.rs
```diff
@@ -0,0 +1,174 @@
...
+            let url = this.client.query_url();
+            let mut request = this.client.prepare_request(url, http::Method::POST);
+
+            add_mandatory_header2(
+                &ContentType::new("application/json; charset=utf-8"),
+                &mut request,
+            )
+            .unwrap();
+            add_mandatory_header2(&Accept::new("application/json"), &mut request).unwrap();
+            add_mandatory_header2(&AcceptEncoding::new("gzip,deflate"), &mut request).unwrap();
```
AsafMah, line 67: This header causes the response to get back as gzip, which the azure core library doesn't seem to decypher, so I get an error unless I remove it. Is it the same experience for you?

sdk/data_kusto/src/request_options.rs
roeap: moved the request options to core crate

sdk/data_kusto/src/operations/query.rs
roeap: Tried to adjust the `into_future` method to the latest approach from the cosmos crate and also added some more optional request options to illustrate the usage of the request builder. Also added the implementation for the so far only in nightly `IntoFuture` trait to show how the explicit `inio_future` call will dissapear once that feature lands in stable.

sdk/data_kusto/src/operations/query.rs
roeap: hmm.. when i run it locally, it actually seems to be required, strange. One way to compare would be to record a query using the mock transport framework and compare how the same response behaves on different machines. Do you maybe have a test cluster to run queries against that we can record? I could then replay the test and see if it fails or passes?

sdk/data_kusto/src/operations/query.rs
roeap: moved to new header handling and removed the panics.

sdk/data_kusto/src/operations/query.rs
roeap: gone :)

sdk/data_kusto/src/connection_string.rs
roeap: moved boolean parsing into its own reusable function.

sdk/data_kusto/src/connection_string.rs
roeap: added a basic function to support aliases for keys based on the link. Need to look into the implementation there a bit more, but i think we need to normalize a bit more via lower casing input strings.

sdk/data_kusto/src/operations/query.rs
AsafMah: You still left the box::pin and copying which again I don't think are needed here

roeap: We had one issue in the past, that i think will re-appear if we drop that. Essentially, when used in API we may want to return the future rather then evaluating the future in the API itself, and then turning the response into a future again. Main reason being that if we don't copy the client and box/pin it, the lifetime of the future is directly tied to the lifetime of the client itself, which sometimes significantly limits usability in async contexts. However I am not entirely sure that I fully understand all implications - I think @rylev and @yoshuawuyts did the majority of the designs around this topic. 

a related issues (i think):
- #164

sdk/core/src/request_options/accept.rs
```diff
@@ -0,0 +1,29 @@
+use crate::headers::{self, Header};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Accept<'a>(&'a str);
```
rylev, line 4: Can we use `String` or `Cow<'static, str>` instead of `&'a str`?

rylev, line 4: Can we add some documentation to these header types?

sdk/data_kusto/src/authorization_policy.rs
rylev: This is correct. Triggering this would mean a bug in the SDK code itself and thus we want to fail fast here.

sdk/data_kusto/src/connection_string.rs
```diff
@@ -0,0 +1,389 @@
...
+    #[error("Missing value for key '{}'", key)]
+    MissingValue { key: String },
+    #[error("Unexpected key '{}'", key)]
+    UnexpectedKey { key: String },
+    #[error("Parsing error: {}", msg)]
+    ParsingError { msg: String },
+}
+
+#[derive(Default)]
+pub struct ConnectionStringBuilder<'a>(ConnectionString<'a>);
```
rylev, line 100: Let's add some more docs to these types.

sdk/data_kusto/src/connection_string.rs
```diff
@@ -0,0 +1,389 @@
...
+        let mut msi_params = None;
+        let mut az_cli = None;
+
+        let kv_str_pairs = connection_string
+            .split(';')
+            .filter(|s| !s.chars().all(char::is_whitespace));
+
+        for kv_pair_str in kv_str_pairs {
+            let mut kv = kv_pair_str.trim().split('=');
+            let k = match kv.next() {
```
rylev, line 259: You can avoid the first match variant if you do `kv.next().filter(|k| !k.chars().all(char::is_whitespace))`

sdk/data_kusto/src/connection_string.rs
```diff
@@ -0,0 +1,389 @@
...
+                    })
+                }
+                None => {
+                    return Err(ConnectionStringError::ParsingError {
+                        msg: "No key found".to_owned(),
+                    })
+                }
+                Some(k) => k,
+            };
+            let v = match kv.next() {
```
rylev, line 272: Same point as above.

sdk/data_kusto/src/error.rs
```diff
@@ -0,0 +1,41 @@
+//! Defines `KustoRsError` for representing failures in various operations.
+use std::fmt::Debug;
+use thiserror;
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
```
rylev, line 6: Just as an FYI: we want to move away from crates have their own error types in favor of just a single error type defined in azure_core. This is fine for now though - we can make that change later once we finish making the change in the cosmos crate.

sdk/data_kusto/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+#[macro_use]
+extern crate serde_derive;
```
rylev, line 2: Can we get rid of this and just refer to `Serialize` and `Deserialize` through full paths: e.g., `serde::Deserialize`

sdk/data_kusto/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[dependencies]
+azure_core = { path = "../core", version = "0.1" }
+azure_identity = { path = "../identity", version = "0.1" }
+async-trait = "0.1"
+bytes = "1"
+futures = "0.3"
+http = "0.2"
+serde = "1"
+serde_derive = "1"
```
rylev, line 22: Can we just bring this in through `serde`: `serde = { version = "1.0", features = ["derive"] }`

sdk/data_kusto/src/operations/query.rs
```diff
@@ -0,0 +1,191 @@
...
+use azure_core::setters;
+use azure_core::{collect_pinned_stream, Response as HttpResponse};
+use futures::future::BoxFuture;
+
+type ExecuteQuery = BoxFuture<'static, crate::error::Result<KustoResponseDataSetV2>>;
+
+#[derive(Debug, Serialize, Deserialize)]
+struct QueryBody {
+    db: String,
+    csl: String,
```
rylev, line 12: If we are going to use short names like this we should document what they need.

sdk/core/src/request_options/accept.rs
roeap: Used `String` since this seems more "common" right now. Was almost about to migrate more headers, but that's better left for another PR :)

roeap: Yes, done. For the headers also available in the HTTP crate I just copied some docs from there. For other I went for the text in the Kusto docs.

sdk/data_kusto/src/lib.rs
roeap: Done, much cleaner now. Do I understand correctly that the use of  `macro_use` could be considered deprecated in recent Rust versions in favour of direct imports?

sdk/data_kusto/src/operations/query.rs
roeap: added some docs.

sdk/data_kusto/Cargo.toml
roeap: yes :)

sdk/data_kusto/src/connection_string.rs
roeap: nice, again saved a few lines.

sdk/data_kusto/src/operations/query.rs
roeap: Trying to understand this. Aside from the `Box::pin`, if we do not clone the client and do the `async move`, my understanding was we would again run into the situation, where the futures and clients lifetimes would be correlated, and could not have a function that return the future directly for use in downstream functions, but rather have to await the future and then wrap the returned value into a future again. We ran into a situation like this with the `list_blobs` function from storage in the delta-rs crate, and I wanted to understand the actual root cause, since we have yet to clean that um in delta.

sdk/data_kusto/src/operations/query.rs
roeap: turns out, now it's the same experience for me :D. SO removed the header for now.

roeap: this was related to the core crate not including the `gzip` feature of `reqwest`, added it since it as an optional feature so we can use gzip in this crate.

sdk/data_kusto/src/operations/query.rs
```diff
@@ -0,0 +1,195 @@
...
+            request.insert_headers(&ContentLength::new(bytes.len() as i32));
+            request.set_body(bytes.into());
+
+            let response = self
+                .client
+                .pipeline()
+                .send(&mut ctx.clone(), &mut request)
+                .await?;
+
+            <KustoResponseDataSetV2 as TryFrom<HttpResponse>>::try_from(response).await
```
roeap, line 89: @yoshuawuyts - I tried using your `async-convert` crate here to eventually make this builder generic over the response type. The management and query request builders would likely look more or less the same, and differ only in the response type. I wanted to end up up with something like

```rust
response.try_into().await
```

but I am doing something wrong. Or am completely off track :). Could you give us a hint if this is the right place to use it, and if so, how?

roeap:
@AsafMah - when writing some tests i took data from the python implementation and ran into some errors. specifically in some example data over there the json response contains the tokens `Infinity` / `-Infinity` which `serde_json` considers to be invalid and thus panics on parsing. it seems that the python package is the only one from go / java / node / python where this scenario is covered, so not sure how the other SDKs behave in that case.

i found the crate [python-json-read-adapter](https://github.com/mitsuhiko/python-json-read-adapter) which replaces these tokens with 0.0 so it can be parsed, but altering data seems like a bad idea to me. I thought about converting the bytes to string and doing some string manipulation, but that feels like it would yield a significant performance hit.

do you have any ideas / thoughts how to best handle that situation? Also, do you know if its just Rust that thinks this is invalid, or doe json parsers in other languages share that opinion?

AsafMah:
Yeah, I've been through this exact thing - sorry for not updating you.

Btw - have you seen my PR? https://github.com/roeap/azure-sdk-for-rust/pull/5

AsafMah:
Also we need to consider where we want this project to live - 
From our team's end, we want it to be in its own repo like the other sdks,
But for now it seems very rooted here, and putting it in a different repo would mean waiting until changes will be inserted into core, which may take a while.

roeap:
> Yeah, I've been through this exact thing - sorry for not updating you.
> 
> Btw - have you seen my PR? [roeap#5](https://github.com/roeap/azure-sdk-for-rust/pull/5)

no i had not. some GitHub failed to notify me. Look much better and merged it!

roeap:
> From our team's end, we want it to be in its own repo like the other sdks

This pretty much answers it for me :).

As you mentioned, since the repo is still moving quite fast, there is a question if this should be now or later. But from a technical standpoint we could keep using git dependencies for a while. This would block releasing it to crates.io but my personal experience is that integrating changes in this repo is quite fast and I also believe that the maintainers here are quite willing to do fairly frequent updates / releases. As such this does not need to be the case for long ...

AsafMah:
@roeap 
Hey, sorry for "abusing" this platform, but it seems like you don't get my emails, so I'll write to you here.
Maybe I got your address wrong - I'm in asafmahlev@microsoft.com .
How do you prefer I contact you? Thanks!

sdk/data_kusto/src/operations/query.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    /// Consumes the response into an iterator over all PrimaryResult tables within the response dataset
+    pub fn into_primary_results(self) -> impl Iterator<Item = DataTable> {
+        self.tables
+            .into_iter()
+            .filter(|t| matches!(t, ResultTable::DataTable(tbl) if tbl.table_kind == TableKind::PrimaryResult))
+            .map(|t| match t {
+                ResultTable::DataTable(tbl) => tbl,
+                _ => unreachable!("All other variants are excluded by filter"),
+            })
```
AsafMah, line 137: .filter_map(|table| match table {
                ResultTable::DataTable(table) if table.table_kind == TableKind::PrimaryResult => Some(table),
                _ => None,
            })

sdk/data_kusto/src/arrow.rs
```diff
@@ -0,0 +1,272 @@
...
+use std::sync::Arc;
+
+const SECOND_TO_NANOSECONDS: i64 = 1000000000;
+const MINUTES_TO_SECONDS: i64 = 60;
+const HOURS_TO_SECONDS: i64 = 60 * MINUTES_TO_SECONDS;
+const DAYS_TO_SECONDS: i64 = 24 * HOURS_TO_SECONDS;
+const TICK_TO_NANOSECONDS: i64 = 100;
+
+#[inline]
+fn to_nanoseconds(days: i64, hours: i64, minutes: i64, seconds: i64, ticks: i64) -> i64 {
```
AsafMah, line 20: You might have easier time with this using the time or chrono crates

AsafMah:
@roeap but basically - I've created a repo detached from this one that follows the conventions of our other kusto SDKs, and has the code from here - https://github.com/Azure/azure-kusto-rust/pull/3

Also there is a roadmap onenote document.

We obviously don't want you to implement the whole SDK yourself - so I'd like to discuss which features you'd like to do and what you'd want to leave for us.

I'd like to speak to you more, but again - we need a better way to communicate :)

roeap:
@AsafMah - I finally read my emails :).

sdk/data_kusto/src/lib.rs
rylev: Deprecated is perhaps too strong of a word, but it is becoming less and less idiomatic over time, and there is movement in the language team to deprecate this perhaps in the next edition.

roeap:
After talking with the Kusto PG we moved the kusto specific code into a separate repository - https://github.com/Azure/azure-kusto-rust, so I removed the specific code here. I think the changes to header / request properties in the core crate are still relevant as they are used by other azure services as well, so it would be great to still get them merged here.

@cataggar

cataggar (approved):
Looks like it is just adding support for a few standard headers now.

PR #686, opened 2022-03-14
Use the new `IntoFuture` signature on nightly

yoshuawuyts:
We changed the signature of `IntoFuture` on nightly in https://github.com/rust-lang/rust/pull/94818. It appears there are no nightly releases with this change yet, but once there are (which shouldn't be more than a few days away) we should merge this to match the new signature. Thanks!


yoshuawuyts:
Looking at CI https://github.com/Azure/azure-sdk-for-rust/runs/5540261760?check_suite_focus=true and the [Rust nightly docs](https://doc.rust-lang.org/nightly/std/future/trait.IntoFuture.html#associatedtype.IntoFuture) it seems `IntoFuture::IntoFuture` has landed on a nightly release. Seems like it's time to merge this!

PR #691, opened 2022-03-21
Better header handling

rylev:
*worked on with @yoshuawuyts! 🎉*

This PR introduces a few mechanisms which make it easier to handle headers.

In general, we provide rich high-level types that can be mapped to some set of headers. This is accomplished through the `AsHeaders` trait which allows a type to define an iterator over some set of `(HeaderName, HeaderValue)`. 

As a convenience there is the `Header` trait which allows a type to specify itself as mapping of a type 1 to 1 to a header. This is provided since a 1 to 1 mapping is the most common case and so makes definitions a bit more straight forward. 

This change is largely internal facing (i.e., users won't really notice a difference). However, internally, the definition of how to turn a type into headers is much shorter and adding those headers to a request is a bit more succinct.

### DIFF
--- a/sdk/data_cosmos/src/operations/replace_document.rs
+++ b/sdk/data_cosmos/src/operations/replace_document.rs
@@ -65,13 +65,13 @@ impl<D: Serialize + Send + 'static> ReplaceDocumentBuilder<D> {
                 .unwrap_or_else(|| self.client.partition_key_serialized());
             add_as_partition_key_header_serialized2(partition_key, &mut request);
 
-            azure_core::headers::add_mandatory_header2(&self.indexing_directive, &mut request)?;
-            azure_core::headers::add_optional_header2(&self.if_match_condition, &mut request)?;
-            azure_core::headers::add_optional_header2(&self.if_modified_since, &mut request)?;
+            request.insert_headers(&self.indexing_directive);
+            request.insert_headers(&self.if_match_condition);
+            request.insert_headers(&self.if_modified_since);
             if let Some(cl) = &self.consistency_level {
                 request.insert_headers(cl);
             }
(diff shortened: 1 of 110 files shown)


cataggar (approved):

sdk/data_cosmos/src/operations/replace_document.rs
```diff
@@ -65,11 +65,13 @@ impl<D: Serialize + Send + 'static> ReplaceDocumentBuilder<D> {
...
-            azure_core::headers::add_optional_header2(&self.if_match_condition, &mut request)?;
-            azure_core::headers::add_optional_header2(&self.if_modified_since, &mut request)?;
-            azure_core::headers::add_optional_header2(&self.consistency_level, &mut request)?;
-            azure_core::headers::add_mandatory_header2(&self.allow_tentative_writes, &mut request)?;
+            request.insert_headers(&self.indexing_directive);
+            request.insert_headers(&self.if_match_condition);
+            request.insert_headers(&self.if_modified_since);
+            if let Some(cl) = &self.consistency_level {
+                request.insert_headers(cl);
+            }
```
yoshuawuyts, line 73: Something which stands out to me here is that `insert_headers` no longer returns a `Response`. I believe this would allow us to write the following:

```rust
&self.consistency_level.map(request.insert_headers);
```

rylev: Unfortunately, we need to do the following to get it to compile:

```rust
self.consistency_level
                .as_ref()
                .map(|c| request.insert_headers(c));
```
This doesn't really save us very much in terms of readability...

yoshuawuyts: oh no, you're right!

PR #694, opened 2022-03-23
Add Device update functionality

rerkcp:
This PR implements the Azure Device Update API. The code is based on the security_keyvault crate.
Feedback welcome even if the PR is not merged.


sdk/device_update/examples/list_operations.rs
```diff
@@ -0,0 +1,45 @@
...
+        client_id,
+        client_secret,
+        TokenCredentialOptions::default(),
+    );
+    let mut client = DeviceUpdateClient::new(&device_update_url, &creds)?;
+
+    let s_filter= match env::var("DEVICE_UPDATE_FILTER") {
+        Err(_e) => { "".to_owned() },
+        Ok(s) => { s }
+    };
```
bmc-msft, line 26:
```suggestion
    let s_filter = env::var("DEVICE_UPDATE_FILTER").unwrap_or_default();
```

bmc-msft:
Perhaps this would be a good candidate to wrap the service-generated code in a rust-idiomatic layer.  

For reference: [generated crate](https://github.com/Azure/azure-sdk-for-rust/tree/main/services/svc/deviceupdate) and [docs](https://docs.rs/azure_svc_deviceupdate/latest/azure_svc_deviceupdate/)

sdk/device_update/examples/list_operations.rs
rerkcp: done, thank you

rerkcp:
> Perhaps this would be a good candidate to wrap the service-generated code in a rust-idiomatic layer.
> 
> For reference: [generated crate](https://github.com/Azure/azure-sdk-for-rust/tree/main/services/svc/deviceupdate) and [docs](https://docs.rs/azure_svc_deviceupdate/latest/azure_svc_deviceupdate/)

I tried to do that in the beginning so as not to copy all the struct and enum definitions, but I ran into problems with integrating the svc crate and the sdk crate together due to Rusts checks on how crates can reference each other and could not get it to even compile. Is there an example on how to do this I could look at?

sdk/device_update/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+thiserror = "1.0"
+reqwest = { version = "0.11", features = ["json"] }
+chrono = { version = "0.4", features = ["serde"] }
+const_format = "0.2.13"
+serde_json = "1.0"
+url = "2.2"
+serde = { version = "1.0", features = ["derive"] }
+getset = "0.1"
+azure_core = { path = "../core", version = "0.2" }
+tokio = { version = "1.0", features = ["full"] }
```
rylev, line 26: We shouldn't need to depend on `tokio`.

sdk/device_update/examples/delete_update.rs
```diff
@@ -0,0 +1,36 @@
+use azure_device_update::DeviceUpdateClient;
```
rylev, line 1: I think a lot of these examples should be tests instead and combined together.

sdk/device_update/src/client.rs
```diff
@@ -0,0 +1,209 @@
...
+/// # Example
+///
+/// ```no_run
+/// use azure_device_update::DeviceUpdateClient;
+/// use azure_identity::token_credentials::DefaultAzureCredential;
+/// let creds = DefaultAzureCredential::default();
+/// let client = DeviceUpdateClient::new("contoso.api.adu.microsoft.com", &creds).unwrap();
+/// ```
+#[derive(Debug)]
+pub struct DeviceUpdateClient<'a, T> {
```
rylev, line 20: We're generally avoiding tying clients to some specific lifetime. We should just make it easily cloneable.

sdk/device_update/src/client.rs
```diff
@@ -0,0 +1,209 @@
...
+
+        let resp = reqwest::Client::new()
+            .put(&uri)
+            .bearer_auth(self.token.as_ref().unwrap().token.secret())
+            .header("Content-Type", "application/json")
+            .body(body)
+            .send()
+            .await
+            .unwrap();
+        let body = resp.text().await?;
```
rylev, line 94: We'll want to use the pipeline architecture we're using in `azure_data_ cosmos`.

rylev (approved):
Thanks for contributing! I think this is a good start though we'll want to start moving towards things like using the pipeline architecture like we do in `azure_data_cosmos`. I'll merge this for now, and we can make improvements in follow-up PRs.

PR #695, opened 2022-03-23
add links & README.md for services

### LINKED ISSUE #632: add links & README.md for services
> warning: manifest has no documentation, homepage or repository.

When publishing 0.1 of the generated service crates, that warning can be seen.

We should add those properties for the next service crate publishing, so that we get 

image

We should also generated a README.md, else things look pretty blank:

image

cataggar:
The primary purpose of the pull request is to add links & README.md for services. Fix #632.

For AutoRust, it also makes a couple of additional changes:
- switch to `Utf8Path` & `Utf8PathBuf` from `camino`
- some error cleanup
- switched some io function to its own module

I'll be making a separate pull request to explain the tags.


services/autorust/codegen/src/cargo_toml.rs
```diff
@@ -28,9 +33,12 @@ pub fn create(crate_name: &str, tags: &[&Tag], default_tag: Option<&str>, path:
 [package]
 name = "{}"
 version = "0.2.0"
-edition = "2018"
+edition = "2021"
 license = "MIT"
 description = "generated REST API bindings"
+repository = "https://github.com/azure/azure-sdk-for-rust"
+homepage = "https://github.com/azure/azure-sdk-for-rust"
+documentation = "https://docs.rs/{}"
```
heaths, line 41: Since we're using `format!()` could we switch to named parameters? As more replacements are added it gets more difficult to keep them properly serialized.

services/autorust/codegen/src/io.rs
```diff
@@ -0,0 +1,60 @@
...
+    let mut c = a.as_ref();
+    if c.extension().is_some() {
+        c = c.parent().ok_or(Error::PopUpPath)?; // to directory
+    }
+    let mut c = PathBuf::from(c);
+    c.append(b.as_ref()).map_err(Error::AppendPath)?;
+    Utf8PathBuf::from_path_buf(c).map_err(Error::FromPathBuf)
+}
+
+pub fn join_several<P1: AsRef<Utf8Path>>(a: P1, b: &[Utf8PathBuf]) -> Result<Vec<Utf8PathBuf>> {
```
heaths, line 41: Test for this as well?

services/autorust/codegen/templates/readme.md.jinja
```diff
@@ -0,0 +1,5 @@
+# {{ crate_name }} crate
+
+The is an [Azure SDK for Rust](https://github.com/Azure/azure-sdk-for-rust) crate that is generated from the Azure REST API Specifications listed in:
```
heaths, line 3: Nit: "specifications" used in this way is not part of a proper name and should be lower case.
```suggestion
The is an [Azure SDK for Rust](https://github.com/Azure/azure-sdk-for-rust) crate that is generated from the Azure REST API specifications listed in:
```

heaths (approved):
Nice!

services/autorust/codegen/src/cargo_toml.rs
cataggar: I'd like to switch over to a `askama` "jinja" template in a future pull request for this too.

heaths (approved):

PR #696, opened 2022-03-23
bump edition to 2021

cataggar:
The edition was stable in https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html last October.

### DIFF
--- a/sdk/data_cosmos/src/operations/list_attachments.rs
+++ b/sdk/data_cosmos/src/operations/list_attachments.rs
@@ -46,6 +46,7 @@ impl ListAttachmentsBuilder {
             let this = self.clone();
             let ctx = self.context.clone();
             async move {
+                let _ = &this;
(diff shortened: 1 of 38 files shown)


sdk/data_cosmos/src/operations/list_attachments.rs
```diff
@@ -46,6 +46,7 @@ impl ListAttachmentsBuilder {
             let this = self.clone();
             let ctx = self.context.clone();
             async move {
+                let _ = &this;
```
cataggar, line 49: `cargo fix --edition` added this for the 2021 edition. Why the change? What is this doing? Can it be simplified?

cataggar: May be it is not needed.

rylev (approved):

PR #625, opened 2022-01-24
[Proposal] Error Handling

rylev:
This is a proposal for a new form of error handling in the SDKs I have been working on with @yoshuawuyts. This is an attempt to reduce the large amount of errors into as compact a form as possible while still offering good ergonomics and quality error messages. 

In the grand scheme of things, this proposal is not a radical departure from the current approach taken, but is an attempt to rethink error handling from first principles. 

## The current approach

The current approach to error handling is to create a large, all-encompassing error type modeled as a Rust enum. This looks like this:

```rust
/// A general Azure error type.
#[non_exhaustive]
#[derive(Debug, thiserror::Error)]
pub enum Error {
    #[error("parse error")]
    Parse(#[from] ParseError),
    #[error("error getting token")]
    GetToken(#[source] Box<dyn std::error::Error + Send + Sync>),
    #[error("http error")]
    Http(#[from] HttpError),
    /// .... 
}
````

This approach does have some nice properties:
* The `thiserror` crate makes it pretty easy to declare new error variants together with how they are displayed to the user with minimal boiler plate and that conform to best practices for error types including provide an implementation for the `Error::source` method which shows what other error *caused* this error. 
* Error messages are relatively consistent (i.e., if two parsing errors can happen in two different SDKs, they'll largely have the same error message)
* More error types can be added without breaking existing user code (through the use of `non_exhaustive`. 
* Conversions from 3rd party error types to this error type are generally fairly easy.

However, this approach does have downsides:
* The specific variant of `Error` in question is tied directly to the underlying data or error cause. For example, in the code above, an `Error::Parse` must always have a corresponding `ParseError`. This encourages an ever expanding tree of errors. 
* Context dependent error messages are hard to add. Since the data associated with an error is directly tied to the error variant, it's hard to add additional context to errors that happen for unique reasons. For example,. say a parsing error when the key "foo" must have a value with the prefix "bar" - this happens only in one place in all the SDKs. If you want to display to the user this important context, you need to add a new variant to the `ParseError` type that explains this. This adds even more pressure for an ever expanding tree of error types.

The above list is not exhaustive but in general should give you the impression that the way we're modelling errors encourages error types with *lots* of variants and sub errors. This is exactly what we see in practice, where the error types in `azure_core` are bloated and often hard to keep track of. 

## What do we need?

Before getting into solutions, I'd like to briefly discuss the general needs we have from our errors. Errors are generally used for two things:
* Control flow (i.e., the user inspects the errors and writes code specifically to handle a certain error case)
[...]

PR #702, opened 2022-03-25
Remove `azure_data_cosmos::Error`

rylev:
As a continuation of #625, this removes the bespoke error type for `azure_data_cosmos` in favor of using the new error type from `azure_core`. 

I think this has largely proven to be a win. Code was removed but the error messages we're now producing are better than they were before. 

It might be easiest to review this by going commit by commit.


rylev:
Looks like I broke some of the generated services. I'll try to clean that up so this becomes mergeable.

yoshuawuyts (approved):
This looks good!

Issue #693, opened 2022-03-22
Blob Storage: Set blob tier endpoint is missing


rminderhoud:
It appears that the endpoint to set a blob access tier is missing from the `azure_storage_blob` crate: https://docs.microsoft.com/en-us/rest/api/storageservices/set-blob-tier.

**Note**
For a future implementer or someone who needs this functionality in the interim, the tier can be set by crafting the request manually. Looks something like this:

```rust
let storage_url = storage_account_client.blob_storage_url();
let container_name = container_client.container_name();

let blob_url = format!(
            "{}{}/{}?comp=tier",
            storage_url.as_str(),
            container_name,
            blob_name
);

// Manually update the access tier using an http request since it's
// currently not directly supported by the SDK

let (request, _url) = storage_client
    .prepare_request(
        &blob_url,
        &http::Method::PUT,
        &|request| {
            request.header(
                azure_core::headers::BLOB_ACCESS_TIER,
                http::header::HeaderValue::from_str(tier.as_ref()).unwrap(),
            )
        },
        ServiceType::Blob,
        None,
    )
    })?;

let response = storage_client
    .http_client()
    .execute_request_check_status(request, StatusCode::OK)
    .await?;

// Parse response
```

rickrain:
Wondering if this might be labeled as a **good_first_issue**.  There seems to be a reasonable amount of guidance for one to embark on this given that we have a code reference above _and_ implementations of other API's to refer to.
 
I'm relatively new to Rust and looking for some opportunities to further my learning while also contribute.

PR #703, opened 2022-03-28
Blob Storage: Add Set Blob Access Tier

rickrain:
This addresses #693.

In the integration tests, I structured this so there is at least one transition to/from the access tiers (hot/cold/archive).  I didn't test every permutation.  Let me know if this is OK or if I should create a separate test for every permutation.


rylev (approved):
🎉 Thanks! This looks like a good addition.

PR #706, opened 2022-03-30
set azure_core to 0.2.1

cataggar:
It is set to 0.2.1 because 0.2.0 is [yanked](https://crates.io/crates/azure_core/versions). This updates the README and adds a CHANGELOG.

### DIFF
--- a/sdk/core/README.md
+++ b/sdk/core/README.md
@@ -1,3 +1,12 @@
-# azure_core crate
+# Azure SDK for Rust - Core crate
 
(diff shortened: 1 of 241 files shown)


sdk/core/README.md
```diff
@@ -1,12 +1,3 @@
-# Azure SDK for Rust - Core crate
+# azure_core crate
```
rylev, line 1: I don't really love these changes to the readme...

Perhaps reverting this and then opening a new PR where we can discuss the needs of a README and how we can address them?

rylev (changes requested):

rylev (approved):

PR #707, opened 2022-03-30
update azure_core readme

cataggar:
Extracted from #706 for discussion.


heaths (approved):
For now this will suffice, but eventually we'll want it to look like the others e.g., https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/README.md

PR #708, opened 2022-04-01
Adding support for certificate based credentials for app authentication

msabansal:
The implementation is inspired by the implementation of client_secret_credential.


sdk/identity/src/token_credentials/client_certificate_credentials.rs
```diff
@@ -0,0 +1,231 @@
...
+        Self {
+            authority_host: authority_hosts::AZURE_PUBLIC_CLOUD.to_owned(),
+            send_certificate_chain: false,
+        }
+    }
+}
+
+impl CertificateCredentialOptions {
+    /// Create a new TokenCredentialsOptions. default() may also be used.
+    pub fn new(authority_host: String, send_certificate_chain: bool) -> Self {
```
heaths, line 38: Ideally we take these as `str` but I see this is consistent with our existing credentials. We should fix that later.

sdk/identity/src/token_credentials/client_certificate_credentials.rs
```diff
@@ -0,0 +1,231 @@
...
+        let der = cert.to_der()?;
+        let digest = hash(MessageDigest::sha1(), &der)?;
+        Ok(digest)
+    }
+
+    fn as_jwt_part(part: &[u8]) -> String {
+        base64::encode(&part)
+            .replace('+', "-")
+            .replace('/', "_")
+            .replace('=', "")
```
heaths, line 118: For consistency with other crates, use `base64::*_config(..., BASE64_URL_SAFE)`.

sdk/identity/src/token_credentials/client_certificate_credentials.rs
```diff
@@ -0,0 +1,231 @@
...
+        let form_data = vec![
+            ("client_id", self.client_id.to_owned()),
+            ("scope", format!("{}/.default", resource)),
+            (
+                "client_assertion_type",
+                "urn:ietf:params:oauth:client-assertion-type:jwt-bearer".to_owned(),
+            ),
+            ("client_assertion", client_assertion),
+            ("grant_type", "client_credentials".to_owned()),
+        ];
```
heaths, line 200: @schaabs any concerns with this particular bit?

heaths (changes requested):
Great start! Could you also add an example / test for this crate like with other credentials? It would be expected that a customer provide their own certificate so that wouldn't need to be part of the sample, of course. But this also helps make sure it works as expected.

sdk/identity/Cargo.toml
```diff
@@ -26,6 +26,9 @@ log = "0.4"
 async-timer = "1.0.0-beta.7"
 thiserror = "1.0"
 async-trait = "0.1"
+openssl = { version = "0.10" }
```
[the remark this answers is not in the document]
msabansal: agreed. fixed.

sdk/identity/src/token_credentials/client_certificate_credentials.rs
msabansal: i was not aware of this part. Fixed.

msabansal:
@heaths Thanks for reviewing the PR and providing feedback. I have added an example and fixed the comments.

msabansal:
I have added an example. If there is some test infra i am happy to write a test aswell.

sdk/identity/src/token_credentials/client_certificate_credentials.rs
msabansal: I can do this in the current PR if you want. not doing it for now

sdk/identity/src/token_credentials/client_certificate_credentials.rs
heaths: @chlowell since Scott is OOF, what do you think about the substance of this method? You (and Scott) know that area of identity far better than I.

sdk/identity/src/token_credentials/client_certificate_credentials.rs
```diff
@@ -0,0 +1,232 @@
...
+            .map_err(ClientCertificateCredentialError::OpensslError)?;
+
+        let uuid = uuid::Uuid::new_v4();
+        let current_time = Utc::now().timestamp();
+        let expiry_time = current_time + DEFAULT_REFRESH_TIME;
+        let x5t = base64::encode(&thumbprint);
+
+        let header = match options.send_certificate_chain {
+            true => format!(
+                r#"{{"alg":"RS256","typ":"JWT", "x5t":"{}", "x5c":["{}"]}}"#,
```
chlowell, line 173: Looks like x5c can contain only one cert? A client may provide any number of intermediate certs in addition to the signing cert, in which case x5c must contain all of them.

sdk/identity/src/token_credentials/client_certificate_credentials.rs
```diff
@@ -0,0 +1,232 @@
...
+    client_certificate_pass: String,
+    options: CertificateCredentialOptions,
+}
+
+impl ClientCertificateCredential {
+    /// Create a new ClientCertificateCredential
+    pub fn new(
+        tenant_id: String,
+        client_id: String,
+        client_certificate: String,
```
chlowell, line 89: Docs should underscore that this is the base64 encoded bytes of a cert in PKCS12 format, not a cert in PEM format. Also, to fully support SNI, the credential must accept multiple certs.

msabansal: The api was inspired from the js sdk and there aswell a single certificate is expected. I am not sure if AAD mandates to send the entire chain If there is a change needed i am happy to follow up.

sdk/identity/src/token_credentials/client_certificate_credentials.rs
msabansal: Uncertain if there is a change requested here?

sdk/identity/src/token_credentials/client_certificate_credentials.rs
```diff
@@ -0,0 +1,232 @@
...
+
+        let jwt = format!("{}.{}", header, payload);
+        let signature = ClientCertificateCredential::sign(&jwt, &certificate.pkey)
+            .map_err(ClientCertificateCredentialError::OpensslError)?;
+        let sig = ClientCertificateCredential::as_jwt_part(&signature);
+        let client_assertion = format!("{}.{}", jwt, sig);
+
+        let form_data = vec![
+            ("client_id", self.client_id.to_owned()),
+            ("scope", format!("{}/.default", resource)),
```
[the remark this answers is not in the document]
msabansal: Uncertain if there is a change requested here?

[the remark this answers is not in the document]
heaths: In other languages we have credential options that let you specify 1 or more scopes, but as @chlowell implied that could be left for another PR since it will affect multiple credentials. I've opened https://github.com/Azure/azure-sdk-for-rust/issues/714 to track.

sdk/identity/src/token_credentials/client_certificate_credentials.rs
heaths: The "x5c" array should take one or more certificates.

msabansal: Fixed. Sending the entire chain if available.

bmc-msft:
Can this be an optional feature?  There are occasions where linking against OpenSSL is a significant hurdle.

sdk/identity/src/token_credentials/client_certificate_credentials.rs
chlowell: AAD requires any intermediate certs in the chain between the signing cert and one signed by a trusted issuer. I expect zero intermediate certs is the common case but if there are any, the client can't authenticate without providing them, so I suggest designing with that in mind.

Re the JS SDK, its certificate credential API was designed before SNI was on the roadmap, and the constructor took a path to a cert file on disk. That didn't preclude supporting SNI because a file can contain multiple certs, at least in PEM format (I'm less familiar with PKCS12). (I'm just trying to provide context here, not suggesting you follow along and take a file path. I think a collection of strings or bytes is better.)

msabansal: @chlowell Thanks for providing context. The chain is supported in the current model. If a chain is needed client_certificate is expected to contain the entire chain. The previous code had the bug where i was also sending the private key and according the docs the private key must not be sent. I have fixed that issue.
For now i would like to resolve this as wont fix and opt for a cleaner, simpler API which aligns with other SDKs. (also i dislike the filepath api so i dont want to add support for that. If somebody else needs it, that can be done in a future iteration)

sdk/identity/Cargo.toml
```diff
@@ -50,3 +53,7 @@ enable_reqwest_rustls = [
 ]
 development = []
 test_e2e = []
+
+[[example]]
+name="client_certificate_credentials"
+required-features = ["openssl"]
```
bmc-msft, line 59: Perhaps make this a named feature such as `client_certificate` rather than `openssl`?

msabansal: Resolved. Made it an optional feature. I was debating about it but initially opted not to.

heaths (changes requested):
LGTM, apart from a crate-specific feature name of `client_certificate` as opposed to an implementation detail like `openssl`.

sdk/identity/examples/client_certificate_credentials/src/main.rs
```diff
@@ -0,0 +1,76 @@
...
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let client_id =
+        ClientId::new(env::var("CLIENT_ID").expect("Missing CLIENT_ID environment variable."));
+    let tenant_id = env::var("TENANT_ID").expect("Missing TENANT_ID environment variable.");
+    let subscription_id =
+        env::var("SUBSCRIPTION_ID").expect("Missing SUBSCRIPTION_ID environment variable.");
+
+    let keyvault_id = env::var("KEYVAULT_ID").expect("Missing KEYVAULT_ID environment variable.");
```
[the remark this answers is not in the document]
msabansal: this file was added by accident. Removed this file

sdk/identity/examples/client_certificate_credentials.rs
```diff
@@ -0,0 +1,76 @@
...
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let client_id =
+        ClientId::new(env::var("CLIENT_ID").expect("Missing CLIENT_ID environment variable."));
+    let tenant_id = env::var("TENANT_ID").expect("Missing TENANT_ID environment variable.");
+    let subscription_id =
+        env::var("SUBSCRIPTION_ID").expect("Missing SUBSCRIPTION_ID environment variable.");
+
+    let keyvault_id = env::var("KEYVAULT_ID").expect("Missing KEYVAULT_ID environment variable.");
```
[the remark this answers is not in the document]
msabansal: Sounds good. Resolved

msabansal:
The error i see is in security_keyvault crate which i have not changed (but imported). Do i need to fix it?

bmc-msft:
@msabansal , pr #719 will address the issues you're experiencing.

sdk/identity/Cargo.toml
heaths: Could you rename the feature to `client_certificate` instead of `openssl`?

sdk/identity/Cargo.toml
```diff
@@ -26,11 +26,18 @@ log = "0.4"
 async-timer = "1.0.0-beta.7"
 thiserror = "1.0"
 async-trait = "0.1"
+openssl = { version = "0.10",  optional=true }
+base64 = "0.13.0"
+uuid = { version = "0.8",  features = ["v4"] }
+
+
```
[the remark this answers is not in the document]
msabansal: fixed

sdk/identity/Cargo.toml
msabansal: done

sdk/identity/src/token_credentials/client_certificate_credentials.rs
heaths: Sounds like this is fine for now.

heaths (approved):
Thanks!

PR #712, opened 2022-04-03
test all service crates in parallel

cataggar:
This makes it possible to check all the services in parallel jobs. The workflow definitions can be generated when the services are updated.


ctaggart:
I was hoping it would publish 1 job at a time. Looks like it is doing 20 concurrently. :/

https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration

cataggar:
crates.io had to problem with the concurrent crate updates. The backoff appears that it was just for new crates. docs.rs put the `azure_mgmt` crates in a "priority: -1, so they get processed when capacity is available. It may take a few days. https://docs.rs/releases/queue

.github/workflows/publish.yml
```diff
@@ -0,0 +1,31 @@
+name: Publish Crate
```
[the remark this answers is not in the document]
cataggar: Since both of the large templates listing all of the services are generated using templates, I don't think there is a benefit to using the reusable actions. I was using them and removed them.

services/autorust/codegen/templates/check-all-services.yml.jinja
```diff
@@ -10,22 +10,20 @@ env:
   CARGO_INCREMENTAL: 0
 
 jobs:
-  test-services:
-    name: Services Tests
-    runs-on: macos-latest
+{% for package in packages %}
```
cataggar, line 13: This template generates the `check-all-services.yml`.

services/autorust/codegen/examples/list_crates.rs
```diff
@@ -55,23 +57,33 @@ fn main() -> Result<()> {
...
+    //         for (i, name) in names.iter().enumerate() {
+    //             println!("{} {}", i + 1, name);
+    //         }
+    //     }
+    // }
+
+    let yml = PublishServicesYml {
+        packages: &names.iter().map(String::as_str).collect(),
+    };
+    yml.create("../../.github/workflows/publish-services.yml")?;
```
cataggar, line 81: generated here

bmc-msft (approved):

bmc-msft (changes requested):
Given these have access to CARG_REGISTRY_TOKEN, this should probably include documentation as to who has access to run these and what the process is to enable giving someone access to run these.

cataggar:
> Given these have access to CARG_REGISTRY_TOKEN, this should probably include documentation as to who has access to run these and what the process is to enable giving someone access to run these.

The publish actions are intended for running in your own fork where only you have access to run the actions.  Do not add the `CARGO_REGISTRY_TOKEN` to the main repository.  Only add it to your repository.

How is that ^^^?

bmc-msft:
Discussion moved to the internal Teams instance for Azure SDK For Rust.

.gitattributes
```diff
@@ -6,3 +6,5 @@ Cargo.lock merge=binary
 * text=auto eol=lf
 services/mgmt/**/src/**/*.rs linguist-generated=true
 services/mgmt/**/Cargo.toml linguist-generated=true
+.github/workflows/check-all-services.yml linguist-generated=true
+.github/workflows/publish-services.yml linguist-generated=true
```
bmc-msft, line 10:
```suggestion
```

bmc-msft (approved):

PR #716, opened 2022-04-07
address some concerns in initial device_update PR

bmc-msft:
This PR adds the following changes:
* moves to enabling the client to support light-weight clones rather than managing lifetimes for the client.
* removes the need for the client to be mutable
* moves to using `azure_identity`'s `AutoRefreshingTokenCredential` rather than by-hand token refresh.
* replaces runtime panics with returning errors
* moves helper functions to return the `DeserializeOwned` trait, rather than strings which get immediately deserialized
* moves many of the errors to use the `azure_core` error types (in prep for moving to the pipeline archtecture)
* moves to using `query_pairs_mut` to add URL params, rather than building query strings by hand
* simplfies creating the mock client used for testing

Things still to do (in follow-up PRs):
* Move to using `azure_svc_deviceupdate` under the hood
* Rather than iteratively fetching all the results at once before returning to the client, move to using a streaming api, akin to [azure_data_tables::requests::QueryEntityBuilder.stream](https://docs.rs/azure_data_tables/latest/azure_data_tables/requests/struct.QueryEntityBuilder.html#method.stream)

### DIFF
--- a/sdk/device_update/Cargo.toml
+++ b/sdk/device_update/Cargo.toml
@@ -15,9 +15,9 @@ edition = "2018"
 [dependencies]
 base64 = "0.13"
 thiserror = "1.0"
-reqwest = { version = "0.11", features = ["json"] }
+reqwest = { version = "0.11", features = ["json"], default_features=false }
 chrono = { version = "0.4", features = ["serde"] }
-const_format = "0.2.13"
+const_format = "0.2"
 serde_json = "1.0"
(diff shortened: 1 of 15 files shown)


rylev (approved):
This looks good. I think we should maybe start moving over to the pipeline architecture ASAP as we might be doing some work that will just go away once the transition is made. I, however, still own a written transition guide...

PR #721, opened 2022-04-09
Add setters for `range` property on `GetPath` builder

roeap:
Fund that we missed adding a setter for the `range` property on the `GetPathBuilder`.


rylev (approved):
Thanks!

Issue #496, opened 2021-11-07
[Tracking Issue] Create/Convert ADLS Gen2 (data_lake) using Pipeline Architecture


thovoll:
This is a tracking issue for creating/converting ADLS Gen2 (data_lake) operations using the pipeline architecture.

### Overview 

Previous to the pipeline architecture, each operation was independently controlled meaning there was no central way to control the way operations were performed. 

The pipeline architecture on the other hand allows users of the SDK to create pipelines through which outgoing requests and incoming responses are processed. The pipeline are composed of "policies" which can both change the outgoing request and the incoming response. Policies are usually either on a "per-request" basis (meaning they are called only once per operation) or "per-retry" (meaning they are applied every time an operation is attempted). 

Examples of policies are the `TelemetryPolicy` which adds various telemetry headers to the request and the retry policy which checks whether an incoming response was successful and if it detects a transient failure, retries the request. 

### Converting to the Architecture

For an example of adding the first operation using the pipeline architecture, take a look at #358 and its follow-ups #451 and #491. 
Once the basic structure is established, adding new operations is easier: #495.

No **existing** operations have been converted yet, but the basic steps are shown in the CosmosDB tracking issue: #290.

## Todos

- [ ] **Files** #610 
    - [X] [`create_file`](https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/create) #358, #451, #491
    - [X] [`create_file_if_not_exists`](https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/create) #495
    - [X] [`rename_file`](https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/create) #495 
    - [X] [`rename_file_if_not_exists`](https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/create) #506 
    - [X] [`delete_file`](https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/delete) #512
    - [ ] [`lease_file`](https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/lease)
    - [X] [`get_file_properties`](https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/get-properties)
    - [x] [`get_file_status`](https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/get-properties)
[...]

PR #722, opened 2022-04-11
Add 'get_status' to datalake file client

rickrain:
This implements `get_status` for the datalake file client, as mentioned in #496.


rickrain:
Ping...  Any thoughts yet on this PR?  It's really pretty lightweight/simple.  Most of the files in this PR are due to the new test (didn't exist before) and all the transaction files needed to pass CI checks.

sdk/storage_datalake/src/operations/path_head.rs
```diff
@@ -98,7 +98,7 @@ impl HeadPathResponse {
             common_storage_response_headers: (&headers).try_into()?,
             etag: etag_from_headers(&headers)?,
             last_modified: last_modified_from_headers(&headers)?,
-            properties: (&headers).try_into()?,
+            properties: (&headers).try_into().ok(),
```
cataggar, line 101: I don't understand this change. Can't we try it without? I'll give you a PR.

cataggar, line 101: I don't understand this change. Can't we try it without? I'll give you a PR.

cataggar: https://github.com/rickrain/azure-sdk-for-rust/pull/2

rickrain: @cataggar, just checking that you saw my comments [here](https://github.com/rickrain/azure-sdk-for-rust/pull/2/commits/ac6fe5f90d09fff444d72c07de7cc28d190947b8#r859143519) and [here](https://github.com/rickrain/azure-sdk-for-rust/pull/2/commits/ac6fe5f90d09fff444d72c07de7cc28d190947b8#r859145288).  I updated my PR to to [remove unused variables](https://github.com/Azure/azure-sdk-for-rust/pull/722/commits/924081078b808e92c5243f8271784c33cb85cee4), but the other change is necessary. Let me know if there's anything else I can do to move this forward.

sdk/storage_datalake/src/operations/path_head.rs
```diff
@@ -87,18 +87,18 @@ pub struct HeadPathResponse {
...
 
         Ok(Self {
-            common_storage_response_headers: (&headers).try_into()?,
-            etag: etag_from_headers(&headers)?,
-            last_modified: last_modified_from_headers(&headers)?,
-            properties: (&headers).try_into()?,
+            common_storage_response_headers: headers.try_into()?,
+            etag: etag_from_headers(headers)?,
+            last_modified: last_modified_from_headers(headers)?,
+            properties: headers.try_into().ok(),
```
cataggar, line 101: Thanks for the details on why it is Option<Properties> now, but why use `.ok()` here instead of `?`?

rickrain: Ahh, that's because I need to convert `Result<T, E>` to `Option<T>` since `properties` is now an `Option`.  However, I just realized a problem with this whereby the `E` get's discarded silently in this implementation. A better implementation would be to only discard the error if it is `HeaderNotFound` since this is now an expected condition, and propagate all other errors that may come out of `try_into()`.  So, basically this:

```rust
properties: match headers.try_into() {
    Ok(p) => Some(p),
    // 'x-ms-properties' will not be returned in some cases, such as when the request 'action' is set to 'getStatus'
    Err(crate::Error::HeaderNotFound(_)) => None,
    // Propogate all other errors
    Err(e) => return Err(e),
},
```

I just tested this locally and it works as expected.  If you're OK with this implementation, I'll push it into my PR.

rickrain: @cataggar, just heads up that I went ahead and applied the change above. Let me know if this resolves your concerns.

cataggar: Sorry @rickrain, I'm lacking time to dig into this. An error case shouldn't be a normal workflow. If the header is optional for an operation and it is not found, it should return None. It should return an error and then map that to None.

rickrain: Ok.  So, you would rather me write the same logic that `try_into()` already does so I can check if the header is there first, rather than letting `try_into()` just return the error and respond accordingly. This seems to be very opinionated, but ok.  In other languages, where _exceptions_ are expensive, I would agree with this opinion. But, I'm not sure I'm in agreement here given how errors are handled in Rust.

cataggar: You can use get_option_str_from_headers https://github.com/rickrain/azure-sdk-for-rust/pull/3 .

rickrain: I had no idea that function existed. Thank you for steering me in the right direction!  Had a different implementation I was about to commit but like yours better given it's using an existing function from the SDK.

Issue #741, opened 2022-05-04
generated crates need to support long running operations


bmc-msft:
The generated crates need to support long-running operations.

Ref: https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#long-running-operations-with-status-monitor

cataggar:
More info in #174.

cataggar:
Fixed by #1370

Issue #446, opened 2021-10-25
operation response type with headers


ctaggart:
- #746
- #1034
- #1037
- ~remove special handling for single responses~
-  adjust non-single responses to be like single responses. See analysis in  https://github.com/ctaggart/azure-sdk-for-rust/pull/17
- return new type `operation::Response(azure_core::Response)`
  - typed `headers()` `body()` functions
  - allow getting the `azure_core::Response`. may be implement `AsRef<azure_core::Response>`

---

https://github.com/Azure/azure-sdk-for-rust/blob/2eadbde60c2653a9dc172552ca617450679fd512/services/svc/blobstorage/src/package_2020_10/operations.rs#L407-L412

specification\storage\data-plane\Microsoft.BlobStorage\preview\2020-10-02\blob.json

``` json

    "/?restype=account&comp=properties": {
      "get": {
        "tags": [
          "service"
        ],
        "operationId": "Service_GetAccountInfo",
        "description": "Returns the sku name and account kind ",
        "parameters": [
          {
            "$ref": "#/parameters/ApiVersionParameter"
          }
        ],
        "responses": {
          "200": {
            "description": "Success (OK)",
            "headers": {
              "x-ms-client-request-id": {
                "x-ms-client-name": "ClientRequestId",
                "type": "string",
                "description": "If a client request id header is sent in the request, this header will be present in the response with the same value."
              },
              "x-ms-request-id": {
                "x-ms-client-name": "RequestId",
                "type": "string",
                "description": "This header uniquely identifies the request that was made and can be used for troubleshooting the request."
              },
              "x-ms-version": {
                "x-ms-client-name": "Version",
                "type": "string",
                "description": "Indicates the version of the Blob service used to execute the request. This header is returned for requests made against version 2009-09-19 and above."
              },
              "Date": {
                "type": "string",
                "format": "date-time-rfc1123",
                "description": "UTC date/time value generated by the service that indicates the time at which the response was initiated"
              },
              "x-ms-sku-name": {
                "x-ms-client-name": "SkuName",
                "type": "string",
                "enum": [
                  "Standard_LRS",
                  "Standard_GRS",
                  "Standard_RAGRS",
                  "Standard_ZRS",
                  "Premium_LRS"
                ],
                "x-ms-enum": {
                  "name": "SkuName",
                  "modelAsString": false
                },
                "description": "Identifies the sku name of the account"
              },
              "x-ms-account-kind": {
[...]

PR #742, opened 2022-05-04
initial work to support x-ms-pageable for generated specs

bmc-msft:
API endpoints that define x-ms-pageable supports generating a `Pageable` stream (via `into_stream`) instead of an `into_future` builder.

At the moment, the generator does not support endpoints that use the HTTP header `x-ms-continuation` (seen as not providing a next_link_name in the x-ms-pageable spec).  For functions that this does not work, there is now a doc comment added such that users are informed that the generated API does not page.

In order to fully flesh this out, we need to support including header values in the response object due to how Pageable works.   See #446 for more information.

As this is similar to LRO support, endpoints that would typically be polled but do not yet poll, a doc comment is also added such that the users are informed that the polling does not occur. 
 Adding support for this is tracked as #741.


cataggar (approved):
This is fantastic! I reviewed the commits after a `gh pr checkout 742`.

PR #744, opened 2022-05-04
prep release for idiomatic crates

bmc-msft:
All of the idiomatic crates had their dependencies (primarily `azure_core`) bumped in source but not released.

In the future, it would be good if tags were added for each release such that generating changelogs between releases was straight forward.


bmc-msft:
Note, rather than bumping the `dev-dependencies` in all of the generated crates, I will be opening a PR to only use the path, which will reduce the burden on developers releasing dependencies.

cataggar (approved):

PR #746, opened 2022-05-05
sketch out response type for headers

cataggar:
This sketches out what needs to be done to support response types for headers #446. It is a draft and a start of the design. To be continued.

It also highlights a few things that need to be done to make `azure_svc_blobstoge` work.


cataggar:
Replaced by #855

PR #747, opened 2022-05-06
add autorust.toml with tags_allow to limit tags

cataggar:
This allows specifying an allow list of tags to generate for a crate. The configuration for the package is an `autorust.toml` file in the destination directory. We can extend it and move package specific configuration there.


services/mgmt/network/autorust.toml
```diff
@@ -0,0 +1 @@
+tags_allow = ["package-2021-08", "package-2021-05"]
```
cataggar, line 1: This is limiting the tags for azure_mgmt_network to two of the most recent stable tags.

services/autorust/codegen/src/autorust_toml.rs
```diff
@@ -0,0 +1,57 @@
...
+        Ok(PackageConfig::default())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_tags_allow() -> Result<(), Error> {
```
heaths, line 49: Would be good to cover the case where no autorest.toml exists and the passed tags are returned back. That'll happen far more frequently.

services/mgmt/network/autorust.toml
heaths: So no root section is necessary? Should we have one to avoid future issues, or likely not a problem?

FWIW, our autorest tooling makes heavy use (required, really) of configuration so I definitely could see adding more in the future here...assuming we don't eventually adopt autorest instead.

heaths (approved):
Awesome! Just a couple nits/thoughts.

PR #751, opened 2022-05-16
datetime from azure cli token is in the local timezone

yvespp:
While working on auto refresh I noticed that the token expiry from azure cli is in the local timezone but is interpreted as UTC in code. This causes the token not being refreshed in my timezone. (My time: 12:00 GMT+2, toke `expiresOn` time 13:20, token expiry in UTC 15:20).
This can be verified by running `az account get-access-token --output json` and comparing `exp` of the jwt with `expiresOn` from the cli.

### DIFF
--- a/sdk/identity/src/token_credentials/cli_credentials.rs
+++ b/sdk/identity/src/token_credentials/cli_credentials.rs
@@ -8,7 +8,7 @@ use std::str;
 use std::{io::ErrorKind, str::Utf8Error};
 
 mod az_cli_date_format {
-    use chrono::{DateTime, TimeZone, Utc};
+    use chrono::{DateTime, Local, TimeZone, Utc};
     use serde::{self, Deserialize, Deserializer};
 
     const FORMAT: &str = "%Y-%m-%d %H:%M:%S.%6f";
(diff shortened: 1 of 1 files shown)


bmc-msft:
In my testing, I saw a similar difference with `expiresOn` as indicated by @yvespp .  I went to check how the CLI implements this to make sure we're not relying on a local configuration regarding time zones.   That doesn't appear to be the case.

However, in my investigation, I came across the following from the [SDK](https://github.com/Azure/azure-cli/blob/a7d5d6d53f1b8ecd8a28752ea06e3f5499f20856/src/azure-cli-core/azure/cli/core/_profile.py#L385-L394)

```python
        # Convert epoch int 'expires_on' to datetime string 'expiresOn' for backward compatibility
        # WARNING: expiresOn is deprecated and will be removed in future release.
        import datetime
        expiresOn = datetime.datetime.fromtimestamp(sdk_token.expires_on).strftime("%Y-%m-%d %H:%M:%S.%f")


        token_entry = {
            'accessToken': sdk_token.token,
            'expires_on': sdk_token.expires_on,  # epoch int, like 1605238724
            'expiresOn': expiresOn  # datetime string, like "2020-11-12 13:50:47.114324"
        }
```

In the future, we may want to consider parsing the JWT token itself, which includes more details as to the expiry state (including a not-before and expires-on timestamp).

sdk/identity/src/token_credentials/cli_credentials.rs
```diff
@@ -18,8 +18,11 @@ mod az_cli_date_format {
         D: Deserializer<'de>,
     {
         let s = String::deserialize(deserializer)?;
-        Utc.datetime_from_str(&s, FORMAT)
-            .map_err(serde::de::Error::custom)
+        // expiresOn from azure cli uses the local timezone and needs to be converted to UTC
+        let local_datetime = Local
+            .datetime_from_str(&s, FORMAT)
+            .map_err(serde::de::Error::custom)?;
+        return Ok(local_datetime.with_timezone(&Utc));
```
bmc-msft, line 25: To address the issue with clippy:

```suggestion
        Ok(local_datetime.with_timezone(&Utc))
```

bmc-msft (approved):

Issue #739, opened 2022-05-04
StorageAccountClient bearer token refresh


yvespp:
While looking into https://github.com/vectordotdev/vector/issues/8667 I noticed that the client doesn't refresh the bearer token and will stop working once it expires.

Is there a way to refresh the token automatically?

I've seen that there is [AutoRefreshingTokenCredential](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/identity/src/token_credentials/auto_refreshing_credentials.rs) but it can't be passed to the StorageAccountClient at the moment.

cataggar:
@yvespp, new crates were published on Friday. AutoRefreshingTokenCredential is in azure_identity 0.2.0.

yvespp:
@cataggar ok, but I don't see a way to use AutoRefreshingTokenCredential with the StorageAccountClient. The client takes the token as a string https://github.com/Azure/azure-sdk-for-rust/blob/beb3375631cfa31849775682ad82d39394e1dfee/sdk/storage/src/core/clients/storage_account_client.rs#L37 and not as a TokenCredential?

PR #757, opened 2022-05-19
add AutoRefreshingTokenCredential support to StorageAccountClient

yvespp:
For https://github.com/vectordotdev/vector/issues/8667 a way to refresh the managed identity token is needed. Vector is a log agent that can forward logs to azure blob storage. See also https://github.com/Azure/azure-sdk-for-rust/issues/739.

I add a way to pass TokenCredential to the StorageAccountClient so AutoRefreshingTokenCredential can be passed to it and the client can refresh the token automatically. I tested it locally using the added example.

I don't what's preferred here and can easily drop the second commit:
* In the first commit I used arc for TokenCredential and implemented Debug, PartialEq and Eq manually.
* In the second commit I just used box for TokenCredential and removed currently unused traits Clone, PartialEq and Eq.

### DIFF
--- a/sdk/storage/src/core/clients/storage_account_client.rs
+++ b/sdk/storage/src/core/clients/storage_account_client.rs
@@ -6,6 +6,7 @@ use crate::{
         AccountSharedAccessSignatureBuilder, ClientAccountSharedAccessSignature,
     },
 };
+use azure_core::auth::TokenCredential;
 use azure_core::headers::*;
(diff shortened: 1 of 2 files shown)


bmc-msft (approved):
LGTM. 

Thanks for the contribution!

With this adding support for the `azure-identity` crate, we can probably remove the bearer token methods, as `azure-identity` is a significantly better mechanism for managing bearer tokens.

PR #760, opened 2022-05-20
remove defunct enable_hyper feature

### LINKED ISSUE #688: Attempting to enable hyper causes azure_core to fail to compile
Hello, I'm attempting to use these crates to access Azure Blob Storage. Because this is a new feature for an existing project, the code already uses Hyper under the hood, so I was attempting to replicate that in the use of these crates as well. However, when I attempt to enable hyper on the `azure_core` crate, it fails to compile with a large number of error messages. This is true even if I use `default-features = false` to attempt to disable the default reqwest support.

Among the issues are misuse of the hyper_rustls crate (`with_native_roots` is a function on the HttpsConnectorBuilder in 0.23, not the HttpsConnector), there is no `new_unexpected_status_code` in HttpError, and more.

Is the use of `hyper` currently supported with these crates, and if not, is there anything that can be done to help get support for it off the ground? Thanks!

cataggar:
This is a subset of #755. It hasn't been supported for months. A user could implement the `HttpClient` trait to pass in their own version. Fix #688.

### DIFF
--- a/sdk/core/Cargo.toml
+++ b/sdk/core/Cargo.toml
@@ -20,7 +20,6 @@ chrono = "0.4"
 dyn-clone = "1.0"
 futures = "0.3"
 http = "0.2"
-hyper = { version = "0.14", optional = true }
 hyper-rustls = { version = "0.23", optional = true }
 log = "0.4"
 oauth2 = { version = "4.0", default-features = false }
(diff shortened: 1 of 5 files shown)


rylev (approved):
I agree with having only one built-in solution (and requiring the user implement their own if they don't like the default), but I'm unsure that `reqwest` is the right default. Since `reqwest` is built on top of `hyper`, it seems reasonable to use the lowest common denominator as the default. Either way, I'm fine with merging this for now.

PR #761, opened 2022-05-21
limit tags to 5

### LINKED ISSUE #740: generated crates should include the latest version of each module by default
Consider `azure_mgmt_resources`.   In go, the modules are split up by module name such as:

```
❯ grep ^module-name azure-rest-api-specs/specification/resources/resource-manager/readme.go.md
module-name: sdk/resourcemanager/resources/armresources
module-name: sdk/resourcemanager/resources/armdeploymentscripts
module-name: sdk/resourcemanager/resources/armfeatures
module-name: sdk/resourcemanager/resources/armlinks
module-name: sdk/resourcemanager/resources/armlocks
module-name: sdk/resourcemanager/resources/armmanagedapplications
module-name: sdk/resourcemanager/resources/armpolicy
module-name: sdk/resourcemanager/resources/armsubscriptions
module-name: sdk/resourcemanager/resources/armtemplatespecs
module-name: sdk/resourcemanager/resources/armchanges
```

Python does something similar by dividing them up by names:Ace
```
❯ grep ^namespace azure-rest-api-specs/specification/resources/resource-manager/readme.python.md |head -n 10
namespace: azure.mgmt.resource.changes.v2022_05_01
namespace: azure.mgmt.resource.privatelinks.v2020_05_01
namespace: azure.mgmt.resource.features.v2021_07_01
namespace: azure.mgmt.resource.features.v2015_12_01
namespace: azure.mgmt.resource.links.v2016_09_01
namespace: azure.mgmt.resource.locks.v2016_09_01
namespace: azure.mgmt.resource.locks.v2015_01_01
namespace: azure.mgmt.resource.managedapplications
namespace: azure.mgmt.resource.policy.v2021_06_01
namespace: azure.mgmt.resource.policy.v2020_09_01
```

It would be useful to have the latest version of each of the module to be our default set when using the module.

cataggar:
Fix #740. The primary change is to limit the number of generated tags for each service to 5 by default. The limit can be configured in `autorust.toml`. A `[tags]` section was added along with additional filtering options.


services/svc/servicefabric/Cargo.toml
```diff
@@ -37,5 +37,3 @@ no-default-tag = []
 "7_1" = []
 "7_2" = []
 "8_0" = []
-"8_1" = []
-"8_2" = []
```
cataggar, line 41: File order doesn't work for a few services. A `sort` option would be useful, but I'm not sure it would be best for all. Will try.

services/mgmt/operationalinsights/Cargo.toml
```diff
@@ -25,21 +25,15 @@ tokio = { version = "1.0", features = ["macros"] }
 env_logger = "0.9"
 
 [package.metadata.docs.rs]
-features = ["no-default-tag", "package-2015-11-preview", "package-2019-08-preview", "package-2019-09-preview", "package-2020-03-preview"]
+features = ["no-default-tag", "package-2021-12-01-preview", "package-2021-06", "package-2020-10", "package-2020-08"]
 
 [features]
-default = ["package-2015-03", "enable_reqwest"]
+default = ["package-2022-02", "enable_reqwest"]
```
cataggar, line 31: Duplicate SystemData. It is a mistake in the spec.

services/autorust/codegen/examples/gen_workspace.rs
```diff
@@ -0,0 +1,25 @@
...
+    jinja::{CargoToml, CheckAllServicesYml},
+};
+pub type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
+
+fn main() -> Result<()> {
+    let dirs = list_dirs()?;
+    let dirs: Vec<String> = dirs.iter().map(|dir| dir.as_str().replace("\\", "/").replace("../", "")).collect();
+
+    let yml = CargoToml { dirs };
+    yml.create("../Cargo.toml")?;
```
bmc-msft, line 14: Is there a benefit of writing out all of the crate names, rather than using wildcards?  I imagine this will just cause churn in git without providing value.

services/autorust/codegen/src/crates.rs
```diff
@@ -0,0 +1,72 @@
...
+
+pub fn has_version(name: &str, version: &str) -> Result<bool> {
+    Ok(get_versions(name)?.iter().any(|v| v.vers.as_str() == version))
+}
+
+/// Gets all the versions for a given crate
+/// Expects https://github.com/rust-lang/crates.io-index to be cloned as a sibling directory.
+fn get_versions(crate_name: &str) -> Result<Vec<CrateVersion>> {
+    // all of these crates begin with "azure".
+    let path = format!("../../../crates.io-index/az/ur/{}", crate_name);
```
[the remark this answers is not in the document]
bmc-msft: Nevermind, I figured out how this is used and it's reasonable.  Though, we may want to do a shallow checkout at runtime rather than rely on this being up-to-date by the user.

bmc-msft:
This is a rather positive improvement from the current `gen_svc` and `gen_mgmt` examples.

bmc-msft:
There are a handful of the autorust.toml files that exclude versions without explaining why, in one such example purview now excludes a tag that as far as I can tell isn't excluded in main.  These should all have comments as to why limits are included.

services/autorust/codegen/src/crates.rs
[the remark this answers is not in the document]
cataggar: This utility function is only used in the examples where I am auditing published crates. I prefer sibling projects over git submodules.

bmc-msft:
Do you have any thoughts on expanding on your autorust.toml work to include the other components currently defined in source, such as the workarounds for `FIX_CASE_PROPERTIES`, `BOX_PROPERTIES`, and `INVALID_TYPE_WORKAROUND`?

services/autorust/codegen/src/crates.rs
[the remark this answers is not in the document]
bmc-msft: I'm not suggesting submodules.  I'm suggesting something to the effect of creating a temp dir, doing a `git checkout --depth 1`, getting the data you need, then cleaning the temp dir.

Both submodules and sibling projects rely on manual user interaction, which we should avoid.

cataggar:
> There are a handful of the autorust.toml files that exclude versions without explaining why, in one such example purview now excludes a tag that as far as I can tell isn't excluded in main. These should all have comments as to why limits are included.

I agree. It is only a few of them. This filter is useful for reviewing:

image

services/mgmt/databoxedge/autorust.toml
```diff
@@ -0,0 +1,5 @@
+[tags]
+deny=[
+  "package-2022-04-01-preview", # duplicate SystemData
+  "profile-hybrid-2020-09-01",
```
bmc-msft, line 4: Why was the second entry denied?

services/mgmt/dataprotection/autorust.toml
```diff
@@ -0,0 +1,2 @@
+[tags]
+deny_contains_preview = true
```
bmc-msft, line 2: please add explanation

services/mgmt/servicefabricmanagedclusters/autorust.toml
```diff
@@ -0,0 +1,2 @@
+[tags]
+limit = 0
```
bmc-msft, line 2: please add explanation

services/svc/purview/autorust.toml
```diff
@@ -0,0 +1,4 @@
+[tags]
+deny = [
+    "package-2021-05-01-preview"
```
bmc-msft, line 3: please add explanation.  This isn't excluded in `main`

cataggar:
> Do you have any thoughts on expanding on your autorust.toml work to include the other components currently defined in source, such as the workarounds for FIX_CASE_PROPERTIES, BOX_PROPERTIES, and INVALID_TYPE_WORKAROUND?

Yes, I was thinking we could migrate them directly over doing something like this:

``` toml
[properties]
box = [
    ["../../../azure-rest-api-specs/specification/databox/resource-manager/Microsoft.DataBox/stable/2022-03-01/databox.json", "transferAllDetails", "include"]
  ]
fix_case =
invalid_type = 
```

cataggar:
A run of [Check All Features for Services](https://github.com/cataggar/azure-sdk-for-rust/actions/runs/2368594321) passed last night. An [earlier run,](https://github.com/cataggar/azure-sdk-for-rust/actions/runs/2366705450) showed these four failing:

azure_mgmt_databoxedge
azure_mgmt_dataprotection
azure_svc_mixedreality
azure_svc_purview

I suspect, `main` also currently has issues for those four.

Basic explanations have been added where requested in the autorust.toml files. They can be refined in future pull requests. This should be good to merge. I started [another run](https://github.com/cataggar/azure-sdk-for-rust/actions/runs/2372370833) for the latest commit here.

services/mgmt/containerservice/README.md
```diff
@@ -6,70 +6,12 @@ https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containers
 
 To get started with these generated service crates, see the [examples](https://github.com/Azure/azure-sdk-for-rust/blob/main/services/README.md#examples).
 
-The default tag is `package-2022-03`.
```
cataggar, line 9: There will probably be a few places where sorting alphabetically doesn't work. This is one. I don't think we need to fix every case in this PR though.

services/mgmt/costmanagement/README.md
```diff
@@ -6,21 +6,12 @@ https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cost-manag
 
 To get started with these generated service crates, see the [examples](https://github.com/Azure/azure-sdk-for-rust/blob/main/services/README.md#examples).
 
-The default tag is `package-2021-10`.
```
cataggar, line 9: This is another case where sorting does not work.

services/mgmt/keyvault/Cargo.toml
```diff
@@ -25,23 +25,15 @@ tokio = { version = "1.0", features = ["macros"] }
...
+all-features = true
 
 [features]
 default = ["package-preview-2021-11", "enable_reqwest"]
 enable_reqwest = ["azure_core/enable_reqwest"]
 enable_reqwest_rustls = ["azure_core/enable_reqwest_rustls"]
 no-default-tag = []
+"profile-hybrid-2020-09-01" = []
 "package-preview-2021-11" = []
-"package-2021-10" = []
```
cataggar, line 36: This should probably be the default. It is the most recent stable. It may be useful to have a `limit_preview = 2` option so that there are stable tags.

services/mgmt/kusto/Cargo.toml
```diff
@@ -25,25 +25,13 @@ tokio = { version = "1.0", features = ["macros"] }
 env_logger = "0.9"
 
 [package.metadata.docs.rs]
-features = ["no-default-tag", "package-2021-08-27", "package-2021-01", "package-2020-09-18", "package-2020-06-14"]
+all-features = true
 
 [features]
-default = ["package-2022-02", "enable_reqwest"]
```
cataggar, line 31: Will probably need `deny_contains = ["schema"]` here.

services/mgmt/machinelearning/Cargo.toml
```diff
@@ -25,15 +25,15 @@ tokio = { version = "1.0", features = ["macros"] }
 env_logger = "0.9"
 
 [package.metadata.docs.rs]
-features = ["no-default-tag", "package-commitmentPlans-2016-05-preview", "package-workspaces-2016-04", "package-workspaces-2019-10", "package-webservices-2016-05-preview"]
+all-features = true
 
 [features]
```
cataggar, line 30: An example where both file order of tags and alphabetical order of tags fail.

PR #762, opened 2022-05-22
iot_hub: Add support for configurations

corollaries:
1. Add support for applying a configuration on an edge device.
2. Add support for: Get, create, update, or delete a configuration.
3. Some minor fixes to documentation and typos. 

The deployment manifest that is required for applying a configuration on an edge device is quite complex. I considered creating a builder for this but this would lead to a lot of code that might need changes again once this deployment manifest gets updated (and it does every once in a while). Therefore just went with supplying the raw JSON as other Microsoft SDK's follow a similar approach.


bmc-msft (approved):
While we should not take such a strong dependency on consuming `Serde::Value`, this is a reasonable starting point for these.

bmc-msft:
Thanks for the contribution!

PR #765, opened 2022-05-25
use azure_core errors in services (no gen)

cataggar:
This replaces the errors in the services. The generated code is included in #766.

### DIFF
--- a/services/autorust/codegen/src/codegen_operations.rs
+++ b/services/autorust/codegen/src/codegen_operations.rs
@@ -173,15 +173,6 @@ pub fn create_operations(cg: &CodeGen) -> Result<TokenStream, Error> {
         });
     }
 
-    file.extend(quote! {
-        #[non_exhaustive]
-        #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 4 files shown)


services/autorust/codegen/src/codegen_operations.rs
```diff
@@ -664,90 +625,29 @@ fn create_operation_code(cg: &CodeGen, operation: &WebOperationGen) -> Result<Op
...
-                                Err(Error::#response_type_name{})
-                            }
-                        });
-                    }
-                }
-            }
-            autorust_openapi::StatusCode::Default => {}
+    match_status.extend(quote! {
+        status_code => {
+            Err(azure_core::error::Error::from(azure_core::error::ErrorKind::HttpResponse { status: status_code.as_u16(), error_code: None }))
```
cataggar, line 630: It would be nice to have a `.into_error()` for an `ErrorKind`. I'm punting on `error_code` here to a future PR. I think core needs to slights adjustments.

bmc-msft (approved):
Moving to use core's errors is a positive.

PR #770, opened 2022-05-27
Implemented `messaging_servicebus` crate

danbugs:
This PR is addressing and hoping to close issue #255 — That is, I've updated the crate to use the new Azure core's `HttpClient`, as requested. Plus, coming from MindFlavor's AzureSDKForRust, I had to make some changes to error handling, so I did that too following the pattern used by other crates under sdk/.

Signed-off-by: Dan Chiarlone <dchiarlone@microsoft.com>


sdk/messaging_servicebus/Cargo.toml
```diff
@@ -15,13 +15,17 @@ categories = ["api-bindings"]
 edition = "2021"
 
 [dependencies]
-azure_core = { path = "../core", version = "0.2" }
+azure_core = "0.2"
```
cataggar, line 18: Why?

cataggar (changes requested):
Thanks for this. Sorry, we moved the goal post. Instead of `thiserror`, can you please use `azure_core::error`. See #771.

sdk/messaging_servicebus/Cargo.toml
danbugs: I was using this crate in a local project to test it and, with `azure_core = { path = "../core", version = "0.2" }`, the compiler screamed at me that I was using two different `azure_core` versions (i.e., locally I had `azure_core = "0.2"`). After I changed it like so, it all worked 🤷

danbugs:
@cataggar ~ Thank you for the review! Like I mentioned in an earlier comment, I was testing this project locally, and, while doing so, I realized none of the functions related to getting messages (i.e., `receive_and_delete`, `peek_lock`, or `peek_lock_full`) work — no wonder there's no test coverage for anything but `send_event` lol

Looking further into it, I realized that's because, rather than asking for a simple queue, we are sneakily asking for an event hub, which doesn't really have that sort of functionality as can be seen per our REST API reference:
image
> Note: [source](https://docs.microsoft.com/en-us/rest/api/eventhub/)

Considering event hubs are built on top of service buses, the user could just pass in a queue name instead of an event hub name and that would work fine-ish, but I figured that's just bad UX for a crate called messaging_**servicebus**.

I have an upcoming PR where I'm re-organizing things:
- renamed stuff (e.g., `event_hub` to `queue`).
- because we don't need different functions for every single type of request we want to prepare (e.g., `send_event_prepare`, and so on), so I'm unifying all those into a single `prepare_request` function — similar to how other crates under sdk/ do it.
- considering `delete_message`, `renew_lock`, and `unlock_message` only work when we have a `lock-id`, which is only ever returned in the `PrepareLockResponse` object, I made them member functions to that struct instead.
- added documentation comments and dev comments.
- removed the `duration` parameter that was passed to pretty much every function of our crate and hid that detail from the user — mostly because I think it's confusing to have to pass a SAS duration when I want to, for example, just get a message. I just made it a set (and safe) expiry duration of 1 hour.
- updated the existing example (we might want to add more later).
- added test coverage for everything. 

Also, per your request, I'll drop the `thiserror` dependency and review the `errors` module (some stuff there might not be being used).

Considering this was a _lot_ of work, what are your thoughts on making this a `0.3`?

danbugs:
Another thing I'd like to point out is that, as soon as I push these latest changes, the tests from `messaging_servicebus` will probably fail because I changed one of the required environment variables from `AZURE_EVENT_HUB_NAME` to `AZURE_QUEUE_NAME`.

sdk/messaging_servicebus/Cargo.toml
rylev: This should be changed back. If you also want to use `azure_core` as a direct dependency locally, you should write `azure_core = { path = "../YOUR/AZURE_CORE/CHECKOUT", version = "0.2" }` .

sdk/messaging_servicebus/src/lib.rs
```diff
@@ -1,9 +1,7 @@
 //! Azure service bus crate for the unofficial Microsoft Azure SDK for Rust. This crate is part of a collection of crates: for more information please refer to [https://github.com/azure/azure-sdk-for-rust](https://github.com/azure/azure-sdk-for-rust).
 #![recursion_limit = "128"]
 
-// #[macro_use]
-// extern crate log;
+extern crate log;
```
rylev, line 4: This line can be removed `extern crate` declarations are not necessary post Rust 2015

sdk/messaging_servicebus/src/service_bus/client.rs
```diff
@@ -0,0 +1,132 @@
...
+use ring::hmac::Key;
+
+use azure_core::{
+    error::{Error, ErrorKind, ResultExt},
+    HttpClient,
+};
+
+/// Client object that allows interaction with the ServiceBus API
+pub struct Client {
+    http_client: Arc<dyn HttpClient>,
```
rylev, line 17: This is fine, but do note that we will want to move this crate to using the pipeline architecture (as can be seen [here](https://github.com/Azure/azure-sdk-for-rust/blob/a76ebf194eddf0a1846ffec77078359a5735f5a4/sdk/data_cosmos/src/clients/cosmos.rs#L69-L76)) ASAP. However, we don't need to block this PR on that.

sdk/messaging_servicebus/src/service_bus/client.rs
```diff
@@ -0,0 +1,132 @@
...
+                &self.queue,
+                &self.policy_name,
+                &self.signing_key,
+            )
+            .await?
+            .into_body(),
+        )
+        .context(
+            ErrorKind::DataConversion,
+            "Failed to convert body bytes to UTF8",
```
rylev, line 78: We should move this error handling to a common method so that we don't have to repeat the error handling in every method. 

Also nit: error messages should all start with lowercase letters.

sdk/messaging_servicebus/src/service_bus/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+};
+
+mod client;
+pub use self::client::Client;
+
+/// Default duration for the SAS token in days — We might want to make this configurable at some point
+const DEFAULT_SAS_DURATION: i64 = 1;
+
+/// Prepares an HTTP request
+#[inline]
```
rylev, line 26: I don't think this should be inlined (it's called pretty rarely). If the compiler thinks it better to inline, we can let it do so.

sdk/messaging_servicebus/src/service_bus/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+            request = request.header(CONTENT_LENGTH, 0); // added to avoid truncation errors
+            request.body(azure_core::EMPTY_BODY)
+        }
+    }?;
+
+    Ok(ret)
+}
+
+/// Generates a SAS signature
+#[inline]
```
rylev, line 64: Same as above

sdk/messaging_servicebus/tests/service_bus.rs
```diff
@@ -0,0 +1,101 @@
...
+async fn send_message_test() {
+    let mut client = create_client().unwrap();
+    client
+        .send_message("hello, world!")
+        .await
+        .expect("Failed to send message");
+}
+
+#[tokio::test]
+async fn receive_and_delete_message_test() {
```
rylev, line 17: Does this expect `send_message_test` to run first? it seems like we want these tests to be one test instead of many.

sdk/messaging_servicebus/Cargo.toml
danbugs: I don't want to use `azure_core` as a local dependency on my project, I want to use `azure_core = "0.2"`. With that said, why do I get errors for using two different `azure_core` versions if both `messaging_servicebus` and my local project use version `0.2`? That is, regardless of it being local or not...? 

For reference, this is the error I get:
```rs
error[E0308]: mismatched types
  --> crates/mq-azure-servicebus/src/lib.rs:33:13
   |
33 |             http_client,
   |             ^^^^^^^^^^^ expected trait `azure_core::http_client::HttpClient`, found trait `HttpClient`
   |
   = note: expected struct `Arc<(dyn azure_core::http_client::HttpClient + 'static)>`
              found struct `Arc<dyn HttpClient>`
   = note: perhaps two different versions of crate `azure_core` are being used?

For more information about this error, try `rustc --explain E0308
```

This is w/ my local project depending on `azure_core = "=0.2"`, and `messaging_servicebus` depending on `azure_core =  { path = "../core", version = "=0.2" }`. If I change `messaging_servicebus` to depend on `azure_core = "0.2"`, the problem is fixed. 

P.S. I'm getting the `HttpClient` the exact same way I do in the tests: `let http_client = azure_core::new_http_client();
`

danbugs: In any case~ done ✅

sdk/messaging_servicebus/src/lib.rs
danbugs: done ✅

sdk/messaging_servicebus/src/service_bus/client.rs
danbugs: Sounds good~ I'll leave this for now, so we can at least get a working `messaging_servicebus` crate out!

sdk/messaging_servicebus/src/service_bus/client.rs
danbugs: done ✅

sdk/messaging_servicebus/src/service_bus/mod.rs
danbugs: done ✅

sdk/messaging_servicebus/tests/service_bus.rs
danbugs: That's how I had done it initially, I'll revert back to that ~ done ✅

sdk/messaging_servicebus/src/service_bus/mod.rs
danbugs: done ✅

sdk/messaging_servicebus/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "azure_messaging_servicebus"
-version = "0.2.0"
+version = "0.3.0"
```
cataggar, line 3: You can revert this back to `0.2.0`. There is only a `0.1.0` published. https://crates.io/crates/azure_messaging_servicebus/versions

danbugs: done ✅

sdk/messaging_servicebus/Cargo.toml
```diff
@@ -22,6 +22,10 @@ log = "0.4"
 url = "2.2"
 hmac = "0.12"
 sha2 = "0.10"
+ring = "0.16"
+http = "0.2"
+bytes = "1.0"
+thiserror = "1.0"
```
cataggar, line 28: Please remove `thiserror`.

sdk/messaging_servicebus/src/utils.rs
```diff
@@ -0,0 +1,35 @@
...
+    // add timeout, if given
+    if let Some(t) = lock_expiry {
+        url.query_pairs_mut()
+            .append_pair("timeout", &t.num_seconds().to_string());
+    };
+
+    Ok(url)
+}
+
+pub fn body_bytes_to_utf8(bytes: &[u8]) -> Result<String, Error> {
```
cataggar, line 28: I haven't looked at this library enough to know if a conversion to a `String` for a body is the right thing or if it should be passed around as `Bytes`.

danbugs: I followed what the crate initially did there. Plus, as a user of the crate, I'd want `String` rather than having to add a dependency for `Bytes` and handle that.

sdk/messaging_servicebus/Cargo.toml
danbugs: done ✅

sdk/messaging_servicebus/Cargo.toml
rylev: The note `perhaps two different versions of crate `azure_core` are being used?` points to exactly what is happening.

You are using two versions of `azure_core` even though they are both labeled as version 0.2. One is the local copy (from the local `messaging_servicebus`) and one is from crates.io (from your direct dependency). The correct way to fix this is for your project to also depend on the local copy of `azure_core`.

danbugs: @rylev ~ I feel like that can't be the case. I, for example, use the `storage_blobs` crate just fine without having to depend on a local `azure_core`, and that crate itself depends on a local `azure_core` dependency...

danbugs: You're right~ I didn't know Rust would get confused by using two crates of equal versions from different registries. Thanks, Ryan!

danbugs:
@rylev, and @cataggar ~ Is this good to merge? Once we do, we'll have to trigger the new release of the crate!

sdk/messaging_servicebus/src/utils.rs
cataggar: We are okay with returning `Bytes`. A recent similar change I made is [here](https://github.com/Azure/azure-sdk-for-rust/pull/776/files#diff-5e3e9406014c7c90e2c7dd3f7616bc517d15bfdc63b60e47b55eb7bdd7d2a024L66). I'm not going to hold up this PR for this.

sdk/messaging_servicebus/examples/service_bus00.rs
```diff
@@ -1,65 +1,43 @@
...
 
-    // let messages = vec![
-    //     "These", "are", "useless", "messages", "provided", "for", "free", "with", "love",
-    // ];
-    // println!(
-    //     "Sending the following messages: {:?}. \
-    //      Please note they will be sent out of order!",
-    //     messages
-    // );
+    client
```
Mossaka, line 30: it would be great to have 
1. multi-thread example using rust channels, where one thread sends msg to the queue, and another thread recieves the msg from the queue. 
2. multi sender-receiver where one sender sends msg to multiple queues, and mutiple recivers are able to subscribe to different queues. 

what do you think?

sdk/messaging_servicebus/src/service_bus/client.rs
```diff
@@ -0,0 +1,122 @@
...
+            http_client,
+            namespace: namespace.into(),
+            queue: queue.into(),
+            policy_name: policy_name.into(),
+            signing_key,
+        })
+    }
+
+    /// Sends a message to the queue
+    pub async fn send_message(&mut self, msg: &str) -> Result<(), Error> {
```
Mossaka, line 51: no need to make self mutable reference. 
```suggestion
    pub async fn send_message(&self, msg: &str) -> Result<(), Error> {
```

sdk/messaging_servicebus/src/service_bus/client.rs
```diff
@@ -0,0 +1,122 @@
...
+            &self.queue,
+            &self.policy_name,
+            &self.signing_key,
+            msg,
+        )
+        .await
+    }
+
+    /// Receive and delete a message
+    pub async fn receive_and_delete_message(&mut self) -> Result<String, Error> {
```
Mossaka, line 64:
```suggestion
    pub async fn receive_and_delete_message(&self) -> Result<String, Error> {
```

sdk/messaging_servicebus/src/service_bus/client.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// Non-destructively read a message
+    ///
+    /// Note: This function does not return the delete location
+    /// of the message, so, after reading, you will lose
+    /// "track" of it until the lock expiry runs out and
+    /// the message can be consumed by others. If you want to keep
+    /// track of this message (i.e., have the possibility of deletion),
+    /// use `peek_lock_message2`.
+    pub async fn peek_lock_message(
+        &mut self,
```
Mossaka, line 87:
```suggestion
        &self,
```

sdk/messaging_servicebus/src/service_bus/client.rs
```diff
@@ -0,0 +1,122 @@
...
+            .into_body(),
+        )
+    }
+
+    /// Non-destructively read a message but track it
+    ///
+    /// Note: This function returns a `PeekLockResponse`
+    /// that contains a helper `delete_message` function.
+    pub async fn peek_lock_message2(
+        &mut self,
```
Mossaka, line 109:
```suggestion
        &self,
```

Mossaka (changes requested):

sdk/messaging_servicebus/examples/service_bus00.rs
danbugs: Hmmm, I don't know if we should block the PR on examples.

Also, I should note: the Azure Service Bus API does not include the concept of "subscribing".

sdk/messaging_servicebus/src/service_bus/client.rs
danbugs: done ✅

sdk/messaging_servicebus/src/service_bus/client.rs
danbugs: done ✅

sdk/messaging_servicebus/src/service_bus/client.rs
danbugs: done ✅

sdk/messaging_servicebus/src/service_bus/client.rs
```diff
@@ -0,0 +1,122 @@
...
+    },
+    utils::body_bytes_to_utf8,
+};
+use chrono::Duration;
+use ring::hmac::Key;
+
+use azure_core::{error::Error, HttpClient};
+
+/// Client object that allows interaction with the ServiceBus API
+pub struct Client {
```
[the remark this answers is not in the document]
danbugs: done ✅

Issue #771, opened 2022-05-27
migrate SDKs to `azure_core::error::Error`


cataggar:
Three months ago, in late February, #625 was merged. It added `azure_core::error::Error` and migrated a few SDKs. Earlier this week, #766 was merged to migrate all of the generated services. These still need to be migrated.

  - [x] sdk/core HttpClient::execute_request2 @cataggar #779
  - [x] sdk/identity @cataggar #781
  - [x] sdk/security_keyvault @johnbatty #782 
  - [x] sdk/data_tables @johnbatty #805
  - [x] sdk/device_update @mfrw #794
  - [x] sdk/iot_hub @cataggar #801 
  - [x] sdk/storage @johnbatty #792
  - [x] sdk/storage_blobs @rickrain #794
  - [x] sdk/storage_datalake @roeap #795
  - [x] sdk/storage_queues @johnbatty #800 
  - [x] sdk/messaging_servicebus @danbugs #770
  - [x]  sdk/core #806
 
After that, we can remove the existing `azure_core::Error` & `azure_core::HttpError` and friends.

Migration is more tedious than difficult. A bunch of `.map_error` can be removed. others can be replaced with `.context(ErrorKind::SomeKind, "some message"`. If a formatted message is useful, then something like:

``` rust
HeaderValue::from_str(&self.header).with_context(
    ErrorKind::DataConversion,
    || {
        format!(
            "user agent '{}' cannot be serialized as an ascii-only HTTP header
            self.header
        )
    },
)?;
```

cataggar:
@mfrw, yes please. I'll have a PR for identity soon that we can review and may be use as an example.

johnbatty:
I've just implemented this for `security_keyvault`, before noticing that it isn't on your list above!
Was this a deliberate omission?

johnbatty:
In your example above for a formatted message, the method should be `with_context(...)` rather than `context` when passing a closure.

rickrain:
@cataggar, I'm interested in helping on the `sdk/storage_blobs`.

cataggar:
@rickrain, I put your name on it.

roeap:
@cataggar - took a stab at `sdk/storage_datalake`.

cataggar:
Thank you everyone!

PR #775, opened 2022-05-30
set MSI secret env.var optional

### LINKED ISSUE #774: IMDS does not support pod identities
Current implementation of `IMDS` credential requires the user to set `MSI_SECRET_ENV_KEY` environment variable which then hinders usage of the credential when pod identities are enabled on the AKS cluster. Pod identities authentication does not require setting `MSI_SECRET_ENV_KEY` environment variable. Making it optional would enable to use `IMDS` credential with pod identities.

pavelm10:
By allowing `MSI` secret env.var. to be optional the `IMDS` credential can be used with pod identities authentication.

Fixes #774

### DIFF
--- a/sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs
+++ b/sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs
@@ -2,6 +2,7 @@ use super::TokenCredential;
 use azure_core::auth::TokenResponse;
 use chrono::{DateTime, TimeZone, Utc};
 use oauth2::AccessToken;
+use reqwest::header::HeaderMap;
 use serde::{
     de::{self, Deserializer},
     Deserialize,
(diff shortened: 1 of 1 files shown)


sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs
```diff
@@ -97,6 +93,9 @@ impl TokenCredential for ImdsManagedIdentityCredential {
 
         let mut query_items = vec![("api-version", MSI_API_VERSION), ("resource", resource)];
 
+        let mut headers = HeaderMap::new();
+        headers.insert("Metadata", "true".parse().unwrap());
```
ctaggart, line 97: This new `unwrap` isn't my favorite, but I'll be updating the error handling for this crate for #771 very soon.

ctaggart (approved):

cataggar (approved):

PR #776, opened 2022-05-30
convert error responses from bytes

ctaggart:
The conversion to a String first is not needed. After this change, `http_response_from_body` can be used from services. I skipped using it [here](https://github.com/Azure/azure-sdk-for-rust/pull/765#discussion_r881203828).

### DIFF
--- a/sdk/core/src/error/mod.rs
+++ b/sdk/core/src/error/mod.rs
@@ -1,14 +1,13 @@
 use std::borrow::Cow;
 use std::fmt::{Debug, Display};
-
 mod azure_core_errors;
 mod http_error;
 mod hyperium_http;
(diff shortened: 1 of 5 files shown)


sdk/core/src/error/mod.rs
```diff
@@ -39,7 +37,7 @@ impl ErrorKind {
         Self::HttpResponse { status, error_code }
     }
 
-    pub fn http_response_from_body(status: u16, body: &str) -> Self {
+    pub fn http_response_from_body(status: u16, body: &[u8]) -> Self {
```
cataggar, line 40: No need to convert a response body to a String. With this change, it should be usable from services.

Issue #735, opened 2022-05-02
make the sdk wasm32-wasi compatible


Mossaka:
Currently the `azure-core` has a dependency on `reqwest`, which is depended on `wasm-bindgen`. This package can only be used to compile to `wasm32-unknown-unknown` target, used in browser. 

However, this prevent the crate to be compatible to `wasm32-wasi` target. 

Could we make the dependency on `reqwest` an optional feature?

cataggar:
Yes, we'll accept pull requests to make it truly optional. It is already marked as optional in `azure_core`:

https://github.com/Azure/azure-sdk-for-rust/blob/66db4b485ce56b68be148708d9c810960a50be51/sdk/core/Cargo.toml#L28-L30

cataggar:
#755 is in draft to fix this.

Mossaka:
Huge thanks!

cataggar:
I think #939 completed this.

PR #777, opened 2022-05-30
make reqwest optional for azure_core

ctaggart:
For #735, it adds a check for `cargo check -p azure_core --no-default-features` which does not include `enable_reqwest`. Note, that these legacy errors can't be removed until  #771.

### DIFF
--- a/sdk/core/src/options.rs
+++ b/sdk/core/src/options.rs
@@ -1,5 +1,5 @@
 use crate::policies::{ExponentialRetryPolicy, FixedRetryPolicy, NoRetryPolicy, Policy};
-use crate::{new_http_client, HttpClient};
+use crate::HttpClient;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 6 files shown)


sdk/core/src/options.rs
```diff
@@ -15,7 +15,7 @@ use std::time::Duration;
 ///     .retry(RetryOptions::default().max_retries(10u32))
 ///     .telemetry(TelemetryOptions::default().application_id("my-application"));
 /// ```
-#[derive(Clone, Debug, Default)]
```
rylev, line 18: Why was this removed? I think what there is now is equivalent only more verbose.

sdk/core/src/response.rs
```diff
@@ -13,6 +13,7 @@ pub(crate) struct ResponseBuilder {
 }
 
 impl ResponseBuilder {
+    #[allow(dead_code)]
```
rylev, line 16: This maybe points to that this module should only be built if the reqwest feature is turned on.

ctaggart: I updated to use cfg.

sdk/core/src/options.rs
rylev: Ah, can you use [`cfg_attr`](https://stackoverflow.com/questions/42551113/is-it-possible-to-conditionally-enable-an-attribute-like-derive) for this?

PR #778, opened 2022-05-31
add ResultExt::map_kind

ctaggart:
This will help with migration of errors #771.


sdk/core/src/error/mod.rs
```diff
@@ -234,11 +234,18 @@ impl Display for Error {
 ///
 /// This trait cannot be implemented on custom types and is meant for usage with `Result`
 pub trait ResultExt<T>: private::Sealed {
+    /// Creates a new error with the specified kind
+    fn kind(self, kind: ErrorKind) -> Result<T>
```
rylev, line 238: Maybe `map_kind` would be a better name?

rylev (approved):
Looks good though I'm not 100% sure about the name

PR #779, opened 2022-05-31
update execute_request2 to use error::Error instead of HttpError

ctaggart:
For #771. It migrates the HttpClient execute_request2 API in azure_core.

### DIFF
--- a/sdk/core/src/http_client.rs
+++ b/sdk/core/src/http_client.rs
@@ -1,6 +1,9 @@
 #[allow(unused_imports)]
 use crate::Body;
-use crate::HttpError;
+use crate::{
+    error::{Error, ErrorKind, ResultExt},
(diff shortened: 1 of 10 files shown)


sdk/core/src/http_client.rs
```diff
@@ -138,29 +134,15 @@ impl HttpClient for reqwest::Client {
...
                     e,
                     "error converting `reqwest` request into a byte stream",
                 )
             })));
 
         Ok(response)
     }
-
-    #[cfg(target_arch = "wasm32")]
-    /// Stub implementation. Will remove as soon as reqwest starts
```
ctaggart, line 152: This also removes this stub. HttpClient is not implemented by us for wasm.

cataggar (approved):

PR #781, opened 2022-06-04
migrate azure_identity to `azure_core::error::Error`

ctaggart:
Changes:
- Removes `thiserror` errors & replaces with `azure_core::error::Error`
- Removes `identity::TokenCredential`
- `get_token` switched to `azure_core::error::Error`
- changes `Error::with_message` to accept a `FnOnce` like `ResultExt::with_context`.
- Adds `Error::message`, similar to `ResultExt::context`.
- refresh_token error handling now checks response status

PR #782, opened 2022-06-05
Migrate security_keyvault to new azure_core::error scheme

johnbatty:


johnbatty:
@rylev, @cataggar Please could you review, as this is the first of these conversions that I've done.
Please also review/comment on my notes below, in case I have misunderstood anything!

Notes on the process, to help other crate migrations:
- Update `Cargo.toml`
  - Remove `thiserror` crate
  - Update `azure_core` version to 0.2.1
- Remove the `thiserror` error definitions
  - Tagged by `#[derive(thiserror::Error...)]`
- Build... fixup errors... repeat (until it clean compiles)
- `cargo test`
- `cargo clippy`

## Fixing up errors
Existing `thiserror` error types need to be replaced with the new `azure_core::error` errors.  These have a "kind" (`azure_core::error::ErrorKind`), plus a text description of the error/context with any helpful values included.  The main work  is (a) deciding what `ErrorKind` to use (b) creating a helpful error message describing the context and including appropriate values that will be useful to diagnose issues.
```
/// The kind of error
///
/// The classification of error is intentionally fairly coarse.
#[derive(Clone, Debug, PartialEq)]
pub enum ErrorKind {
    /// An HTTP status code that was not expected
    HttpResponse {
        status: u16,
        error_code: Option<String>,
    },
    /// An error performing IO
    Io,
    /// An error converting data
    DataConversion,
    /// An error getting an API credential token
    Credential,
    #[cfg(feature = "mock_transport_framework")]
    /// An error having to do with the mock framework
    MockFramework,
    /// A catch all for other kinds of errors
    Other,
}
```

I found `.with_context(...)` and `.context(...)` to be the main way to convert existing error types into new-style errors.  `.with_context` should be used if formatting the message via `format!(...)`, as it allows the formatting to be passed via a closure which will only get executed if the error is required.  In some situations you may need to directly construct an error to return, which you can do via `Error::with_message(kind, message)`

Example replacements:

Using `with_context(...)`
```
// before:
        let vault_url = Url::parse(vault_url)?;
// after:
        let vault_url = Url::parse(vault_url).with_context(ErrorKind::DataConversion, || {
            format!("failed to parse vault url: {}", vault_url)
        })?;
```

Using `context(...)`
```
// before:
        let client = KeyClient {
            vault_url,
            .token_credential
            .get_token(&self.endpoint)
            .await
            .map_err(|_| Error::Authorization)?;

// after:
        let client = KeyClient {
            vault_url,
            .token_credential
            .get_token(&self.endpoint)
            .await
            .context(ErrorKind::Credential, "get token failed")?;  // .map_err(...) replaced by .context(...)
```

Directly constructing errors:
```
// before:
        if let Some(err) = body_deserialized.get("error") {
            let msg = err.get("message").ok_or(Error::UnparsableError)?;
            Err(Error::General(msg.to_string()))
        }
// after:
        if let Some(err) = body_deserialized.get("error") {
            let msg = err.get("message").ok_or_else(|| {
                Error::with_message(
                    ErrorKind::DataConversion,
                    format!(
                        "failed to read message field from error response. uri:{} body:{}",
                        uri, body
                    ),
                )
            })?;
            Err(Error::with_message(
                ErrorKind::Other,
                format!("post response error: {}", msg),
            ))
```

johnbatty:
Aside...  I noticed that there are a large number of `unwrap()` calls in this crate, which we should replace with more graceful error handling.  I replaced a few, but when I realised how many there were I thought that perhaps I should deal with as a separate issue.  Many of these are related to send/receive operations, so perhaps this code will just be replaced when we switch to the pipeline architecture...?

Example:
```
        let resp = reqwest::Client::new()
            .patch(&uri)
            .bearer_auth(self.token.as_ref().unwrap().token.secret())
            .header("Content-Type", "application/json")
            .body(body)
            .send()
            .await
            .unwrap();
```
Thoughts?

sdk/security_keyvault/src/certificate.rs
```diff
@@ -205,10 +205,8 @@ impl<'a, T: TokenCredential> CertificateClient<'a, T> {
 
         let response_body = self.get_authed(uri.to_string()).await?;
         let response = serde_json::from_str::<KeyVaultGetCertificateResponse>(&response_body)
-            .map_err(|error| Error::BackupCertificateParseError {
-                error,
-                certificate_name: name.to_string(),
-                response_body,
+            .with_context(ErrorKind::DataConversion, || {
+                format!("failed to parse get certificate response. uri:{} certificate_name:{} response_body:{}", uri, name, response_body)
```
cataggar, line 209: May be add spaces after the colons?

sdk/security_keyvault/src/certificate.rs
```diff
@@ -468,17 +466,17 @@ impl<'a, T: TokenCredential> CertificateClient<'a, T> {
...
+            serde_json::from_str::<KeyVaultCertificateBackupResponseRaw>(&response_body)
+                .with_context(ErrorKind::DataConversion, || {
+                    format!("failed to parse certificate backup response. uri:{}", uri)
+                })?;
 
         Ok(CertificateBackupResult {
-            backup: base64::decode(backup_blob.value)?,
+            backup: base64::decode(backup_blob.value)
+                .with_context(ErrorKind::DataConversion, || {
+                    "failed base64 decode of backup blob"
```
cataggar, line 478: This can just be `context`.

sdk/security_keyvault/src/client.rs
```diff
@@ -57,7 +59,7 @@ impl<'a, T: TokenCredential> KeyClient<'a, T> {
             .token_credential
             .get_token(&self.endpoint)
             .await
-            .map_err(|_| Error::Authorization)?;
+            .with_context(ErrorKind::Credential, || "get token failed")?;
```
cataggar, line 62: just `context`

sdk/security_keyvault/src/client.rs
```diff
@@ -137,13 +157,25 @@ impl<'a, T: TokenCredential> KeyClient<'a, T> {
...
+        let body = resp.text().await.with_context(ErrorKind::Io, || {
+            format!("failed to read response body text. uri:{}", uri)
+        })?;
+        let body_deserialized = serde_json::from_str::<serde_json::Value>(&body).unwrap();
+
+        if let Some(err) = body_deserialized.get("error") {
+            let msg = err.get("message").ok_or_else(|| {
+                Error::with_message(
+                    ErrorKind::DataConversion,
+                    format!(
```
cataggar, line 169: In #781, I change this to allow a `|| ` as well.

cataggar (commented):
Looking good. Can you review #781? It was a change that will help improve this PR.

sdk/security_keyvault/src/certificate.rs
johnbatty: Added a space (for all the messages)

sdk/security_keyvault/src/client.rs
johnbatty: Cool.  I can change this once #781 is merged.

sdk/security_keyvault/Cargo.toml
```diff
@@ -14,15 +14,14 @@ edition = "2021"
...
-thiserror = "1.0"
 reqwest = { version = "0.11", features = ["json"] }
 chrono = { version = "0.4", features = ["serde"] }
 const_format = "0.2.13"
 serde_json = "1.0"
 url = "2.2"
 serde = { version = "1.0", features = ["derive"] }
 getset = "0.1"
-azure_core = { path = "../core", version = "0.2" }
+azure_core = { path = "../core", version = "0.2.1" }
```
bmc-msft, line 24: Why do you pick a specific version here?

johnbatty: I'll revert that back.  I changed it to match the version number I had locally, in a vain attempt to fix some build issues (which turned out to be because I hadn't properly synced the latest code).  Functionally there won't be any difference (as 0.2.0 was yanked, and specifying 0.2.1 will match the latest 0.2.x available).  However I think it is clearer to not specify the patch version.

cataggar (approved):

PR #785, opened 2022-06-08
use HttpClient in azure_identity

ctaggart:
The main goal of this PR is to move `reqwest` out of `dependencies` to `dev-dependencies` in `azure_identity`.

Other changes:
- added many standard header constants
- added some content type constants
- Request set_body now allows `Into<Body>`
- Any <Into<Bytes>> can be set as the body
- Simplified setting the body several places
- prefer `serde_json::to_vec` `serde_json::to_string` to skip going to a String


cataggar (approved):
The flows probably need some manual testing.

cataggar:
I would appreciate a review of this.

sdk/identity/src/client_credentials_flow/mod.rs
```diff
@@ -39,13 +39,18 @@
...
+    headers, HttpClient, Request,
+};
+use http::Method;
 use login_response::LoginResponse;
 use url::form_urlencoded;
 
 /// Perform the client credentials flow
 pub async fn perform(
-    client: reqwest::Client,
+    http_client: &dyn HttpClient,
```
bmc-msft, line 53: This feels weird from the consumer side, given the client we get from `new_http_client` is a `Arc<dyn HttpClient>`.   In other areas in the API, we use `.clone()` instead of `.as_ref()`, which feels slightly more natural.

bmc-msft: Other than that, this LGTM and works in all of my use cases.

bmc-msft (approved):

PR #791, opened 2022-06-10
sdk: device_update: migrate to `azure_core::error::Error`

mfrw:
For:
- #771

Changes:

- Remove `thiserror` & replace with `azure_core::error::Error`

### DIFF
--- a/sdk/device_update/src/client.rs
+++ b/sdk/device_update/src/client.rs
@@ -107,12 +107,12 @@ impl DeviceUpdateClient {
             return match headers.get("operation-location") {
                 Some(location) => location.to_str().map(|x| x.to_string()).map_err(|_| {
                     Error::with_message(ErrorKind::Other, || {
-                        "successful import (202 status) but no operation-location header found"
-                            .to_string()
+                        "invalid characters in operation-location path".to_string()
                     })
                 }),
                 None => Err(Error::with_message(ErrorKind::Other, || {
-                    "invalid characters in operation-location path".to_string()
+                    "successful import (202 status) but no operation-location header found"
+                        .to_string()
                 })),
             };
         }
--- a/sdk/device_update/src/client.rs
+++ b/sdk/device_update/src/client.rs
@@ -118,7 +118,7 @@ impl DeviceUpdateClient {
         }
 
         Err(Error::with_message(ErrorKind::Other, || {
-            format!("import unseccessful, status: {}", resp.status())
+            format!("import unsuccessful, status: {}", resp.status())
         }))
     }
 
--- a/sdk/device_update/src/client.rs
+++ b/sdk/device_update/src/client.rs
@@ -1,7 +1,6 @@
-use crate::{Error, Result};
 use azure_core::{
     auth::{TokenCredential, TokenResponse},
-    Error as CoreError, HttpError,
+    error::{Error, ErrorKind, Result, ResultExt},
 };
 use azure_identity::AutoRefreshingTokenCredential;
 use const_format::formatcp;
@@ -45,8 +44,10 @@ impl DeviceUpdateClient {
         device_update_url: &str,
         token_credential: Arc<dyn TokenCredential>,
     ) -> Result<Self> {
-        let device_update_url =
-            Url::parse(device_update_url).map_err(|e| CoreError::Http(HttpError::Url(e)))?;
+        let device_update_url = Url::parse(device_update_url)
+            .with_context(ErrorKind::DataConversion, || {
+                format!("failed to parse update url: {device_update_url}")
+            })?;
         let endpoint = extract_endpoint(&device_update_url)?;
         let token_credential = AutoRefreshingTokenCredential::new(token_credential);
 
@@ -62,7 +63,7 @@ impl DeviceUpdateClient {
         self.token_credential
             .get_token(&self.endpoint)
             .await
-            .map_err(|e| Error::Core(e.into()))
+            .context(ErrorKind::Credential, "get token failed")
     }
 
     pub(crate) async fn get<R>(&self, uri: String) -> Result<R>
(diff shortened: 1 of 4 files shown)


sdk/device_update/src/client.rs
```diff
@@ -94,23 +98,28 @@ impl DeviceUpdateClient {
...
             return match headers.get("operation-location") {
-                Some(location) => location
-                    .to_str()
-                    .map(|x| x.to_string())
-                    .map_err(|_| Error::InvalidOperationPath),
-                None => Err(Error::NoOperationLocation),
+                Some(location) => location.to_str().map(|x| x.to_string()).map_err(|_| {
+                    Error::with_message(ErrorKind::Other, || {
+                        "successful import (202 status) but no operation-location header found"
+                            .to_string()
```
johnbatty, line 111: This error needs to be swapped with the following one - i.e. this one should be "invalid characters..."

sdk/device_update/src/client.rs
```diff
@@ -94,23 +98,28 @@ impl DeviceUpdateClient {
...
+                }),
+                None => Err(Error::with_message(ErrorKind::Other, || {
+                    "invalid characters in operation-location path".to_string()
+                })),
             };
         }
 
-        Err(Error::ImportError(resp.status()))
+        Err(Error::with_message(ErrorKind::Other, || {
+            format!("import unseccessful, status: {}", resp.status())
```
johnbatty, line 121: typo: "unseccessful" -> "unsuccessful"

sdk/device_update/src/device_update.rs
```diff
@@ -183,11 +186,27 @@ impl DeviceUpdateClient {
             let update_operation: UpdateOperation = self.get(uri.to_string()).await?;
 
             match update_operation.status {
-                OperationStatus::Failed => return Err(Error::ImportFailed(update_operation)),
+                OperationStatus::Failed => {
+                    return Err(Error::with_message(ErrorKind::Other, {
+                        || {
+                            format!(
+                                "import unsccessful with status failed. status: {:?}",
```
johnbatty, line 193: typo: "unsccessful" => "unsuccessful"

johnbatty (commented):
Looking good! Thanks very much for doing this.  Just a few minor markups.

sdk/device_update/src/client.rs
mfrw: Ahh :( my spellings :(

sdk/device_update/src/client.rs
mfrw: Sure

sdk/device_update/src/client.rs
```diff
@@ -74,13 +75,16 @@ impl DeviceUpdateClient {
...
-            .map_err(|e| Error::Core(CoreError::Http(HttpError::ReadBytes(e.into()))))?;
-        serde_json::from_slice(&body).map_err(|e| Error::Core(CoreError::Json(e)))
+            .with_context(ErrorKind::Io, || {
+                format!("failed to send request. uri: {uri}")
+            })?;
+
+        let body = resp.bytes().await.with_context(ErrorKind::Io, || {
+            format!("failed to read response body text. uri: {uri}")
+        })?;
+        serde_json::from_slice(&body).with_context(ErrorKind::DataConversion, || {
```
cataggar, line 85: You can use `.context` here.

sdk/device_update/src/client.rs
```diff
@@ -74,13 +75,16 @@ impl DeviceUpdateClient {
...
+            "failed to deserialze json response body".to_string()
```
cataggar, line 86: deserialize

sdk/device_update/src/client.rs
```diff
@@ -94,23 +98,28 @@ impl DeviceUpdateClient {
...
 
         if resp.status() == 202u16 {
             let headers = resp.headers();
             return match headers.get("operation-location") {
-                Some(location) => location
-                    .to_str()
-                    .map(|x| x.to_string())
-                    .map_err(|_| Error::InvalidOperationPath),
-                None => Err(Error::NoOperationLocation),
+                Some(location) => location.to_str().map(|x| x.to_string()).map_err(|_| {
```
cataggar, line 108: Use `context` instead of `map_err` to keep the source error.

sdk/device_update/src/client.rs
```diff
@@ -94,23 +98,28 @@ impl DeviceUpdateClient {
...
+                    Error::with_message(ErrorKind::Other, || {
+                        "invalid characters in operation-location path".to_string()
+                    })
+                }),
+                None => Err(Error::with_message(ErrorKind::Other, || {
```
cataggar, line 113: You are not formatting the error message, so you can just use `Error::message`

sdk/device_update/src/client.rs
```diff
@@ -134,40 +144,57 @@ impl DeviceUpdateClient {
...
     use super::*;
+    use azure_core::error::{Result, ResultExt};
 
     #[test]
     fn can_extract_endpoint() -> Result<()> {
+        let url = "https://myadu.api.adu.microsoft.com";
         let suffix = extract_endpoint(
-            &Url::parse("https://myadu.api.adu.microsoft.com")
-                .map_err(|e| CoreError::Http(HttpError::Url(e)))?,
+            &Url::parse(url).with_context(ErrorKind::DataConversion, || {
```
cataggar, line 171: These are tests, so context isn't really needed. You can just do `&Url::parse(url)?`. `From` is implemented.
https://github.com/Azure/azure-sdk-for-rust/blob/a6625a5f3b72cc6bf6095c7116d325a79e7816ce/sdk/core/src/error/mod.rs#L237-L241

cataggar: May be check if `url::ParseError` provides the url that it fails on. If it does, then context probably isn't needed in a lot of these.

sdk/device_update/src/client.rs
```diff
@@ -134,42 +144,48 @@ impl DeviceUpdateClient {
...
-        .ok_or(Error::DomainParse)?
-        .splitn(2, '.') // FIXME: replace with split_once() when it is in stable
-        .last()
-        .ok_or(Error::DomainParse)?;
+        .ok_or_else(|| {
+            Error::with_message(ErrorKind::DataConversion, || {
+                format!("could not get device update domain. url: {url}")
+            })
+        })?
+        .split_once('.')
```
mfrw, line 152: I also noticed the `FIXME: use split_once`
Might as well do it as it is in stable now :)

sdk/device_update/src/client.rs
mfrw: 😢

sdk/device_update/src/client.rs
mfrw: It does not :(, but I think we should not clutter the tests.
I've added one test that should fail though.

bmc-msft (approved):

PR #792, opened 2022-06-10
Migrate storage to new azure_core::error scheme

johnbatty:


johnbatty:
I've believe I've made all the necessary changes to `storage`, and the crate itself builds.  However the changes cause the other storage crates to fail to build because they have a dependency on `storage`, and in particular they expect old-style error types.  Therefore we need to push on and get all of the storage crate family migrated, at which point this should build/pass the CI tests.

johnbatty:
Ok, now builds with the latest storage crates.  Had to tweak `storage_blobs` to fix up a couple of API calls to match the new API.

ctaggart (approved):

PR #794, opened 2022-06-10
Migrate sdk/storage_blob to azure_core::error scheme

rickrain:
This PR migrates the sdk/storage_blob to the new azure_core::error scheme as described in #771.


cataggar (dismissed):
Thank you!

johnbatty:
Hi Rick.  Many thanks for this - looks great to me, so far...

However, (unfortunately!) there are still quite a few more changes that need to be made for this crate.  This crate not only has its own error definitions, but pulls in most of its error definitions from `azure_storage`.  See `lib.rs`:
```
pub use azure_storage::{Error, Result};
```
If you change this to pull in the new-style error definitions from `azure_core` you'll get a load more compiler errors that will need fixing up.
```
pub use azure_core::error::{Error, Result};
```

rickrain:
@johnbatty, thanks for pointing that out.  I was a little suspicious there weren't more changes. Following your guidance, I'm now seeing 90+ errors that need to be resolved.  I'll work on this and ping you and @cataggar to re-review when I'm done.

johnbatty:
@rickrain I just had similar issues working on storage_queues.  I decided to get it building independently with the existing (non-converted) storage crate using a map_kind(...) on the non-azure_core error(s) returned by storage.  Here's an example from my storage_queues PR:
```rust
        self.storage_client
            .queue_url_with_segments(Some(self.queue_name.as_str()).into_iter().chain(segments))
            .map_kind(ErrorKind::DataConversion)
```
If we can get the higher level storage crates building/merged, then we can merge the storage crate and have a pass through to clean up any unnecessary ErrorKind maps.

rickrain:
@johnbatty, thanks a ton for that pointer. That helped a lot!

rickrain:
@johnbatty, @cataggar This PR is ready for review.

bmc-msft (approved):

cataggar (approved):

PR #795, opened 2022-06-10
Migrate storage_datalake to `azure_core::error::Error`

roeap:
towards #771

Took a stab at migrating datalake - thought this might be dependent on #792, but seems to compile...

### DIFF
--- a/sdk/storage_datalake/src/properties.rs
+++ b/sdk/storage_datalake/src/properties.rs
@@ -1,3 +1,4 @@
+use azure_core::error::{Error, ErrorKind, ResultExt};
 use azure_core::headers::{self, Header, PROPERTIES};
 use http::HeaderMap;
 use std::borrow::Cow;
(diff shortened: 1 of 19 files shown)


sdk/storage_datalake/src/properties.rs
```diff
@@ -54,8 +55,9 @@ impl TryFrom<&HeaderMap> for Properties {
     fn try_from(headers: &HeaderMap) -> Result<Self, Self::Error> {
         let header_value = headers
             .get(PROPERTIES)
```
ctaggart, line 57: Hmm, may be we need to rename to `.try_get` and add a `get` that errors like this.

ctaggart (approved):
Thanks for the conversion!

johnbatty:
Looks good to me too.  Thanks very much for working on this!

johnbatty (approved):

PR #798, opened 2022-06-13
update datalake head and list operations - more blob migration

### LINKED ISSUE #786: [datalake] Expose Content-Length in get_properties
Amongst the headers returned by the [Get Properties](https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/get-properties) endpoint is `Content-Length`, but that is not present on [HeadPathResponse](https://docs.rs/azure_storage_datalake/0.2.0/azure_storage_datalake/operations/struct.HeadPathResponse.html).

Is the a way to access that now that I'm missing? Or could we add this?

roeap:
This PR Closes #786 and addresses an issue with the list operations in datalake crate.

Not sure if I went the right way and still need to add a test, but looking for some feedback :).


sdk/storage_datalake/src/file_system.rs
```diff
@@ -41,14 +43,72 @@ pub struct Path {
...
 impl TryFrom<Bytes> for PathList {
     type Error = azure_core::error::Error;
 
     fn try_from(response: Bytes) -> Result<Self, Self::Error> {
-        Ok(serde_json::from_slice::<PathList>(response.as_ref())?)
+        Ok(serde_json::from_slice(response.as_ref())?)
+    }
+}
+
+mod i64_or_string {
```
roeap, line 58: In one of our programs we get a quoted value back for bot the `is_directory` and `content_length` fields. In that case serde fails to deserialize. But since encding numbers or bool as strings seems to not be done in other places in the API, i felt it was preferable to just have a custom deserialize that can handle both cases. IN any case we would want to do the conversion from sting... the added code complexity is hopefully bearable... then again there may be things I am overlooking :)

rylev (changes requested):
Looks good! I think we can somewhat simplify the custom serialization but otherwise looks good.

sdk/core/src/headers/utilities.rs
```diff
@@ -207,6 +207,10 @@ pub fn content_type_from_headers(headers: &HeaderMap) -> Result<&str> {
     get_str_from_headers(headers, http::header::CONTENT_TYPE.as_str())
 }
 
+pub fn content_length_from_headers(headers: &HeaderMap) -> Result<i64> {
+    get_from_headers(headers, CONTENT_LENGTH)
+}
```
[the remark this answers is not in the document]
roeap: did it fir this case - i think this is cleaner and just as convenient

roeap:
@rylev - implemented your suggestions and fixed a required parameter in `list_paths` with some corresponding tests.

@ctaggart - this PR contains some fixed that would help us out a lot in delta-io/delta-rs#625, would it be possible to release a patch it this gets merged?

cataggar (dismissed):

cataggar:
>  this PR contains some fixed that would help us out a lot in https://github.com/delta-io/delta-rs/pull/625, would it be possible to release a patch if this gets merged?

@roeap , please bump the version. This contains more than what normally goes in a patch. May be 0.4.0 instead of 0.3.1.

sdk/storage_datalake/src/file_system.rs
```diff
@@ -41,14 +44,50 @@ pub struct Path {
...
 impl TryFrom<Bytes> for PathList {
     type Error = azure_core::error::Error;
 
     fn try_from(response: Bytes) -> Result<Self, Self::Error> {
-        Ok(serde_json::from_slice::<PathList>(response.as_ref())?)
+        Ok(serde_json::from_slice(response.as_ref())?)
+    }
+}
+
+fn deserialize_i64<'de, D>(deserializer: D) -> Result<i64, D::Error>
```
cataggar, line 59: Can you add a unit test?

sdk/storage_datalake/src/file_system.rs
```diff
@@ -41,14 +44,50 @@ pub struct Path {
...
+            ))),
+        },
+        other => Err(serde::de::Error::custom(format!(
+            "unexpected data format - expected string or number, got: {:?}",
+            other
+        ))),
+    }
+}
+
+fn deserialize_bool<'de, D>(deserializer: D) -> Result<bool, D::Error>
```
cataggar, line 80: Can you add a unit test?

sdk/storage_datalake/src/operations/path_head.rs
```diff
@@ -105,6 +108,8 @@ impl HeadPathResponse {
             common_storage_response_headers: headers.try_into()?,
             etag: etag_from_headers(headers)?,
             last_modified: last_modified_from_headers(headers)?,
+            content_length: get_from_headers(headers, CONTENT_LENGTH).ok(),
```
cataggar, line 111: Should this be `get_option_from_headers(headers, CONTENT_LENGTH)?,` ?

sdk/storage_datalake/src/operations/path_head.rs
```diff
@@ -105,6 +108,8 @@ impl HeadPathResponse {
...
+            content_type: get_from_headers(headers, CONTENT_TYPE).ok(),
```
cataggar, line 112: get_option_from_headers ?

roeap: absolutely...

sdk/storage_datalake/src/file_system.rs
roeap: added a test for both deserializers.

cataggar (approved):

bmc-msft (approved):

roeap:
Thanks @cataggar, @bmc-msft for making this happen so quickly!!

PR #799, opened 2022-06-13
move to using `Arc<dyn HttpClient>` in azure_identity

bmc-msft:
From discussions in a previous PR, the more common method in the SDK is to clone the `Arc<dyn HttpClient>`, rather than pass a reference.

https://github.com/Azure/azure-sdk-for-rust/pull/785#discussion_r895938450

Note, this uses the crate [fix-hidden-lifetime-bug](https://crates.io/crates/fix-hidden-lifetime-bug) to work around an issue with async fn trait issues (ref: https://github.com/rust-lang/rust/issues/63033).

Without this workaround, we see the error:
```
error[E0700]: hidden type for `impl Trait` captures lifetime that does not appear in bounds
  --> sdk/identity/src/client_credentials_flow/mod.rs:59:28
   |
59 |   ) -> Result<LoginResponse> {
   |  ____________________________^
60 | |     let encoded: String = form_urlencoded::Serializer::new(String::new())
61 | |         .append_pair("client_id", client_id.as_str())
62 | |         .append_pair("scope", &scopes.join(" "))
...  |
89 | |     serde_json::from_slice(&rsp_body).map_kind(ErrorKind::DataConversion)
90 | | }
   | |_^
   |
   = note: hidden type `impl futures::Future<Output = std::result::Result<client_credentials_flow::login_response::LoginResponse, azure_core::error::Error>>` captures lifetime '_#42r
```

### DIFF
--- a/sdk/identity/Cargo.toml
+++ b/sdk/identity/Cargo.toml
@@ -28,6 +28,8 @@ openssl = { version = "0.10",  optional=true }
 base64 = "0.13.0"
 uuid = { version = "1.0",  features = ["v4"] }
 http = "0.2"
+# work around https://github.com/rust-lang/rust/issues/63033
+fix-hidden-lifetime-bug = "0.2"
 
 [dev-dependencies]
 reqwest = { version = "0.11", features = ["json"], default-features = false }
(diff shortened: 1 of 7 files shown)


sdk/identity/Cargo.toml
```diff
@@ -28,6 +28,8 @@ openssl = { version = "0.10",  optional=true }
 base64 = "0.13.0"
 uuid = { version = "1.0",  features = ["v4"] }
 http = "0.2"
+# work around https://github.com/rust-lang/rust/issues/63033
+fix-hidden-lifetime-bug = "0.2"
```
cataggar, line 32: 😱

cataggar (approved):

PR #800, opened 2022-06-13
Migrate storage_queues to new azure_core::error scheme

johnbatty:

### DIFF
--- a/sdk/storage_queues/src/responses/get_queue_service_properties_response.rs
+++ b/sdk/storage_queues/src/responses/get_queue_service_properties_response.rs
@@ -1,4 +1,5 @@
 use crate::QueueServiceProperties;
+use azure_core::error::{Error, ErrorKind, Result, ResultExt};
 use azure_storage::core::headers::CommonStorageResponseHeaders;
 use azure_storage::core::xml::read_xml;
 use bytes::Bytes;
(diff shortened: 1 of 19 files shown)


sdk/storage_queues/src/responses/get_queue_service_properties_response.rs
```diff
@@ -12,15 +13,16 @@ pub struct GetQueueServicePropertiesResponse {
...
-    fn try_from(response: &Response<Bytes>) -> Result<Self, Self::Error> {
+    fn try_from(response: &Response<Bytes>) -> Result<Self> {
         let headers = response.headers();
         let body = response.body();
 
         debug!("headers == {:?}", headers);
         debug!("body == {:#?}", body);
-        let queue_service_properties: QueueServiceProperties = read_xml(body)?;
+        let queue_service_properties: QueueServiceProperties =
+            read_xml(body).map_kind(ErrorKind::DataConversion)?;
```
cataggar, line 25: I imagine that  `.map_kind` will not be needed here when `read_xml` is migrated.

cataggar (approved):

sdk/storage_queues/src/responses/get_queue_service_properties_response.rs
johnbatty: I did a similar thing for the clients that are also defined by the `storage` crate.

bmc-msft (approved):

PR #801, opened 2022-06-14
`azure_core::error::Error` for `azure_iot_hub`

ctaggart:


sdk/iot_hub/src/service/requests/get_twin.rs
```diff
@@ -27,7 +26,9 @@ where
     };
 
     let request = service_client.prepare_request(&uri, Method::GET);
-    let request = request.body(azure_core::EMPTY_BODY)?;
+    let request = request
+        .body(azure_core::EMPTY_BODY)
+        .map_kind(ErrorKind::Io)?;
```
johnbatty, line 31: Is this `map_kind(...)` required?  Other similar functions seem to build with `.body(azure_core::EMPTY_BODY)?`

johnbatty (approved):
Looks good!

sdk/iot_hub/src/service/requests/get_twin.rs
cataggar: Good catch. Removed.

sdk/iot_hub/src/service/mod.rs
```diff
@@ -116,10 +90,14 @@ impl ServiceClient {
             iot_hub_name, &expiry_date_seconds
         );
 
-        let key = decode(private_key).map_err(GenerateSasTokenError::DecodePrivateKeyError)?;
+        let key = decode(private_key).with_context(ErrorKind::Other, || {
```
rickrain, line 93: Would it be appropriate for errors relating to the construction of a SAS token to be `ErrorKind::Credential` instead of `ErrorKind::Other`?

Issue #809, opened 2022-06-15
crate for URL query params


cataggar:
Consider using `Url` from `url` crate instead of `Uri` from `http` crate in the HTTP Pipeline.

http::Uri does not Url supports query params
https://github.com/Azure/azure-sdk-for-rust/pull/785/files#r892727411

qstring crate used in AutoRust code gen
https://github.com/Azure/azure-sdk-for-rust/pull/796/files#diff-e58972f7fb42139dc66ee720139c18d34d3ea0ef036688c9411f14da6511dae0R127

PR #811, opened 2022-06-15
Add pipeline to storage_account_client

rylev:
This is a WIP PR after a shared "mob programming" session with some folks. 

This shows an (almost) complete move of one operation to the pipeline architecture.


sdk/storage/src/account/operations/get_account_information.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub fn into_future(mut self) -> GetAccountInformation {
+        Box::pin(async move {
+            let mut request = self
+                .storage_client
+                .storage_account_client()
+                .blob_storage_request("", http::Method::GET);
+
+            // TODO: add the query pairs
+            // request.uri_mut().query_pairs_mut().append_pair("restype", "account");
+            // request.uri_mut().query_pairs_mut().append_pair("comp", "properties");
```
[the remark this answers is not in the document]
cataggar: I think we should probably use `url::Url` instead. #809 We may wish to review why we are using `http` crate at all.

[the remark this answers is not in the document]
rylev: Very much agree. Should we do that in this PR or in another? Might be easier in another.

sdk/identity/src/device_code_flow/mod.rs
```diff
@@ -171,7 +171,7 @@ async fn post_form(
     url: &str,
     form_body: String,
 ) -> Result<Response> {
-    let url = Request::parse_uri(url)?;
+    let url = Url::parse(url).map_kind(ErrorKind::DataConversion)?;
```
cataggar, line 174: `map_kind` not needed. 

https://github.com/Azure/azure-sdk-for-rust/blob/02e025a49c1e081d2be7ac4e57287dd0727e551b/sdk/core/src/error/mod.rs#L236-L240

cataggar, line 174: `map_kind` not needed. 

https://github.com/Azure/azure-sdk-for-rust/blob/02e025a49c1e081d2be7ac4e57287dd0727e551b/sdk/core/src/error/mod.rs#L236-L240

sdk/core/src/request.rs
```diff
@@ -36,25 +35,38 @@ impl From<Box<dyn SeekableStream>> for Body {
...
+    pub(crate) url: Url,
     pub(crate) method: Method,
     pub(crate) headers: Headers,
     pub(crate) body: Body,
 }
 
 impl Request {
     /// Create a new request with an empty body and no headers
-    pub fn new(uri: Uri, method: Method) -> Self {
+    pub fn new(url: Url, method: Method) -> Self {
```
cataggar, line 46: Fixes #809

cataggar (approved):

PR #820, opened 2022-06-16
2022 June Release update

ctaggart:
After publishing #813 yesterday, going to republish a few crates with these changes:

- #798
- #814
- #815 
- #816
- #817
- #818
- #819

Publishing these crates:

- azure_data_cosmos 0.4.0
- azure_data_tables 0.4.0
- azure_storage 0.4.0
- azure_storage_blobs 0.4.0
- azure_storage_datalake 0.4.0
- azure_storage_queues 0.4.0

PR #813, opened 2022-06-15
2022 June Release

ctaggart:
azure_core 0.3.0
azure_data_cosmos 0.3.0
azure_data_tables 0.3.0
azure_identity 0.4.0
azure_iot_hub 0.3.0
azure_messaging_eventgrid 0.3.0
azure_messaging_servicebus 0.3.0
azure_security_keyvault 0.3.0
azure_storage 0.3.0
azure_storage_blobs 0.3.0
azure_storage_datalake 0.3.0
azure_storage_queues 0.3.0

services 0.4.0
azure_svc_*
azure_mgmt_*

PR #822, opened 2022-06-17
update tools for publishing

ctaggart:
These are updated tools I used for publishing #813 &  #820.

### DIFF
--- a/services/autorust/codegen/examples/gen_workspace.rs
+++ b/services/autorust/codegen/examples/gen_workspace.rs
@@ -2,9 +2,9 @@
 
 use autorust_codegen::{
     crates::{list_crate_names, list_dirs},
-    jinja::{CargoToml, CheckAllServicesYml},
+    jinja::{CargoToml, CheckAllServicesYml, PublishSdksYml, PublishServicesYml},
+    Result,
 };
-pub type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
 
 fn main() -> Result<()> {
     let dirs = list_dirs()?;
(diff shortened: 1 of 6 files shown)


services/autorust/codegen/examples/gen_workspace.rs
```diff
@@ -19,7 +19,37 @@ fn main() -> Result<()> {
     let yml = CheckAllServicesYml { packages };
     yml.create("../../.github/workflows/check-all-services.yml")?;
 
-    // let yml = PublishServicesYml { packages };
-    // yml.create("../../.github/workflows/publish-services.yml")?;
+    // publish_sdks()?;
+    // publish_services()?;
```
rylev, line 23: Can we put these behind a command line flag so that you're not expected to uncomment these to actually run the thing?

services/autorust/codegen/examples/gen_workspace.rs
```diff
@@ -19,7 +19,37 @@ fn main() -> Result<()> {
...
+        "azure_messaging_eventgrid",
+        "azure_messaging_servicebus",
+        "azure_security_keyvault",
+        "azure_storage",
+        "azure_storage_blobs",
+        "azure_storage_datalake",
+        "azure_storage_queues",
+    ];
+    let yml = PublishSdksYml { packages };
+    yml.create("../../.github/workflows/publish-sdks.yml")?;
```
rylev, line 44: nit: would be awesome to not have to rely on these very awkward relative paths.

services/autorust/codegen/examples/gen_workspace.rs
```diff
@@ -19,7 +19,37 @@ fn main() -> Result<()> {
...
+    Ok(())
+}
+
+#[allow(dead_code)]
```
rylev, line 48: With the above comment fixed we could remove this.

services/autorust/codegen/examples/gen_workspace.rs
cataggar: I added a `publish` arg.

rylev (approved):

PR #824, opened 2022-06-17
fix: `azure_core` fails to build with `azurite_workaround` feature

### LINKED ISSUE #823: `azure_core` fails to build with `azurite_workaround` feature
after the error migration (#771) one conflicting result definition is left behind the `azurite_workaround` feature gate.

roeap:
This PR fixes a conflicting error definition left behind the `azurite_workaround` feature gate.

closes #823

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,6 +65,9 @@ jobs:
       - name: check core for wasm
         run: cargo check -p azure_core --target=wasm32-unknown-unknown
 
+      - name: check for azurite_workaround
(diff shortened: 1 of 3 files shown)


rylev (changes requested):
Good catch! Can we add a run of `cargo check --feature azurite_workaround` to CI so that this won't happen again in the future?

roeap:
@rylev - absolutely, and good thing we did, b/c there was another one hidden in the `storage_blobs` crate.

rylev (approved):

PR #825, opened 2022-06-17
Move Result to top of core crate

rylev:
This rather large PR is quite simple. It adds a `pub use` declaration to `azure_core` so that users can access the `Result` type alias from the root of `azure_core` like so: `azure_core::Result`. This is quite a common pattern in the Rust world. 

The rest of the changes are simply moving all uses of `azure_core`'s `Result` type to use this path so that we have consistency across the code base.

### DIFF
--- a/sdk/core/src/lib.rs
+++ b/sdk/core/src/lib.rs
@@ -40,6 +40,7 @@ use uuid::Uuid;
 pub use bytes_stream::*;
 pub use constants::*;
 pub use context::Context;
+pub use error::Result;
 #[doc(inline)]
 pub use headers::Header;
 #[cfg(any(feature = "enable_reqwest", feature = "enable_reqwest_rustls"))]
--- a/sdk/core/src/auth.rs
+++ b/sdk/core/src/auth.rs
@@ -1,6 +1,5 @@
 //! Azure authentication and authorization.
 
-use crate::error::Error;
 use chrono::{DateTime, Utc};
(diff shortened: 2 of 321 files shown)


bmc-msft:
I was hoping to ask about doing the same.

bmc-msft:
This LGTM, the only thing I would have done differently is doing `use azure_core::Result` and then using `Result`, rather than repeatedly using `azure_core::Result`.

bmc-msft (approved):

sdk/core/src/lib.rs
```diff
@@ -40,6 +40,7 @@ use uuid::Uuid;
 pub use bytes_stream::*;
 pub use constants::*;
 pub use context::Context;
+pub use error::Result;
```
cataggar, line 43: Why just `Result` and not `errror::*`?

rylev: We could do that though I think I'd leave that for a future PR.

rylev:
> This LGTM, the only thing I would have done differently is doing `use azure_core::Result` and then using `Result`, rather than repeatedly using `azure_core::Result`.

I purposefully went this route since this tends to be more idiomatic. Generally speaking, unqualified `Result` is reserved for `std::result::Result` while qualified `Result` is used for others. This is why in the std lib docs, you often see `std::io::Result` used as `io::Result` instead of just as `Result`.

PR #832, opened 2022-06-18
Fix 401 Unauthorized error in CosmosDB operations

### LINKED ISSUE #831: Unauthorized error in CosmosDB operations
CosmosDB operations using azure_data_cosmos throw 401 Unauthorized.

For example, an execution of code of `examples/readme.rs` results below.
```
Inserting 10 documents...
Error: Error { context: Full(Custom { kind: HttpResponse { status: 401, error_code: None }, error: HttpError { status: 401, error_code: None, headers: {"transfer-encoding": "chunked", "server": "Microsoft-HTTPAPI/2.0", "x-ms-activity-id": "746a4599-4fe5-4956-962e-cd79b7de043f", "x-ms-gatewayversion": "version=2.14.0", "content-type": "application/json", "strict-transport-security": "max-age=31536000", "date": "Sat, 18 Jun 2022 08:34:53 GMT"}, body: b"{\"code\":\"Unauthorized\",\"message\":\"The input authorization token can't serve the request. The wrong key is being used or the expected payload is not built as per the protocol. For more info: https://aka.ms/cosmosdb-tsg-unauthorized. Server used the following payload to sign: 'post\\ndocs\\ndbs/TestDatabase/colls/TestContainer\\nsat, 18 jun 2022 08:34:53 gmt\\n\\n'\\r\\nActivityId: 746a4599-4fe5-4956-962e-cd79b7de043f, Microsoft.Azure.Documents.Common/2.14.0\"}" } }, "server returned error status which will not be retried") }
```

I confirmed 
- It failed at (least) 55dc0464117dbe7331a97a265c191b238b13b26e
    - (with dev-dependencies `tokio = { version = "1", features = ["macros", "rt-multi-thread"] }`)
- It succeed at 267d8cea2ce5d3bcdca9f7005d7620d7ac645ce3

spica314:
fixes #831

### DIFF
--- a/sdk/data_cosmos/src/authorization_policy.rs
+++ b/sdk/data_cosmos/src/authorization_policy.rs
@@ -57,18 +57,15 @@ impl Policy for AuthorizationPolicy {
 
         let time_nonce = TimeNonce::new();
 
-        let uri_path = &request.path_and_query()[1..];
-        trace!("uri_path used by AuthorizationPolicy == {:#?}", uri_path);
-
         let auth = {
-            let resource_link = generate_resource_link(uri_path);
+            let resource_link = generate_resource_link(request);
             trace!("resource_link == {}", resource_link);
             generate_authorization(
                 &self.authorization_token,
                 &request.method(),
                 ctx.get()
                     .expect("ResourceType must be in the Context at this point"),
-                resource_link,
+                &resource_link,
                 time_nonce,
             )
         };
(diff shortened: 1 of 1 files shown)


cataggar (commented):
This is fine to me. I'd like @rylev to review. Would prefer a test added. I wonder if the query part is actually needed for the AuthorizationPolicy.

sdk/data_cosmos/src/authorization_policy.rs
```diff
@@ -57,11 +57,11 @@ impl Policy for AuthorizationPolicy {
 
         let time_nonce = TimeNonce::new();
 
-        let uri_path = request.path_and_query();
+        let uri_path = &request.path_and_query()[1..];
```
cataggar, line 60: This matches what the code was, prior to this change:
https://github.com/Azure/azure-sdk-for-rust/pull/811/files#diff-1d8d32914ddb120499d9acbf8fb856f3757b01bf2d612c4eef44e5ff9590dcbaL57

I think in #811, there was a switch from `http::Uri` to `url::Url` and `path_and_query` existed in the former.

https://docs.rs/http/latest/http/uri/struct.Uri.html#method.path_and_query
https://docs.rs/url/latest/url/struct.Url.html#method.path

rylev (changes requested):
Whoops! Good find. I think we should probably change the `generate_resource_link` to take a `&Request` and have that function do the magical `&request.path_and_query()[1..]` dance. That way it's all contained within that function and not the callers responsibility to give the `generate_resource_link` function a "uri path" in the right format.

spica314:
@rylev 
Thank you for your review!

I moved the code for striping the leading slash to `generate_resource_link`.
(Since my English skill is poor, if my comments are incorrect, please tell me...)

And I also rewrite tests for `generate_resource_links`, but I'm not sure I'm following your suggestion. 
If not, please tell me more (example or functions to use).

rylev (approved):
Thank you! 🎉

PR #833, opened 2022-06-18
migrate SDKs from `http::Request` to `azure_core::Request`

ctaggart:
This completes the migration from `http::Request` to `azure_core::Request`. This should make moving to HTTP pipelines easier. Some highlights:

- rewrote `execute_request_check_status` function
  - instead of `http::Response<Bytes>`, it returns a new `azure_core::CollectedResponse`
- replaced `http:HeaderMap` with `azure_core::headers::Headers`
  - migrated several helper functions like `get_as_str`
  - added some `or_err` variants like `get_as_str_or_err`
- replaced some helper functions for `http:request::Builder` with functions in `azure_core::Request`
  - add_optional_header
  -  add_optional_header_ref
  - add_mandatory_header


sdk/core/src/http_client.rs
```diff
@@ -20,90 +21,24 @@ pub fn new_http_client() -> std::sync::Arc<dyn HttpClient> {
...
     ///
-    /// This function will be the only one remaining in the trait as soon as the trait stabilizes.
-    /// It will be renamed to `execute_request`. The other helper functions (ie
-    /// `execute_request_check_status`) will be removed since the status check will be
-    /// responsibility of another policy (not the transport one). It does not consume the request.
-    /// Implementors are expected to clone the necessary parts of the request and pass them to the
-    /// underlying transport.
+    /// It does not consume the request. Implementors are expected to clone the necessary parts
+    /// of the request and pass them to the underlying transport.
     async fn execute_request2(&self, request: &crate::Request) -> crate::Result<crate::Response>;
```
bmc-msft, line 28: If we're entirely removing the old style of `execute_request`, can we rename this to `execute_request`?

sdk/core/src/http_client.rs
```diff
@@ -20,90 +21,24 @@ pub fn new_http_client() -> std::sync::Arc<dyn HttpClient> {
...
-
-    /// Send out a request and validate it was in the `2xx` range, using
-    /// `hyperium/http`'s types.
-    ///
-    /// Note: the `expected_status` parameter is never used, and instead we
-    /// always validate the status was in the `2xx` range. This method should
-    /// be considered deprecated, and should not be used in new code.
-    async fn execute_request_check_status(
```
bmc-msft, line 44: Why are you moving this trait function to a bare function?

cataggar: I guess it does not need to be. I'll move it back.

sdk/core/src/http_client.rs
cataggar: Yes, I'll rename it.

sdk/core/src/http_client.rs
```diff
@@ -20,90 +21,39 @@ pub fn new_http_client() -> std::sync::Arc<dyn HttpClient> {
...
-            Ok(response)
-        } else {
-            let body = response.into_body();
-            Err(ErrorKind::http_response_from_body(status.into(), &body).into())
+        request: &crate::Request,
+    ) -> crate::Result<crate::CollectedResponse> {
+        let rsp = self.execute_request(request).await?;
+        let (status, headers, body) = rsp.deconstruct();
+        let body = crate::collect_pinned_stream(body).await?;
+        if !status.is_success() {
```
cataggar, line 39: Need to double check this is the same range.

sdk/core/src/http_client.rs
```diff
@@ -20,90 +21,39 @@ pub fn new_http_client() -> std::sync::Arc<dyn HttpClient> {
...
-    ///
-    /// Note: the `expected_status` parameter is never used, and instead we
-    /// always validate the status was in the `2xx` range. This method should
-    /// be considered deprecated, and should not be used in new code.
+    /// Send out the request and collect the response body.
+    /// An error is returned if the status is not success.
     async fn execute_request_check_status(
         &self,
-        request: Request<Bytes>,
-        _expected_status: StatusCode,
```
cataggar, line 47: `_expected_status` was not used, so I removed it.

sdk/core/src/http_client.rs
```diff
@@ -20,90 +21,40 @@ pub fn new_http_client() -> std::sync::Arc<dyn HttpClient> {
...
-    async fn execute_request2(&self, request: &crate::Request) -> crate::Result<crate::Response>;
+    /// It does not consume the request. Implementors are expected to clone the necessary parts
+    /// of the request and pass them to the underlying transport.
+    async fn execute_request(&self, request: &crate::Request) -> crate::Result<crate::Response>;
 
-    /// Send out a request and validate it was in the `2xx` range, using
-    /// `hyperium/http`'s types.
-    ///
-    /// Note: the `expected_status` parameter is never used, and instead we
-    /// always validate the status was in the `2xx` range. This method should
```
bmc-msft, line 42: I have concerns regarding the implementation here.

While this comment says 2xx range, the original implementation checks 200..400 (which this PR matches) does not align with this comment.  More importantly, accepting a response within the 3xx range is probably not what the consumers of this API are likely to expect.

IMO, this is easily handled from the consumer side as:

`Response.status().is_success()`, or `Response.status() == HttpStatus::CREATED`

sdk/core/src/http_client.rs
```diff
@@ -20,90 +21,40 @@ pub fn new_http_client() -> std::sync::Arc<dyn HttpClient> {
...
-    /// be considered deprecated, and should not be used in new code.
```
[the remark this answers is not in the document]
cataggar: I was deprecated because it was using the original `execute_request`. It is no longer deprecated.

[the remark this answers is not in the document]
bmc-msft: The previous `execute_request` and `execute_request_check_status` were considered deprecated, and `execute_request2` (now renamed to `execute_request` is what we should be using.

Note this comment:

https://github.com/Azure/azure-sdk-for-rust/blob/edade075be4b0a0fdc8c644a6bb998b93e0046e7/sdk/core/src/http_client.rs#L30-L35

In my opinion, this comment was more in line with what we should be doing here.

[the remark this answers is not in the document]
cataggar: Okay, I’ll leave it deprecated with the reason as: status check will be responsibility of another policy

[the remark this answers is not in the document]
cataggar: 54d9430cc971687b5823a009fd25264a4b3c463f added the deprecated message back.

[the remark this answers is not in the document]
cataggar: Back to deprecated.

sdk/core/src/headers/mod.rs
```diff
@@ -65,6 +64,80 @@ impl Headers {
...
+            Error::with_message(ErrorKind::Other, || {
+                format!("header not found {}", key.as_str())
+            })
+        })
+    }
+
+    /// Get a header value as a str
+    pub fn get_as_str<T: Into<HeaderName>>(&self, key: T) -> Option<&str> {
+        self.get(key).map(|v| v.as_str())
+    }
```
rylev, line 81: I think we should just get rid of the variants of this methods and only have the variants which return a `crate::Result`. If you need an option, you can just call `.ok()` on the `Result`. 

And of course, then we just use the simpler naming convention and not the `or_err` names.

sdk/core/src/headers/mod.rs
```diff
@@ -65,6 +64,80 @@ impl Headers {
...
+        let key = key.into();
+        let v = self.get_or_err(key.clone())?;
+        let v = v.as_str();
+        v.parse::<u64>()
+            .with_context(ErrorKind::DataConversion, || {
+                format!("unable to parse header into u64 {key:?}: {v}")
+            })
+    }
+
+    pub fn get_as_enum<T: Into<HeaderName>, V: FromStr<Err = E>, E>(
```
rylev, line 123: This is a bit more general than for enums. Can we call this `get_as_parsed` or something like that? I think `get_as_u64_or_err`  can be implemented using this function.

sdk/core/src/options.rs
```diff
@@ -33,6 +33,7 @@ pub struct ClientOptions {
     /// Telemetry options.
     pub(crate) telemetry: TelemetryOptions,
     /// Transport options.
+    #[allow(dead_code)] // wasm
```
rylev, line 36: I don't love these... I wish we could just make the code conditionally compile based on features.

sdk/core/src/error/http_error.rs
```diff
@@ -76,9 +75,8 @@ fn get_error_code_from_header(response: &Response) -> Option<String> {
     Some(
         response
             .headers()
-            .get(http::header::HeaderName::from_static("x-ms-error-code"))?
-            .to_str()
-            .ok()?
+            .get("x-ms-error-code")?
+            .as_str()
```
rylev, line 79:
```suggestion
            .get_as_str("x-ms-error-code")?;
```

sdk/core/src/response.rs
```diff
@@ -107,3 +76,38 @@ pub async fn collect_pinned_stream(mut pinned_stream: PinnedStream) -> crate::er
...
+impl CollectedResponse {
+    /// Create a new instance
+    pub fn new(status: StatusCode, headers: Headers, body: Bytes) -> Self {
+        Self {
+            status,
+            headers,
+            body,
+        }
+    }
+    /// Get the status
```
rylev, line 97: Nit: can we add spaces between these methods?

sdk/data_tables/src/continuation_next_partition_and_row_key.rs
```diff
@@ -18,24 +15,14 @@ impl ContinuationNextPartitionAndRowKey {
...
-            .map(|item| item.to_str())
-            .transpose()
-            .context(
-                ErrorKind::DataConversion,
-                "failed to convert x-ms-continuation-NextPartitionKey header value to string",
-            )?;
+            .map(|item| item.as_str());
 
         let row_header_as_str = headers
             .get("x-ms-continuation-NextRowKey")
```
rylev, line 24:
```suggestion
            .get_as_str("x-ms-continuation-NextRowKey")
```

rylev (approved):

sdk/core/src/headers/mod.rs
cataggar: Sounds good. I added as #837.

sdk/core/src/options.rs
cataggar: Noted. I'll try to reduce these in future changes.

sdk/core/src/headers/mod.rs
cataggar: Also captured in #837.

sdk/data_tables/src/requests/delete_entity_builder.rs
```diff
@@ -38,41 +35,31 @@ impl<'a> DeleteEntityBuilder<'a> {
...
-        let request = add_optional_header(&self.client_request_id, request);
-        let request = request.header("Accept", "application/json;odata=minimalmetadata");
-        let request = request.header("If-Match", "*");
+        let mut request = Request::new(url.clone(), Method::DELETE);
+        request.add_optional_header(&self.client_request_id);
+        request.insert_header("Accept", "application/json;odata=minimalmetadata");
+        request.insert_header("If-Match", "*");
 
-        let request = request.body("".to_owned())?;
+        request.set_body("");
```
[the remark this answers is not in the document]
cataggar: I'm not sure.

sdk/data_tables/src/responses/insert_entity_response.rs
```diff
@@ -22,35 +20,31 @@ where
...
-            common_storage_response_headers: response.headers().try_into()?,
-            etag: etag_from_headers(response.headers())?.into(),
-            location: response
-                .headers()
-                .get_as_str("location")
-                .map(Url::parse)
+            common_storage_response_headers: headers.try_into()?,
+            etag: etag_from_headers(headers)?.into(),
+            location: headers
+                .get("location")
```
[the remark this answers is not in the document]
cataggar: No `get_as_parsed` yet. Would be nice.

bmc-msft (approved):

PR #838, opened 2022-06-21
move storage_blob operations to follow the into_future pattern

bmc-msft:
Changes beyond the into_future pattern updates:
1. This updates ContentDisposition, ContentEncoding, ContentLanguage, and ContentType to follow other header types.
2. This adds a `BlobClient.get_content` helper that iterates over `BlobClient.get().stream()` and writes the data into a single Vec, returning the Vec.  This was an extremely common pattern in the examples and something akin to it is available in most of the other SDKs as well.
3. Makes `ContainerClient.list_blobs`, `BlobServiceClient.list_containers`, and `BlobClient.get` only return streams.  Having reviewed a non-trivial amount of code that assumed that list_blobs.execute returned the full set, I think this is a footgun we should avoid.

### DIFF
--- a/sdk/storage_blobs/src/blob/operations/append_block.rs
+++ b/sdk/storage_blobs/src/blob/operations/append_block.rs
@@ -18,7 +18,7 @@ impl AppendBlockBuilder {
     pub(crate) fn new(blob_client: BlobClient, body: Bytes) -> Self {
         Self {
             blob_client,
-            body: body.into(),
+            body,
             hash: None,
             condition_max_size: None,
             condition_append_position: None,


sdk/core/src/request_options/content_encoding.rs
```diff
@@ -1,23 +1,48 @@
...
-    S: Into<&'a str>,
-{
-    fn from(s: S) -> Self {
-        Self(s.into())
+impl ContentEncoding {
+    pub fn as_str(&self) -> &str {
+        self.0.as_ref()
+    }
+
+    const fn from_static(s: &'static str) -> Self {
```
cataggar, line 11: `pub` please

cataggar (approved):
This looks great!

sdk/core/src/request_options/content_language.rs
```diff
@@ -1,23 +1,48 @@
...
+    pub fn as_str(&self) -> &str {
+        self.0.as_ref()
+    }
+
+    const fn from_static(s: &'static str) -> Self {
+        Self(std::borrow::Cow::Borrowed(s))
+    }
+}
+
+impl From<&'static str> for ContentLanguage {
```
cataggar, line 16: I'm new to `const fn`. Would it be better for remove this and require the use of `ContentLanguage::from_static` which is a `const fn`? In general, are `const fn` able to provide optimations?

sdk/core/src/request_options/content_language.rs
```diff
@@ -1,23 +1,48 @@
...
-    S: Into<&'a str>,
-{
-    fn from(s: S) -> Self {
-        Self(s.into())
+impl ContentLanguage {
+    pub fn as_str(&self) -> &str {
+        self.0.as_ref()
+    }
+
+    const fn from_static(s: &'static str) -> Self {
```
cataggar, line 11: `pub` please, I think

sdk/storage_blobs/examples/blob_01.rs
```diff
@@ -22,18 +23,11 @@ async fn main() -> azure_core::Result<()> {
...
-        .include_metadata(true)
-        .include_snapshots(true)
-        .include_uncommitted_blobs(true)
-        .execute()
-        .await?;
-
-    let result = blob_client.get().execute().await?;
-
+    // only get the first 8k chunk
+    let result = Box::pin(blob_client.get().stream(1024 * 8))
```
rylev, line 27: Is the `Box::pin` necessary here? I believe that `Pageable` streams are already `Unpin`.

sdk/storage_blobs/src/blob/operations/append_block.rs
```diff
@@ -0,0 +1,78 @@
...
+    hash: Option<Hash>,
+    condition_max_size: Option<ConditionMaxSize>,
+    condition_append_position: Option<ConditionAppendPosition>,
+    lease_id: Option<LeaseId>,
+    client_request_id: Option<ClientRequestId>,
+    timeout: Option<Timeout>,
+}
+
+impl AppendBlockBuilder {
+    pub(crate) fn new(blob_client: BlobClient, body: impl Into<Bytes>) -> Self {
```
rylev, line 18: We normally keep these constructors non-generic and instead do the conversion inside the client operation method.

sdk/storage_blobs/src/blob/operations/append_block.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[derive(Debug, Clone)]
+pub struct AppendBlockBuilder {
+    blob_client: BlobClient,
+    body: Bytes,
+    hash: Option<Hash>,
+    condition_max_size: Option<ConditionMaxSize>,
+    condition_append_position: Option<ConditionAppendPosition>,
+    lease_id: Option<LeaseId>,
+    client_request_id: Option<ClientRequestId>,
```
rylev, line 13: Same as above

sdk/storage_blobs/src/blob/operations/break_lease.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    pub fn into_future(self) -> Response {
+        Box::pin(async move {
+            let mut url = self.blob_client.url_with_segments(None)?;
+
+            url.query_pairs_mut().append_pair("comp", "lease");
+            self.timeout.append_to_url_query(&mut url);
+
+            trace!("url == {:?}", url);
```
rylev, line 39: Can we remove these traces?

sdk/storage_blobs/src/blob/operations/get_blob.rs
```diff
@@ -58,12 +38,12 @@ impl<'a> GetBlobBuilder<'a> {
...
         blob_versioning: &'a BlobVersioning => Some(blob_versioning),
-        client_request_id: ClientRequestId<'a> => Some(client_request_id),
+        client_request_id: &'a ClientRequestId => Some(client_request_id),
         timeout: Timeout => Some(timeout),
-        lease_id: &'a LeaseId => Some(lease_id),
+        lease_id: LeaseId => Some(lease_id),
     }
 
-    pub async fn execute(&self) -> azure_core::Result<GetBlobResponse> {
+    async fn execute(&self) -> azure_core::Result<GetBlobResponse> {
```
rylev, line 46: Is there a reason this wasn't moved to using the pipeline architecture?

sdk/storage_blobs/src/blob/operations/get_metadata.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub server: String,
+    pub date: DateTime<Utc>,
+    pub metadata: Metadata,
+}
+
+impl TryFrom<&Headers> for GetMetadataResponse {
+    type Error = crate::Error;
+
+    fn try_from(headers: &Headers) -> Result<Self, Self::Error> {
+        debug!("headers == {:#?}", headers);
```
rylev, line 71:
```suggestion
```

sdk/storage_blobs/src/blob/operations/get_properties.rs
```diff
@@ -0,0 +1,92 @@
...
+            request.add_optional_header(&self.client_request_id);
+
+            let response = self
+                .blob_client
+                .http_client()
+                .execute_request_check_status(&request)
+                .await?;
+
+            // TODO: Fix this
+            //let blob = Blob::from_headers(&blob_name, &container_name, snapshot_time, &headers)?;
```
rylev, line 52: Can we get more context on this?

sdk/storage_blobs/src/blob/operations/mod.rs
```diff
@@ -0,0 +1,50 @@
+mod acquire_lease;
+pub use acquire_lease::*;
```
rylev, line 2: nit: in other SDKs we do the mods first and then the `pub use` statements

rylev (changes requested):
Looks good! I had some small feedback though. I was also wondering why some of the operations were not fully converted over to the pipeline.

sdk/storage_blobs/src/blob/operations/get_properties.rs
bmc-msft: While I used `git mv` to rename all of the builders to operations, then appended the results, git does not show these as being renamed and therefore keeping the previous history.

I don't have context on this, it's a holdover from the previous implementation.

sdk/storage_blobs/src/blob/operations/get_metadata.rs
bmc-msft: (removed all the other similar messages in the process)

sdk/storage_blobs/src/blob/operations/get_blob.rs
bmc-msft: Note, this didn't move everything to the pipeline architecture, just into_future.  There is more work to be done to enable that, primarily exposing a Context.

Note, this is still a hold over because Pageable isn't used yet, as this doesn't use pipelines yet.

sdk/storage_blobs/examples/blob_01.rs
bmc-msft: This will get addressed in a follow-on PR moving to Pipelines (and therefore Pageable)

sdk/storage_blobs/src/blob/operations/acquire_lease.rs
```diff
@@ -0,0 +1,86 @@
...
+};
+use chrono::{DateTime, Utc};
+
+#[derive(Debug, Clone)]
+pub struct AcquireLeaseBuilder {
+    blob_client: BlobClient,
+    lease_duration: LeaseDuration,
+    lease_id: Option<LeaseId>,
+    proposed_lease_id: Option<ProposedLeaseId>,
+    client_request_id: Option<ClientRequestId>,
```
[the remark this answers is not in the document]
bmc-msft: Addressed when this moves to pipelines

sdk/storage_blobs/src/blob/operations/append_block.rs
bmc-msft: Addressed when this moves to pipelines

sdk/storage_blobs/src/blob/operations/break_lease.rs
```diff
@@ -0,0 +1,77 @@
+use crate::prelude::*;
+use azure_core::{headers::*, prelude::*, RequestId};
+use chrono::{DateTime, Utc};
+
+#[derive(Debug, Clone)]
+pub struct BreakLeaseBuilder {
+    blob_client: BlobClient,
+    lease_break_period: Option<LeaseBreakPeriod>,
+    lease_id: Option<LeaseId>,
+    client_request_id: Option<ClientRequestId>,
```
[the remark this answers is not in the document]
bmc-msft: Addressed when this moves to pipelines

sdk/storage_blobs/src/blob/operations/clear_page.rs
```diff
@@ -0,0 +1,95 @@
...
+use chrono::{DateTime, Utc};
+
+#[derive(Debug, Clone)]
+pub struct ClearPageBuilder {
+    blob_client: BlobClient,
+    ba512_range: BA512Range,
+    sequence_number_condition: Option<SequenceNumberCondition>,
+    if_modified_since_condition: Option<IfModifiedSinceCondition>,
+    if_match_condition: Option<IfMatchCondition>,
+    client_request_id: Option<ClientRequestId>,
```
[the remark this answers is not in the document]
bmc-msft: Addressed when this moves to pipelines

rylev (approved):
🎉

PR #839, opened 2022-06-21
header name constants changed from `&str` to `HeaderName`

ctaggart:
- header name constants changed from `&str` to `azure_core::headers::HeaderName`
- replaced types from `http:header` with `azure_core::headers` types
- create `const fn` `HeaderName::from_static`
- `Header::get` takes a `&HeaderName` instead of `Into<HeaderName>`
- similar changes for `HeaderValue`
- removed `Request::headers_mut`


sdk/storage_datalake/tests/files.rs
```diff
@@ -77,7 +77,8 @@ async fn file_upload() -> azure_core::Result<()> {
     Ok(())
 }
 
-#[tokio::test]
+// #[tokio::test]
+#[allow(dead_code)]
```
cataggar, line 81: I need help figuring out why this one mock test for storage_datalake is failing.

sdk/core/src/request_options/content_language.rs
```diff
@@ -33,7 +33,7 @@ impl From<&String> for ContentLanguage {
 
 impl Header for ContentLanguage {
     fn name(&self) -> headers::HeaderName {
-        http::header::CONTENT_TYPE.into()
```
cataggar, line 36: Was this a bug in a recent commit?

sdk/storage_datalake/tests/files.rs
cataggar: > received request have header 'range' but it was not present in the read request

sdk/core/src/request_options/range.rs
```diff
@@ -28,9 +28,12 @@ impl AsHeaders for Range {
     type Iter = std::vec::IntoIter<(HeaderName, HeaderValue)>;
 
     fn as_headers(&self) -> Self::Iter {
-        let mut headers = vec![("x-ms-range".into(), format!("{}", self).into())];
+        let mut headers = vec![(headers::RANGE, format!("{}", self).into())];
```
cataggar, line 31: should be MS_RANGE

cataggar: This was the bug caught by the test.

bmc-msft (approved):

PR #840, opened 2022-06-22
add `#[must_use]` across the SDK

bmc-msft:
adding `#[must_use]` to identify inadvertent footguns in the SDK.

This comes about as a continued attempt to automation prevent issues found during a review of code that uses this SDK.

### DIFF
--- a/sdk/core/src/auth.rs
+++ b/sdk/core/src/auth.rs
@@ -8,9 +8,11 @@ use std::fmt::Debug;
 pub struct AccessToken(String);
 
 impl AccessToken {
+    #[must_use]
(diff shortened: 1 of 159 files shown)


bmc-msft:
I was reviewing code that had a handful of places were using code that implemented the builder pattern, but they did not understand that the builder then had to be used.

The simplest fix was to slap `#[must_use]` on almost every single function using `clippy::must_use_candidate`.  :)

rylev:
Could we start by just doing the operation methods on clients? That seems pretty uncontroversial. Then we can add new classes of places in follow up PRs. This seems a little too willy-nilly for my taste.

PR #843, opened 2022-06-22
Move storage blobs to pipeline architecture

bmc-msft:


bmc-msft:
Note, this does *not* move `get_blob`, as that implements paging by shifting the Range for the blob.  The current Pageable support only supports Continuation as a String, while this needs the current offset.

sdk/storage/src/core/clients/storage_account_client.rs
```diff
@@ -8,11 +8,12 @@ use crate::{
...
-use azure_core::error::{Error, ErrorKind, ResultExt};
-use azure_core::Request;
-use azure_core::{headers::*, Pipeline};
-use azure_core::{ClientOptions, HttpClient};
+use azure_core::{
+    auth::TokenCredential,
+    error::{Error, ErrorKind, ResultExt},
+    headers::*,
+    ClientOptions, Context, HttpClient, Pipeline, Request, Response,
+};
```
rylev, line 16: Nit: we don't seem to have a consistent way of structuring imports. I'd like to figure out a consistent way of structuring these things and stick to that across all of the SDKs.

sdk/storage_blobs/examples/container_00.rs
```diff
@@ -23,30 +23,26 @@ async fn main() -> azure_core::Result<()> {
...
+        .list_containers()
+        .max_results(max_results)
+        .into_stream();
 
     let mut count = 0;
     while let Some(result) = iv.next().await {
         let container = result?;
-        count += container.incomplete_vector.len();
-        for container in container.incomplete_vector.iter() {
+        count += container.containers.len();
```
rylev, line 34: nit: `container.containers` reads a bit weird. Perhaps the `container` binding should be renamed to `page` or `response`?

sdk/storage_blobs/examples/list_blobs_00.rs
```diff
@@ -22,16 +22,14 @@ async fn main() -> azure_core::Result<()> {
     let blob_service = storage_client.as_blob_service_client();
     let container_client = storage_client.as_container_client(&container_name);
 
-    let iv = Box::pin(blob_service.list_containers().stream())
+    let iv = blob_service
```
rylev, line 25: nit: rename `iv` to something else: `response` or `page`?

sdk/storage_blobs/src/blob/operations/put_block.rs
```diff
@@ -9,14 +9,13 @@ use azure_storage::{headers::consistency_from_headers, ConsistencyCRC64, Consist
...
 
-#[derive(Debug, Clone)]
 pub struct PutBlockBuilder {
     blob_client: BlobClient,
     block_id: BlockId,
     body: Bytes,
     #[allow(unused)]
     hash: Option<Hash>,
-    client_request_id: Option<ClientRequestId>,
+    context: Context,
```
rylev, line 18: nit: we put `context` last. Might be nice to keep that consistent.

sdk/storage_blobs/src/blob/operations/put_block_blob.rs
```diff
@@ -22,7 +22,7 @@ pub struct PutBlockBlobBuilder {
     access_tier: Option<AccessTier>,
     // TODO: Support tags
     lease_id: Option<LeaseId>,
-    client_request_id: Option<ClientRequestId>,
+    context: Context,
```
rylev, line 25: Same here, and there's other places as well.

sdk/storage_blobs/src/clients/blob_service_client.rs
```diff
@@ -28,6 +31,16 @@ impl BlobServiceClient {
     }
 
     pub fn list_containers(&self) -> crate::container::operations::ListContainersBuilder {
-        crate::container::operations::ListContainersBuilder::new(&self.storage_client)
+        crate::container::operations::ListContainersBuilder::new(self.clone())
```
rylev, line 34: Can we use the unqualified name?

sdk/storage_blobs/examples/list_containers_and_blobs.rs
```diff
@@ -0,0 +1,43 @@
...
+        .expect("please specify the account name as first command line parameter");
+
+    let account_key =
+        std::env::var("STORAGE_ACCOUNT_KEY").expect("Set env variable STORAGE_ACCOUNT_KEY first!");
+
+    let http_client = azure_core::new_http_client();
+    let storage_account_client =
+        StorageAccountClient::new_access_key(http_client, &account, &account_key);
+
+    let blob_service_client = storage_account_client.as_blob_service_client();
```
rylev, line 21: In Cosmos, we would use the naming convention  `storage_account_client.blob_client()`. This was discussed in [a previous PR](https://github.com/Azure/azure-sdk-for-rust/pull/676#issuecomment-1062310005)

rylev (approved):
*Reviewed together with @yoshuawuyts 🧡* 

There are some small things that we think we should change, but feel free to do those in a follow up PR if you want.

sdk/storage/src/core/clients/storage_account_client.rs
bmc-msft: Agreed, preferably using [imports_granularity](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#imports_granularity) once it stabilizes.

sdk/storage_blobs/examples/list_containers_and_blobs.rs
bmc-msft: I'll do this change as a follow-up PR, as I'd like feedback on it uniquely.

PR #844, opened 2022-06-22
align storage key name to documentation

bmc-msft:
Azure documentation refers storage account keys as "access keys".

https://docs.microsoft.com/en-us/azure/storage/common/storage-account-keys-manage?tabs=azure-portal#regenerate-access-keys


cataggar (approved):
💖

Issue #848, opened 2022-06-23
trouble mapping TryFrom errors


cataggar:
I'd like to be able to have the error be the source and have a formatted message.

``` rust
impl TryFrom<u16> for StatusCode {
    type Error = Error;
    fn try_from(value: u16) -> crate::Result<Self> {
        Ok(StatusCode(
            http_types::StatusCode::try_from(value).with_context(ErrorKind::DataConversion, || {
                format!("StatusCode::try_from failed for '{value}'")
            }),
        ))
    }
}
```

results in this error:
```
error[E0599]: the method `with_context` exists for enum `std::result::Result<http_types::StatusCode, http_types::Error>`, but its trait bounds were not satisfied
   --> sdk\core\src\status_code.rs:50:53
    |
50  |             http_types::StatusCode::try_from(value).with_context(ErrorKind::DataConversion, || {
    |                                                     ^^^^^^^^^^^^ method cannot be called on `std::result::Result<http_types::StatusCode, http_types::Error>` due to unsatisfied trait bounds
    |
   ::: C:\Users\cataggar\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib/rustlib/src/rust\library\core\src\result.rs:504:1
    |
504 | pub enum Result<T, E> {
    | --------------------- doesn't satisfy `_: error::ResultExt<http_types::StatusCode>`
    |
   ::: C:\Users\cataggar\.cargo\registry\src\github.com-1ecc6299db9ec823\http-types-2.12.0\src\error.rs:16:1
    |
16  | pub struct Error {
    | ---------------- doesn't satisfy `http_types::Error: StdError`
    |
note: trait bound `http_types::Error: StdError` was not satisfied
   --> sdk\core\src\error\mod.rs:285:8
    |
283 | impl<T, E> ResultExt<T> for std::result::Result<T, E>
    |            ------------     -------------------------
284 | where
285 |     E: std::error::Error + Send + Sync + 'static,
    |        ^^^^^^^^^^^^^^^^^ unsatisfied trait bound introduced here
```

---

This works, but `full` does not accept a formatted message. Should a `Error::with_full` be added?
``` rust
impl TryFrom<u16> for StatusCode {
    type Error = Error;
    fn try_from(value: u16) -> crate::Result<Self> {
        Ok(StatusCode(
            http_types::StatusCode::try_from(value).map_err(|error| {
                Error::full(
                    ErrorKind::DataConversion,
                    error,
[...]

PR #849, opened 2022-06-23
add `Method` & `StatusCode` to azure_core

ctaggart:
This removes the last dependency on the `http` crate by adding a `Method` and `StatusCode` types.  I'm interested in having compatibility with [http-types](https://github.com/http-rs/http-types) and supporting [surf](https://github.com/http-rs/surf) in addition to `reqwest`. This draft wraps `http-types` implementations of `Method` and `StatusCode`, but exposes the API from `http` crate that we are using. There are a few different ways we could go here. Looking for feedback.


sdk/core/src/http_client/reqwest.rs
```diff
@@ -0,0 +1,84 @@
...
+
+fn try_from_status(status: ::reqwest::StatusCode) -> crate::Result<crate::StatusCode> {
+    let status = status.as_u16();
+    http_types::StatusCode::try_from(status)
+        .map_err(|_| {
+            Error::with_message(ErrorKind::DataConversion, || {
+                format!("invalid status code {status}")
+            })
+        })
+        .map(crate::StatusCode)
```
rylev, line 83: We should use `crate::StatusCode`'s `TryFrom<u16>` impl

sdk/core/src/method.rs
```diff
@@ -0,0 +1,46 @@
...
+#[derive(PartialEq, Eq, Clone, Debug)]
+pub struct Method(pub(crate) http_types::Method);
+
+impl Method {
+    pub fn as_str(&self) -> &str {
+        self.as_ref()
+    }
+}
+
+impl Deref for Method {
```
rylev, line 13: This puts `http_types` into the public API. Do we want that?

sdk/core/src/status_code.rs
```diff
@@ -0,0 +1,59 @@
...
+        StatusCode(http_types::StatusCode::InternalServerError);
+    pub const NOT_MODIFIED: StatusCode = StatusCode(http_types::StatusCode::NotModified);
+    pub const OK: StatusCode = StatusCode(http_types::StatusCode::Ok);
+    pub const REQUEST_TIMEOUT: StatusCode = StatusCode(http_types::StatusCode::RequestTimeout);
+    pub const SERVICE_UNAVAILABLE: StatusCode =
+        StatusCode(http_types::StatusCode::ServiceUnavailable);
+    pub const TOO_MANY_REQUESTS: StatusCode = StatusCode(http_types::StatusCode::TooManyRequests);
+}
+
+impl Default for StatusCode {
```
rylev, line 40: Not sure I like this impl. `OK` does not logically feel like the "default" status code to me.

sdk/data_tables/src/transaction.rs
```diff
@@ -40,7 +40,7 @@ impl Transaction {
         for transaction_operation in self.transaction_operations.iter() {
             s.push_str(&format!("--changeset_{}\nContent-Type: application/http\nContent-Transfer-Encoding: binary\n\n", self.change_set_uuid.hyphenated()));
             s.push_str(&format!(
-                "{} {} HTTP/1.1\n",
+                "{:?} {} HTTP/1.1\n",
```
rylev, line 43: Not sure I like that `method` doesn't have a display. The string representations of the HTTP methods are cononical no?

rylev (changes requested):

sdk/core/src/http_client/reqwest.rs
```diff
@@ -0,0 +1,84 @@
+use super::*;
+use std::{collections::HashMap, str::FromStr};
+
+/// Construct a new `HttpClient` with the `reqwest` backend.
+pub fn new_http_client() -> std::sync::Arc<dyn HttpClient> {
+    std::sync::Arc::new(::reqwest::Client::new())
```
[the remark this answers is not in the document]
cataggar: I put these in a module named reqwest, so needed that prefix. Could rename the module.

sdk/data_tables/Cargo.toml
```diff
@@ -18,9 +18,8 @@ azure_storage = { path = "../storage", version = "0.4", default-features=false,
 bytes = "1.0"
 chrono = { version = "0.4", features = ["serde"] }
 futures = "0.3"
-http = "0.2"
 log = "0.4"
-once_cell = "1.0"
+# once_cell = "1.0"
```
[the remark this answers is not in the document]
cataggar: Can remove. http crate doesn’t have Mehod::MERGE.

sdk/data_tables/src/transaction.rs
yoshuawuyts: > The string representations of the HTTP methods are cononical no?

Correct. As per [RFC 9110, 9: Methods](https://www.rfc-editor.org/rfc/rfc9110#name-methods):

> The method token is case-sensitive because it might be used as a gateway to object-based systems with case-sensitive method names. By convention, standardized methods are defined in all-uppercase US-ASCII letters.

sdk/core/src/status_code.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub const GATEWAY_TIMEOUT: StatusCode = StatusCode(http_types::StatusCode::GatewayTimeout);
+    pub const INTERNAL_SERVER_ERROR: StatusCode =
+        StatusCode(http_types::StatusCode::InternalServerError);
+    pub const NOT_MODIFIED: StatusCode = StatusCode(http_types::StatusCode::NotModified);
+    pub const OK: StatusCode = StatusCode(http_types::StatusCode::Ok);
+    pub const REQUEST_TIMEOUT: StatusCode = StatusCode(http_types::StatusCode::RequestTimeout);
+    pub const SERVICE_UNAVAILABLE: StatusCode =
+        StatusCode(http_types::StatusCode::ServiceUnavailable);
+    pub const TOO_MANY_REQUESTS: StatusCode = StatusCode(http_types::StatusCode::TooManyRequests);
+}
```
yoshuawuyts, line 38: To be honest, I don't like this construct much. In my opinion status codes are best represented in Rust as a non-exhaustive enum.

I feel like we'd be better off either re-exporting `http_types::Method` as-is, or if we don't want to rely on using `http-types` in our public API, either copying over the implementation - or creating a wrapper/type-alias to it.

yoshuawuyts (commented):
I feel like I'm missing some context about direction. Obviously as the lead author on `http-types` I'm thrilled by the prospect that we may want to use it for the azure SDK. But I want to make sure that we properly understand and discuss the tradeoffs of doing so. Because if we're going to adopt it, I want to make sure we're doing so for the right reasons.

For example: should we expose `http-types` types from the Rust Azure SDK's public API? If we do that, how do we feel about the governance? Are there any reasons beyond the code changes in this API why we may want to adopt this?

I think it would be worth laying out a vision of how you see the integration working out in the longer term. And then maybe we can discuss it offline here / perhaps bring it to a meeting to discuss further?

cataggar:
We'll discuss this in our next meeting. A purpose of this PR was to show the differences between `http` and `http-type` implementations and start a discussion. I prefer `http-types` representations of both `Method` & `StatusCode` as enums.

https://docs.rs/http-types/latest/src/http_types/method.rs.html
``` rust
#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
pub enum Method {
```

https://docs.rs/http-types/latest/src/http_types/status_code.rs.html

``` rust
#[repr(u16)]
#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
pub enum StatusCode {
```

`http` instead has consts for all the values.

```
#[derive(Clone, PartialEq, Eq, Hash)]
pub struct Method(Inner);

enum Inner {
    Options,
    Get,
```

https://docs.rs/http/latest/src/http/status.rs.html
```
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
pub struct StatusCode(NonZeroU16);
```


Differences
- The biggest change needed for http-types is error the need for its error type to implement `std::error::Error` #848 https://github.com/http-rs/http-types/pull/395. 
- `http::StatusCode` has a `fn as_u16`, http-types is missing it and uses `as u16` everywhere. The convenience function would be nice.
- `http::StatusCode` has a default `StatusCode::OK`. Some of our Default implementations use this. Should be easy to figure out.
- Because http-types uses an enum instead of constants, the name casing is different. Please `BAD_GATEWAY` with `BadGateway` and the 7 other status codes we use. This is just a few global replaces.

I created #852 with a subset of changes that leave `http` in place for now.

PR #850, opened 2022-06-24
Move get blob to pageable

bmc-msft:
This PR does two main things:

1. Extends the `Continuation` model to support String (as used by most of the SDKs) and `Range` as used by `get_blob`.
2. Moves the `get_blob` to not require the user to specify a chunk size, but by default will download in 1MB chunks.  I think this simplifies the interface while also not shooting the user in the foot by default downloading the entire blob into memory.


bmc-msft:
Note, the Continuation code for all of the generated crates will need to be updated, though the update is minimal.  I will do that tomorrow.

sdk/data_cosmos/src/operations/list_collections.rs
```diff
@@ -116,7 +116,7 @@ impl ListCollectionsResponse {
 }
 
 impl Continuable for ListCollectionsResponse {
-    fn continuation(&self) -> Option<String> {
-        self.continuation_token.clone()
+    fn continuation(&self) -> Option<Continuation> {
+        self.continuation_token.clone().map(|x| x.into())
```
rylev, line 120: This might be ever so clearer:
```suggestion
        self.continuation_token.clone().map(Continuation::from)
```

sdk/storage_blobs/src/blob/operations/get_blob.rs
```diff
@@ -1,125 +1,129 @@
...
     range: Option<Range>,
-    blob_versioning: Option<&'a BlobVersioning>,
-    client_request_id: Option<&'a ClientRequestId>,
+    blob_versioning: Option<BlobVersioning>,
     timeout: Option<Timeout>,
     lease_id: Option<LeaseId>,
+    chunk_size: u64,
+    context: Context,
+}
+
```
rylev, line 25: Can we move these helper functions towards the end? It's nice to not have to scroll to see the `impl` block

sdk/storage_blobs/src/blob/operations/get_blob.rs
```diff
@@ -1,125 +1,129 @@
...
+    match request_range {
+        Some(range) => {
+            let len = std::cmp::min(range.len(), chunk_size);
+            Range::new(range.start, range.start + len)
+        }
+        None => Range::new(0, chunk_size),
+    }
+}
+
+fn remaining_range(
```
rylev, line 36: Can you document what this function is doing? The docs inside the function are helpful but having a summary would also be nice.

sdk/core/src/pageable.rs
```diff
@@ -82,12 +84,12 @@ impl<T, O> std::fmt::Debug for Pageable<T, O> {
 
 /// A type that can yield an optional continuation token
 pub trait Continuable {
-    fn continuation(&self) -> Option<String>;
+    fn continuation(&self) -> Option<Continuation>;
```
rylev, line 87: Do any endpoints allow for multiple types of continuations? It seems like right now the `Continuation` header with a continuation token is the happy path and we're using panics to ensure that we don't use a `Continuation::Range` when we don't meant time.

What if we were to trying the following;

```rust
pub trait Continuable {
   type Continuation: crate::AsHeaders;
    fn continuation(&self) -> Option<Self::Continuation>;
}
```

The `make_request` function then becomes `impl Fn(Option<T::Continuation>) -> F + Clone + 'static + Send`

This would make the system more robust. Every continuable response just needs to provide the headers it uses for specifying how it continues.

rylev (changes requested):

sdk/core/src/pageable.rs
bmc-msft: Examples of different Continuation methods:

* BlobClient.get uses Range header
* Cosmos's list_attachments uses `Continuation` header
* ContainerClient.list_blobs uses `marker` URI parameter
* datalake's file_systems_list uses `maker` URI parameter.  Note, this one also accepts next_marker from the builder, which probably shouldn't be available.

sdk/storage_blobs/src/blob/operations/get_blob.rs
```diff
@@ -173,6 +129,62 @@ impl Continuable for GetBlobResponse {
     }
 }
 
+// caclculate the first Range for use at the beginning of the Pageable.
```
rylev, line 130: FYI: even though these are internal APIs they should still be `///` doc comments so that they show up in rust-analyzer.

rylev (approved):

PR #852, opened 2022-06-25
add `Method` & `StatusCode` to azure_core

cataggar:
After getting feedback on #849, this is a subset of that change.
- It does not change the `StatusCode` or `Method` being used. They are still from `http` in this pull request.


sdk/core/src/error/http_error.rs
```diff
@@ -1,4 +1,4 @@
-use crate::{headers, Response};
+use crate::{headers, response, Response, INVALID_BODY};
```
rylev, line 1: nit: I'm not sure exposing `INVALID_BODY` from the top of the crate is the right thing to do. It's used rarely enough that keeping it in a submodule would make sense IMO.

sdk/core/src/error/http_error.rs
```diff
@@ -16,19 +16,21 @@ impl HttpError {
...
-        let status = response.status();
         let mut error_code = get_error_code_from_header(&response);
-        let mut headers = HashMap::new();
+        let (status, headers, body) = response.deconstruct();
+        let mut error_headers = HashMap::new();
 
-        for (name, value) in response.headers().iter() {
-            headers.insert(name.as_str().to_string(), value.as_str().to_string());
+        for (name, value) in headers.iter() {
+            error_headers.insert(name.as_str().to_string(), value.as_str().to_string());
```
rylev, line 24:
```rust
let headers: HashMap<String, String> =  headers.iter().map(|(name, value)| (name.as_str().to_owned(), value.as_str().to_owned())).collect();
```

sdk/core/src/http_client/reqwest.rs
```diff
@@ -0,0 +1,65 @@
...
+            req = req.header(name.as_str(), value.as_str());
+        }
+
+        let body = request.body().clone();
+
+        let reqwest_request = match body {
+            Body::Bytes(bytes) => req
+                .body(bytes)
+                .build()
+                .context(ErrorKind::Other, "failed to build request")?,
```
rylev, line 24: Nit: can we move this outside this match so that we don't have to repeat it in both arms

sdk/core/src/http_client/reqwest.rs
```diff
@@ -0,0 +1,65 @@
...
+            Body::SeekableStream(mut seekable_stream) => {
+                seekable_stream.reset().await.unwrap(); // TODO: remove unwrap when `HttpError` has been removed
+                req.body(::reqwest::Body::wrap_stream(seekable_stream))
+                    .build()
+                    .context(ErrorKind::Other, "failed to build request")?
```
rylev, line 29:
```suggestion
                    .context(ErrorKind::Other, "failed to build `reqwest` request")?
```

sdk/core/src/http_client/reqwest.rs
```diff
@@ -0,0 +1,65 @@
...
+            .await
+            .context(ErrorKind::Io, "failed to execute request")?;
+
+        let status = rsp.status();
+        let headers = to_headers(rsp.headers())?;
+        let body: PinnedStream = Box::pin(rsp.bytes_stream().map_err(|error| {
+            Error::full(
+                ErrorKind::Io,
+                error,
+                "error converting `reqwest` request into a byte stream",
```
rylev, line 44:
```suggestion
                "error converting `reqwest` response into a byte stream",
```

sdk/core/src/http_client/reqwest.rs
```diff
@@ -0,0 +1,65 @@
...
+        Ok(crate::Response::new(status, headers, body))
+    }
+}
+
+fn to_headers(map: &::reqwest::header::HeaderMap) -> crate::Result<crate::headers::Headers> {
+    let map = map
+        .iter()
+        .map(|(k, v)| {
+            let key = k.as_str();
+            let value = std::str::from_utf8(v.as_bytes()).expect("non-UTF8 header value");
```
rylev, line 57: I think we probably shouldn't panic here. Maybe for now, we should just skip headers whose values are not UTF8 and log an error.

sdk/core/src/lib.rs
```diff
@@ -57,6 +57,13 @@ pub use response::*;
 pub use seekable_stream::*;
 pub use sleep::sleep;
 
+// re-export packages
+pub use http;
+pub use url;
```
rylev, line 62: Why? It would be really nice not to expose these entire crates. This means the core crate's semver story is directly tied to two other crates.

sdk/data_tables/src/responses/submit_transaction_response.rs
```diff
@@ -40,13 +39,18 @@ impl TryFrom<CollectedResponse> for SubmitTransactionResponse {
...
+                    let status_code = line
                         .split_whitespace()
                         .nth(1)
                         .ok_or_else(|| {
                             Error::message(ErrorKind::Other, "missing HTTP status code")
                         })?
-                        .parse()?;
+                        .parse::<u16>()
+                        .map_err(|_| {
+                            Error::message(ErrorKind::DataConversion, "invalid HTTP status code")
```
rylev, line 50: nit: this would completely swallow what the status code actually was.

rylev (dismissed):
Looks good, just a few nitpicks.

sdk/core/src/lib.rs
cataggar: I'm okay with that. I added this given previous feedback where @bmc-msft preferred whole packages over individual types. `form_urlencoded` is also needed. It is a [re-export from Url](https://docs.rs/url/latest/url/#reexports).

cataggar: I removed the package exports to get this PR merged.

sdk/core/src/response.rs
```diff
@@ -26,8 +26,8 @@ impl Response {
     }
 
     /// Get the status code from the response.
-    pub fn status(&self) -> StatusCode {
-        self.status
+    pub fn status(&self) -> &StatusCode {
```
cataggar, line 29: undo

sdk/core/src/mock/player_policy.rs
```diff
@@ -115,7 +115,7 @@ impl Policy for MockTransportPlayerPolicy {
         if expected_request.method() != request.method() {
             return Err(Error::with_message(ErrorKind::MockFramework, || {
                 format!(
-                    "mismatched HTTP request method. Actual: {0}, Expected: {1}",
+                    "mismatched HTTP request method. Actual: {0:?}, Expected: {1:?}",
```
cataggar, line 118: undo

sdk/core/src/mock/mock_response.rs
```diff
@@ -48,7 +46,7 @@ impl MockResponse {
         )?;
 
         let response = Response::new(
-            status_code,
+            status_code.clone(),
```
cataggar, line 49: undo

rylev (approved):

Issue #802, opened 2022-06-14
migrate SDKs to use HTTP pipeline


cataggar:
An HTTP pipeline general guidelines:
https://azure.github.io/azure-sdk/general_azurecore.html

- [x] sdk/data_cosmos
  - #290
  - #510 
  - #677
- [x] sdk/storage
  - #826
- [x] sdk/storage_datalake
  - #496 
- [x] sdk/storage_blobs
  - #843 @bmc-msft 
- [x] sdk/storage_queues
  -  #851 @bmc-msft 
- [x] sdk/iothub
  - #960 @rylev 
  - #972 @roeap 
- [ ] sdk/messaging_servicebus
- [x] sdk/data_tables
  - #547 @ShayMusachanov-dev
  - #874 @bmc-msft 
- [x] sdk/security_keyvault 
  - #959 @bmc-msft
  - #962 @bmc-msft 

A description copied from #290:

---
Previous to the pipeline architecture, each operation was independently controlled meaning there was no central way to control the way operations were performed. 

The pipeline architecture on the other hand allows users of the SDK to create pipelines through which outgoing requests and incoming responses are processed. The pipeline are composed of "policies" which can both change the outgoing request and the incoming response. Policies are usually either on a "per-request" basis (meaning they are called only once per operation) or "per-retry" (meaning they are applied every time an operation is attempted). 

Examples of policies are the `TelemetryPolicy` which adds various telemetry headers to the request and the retry policy which checks whether an incoming response was successful and if it detects a transient failure, retries the request. 

### Converting to the Architecture

For an example of converting one operation (Cosmos's `get_database` operation) over to the pipeline architecture, take a look at https://github.com/Azure/azure-sdk-for-rust/pull/286.

The basic steps are:
* Create a new submodule of the `operations` module where the operation will live. This submodule should be named after the operation (e.g., for `get_database` it's called `get_database`). 
* Move the "request builder" (e.g., for `get_database` the `GetDatabaseBuilder`) to new operation submodule and rename it from `Builder` to `Options` (e.g., `GetDatabaseBuilder => `GetDatabaseOptions`). This now represents the options that can be supplied when performing the operation. Note that many options that were part of the builder are no longer valid options because there will be individual policies that take their place. For instance, previously many operations took an option to set the user agent. This is now done by a policy. 
* Change the `execute` method to a `decorate_request` method that instead of performing the request, simply mutates a request parameter with any of the options supplied. 
[...]

PR #853, opened 2022-06-25
[datalake] add missing file and directory methods

roeap:
This PR adds the last remaining methods to the file and directory clients.

From what I can tell the last missing functionality is leases, but I left those for a follow-up PR, since they seem to be usually implemented using a dedicated client.

towards #802, #496

### DIFF
--- a/sdk/storage_datalake/src/properties.rs
+++ b/sdk/storage_datalake/src/properties.rs
@@ -14,6 +14,12 @@ impl Default for Properties {
     }
 }
 
+impl Into<Properties> for BTreeMap<Cow<'static, str>, Cow<'static, str>> {
+    fn into(self) -> Properties {
+        Properties(self)
+    }
+}
+
 impl Properties {
     pub fn new() -> Self {
         Self(BTreeMap::new())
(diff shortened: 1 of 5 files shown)


sdk/storage_datalake/src/properties.rs
```diff
@@ -14,6 +14,12 @@ impl Default for Properties {
     }
 }
 
+impl From<BTreeMap<Cow<'static, str>, Cow<'static, str>>> for Properties {
+    fn from(value: BTreeMap<Cow<'static, str>, Cow<'static, str>>) -> Self {
+        Self(value)
+    }
+}
```
roeap, line 21: Thought about providing a more flexible implementation, like

```rs
impl<T> From<T> for Properties
where
    T: Into<BTreeMap<Cow<'static, str>, Cow<'static, str>>>,
{
    fn from(value: T) -> Self {
        Self(value.into())
    }
}
```
however that conflicts with the `TryFrom<&Headers>` implementation, which needs to look into a specific header key. This seems to be related to some generic implementation in core - this was a bit surprising to me, since its different traits. but did not dig any deeper...

ctaggart (approved):
Thanks @roeap. I'm not an expert here. It looks fine to me.

PR #855, opened 2022-06-27
sketch out response type for headers

ctaggart:
Replaces #746, this is a design sketch for adjusting the generated responses to support headers. It also allows access to the inner `azure_core::Request`. It makes manual adjustments to Service_GetAccountInfo from azure-rest-api-specs/specification/storage/data-plane/Microsoft.BlobStorage/preview/2021-08-06/blob.json . The SDK implementation is azure_storage StorageAccountClient get_account_information.


yoshuawuyts (commented):
To Ryan's point from the meeting today: it'd be good if we could align the APIs from the hand-rolled crates and the generated crates.

I agree that ideally we should use the hand-rolled crates as a starting point, see how close we can get with the generated crates, and then adjust the designs from there. Making consistency between both types of crates a primary goal. What do you think?

rylev:
@ctaggart what's the state of this. Should this be closed or are you waiting on something specific before this can move forward?

PR #851, opened 2022-06-24
move storage queues to pipeline architecture

bmc-msft:

PR #856, opened 2022-06-27
use u8 slice for read_xml instead of Bytes

bmc-msft:

PR #858, opened 2022-06-27
Address feedback from Storage Queue pipeline pr

bmc-msft:
This addresses the comments in #851.

Note, this works around an issue flattening nested XML structures (https://github.com/RReverser/serde-xml-rs/issues/83) by adding `message_id` and `pop_receipt` to `Message`, then adding a `pop_receipt` method that generates a PopReceipt.


bmc-msft:
Note, this builds on #856

rylev (approved):

PR #860, opened 2022-06-27
switch `Method` and `StatusCode` to `http-types`

ctaggart:
`http-types` has better representations of both `Method` and `StatusCode` as I described in more [detail earlier](https://github.com/Azure/azure-sdk-for-rust/pull/849#issuecomment-1166333922).

I reverted the generated services code to make it easier to review.


sdk/core/src/error/http_error.rs
```diff
@@ -26,7 +26,7 @@ impl HttpError {
         let body = response.into_body().await;
         error_code = error_code.or_else(|| get_error_code_from_body(&body));
         HttpError {
-            status: status.as_u16(),
+            status: status as u16,
```
bmc-msft, line 29: Since we are re-exporting StatusCode now, can we make the error include the enum value instead of a u16?  This will make dealing with contextualized errors significantly more straightforward.

sdk/core/src/http_client/reqwest.rs
```diff
@@ -65,3 +69,16 @@ fn to_headers(map: &::reqwest::header::HeaderMap) -> crate::Result<crate::header
...
     Ok(crate::headers::Headers::from(map))
 }
+
+fn try_from_method(method: &crate::Method) -> crate::Result<::reqwest::Method> {
+    ::reqwest::Method::from_str(method.as_ref()).map_kind(ErrorKind::DataConversion)
+}
+
+fn try_from_status(status: ::reqwest::StatusCode) -> crate::Result<crate::StatusCode> {
+    let status = u16::from(status);
+    http_types::StatusCode::try_from(status).map_err(|_| {
```
bmc-msft, line 79:
```suggestion
    crate::StatusCode::try_from(status).map_err(|_| {
```

sdk/core/src/http_client/reqwest.rs
```diff
@@ -65,3 +69,16 @@ fn to_headers(map: &::reqwest::header::HeaderMap) -> crate::Result<crate::header
         .collect::<HashMap<_, _>>();
     Ok(crate::headers::Headers::from(map))
 }
+
+fn try_from_method(method: &crate::Method) -> crate::Result<::reqwest::Method> {
+    ::reqwest::Method::from_str(method.as_ref()).map_kind(ErrorKind::DataConversion)
```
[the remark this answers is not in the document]
cataggar: 7dfcf0125190ada447e2968a75282a5981f14f74 makes it more optimal.

sdk/core/src/error/http_error.rs
ctaggart: added #870

yoshuawuyts (approved):
LGTM

Issue #868, opened 2022-06-28
[Feature] Add support for setting blob properties


aaron-hardin:
Specifically would like to be able to change the md5 of a blob without editing the content. Use case: sometimes files uploaded to blob storage are missing the md5 and in cases where it is already known we can set it easily.
Set properties defined here: https://docs.microsoft.com/en-us/rest/api/storageservices/set-blob-properties

Will be doing a pull request for this shortly but most projects I've worked on like an issue to exist.

PR #869, opened 2022-06-28
[Feature] Add support for setting blob properties

aaron-hardin:
Adds an operation for set_properties
Set properties defined here: https://docs.microsoft.com/en-us/rest/api/storageservices/set-blob-properties

Related to #868


sdk/storage_blobs/src/blob/operations/set_properties.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    setters! {
+        lease_id: LeaseId => Some(lease_id),
+        timeout: Timeout => Some(timeout),
+        cache_control: BlobCacheControl => Some(cache_control),
+        content_type: BlobContentType => Some(content_type),
+        content_encoding: BlobContentEncoding => Some(content_encoding),
+        content_language: BlobContentLanguage => Some(content_language),
+        content_disposition: BlobContentDisposition => Some(content_disposition),
+        content_md5: BlobContentMD5 => Some(content_md5),
```
rylev, line 74: You should also add

```rust
context: Context => context,
```

sdk/storage_blobs/src/blob_cache_control.rs
```diff
@@ -0,0 +1,42 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct BlobCacheControl(std::borrow::Cow<'static, str>);
+
+impl BlobCacheControl {
+    pub fn as_str(&self) -> &str {
+        self.0.as_ref()
+    }
+
+    pub const fn from_static(s: &'static str) -> Self {
```
rylev, line 11: nit: put constructors first in the `impl`

sdk/storage_blobs/src/clients/blob_client.rs
```diff
@@ -93,6 +93,13 @@ impl BlobClient {
         GetPropertiesBuilder::new(self.clone())
     }
 
+    /// Creates a builder for setting blob properties
+    /// NOTE: several properties are cleared from the blob if not passed
+    /// Consider calling set_from_blob_properties with existing blob properties
```
rylev, line 98:
```suggestion
    /// Creates a builder for setting blob properties.
    ///
    /// Several properties are cleared from the blob if not passed.
    /// Consider calling `set_from_blob_properties` with existing blob properties.
```

rylev (approved):
👍Thanks!

bmc-msft:
Thanks for the contribution!  It looks like you're hitting conflict from a change that landed just a few minutes ago.  #860.

Can you update this PR use the new HTTP types?

sdk/storage_blobs/src/blob/operations/set_properties.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    pub fn into_future(mut self) -> Response {
+        Box::pin(async move {
+            let mut url = self.blob_client.url_with_segments(None)?;
+
+            url.query_pairs_mut().append_pair("comp", "properties");
+            self.timeout.append_to_url_query(&mut url);
+
+            let mut request = self.blob_client.prepare_request(url, Method::PUT, None)?;
```
bmc-msft, line 85:
```suggestion
            let mut request = self.blob_client.prepare_request(url, Method::Put, None)?;
```

aaron-hardin:
> Needs updated types from #860, which just landed.

👍 Thanks for the update

bmc-msft:
After changing the HTTP Verb to match the updated types, LGTM.

Again, thanks for the contributions!

Issue #871, opened 2022-06-28
Azure Data Tables SharedKey does not work


bmc-msft:
The Access Key authentication for Azure Data Tables does not work as it generates an invalid signed request.

Consider the `create_table` request.  This request uses `prepare_request`, then adds headers as appropriate:

https://github.com/Azure/azure-sdk-for-rust/blob/42f89966d1d76bf4916979b091fb6af4ebeaefe9/sdk/data_tables/src/requests/create_table_builder.rs#L37-L44

The `prepare_request` does the signing, which goes through a call stack that gets to `azure_storage::core::clients::StorageAccountClient.prepare_request` which does the signing.

For account & key based authentication, it does calls `generate_authorization` with the headers in the request thus far:

https://github.com/Azure/azure-sdk-for-rust/blob/42f89966d1d76bf4916979b091fb6af4ebeaefe9/sdk/storage/src/core/clients/storage_account_client.rs#L444-L447

Later, the content-type is used to generate the signature:

https://github.com/Azure/azure-sdk-for-rust/blob/42f89966d1d76bf4916979b091fb6af4ebeaefe9/sdk/storage/src/core/clients/storage_account_client.rs#L549-L557

Because `content-type` doesn't exist as a header yet, it's not available at the time of `prepare_request`, which results in an invalid signature.

I validated this by temporarily hard-coding the content-type within `prepare_request` to include `application/json`, which generated a signed request that worked to create a table.

PR #872, opened 2022-06-28
pass Headers as part of prepare_request

bmc-msft:
`prepare_request` adds the authentication related headers.  When using Access Key methods, both Blob Storage and Data Tables uses data from the headers as part of canonicalized data that gets signed, which is checked at the server side.

As such, the headers must be part of what is sent to `prepare_request`.

See #871 for a walk through for Data Tables.

### DIFF
--- a/sdk/core/src/headers/mod.rs
+++ b/sdk/core/src/headers/mod.rs
@@ -181,6 +181,18 @@ impl HeaderName {
         Self(std::borrow::Cow::Borrowed(s))
     }
 
+    fn from_cow<C>(c: C) -> Self
+    where
+        C: Into<std::borrow::Cow<'static, str>>,
+    {
+        let c = c.into();
+        assert!(
+            c.chars().all(|c| c.is_lowercase() || !c.is_alphabetic()),
+            "header names must be lowercase: {c}"
+        );
+        Self(c)
+    }
+
     pub fn as_str(&self) -> &str {
         self.0.as_ref()
     }
@@ -188,14 +200,13 @@ impl HeaderName {
 
 impl From<&'static str> for HeaderName {
     fn from(s: &'static str) -> Self {
-        assert_eq!(s.to_lowercase(), s, "header names must be lower-case");
-        Self::from_static(s)
+        Self::from_cow(s)
     }
 }
 
 impl From<String> for HeaderName {
     fn from(s: String) -> Self {
-        Self(std::borrow::Cow::Owned(s.to_lowercase()))
+        Self::from_cow(s.to_lowercase())
     }
 }
 
@@ -208,6 +219,13 @@ impl HeaderValue {
         Self(std::borrow::Cow::Borrowed(s))
     }
 
+    fn from_cow<C>(c: C) -> Self
+    where
+        C: Into<std::borrow::Cow<'static, str>>,
+    {
+        Self(c.into())
+    }
+
     pub fn as_str(&self) -> &str {
         self.0.as_ref()
     }
@@ -215,13 +233,13 @@ impl HeaderValue {
 
 impl From<&'static str> for HeaderValue {
     fn from(s: &'static str) -> Self {
-        Self::from_static(s)
+        Self::from_cow(s)
     }
 }
 
 impl From<String> for HeaderValue {
     fn from(s: String) -> Self {
-        Self(std::borrow::Cow::Owned(s))
+        Self::from_cow(s)
     }
 }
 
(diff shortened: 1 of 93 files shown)


bmc-msft:
Note: With this PR, the table_00 example from azure_data_tables now works.

sdk/core/src/headers/mod.rs
```diff
@@ -178,13 +193,14 @@ impl HeaderName {
 
 impl From<&'static str> for HeaderName {
     fn from(s: &'static str) -> Self {
+        assert_eq!(s.to_lowercase(), s, "header names must be lower-case");
```
bmc-msft, line 196: This forces a canonicalization of header names to prevent issues such as what I found in data tables, where even after adding headers to the `prepare_request`, I found that data_tables was using `Content-Type` rather than `content-type`, which did show up in `Headers.get("content-type")`.  

By forcing a canonicalized form here, we prevent footguns later.

sdk/core/src/headers/mod.rs
```diff
@@ -140,12 +140,27 @@ impl Headers {
...
     pub fn iter(&self) -> impl Iterator<Item = (&HeaderName, &HeaderValue)> {
         self.0.iter()
     }
 }
 
+impl Default for Headers {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
rylev, line 162: It doesn't seem like we really need this. Can we remove it?

sdk/core/src/headers/mod.rs
```diff
@@ -140,12 +140,27 @@ impl Headers {
         self.0.insert(key.into(), value.into());
     }
 
+    pub fn add<H>(&mut self, header: H)
```
rylev, line 143: Can you add a doc?

sdk/core/src/headers/mod.rs
rylev, line 196: This probably should be inside `HeaderName::from_static` or even better in a private constructor that takes a `Cow<'static, str>` and constructs the `HeaderName` which everything uses instead of constructing the `HeaderName` directly.

sdk/data_tables/src/clients/entity_client.rs
```diff
@@ -102,10 +102,11 @@ impl EntityClient {
         &self,
         url: Url,
         method: Method,
+        headers: Headers,
         request_body: Option<Bytes>,
     ) -> azure_core::Result<Request> {
         self.partition_key_client
-            .prepare_request(url, method, request_body)
```
rylev, line 108: Since all of these `prepare_request` methods need to be called with the final set of headers, maybe renaming them to `finalize_request` would be better. That way it's more clear that this should be called *last*.

sdk/storage_blobs/src/blob/operations/acquire_lease.rs
```diff
@@ -44,13 +37,15 @@ impl AcquireLeaseBuilder {
             url.query_pairs_mut().append_pair("comp", "lease");
             self.timeout.append_to_url_query(&mut url);
 
+            let mut headers = Headers::new();
+            headers.insert(LEASE_ACTION, "acquire");
```
rylev, line 41: It would be nice if we could get `headers.add(azure_core::LeaseAction::Acquire)` to work, but we can leave that for a future PR.

sdk/storage_blobs/src/blob/operations/clear_page.rs
```diff
@@ -47,17 +47,18 @@ impl ClearPageBuilder {
             self.timeout.append_to_url_query(&mut url);
             url.query_pairs_mut().append_pair("comp", "page");
 
+            let mut headers = Headers::new();
+            headers.insert(PAGE_WRITE, "clear");
+            headers.insert(BLOB_TYPE, "PageBlob");
```
rylev, line 52: Same here: it would be nice if `headers.add(BlobType::PageBlob)` worked.

rylev (changes requested):
Looks good, but I'd like to address a few small things first.

sdk/core/src/headers/mod.rs
bmc-msft: While I'd prefer this to live in from_static, I did not want to break the `const` attribute, which @cataggar was invested in adding, none of the methods I can come up with are const.  (for loops, chars iterator, and to_lowercase)

I added a private from_cow method which is used by the From impls that does this.

rylev (approved):

PR #873, opened 2022-06-29
Make `Continuation` an associated type

rylev:
This adds an associated type to `Continuable` called `Continuation`. This means that implementors of `Continuable` are in charge of specifying which type is used to indicate how they continue. This removes a bunch of run time checks that the right type of `Continuation` was used.


bmc-msft (approved):
This is much more elegant.

PR #547, opened 2021-12-25
[Data Tables] Migrate table service to pipeline architecture

ShayMusachanov-dev:
Before I continue implementing the missing features, could someone please verify that I am up to date with the latest changes in the architecture?

PR #874, opened 2022-06-29
Move data tables to pipeline

bmc-msft:
This is heavily inspired by the work done by @ShayMusachanov-dev in #547, but uses the `azure_storage` pipeline APIs added in #811.

Once #873 lands, this should be updated to clean up the Continuation usage.

This changes the transaction model way from consuming existing operation builders and instead adds methods to the transaction builder that inserts transaction operations.  This provides a significantly simpler method to collect operations into a single builder.

As an example, this snippit is from the example code:
```rust
    let response = partition_key_client
        .transaction()
        .delete(&entity1.surname)?
        .insert(&entity2)?
        .update(&entity3.surname, &entity3)?
        .merge(&entity4.surname, &entity4)?
        .insert_or_replace(&entity5.surname, &entity5, IfMatchCondition::Any)?
        .insert_or_merge(&entity6.surname, &entity6, IfMatchCondition::Any)?
        .into_future()
        .await?;
```


sdk/data_tables/src/operations/get_entity.rs
```diff
@@ -0,0 +1,105 @@
...
+        })
+    }
+}
+
+pub type FutureResponse<E> =
+    futures::future::BoxFuture<'static, azure_core::Result<GetEntityResponse<E>>>;
+
+#[cfg(feature = "into_future")]
+impl std::future::IntoFuture for GetEntityBuilder {
+    type IntoFuture = FutureResponse<E>;
```
rylev, line 61: This doesn't compile. The Builder should be made generic. See `CreateDocument` in cosmos db.

rylev (changes requested):
Looking good. There is one issue I've found though

rylev (approved):

PR #875, opened 2022-06-29
Only add params if they don't already exist

### LINKED ISSUE #867: SASToken auth broken
A recent commit (dd6b8523033d6443ef6a6e982d9e579fc8cf7caa) added the Authorization policy which adds the query string for SASToken however, StorageAccountClient::prepare_request also adds it, the result is that trying to use SASToken for blob storage requests (at least) fails with AuthenticationFailed (403) response.
URL looks like: https://[account].blob.core.windows.net/[container]/[blob]?[sasToken][sasToken]

rylev:
Fixes #867 

Once we move all of the SDKs that use the `StorageAccountClient` to the pipeline architecture, this workaround can be removed.


bmc-msft (approved):
This whole `if` block could probably move to `match &self.storage_credentials` which happens a few lines later.

rylev:
@bmc-msft that match expression will be deleted when we move all the Storage SDKs to the pipeline architecture since this is all handled by the authentication policy.

PR #878, opened 2022-06-30
Try out an operation macro

rylev:
This introduces an `operation!` macro which reduces the amount of boiler plate around declaring operations considerably.

This is how it looks when called:

```rust
operation! {
    CreateCollection,
    client: DatabaseClient,
    collection_name: String,
    partition_key: PartitionKey,
    ?consistency_level: ConsistencyLevel,
    ?indexing_policy: IndexingPolicy,
    ?offer: Offer
}
```

And it expands to:

```rust
#[derive(Debug, Clone)]
pub struct CreateCollectionBuilder {
    client: DatabaseClient,
    collection_name: String,
    partition_key: PartitionKey,
    consistency_level: Option<ConsistencyLevel>,
    indexing_policy: Option<IndexingPolicy>,
    offer: Option<Offer>,
    context: Context,
}

impl CreateCollectionBuilder {
    pub(crate) fn new(
        client: DatabaseClient,
        collection_name: String,
        partition_key: PartitionKey,
    ) -> Self {
        Self {
            client,
            collection_name,
            partition_key,
            consistency_level: None,
            indexing_policy: None,
            offer: None,
            context: Context::new(),
        }
    }

    setters! {
        consistency_level: ConsistencyLevel => Some(consistency_level),
        indexing_policy: IndexingPolicy => Some(indexing_policy),
        offer: Offer => Some(offer),
        context: Context => context,
    }
}

#[cfg(feature = "into_future")]
impl std::future::IntoFuture for CreateCollectionBuilder {
    type IntoFuture = CreateCollection;
    type Output = <CreateCollection as std::future::Future>::Output;
    fn into_future(self) -> Self::IntoFuture {
        Self::into_future(self)
    }
}

/// The future returned by calling `into_future` on the builder.
pub type CreateCollection =
    futures::future::BoxFuture<'static, azure_core::Result<CreateCollectionResponse>>;
```

This would ensure all of our operation builders follow the same pattern.

Some current limitations:
* Does not work for pageable operations
* Does not work for operations that require a generic param

I believe both of these limitations are pretty easy to overcome, but I wanted to see what folks thought.

Thoughts?

### DIFF
--- a/sdk/data_cosmos/src/resources/document/indexing_directive.rs
+++ b/sdk/data_cosmos/src/resources/document/indexing_directive.rs
@@ -13,6 +13,12 @@ pub enum IndexingDirective {
     Exclude,
 }
 
+impl Default for IndexingDirective {
+    fn default() -> Self {
+        Self::Default
+    }
+}
+
 impl<'a> From<&'a IndexingDirective> for &'a str {
     fn from(s: &'a IndexingDirective) -> &'a str {
         match s {
--- a/sdk/core/Cargo.toml
+++ b/sdk/core/Cargo.toml
@@ -30,6 +30,7 @@ serde_json = "1.0"
 url = "2.2"
 uuid = { version = "1.0" }
 pin-project = "1.0.10"
+paste = "1.0"
 
 # Add dependency to getrandom to enable WASM support
 [target.'cfg(target_arch = "wasm32")'.dependencies]
--- a/sdk/core/src/lib.rs
+++ b/sdk/core/src/lib.rs
@@ -89,3 +89,9 @@ where
         }
     }
 }
+
+#[doc(hidden)]
+/// Used by macros as an implementation detail
+pub mod __private {
+    pub use paste::paste;
+}
--- a/sdk/core/src/macros.rs
+++ b/sdk/core/src/macros.rs
@@ -121,8 +121,8 @@ macro_rules! setters {
 ///     futures::future::BoxFuture<'static, azure_core::Result<CreateCollectionResponse>>;
 /// ```
 ///
-/// Additionally, `@list` can be used before the operation name to generate code appropriate for list operations
-/// and `??` can be used at the end of the list of options for options where we should not generate a setter.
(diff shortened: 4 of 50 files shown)


bmc-msft:
> @bmc-msft any thoughts on what we should use for option instead of the `?`?

Not at this time.  It's livable and given it's only for internal use, we can move to it and adjust if we need to.

It would be good to document the macro with the example you provided.

rylev:
Ok I'm happy where this ended up. There are about 1000 fewer lines of operations code in Cosmos, and this will only grow as we add more operations from other SDKs. 

The following is the syntax we currently have. In the interest of keeping the macro definition simple I'd like to avoid getting too fancy, but we could certainly improve some things. I'll make note of some ideas. Also, please note that the usage below is the most gnarly it can possibly be (i.e., it uses all the features of the macro for instruction purposes). No actual usage is this complex. 

```rust
operation! {
    #[stream] // This instructs the macro to avoid generating the `IntoFuture` definition and instead assume it's a stream
    ListAttachments<T: Serialize + Send>,
    client: DocumentClient, // Which client is used always comes next
    document: T,  // Then a list of all the required options. We could require the user to annotate these with `#[required]`
    ?consistency_level: ConsistencyLevel // Then all the optional options marked by `?`. If we require the use of `#[required]` then we could drop the `?` here.
    #[skip] // This instructs the follow field to not generate a setter automatically. This is useful when you want a more complex setter 
    partition_key_serialized: String
}
```

Thoughts?

sdk/data_cosmos/src/resources/document/indexing_directive.rs
```diff
@@ -13,6 +13,12 @@ pub enum IndexingDirective {
     Exclude,
 }
 
+impl Default for IndexingDirective {
```
bmc-msft, line 16: Consider using `derive_default_enum` with a bumped MSRV?  Given the soon-to-be-stabilized IntoFuture, we're likely to use this in the future.

rylev: We can switch to that when we bump the MSRV for `IntoFuture`.

PR #882, opened 2022-07-01
use StatusCode rather than u16 in HttpError

### LINKED ISSUE #870: consider using `StatusCode` instead of `u16` in `HttpError`
From https://github.com/Azure/azure-sdk-for-rust/pull/860#pullrequestreview-1021792214

bmc-msft:
As we're exposing the StatusCode enum, this makes the consumption of the
errors significantly easier to deal with.

Resolves #870


cataggar (approved):

rylev (approved):
I'm surprised to see we don't have any examples that are impacted by this change. Would you mind adding an issue for creating such an example or test?

PR #884, opened 2022-07-01
use `clap` parser in cosmos examples

yoshuawuyts:
This is a relatively small change, but it moves the cosmos examples from our custom parsing code to the commonly-used [`clap`](https://docs.rs/clap) command line parser crate. The benefit of doing this is that it removes bespoke parsing code from inside `fn main` to a declarative struct at the top of the file. The focus of the examples should be on how to use the Azure APIs, and not having several lines of command line parsing code at the start of each `fn main` should make it a lot easier to skip to the core of the examples.

I was reading over some examples, and found myself distracted by the initial parsing code in each file. Taken as a whole I think this should make the examples a lot easier to follow. Thanks!

### DIFF
--- a/sdk/data_cosmos/examples/attachments_00.rs
+++ b/sdk/data_cosmos/examples/attachments_00.rs
@@ -1,4 +1,5 @@
 use azure_data_cosmos::prelude::*;
+use clap::Parser;
 use futures::StreamExt;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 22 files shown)


sdk/data_cosmos/examples/attachments_00.rs
```diff
@@ -19,32 +20,33 @@ impl azure_data_cosmos::CosmosEntity for MySampleStruct {
...
+    let authorization_token = AuthorizationToken::primary_from_base64(&args.primary_key)?;
 
-    let id = format!("unique_id{}", 100);
+    let client = CosmosClient::new(args.account, authorization_token, CosmosOptions::default())
+        .database_client(args.database_name)
+        .collection_client(args.collection_name);
 
     let doc = MySampleStruct {
-        id,
+        id: format!("unique_id{}", 100),
```
rylev, line 49: I know it was this way before, but `"unique_id100".to_owned()` might be better.

rylev (approved):
Looks good! 

Two thoughts:
- I'm slightly worried that clap will make compile times for examples worse.
- Maybe we could share some of the auth related functionality across the different examples.

PR #889, opened 2022-07-04
Use `Headers::get_as` where appropriate

rylev:
This just makes getting headers more regular.


cataggar (approved):
Appreciate these cleanups.

bmc-msft (approved):

PR #890, opened 2022-07-04
Use `Response::is_success` for retry policy

### LINKED ISSUE #836: check status code range for `execute_request_check_status`
> I have concerns regarding the implementation here.

> While this comment says 2xx range, the original implementation checks 200..400 (which this PR matches) does not align with this comment.  More importantly, accepting a response within the 3xx range is probably not what the consumers of this API are likely to expect.

> IMO, this is easily handled from the consumer side as:

> `Response.status().is_success()`, or `Response.status() == HttpStatus::CREATED`

https://github.com/Azure/azure-sdk-for-rust/pull/833#pullrequestreview-1012667358

rylev:
Fixes #836

### DIFF
--- a/sdk/core/src/policies/retry_policies/retry_policy.rs
+++ b/sdk/core/src/policies/retry_policies/retry_policy.rs
@@ -48,13 +48,12 @@ where
 
         loop {
             let error = match next[0].send(ctx, request, &next[1..]).await {
-                Ok(response) if (200..400).contains(&u16::from(response.status())) => {
+                Ok(response) if response.status().is_success() => {
                     log::trace!(
                         "Succesful response. Request={:?} response={:?}",
                         request,
                         response
                     );
-                    // Successful status code
                     return Ok(response);
                 }
                 Ok(response) => {


sdk/core/src/policies/retry_policies/retry_policy.rs
```diff
@@ -48,13 +48,12 @@ where
 
         loop {
             let error = match next[0].send(ctx, request, &next[1..]).await {
-                Ok(response) if (200..400).contains(&u16::from(response.status())) => {
+                Ok(response) if response.status().is_success() => {
```
cataggar, line 51: A concern is that it is a change in behavior. What if it [.is_redirection](https://docs.rs/http-types/latest/src/http_types/status_code.rs.html#434-450) in the 300..400 range?

bmc-msft: We should have a policy with regards to redirection and it should occur before retry, but it should not be intermingled with retry.

bmc-msft (approved):

PR #891, opened 2022-07-04
House cleaning for blob clients

rylev:
A bit of house cleaning:
* Add docs to `BlobClient` operation methods
* Rearrange the methods so that they are in some order (generally: first general get, crate, update, and delete operations and then operations specific to certain types of blobs)
* Move methods in other clients around so public ones are on top
* Rename `update_page` to `put_page` to mirror other "put" operations and the docs.


sdk/storage_blobs/src/clients/blob_client.rs
```diff
@@ -234,18 +272,6 @@ impl BlobClient {
     ) -> azure_core::Result<Response> {
         self.container_client.send(context, request).await
     }
-
-    pub async fn exists(&self) -> azure_core::Result<bool> {
```
bmc-msft, line 238: This helper similarly exists in the Python SDK.

https://github.com/Azure/azure-sdk-for-python/blob/7eeaca97c41bc18ec66f325eef5e6910efde6e12/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py#L1154-L1180

bmc-msft (approved):

PR #896, opened 2022-07-05
add azure-autorust bin

ctaggart:
This consolidates the scripts used to do the code generation into an executable. It is a direct port.  This will open up the door for adding additional args.


services/autorust/azure-autorust/src/main.rs
```diff
@@ -0,0 +1,112 @@
...
+        "azure_security_keyvault",
+        "azure_storage",
+        "azure_storage_blobs",
+        "azure_storage_datalake",
+        "azure_storage_queues",
+    ];
+    let yml = PublishSdksYml { packages };
+    yml.create("../../.github/workflows/publish-sdks.yml")?;
+    Ok(())
+}
```
bmc-msft, line 104: It feels weird that configs for the SDK crates are written by azure-autorust.

bmc-msft (approved):

services/autorust/azure-autorust/src/main.rs
cataggar: I agree. It is behind a `--publish` arg and I'm the only one who runs it. It would be nice to keep the creation of the workflows together. It just saves me time. I can move this back into an example if you would prefer that.

PR #898, opened 2022-07-05
replace block_on witch await for token_credential.get_token in storage

### LINKED ISSUE #808: Storage account client with TokenCredential hangs
My changes https://github.com/Azure/azure-sdk-for-rust/pull/757 seem to cause a hang in [Vector](https://github.com/vectordotdev/vector) when managed identity is used. It hangs when requesting the token from the IMDS: https://github.com/Azure/azure-sdk-for-rust/blob/490d34b7a51af853ec2c39eb5820fae0a44dcc4d/sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs#L125

The log "starting new connection:" is the last thing I see before it hangs: https://github.com/seanmonstar/reqwest/blob/8b37ae4b15f668e3e3f9b8cf43b547fc4f95e444/src/connect.rs#L560

I think this has something to do with that Vector calls it from a async method, which calls a sync method (prepare_request of sdk), which call a async function again (get_token of sdk).
Example call in Vector: https://users.rust-lang.org/t/simplest-possible-block-on/48364/29
This seems to be problematic according to this post: https://users.rust-lang.org/t/simplest-possible-block-on/48364/26
But I couldn't verify it's actually the case.

To verify it I tried to remove `block_on` by making prepare_request async but I'm running into compile errors here: https://github.com/yvespp/azure-sdk-for-rust/blob/86ebddbbb80f4527cd6d99c40b714957ff2d6a68/sdk/storage_blobs/tests/blob.rs#L380
```
generator cannot be sent between threads safely
the trait `Sync` is not implemented for `dyn Fn(http::request::Builder) -> http::request::Builder`rustc
blob.rs(330, 42): required by a bound in `requires_send_future`
```
And also while compiling Vector:
```
make build
cargo build --release --no-default-features --features default
   Compiling vector v0.23.0 (/home/azureuser/vector)
error: generator cannot be sent between threads safely
  --> src/sinks/azure_common/service.rs:43:9
   |
43 |         Box::pin(async move {
   |         ^^^^^^^^^^^^^^^^^^^ future created by async block is not `Send`
   |
   = help: the trait `std::marker::Sync` is not implemented for `dyn std::ops::Fn(http::request::Builder) -> http::request::Builder`
   = note: required for the cast to the object type `dyn futures_util::Future<Output = std::result::Result<AzureBlobResponse, Box<(dyn snafu::Error + std::marker::Send + std::marker::Sync + 'static)>>> + std::marker::Send`

error: could not compile `vector` due to previous error
make: *** [Makefile:172: build] Error 101
```
The branch with my changes is here: https://github.com/Azure/azure-sdk-for-rust/compare/main...yvespp:storage_account_TokenCredential_async_fix2

It would be nice if you coud give me some pointers on how to go forward with this?

I also thought about using a blocking method from tokio instead of `futures::executor::block_on` but that would bind the sdk to tokio which is probably not wanted.

yvespp:
As get_token is now inside an async function it doesn't make sens to use block_one.
Should fix #808


bmc-msft (approved):
I agree this should just be an .await.

bmc-msft (approved):

PR #899, opened 2022-07-05
add missing restype=container query param for container get_props

yvespp:
Without it get_properties returns a 404.

I also noticed that list_containers should be a storage account operation and not a container operation: https://docs.microsoft.com/en-us/rest/api/storageservices/list-containers2
Should it be moved?


bmc-msft:
Thanks for the contribution!

PR #909, opened 2022-07-06
lowercase x-identity-header used in IMDS

bmc-msft:


heaths (approved):
Seems fine, but I'm concerned why this is necessary. HTTP headers should be read case-insensitively. Is the service not doing so?

cataggar:
This is inserting headers. The reason to standardize on lowercase is in https://github.com/Azure/azure-sdk-for-rust/pull/872#discussion_r909066411.

PR #922, opened 2022-07-12
expose Storage Options when creating storage clients

bmc-msft:
This enables setting pipeline options at client configuration time,
which is required to set policies, such as request retry configuration.

### DIFF
--- a/sdk/storage/tests/account.rs
+++ b/sdk/storage/tests/account.rs
@@ -8,7 +8,8 @@ async fn get_account_information() {
     let access_key =
         std::env::var("STORAGE_ACCESS_KEY").expect("Set env variable STORAGE_ACCESS_KEY first!");
 
-    let storage_client = StorageClient::new_access_key(&account, &access_key);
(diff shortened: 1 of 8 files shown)


bmc-msft:
Note, I recognize this goes in the opposite direction of #886 , but think this is a path to enable setting policies until a design for the builder pattern used for clients has been defined.

rylev (commented):
I'll let @yoshuawuyts comment on what the plans are for the client builder pattern, but I'm unsure about merging this if it's just going to change in a few days.

yoshuawuyts:
Thanks for tagging me @rylev! - Yeah, this pulls us in the exact opposite direction of https://github.com/Azure/azure-sdk-for-rust/pull/886, and it would probably make more sense if we figure the API out in the Cosmos SDK first, and then migrate this API over after to match that.

bmc-msft:
Agreed.  This was discussed at the SDK meeting.  I will close for now.

PR #927, opened 2022-07-13
add tags support to storage blobs

### LINKED ISSUE #593: [storage] add x-ms-tags header to PutBlockBlobBuilder
This PR stalled, so this is just moving it to an issue.
- #429

bmc-msft:
This adds support for setting blobs during operations that support it (via headers) as well as getting & setting tags via operations.

The result of the example looks something like this:

```
get tags result: GetTagsResponse { request_id: 644618ba-701e-000d-2ca6-97d662000000, date: 2022-07-14T17:23:31Z, tags: Tags { tag_set: TagSet { tags: [Tag { key: "tag1", value: "value1" }] } } }
set tags result: SetTagsResponse { request_id: 644618c6-701e-000d-37a6-97d662000000, date: 2022-07-14T17:23:31Z }
get tags result: GetTagsResponse { request_id: 644618cc-701e-000d-3da6-97d662000000, date: 2022-07-14T17:23:31Z, tags: Tags { tag_set: TagSet { tags: [Tag { key: "tag2", value: "value2" }, Tag { key: "tag3", value: "value3" }] } } }
key:tag2 value:value2
key:tag3 value:value3
get tags without tags result: GetTagsResponse { request_id: 644618dc-701e-000d-4ba6-97d662000000, date: 2022-07-14T17:23:31Z, tags: Tags { tag_set: TagSet { tags: [] } } }```

closes #593 

Inspired by the work done by @JensWalter in #429

### CODE REVIEW

File: sdk/storage_blobs/src/options/tags.rs [hunk 1]
```diff
@@ -0,0 +1,125 @@
+use azure_core::{
+    error::Error,
+    headers::{Header, HeaderName, HeaderValue, TAGS},
+};
+use std::{collections::HashMap, str::FromStr};
+use url::form_urlencoded;
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Default)]
+#[serde(rename_all = "PascalCase")]
+pub struct Tags {
```
rylev, line 10: Can we add comments to more of this?

sdk/storage_blobs/src/options/tags.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub tag_set: Option<TagSet>,
```
rylev, line 11: Can this be `#[serde(default)]` so it doesn't have to be an option?

sdk/storage_blobs/src/options/tags.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+impl Tags {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn insert<S>(&mut self, key: S, value: S)
+    where
+        S: Into<String>,
```
rylev, line 34: This makes the following not possible:

```rust
tags.insert("Hello", String::from("world"));
```

sdk/storage_blobs/src/options/tags.rs
```diff
@@ -0,0 +1,125 @@
...
+        let tag_set = self.tag_set.as_mut().unwrap();
+        tag_set.tag.push(Tag {
+            key: key.into(),
+            value: value.into(),
+        });
+    }
+
+    pub fn to_xml(&self) -> String {
+        let mut s = String::new();
+        s.push_str("<?xml version=\"1.0\" encoding=\"utf-8\"?>\n<Tags>\n<TagSet>\n");
```
rylev, line 49:
```suggestion
        let mut s = String::from("<?xml version=\"1.0\" encoding=\"utf-8\"?>\n<Tags>\n<TagSet>\n");
```

sdk/storage_blobs/src/options/tags.rs
```diff
@@ -0,0 +1,125 @@
...
+        if let Some(tag_set) = &self.tag_set {
+            for tag in &tag_set.tag {
+                let node = format!(
+                    "\t<Tag><Key>{}</Key><Value>{}</Value></Tag>\n",
+                    tag.key, tag.value
+                );
+                s.push_str(&node);
```
rylev, line 56:
```suggestion
               write!(
                    &mut s, 
                    "\t<Tag><Key>{}</Key><Value>{}</Value></Tag>\n",
                    tag.key, tag.value
                ).unwrap();
```

sdk/storage_blobs/src/options/tags.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+impl From<Tags> for HashMap<String, String> {
+    fn from(tags: Tags) -> Self {
+        let mut map = Self::new();
+        if let Some(tag_set) = &tags.tag_set {
+            for tag in &tag_set.tag {
+                map.insert(tag.key.clone(), tag.value.clone());
+            }
+        }
```
rylev, line 92:
```suggestion
        if let Some(tag_set) = tags.tag_set {
           map.extend(tag_set.tag);
        }
```

sdk/storage_blobs/src/options/tags.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+impl FromStr for Tags {
+    type Err = Error;
+    fn from_str(value: &str) -> azure_core::Result<Tags> {
+        let mut tags = Self::new();
+        let pairs = form_urlencoded::parse(value.as_bytes());
+        for (key, value) in pairs {
+            tags.insert(key.to_owned(), value.to_owned());
+        }
```
rylev, line 104: Might be nice to have an [`extend` method](https://doc.rust-lang.org/stable/std/collections/hash_map/struct.HashMap.html#method.extend) on `Tags` so that we don't always have to manually loop.

rylev (changes requested):

rylev (approved):

sdk/storage_blobs/src/options/tags.rs
```diff
@@ -28,67 +45,68 @@ impl Tags {
...
-            }
+    pub fn to_xml(&self) -> azure_core::Result<String> {
+        let mut s = String::from("<?xml version=\"1.0\" encoding=\"utf-8\"?><Tags><TagSet>");
+        for tag in &self.tag_set.tag {
+            write!(
+                &mut s,
+                "<Tag><Key>{}</Key><Value>{}</Value></Tag>",
+                tag.key, tag.value
+            )
+            .context(ErrorKind::DataConversion, "failed to write Tags xml")?;
```
rylev, line 72: Just as an FYI: the call can't fail here (writing to a `String` always works), so `unwrap` would be arguably more appropriate here.

Issue #930, opened 2022-07-14
Download of single byte range from large file fails


evanxg852000:
When downloading a small range from a large blob, the byte range gets downloaded successfully but a subsequent request is sent that yield the following error.
```bash
Error: Error { context: Full(Custom { kind: HttpResponse { status: BadRequest, error_code: Some("InvalidHeaderValue") }, error: HttpError { status: BadRequest, error_code: Some("InvalidHeaderValue"), headers: {"date": "Thu, 14 Jul 2022 08:15:00 GMT", "x-ms-request-id": "e4deb301-701e-0057-3359-977c9c000000", "server": "Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0", "x-ms-version": "2019-12-12", "content-type": "application/xml", "x-ms-error-code": "InvalidHeaderValue", "content-length": "338"}, body: b"\xef\xbb\xbf<?xml version=\"1.0\" encoding=\"utf-8\"?><Error><Code>InvalidHeaderValue</Code><Message>The value for one of the HTTP headers is not in the correct format.\nRequestId:e4deb301-701e-0057-3359-977c9c000000\nTime:2022-07-14T08:15:01.5908400Z</Message><HeaderName>x-ms-range-get-content-crc64</HeaderName><HeaderValue>true</HeaderValue></Error>" } }, "server returned error status which will not be retried") }
```

PR #931, opened 2022-07-14
Fix downloading single range from large blob 

evanxg852000:
Avoid sending subsequent request when downloading single range from a large blob.
Closes [this issue](https://github.com/Azure/azure-sdk-for-rust/issues/930)


evanxg852000:
@bmc-msft  You are welcome, Thanks for the work you have put into this. It's really useful.

PR #933, opened 2022-07-14
Some cleanup of cosmos examples

rylev:
This does some small cleanup to the Cosmos examples:
* Delete a few examples that simply did the exact same thing as another example
* Renamed all examples from `$NAME_01` to just `$NAME`
* Cleaned up some out of date or incorrect comments
* Cleaned up some of the output to stdout


yoshuawuyts (approved):
Looks great!

PR #934, opened 2022-07-14
Prefer usage of ContentType type over inserting header directly

rylev:
I'd like to see us move more in this direction where we're not manually inserting headers but rather using strongly typed wrappers.

### DIFF
--- a/sdk/core/src/lib.rs
+++ b/sdk/core/src/lib.rs
@@ -22,7 +22,6 @@ mod pageable;
 mod pipeline;
 mod policies;
 mod request;
-mod request_options;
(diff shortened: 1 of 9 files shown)


bmc-msft (approved):

cataggar:
Can we consolidate?

https://github.com/Azure/azure-sdk-for-rust/blob/a001d10bd59406d6651cdbc717253a9cb8268f51/sdk/core/src/constants.rs#L16-L34

rylev:
@cataggar yes, eventually I'd like the constants file to just go away.

PR #935, opened 2022-07-14
simplify blob range example

bmc-msft:


rylev (approved):
Is this better as a test rather than an example?

PR #939, opened 2022-07-15
Allow wasm to compile

rylev:
This cleans up our wasm support a bit and paves the way for actual wasm support to be added in the near future.

## User experience 

Users can now compiler for `wasm32` targets as long as the `reqwest` features are not turned on (since reqwest does not have wasm support). 

If the user compiles for `wasm32` targets with the `reqwest` features enabled (as they are by default), they see the following error.

```
$ cargo c --target=wasm32-unknown-unknown -p azure_data_cosmos 

Checking azure_core v0.3.0 (/home/rylevick/code/azure-sdk/azure-sdk-for-rust/sdk/core)
error: The `enable_request` and `enable_reqwest_rustls` features are not allowed for `wasm32` targets`
  --> sdk/core/src/http_client/mod.rs:11:1
   |
11 | compile_error!("The `enable_request` and `enable_reqwest_rustls` features are not allowed for `wasm32` targets");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `azure_core` due to previous error
```

## The noop client

We currently don't support any way for users to make requests when compiling for the `wasm32` target. In the future, we will provide an easy way for users to supply there own transport policies. Users who compile for wasm in the browser, can provide a transport policy that uses the browser APIs, while users compiling for wasi, can use a transport policy that uses wasi based APIs.

Since there is (IMO) no obvious default, we have a `NoopClient` that simply panics and prints out the outgoing request. Most users who will be using the `reqwest` feature, will never run into this.

## Some challenges

This solution isn't 100% elegant, but is IMO an improvement over the status quo. There are some downsides though:
* We do enforce `Send` futures for wasm targets. This causes us to have to propogate this out since by default `BoxFuture` and the `Future`s created by the `async_trait` macro are all constrained to be `Send`. So, we still have a lot of `cfg` markers for wasm, but there in much more predictable places and we don't have to add a bunch of `allow(unused)` directives everywhere.
*

### DIFF
--- a/sdk/core/src/auth.rs
+++ b/sdk/core/src/auth.rs
@@ -33,7 +33,8 @@ impl TokenResponse {
 }
 
 /// Represents a credential capable of providing an OAuth token.
-#[async_trait::async_trait]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 pub trait TokenCredential: Send + Sync {
     /// Gets a `TokenResponse` for the specified resource
     async fn get_token(&self, resource: &str) -> crate::Result<TokenResponse>;
--- a/sdk/core/src/bytes_stream.rs
+++ b/sdk/core/src/bytes_stream.rs
@@ -56,7 +56,8 @@ impl Stream for BytesStream {
     }
 }
 
-#[async_trait::async_trait]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 impl SeekableStream for BytesStream {
     async fn reset(&mut self) -> crate::Result<()> {
         self.bytes_read = 0;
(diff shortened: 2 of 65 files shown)


rylev:
@ctaggart @bmc-msft this should be ready to merge!

bmc-msft (commented):
In all of our uses, `azure_core` is set to default-features=false.  Perhaps we should not set default features for azure_core, and have it's dependencies do that?

```
❯ rg -I ^azure_core -g '**/Cargo.toml' | sort -u
azure_core = { path = "../../../sdk/core", version = "0.3", default-features = false }
azure_core = { path = "../core", version = "0.3", default-features = false }
azure_core = { path = "../core", version = "0.3", default-features = false}
azure_core = { path = "../core", version = "0.3", default-features=false }
azure_core = { path = "../core", version = "0.3", default_features = false }
❯
```

bmc-msft (approved):

cataggar (approved):

rylev:
@radu-matei awesome! I hope to have a change soon that allows for easily swapping out the transport policy. Once that's done, it should hopefully be easy to have a `wasi` backed transport policy that can be swapped in without the need for modifying any of the code.

PR #943, opened 2022-07-19
move to using azure_core::Body instead of Bytes

bmc-msft:
While this doesn't actually enable users to AsyncRead+AsyncSync yet, this is a step towards that.

### DIFF
--- a/sdk/core/src/request.rs
+++ b/sdk/core/src/request.rs
@@ -20,6 +20,10 @@ impl Body {
             Body::SeekableStream(stream) => stream.len(),
         }
     }
+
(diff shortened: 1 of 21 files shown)


sdk/core/src/request.rs
```diff
@@ -13,6 +13,19 @@ pub enum Body {
     SeekableStream(Box<dyn SeekableStream>),
 }
 
+impl Body {
+    pub fn len(&self) -> usize {
```
rylev, line 17: While these functions are pretty self-explanatory, I think it's a good habit to always add documentation for all public methods.

rylev (approved):

PR #944, opened 2022-07-19
Fix `IntoFuture` bug

rylev:
The use of `into_future` broke, and we didn't notice. This fixes the issue, and adds `into_future` testing to CI. Once `IntoFuture` lands on stable, this can be removed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,6 +71,9 @@ jobs:
       - name: sdk tests
         run: cargo test --all --features mock_transport_framework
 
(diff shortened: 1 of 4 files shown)


yoshuawuyts (approved):

cataggar:
Can #687 be closed?

rylev:
@cataggar not quite. When the `IntoFuture` language feature stabilizes, the `into_future` feature flag needs to be removed and the example code needs to be updated to not directly call `into_future`. This is what #687 still does.

PR #951, opened 2022-07-20
specify additional conditional headers for blob & container operations

bmc-msft:
Flush out the rest of the conditional operations headers for blob & container operations

Ref: https://docs.microsoft.com/en-us/rest/api/storageservices/specifying-conditional-headers-for-blob-service-operations 

With this change, users will be able to do things like:

```
    let yesterday = Utc::now() - chrono::Duration::days(1);
    blob_client
        .delete()
        .if_modified_since(IfModifiedSinceCondition::Modified(yesterday))
        .into_future()
        .await?;
```


sdk/storage_blobs/src/blob/operations/clear_page.rs
```diff
@@ -10,9 +10,9 @@ operation! {
     ClearPage,
     client: BlobClient,
     ba512_range: BA512Range,
-    ?sequence_number_condition: SequenceNumberCondition,
-    ?if_modified_since_condition: IfModifiedSinceCondition,
-    ?if_match_condition: IfMatchCondition,
+    ?if_sequence_number: SequenceNumberCondition,
```
rylev, line 13: Should we change `SequenceNumberCondition` to `IfSequenceNumber`?

sdk/storage_blobs/src/blob/operations/clear_page.rs
```diff
@@ -27,9 +27,9 @@ impl ClearPageBuilder {
             headers.insert(PAGE_WRITE, "clear");
             headers.insert(BLOB_TYPE, "PageBlob");
             headers.add(self.ba512_range);
-            headers.add(self.sequence_number_condition);
-            headers.add(self.if_modified_since_condition);
-            headers.add(self.if_match_condition);
+            headers.add(self.if_sequence_number);
```
rylev, line 30: It would be nice if we could figure out how to apply these headers without having to write this code over and over in different operations.

sdk/storage_blobs/src/options/condition_tags.rs
```diff
@@ -0,0 +1,26 @@
+use azure_core::headers::{Header, HeaderName, HeaderValue};
+
+#[derive(Debug, Clone)]
+pub struct IfTagsCondition(String);
```
rylev, line 4: We've been using `Cow<'static, str>` with great success. Should we consider that here?

rylev (approved):
Looks good. Just had some thoughts for us to noodle on, but nothing actionable right now.

sdk/storage_blobs/src/options/condition_tags.rs
bmc-msft: Opened PR to do that

sdk/storage_blobs/src/blob/operations/clear_page.rs
bmc-msft: Opened PR

PR #952, opened 2022-07-21
Remove ToJsonVector

### LINKED ISSUE #104: Cosmos: Consider removing ToJsonVector
Currently `PartitionKeys` and `stored_procedure::Parameters` are simple aliases of `ToJsonVector`. I'm not convinced that this wins us anything over simply using a wrapper around `Vec<String>` for each of these (or simply `Vec<String>` itself in the case of `Parameters`). Switching to a simpler representation would most likely make these types easier to understand as well.

rylev:
Fixes #104


yoshuawuyts (approved):
It makes sense to me to move the logic into cosmos, and to simplify the construction of it by keeping a `Vec<String>` instead of `String` internally. LGTM.

PR #953, opened 2022-07-21
move to using Cow<'static, str> for most headers/request options

bmc-msft:


bmc-msft:
Based on comments from #951

sdk/core/src/macros.rs
```diff
@@ -290,6 +290,98 @@ macro_rules! future {
...
+}
+
+/// The following macro invocation:
+/// ```
+/// create_request_query_cow!(Prefix, "prefix");
+/// ```
+/// Turns into a request query option used to construct requests
+#[macro_export]
+macro_rules! create_request_query_cow {
+    ($name:ident, $option:expr $(, $doc:expr)?) => {
```
rylev, line 352: Can we use `($(#[$doc:meta])*` for the docs so that we don't have to write docs as string literals?

sdk/core/src/macros.rs
```diff
@@ -290,6 +290,98 @@ macro_rules! future {
...
+    };
+}
+
+/// The following macro invocation:
+/// ```
+/// create_request_query_cow!(Prefix, "prefix");
+/// ```
+/// Turns into a request query option used to construct requests
+#[macro_export]
+macro_rules! create_request_query_cow {
```
rylev, line 351: Don't love the name -  perhaps `request_query_option!`?

sdk/core/src/macros.rs
```diff
@@ -290,6 +290,100 @@ macro_rules! future {
...
+        $( #[doc = $doc] )?
+        #[derive(Debug, Clone)]
+        pub struct $name(std::borrow::Cow<'static, str>);
+
+        impl $name {
+            $(
+                pub const $variant: $name = $name::from_static($value);
+            )*
+
+            pub fn new<S>(s: S) -> Self
```
rylev, line 311: Seems like most of this stuff is shared. Can we share the definition? We can probably do all of this in one macro since the invocation is unambiguous between the two use cases.

bmc-msft: There needs to be some disambiguation because of either Header or AppendToUrlQuery implementation.

rylev (approved):

PR #954, opened 2022-07-21
rename SequenceNumberCondition to IfSequenceNumber (per #951)

bmc-msft:


sdk/core/src/request_options/if_sequence_number.rs
```diff
@@ -0,0 +1,26 @@
+use crate::{headers, Header};
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum IfSequenceNumber {
```
rylev, line 4: It would be good to have docs for all these options.

sdk/core/src/request_options/if_sequence_number.rs
```diff
@@ -0,0 +1,26 @@
...
+    Less(u64),
```
rylev, line 5: Maybe `LessThan` is clearer?

rylev (approved):

PR #956, opened 2022-07-22
Improve storage operations

rylev:
* Improve the example for the `get_account` storage operation
* Make `find_blobs_by_tags` a stream instead of future
* Move `find_blobs_by_tags` into the `azure_blob_stroage` since it's blob specific and not general to all storage


bmc-msft (approved):
Note:

Prior to this change, the operations that live in `blob/operations` could be summarized as "For a given blob, do these things" and the operations that live in `container/operations` could be summarized as "For a given container, do these things".

While `find_blobs_by_tags` is finds blobs, it operates on an account, not a blob.

rylev:
@bmc-msft I moved the operation to the "service" module which seems to be the right place for it to live.

PR #959, opened 2022-07-25
move keyvault to use builder pattern

bmc-msft:
NOTE: This does *not* update the keyvault to use the pipelines yet, that will be a follow-on PR


sdk/identity/examples/client_certificate_credentials.rs
```diff
@@ -7,7 +7,7 @@ use azure_core::auth::TokenCredential;
 use azure_identity::{
     CertificateCredentialOptions, ClientCertificateCredential, DefaultAzureCredential,
 };
-use azure_security_keyvault::KeyClient;
+use azure_security_keyvault::KeyvaultClient;
```
heaths, line 10: As with all our other language SDKs, we have separate clients for Keys and Cryptography, Secrets, and Certificates. They were named correctly already. These should idiomatically but consistently match our other SDKs unless there's a good reason not to.

sdk/security_keyvault/src/account/operations/list_secrets.rs
```diff
@@ -0,0 +1,55 @@
...
+
+                secrets.extend(
+                    response
+                        .value
+                        .into_iter()
+                        .map(|s| KeyVaultSecretBaseIdentifier {
+                            id: s.id.clone(),
+                            name: s.id.split('/').last().unwrap().to_owned(),
+                            enabled: s.attributes.enabled,
+                            time_created: s.attributes.created,
```
heaths, line 38: `created_on`. See https://docs.microsoft.com/en-us/dotnet/api/azure.security.keyvault.secrets.secretproperties?view=azure-dotnet

sdk/security_keyvault/src/account/operations/list_secrets.rs
```diff
@@ -0,0 +1,55 @@
...
+                            time_updated: s.attributes.updated,
```
heaths, line 39: `updated_on`

sdk/security_keyvault/src/keys/operations/get_key.rs
```diff
@@ -0,0 +1,32 @@
...
+
+impl GetKeyBuilder {
+    pub fn into_future(mut self) -> GetKey {
+        Box::pin(async move {
+            let mut uri = self.client.client.vault_url.clone();
+            let path = if let Some(ver) = self.version {
+                format!("keys/{}/{}", self.client.name, ver)
+            } else {
+                format!("keys/{}", self.client.name)
+            };
```
heaths, line 17: FWIW, this doesn't hurt but it's not necessary. The Key Vault service already correctly handles if the version is elided and you try to GET something like `/keys/{key-name}//?api-version=7.1`.

heaths (changes requested):
A few nits, but the main blocker is that the existing client names e.g., `KeyClient` are what customers should be familiar with and invoke. Perhaps this `KeyvaultClient` is an internal implementation used to make it easier, but we should try to be idiomatic yet consistent.

sdk/identity/examples/client_certificate_credentials.rs
bmc-msft: I see the confusion here.  I was trying to be consistent with other services thinking the KeyVault client was not consistent.

Consider Storage Queues, with [QueueServiceClient.get_queue_client](https://docs.microsoft.com/en-us/python/api/azure-storage-queue/azure.storage.queue.queueserviceclient?view=azure-python#azure-storage-queue-queueserviceclient-get-queue-client), or Service Bus, with [ServiceBusClient.get_queue_receiver](https://docs.microsoft.com/en-us/python/api/azure-storage-queue/azure.storage.queue.queueserviceclient?view=azure-python#azure-storage-queue-queueserviceclient-get-queue-client), or Storage Blobs with [BlobServiceClient.get_container_client](https://docs.microsoft.com/en-us/azure/developer/python/sdk/storage/azure-storage-blob/azure.storage.blob.blobserviceclient?view=storage-py-v12#get-container-client-container-)(...).[get_blob_client](https://docs.microsoft.com/en-us/azure/developer/python/sdk/storage/azure-storage-blob/azure.storage.blob.containerclient?view=storage-py-v12#get-blob-client-blob--snapshot-none-).

sdk/security_keyvault/src/keys/operations/get_key.rs
bmc-msft: This was just refactored original code, but I will address.

sdk/security_keyvault/examples/get_secret.rs
```diff
@@ -18,9 +18,9 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         client_secret,
         TokenCredentialOptions::default(),
     ));
-    let client = KeyvaultClient::new(&keyvault_url, creds)?.secret_client(secret_name);
+    let client = SecretClient::new(&keyvault_url, creds)?;
```
heaths, line 21: So what does setting options look like now with a builder pattern? When I last updated this crate, it was using the `ClientOptions` I originally proposed.

bmc-msft: Per the discussion at SDK meeting, the builder pattern for creating clients is a work in progress by @yoshuawuyts .  We'll update this crate once we have flushed out that design.

heaths (approved):

PR #960, opened 2022-07-27
Move IoT Hub to operation builder pattern

rylev:
This changed the user interface a little bit, and made it a tiny bit worse IMO, but I think we can easily regain that usability in a follow up. 

NOTE: This does *not* update IoT Hub to use the pipeline yet. That will be a follow-up PR.

This is part of the work for #802


sdk/iot_hub/examples/configuration.rs
```diff
@@ -38,7 +38,11 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync>> {
     );
 
     println!("Getting configuration: {}", configuration_id);
-    let configuration = service_client.get_configuration(configuration_id).await?;
+    let configuration = service_client
+        .get_configuration(configuration_id)
+        .into_future()
+        .await?;
+    let configuration: Configuration = configuration.try_into()?;
```
rylev, line 45: This is a bit of a shame that we are now returning an unstructured type.

sdk/core/src/macros.rs
```diff
@@ -139,6 +140,7 @@ macro_rules! operation {
         ) => {
         azure_core::__private::paste! {
         #[derive(Debug, Clone)]
+        $(#[$outer])*
```
bmc-msft, line 143: 👍

sdk/iot_hub/src/service/requests/create_or_update_configuration.rs
```diff
@@ -0,0 +1,127 @@
...
+
+azure_core::operation! {
+    /// The CreateOrUpdateConfigurationBuilder is used to construct a new configuration
+    /// or update an existing one.
+    CreateOrUpdateConfiguration,
+    client: ServiceClient,
+    configuration_id: String,
+    priority: u64,
+    target_condition: String,
+    etag: Option<String>,
```
bmc-msft, line 19: Instead of exposing two operations, we could do what the Python SDK does, which is to have a single `create_or_update` function, and provide the optional parameter `if_match`.

https://docs.microsoft.com/en-us/python/api/azure-mgmt-iothub/azure.mgmt.iothub.v2021_07_02.operations.iothubresourceoperations?view=azure-python#azure-mgmt-iothub-v2021-07-02-operations-iothubresourceoperations-begin-create-or-update

> if_match
> ETag of the IoT Hub. Do not specify for creating a brand new IoT Hub. Required to update an existing IoT Hub.

bmc-msft: This could be done across the crate.

sdk/iot_hub/src/service/requests/delete_configuration.rs
```diff
@@ -0,0 +1,37 @@
+use crate::service::{ServiceClient, API_VERSION};
+
+use azure_core::headers;
+use azure_core::Method;
+
+azure_core::operation! {
+    /// The DeleteConfigurationBuilder is used to construct a request to delete a configuration.
+    DeleteConfiguration,
+    client: ServiceClient,
+    if_match: String,
```
bmc-msft, line 10: Similar with above, we should standardize on `etag` vs `if_match`.  Additionally, this should be optional.

https://docs.microsoft.com/en-us/rest/api/iothub/service/configuration/delete#request-headers

sdk/iot_hub/src/service/requests/create_or_update_configuration.rs
rylev: That sounds fine. I think the entire client interface needs to be rethought through[^1] so maybe we leave that for a different PR?

[^1] This SDK does not follow the "client tree" model and instead ops for one mega client. We should probably move away from that.

bmc-msft (approved):

PR #964, opened 2022-07-29
Add `CosmosClientBuilder` to the Cosmos SDK

yoshuawuyts:
This patch adds a `ClientBuilder` to cosmos, which allows changing the settings on the client prior to construction.

## Examples

```rust
use azure_core::{RetryOptions, TransportOptions};
use azure_data_cosmos::CosmosClient;

let _database = CosmosClient::builder(args.account, authorization_token)
    .retry(RetryOptions::default())
    .transport(TransportOptions::default())
    .build()
    .database_client(args.database_name);
```

## Design

This replaces the `with_` methods on `CosmosClient` and moves them to `CosmosClientBuilder`. Using the builder various things can be configured on the client, such as the transport and retry policy. We've chosen to go with a separate builder type instead of adding the methods on the main client, because once constructed the pipeline inside the client needs to be `Arc`'ed - and we didn't want to require it to require an additional `RwLock` to function.

This patch intentionally has decided to go with adding methods to set specific groups of options. We could've taken this further by allowing each setting of each option to directly be set on the client. But because we have many different SDKs, the set of options may expand over time, and we will want to support reuse of options between SDKs, this should work as a reasonable middle-ground.

## Future Directions

This PR doesn't yet touch the options themselves, which currently support three different styles:

- `Default::default`
- struct literal instantiation (which means adding new fields to the struct is semver-incompatible)
- setters/getters

We almost certainly will want to convert the options to be in a builder style as well, but that's intentionally left for a follow-up patch. The intent for this PR was to setup just enough of an API surface that the other SDKs can follow the design without needing to do much more breakage in the future.

The other bit that this PR doesn't yet address is setting custom policies in the pipeline. The most important item to address was providing a way to disable the built-in retry middleware, which is now possible by passing the `retry` setting on the builder.

---

I hope this all makes sense; keen to hear what folks think!

### DIFF
--- a/sdk/data_cosmos/src/clients/cosmos.rs
+++ b/sdk/data_cosmos/src/clients/cosmos.rs
@@ -23,6 +23,7 @@ pub struct CosmosClientBuilder {
 
 impl CosmosClientBuilder {
     /// Create a new instance of `CosmosClientBuilder`.
+    #[must_use]
     pub fn new(account: impl Into<String>, auth_token: AuthorizationToken) -> Self {
         Self::with_location(CloudLocation::Public {
             account: account.into(),
@@ -31,6 +32,7 @@ impl CosmosClientBuilder {
     }
 
     /// Create a new instance of `CosmosClientBuilder` with a cloud location.
+    #[must_use]
     pub fn with_location(cloud_location: CloudLocation) -> Self {
         Self {
             options: ClientOptions::default(),
@@ -39,6 +41,7 @@ impl CosmosClientBuilder {
     }
 
     /// Convert the builder into a `CosmosClient` instance.
+    #[must_use]
     pub fn build(self) -> CosmosClient {
         let auth_token = self.cloud_location.auth_token();
         CosmosClient {
@@ -48,25 +51,21 @@ impl CosmosClientBuilder {
     }
 
     /// Set the cloud location.
+    #[must_use]
     pub fn cloud_location(mut self, cloud_location: CloudLocation) -> Self {
         self.cloud_location = cloud_location;
         self
     }
 
-    /// Enable the mocking framework.
-    #[cfg(feature = "mock_transport_framework")]
-    pub fn mock(self, enable: bool) -> Self {
-        self.mock = enable;
-        self
-    }
-
     /// Set the retry options.
+    #[must_use]
     pub fn retry(mut self, retry: impl Into<azure_core::RetryOptions>) -> Self {
         self.options = self.options.retry(retry);
         self
     }
 
     /// Set the transport options.
+    #[must_use]
     pub fn transport(mut self, transport: impl Into<azure_core::TransportOptions>) -> Self {
         self.options = self.options.transport(transport);
         self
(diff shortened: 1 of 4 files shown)


sdk/data_cosmos/src/clients/cosmos.rs
```diff
@@ -14,60 +14,77 @@ use std::sync::Arc;
...
     }
 
-    #[cfg(feature = "mock_transport_framework")]
-    /// Create a new instance of `CosmosClient` using a mock backend. The
-    /// transaction name is used to look up which files to read to validate the
-    /// request and mock the response.
-    // TODO(yosh): consider adding a general way to replace transports, and remove this method.
-    pub fn with_mock(
+    /// Set the cloud location.
+    pub fn cloud_location(mut self, cloud_location: CloudLocation) -> Self {
```
rylev, line 51: Do these need a `#[must_use]` on them?

rylev (commented):
I'm not sure how I feel about this change. I think I like *most* of it, but there's a part I'm not sure I like.

In particular, I'm unsure about the requirement to forward options methods through `ClientBuilder` to the underlying `ClientOptions` instead of just exposing `ClientOptions` to the user. Would it make sense to instead have a method `client_options` that seems allows users to set whatever `ClientOptions` they want? We can then make the API for building a `ClientOptions` follow the builder pattern. This removes the need to forward everything to the `ClientOptions` through the `ClientBuilder`.

gorzell:
> Would it make sense to instead have a method `client_options` that seems allows users to set whatever `ClientOptions` they want?

I agree with this, I think it is rather cumbersome to proxy every sub-option onto the builder, instead we should try to make the options themselves more composable/consistent/ergonomic.

> This PR doesn't yet touch the options themselves, which currently support three different styles:
> 
>     * `Default::default`
> 
>     * struct literal instantiation (which means adding new fields to the struct is semver-incompatible)
> 
>     * setters/getters

I noticed this in the context of `StorageOptions` which actually has private struct values, with **no!?!** setters, but then `ClientOptions` has `pub(crate)`, and `RetryOptions` is just `pub`. We need to decide on an approach and make these all consistent.

sdk/data_cosmos/src/clients/cosmos.rs
yoshuawuyts: Oh yeah, good call!

rylev (approved):
I talked with Yosh about this, and we both believe this is likely the best route forward. This offers a way to control options in a form that is both idiomatic and not too much of a maintenance burden. 

You can see discussion of the alternative presented in https://github.com/Azure/azure-sdk-for-rust/pull/975.

gorzell:
I thought that there had been some consensus that it would be better to just have a `with_options(azure_core::ClientOptions)` function rather than separate functions for `retry` and `transport`. I looked at #975 and it says this isn't possible, but it is not obvious to me from the code in either PR why it wouldn't be. What am I missing?

I still think that is a much more maintainable approach and the ergonomics don't really suffer for it.

rylev:
Talked with @gorzell about this. Some points:
* There's agreement that an `with_options(azure_core::ClientOptions)` approach would have probably been better if we never anticipated SDK specific options. 
* However, we want to avoid confronting the user with fact that some options as SDK specific and some are shared across many SDKs. Having `with_options(azure_core::ClientOptions)`  and `with_sdk_options(azure_data_cosmos::ClientOptions)` would not be a great user experience.
* The maintence burden is somewhat small with the given proposal. However, it is true that every time `azure_core::ClientOptions` gains new functionality we might have to update the SDKs. This might be avoidable with a macro, but it is a bit of a bummer. 

All in all, there's no solution that is clearly ideal. I believe the option merged in this PR is the best we have available. We can continue to experiment though.

heaths:
> Talked with @gorzell about this. Some points:
> 
> * There's agreement that an `with_options(azure_core::ClientOptions)` approach would have probably been better if we never anticipated SDK specific options.
> * However, we want to avoid confronting the user with fact that some options as SDK specific and some are shared across many SDKs. Having `with_options(azure_core::ClientOptions)`  and `with_sdk_options(azure_data_cosmos::ClientOptions)` would not be a great user experience.
> * The maintence burden is somewhat small with the given proposal. However, it is true that every time `azure_core::ClientOptions` gains new functionality we might have to update the SDKs. This might be avoidable with a macro, but it is a bit of a bummer.
> 
> All in all, there's no solution that is clearly ideal. I believe the option merged in this PR is the best we have available. We can continue to experiment though.

@yoshuawuyts @rylev this makes sense, but what about an escape hatch that allows passing a ClientOptions? Or is it even worth it since conceptually adding options to the base ClientOptions means we have to touch everything anything?

PR #965, opened 2022-08-01
switch from `chrono` crate to `time` crate

ctaggart:
The `time` crate has been well maintained and offers a better API than `chrono`. The `time::OffsetDateTime` type is similar to .NET's [System.DateTimeOffset](https://docs.microsoft.com/en-us/dotnet/api/system.datetimeoffset?view=net-6.0). `chrome::Duration` has been replaced by `std::time::Duration`. A new `azure_core::date` module contains functions for parsing and formatting formats and creating new durations.

The Azure REST API Specifications use either `"format": "date-time"` or `"format": "date-time-rfc1123"`.  These correspond to RFC 3339 and RFC 1123 date formats. The pull request replaces usage of RFC 2822 with RFC 1123 where the zone must be `GMT`.


sdk/core/src/date.rs
```diff
@@ -0,0 +1,185 @@
...
+/// RFC 2822: Internet Message Format
+///
+/// https://www.rfc-editor.org/rfc/rfc2822#section-3.3
+/// https://www.rfc-editor.org/rfc/rfc2822#section-4.3
+/// A "GMT" zone is obsolete, instead "+0000" is used.
+///
+/// Date: Fri, 21 Nov 1997 10:01:10 -0600
+pub fn parse_rfc2822(s: &str) -> crate::Result<OffsetDateTime> {
+    OffsetDateTime::parse(s, &Rfc2822).with_context(ErrorKind::DataConversion, || {
+        format!("unable to parse smtp date '{s}")
```
rylev: Is this the right error message?

sdk/core/src/date.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+/// RFC 3339: Date and Time on the Internet: Timestamps
+///
+/// https://www.rfc-editor.org/rfc/rfc3339
+///
+/// 1985-04-12T23:20:50.52Z
+pub fn parse_rfc3339(s: &str) -> crate::Result<OffsetDateTime> {
+    OffsetDateTime::parse(s, &Rfc3339).with_context(ErrorKind::DataConversion, || {
+        format!("unable to parse internet date '{s}")
```
rylev: I think these error messages should reflect the same nomenclature used in the method names.

sdk/core/src/date.rs
```diff
@@ -0,0 +1,185 @@
...
+/// Preferred HTTP date format
+///
+/// https://httpwg.org/specs/rfc9110.html#http.date
+///
+/// In Azure REST API specification it is specified as `"format": "date-time-rfc1123"`. In .NET it is:
+///
+/// https://docs.microsoft.com/dotnet/api/system.globalization.datetimeformatinfo.rfc1123pattern
+///
+/// Sun, 06 Nov 1994 08:49:37 GMT
+pub fn parse_http_date(s: &str) -> crate::Result<OffsetDateTime> {
```
rylev, line 71: If we're using the rfc naming for others, perhaps we should do that here too?

sdk/core/src/error/mod.rs
```diff
@@ -257,6 +257,12 @@ impl From<url::ParseError> for Error {
     }
 }
 
+impl From<time::error::Parse> for Error {
```
rylev, line 260: I'm unsure if we should have this. It might be nice to require that users add a context when this error occurs. Otherwise the error message might be a bit obtuse.

sdk/core/src/serde/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod http_date;
```
rylev, line 1: Not sure how I feel about having a `serde` module. Is there any reason this can't live in the date module?

sdk/core/src/sleep.rs
```diff
@@ -25,7 +25,7 @@ impl Future for Sleep {
             let waker = cx.waker().clone();
             let duration = self.duration;
             self.get_mut().thread = Some(thread::spawn(move || {
-                thread::sleep(duration);
+                thread::sleep(duration.try_into().unwrap_or_default());
```
rylev, line 28: Is this because the std `Duration` is not signed? I'm not sure I like this... Might be best to keep using the std library's duration in this case?

sdk/core/src/date.rs
cataggar: I thought I updated this last night in commit 82c63f8c6166da225e11092fc0091c79e48868ca. It should say `parse rfc2822 date`. I'll check again.

sdk/core/src/sleep.rs
cataggar: [time::Duration](https://docs.rs/time/latest/time/struct.Duration.html) uses a i64 of seconds and `std::time::Duration` uses an u64 of seconds to allow for negative durations. It felt a bit more consistent to switch to `time::Duration`. The API felt a bit nicer as well, but we could use `std::time::Duration`.

sdk/core/src/date.rs
cataggar: Me too. 82c63f8c6166da225e11092fc0091c79e48868ca had already changed it.

sdk/core/src/date.rs
cataggar: I renamed it `parse_rfc1123`. This matches the Azure specs and .NET library.

sdk/core/src/serde/mod.rs
cataggar: Done.

sdk/core/src/error/mod.rs
cataggar: I've removed `From<time::error::Parse>`  as requested.

sdk/core/src/sleep.rs
cataggar: I migrated to `std::time::Duration`.

cataggar:
@rylev, thanks for the review! All your requested changes have been made.

rylev (approved):
This is a pretty big change so I didn't go through every line with a fine-tooth comb, but overall, it looks like a solid change.

PR #966, opened 2022-08-01
storage_blobs: Add support for snapshot blob.

gorzell:
Add support to `BlobClient` for taking snapshots of a blob. For now this
reuses the opaque type that already exists for `Snapshot`, but it might
be worth considering adding support for treating it as a `DateTime`.


rylev (changes requested):
Thanks! This looks good. 

Would you mind changing the test to use the mocking framework? There is information [here](https://github.com/Azure/azure-sdk-for-rust/blob/main/docs/mock_transport.md) about how that's supposed to work. If the docs are clear enough, let me know and we can update them.

gorzell:
> Thanks! This looks good.
> 
> Would you mind changing the test to use the mocking framework? There is information [here](https://github.com/Azure/azure-sdk-for-rust/blob/main/docs/mock_transport.md) about how that's supposed to work. If the docs are clear enough, let me know and we can update them.

Sure thing. Is there a standard place to land those tests? Should get go in the actual `src` file as unit tests since they don't have any external dependencies?

rylev:
@gorzell indeed - the cosmos crate is the one to follow. Putting them in a separate tests directory is the way to go. I don't believe `DefaultAzureCredential` is very full featured and so it might not offer any real benefit and would make the examples potentially slightly harder to understand.

sdk/storage/src/core/clients/storage_client.rs
```diff
@@ -337,6 +337,42 @@ impl StorageClient {
         }
     }
 
+    #[cfg(feature = "mock_transport_framework")]
+    /// Create a new instance of `CosmosClient` using a mock backend. The
+    /// transaction name is used to look up which files to read to validate the
+    /// request and mock the response.
+    // TODO(yosh): consider adding a general way to replace transports, and remove this method.
```
rylev, line 344: You can remove this (it refers to Cosmos anyway) - when we change how the mock stuff is initialized, I'll make sure this gets changed.

sdk/storage_blobs/src/blob/operations/snapshot_blob.rs
```diff
@@ -0,0 +1,68 @@
...
+            request_id: request_id_from_headers(headers)?,
+            etag: etag_from_headers(headers)?,
+            date: date_from_headers(headers)?,
+            snapshot: Snapshot::new(headers.get_str(&SNAPSHOT)?.to_string()),
+            last_modified: last_modified_from_headers(headers)?,
+        })
+    }
+}
+
+pub const SNAPSHOT: HeaderName = HeaderName::from_static("x-ms-snapshot");
```
rylev, line 68: Unsure if this is the right place for this. I'll think about it a bit more...

sdk/storage_blobs/tests/blob_operations.rs
```diff
@@ -0,0 +1,61 @@
...
+        .hash(digest)
+        .into_future()
+        .await
+        .unwrap();
+
+    trace!("created {:?}", BLOB_NAME);
+
+    let snapshot = blob.snapshot().into_future().await.unwrap().snapshot;
+
+    trace!("crated snapshot: {:?} of {:?}", snapshot, BLOB_NAME);
```
rylev, line 56: It would be nice to assert something here about the snapshot. Otherwise we're only testing that the deserialization doesn't outright fail.

test/transactions/put_block_blob_and_snapshot/0_response.json
```diff
@@ -0,0 +1 @@
...1Nb2RpZmllZD5UaHUsIDE0IEp1bCAyMDIyIDIyOjQzOjUxIEdNVDwvTGFzdC1Nb2RpZmllZD48RXRhZz4iMHg4REE2NUVBNTMzQkVFRjgiPC9FdGFnPjxMZWFzZVN0YXR1cz51bmxvY2tlZDwvTGVhc2VTdGF0dXM+PExlYXNlU3RhdGU+YXZhaWxhYmxlPC9MZWFzZVN0YXRlPjxEZWZhdWx0RW5jcnlwdGlvblNjb3BlPiRhY2NvdW50LWVuY3J5cHRpb24ta2V5PC9EZWZhdWx0RW5jcnlwdGlvblNjb3BlPjxEZW55RW5jcnlwdGlvblNjb3BlT3ZlcnJpZGU+ZmFsc2U8L0RlbnlFbmNyeXB0aW9uU2NvcGVPdmVycmlkZT48SGFzSW1tdXRhYmlsaXR5UG9saWN5PmZhbHNlPC9IYXNJbW11dGFiaWxpdHlQb2xpY3k+PEhhc0xlZ2FsSG9sZD5mYWxzZTwvSGFzTGVnYWxIb2xkPjwvUHJvcGVydGllcz48L0NvbnRhaW5lcj48Q29udGFpbmVyPjxOYW1lPnRlc3Q8L05hbWU+PFByb3BlcnRpZXM+PExhc3QtTW9kaWZpZWQ+VGh1LCAxNCBKdWwgMjAyMiAyMjo0Mzo1MSBHTVQ8L0xhc3QtTW9kaWZpZWQ+PEV0YWc+IjB4OERBNjVFQTUzNDE2OTI3IjwvRXRhZz48TGVhc2VTdGF0dXM+dW5sb2NrZWQ8L0xlYXNlU3RhdHVzPjxMZWFzZVN0YXRlPmF2YWlsYWJsZTwvTGVhc2VTdGF0ZT48RGVmYXVsdEVuY3J5cHRpb25TY29wZT4kYWNjb3VudC1lbmNyeXB0aW9uLWtleTwvRGVmYXVsdEVuY3J5cHRpb25TY29wZT48RGVueUVuY3J5cHRpb25TY29wZU92ZXJyaWRlPmZhbHNlPC9EZW55RW5jcnlwdGlvblNjb3BlT3ZlcnJpZGU+PEhhc0ltbXV0YWJpbGl0eVBvbGljeT5mYWxzZTwvSGFzSW1tdXRhYmlsaXR5UG9saWN5PjxIYXNMZWdhbEhvbGQ+ZmFsc2U8L0hhc0xlZ2FsSG9sZD48L1Byb3BlcnRpZXM+PC9Db250YWluZXI+PC9Db250YWluZXJzPjxOZXh0TWFya2VyIC8+PC9FbnVtZXJhdGlvblJlc3VsdHM+"}
```
rylev, line 1: I think you need to run `./eng/scripts/format_json.sh` so that all json is formatted.

rylev (approved):
Looks good! Thanks for taking the time to add the mock testing. I pretty sure you'll need to update the json docs by running the formatting script as I noted to get CI to pass, but once that happens, we can merge this.

sdk/storage_blobs/src/blob/operations/snapshot_blob.rs
gorzell: It seemed like most of the other `const` of this type were in `core` which was a bit extreme for something storage specific. It could go in `lib`, or some other central place in the `storage_blobs` crate for sure.

gorzell: More specifically: https://github.com/gorzell/azure-sdk-for-rust/blob/main/sdk/core/src/headers/mod.rs

I think it would be fine to move this to a `headers.rs` file at the top level of `storage_blob`.

sdk/storage_blobs/tests/blob_operations.rs
gorzell: I'll think about this, but it is a bit tricky since you get back an "opaque" value that happens to be a timestamp. You could assert that the value is in the past, but then you wouldn't be treating it as opaque. I plan to add `list_snaphots` in the near future, which might be the better way to improve this test?

Issue #961, opened 2022-07-27
Parse Error when calling get_block_list()


juliusl:
When calling get_block_list() I run into this problem,  

Repro:
```
    let container_client = client.container_client("testcontainer");
    let blob_client = container_client.blob_client("test");

    let block_list = blob_client.get_block_list().into_future().await?;

    println!("{:#?}", block_list);
```

Error:
```
    Error: Error { context: Full(Custom { kind: DataConversion, error: ParseError(Invalid) }, "unable to parse header 'HeaderName(\"last-modified\"): HeaderValue(\"Wed, 27 Jul 2022 20:52:56 GMT\")' into chrono::datetime::DateTime<chrono::offset::utc::Utc>") }
```

cataggar:
@juliusl, are you using `main` or a particular version?

juliusl:
> @juliusl, are you using `main` or a particular version?

I'm working directly in 'main' in the example dir for storage.

ctaggart:
Hi @juliusl, could you test pr #965? You could get the branch with `gh pr checkout 965`.

juliusl:
@cataggar  so I think it gets past the time issue but now I've run into this,

```
Error: Error { context: Custom(Custom { kind: DataConversion, error: Custom { field: "missing field UncommittedBlocks" } }) }
```

Is there a way to make these requests but get back a pointer to the unconverted response, so I can plug in my own deserialization?

juliusl:
Also, here's a patch for the bug above,

```
diff --git a/sdk/storage_blobs/src/blob/block_with_size_list.rs b/sdk/storage_blobs/src/blob/block_with_size_list.rs
index 5ac184d60..ece3ce10c 100644
--- a/sdk/storage_blobs/src/blob/block_with_size_list.rs
+++ b/sdk/storage_blobs/src/blob/block_with_size_list.rs
@@ -30,9 +30,9 @@ struct OuterBlock {
 #[derive(Debug, Deserialize)]
 struct BlockList {
     #[serde(rename = "CommittedBlocks")]
-    pub committed_blocks: OuterBlock,
+    pub committed_blocks: Option<OuterBlock>,
     #[serde(rename = "UncommittedBlocks")]
-    pub uncommitted_blocks: OuterBlock,
+    pub uncommitted_blocks: Option<OuterBlock>,
 }
 
 #[derive(Default, Debug, Clone, PartialEq)]
[...]

PR #968, opened 2022-08-01
Fix. deserialization for block blob list resp

juliusl:
Signed-off-by: Julius <juliusl@microsoft.com>

Fixes deserialization issue found in #961

### DIFF
--- a/sdk/storage_blobs/src/blob/block_with_size_list.rs
+++ b/sdk/storage_blobs/src/blob/block_with_size_list.rs
@@ -30,9 +30,9 @@ struct OuterBlock {
 #[derive(Debug, Deserialize)]
 struct BlockList {
     #[serde(rename = "CommittedBlocks")]
-    pub committed_blocks: OuterBlock,
+    pub committed_blocks: Option<OuterBlock>,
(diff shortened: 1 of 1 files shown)


sdk/storage_blobs/src/blob/block_with_size_list.rs
```diff
@@ -126,4 +126,60 @@ mod test {
         assert!(bl.blocks[1].size_in_bytes == 62);
         assert!(bl.blocks[2].size_in_bytes == 62);
     }
+
+    #[test]
+    fn test_incomplete_response() {
```
rylev, line 131: Can you add a comment here that explains that we want to explicitly test for responses that both have and don't have committed and uncommitted blocks? This is clear from this PR but might not be as clear in the future when someone is reading this code.

rylev (changes requested):

sdk/storage_blobs/src/blob/block_with_size_list.rs
juliusl: updated

rylev (approved):

PR #970, opened 2022-08-02
storage: Add support for setting StorageClient options to something other than the default.

gorzell:
As far as I could tell there was no way to change the `ClientOptions`,
retries, etc from the default values. All of the existing public
constructors just call `StorageOptions::new()`.

I chose to just add a builder type method, but I am open to other
implementations. The important thing is to provide the ability to change
these values.

### DIFF
--- a/sdk/storage/src/core/clients/storage_client.rs
+++ b/sdk/storage/src/core/clients/storage_client.rs
@@ -336,6 +336,11 @@ impl StorageClient {
         }
     }
 
+    pub fn with_options(mut self, options: StorageOptions) -> Self {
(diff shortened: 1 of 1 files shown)


rylev:
@gorzell Thanks for the PR! This is an active area of discussion. You can see a proposal in #964 for how this might work in Cosmos. We need to first decide on how we want this to look in general, and then once we do, we can make the changes to the storage crate.

So with that in mind, I'm going to close this PR. I encourage you to participate in the discussion on #964 to help push that through. Once that's merged, we can open a new PR with the changes to the Storage clients.

PR #971, opened 2022-08-02
Add better retry policy defaults

rylev:
After talking with @gorzell about retry policies I decided to take a look at improving the defaults. 

This moves to having the default be an exponential back off starting at 200ms and increasing quadratically every retry until hitting a ceiling of 30seconds. 

This seems like a better default than only retrying 3 times though it will not work for everyone including for @gorzell.

### DIFF
--- a/sdk/core/src/options.rs
+++ b/sdk/core/src/options.rs
@@ -107,17 +107,23 @@ pub struct RetryOptions {
     /// The delay between retry attempts for a fixed algorithm
     /// or the delay on which to base calculations for a back-off-based approach.
     ///
-    /// The default is 800 milliseconds.
+    /// The default is 200 milliseconds.
     pub delay: Duration,
 
     /// The maximum number of retry attempts before giving up.
     ///
-    /// The default is 3.
+    /// The default is 10.
     pub max_retries: u32,
 
-    /// The maximum permissible delay between retry attempts.
+    /// The maximum permissible elapsed time since starting to retry.
     ///
     /// The default is 1 minute.
+    pub max_elapsed: Duration,
+
+    /// The maximum permissible time between retries.
+    ///
+    /// The default is 30 seconds. For SRE reasons, this is only respected when above 1 second.
+    /// This option is ignored when using retry modes that do not change their delay time.
     pub max_delay: Duration,
 }
 
@@ -126,7 +132,7 @@ impl RetryOptions {
         mode: RetryMode => mode,
         delay: Duration => delay,
         max_retries: u32 => max_retries,
-        max_delay: Duration => max_delay,
+        max_elapsed: Duration => max_elapsed,
     }
 }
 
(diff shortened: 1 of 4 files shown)


sdk/core/src/options.rs
```diff
@@ -126,33 +132,36 @@ impl RetryOptions {
         mode: RetryMode => mode,
         delay: Duration => delay,
         max_retries: u32 => max_retries,
-        max_delay: Duration => max_delay,
+        max_elapsed: Duration => max_elapsed,
```
gorzell, line 135: **Bug:** I think that you still want to have a setter for `max_delay`?

sdk/core/src/policies/retry_policies/exponential_retry.rs
```diff
@@ -12,31 +12,75 @@ use time::OffsetDateTime;
...
+            let duration = policy.sleep_duration(retry_count);
+            durations.push(duration);
+            elapsed_time += duration; // simulate sleep
+        }
+
+        let actual = durations
+            .into_iter()
+            .map(|d| d.as_secs())
+            .collect::<Vec<_>>();
+        let expected = &[0, 0, 1, 3, 6, 12, 25, 30];
```
gorzell, line 76: **Nit:** Do this as millis so that there isn't rounding going on?

sdk/core/src/policies/retry_policies/exponential_retry.rs
```diff
@@ -12,31 +12,75 @@ use time::OffsetDateTime;
...
+
+        for (&a, &e) in actual.iter().zip(expected.iter()) {
+            // Check within one second to account for the gitter
```
gorzell, line 79:
```suggestion
            // Check within one second to account for the jitter
```

sdk/core/src/policies/retry_policies/exponential_retry.rs
```diff
@@ -12,31 +12,75 @@ use time::OffsetDateTime;
...
+            elapsed_time += duration; // simulate sleep
+        }
+
+        let actual = durations
+            .into_iter()
+            .map(|d| d.as_secs())
+            .collect::<Vec<_>>();
+        let expected = &[0, 0, 1, 3, 6, 12, 25, 30];
+
+        for (&a, &e) in actual.iter().zip(expected.iter()) {
```
gorzell, line 78: **Concern:** While still valid, I think this is truncating because you get to the default `max_elapsed` before you get to the end of `expected`.

sdk/core/src/policies/retry_policies/exponential_retry.rs
gorzell: Another benefit of doing this as millis is that you can use that actual max value of the jitter which should be 255.

sdk/core/src/options.rs
```diff
@@ -126,33 +132,36 @@ impl RetryOptions {
...
     fn default() -> Self {
         RetryOptions {
             mode: RetryMode::default(),
-            delay: Duration::from_millis(800),
-            max_retries: 3,
-            max_delay: Duration::from_secs(60),
+            delay: Duration::from_millis(200),
+            max_retries: 10,
+            max_elapsed: Duration::from_secs(60),
+            max_delay: Duration::from_secs(30),
```
gorzell, line 146: **Suggestion:** It might be worth trying to align the defaults, i.e. so that you aren't setting `max_retries` to a value that is unattainable due to `max_elapsed`. Otherwise it can be a bit misleading wrt to what behavior you would expect from the default. I guess this will be a bit hard because it differs by policy, so maybe a doc comment would also be enough?

sdk/core/src/policies/retry_policies/exponential_retry.rs
```diff
@@ -12,31 +12,75 @@ use time::OffsetDateTime;
...
         Duration::from_millis(sleep_ms)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn increases_correctly() {
```
gorzell, line 53:
```suggestion
    fn exponential_increases_correctly() {
```

gorzell (commented):
This looks like a good improvement to me. I left a few comments about things that might be worth addressing before merging.

sdk/core/src/policies/retry_policies/exponential_retry.rs
rylev: I'm not sure we care enough about that level of fidelity. I think we really just want to make sure that the general path of increasing feels correct.

sdk/core/src/options.rs
rylev: I added a document explaining this.

sdk/core/src/policies/retry_policies/exponential_retry.rs
rylev: This is no longer the case.

gorzell (commented):
I am bit concerned that not all policies support `max_elapsed`, I can kind of see the logic behind them not needing to (because it is a more straight forward calculation), but it is still confusing.

When looking through that I also saw that `FixedRetryPolicy` is actually using `max_delay: Duration` which doesn't make any sense at all given that it also take `delay`.

gorzell (approved):

cataggar (approved):

PR #972, opened 2022-08-02
Migrate iothub to pipelines

roeap:
part of #802

Migrating the iothub sdk to pipelines architecture.

I tried to retain the API surface. There are some methods that return generic `CollectedResponse` and implement the `TryInto` for a specific return type. In these cases I implemented the "async try_from" that is generally used on the type alias, as to not have conflicts / ambiguities. An example being `get_identity` used for module and device identity methods.

I also added an authorization policy along with the option to use identities for authorization. I never really worked with iothub, but looking at the docs it seems that is a valid auth method. Currently there is no way to specify client options, albeit them being used for constructing the pipeline. This was deliberate, as I understand a new builder pattern for the clients is being worked on and there were no options before as well.

Lastly I reused the timeout policy from storage and moved it to core. 

@rylev @ctaggart

### DIFF
--- a/sdk/iot_hub/src/service/mod.rs
+++ b/sdk/iot_hub/src/service/mod.rs
@@ -47,10 +47,6 @@ pub enum IoTHubCredentials {
 impl std::fmt::Debug for IoTHubCredentials {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match &self {
-            // IoTHubCredentials::Key(_, _) => f
-            //     .debug_struct("IoTHubCredentials")
-            //     .field("credential", &"Key")
-            //     .finish(),
             IoTHubCredentials::SASToken(_) => f
                 .debug_struct("IoTHubCredentials")
                 .field("credential", &"SASToken")
--- a/sdk/iot_hub/src/service/mod.rs
+++ b/sdk/iot_hub/src/service/mod.rs
@@ -92,9 +92,9 @@ impl ServiceClient {
     /// let iot_hub_name = "cool-iot-hub";
     /// let sas_token = "<a generated sas token>";
     ///
-    /// let iot_hub = ServiceClient::from_sas_token(iot_hub_name, sas_token);
+    /// let iot_hub = ServiceClient::new_sas_token(iot_hub_name, sas_token);
     /// ```
-    pub fn from_sas_token<S, T>(iot_hub_name: S, sas_token: T) -> Self
+    pub fn new_sas_token<S, T>(iot_hub_name: S, sas_token: T) -> Self
     where
         S: Into<String>,
         T: Into<String>,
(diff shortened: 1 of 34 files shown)


sdk/iot_hub/src/service/mod.rs
```diff
@@ -29,6 +33,40 @@ use self::resources::{AuthenticationMechanism, Status};
...
+    TokenCredential(Arc<dyn TokenCredential>),
+}
+
+impl std::fmt::Debug for IoTHubCredentials {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match &self {
+            // IoTHubCredentials::Key(_, _) => f
+            //     .debug_struct("IoTHubCredentials")
+            //     .field("credential", &"Key")
+            //     .finish(),
```
rylev, line 53: Should this be removed?

sdk/iot_hub/src/service/operations/create_or_update_device_identity.rs
```diff
@@ -67,17 +66,23 @@ impl CreateOrUpdateDeviceIdentityBuilder {
...
+
+            CreateOrUpdateDeviceIdentityResponse::try_from(response).await
         })
     }
 }
 
 pub type CreateOrUpdateDeviceIdentityResponse = DeviceIdentityResponse;
 
+impl CreateOrUpdateDeviceIdentityResponse {
+    pub(crate) async fn try_from(response: azure_core::Response) -> azure_core::Result<Self> {
```
rylev, line 79: I think this should probably live with the definition of `DeviceIdentityResponse`

sdk/iot_hub/src/service/operations/create_or_update_module_identity.rs
```diff
@@ -51,17 +50,23 @@ impl CreateOrUpdateModuleIdentityBuilder {
...
+
+            CreateOrUpdateModuleIdentityResponse::try_from(response).await
         })
     }
 }
 
 pub type CreateOrUpdateModuleIdentityResponse = ModuleIdentityResponse;
 
+impl CreateOrUpdateModuleIdentityResponse {
+    pub(crate) async fn try_from(response: azure_core::Response) -> azure_core::Result<Self> {
```
rylev, line 63: Same as above

rylev (approved):
Looks great! I agree that there are still some open questions, but this brings us much closer to where we want to be. I had some small nits - feel free to address them here or merge, and we can do a follow up.

sdk/iot_hub/src/service/mod.rs
roeap: yes :)

sdk/iot_hub/src/service/operations/create_or_update_device_identity.rs
roeap: done.

roeap:
Thanks @rylev, adressed the PR comments, you'll have to do the merging :)

PR #975, opened 2022-08-04
Define an options-based constructor for the Cosmos SDK

yoshuawuyts:
This is an alternative to https://github.com/Azure/azure-sdk-for-rust/pull/964, following the feedback provided in https://github.com/Azure/azure-sdk-for-rust/pull/964#pullrequestreview-1057377295 filed as a draft to use as a comparison for design discussion. Thanks!

## Discussion

Similar to #964 this attempts to keep a fluent builder-style API for options construction. I attempted to implement `Deref/DerefMut` for the `ClientOptions`, but that's incompatible with a builder-style API - the `deref` call will never be able to return the cosmos builder. We could rewrite this using a getter/setter-based approach instead, but that would be a separate PR (which I'm happy to file a separate draft for if people want to see it).

We also cannot directly pass the `azure_core::ClientOptions` to the Cosmos SDK since the cosmos SDK requires custom extensions. We could, of course, create layered options constructors - but I strongly suspect that will not help progress our goal of creating an ergonomic end-user experience.

The other decision I made is to keep the declaration of the `account` and `auth_token` tied to `CloudLocation`. This means that the client constructor is `CosmosClient::with_options(options)` instead of `CosmosClient::with_options(account, auth_token, options)`. We could change this, but the current design correctly surfaces that not all `CloudLocation` variants require an account name + auth token - which removes the need to just make up values to satisfy the interface.

In terms of implementation this appears to be roughly the same amount of code as https://github.com/Azure/azure-sdk-for-rust/pull/964. Which I think means, modulo any further changes, the decision for which API we prefer will mostly come down to which API we like best.

### DIFF
--- a/sdk/data_cosmos/examples/client_builder.rs
+++ b/sdk/data_cosmos/examples/client_builder.rs
@@ -1,0 +1,32 @@
+use azure_core::{RetryOptions, TransportOptions};
+use clap::Parser;
+
+use azure_data_cosmos::prelude::*;
+
+#[derive(Debug, Parser)]
+struct Args {
+    /// Cosmos primary key name
+    #[clap(env = "COSMOS_PRIMARY_KEY")]
+    primary_key: String,
+    /// The cosmos account your're using
+    #[clap(env = "COSMOS_ACCOUNT")]
+    account: String,
+    /// The name of the database
+    database_name: String,
+}
+
+#[tokio::main]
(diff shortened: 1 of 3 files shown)


rylev:
Closing this as we've decided to go with the original proposal https://github.com/Azure/azure-sdk-for-rust/pull/964.

PR #978, opened 2022-08-04
Use specific minimum version of time dependency

bittrance:
Switching to latest main version of the SDK in my latest project produced this compiler error:
```
error[E0599]: no method named `unsigned_abs` found for struct `time::Duration` in the current scope
   --> /home/bittrance/.cargo/git/checkouts/azure-sdk-for-rust-c7e81752dfd35dc6/c8ccf8d/sdk/core/src/date/mod.rs:138:22
    |
138 |     (first - second).unsigned_abs()
    |                      ^^^^^^^^^^^^ method not found in `time::Duration`
```
This is a consequence of the switch to the `time` crate as introduced in https://github.com/Azure/azure-sdk-for-rust/pull/965 which specifies `time` version `0.3`. However, the method was introduced in time 0.3.10, see https://github.com/time-rs/time/commit/c3df3f9f79181604c224bd6ecebbf2c718513a3d . 

I'm not sure exactly what combination of dependencies in my project bounded `time` to version `0.3.9`, but I'm thinking this SDK should declare a viable minimum version?

### DIFF
--- a/sdk/core/Cargo.toml
+++ b/sdk/core/Cargo.toml
@@ -16,7 +16,7 @@ edition = "2021"
 async-trait = "0.1"
 base64 = "0.13"
 bytes = "1.0"
-time = { version = "0.3", features = ["serde-well-known", "macros", "local-offset"] }
+time = { version = "0.3.10", features = ["serde-well-known", "macros", "local-offset"] }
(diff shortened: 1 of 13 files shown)


cataggar (approved):
Thanks!

PR #980, opened 2022-08-06
Rearrange mock transport support

rylev:
This PR moves the mock transport layer out of azure_core into its own dependency. This removes a lot of complexity introduced by previously requiring a feature to enable the mock transport framework. Now this functionality can simply be added as a dev-dependency. 

What's more this now enables arbitrary custom transport policies. If customers want to exactly fine tune how the requests make it to the network, they can now do so. 

Check out the [updated docs](https://github.com/Azure/azure-sdk-for-rust/compare/main...rylev:azure-sdk-for-rust:rearrange-mock?expand=1#diff-e61041fbb04ad6af8b81e52e6b987d9cdeb5564ff89806e6bb9f24c370e1601b) for a peek at how this feature now looks.


cataggar (approved):
I really like seeing the separated out.

PR #982, opened 2022-08-08
add `--package` to `azure-autorust`

ctaggart:
This allows the packages to generate to be limited. This is useful for development. Previously, `ONLY_SERVICES` was being hardcoded.

### DIFF
--- a/services/autorust/azure-autorust/src/main.rs
+++ b/services/autorust/azure-autorust/src/main.rs
@@ -15,13 +15,13 @@ struct Args {
     #[clap(long)]
     publish: bool,
 
-    /// Specify specific package to generate
+    /// Specify specific package to generate. Multiple accepted.
     #[clap(long = "package", short = 'p')]
-    packages: Vec<String>,
+    package: Vec<String>,
 }
 impl Args {
     pub fn packages(&self) -> Vec<&str> {
-        self.packages.iter().map(String::as_str).collect()
+        self.package.iter().map(String::as_str).collect()
     }
 }
 
--- a/services/autorust/azure-autorust/src/main.rs
+++ b/services/autorust/azure-autorust/src/main.rs
@@ -1,9 +1,9 @@
 // cargo run --release -p azure-autorust
+// cargo run --release -p azure-autorust -- -p azure_svc_blobstorage
 
 use autorust_codegen::{
     crates::{list_crate_names, list_dirs},
-    gen::gen_crate,
-    get_mgmt_readmes, get_svc_readmes,
+    gen, get_mgmt_readmes, get_svc_readmes,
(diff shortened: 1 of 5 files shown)


services/autorust/azure-autorust/src/main.rs
```diff
@@ -1,9 +1,9 @@
 // cargo run --release -p azure-autorust
+// cargo run --release -p azure-autorust -- -p azure_svc_blobstorage
```
ctaggart, line 2: Example usage where only `azure_svc_blobstorage` will be generated.

services/autorust/azure-autorust/src/main.rs
```diff
@@ -14,52 +14,63 @@ struct Args {
     /// Generate the publish GitHub workflows
     #[clap(long)]
     publish: bool,
+
+    /// Specify specific package to generate
+    #[clap(long = "package", short = 'p')]
```
ctaggart, line 19: `clap` is pretty nice

services/autorust/azure-autorust/src/main.rs
```diff
@@ -14,52 +14,63 @@ struct Args {
     /// Generate the publish GitHub workflows
     #[clap(long)]
     publish: bool,
+
+    /// Specify specific package to generate
```
rylev, line 18: Nit: we probably should mention that this takes a list of packages.

services/autorust/azure-autorust/src/main.rs
```diff
@@ -14,52 +14,63 @@ struct Args {
...
+    #[clap(long = "package", short = 'p')]
+    packages: Vec<String>,
+}
+impl Args {
```
rylev, line 22:
```suggestion
}

impl Args {
```

services/autorust/azure-autorust/src/main.rs
```diff
@@ -14,52 +14,63 @@ struct Args {
...
+    pub fn packages(&self) -> Vec<&str> {
+        self.packages.iter().map(String::as_str).collect()
```
rylev, line 24: Just a thought: This seems somewhat wasteful - maybe returning `impl Iterator<Item = &str>` would work better?

rylev (approved):
Looks good!

services/autorust/azure-autorust/src/main.rs
cataggar: I added to the help that `Multiple accepted.`
```
USAGE:
    azure-autorust.exe [OPTIONS]

OPTIONS:
    -h, --help                 Print help information
    -p, --package <PACKAGE>    Specify specific package to generate. Multiple accepted
        --publish              Generate the publish GitHub workflows     
```

PR #983, opened 2022-08-08
remove `operations` modules from services

ctaggart:
The `models` are still generated in a `models` submodule, but the operations are now directly in the module. This also removes the "generated by AutoRust" header.


ctaggart:
Replaced by #984 with no conflicts.

PR #986, opened 2022-08-08
Allow for a streaming response

### LINKED ISSUE #981: Expose streaming for operations with potentially large data
When reading from blob, response includes the full data as bytes, instead of stream of bytes. It can easily lead to spike in RAM usage for reading large blobs. Reqwest and hyper allows us to stream bytes from the response. 

Is there any reason for this? or is this already in road map?

rylev:
Fixes #981 

This allows operations to expose a streaming response body so that they don't have to collect response bodies into a `Bytes` collection. For example, `get_blob` now returns back a `ResponseBody` as part of its response type. This type implements `Stream` for users who want to stream the data as well as `collect` methods for users that just want the data as `Bytes` or even as a `String`. 

This also makes `PinnedStream` and `collect_pinned_stream` private which were always a bit of an awkward API.  

This is what fetching a blob now looks like:

```rust
let mut result: Vec<u8> = vec![];
let mut stream = blob_client
     .get()
     .chunk_size(0xFFFF_FFFF_FFFFu64)
     .into_stream();
while let Some(response) = stream.next().await {
     let mut body = response?.data;
     while let Some(value) = body.next().await {
         // Instead of this while loop the user can also call `body.collect()` to get back `Bytes`
         // or `body.collect_string()` to get a `String`
         let value = value?;
         result.extend(&value);
     }
}
```


cataggar:
Sorry, #984 caused some conflicts. Probably best to revert 96c66c166f4d3af976f6171e9cd8646283a7b66e, merge main, and rerun codegen.

cataggar (approved):
This looks great!

sdk/core/src/response.rs
```diff
@@ -108,15 +88,66 @@ impl CollectedResponse {
...
+pub struct ResponseBody(#[pin] PinnedStream);
+
+impl ResponseBody {
+    fn new(stream: PinnedStream) -> Self {
+        Self(stream)
+    }
+
+    /// Collect the stream into a `Bytes` collection
+    pub async fn collect(self) -> crate::Result<Bytes> {
+        collect_pinned_stream(self.0).await
```
yoshuawuyts, line 112: Can we inline `collect_pinned_stream` here? I don't believe it's used anywhere else.

yoshuawuyts (approved):
This looks really good!

rajasekarv:
Didn't expect it to be done this quick. Thanks a lot @rylev

PR #987, opened 2022-08-08
WIP: Start creating a builder for the StorageClient.

gorzell:
This is a follow on to the work done in #964 that implements a similar pattern for the `StorageClient`. One of the main benefits in addition to consistency is that it allows for configuring more of the client options like `retries`, etc.

### DIFF
--- a/sdk/storage/src/core/clients/storage_client.rs
+++ b/sdk/storage/src/core/clients/storage_client.rs
@@ -67,6 +67,114 @@ pub enum ServiceType {
 }
 
 #[derive(Clone, Debug)]
+pub struct StorageClientBuilder {
+    storage_credentials: Option<StorageCredentials>,
+    blob_storage_url: Url,
+    table_storage_url: Url,
+    queue_storage_url: Url,
+    queue_storage_secondary_url: Url,
+    filesystem_url: Url,
+    account: String,
+    storage_options: StorageOptions,
+}
+
+impl StorageClientBuilder {
+    #[must_use]
+    pub fn new(account: impl Into<String>) -> Self {
+        let account = account.into();
+        Self {
+            storage_credentials: None,
+            blob_storage_url: get_endpoint_uri(None, &account, "blob").unwrap(),
+            table_storage_url: get_endpoint_uri(None, &account, "table").unwrap(),
+            queue_storage_url: get_endpoint_uri(None, &account, "queue").unwrap(),
+            queue_storage_secondary_url: get_endpoint_uri(
+                None,
+                &format!("{account}-secondary"),
+                "queue",
+            )
+            .unwrap(),
+            filesystem_url: get_endpoint_uri(None, &account, "dfs").unwrap(),
+            account,
+            storage_options: StorageOptions::new(),
+        }
+    }
+
+    #[must_use]
+    pub fn new_emulator() -> Self {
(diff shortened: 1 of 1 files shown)


sdk/storage/src/core/clients/storage_client.rs
```diff
@@ -66,6 +66,114 @@ pub enum ServiceType {
...
+
+    #[must_use]
+    pub fn access_key(mut self, key: impl Into<String>) -> Self {
+        match self.storage_credentials {
+            None => {
+                self.storage_credentials =
+                    Some(StorageCredentials::Key(self.account.clone(), key.into()));
+                self
+            }
+            // FIXME: This gets a bit tricky in the emulator case...
```
rylev, line 133: The way we're handling this in Cosmos is by not having an `access_key` (or in the case of Cosmos an `AuthorizationToken`) setter. As you rightfully pointed out here, the need for authentication comes down to whether the user is talking to an emulator or not. 

This is why this is [all encapsulated in the `CloudLocation`](https://github.com/Azure/azure-sdk-for-rust/blob/bc1d085a8d6ba864a7acc0cccbad78efef348561/sdk/data_cosmos/src/clients/cosmos.rs#L179-L197). I think we'll want a similar construct here as well.

sdk/storage/src/core/clients/storage_client.rs
```diff
@@ -66,6 +66,114 @@ pub enum ServiceType {
...
+            Some(c) => panic!(
+                "credentials already set to: {:?}, you should only set them once.",
+                c
+            ),
```
rylev, line 137: Is there a need for panicking here? Why not let the user override credentials? We allow them to override other items that might already be set.

sdk/storage/src/core/clients/storage_client.rs
```diff
@@ -66,6 +66,114 @@ pub enum ServiceType {
...
+    storage_options: StorageOptions,
+}
+
+impl StorageClientBuilder {
+    #[must_use]
+    pub fn new(account: impl Into<String>) -> Self {
+        let account = account.into();
+        Self {
+            storage_credentials: None,
+            blob_storage_url: get_endpoint_uri(None, &account, "blob").unwrap(),
```
rylev, line 87: I think we'll likely want to rework how we handle urls completely. In Cosmos, we ask the [`CloudLocation` for the right url]((https://github.com/Azure/azure-sdk-for-rust/blob/bc1d085a8d6ba864a7acc0cccbad78efef348561/sdk/data_cosmos/src/clients/cosmos.rs#L132) (more on `CloudLocation` below). 

This should also work here, but we'll just need to include one additional piece of information: the resource type. We can encapsulate that in an enum which gets passed to the `url` method which can then construct the right url based on cloud location and the resource type. 

I think for now, we can just construct the url as a new `String` every time. Once we get an API we're happy with we can add back some sort of memoization should we need to.

rylev (changes requested):
This is a good start, but I think unfortunately, we'll need a more fundamental refactoring here to get to where we want to go.

cataggar:
@gorzell, we are preparing a release very soon #994. In case you wish to squeeze this in, let me know.

rylev:
This has been superseded by #1057 and subsequent PRs.

PR #988, opened 2022-08-08
add retry and transport options for services

ctaggart:
This follows #964 for how to set the retry and transport options in the services. It also follows #964 by adding a `Client::builder`.

### DIFF
--- a/services/autorust/codegen/src/codegen_operations.rs
+++ b/services/autorust/codegen/src/codegen_operations.rs
@@ -83,33 +83,56 @@ pub fn create_client(modules: &[String], endpoint: Option<&str>) -> Result<Token
             credential: std::sync::Arc<dyn azure_core::auth::TokenCredential>,
             endpoint: Option<String>,
             scopes: Option<Vec<String>>,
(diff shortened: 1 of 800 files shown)


services/autorust/codegen/src/codegen_operations.rs
```diff
@@ -83,33 +83,56 @@ pub fn create_client(modules: &[String], endpoint: Option<&str>) -> Result<Token
...
             }
 
+            #[doc = "Set the endpoint."]
+            #[must_use]
             pub fn endpoint(mut self, endpoint: impl Into<String>) -> Self {
                 self.endpoint = Some(endpoint.into());
                 self
             }
-
+            #[doc = "Set the scopes."]
```
rylev, line 109:
```suggestion

            #[doc = "Set the scopes."]
```

rylev: Also, it took me a second to figure out what "scopes" refers to. Can we make the docs clearer here?

rylev (approved):

PR #992, opened 2022-08-09
Fix issue with user defined functions

rylev:
Creating user defined functions was broken - this fixes that and makes a test for user defined functions using the mock framework.

The fix itself was twofold: 
* Using the right format for the date header. I actually think all of the cosmos crate is broken since we don't test dates in local tests since they change on every request. I believe this was an issue introduced by #965. 
* The resource type (which is used to sign the request) was not correct for user defined functions. There was one additional place where the resource type was incorrect - which I fixed. We're missing test coverage for stored procedures unfortunately.

Additionally, I made the following changes:
* We currently only retry HTTP errors that are not on the disallow list *and* all other errors. This makes it such that only IO errors get retried. Generally speaking, any other type of error is something wrong that a retry won't fix, and it should be bubbled up to the user ASAP. 
* There were a few general headers that were duplicated across azure_core and azure_data_cosmos. I deleted the ones in cosmos and just used the ones from azure_core.


sdk/data_cosmos/src/authorization_policy.rs
```diff
@@ -75,8 +74,8 @@ impl Policy for AuthorizationPolicy {
             &auth
         );
 
-        request.insert_header(HEADER_DATE, HeaderValue::from(time_nonce.to_string()));
-        request.insert_header(HEADER_VERSION, HeaderValue::from_static(AZURE_VERSION));
+        request.insert_header(MS_DATE, HeaderValue::from(date::to_rfc1123(&time_nonce)));
```
cataggar, line 77: Yes, looks like this was missed in #965.

cataggar (approved):

PR #995, opened 2022-08-10
allow wasm32 for `azure_identity`

cataggar:
This makes all of these checks to work:

- cargo check --target=wasm32-unknown-unknown --no-default-features
- cargo check --tests --features test_e2e --manifest-path sdk/storage_blobs/Cargo.toml
- cargo check --tests --features test_e2e --manifest-path sdk/storage_datalake/Cargo.toml

The storage_blobs & storage_datalake dependency changes were included in yesterday's release.

### DIFF
--- a/sdk/storage_blobs/Cargo.toml
+++ b/sdk/storage_blobs/Cargo.toml
@@ -38,7 +38,7 @@ oauth2 = { version = "4.0.0", default-features = false }
 mock_transport = { path = "../../eng/test/mock_transport" }
 
 [features]
-default = ["enable_reqwest", "azure_identity/default"]
+default = ["enable_reqwest"]
 test_e2e = []
 azurite_workaround = []
 enable_reqwest = ["azure_core/enable_reqwest", "azure_storage/enable_reqwest"]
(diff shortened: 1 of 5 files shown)


sdk/storage_blobs/Cargo.toml
```diff
@@ -38,7 +38,7 @@ oauth2 = { version = "4.0.0", default-features = false }
 mock_transport = { path = "../../eng/test/mock_transport" }
 
 [features]
-default = ["enable_reqwest", "azure_identity/default"]
```
cataggar, line 41: `azure_identity` is a dev-dependency, so this doesn't work. It fails when publishing.

sdk/storage_datalake/Cargo.toml
```diff
@@ -15,7 +15,6 @@ edition = "2021"
 [dependencies]
 async-trait = "0.1"
 azure_core = { path = "../core", version = "0.4" }
-azure_identity = { path = "../identity", default_features = false }
```
cataggar, line 18: azure_identity should be a dev-dependency.

rylev (changes requested):
Can we just land the dependency changes first? I'd rather not add all of these feature flags (which are really hard to understand), and just focus on fixing #996.

cataggar:
I'd prefer to merge this and then back out the changes once #996 is fixed.

rylev (approved):

PR #997, opened 2022-08-10
implement oauth2 HTTP client

### LINKED ISSUE #996: implement oauth2 HTTP client
#995 disables `ClientSecretCredentials` and other features in `identity` because of the `reqwest` dependency.  oauth2 allows swapping out the HTTP Client https://docs.rs/oauth2/latest/oauth2/#importing-oauth2-selecting-an-http-client-interface . Let's see if we can implement it with `azure_core::http_client::HttpClient`.  An alternative would be to fork the oauth2 code into `azure_identity`.

cataggar:
Fix #996. Two`azure_identity::token_credentials` now require an `Arc<dyn HttpClient>`
- `ClientSecretCredential`
- `EnvironmentCredential`

The `HttpClient` is used to implement the oauth2 http client.

Tested with:
- cargo run -p azure_security_keyvault --example get_secret_versions
- cargo run -p azure_security_keyvault --example list_secrets


sdk/identity/src/token_credentials/client_secret_credentials.rs
```diff
@@ -135,7 +135,7 @@ impl TokenCredential for ClientSecretCredential {
         let token_result = client
             .exchange_client_credentials()
             .add_scope(Scope::new(format!("{}/.default", resource)))
-            .request_async(async_http_client)
+            .request_async(crate::oauth2_async_http_client)
```
rylev, line 138: If would be nice if all functions in this crate had access to a `&dyn azure_core::HttpClient` this would make this crate completely unaware of which implementation it's using.

cataggar: Done. Header conversion and testing is left if you with to pick it up.

sdk/core/src/request.rs
```diff
@@ -99,6 +99,10 @@ impl Request {
         &self.body
     }
 
+    pub fn set_headers(&mut self, headers: impl Into<Headers>) {
+        self.headers = headers.into();
+    }
+
```
rylev, line 105: I'm somewhat worried that this just runs the risk of folks overriding any headers that have already been set. Perhaps an `append_headers` would be better? I wish this just worked with the existing `insert_headers`.

sdk/identity/src/oauth2_http_client.rs
```diff
@@ -0,0 +1,110 @@
...
+        Ok(oauth2::HttpResponse {
+            status_code,
+            headers,
+            body,
+        })
+    }
+}
+
+fn try_from_method(method: &oauth2::http::Method) -> azure_core::Result<azure_core::Method> {
+    if method == oauth2::http::Method::GET {
```
rylev, line 41: Any reason to not use a `match`?

rylev (dismissed):
Looks good, just have some code quality comments.

sdk/identity/src/oauth2_http_client.rs
```diff
@@ -0,0 +1,110 @@
...
+
+pub(crate) struct Oauth2HttpClient {
+    http_client: Arc<dyn HttpClient>,
+}
+
+impl Oauth2HttpClient {
+    /// Create a new Oauth2HttpClient
+    pub fn new(http_client: Arc<dyn HttpClient>) -> Self {
+        Self { http_client }
+    }
```
bmc-msft, line 18: I thought we were moving away from new_http_client and moving to `Pipeline`?

rylev: I guess the question is whether identity should also go through a `Pipeline`. It makes sense in a way. The challenge is that there is specific logic the oauth client handles for us that a simple transport policy would not do.

sdk/core/src/request.rs
cataggar: I updated it to use `insert_headers`.

sdk/identity/src/oauth2_http_client.rs
cataggar: Updated to use `match`.

sdk/identity/src/oauth2_http_client.rs
cataggar: Created #1015 for this.

sdk/identity/src/oauth2_http_client.rs
```diff
@@ -0,0 +1,103 @@
...
+        let body = response.into_body().collect().await?.to_vec();
+        Ok(oauth2::HttpResponse {
+            status_code,
+            headers,
+            body,
+        })
+    }
+}
+
+fn try_from_method(method: &oauth2::http::Method) -> azure_core::Result<azure_core::Method> {
```
bmc-msft, line 42:
```suggestion
fn try_from_method(method: oauth2::http::Method) -> azure_core::Result<azure_core::Method> {
```

sdk/identity/src/oauth2_http_client.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// Create a new Oauth2HttpClient
+    pub fn new(http_client: Arc<dyn HttpClient>) -> Self {
+        Self { http_client }
+    }
+
+    pub(crate) async fn request(
+        &self,
+        oauth2_request: oauth2::HttpRequest,
+    ) -> Result<oauth2::HttpResponse, azure_core::error::Error> {
+        let method = try_from_method(&oauth2_request.method)?;
```
bmc-msft, line 24:
```suggestion
        let method = try_from_method(oauth2_request.method)?;
```

sdk/identity/src/oauth2_http_client.rs
cataggar: Why?

rylev: If `Method` is `Copy` then this is usually more efficient (not that performance matters at all here). Clippy has a warning for this.

cataggar: There is no clippy warning. `oauth2::http::Method` is not `Copy`:

``` rust
#[derive(Clone, PartialEq, Eq, Hash)]
pub struct Method(Inner);
```

rylev (approved):
I was a little hesitant to merge this since I think things could change considerably if/when we move to using a pipeline, but this *is* an improvement over the status quo, so I think we should land it.

PR #1000, opened 2022-08-11
Simplify encoded value generation fixing lifetime issue

satlank:
Shadowing `encoded` makes the async block not send which prevents
sharing it across threads.  Just do the generation of the encoded values
through temporary variables instead of shadowing.

The added code in the test module will fail to compile without the
accompanying code changes.

### DIFF
--- a/sdk/identity/src/device_code_flow/mod.rs
+++ b/sdk/identity/src/device_code_flow/mod.rs
@@ -187,32 +187,16 @@ async fn post_form(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use async_trait::async_trait;
-    use std::error::Error;
 
-    #[async_trait]
-    trait LoginHelper {
-        async fn get_secret(&self) -> Result<String, Box<dyn Error>>;
-    }
+    fn require_send<T: Send>(_t: T) {}
 
-    struct DeviceCodeLoginHelper {
-        client_id: ClientId,
-        tenant_id: String,
-        scopes: Vec<String>,
-    }
-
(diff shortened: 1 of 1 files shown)


sdk/identity/src/device_code_flow/mod.rs
```diff
@@ -184,3 +183,36 @@ async fn post_form(
...
     http_client.execute_request(&req).await
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use async_trait::async_trait;
+    use std::error::Error;
+
+    #[async_trait]
```
rylev, line 193: Is all of this meant to test that the future from `start` is `Send`? I'd rather do that by using a function that constrains some argument [like in this playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c4f87e3bd5d2b746272feffa5222b0a0).

rylev (changes requested):
Great catch! This code is also more readable too. 

I'd like the test to be simpler though (and I offer a way that could look), and we should also make sure the test is clear what is being tested.

sdk/identity/src/device_code_flow/mod.rs
satlank: Thanks for looking at this so speedily. Yes, that was a distilled down version of what caused the compiler to shout at me. Great suggestion to use a function with constrained argument, that makes it much more understandable what this is ensuring. Pushed a new commit with the change.

rylev (approved):
🎉

PR #1004, opened 2022-08-12
Ensure that `exchange` is Send

satlank:
Shadowing the `encoded` Serializer makes it not send if the lifetime is
not restricted to end before the first `await`.

### DIFF
--- a/sdk/identity/src/refresh_token.rs
+++ b/sdk/identity/src/refresh_token.rs
@@ -23,17 +23,19 @@ pub async fn exchange(
     client_secret: Option<&ClientSecret>,
     refresh_token: &AccessToken,
 ) -> azure_core::Result<RefreshTokenResponse> {
-    let mut encoded = form_urlencoded::Serializer::new(String::new());
-    let encoded = encoded.append_pair("grant_type", "refresh_token");
-    let encoded = encoded.append_pair("client_id", client_id.as_str());
-    // optionally add the client secret
-    let encoded = if let Some(client_secret) = client_secret {
-        encoded.append_pair("client_secret", client_secret.secret())
-    } else {
-        encoded
+    let encoded = {
+        let mut encoded = form_urlencoded::Serializer::new(String::new());
+        let encoded = encoded.append_pair("grant_type", "refresh_token");
+        let encoded = encoded.append_pair("client_id", client_id.as_str());
+        // optionally add the client secret
+        let encoded = if let Some(client_secret) = client_secret {
+            encoded.append_pair("client_secret", client_secret.secret())
+        } else {
+            encoded
+        };
+        let encoded = encoded.append_pair("refresh_token", refresh_token.secret());
+        encoded.finish()
     };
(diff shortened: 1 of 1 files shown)


sdk/identity/src/refresh_token.rs
```diff
@@ -23,17 +23,19 @@ pub async fn exchange(
...
+        let encoded = encoded.append_pair("grant_type", "refresh_token");
+        let encoded = encoded.append_pair("client_id", client_id.as_str());
+        // optionally add the client secret
+        let encoded = if let Some(client_secret) = client_secret {
+            encoded.append_pair("client_secret", client_secret.secret())
+        } else {
+            encoded
+        };
+        let encoded = encoded.append_pair("refresh_token", refresh_token.secret());
+        encoded.finish()
```
rylev, line 37:
```suggestion
        let mut encoded = form_urlencoded::Serializer::new(String::new())
            .append_pair("grant_type", "refresh_token");
            .append_pair("client_id", client_id.as_str());
        // optionally add the client secret
        if let Some(client_secret) = client_secret {
            encoded = encoded.append_pair("client_secret", client_secret.secret());
        } 
        encoded.append_pair("refresh_token", refresh_token.secret()).finish()
```

rylev (changes requested):
Looks good! I think we can clean up the code a bit more though.

sdk/identity/src/refresh_token.rs
satlank: I tried that first :) But unfortunately the temporary Serialiser is dropped at the end of the first batch of appends. Couldn't find a neat way to get the conditional append_pair in there. Unless I am missing something obvious?

satlank: Well, can do:
```
         let mut encoded = form_urlencoded::Serializer::new(String::new());
         let mut encoded = encoded
             .append_pair("grant_type", "refresh_token")
             .append_pair("client_id", client_id.as_str());
         // optionally add the client secret
         if let Some(client_secret) = client_secret {
             encoded = encoded.append_pair("client_secret", client_secret.secret())
         };
         encoded.append_pair("refresh_token", refresh_token.secret())
             .finish()
```

That is maybe a reasonable clean-up?

rylev: I would probably write it like this:

```rust
let encoded = {
        let mut encoded = &mut form_urlencoded::Serializer::new(String::new());
        encoded = encoded
            .append_pair("grant_type", "refresh_token")
            .append_pair("client_id", client_id.as_str())
            .append_pair("refresh_token", refresh_token.secret());
        // optionally add the client secret
        if let Some(client_secret) = client_secret {
            encoded = encoded.append_pair("client_secret", client_secret.secret())
        }
        encoded.finish()
};
```

rylev: Sorry - posted mine before I saw yours. Either one would work 😊

satlank: Went with yours, looks a bit cleaner :)

rylev (approved):

PR #1005, opened 2022-08-12
Make `url` on BlobClient pub.

gorzell:
Making this pub so that it can be used with the `copy` function which
requires a URL as an argument. Ideally this probably shouldn't be
fallible given that if you have a BlobClient it should have a valid URL,
but that can be changed later.


rylev (approved):
Looks good!

PR #1006, opened 2022-08-12
Add a rust-toolchain file.

gorzell:
I have found this to be really helpful with my other projects to ensure
that people are developing on the same version, etc. Happy to change the
settings to whatever makes sense based on feedback. `channel` is
probably the only critical thing.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,3 @@
 [toolchain]
 channel = "1.62.0"
-components = [ "rustfmt", "clippy", "llvm-tools-preview" ]
-targets = [ "x86_64-unknown-linux-gnu", "x86_64-apple-darwin" ]
-profile = "default"
+components = [ "rustfmt", "clippy" ]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,5 @@
+[toolchain]
+channel = "1.62.0"
+components = [ "rustfmt", "clippy", "llvm-tools-preview" ]
+targets = [ "x86_64-unknown-linux-gnu", "x86_64-apple-darwin" ]
+profile = "default"
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.62.0"
+channel = "1.63.0"
 components = [ "rustfmt", "clippy" ]


rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "1.62.0"
+components = [ "rustfmt", "clippy", "llvm-tools-preview" ]
+targets = [ "x86_64-unknown-linux-gnu", "x86_64-apple-darwin" ]
```
rylev, line 4: I don't think we need this - folks will have the target that is appropriate for their machine, and we don't want to require the ability to cross compile.

rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
...
+profile = "default"
```
rylev, line 5: I believe this can be removed as `default` is the default

rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "1.62.0"
+components = [ "rustfmt", "clippy", "llvm-tools-preview" ]
```
rylev, line 3: Is "llvm-tools-preview" necessary? I don't believe it is

rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "1.62.0"
```
rylev, line 2: We should require 1.63 which is the latest stable

rylev (changes requested):
I'm not 100% convinced we need this, but we can at least give it a try. The file should be simplified though.

rust-toolchain.toml
gorzell: It's fine with me, since it just came out I didn't want to assume everyone is ready to update.

rust-toolchain.toml
gorzell: I took this from another project as an example, happy to slim this all down.

gorzell:
> I'm not 100% convinced we need this, but we can at least give it a try. The file should be simplified though.

I find it nice for keeping everyone on the same page, especially when it comes to updates and the current rates at which things like clippys are added. That way you don't suddenly have a bunch of lint errors that don't exist in CI just because you happen to have updated your stable  locally, etc.

You also then know exactly what Rust version a release was built and tested with, etc.

rylev (approved):

PR #1008, opened 2022-08-15
blob_storage: Make `Snapshot` type consistent.

gorzell:
Previously many things treated `Snapshot` as an opaque type defined by
the `request_query` macro, but the `Blob` type was deserializing it as a
`DateTime` and not reading it from the header at all. Now everything is
using one shared opaque type.

I chose to redefine `Snapshot` without using the macro so that it could
derive Serialize/Deserialize. This is required because in the cases
`Snapshot` is part of a serialized response body. It is otherwise compatible with the
previous definition. The alternative would have been to change the macro
to do derive Serialize/Deserialize, but since that would be a change to
`core` and a number of other types, it felt heavy handed.

### DIFF
--- a/sdk/storage_blobs/src/blob/mod.rs
+++ b/sdk/storage_blobs/src/blob/mod.rs
@@ -22,10 +22,12 @@ use azure_core::{
     content_type, date,
     headers::{self, Headers},
     parsing::from_azure_time,
-    Etag, LeaseDuration, LeaseState, LeaseStatus,
+    AppendToUrlQuery, Etag, LeaseDuration, LeaseState, LeaseStatus,
 };
 use azure_storage::{ConsistencyCRC64, ConsistencyMD5, CopyId, CopyProgress};
+use std::borrow::Cow;
 use std::collections::HashMap;
+use std::str::FromStr;
 use time::OffsetDateTime;
 
 #[cfg(feature = "azurite_workaround")]
@@ -59,7 +61,10 @@ create_enum!(RehydratePriority, (High, "High"), (Standard, "Standard"));
 
 create_enum!(PageWriteType, (Update, "update"), (Clear, "clear"));
 
+use azure_core::error::Error;
+use azure_core::headers::HeaderName;
 use serde::{self, Deserialize, Deserializer};
+
 fn deserialize_crc64_optional<'de, D>(deserializer: D) -> Result<Option<ConsistencyCRC64>, D::Error>
 where
     D: Deserializer<'de>,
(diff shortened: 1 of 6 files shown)


sdk/storage_blobs/src/blob/mod.rs
```diff
@@ -59,7 +61,10 @@ create_enum!(RehydratePriority, (High, "High"), (Standard, "Standard"));
 
 create_enum!(PageWriteType, (Update, "update"), (Clear, "clear"));
 
+use azure_core::error::Error;
+use azure_core::headers::HeaderName;
```
bmc-msft, line 65: Can these use statements be added in the existing azure_core entry above?

gorzell:
> `request_query` already supports adding arbitrary attributes to the type. So the following works already in the main branch:

Nice, does it still make sense to move it up in the mod tree though?

rylev:
I think it makes sense to leave it where it was with the rest of the request options, but perhaps I'm thinking about things wrong...

sdk/storage_blobs/src/options/mod.rs
```diff
@@ -48,6 +53,17 @@ request_query!(
     /// This type could also be a DateTime but the docs clearly states to treat is as opaque so we do not convert it in any way.
     ///
     /// See: <https://docs.microsoft.com/rest/api/storageservices/get-blob>"]
+    #[derive(PartialEq, Eq, Serialize, Deserialize)]
     Snapshot,
     "snapshot"
 );
+
+impl FromStr for Snapshot {
```
rylev, line 61: What is this needed for? `FromStr` is most oftened used when something is parsed, but we're not really parsing here - we accept any and all strings...

rylev (changes requested):
Looks good, but I do have one open question...

sdk/storage_blobs/src/options/mod.rs
gorzell: This is due to the type signature on `get_optional_as`: https://github.com/Azure/azure-sdk-for-rust/blob/27e1dd45936e77c87cc7bb96bf7b4a7b994a77a7/sdk/core/src/headers/mod.rs#L87-L93 

which is used here: https://github.com/Azure/azure-sdk-for-rust/pull/1008/files#diff-bdeb8cd7bab3372df4aa2feb49411e6cbaf978c135e2dc43330aff4ee30ed32fR233. There are other functions that could be used like `get_optional_str` but then you would have to convert that to the snapshot type via `into`. I am not sure which is better.

rylev (approved):

PR #1011, opened 2022-08-15
set api-version for keyvault list_secrets

ctaggart:
This is a bug fix. Tested with `cargo run -p azure_security_keyvault --example list_secrets`.

### DIFF
--- a/sdk/security_keyvault/src/account/operations/list_secrets.rs
+++ b/sdk/security_keyvault/src/account/operations/list_secrets.rs
@@ -14,6 +14,7 @@ impl ListSecretsBuilder {
 
             let mut uri = self.client.client.vault_url.clone();
             uri.set_path("secrets");
+            uri.set_query(Some(API_VERSION_PARAM));
 
             loop {
                 let resp_body = self


sdk/security_keyvault/src/account/operations/list_secrets.rs
```diff
@@ -14,6 +14,7 @@ impl ListSecretsBuilder {
 
             let mut uri = self.client.client.vault_url.clone();
             uri.set_path("secrets");
+            uri.set_query(Some(API_VERSION_PARAM));
```
rylev, line 17: Is this the only operation where this is necessary?

rylev (approved):

rylev:
It would be great if we could get some tests against the mock framework in for keyvault...

sdk/security_keyvault/src/account/operations/list_secrets.rs
bmc-msft: It is not.  This should be part of the PR I have open for keyvault.

rylev:
Going to close this in favor of https://github.com/Azure/azure-sdk-for-rust/pull/962

PR #1012, opened 2022-08-16
add `AzureCliCredential::new()`

ctaggart:
This is more consistent with other credentials like `DefaultAzureCredential::new()`.

### DIFF
--- a/sdk/identity/src/token_credentials/azure_cli_credentials.rs
+++ b/sdk/identity/src/token_credentials/azure_cli_credentials.rs
@@ -47,9 +47,15 @@ struct CliTokenResponse {
 }
 
 /// Enables authentication to Azure Active Directory using Azure CLI to obtain an access token.
+#[derive(Default)]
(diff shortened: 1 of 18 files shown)


gorzell (approved):
Nice cleanup.

bmc-msft (approved):

sdk/identity/src/token_credentials/azure_cli_credentials.rs
```diff
@@ -47,9 +47,15 @@ struct CliTokenResponse {
 }
 
 /// Enables authentication to Azure Active Directory using Azure CLI to obtain an access token.
+#[derive(Default)]
 pub struct AzureCliCredential;
```
rylev, line 51: FYI: users can still use the `AzureCliCredential` syntax for instantiating this. This means that adding any fields is a breaking change. We may want to consider adding a private field just to prevent this.

```rust
struct AzureCliCredential {
    _private: ()
}
````

PR #1018, opened 2022-08-16
Fixing build break in using client_certificate credentials

msabansal:
The identity SDK doesn't build anymore when trying to build with feature client_certificate. This PR addresses the issue.

Also added a full build on the SDK

### DIFF
--- a/sdk/identity/src/token_credentials/client_certificate_credentials.rs
+++ b/sdk/identity/src/token_credentials/client_certificate_credentials.rs
@@ -3,7 +3,7 @@ use azure_core::{
     auth::{AccessToken, TokenCredential, TokenResponse},
     content_type,
     error::{Error, ErrorKind},
-    headers, new_http_client, Method, Request, StatusCode,
+    headers, new_http_client, Method, Request,
 };
 use base64::{CharacterSet, Config};
 use openssl::{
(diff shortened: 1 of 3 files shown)


rylev:
@msabansal I'm not sure exactly what this fix does. Can you please include a detailed description of the bug being fixed?

sdk/identity/src/token_credentials/client_certificate_credentials.rs
```diff
@@ -130,16 +135,17 @@ struct AadTokenResponse {
     access_token: String,
 }
 
-fn get_encoded_cert(cert: &X509) -> Result<String, ClientCertificateCredentialError> {
+fn get_encoded_cert(cert: &X509) -> Result<String, azure_core::error::Error> {
```
bmc-msft, line 138:
```suggestion
fn get_encoded_cert(cert: &X509) -> azure_core::Result<String> {
```

msabansal:
@rylev  Added the description. I am not sure why the build is passing. 

If someone can point me to how i can fix the test for this i am happy to do that

sdk/identity/src/token_credentials/client_certificate_credentials.rs
```diff
@@ -130,16 +135,17 @@ struct AadTokenResponse {
...
-            cert.to_pem()
-                .map_err(ClientCertificateCredentialError::OpensslError)?
-        )
+        base64::encode(cert.to_pem().map_err(|_| openssl_error())?)
     ))
 }
 
+fn openssl_error() -> azure_core::error::Error {
+    Error::message(ErrorKind::Credential, "Openssl decode error")
+}
```
rylev, line 147: I don't think we want to use the same error message for all of these errors. It makes it hard to tell what actually went wrong.

rylev (approved):
I'm going to merge this since things are broken on master, but if you have time to do a follow up addressing the comments here, that would be amazing.

PR #1023, opened 2022-08-17
Make content_location optional

### LINKED ISSUE #1022: Do not fail for missing content-location header for CosmosDB APIs
It looks like the RUST SDK expects the content-location header to be present in the responses from Azure CosmosDB (documentclient APIs). SDK returns an error if this header is not present. Please note that content-location is not listed in the Common Azure CosmosDB Response headers  [Common response headers - Azure Cosmos DB REST API | Microsoft Docs](https://docs.microsoft.com/en-us/rest/api/cosmos-db/common-cosmosdb-rest-response-headers). Hence taking dependency on this header is not a supported scenario. 
Azure CosmosDB service is not returning this header in the newer versions of the gateway. Hence please fix the SDK to remove dependency on this header.

```rust
let resp = match document_client
            .get_document()
            .execute::<D>()
            .await
            .map_err(into_cosmos_error("Could not get document")) {
                Ok(response) => response,
                Err(err) => {
                    println!("XRAY");
                    return Err(err);
                }
            };
```

Error: "content-location" is not found in headers.

rylev:
Fixes #1022 

This is likely not the only header that is not guaranteed to be there. We should make all headers that are not part of the [common response headers](https://docs.microsoft.com/en-us/rest/api/cosmos-db/common-cosmosdb-rest-response-headers) list optional.

### DIFF
--- a/sdk/data_cosmos/src/headers/from_headers.rs
+++ b/sdk/data_cosmos/src/headers/from_headers.rs
@@ -152,8 +152,10 @@ pub(crate) fn service_version_from_headers(headers: &Headers) -> azure_core::Res
     headers.get_as(&HEADER_SERVICE_VERSION)
 }
 
-pub(crate) fn content_location_from_headers(headers: &Headers) -> azure_core::Result<String> {
-    headers.get_as(&headers::CONTENT_LOCATION)
+pub(crate) fn content_location_from_headers(
+    headers: &Headers,
+) -> azure_core::Result<Option<String>> {
+    headers.get_optional_as(&headers::CONTENT_LOCATION)
 }
 
 pub(crate) fn gateway_version_from_headers(headers: &Headers) -> azure_core::Result<String> {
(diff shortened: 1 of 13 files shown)


ayngling:
We are using this SDK in production (yes, we know it is unofficial, but Rust is awesome ;) ) for many customers (hundred of thousands of users in total), and was affected by the cosmos update. Microsoft helped us by rolling back our cosmos accounts to get it working again, and hence issue #1022 was created.

The long term solution is migrating to a future version of the SDK where this is not an issue, and then upgrade the accounts again. we would appreciate very much if this commit could be merged quickly!

PS: If you arrive at this issue because you are having the same issue with an error about the "content-location" header, the temporary fix is asking Azure support to roll back your cosmos account from the "Compute Gateway for SQL API" to the "legacy gateway for SQL API".

cataggar:
Hi @ayngling, thank you for this info. Are you able to use a main commit or do you need a published crate?

cataggar (approved):

ayngling:
> 

I would love a published crate, but in a pinch I can use a commit :)

ayngling:
@cataggar Actually, if it is possible, a published crate would be much appreciated!

PR #1024, opened 2022-08-18
Move most headers to a common headers struct

rylev:
This change would move most headers that are received back from the Cosmos API to a `CommonHeaders` struct that can be deserialized. The thought behind this is that even for endpoints where it seems like certain headers always come back #1022 showed that this can change. We should probably be pretty lax about what headers we do accept. This allows us to centralize this decision making within Cosmos. 

Thoughts? @ctaggart @bmc-msft

### DIFF
--- a/sdk/data_cosmos/src/headers/common_headers.rs
+++ b/sdk/data_cosmos/src/headers/common_headers.rs
@@ -1,0 +1,49 @@
+use azure_core::{
+    headers::{self, Headers},
+    SessionToken,
+};
+use time::OffsetDateTime;
+
+use crate::ResourceQuota;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct CommonHeaders {
+    pub content_type: String,
+    pub date: OffsetDateTime,
+    pub etag: Option<String>, // todo: make azure_core::Etag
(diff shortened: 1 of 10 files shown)


bmc-msft:
I like the unification for parsing into a central area when possible, though I wonder if it might be more usable if the headers were still available at the top level, rather than having to go through obj.common.  Instead of a nested object, perhaps this could be written as a macro that adds the headers to the struct and another that adds the parsing?

sdk/data_cosmos/src/headers/common_headers.rs
```diff
@@ -0,0 +1,49 @@
+use azure_core::{
+    headers::{self, Headers},
+    SessionToken,
+};
+use time::OffsetDateTime;
+
+use crate::ResourceQuota;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct CommonHeaders {
```
yoshuawuyts, line 10: Can we add documentation about what this is and when it should be used here? Seeing the context of the PR helps me make sense of it, but I'm not sure if I would quite understand what it does if I saw this just on its own.

rylev:
Closing this as we think about a better way to reduce the boiler plate here.

PR #1026, opened 2022-08-18
Minor cleanups as followup of previous PR

msabansal:
Following up on https://github.com/Azure/azure-sdk-for-rust/pull/1018 and incorporating feedback comments. Also moved the http_client creation to be part of the class as i dont want to create a new client on every token refresh.

### DIFF
--- a/sdk/identity/src/token_credentials/client_certificate_credentials.rs
+++ b/sdk/identity/src/token_credentials/client_certificate_credentials.rs
@@ -145,7 +145,7 @@ fn get_encoded_cert(cert: &X509) -> azure_core::Result<String> {
 }
 
 fn openssl_error(err: ErrorStack) -> azure_core::error::Error {
-    Error::new(ErrorKind::DataConversion, err)
+    Error::new(ErrorKind::Credential, err)
 }
 
 #[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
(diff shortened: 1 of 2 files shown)


sdk/identity/src/token_credentials/client_certificate_credentials.rs
```diff
@@ -138,12 +140,12 @@ struct AadTokenResponse {
...
         "\"{}\"",
-        base64::encode(cert.to_pem().map_err(|_| openssl_error())?)
+        base64::encode(cert.to_pem().map_err(openssl_error)?)
     ))
 }
 
-fn openssl_error() -> azure_core::error::Error {
-    Error::message(ErrorKind::Credential, "Openssl decode error")
+fn openssl_error(err: ErrorStack) -> azure_core::error::Error {
+    Error::new(ErrorKind::DataConversion, err)
```
rylev, line 148: Are we sure `DataConversion` is a better choice than `Credential`?

rylev (approved):
Thanks!

sdk/identity/src/token_credentials/client_certificate_credentials.rs
bmc-msft: Thus far, we've used DataConversion is primarily used to indicate parsing errors into and out of the REST endpoints.

msabansal: I wasn't sure what to use. This is an error happening in data conversion from pem so i thought it would be more appropriate. i can update it to credential. Don't have a strong stance for it.

rylev: I don't think it matters too much, but for now, I would keep it as `ErrorKind::Credential`. We may want to revisit these `ErrorKind`s in general, as it's becoming clear when they're useful and when they're just boiler plate...

msabansal: fixed.

bmc-msft:
Thanks for the contribution!

PR #1028, opened 2022-08-19
Improve message from error in retry policy that is no longer retried

rylev:
There are two changes in this PR:
* Previously, the retry policy was wrapping all non-200 responses in an error with the message "server returned error status which will not be retried: {status}" even the ones that *are* retired. This change now only uses that message for errors that are definitely retried. 
* Additional context is now added to the error messages that *were* retried but no longer are because the retry limit was reached.

### DIFF
--- a/sdk/core/src/error/mod.rs
+++ b/sdk/core/src/error/mod.rs
@@ -133,6 +133,23 @@ impl Error {
         }
     }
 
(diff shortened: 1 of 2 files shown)


bmc-msft (approved):

gorzell (approved):
Thanks for the quick fix.

PR #1029, opened 2022-08-19
reduce log level when creating a reqwest instance

bmc-msft:


sdk/core/src/http_client/reqwest.rs
```diff
@@ -7,7 +7,7 @@ use std::{collections::HashMap, str::FromStr};
 
 /// Construct a new `HttpClient` with the `reqwest` backend.
 pub fn new_reqwest_client() -> std::sync::Arc<dyn HttpClient> {
-    log::info!("instantiating an http client using the reqwest backend");
+    log::trace!("instantiating an http client using the reqwest backend");
```
rylev, line 10: Hmmm.. I wonder if we should split the difference and do `debug`. `trace` can be *very* noisy, and this (which happens once per client initialization) seems like something you wouldn't be annoyed to see in debug output.

bmc-msft, line 10:
```suggestion
    log::debug!("instantiating an http client using the reqwest backend");
```

rylev (approved):

PR #1031, opened 2022-08-19
reduce log level on requests that will not be retried

bmc-msft:
This log entry generates errors on multiple common patterns that should not result in the SDK generating error logs.

As an example, the recommended method for checking if a blob exists is to read it's properties.  If a blob does not exist, it generates a 404 response, which causes an error here.

Given the SDK returns an error, this should be left up to the SDK consumer if these warrants logging an error.

### DIFF
--- a/sdk/core/src/policies/retry_policies/retry_policy.rs
+++ b/sdk/core/src/policies/retry_policies/retry_policy.rs
@@ -101,8 +101,9 @@ where
                         );
                         error
                     } else {
-                        log::debug!("non-io error occurred which will not be retried: {}", error);
-                        return Err(error);
+                        return Err(
+                            error.context("non-io error occurred which will not be retried")
+                        );
                     }
                 }
             };
--- a/sdk/core/src/policies/retry_policies/retry_policy.rs
+++ b/sdk/core/src/policies/retry_policies/retry_policy.rs
@@ -73,7 +73,7 @@ where
                     );
 
                     if !RETRY_STATUSES.contains(&status) {
-                        log::error!(
+                        log::debug!(
                             "server returned error status which will not be retried: {}",
                             status
                         );
--- a/sdk/core/src/policies/retry_policies/retry_policy.rs
+++ b/sdk/core/src/policies/retry_policies/retry_policy.rs
@@ -101,7 +101,7 @@ where
                         );
                         error
                     } else {
-                        log::error!("non-io error occurred which will not be retried: {}", error);
+                        log::debug!("non-io error occurred which will not be retried: {}", error);
                         return Err(error);
                     }
                 }


sdk/core/src/policies/retry_policies/retry_policy.rs
```diff
@@ -101,7 +101,7 @@ where
                         );
                         error
                     } else {
-                        log::error!("non-io error occurred which will not be retried: {}", error);
+                        log::debug!("non-io error occurred which will not be retried: {}", error);
```
rylev, line 104: The change above makes a lot of sense. But the errors here are all non-HTTP errors. Wouldn't logging at an `error` level still make sense?

rylev (changes requested):

gorzell:
I am actually wondering if the SDK should log at all in the case where it is returning a `Result`? Shouldn't the caller get to decide if they want to log then, that way the SDK isn't polluting the applications logs. A lower level is a step in the right direction, but unless there is some information in the log that isn't available in the error, I am not sure I see the point, and even then I would think it would be better to just that information in the `Error`.

rylev:
> Shouldn't the caller get to decide if they want to log then, that way the SDK isn't polluting the applications logs

The caller can trivially filter out the SDKs logs - e.g., log_env allows the user to set which crates they want logs from. I think the SDK should be giving enough information for users of the SDK to have a good idea of what's happening. 

The logs here give additional information about what is happening inside of the retry policy. It can be helpful to know what the state of the retry policy is. I've personally used these logs to understand which errors were being retried.

gorzell:
> The logs here give additional information about what is happening inside of the retry policy. It can be helpful to know what the state of the retry policy is. I've personally used these logs to understand which errors were being retried.

Sorry, I wasn't suggesting that **all** logging should be removed. Retry state is a really good example of something that is useful because it can present internal state the caller wouldn't otherwise see. I just think that the pattern of logging an `Error` right before you return that error and with no extra information in the log message is perhaps a bit excessive.

bmc-msft:
> > The logs here give additional information about what is happening inside of the retry policy. It can be helpful to know what the state of the retry policy is. I've personally used these logs to understand which errors were being retried.
> 
> Sorry, I wasn't suggesting that **all** logging should be removed. Retry state is a really good example of something that is useful because it can present internal state the caller wouldn't otherwise see. I just think that the pattern of logging an `Error` right before you return that error and with no extra information in the log message is perhaps a bit excessive.

@gorzell identifies the specific concern I was trying to address with this PR.

My thought is that logging an error condition and returning an Error with the same detail isn't needed.  The `must use` requirement of Results means users will get this information that would have been logged.

rylev:
I believe that the logs *do* encode additional information. The user won't necessarily be familiar with the workings on the retry policy, and thus they may not know that for example, a 419 does not get retried by a 503 does. These logs alert the user that an error has occurred (something they may already know) which will not be retried (potentially new information).

bmc-msft:
Would it be an acceptable compromise to add that information as context to the error such that the control as to what to do with the information is in the hands of the consumer, including acting on the issue?

gorzell:
> Would it be an acceptable compromise to add that information as context to the error such that the control as to what to do with the information is in the hands of the consumer, including acting on the issue?

Improving the information/context in the Error would be my strong preference for providing additional information to the caller (when possible/practical).

rylev:
> Would it be an acceptable compromise to add that information as context to the error such that the control as to what to do with the information is in the hands of the consumer, including acting on the issue?

I suppose so? I still don't quite understand why this log is problematic. It provides additional context in a way that can trivially filtered from logs if users don't want that level of detail. The fact that the error won't be retried isn't really context about the error but rather context to understand how your operation is being processed. It makes more sense to me that it would be logged and not attached to the error, but I'm obviously missing something here 😊

rylev (approved):
I still think logging makes more sense, but I'm happy to see this land if that's what everyone else thinks is the right path forward.

Issue #1032, opened 2022-08-20
converting from Chrono to Time broke SAS signatures


bmc-msft:
The conversion from Chrono to Time broke the SAS signature generation.  

Previously, the expiry & start times for SAS generation was formatted using 
```rust
s.push_str(&sa.start.format("%Y-%m-%dT%H:%M:%SZ").to_string());
```

Now it's generated using:
```rust
s.push_str(&date::to_rfc3339(&sa.start));
```

This increases the precision to include microseconds/nanoseconds.  If you create a SAS using the existing SDK & OffsetDateTime::utc_now(), you will have a timestamp that includes microseconds such as `2022-08-20T03:32:41.627587163Z`.  Azure will take the URL arguments, canonicalize them, and compare against the timestamp without the microseconds.

This can be seen with azcopy, which will generate the error:

```
AuthenticationErrorDetail: Signature did not match. String to sign used was cw
2022-08-20T03:32:41Z                                                                                                                      
```

bmc-msft:
Note, for those that are currently broken, you can work around this by doing `replace_nanosecond(0)` on your expiry and/or start datetimes.

PR #1033, opened 2022-08-20
fix formatting datetime for SAS generation

bmc-msft:
Azure uses RFC3339 format, except without microseconds/nanoseconds

### DIFF
--- a/sdk/storage/src/core/shared_access_signature/mod.rs
+++ b/sdk/storage/src/core/shared_access_signature/mod.rs
@@ -10,7 +10,8 @@ pub trait SasToken {
 }
 
 pub(crate) fn format_date(d: OffsetDateTime) -> String {
-    azure_core::date::to_rfc3339(&d)
+    // replacing nanosecond with 0 is known to not panic
+    azure_core::date::to_rfc3339(&d.replace_nanosecond(0).unwrap())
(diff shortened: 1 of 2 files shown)


bmc-msft:
fixes #1032

cataggar (approved):
Thanks @bmc-msft. Could you add a unit test? Something like the 2nd example:
https://docs.rs/time/latest/time/format_description/well_known/struct.Rfc3339.html

sdk/storage/src/core/shared_access_signature/mod.rs
```diff
@@ -10,7 +10,8 @@ pub trait SasToken {
 }
 
 pub(crate) fn format_date(d: OffsetDateTime) -> String {
-    azure_core::date::to_rfc3339(&d)
+    // replacing nanosecond with 0 is known to not panic
+    azure_core::date::to_rfc3339(&d.replace_nanosecond(0).unwrap())
```
rylev, line 14: I would also write an explanation (much like you did in the associated issue) for why this is necessary.

rylev (changes requested):

ctaggart (approved):

rylev (approved):

PR #1035, opened 2022-08-22
core: Change the RetryPolicy trait to allow waiting on arbitrary things.

gorzell:
Previously the only thing that that a retry would wait for was a period
of time, i.e. `Duration`. This adds another function to the `RetryPolicy` trait
that allows the implementer to wait on any condition they like. The
default implementation of the new method provides backward compatibility
by just sleeping for a period of time.

A real world example of this is flow control. If you are consistently getting errors from the service is too busy, you may want to slow send less requests across multiple threads. Long term it might be better to add a custom policy for this so that you have access to the exact error, but it would also need to be aware of retries that had occurred, and probably sit later in the `Pipeline`. Another option would be to pass the error into this new function.

### DIFF
--- a/sdk/core/src/policies/retry_policies/retry_policy.rs
+++ b/sdk/core/src/policies/retry_policies/retry_policy.rs
@@ -20,6 +20,13 @@ pub trait RetryPolicy {
     fn is_expired(&self, duration_since_start: Duration, retry_count: u32) -> bool;
     /// Determine how long before the next retry should be attempted.
     fn sleep_duration(&self, retry_count: u32) -> Duration;
+    /// A Future that will wait until the request can be retried.
+    fn wait(
+        &self,
+        retry_count: u32,
+    ) -> std::pin::Pin<std::boxed::Box<dyn std::future::Future<Output = ()> + Send>> {
+        Box::pin(sleep(self.sleep_duration(retry_count)))
+    }
 }
 
 /// The status codes where a retry should be attempted.
@@ -116,7 +123,7 @@ where
             }
             retry_count += 1;
 
-            sleep(self.sleep_duration(retry_count)).await;
+            self.wait(retry_count).await;
         }
     }
 }
--- a/sdk/core/src/policies/retry_policies/retry_policy.rs
+++ b/sdk/core/src/policies/retry_policies/retry_policy.rs
@@ -23,7 +23,7 @@ pub trait RetryPolicy {
     /// Determine how long before the next retry should be attempted.
     fn sleep_duration(&self, retry_count: u32) -> Duration;
     /// A Future that will wait until the request can be retried.
-    async fn wait(&self, retry_count: u32) {
+    async fn wait(&self, retry_count: u32, _error: &Error) {
         sleep(self.sleep_duration(retry_count)).await;
     }
 }
@@ -122,7 +122,7 @@ where
             }
             retry_count += 1;
 
-            self.wait(retry_count).await;
+            self.wait(retry_count, &last_error).await;
         }
     }
 }
--- a/sdk/core/src/policies/retry_policies/retry_policy.rs
+++ b/sdk/core/src/policies/retry_policies/retry_policy.rs
@@ -11,9 +11,12 @@ use std::time::Duration;
 
 /// A retry policy.
 ///
-/// All retry policies follow a similar pattern only differing in how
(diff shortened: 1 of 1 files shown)


aneubeck (approved):
👍

sdk/core/src/policies/retry_policies/retry_policy.rs
```diff
@@ -20,6 +20,13 @@ pub trait RetryPolicy {
     fn is_expired(&self, duration_since_start: Duration, retry_count: u32) -> bool;
     /// Determine how long before the next retry should be attempted.
     fn sleep_duration(&self, retry_count: u32) -> Duration;
+    /// A Future that will wait until the request can be retried.
+    fn wait(
+        &self,
+        retry_count: u32,
+    ) -> std::pin::Pin<std::boxed::Box<dyn std::future::Future<Output = ()> + Send>> {
```
rylev, line 27: If we use the `async_trait` macro this would be much simpler.

sdk/core/src/policies/retry_policies/retry_policy.rs
```diff
@@ -20,6 +20,13 @@ pub trait RetryPolicy {
     fn is_expired(&self, duration_since_start: Duration, retry_count: u32) -> bool;
     /// Determine how long before the next retry should be attempted.
     fn sleep_duration(&self, retry_count: u32) -> Duration;
```
rylev, line 22: If we have the `wait` method, do we need `sleep_duration` any more?

rylev (commented):
I'd like to think about this a bit, but I have some questions in the mean time.

sdk/core/src/policies/retry_policies/retry_policy.rs
gorzell: Sure, happy to do that. It only took me 10 min to figure our this signature so I could avoid using it 😭.

sdk/core/src/policies/retry_policies/retry_policy.rs
gorzell: It is a convenience for the simpler cases, but I don't think it is strictly required and it may not always be implemented.

sdk/core/src/policies/retry_policies/retry_policy.rs
```diff
@@ -13,13 +14,18 @@ use std::time::Duration;
 /// All retry policies follow a similar pattern only differing in how
```
rylev, line 14: This comment is not really correct today (and points to the fact that this trait wasn't originally intended to be extended by end users). Can we update the comment to reflect the new reality?

rylev (approved):
I think this is a fine thing to try - since we're still not guaranteeing backwards compatibility I don't mind letting this through and experimenting with it.

sdk/core/src/policies/retry_policies/retry_policy.rs
```diff
@@ -23,7 +23,7 @@ pub trait RetryPolicy {
     /// Determine how long before the next retry should be attempted.
     fn sleep_duration(&self, retry_count: u32) -> Duration;
     /// A Future that will wait until the request can be retried.
-    async fn wait(&self, retry_count: u32) {
+    async fn wait(&self, retry_count: u32, _error: &Error) {
```
rylev, line 26: I think the error should probably go first. We may also want to consider calling it `last_error` or provide documentation that this is the error that caused the retry.

sdk/core/src/policies/retry_policies/retry_policy.rs
```diff
@@ -3,23 +3,33 @@ use crate::policies::{Policy, PolicyResult, Request};
...
+#[async_trait]
 pub trait RetryPolicy {
     /// Determine if no more retries should be performed.
     ///
     /// Must return true if no more retries should be attempted.
     fn is_expired(&self, duration_since_start: Duration, retry_count: u32) -> bool;
     /// Determine how long before the next retry should be attempted.
     fn sleep_duration(&self, retry_count: u32) -> Duration;
+    /// A Future that will wait until the request can be retried.
+    /// `error` is the [`Error`] value the led a retry attempt.
```
rylev, line 29:
```suggestion
    /// `error` is the [`Error`] value the led to a retry attempt.
```

rylev (changes requested):
One small nit, but otherwise looks great!

rylev (approved):

PR #1038, opened 2022-08-23
Make RetryPolicy publicly visible.

gorzell:
I missed this in #1035. I actually tested it this time 😉.


ctaggart (approved):

gorzell:
Note: It is kinda wonky to use this still because you have to set RetryOptions to `NoRetry` and then add a per retry policy that implements this trait. But since we are looking to refactor the client building anyway, that probably doesn't matter much. Technically you can't set any options on the client without a patch.

rylev (approved):

Issue #1039, opened 2022-08-24
adjust generated clients to be more like the .NET generated rest clients


ctaggart:
Related to #446, the generated service clients need some adjustments to be usable by the SDKs #803. The generated services are similar to the generated `RestClient`s in .NET.  Let's look at how `Azure.Storage.Blobs` .NET SDK builds on top of the generated rest client for the operation `Service_GetAccountInfo`.

- [sdk/storage/Azure.Storage.Blobs/src/Generated/ServiceRestClient.cs](https://github.com/Azure/azure-sdk-for-net/blob/568d3730e1f8d9b299599ccaa2d31120352c1031/sdk/storage/Azure.Storage.Blobs/src/Generated/ServiceRestClient.cs#L452-L466)
- [sdk/storage/Azure.Storage.Blobs/src/Generated/ServiceGetAccountInfoHeaders.cs](https://github.com/Azure/azure-sdk-for-net/blob/568d3730e1f8d9b299599ccaa2d31120352c1031/sdk/storage/Azure.Storage.Blobs/src/Generated/ServiceGetAccountInfoHeaders.cs)
- [sdk/storage/Azure.Storage.Blobs/src/BlobServiceClient.cs](https://github.com/Azure/azure-sdk-for-net/blob/568d3730e1f8d9b299599ccaa2d31120352c1031/sdk/storage/Azure.Storage.Blobs/src/BlobServiceClient.cs#L801-L844)
- [sdk/storage/Azure.Storage.Blobs/src/BlobExtensions.cs](https://github.com/Azure/azure-sdk-for-net/blob/568d3730e1f8d9b299599ccaa2d31120352c1031/sdk/storage/Azure.Storage.Blobs/src/BlobExtensions.cs#L234-L246)

``` csharp
        /// <summary> Returns the sku name and account kind. </summary>
        /// <param name="cancellationToken"> The cancellation token to use. </param>
        public async Task<ResponseWithHeaders<ServiceGetAccountInfoHeaders>> GetAccountInfoAsync(CancellationToken cancellationToken = default)
        {
            using var message = CreateGetAccountInfoRequest();
            await _pipeline.SendAsync(message, cancellationToken).ConfigureAwait(false);
            var headers = new ServiceGetAccountInfoHeaders(message.Response);
            switch (message.Response.Status)
            {
                case 200:
                    return ResponseWithHeaders.FromValue(headers, message.Response);
                default:
[...]

PR #1040, opened 2022-08-24
per operation `Response` types for generated clients

ctaggart:
This implements the first part of #1053 design.

``` rust
RequestBuilder
fn send(self) -> Response

Response(azure_core::Response)
fn into_body(self) -> ParsedBody
fn into_raw_reponse(self) -> azure_core::Response
fn as_raw_reponse(&self) -> &azure_core::Response
```


services/svc/blobstorage/src/package_2021_02/mod.rs
```diff
@@ -254,7 +254,14 @@ pub mod service {
     }
     pub mod get_properties {
         use super::models;
-        type Response = models::StorageServiceProperties;
+        struct Response(azure_core::Response);
```
ctaggart, line 257: Each operation has its own `Response`.

services/svc/blobstorage/src/package_2021_02/mod.rs
```diff
@@ -254,7 +254,14 @@ pub mod service {
...
+        impl Response {
+            pub async fn into_body(self) -> azure_core::Result<models::StorageServiceProperties> {
```
ctaggart, line 259: It is now an additional `into_body` to get the response body. If typed headers are available, they will be made available via `headers(&self)` in a future pull request.

services/autorust/codegen/src/codegen_operations.rs
```diff
@@ -663,64 +663,59 @@ impl ResponseCode {
...
-                            match self {
-                                #continuation_response
-                            }
-                        }
-                    }
-                });
-            }
         }
+        // HELP!!!
+        // let mut continuation_response = TokenStream::new();
```
ctaggart, line 695: I could use some help figuring out how to adjust pageable support. @bmc-msft

services/svc/blobstorage/src/package_2021_02/mod.rs
```diff
@@ -524,18 +513,14 @@ pub mod service {
...
-                            }
+                        let rsp = match rsp.status() {
+                            azure_core::StatusCode::Ok => Ok(Response(rsp)),
                             status_code => Err(azure_core::error::Error::from(azure_core::error::ErrorKind::HttpResponse {
                                 status: status_code,
                                 error_code: None,
                             })),
-                        }
+                        };
+                        Ok(rsp?.into_body().await?)
```
cataggar, line 523: I'm not sure this is what we want. @bmc-msft

services/svc/blobstorage/src/package_2021_02/mod.rs
```diff
@@ -475,7 +464,7 @@ pub mod service {
                 self.x_ms_client_request_id = Some(x_ms_client_request_id.into());
                 self
             }
-            pub fn into_stream(self) -> azure_core::Pageable<Response, azure_core::error::Error> {
+            pub fn into_stream(self) -> azure_core::Pageable<models::ListContainersSegmentResponse, azure_core::error::Error> {
```
cataggar, line 467: Should this be left as `Response`?

bmc-msft:
Can you discuss why this change is needed?  It would be helpful to understand the need, as this is increasing the amount of boiler plate developers will need to write.   Consider the existing Application Insights example:

```rust
let response = client.query_client().execute(app_id, body).into_future().await?;
```

With this change, the example looks like this:
```rust
let query_results = client.query_client().execute(app_id, body).into_future().await?.into_body().await?;
```

cataggar:
Thanks for the feedback @bmc-msft. I added a `.into_body()` helper to the `RequestBuilder` that does the same thing as `into_future()` was doing before which is to send the request and get back the response body. It is back to the same amount of boilerplate.

The `send()` function now returns a response. The response with allow access to the typed headers (next PR) and the raw response if needed.  This is the same design as the .NET rest clients #1039.

yoshuawuyts (changes requested):
This PR contains a million-line diff, which is incredibly hard to review. Can you walk us through the API changes you're proposing here? I'd like to be able to understand which changes you're making, and why.

As we discussed in the meeting on Tuesday, there are several ways which would make it easier to talk about API design changes:
- Instead of generating the changes for all files, generate a diff for a single file, or possibly even a single SDK. This allows us to zoom in on a single set of changes and discuss them in detail.
- Alternatively: instead of proposing the API changes for the auto-generated SDKs, file a PR to make the changes to the manually authored SDKs first. It makes sense to not go down this path if you're not comfortable with the manually authored SDKs.
- Author a document or PR description outlining the changes you're making with clear descriptions of the problems and proposed solutions. I would say this is a requirement for API changes, as the auto-generated and manually-authored crates should maintain similar user-facing APIs.

As @rylev mentioned on Tuesday: it's fine for the manually-authored and automatically generated SDKs to deviate in their APIs. But this should be the result of a conscious decision making process, and not because we simply didn't communicate about the design.

Thanks!

cataggar:
> Instead of generating the changes for all files, generate a diff for a single file, or possibly even a single SDK. This allows us to zoom in on a single set of changes and discuss them in detail.

@yoshuawuyts, I opened up #1043 to make is easier to zoom in on a single set of changes for the strorage `Service_GetAccountInfo` operation.

> I would say this is a requirement for API changes, as the auto-generated and manually-authored crates should maintain similar user-facing APIs.

The main difference is the SDKs combine data from the response headers and the response bodies for their responses. The generated rest clients allow type-safe access to individual response headers and response body. In #1039, the `ToAccountInfo` is the manually written conversion function.

services/mgmt/compute/examples/list_free_linux_images.rs
```diff
@@ -24,7 +24,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let mut publishers = client
         .list_publishers(&location, &subscription_id)
-        .into_future()
+        .into_body()
```
cataggar, line 27: The convenience function that sends the request and returns the body is now `into_body`.

services/mgmt/network/examples/list_azure_service_addresses.rs
```diff
@@ -17,7 +17,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let subscription_id = AzureCliCredential::get_subscription()?;
     let client = azure_mgmt_network::Client::builder(credential).build();
 
-    let response = client.service_tags_client().list(location, subscription_id).into_future().await?;
+    let response = client.service_tags_client().list(location, subscription_id).into_body().await?;
```
cataggar, line 20: Another example.

services/svc/batch/examples/create_task.rs
```diff
@@ -31,12 +31,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     println!("creating job");
     let job_params = JobAddParameter::new(job_id.clone(), pool_info);
-    client.job_client().add(job_params).into_future().await?;
+    client.job_client().add(job_params).send().await?;
```
cataggar, line 34: If you don't need the response body, just call `send`. An `into_body` is only available if there is a body.

services/svc/graphrbac/examples/purge_deleted_apps.rs
```diff
@@ -23,7 +23,7 @@ async fn main() -> azure_core::Result<()> {
         for app in apps.value {
             println!("{:?}", app.display_name);
             let obj_id = app.directory_object.object_id.expect("missing object id");
-            client.hard_delete(obj_id, &tenant_id).into_future().await?;
+            client.hard_delete(obj_id, &tenant_id).send().await?;
```
cataggar, line 26: Another `send` example.

services/mgmt/compute/examples/list_free_linux_images.rs
cataggar: Operations like `Service_GetAccountInfo` do not have response bodies. All of the retuned data is in the response headers, which are not currently accessible. The plan is make them available, probably via `Response::headers(&self)` which follows the AutoRest C# design. I'm not sure what `into_future` should map to and would prefer to add it back once we figure that out.

ctaggart: When cadl replaces openapi for the specifications, it will be possible to model the responses the way we with them to be in the SDKs. I created an example for operation in https://github.com/Azure/azure-sdk-for-rust/pull/1043#issuecomment-1228933586 .

cataggar:
@rylev, I provided an overview and switched back to `into_future` as we discussed today. Please review/approve.

bmc-msft (changes requested):
This requires the user know where the response data comes from, either from the header or body, which seems sub-optimal.  The data representation from the HTTP response should be inconsequential to the API consumer.

If we consider some of the other SDKs, this is handled in a handful of different ways.
1. Golang: appears to provide the parsed response
https://github.com/Azure-Samples/azure-sdk-for-go-samples/blob/d9f41170eaf6958209047f42c8ae4d0536577422/services/keyvault/examples/go-keyvault-msi-example.go#L82-L89
2. Python: if you provide the argument `raw`, returns the response with the expectation that the user deserializes it, otherwise the consumer gets a fully contextualized response.
3. Javascript: Users register a callback where you are provided the deserialized response or the base response stream.

It appears the primary concern would be addressed if we provided *either* the contextualized/parsed response _or_ a raw version.  Can we do something akin to [std::str::parse](https://doc.rust-lang.org/stable/std/primitive.str.html#method.parse), and rely on type inference (or turbofishing) to know which response the user wants?

cataggar:
> This requires the user know where the response data comes from, either from the header or body, which seems sub-optimal. The data representation from the HTTP response should be inconsequential to the API consumer.

The is true for SDKs, but not the generated clients. AutoRest clients make this distinction. See `ResponseWithHeaders` in the #1039 example.  The generated clients represent the openapi in Rust code. Some .NET APIs call it response value instead of response body.

> Golang: appears to provide the parsed response

The generated [GetSecret](https://github.com/Azure/azure-sdk-for-go/blob/bc609df91cbdb3a3bec8f332dc323362e282e07f/sdk/keyvault/azsecrets/client.go#L174-L195) is: 

``` golang
// GetSecret - The GET operation is applicable to any secret stored in Azure Key Vault. This operation requires the secrets/get
// permission.
// If the operation fails it returns an *azcore.ResponseError type.
// Generated from API version 7.3
// name - The name of the secret.
// version - The version of the secret. This URI fragment is optional. If not specified, the latest version of the secret
// is returned.
// options - GetSecretOptions contains the optional parameters for the Client.GetSecret method.
func (client *Client) GetSecret(ctx context.Context, name string, version string, options *GetSecretOptions) (GetSecretResponse, error) {
	req, err := client.getSecretCreateRequest(ctx, name, version, options)
	if err != nil {
		return GetSecretResponse{}, err
	}
	resp, err := client.pl.Do(req)
	if err != nil {
		return GetSecretResponse{}, err
	}
	if !runtime.HasStatusCode(resp, http.StatusOK) {
		return GetSecretResponse{}, runtime.NewResponseError(resp)
	}
	return client.getSecretHandleResponse(resp)
}
```

It returns a response type named `GetSecretResponse`:
``` golang
// GetSecretResponse contains the response from method Client.GetSecret.
type GetSecretResponse struct {
	SecretBundle
}
```

The `SecretBundle` is the response body. The spec for the operation does not return any headers.  To skip the deserialization, you would need to call `getSecretCreateRequest` yourself.

For these Rust operation Response types. The intension it to provide `into_raw_response` and `as_raw_response` with `Into<azure_core::Response>` and `AsRef<azure_core::Response>` as well. I can update the pull request.

cataggar:
Looking at generated JavaScript clients, it looks like primary way to respond is with a custom response that contains both the parsed body and parsed headers. Using [getStatistics](https://docs.microsoft.com/en-us/javascript/api/@azure/storage-blob/blobserviceclient?view=azure-node-latest#@azure-storage-blob-blobserviceclient-getstatistics) as an example:

``` ts
function getStatistics(options?: ServiceGetStatisticsOptions): Promise<ServiceGetStatisticsResponse>

type ServiceGetStatisticsResponse = ServiceGetStatisticsHeaders &
  BlobServiceStatistics & {
    _response: HttpResponse & {
      bodyAsText: string,
      parsedBody: BlobServiceStatistics,
      parsedHeaders: ServiceGetStatisticsHeaders,
    },
  }
```

We could provide something similar from `into_future`.

cataggar:
## Design with Parsed Response

Here is a proposed design that provides for parsed responses similar to the generated JavaScript clients. Each operation would have the following:

```
ResponseBuilder
fn into_future(self) -> ParsedResponse
fn send(self) -> Response

ParsedResponse
headers: ParsedHeaders
body: T

ParsedHeaders
x_ms_request_id: Option<String>

Response(azure_core::Response)
fn headers(&self) -> Headers
fn parsed_headers(&self) -> ParsedHeaders
fn into_body(self) -> T
fn into_parsed_response(self) -> ParsedResponse

Headers(&azure_core::headers::Headers)
fn x_ms_request_id(&self) -> azure_core::Result<&str>
```

If we go with this design, it can be broken up into multiple pull requests. This pull request would change to `send` and we would reserve `into_future` for a `ParsedResponse`.

cataggar:
I converted this to a draft while we come to agreement on the proposed design in #1053.

cataggar:
This implements the first part of the #1053 design. It should be good to merge now. @rylev, @bmc-msft

rylev:
I'm fine with merging this if it unblocks folks as long as no one is under the illusion that merging this means we've decided on an actual design which we're still discussing in #1053. Waiting for @bmc-msft's thoughts.

bmc-msft:
I recognize this is a work-in-progress, but is a step backwards in terms of usability as it stands right now.

Currently, nearly all of the published generated crates are broken (fixed in #1020).  Can an update be pushed out before merging this work?

cataggar:
> I recognize this is a work-in-progress, but is a step backwards in terms of usability as it stands right now.

I left `RequestBulider::into_body` in place to prevent it being a step backwards. I intend to add the other features soon. I just need to get unblocked on this review.

> I'm fine with merging this if it unblocks folks as long as no one is under the illusion that merging this means we've decided on an actual design which we're still discussing in #1053.

Sure. We should come to an agreement before the September release in a week or two.

bmc-msft (approved):
Based off of an offline discussion, I am marking this approved as to unblock @cataggar

PR #1042, opened 2022-08-24
enable anonymous access to storage containers

### LINKED ISSUE #1030: Request: allow using anonymous clients for blob storage
If a particular storage account is configured for public access, the Azure Blob Storage service allows anonymously enumerating the blobs in a container (e.g.: as in [this tutorial for .NET](https://docs.microsoft.com/azure/storage/blobs/anonymous-read-access-client?tabs=dotnet)); it would be really helpful if we could create a `StorageClient` value in the Rust SDK without authentication to allow using this functionality from Rust as well. Thank you! 💕

bmc-msft:
fixes #1030

### DIFF
--- a/sdk/storage/src/core/clients/storage_client.rs
+++ b/sdk/storage/src/core/clients/storage_client.rs
@@ -33,6 +33,7 @@ pub enum StorageCredentials {
     SASToken(Vec<(String, String)>),
     BearerToken(String),
     TokenCredential(Arc<dyn TokenCredential>),
+    Anonymous,
 }
 
 impl std::fmt::Debug for StorageCredentials {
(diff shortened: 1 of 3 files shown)


rylev (approved):
Looks good to me!

cc @gorzell as this might interact a bit with the storage refactor.

sdk/storage/src/core/clients/storage_client.rs
```diff
@@ -336,6 +341,35 @@ impl StorageClient {
         }
     }
 
+    /// Create a new anonymous storage client
+    ///
+    /// Ref: https://docs.microsoft.com/en-us/azure/storage/blobs/anonymous-read-access-configure?tabs=portal
```
cataggar, line 346: Cool. nit: it is preferred to drop the `/en-us` segment from the docs URLs.

PR #1045, opened 2022-08-28
move `azure_storage::xml` to `azure_core::xml` behind `xml` feature

ctaggart:
This moves the `xml` module to `azure_core` so that it can be used by the generated service clients too. It is behind an `xml` feature.


rylev (approved):
👍

PR #1048, opened 2022-08-29
address more header names that are not canonicalized before use

bmc-msft:
Because Header Names are canonicalized to lowercase, use of these headers broke when the headername implementation changed.


bmc-msft:
It would be great if we could find a way to ensure header names are always canonicalized moving forwards.  These were not caught during the shift from `http::Request` to `azure_core::Request` as part of #833.

For reference, our existing methods for creating `HeaderName`, which only check when we're creating them from String or Cow, but not when making static versions:
https://github.com/Azure/azure-sdk-for-rust/blob/a2fc3403121f3e03a048469b2fab6c387a6696c5/sdk/core/src/headers/mod.rs#L174-L206

cataggar (approved):

Issue #1051, opened 2022-08-30
security_keyvault model types are inaccessible due to private modules


johnbatty:
Found on the latest (unreleased) main code.

It seems that it is not possible to specify `security_keyvault` model types, as the submodules are private.

The `get_secret` example builds:
```rust
     let secret = client.get(secret_name).into_future().await?;
```

However, if I want to express the type of the return value, I find this impossible.  I can declare secret to be a different type to get the compiler to tell me what it thinks the return type is:
```rust
  --> sdk/security_keyvault/examples/get_secret.rs:25:23
   |
25 |     let secret: i32 = client.get(secret_name).into_future().await?;
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `i32`, found struct `azure_security_keyvault::secrets::models::KeyVaultGetSecretResponse`
```

If I use this type, then I find that it includes a private module:
```rust
  --> sdk/security_keyvault/examples/get_secret.rs:25:42
   |
25 |     let secret: azure_security_keyvault::secrets::models::KeyVaultGetSecretResponse =
   |                                          ^^^^^^^ private module
```

I noticed that the `prelude` does include the submodule types, but declares them as `pub(crate)` so again they are not accessible:
```rust
// prelude.rs
pub use crate::clients::*;
pub(crate) use crate::{account::*, certificates::*, keys::*, secrets::*};
```

The simple fix is clearly to make these module declarations `pub` to make them publicly visible (I've already done this and will create a PR).  However stepping back we should agree on how models and operations should be exposed from all the SDK crates, to make them consistent.  Looking at `data_cosmos` as a reference crate, I see that all model and operation modules are declared as `pub`, plus it exposes many types (all clients and all operations, all resources) via the prelude:
```rust
pub use crate::{ConsistencyLevel, CosmosEntity};

pub use crate::clients::*;

pub use crate::resources::collection::Offer;
pub use crate::resources::document::*;
pub use crate::resources::*;

pub use permission::AuthorizationToken;

pub use crate::operations::*;
```

Do we want the other SDKs to follow this approach?  Any guidance on what should/shouldn't go in the prelude?

PR #1052, opened 2022-08-30
security_keyvault module visibility fixups

johnbatty:
Fixes https://github.com/Azure/azure-sdk-for-rust/issues/1051

### DIFF
--- a/sdk/security_keyvault/examples/backup_secret.rs
+++ b/sdk/security_keyvault/examples/backup_secret.rs
@@ -1,5 +1,5 @@
 use azure_identity::DefaultAzureCredentialBuilder;
-use azure_security_keyvault::SecretClient;
+use azure_security_keyvault::prelude::*;
 use std::{env, sync::Arc};
 
 #[tokio::main]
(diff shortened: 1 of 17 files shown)


rylev:
I think we should land this (once the tests pass), but I think this points to a larger need of thinking through what is included in the prelude and what is not - or if we even need a prelude. I think we should do that after we get these fixes in.

johnbatty:
Yeah, personally I'd be happy without a prelude - think that only makes sense for very common names that are likely to be used by all users - and with such a large API I think it is difficult to determine what is "common".  I'm not keen on including everything in the prelude, as I have done here...

Also need to decide whether to "flatten" the modules as the current code does...
```rust
// secrets/mod.rs
mod models;
mod operations;

pub use models::*;
pub use operations::*;
```

Or to expose the `models` and `operations` modules in the namespacing:
```rust
pub mod models;
pub mode operations;
```

johnbatty:
I think this is good to merge.

We should definitely discuss and agree a preferred module structure - looking at the other SDK crates they all have different approaches.  It would obviously be better if they had a more consistent structure.

rylev (approved):

PR #1054, opened 2022-08-31
Make `azure_data_cosmos::clients::CloudLocation` public

hanhossain:
I was trying to use the SDK against the CosmosDB emulator, and it looks like `CosmosClientBuilder` has a few public methods which aren't usable right now as `CloudLocation` is not public. This one-liner exports `CloudLocation` as public to allow those methods to be used.


rylev (changes requested):
Thanks! It looks like CI doesn't pass because the newly public `CloudLocation` is not properly documented (and we require all public items to be documented). Would you mind trying to document the struct?

rylev (approved):
Thanks!

PR #1057, opened 2022-08-31
Stop using `StorageClient` in blob storage

rylev:
*Note: This builds on work done by @gorzell.*

This change is also a fairly simple one but has far reaching effects.

This removes the use of `StorageClient` from `azure_blob_stroage`. `StorageClient` now has no operations of its own and so is effectively just a helper. This ditches using that type in favor of the hierarchy starting at `BlobServiceClient`. Instead of a whole mess of constructors, construction of clients follows the pattern established in `azure_cosmos`. 

```rust
let storage_credentials = StorageCredentials::Key(account.clone(), access_key);
let service_client = BlobServiceClient::new(account, storage_credentials);
```

`BlobServiceClient` can easily be instantiated with an account name and some sort of `StorageCredentials`. We also offer `BlobServiceClientBuilder` which allows for more advanced creation of a `BlobServiceClient` through tuning of various options such as retry policy, transport policy, etc. This is the same pattern followed in `azure_cosmos`. 

I was not able to fully finish moving everything over, so I'm marking this as a draft for now, but I wanted to make everyone aware.

### DIFF
--- a/sdk/storage/src/clients/mod.rs
+++ b/sdk/storage/src/clients/mod.rs
@@ -1,3 +1,6 @@
 mod storage_client;
 
-pub use storage_client::{ServiceType, StorageClient, StorageCredentials, StorageOptions};
+pub use storage_client::{
+    finalize_request, new_pipeline_from_options, ServiceType, StorageClient, StorageCredentials,
+    StorageOptions,
+};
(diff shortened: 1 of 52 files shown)


bmc-msft:
The builder pattern used here makes setting options used by the pipeline much more reasonable.

bmc-msft (approved):

PR #1060, opened 2022-09-01
provide helper methods for creating StorageCredentials

bmc-msft:
One of the side effects of #1057 is that users now must create `StorageCredentials`.  For SAS tokens, this was previously handled for the user.

I'm not 100% sold on needing helper methods for anything but SAS Tokens, but I added them for completeness.

### DIFF
--- a/sdk/storage/src/clients/storage_client.rs
+++ b/sdk/storage/src/clients/storage_client.rs
@@ -35,12 +35,35 @@ pub enum StorageCredentials {
 }
 
 impl StorageCredentials {
+    pub fn new_storage_key<A, K>(account: A, key: K) -> Self
+    where
+        A: Into<String>,
+        K: Into<String>,
+    {
+        Self::Key(account.into(), key.into())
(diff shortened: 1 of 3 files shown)


sdk/storage/src/clients/storage_client.rs
```diff
@@ -34,6 +34,92 @@ pub enum StorageCredentials {
...
+        Self::Key(account.into(), key.into())
+    }
+
+    /// Create a Shared Access Signature (SAS) token based credential
+    ///
+    /// SAS tokens provide delegated access to resources in a storage account
+    /// with granular control over how the client can access data in the
+    /// account.
+    ///
+    /// ref: https://docs.microsoft.com/azure/storage/common/storage-sas-overview
```
rylev, line 60: This doc and the linked docs don't really reference the fact that `token` is expected to be a sas query string. Can we make that clear?

rylev (changes requested):

PR #1062, opened 2022-09-01
Start moving over `azure_storage_datalake` to use client builder pattern

rylev:
This is the successor to #1057. It moves the `azure_storage_datalake` crate over to the same pattern of client builders as Cosmos and Blob storage. The data lake crate did not use the `StorageClient` as a helper so the changes here are a bit different but the result is the same.  

The biggest change is that the `CloudLocation` which was temporarily in the `BlobServiceClient`'s module has been moved to its own module in the `azure_storage` crate so that it can be used by all the storage crates.

~~I'm not quite done yet, so marking this as a draft. Hopefully I can finish this tomorrow.~~


sdk/storage_datalake/examples/filesystem.rs
```diff
@@ -1,20 +1,20 @@
...
 async fn main() -> azure_core::Result<()> {
-    let data_lake_client = create_data_lake_client().await.unwrap();
+    let data_lake_client = create_data_lake_client();
 
     let file_system_name = format!(
         "azurerustsdk-datalake-example00-{}",
         OffsetDateTime::now_utc().unix_timestamp()
     );
     let file_system_client = data_lake_client
         .clone()
```
cataggar, line 16: Can we have the sub-client factory methods call `self.clone()`? That is the pattern in the generated clients:
https://github.com/Azure/azure-sdk-for-rust/blob/f99f9c1944a25b34420424daa792bb21c37ba6ba/services/svc/blobstorage/src/package_2021_08/mod.rs#L106-L108

cataggar (approved):

bmc-msft (approved):

sdk/storage_datalake/examples/filesystem.rs
rylev: They already do this. The `clone` here is redundant. I will delete it.

PR #1063, opened 2022-09-01
allow shared-key based request signing to fail rather than panic

### LINKED ISSUE #1044: SharedKeyAuthorizationPolicy implementation panics if account_key is not a well-formed HMac string
SharedKeyAuthorizationPolicy panics at https://github.com/Azure/azure-sdk-for-rust/blob/25538f43f8163a1faf31d8d89eb232c7c471730a/sdk/storage_datalake/src/authorization_policies/shared_key.rs#L70 if a given account_key is not a well-formed HMac string.

Implementation should not use simple `unwrap`, because the account key represented by `shared_key` is data coming from the end user, so it can be syntactically incorrect, therefore a better approach could be

1. to change `generate_authorization` function so it returns `PolicyResult`
2. to handle errors coming from the `sign` function and to translate them into corresponding `PolicyResult`

bmc-msft:
fixes #1044

### DIFF
--- a/sdk/storage_datalake/src/authorization_policies/shared_key.rs
+++ b/sdk/storage_datalake/src/authorization_policies/shared_key.rs
@@ -45,7 +45,7 @@ impl Policy for SharedKeyAuthorizationPolicy {
             request.method(),
             &self.credential.account_name,
             &self.credential.account_key,
-        );
+        )?;
 
         request.insert_header(headers::AUTHORIZATION, HeaderValue::from(auth));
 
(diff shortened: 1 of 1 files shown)


cataggar (approved):

sdk/storage_datalake/src/authorization_policies/shared_key.rs
```diff
@@ -61,16 +61,16 @@ fn generate_authorization(
     http_method: &Method,
     storage_account_name: &str,
     shared_key: &str,
-) -> String {
+) -> azure_core::Result<String> {
     let str_to_sign = string_to_sign(http_headers, url, http_method, storage_account_name);
 
     // println!("\nstr_to_sign == {:?}\n", str_to_sign);
```
rylev, line 67: Btw - we should remove all of this.

PR #1065, opened 2022-09-02
Add a conversion from `Arc<dyn TokenCredential` to `StorageCredentials`

rylev:
This conversion makes a lot of sense IMO.


gorzell (approved):
Our code would certainly be easier to read with this. Although you would need to change some of the client methods to take `Into<StorageCredentials>` in order to close the loop completely.

gorzell (approved):

cataggar (approved):

bmc-msft (approved):

PR #1066, opened 2022-09-02
Add a way to override all client options

rylev:
This can be very useful if the user has already created a `ClientOptions` that they want to reuse across many different clients.

### DIFF
--- a/sdk/data_cosmos/src/clients/cosmos.rs
+++ b/sdk/data_cosmos/src/clients/cosmos.rs
@@ -70,6 +70,13 @@ impl CosmosClientBuilder {
         self.options = self.options.transport(transport);
         self
     }
+
+    /// Override all of the client options.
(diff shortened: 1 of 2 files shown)


gorzell (approved):
In addition to allowing re-use, it also provides access to thing that would otherwise not possible to change.

It might be worth adding to the documentation that now that order in which you call builder methods matters. i.e. if you call `retry` and then `client_options`, your `retry` change is lost, but if you call `client_options` and then `retry` you have customized the client options.

rylev:
I've added a warning to the docs to make it clear to users that this overrides all client options previously set. 

As a side note: I think we'll eventually want to abstract over these builders as there is a lot of overlap between the cosmos and blob storage client builders.

gorzell (approved):

cataggar (approved):

PR #1077, opened 2022-09-05
Rename `ConnectionString::new` to ConnectionString::from_str`

yoshuawuyts:
This surfaced during a pairing session with @rylev. Originally it was suggested that this be named `parse`, but on further inspection `from_str` seems more appropriate (`parse` is canonically the method provided on the string, `from_str` on the type resulting from the parse).

We can't implement `std::str::FromStr` because we return a struct containing lifetimes, which is incompatible with the trait. So the next best thing is to implement an inherent method of the same name on the struct. Thanks!


yoshuawuyts:
Rust's Clippy linter seems to dislike this change. I disagree with Clippy, but not enough to say that we should make an exception to the rule.

Issue #1093, opened 2022-09-13
Blob storage fails with invalid content range with 0 byte blob


magicmaaaaan:
Version: f9ec73563f615553ca097b0457bed8347ad2227d

I'm trying to download a zero byte blob from blob storage using `BlobClient::get` and am running into errors. Note that this only happens when the blob is empty (0 bytes).

Here's a simple program that throws an error (similar to `stream_blob_01.rs`).

```rust
use azure_storage_blobs::prelude::*;
use futures::stream::StreamExt;

#[tokio::main]
async fn main() -> azure_core::Result<()> {
    let storage_client = ClientBuilder::emulator();
    let blob_client = storage_client.blob_client("test", "empty.txt");
    let mut stream = blob_client.get().into_stream();
    while let Some(res) = stream.next().await {
        println!("{:?}", res.unwrap());
    }
    Ok(())
}
```

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { context: Full(Custom { kind: DataConversion, error: Error { context: Custom(Custom { kind: DataConversion, error: ParseIntError { kind: Empty } }) } }, "unable to parse header 'HeaderName(\"content-range\"): HeaderValue(\"bytes 0--1/0\")' into azure_core::request_options::content_range::ContentRange") }', sdk/storage_blobs/examples/stream_blob_02.rs:10:30
```

`bytes 0--1/0` cannot be parsed as a `ContentRange` because the end range is -1. (I'm not sure if this is valid in the first place and may actually be an issue with Azurite.)

But I also tried with the actual blob storage service.

```rust
use azure_storage::StorageCredentials;
use azure_storage_blobs::prelude::*;
use futures::stream::StreamExt;

#[tokio::main]
async fn main() -> azure_core::Result<()> {
    let storage_credentials = StorageCredentials::Key("...".into(), "...".into());
    let storage_client = ClientBuilder::new("...", storage_credentials);
    let blob_client = storage_client.blob_client("test", "empty.txt");
    let mut stream = blob_client.get().into_stream();
    while let Some(res) = stream.next().await {
        println!("{:?}", res.unwrap());
    }
    Ok(())
}
```

```
[...]

PR #1094, opened 2022-09-15
work around invalid content-range from azurite

bmc-msft:
When requesting zero byte files from azurite, it can generate invalid content-range headers.  

See #1093 for an example.

### DIFF
--- a/sdk/core/src/request_options/content_range.rs
+++ b/sdk/core/src/request_options/content_range.rs
@@ -40,16 +40,6 @@ impl ContentRange {
 impl FromStr for ContentRange {
     type Err = Error;
     fn from_str(s: &str) -> crate::Result<ContentRange> {
-        if cfg!(featue = "azurite_workaround ") {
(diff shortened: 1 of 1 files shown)


cataggar (approved):
Thanks for chasing this down. It would be good to report this difference as an azurite bug and add a code comment with a link or in a follow-up issue.

bmc-msft:
There is an open work item that discusses this inconsistency.  https://github.com/Azure/Azurite/issues/1682

sdk/core/src/request_options/content_range.rs
```diff
@@ -49,6 +49,14 @@ impl FromStr for ContentRange {
             })
         })?;
 
+        if cfg!(feature = "azurite_workaround") && remaining == "0--1/0" {
```
rylev, line 52: Can we add a comment here? It would be nice to know *why* this code is necessary without having to have the associated issue in memory.

PR #1101, opened 2022-09-16
add TryFrom<Url> to Blob & Container clients

bmc-msft:
An common pattern is to generate, pass around, and use SAS URLs to blob resources.

This makes it easy to create Blob and Container clients from SAS URLs


rylev (approved):
Sorry for the delay! I have some nits, but I don't want to keep blocking this.

PR #1102, opened 2022-09-22
make IntoFuture broadly usable

bmc-msft:
With 1.64, IntoFuture is now stable.  Therefore, most of our calls to .into_future() are no longer needed.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.63.0"
+channel = "1.64.0"
 components = [ "rustfmt", "clippy" ]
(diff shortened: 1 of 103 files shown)


heaths (approved):
Based on a sampling of a dozen or so files, LGTM.

bmc-msft:
We pin the rust version in `rust-toolchain.toml`.   Currently, this is specified here:

https://github.com/Azure/azure-sdk-for-rust/blob/c5ed21f6e948b8040bc2a9c3cca04ebecf1194c6/rust-toolchain.toml#L2

cataggar (approved):

rylev:
@bmc-msft that only impacts people building directly from source. Users who take a dependency through crates.io or even GitHub will not be notified of this.

cataggar (approved):

PR #1104, opened 2022-09-22
Allow support to use vendored openssl

msabansal:
Added support to use the vendored version of openssl. This makes it easier to build an app with this crate where the openssl dependencies are not already installed

### DIFF
--- a/sdk/identity/Cargo.toml
+++ b/sdk/identity/Cargo.toml
@@ -47,6 +47,7 @@ enable_reqwest_rustls = [
 development = []
 test_e2e = []
 client_certificate = ["openssl"]
+vendored_openssl = ["openssl/vendored"]
 
 [[example]]
 name="client_certificate_credentials"


bmc-msft:
@msabansal , thanks for the contribution.

The current build failure is unrelated to your PR and was addressed in #1103.  Please rebase your PR to `main` so we can review it.

sdk/identity/Cargo.toml
```diff
@@ -47,6 +47,7 @@ enable_reqwest_rustls = [
 development = []
 test_e2e = []
 client_certificate = ["openssl"]
+vendored-openssl = ["openssl/vendored"]
```
cataggar, line 50: All other features use `_`.

msabansal: i didn't put a lot of thought into this. Almost every other library i have used uses vendored-openssl for this intent. I will change it to vendored_openssl for now to follow convention of this library

bmc-msft (approved):

PR #1108, opened 2022-09-27
analysis of specifications using `Cancelled` as an enum value

cataggar:
This is analysis for https://github.com/Azure/azure-resource-manager-rpc/issues/144 .

### DIFF
--- a/services/autorust/codegen/examples/cancelled.rs
+++ b/services/autorust/codegen/examples/cancelled.rs
@@ -7,26 +7,26 @@ use autorust_codegen::{get_mgmt_readmes, get_svc_readmes, io, Result, Spec, Spec
 use std::collections::BTreeSet;
 
 fn main() -> Result<()> {
-    println!("CONTROL PLANE");
-    check(&get_mgmt_readmes()?)?;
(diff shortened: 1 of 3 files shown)


heaths:
I'm not clear on what the intent is, but the [Azure guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/ConsiderationsForServiceDesign.md#long-running-operations) dictate the following values:

* notStarted
* running
* succeeded
* failed
* canceled

Technically, that spec still has them as PascalCase, but more recent conversations have us recommending camelCase for enum values. Whatever the case, the public API would be using snake_case anyway, but we should probably handle these values case-insensitively as well as "cancelled" == "canceled" like we do in other languages.

I'll ping @JeffreyRichter and others offline about the casing of these values as we recommended them here, but the case-insensitive text is still correct.

cataggar:
I used this example to locate the 8 services that are using `Cancelled` in `provisioningState`, when they should be using `Canceled`.  This includes the vmware service, which I manage. I still think this example is worth merging. It shows how to loop through the models for analysis.

bmc-msft (approved):

heaths (approved):

PR #1114, opened 2022-10-06
Exposing public queue api data

rguerreiromsft:
**What**
Making Message, PopReceipt, message_id, and pop_receipt public available.

**Why**
Implementing saga design pattern in an internal Microsoft project where a different node can ack or nack a message.

I believe all of these values must be public because one could use the Rest API and retrieve them, so they aren't necessarily secrets to the SDK user.

It also makes it easier to create abstractions that work with a Storage Account Queue under the hood.

Lastly, although not implemented in this PR, Message could benefit from being `Serialize` as well.

### DIFF
--- a/sdk/storage_queues/src/operations/get_messages.rs
+++ b/sdk/storage_queues/src/operations/get_messages.rs
@@ -40,8 +40,8 @@ pub struct GetMessagesResponse {
 #[derive(Debug, Clone, Deserialize)]
 #[serde(rename_all = "PascalCase")]
 pub struct Message {
-    message_id: String,
-    pop_receipt: String,
+    pub message_id: String,
+    pub pop_receipt: String,
     #[serde(with = "azure_core::date::rfc1123")]
     pub insertion_time: OffsetDateTime,
     #[serde(with = "azure_core::date::rfc1123")]
--- a/sdk/storage_queues/src/operations/mod.rs
+++ b/sdk/storage_queues/src/operations/mod.rs
@@ -19,6 +19,7 @@ pub use clear_messages::ClearMessagesBuilder;
 pub use create_queue::CreateQueueBuilder;
 pub use delete_message::DeleteMessageBuilder;
 pub use delete_queue::DeleteQueueBuilder;
+pub use get_messages::Message;
 pub use get_messages::GetMessagesBuilder;
 pub use get_queue_acl::GetQueueACLBuilder;
 pub use get_queue_metadata::GetQueueMetadataBuilder;
--- a/sdk/storage_queues/src/pop_receipt.rs
+++ b/sdk/storage_queues/src/pop_receipt.rs
@@ -6,7 +6,7 @@ pub struct PopReceipt {
 }
 
 impl PopReceipt {
-    pub(crate) fn new(message_id: impl Into<String>, pop_receipt: impl Into<String>) -> Self {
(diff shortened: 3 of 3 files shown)


rylev (changes requested):
Thanks! This seems reasonable. Looks like you need to run `cargo fmt` in order for CI to pass.

rguerreiromsft:
Hi @rylev, should I also make Message (and probably PopReceipt) `Serialize` and `Deserialize`?

sdk/storage_queues/src/operations/mod.rs
```diff
@@ -20,6 +20,7 @@ pub use create_queue::CreateQueueBuilder;
 pub use delete_message::DeleteMessageBuilder;
 pub use delete_queue::DeleteQueueBuilder;
 pub use get_messages::GetMessagesBuilder;
+pub use get_messages::Message;
```
rylev, line 23:
```suggestion
pub use get_messages::{GetMessagesBuilder, Message};
```

sdk/storage_queues/src/operations/get_messages.rs
```diff
@@ -40,8 +40,8 @@ pub struct GetMessagesResponse {
 #[derive(Debug, Clone, Deserialize)]
 #[serde(rename_all = "PascalCase")]
 pub struct Message {
-    message_id: String,
-    pop_receipt: String,
+    pub message_id: String,
+    pub pop_receipt: String,
```
rylev, line 44: Actually now I'm confused why you would want to use these fields directly instead of using the `pop_receipt` method below. Can you clarify? If there's a good answer, we should document it on the `pop_receipt` method.

rylev (changes requested):
I just want to make sure I understand what the need is here. These changes (plus implementing `Serialize`) all seem reasonable, but we just need to make sure it's clear when the user would use the individual fields vs the `pop_receipt` method.

rylev:
@rguerreiromsft ping. Would you be able to get this over the finish line?

sdk/storage_queues/src/operations/get_messages.rs
rguerreiromsft: Both are acceptable, but before this PR, even PopReceipt didn't offer access as well. Now I do worry about cloning the strings when we need this access.

Although I think it's fairly cheap to clone both of them, but I think it has a potential to be costly.

To avoid mutations, we can add methods to expose these by reference, like the PopReceipt.

rguerreiromsft:
Sorry, I was on vacation without access to my computer. I'm working on it now.

rguerreiromsft:
> @rguerreiromsft please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> 
> Contributor License Agreement

rylev (approved):

PR #1115, opened 2022-10-08
add `fn deserialize_null_as_default` for deserializing JSON null values

### LINKED ISSUE #1113: azure_mgmt_web: Error("invalid type: null, expected a sequence", line: 1, column: 994)
Hello,

I try to get a web app configuration in JSON but still got the error:
```
Error: Error { context: Custom(Custom { kind: DataConversion, error: Error("invalid type: null, expected a sequence", line: 1, column: 994) }) }
```

Code to reproduce:
```rust
use azure_identity::AzureCliCredential;
use azure_mgmt_web::ClientBuilder;
use serde_json::{to_string_pretty, Value};
use std::{error::Error, sync::Arc};

#[tokio::main]
async fn main()-> Result<(), Box<dyn Error>> {
    let resource_group = "";
    let webapp_name = "";
    let sub_id = "";

    let creds = Arc::new(AzureCliCredential::default());
    let client = ClientBuilder::new(creds).build();

    let get = client.web_apps_client().get_configuration(resource_group, webapp_name, sub_id).into_future().await?;
    let data = to_string_pretty(&get.properties.unwrap())?;
    println!("{:?}", data);

    Ok(())
}
```
---
I also tried to convert to Value as an intermediate step and got another error:
```
the trait bound `Value: From<SiteConfig>` is not satisfied
```

Code to reproduce:
```rust
use azure_identity::AzureCliCredential;
use azure_mgmt_web::ClientBuilder;
use serde_json::{to_string_pretty, Value, to_vec_pretty};
use std::{error::Error, sync::Arc};

#[tokio::main]
async fn main()-> Result<(), Box<dyn Error>> {
    let resource_group = "";
    let webapp_name = "";
    let sub_id = "";

    let creds = Arc::new(AzureCliCredential::default());
    let client = ClientBuilder::new(creds).build();

    let config = client.web_apps_client().get_configuration(resource_group, webapp_name, sub_id).into_future().await?;
    let config_json: Value = config.properties.unwrap().into();
    println!("{:?}", to_string_pretty(&config_json).unwrap());

    Ok(())
}
```

ctaggart:
Fix #1113. This adds the workaround from https://github.com/serde-rs/serde/issues/1098. I added unit tests and an example related to the fixed issue. No Azure services are supposed to return a JSON null when it is an array. This protects against those that do.

check all: https://github.com/ctaggart/azure-sdk-for-rust/actions/runs/3208461860


sdk/core/src/util.rs
```diff
@@ -15,3 +15,14 @@ where
         .or_else(|_| T::deserialize(serde_json::Value::String(v.to_lowercase())))
         .map_err(de::Error::custom)
 }
+
+/// Deserialize JSON null as default
+/// https://github.com/serde-rs/serde/issues/1098
+pub fn deserialize_null_default<'de, D, T>(deserializer: D) -> Result<T, D::Error>
```
rylev, line 21: Nit: `deserialize_null_as_default` is ever so slightly clearer IMO

rylev (approved):
Looks good. Had a small thought on naming, but it's not a blocking concern.

PR #1125, opened 2022-10-17
Add messages count to get-queue-metadata operation

### LINKED ISSUE #1118: Value of `x-ms-approximate-messages-count` header not returned in `queue.get_metadata()` result
It looks like [QueueClient::get_metadata](https://github.com/Azure/azure-sdk-for-rust/blob/fb8fe611f3406d577434f5cf1a398b1d02fa6802/sdk/storage_queues/src/operations/get_queue_metadata.rs#L42) returns parsed headers in `common_storage_response_headers` which [doesn't parse](https://github.com/Azure/azure-sdk-for-rust/blob/fb8fe611f3406d577434f5cf1a398b1d02fa6802/sdk/storage/src/headers/mod.rs#L23) the value of [x-ms-approximate-messages-count](https://learn.microsoft.com/en-us/rest/api/storageservices/get-queue-metadata#response-headers). We currently need access to this property for crucial infra monitoring and scaling purposes. Is there a way to access this value or the raw headers directly?

rylev:
Fixes #1118

### DIFF
--- a/sdk/storage_queues/src/operations/get_queue_metadata.rs
+++ b/sdk/storage_queues/src/operations/get_queue_metadata.rs
@@ -1,5 +1,10 @@
 use crate::clients::QueueClient;
-use azure_core::{error::Error, headers::Headers, prelude::*, Method, Response as AzureResponse};
+use azure_core::{
+    error::Error,
+    headers::{HeaderName, Headers},
+    prelude::*,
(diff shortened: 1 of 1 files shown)


bmc-msft (approved):

wilsonzlin:
Thanks @rylev!

PR #1133, opened 2022-10-26
handle removing crates when all tags have been disabled

bmc-msft:
In the case of agrifood, generating any of the tags is currently broken.

Rather than needing to do a manual cleanup after trying to generate a tag that's broken, this does the cleanup as part of gen_crate.


cataggar (approved):
thanks!

Issue #1146, opened 2022-11-06
Set KeyVault secret is missing Content-Type header attribute


pavelm10:
When trying to create new secret using `azure_security_keyvault` crate the following error is shown:
```
Error: Error { context: Full(Custom { kind: HttpResponse { status: UnsupportedMediaType, error_code: None }, error: HttpError { status: UnsupportedMediaType, details: ErrorDetails { code: None, message: None }, headers: {"content-length": "164", "x-ms-request-id": "a080e13b-1e39-42ac-8894-698e7bb96e49", "date": "Sun, 06 Nov 2022 10:24:48 GMT", "content-type": "application/json; charset=utf-8", "x-ms-keyvault-service-version": "1.9.576.1", "x-ms-keyvault-region": "westeurope", "x-content-type-options": "nosniff", "cache-control": "no-cache", "expires": "-1", "pragma": "no-cache", "x-ms-keyvault-network-info": "conn_type=Ipv4;addr=86.49.229.122;act_addr_fam=InterNetwork;", "strict-transport-security": "max-age=31536000;includeSubDomains"}, body: b"{\"Message\":\"The request contains an entity body but no Content-Type header. The inferred media type 'application/octet-stream' is not supported for this resource.\"}" } }, "server returned error status which will not be retried: 415") }
```
In the `SetSecretBuilder.into_future()` method the provided body is in JSON format hence the `Content-Type` header must be set to `application/json`.

PR #1147, opened 2022-11-06
Add Content-Type to finalize request in KV client

pavelm10:
Fixes https://github.com/Azure/azure-sdk-for-rust/issues/1146

### DIFF
--- a/sdk/security_keyvault/src/clients/keyvault_client.rs
+++ b/sdk/security_keyvault/src/clients/keyvault_client.rs
@@ -4,6 +4,7 @@ use azure_core::{
     date,
     error::{Error, ErrorKind},
     headers::*,
+    prelude::*,
     Body, Context, Method, Pipeline, Request, Response,
 };
 use const_format::formatcp;
@@ -84,6 +85,9 @@ impl KeyvaultClient {
         request.insert_header(MS_DATE, time);
 
         if let Some(request_body) = request_body {
+            if request.headers().get_optional_str(&CONTENT_TYPE).is_none() {
+                request.insert_headers(&ContentType::APPLICATION_JSON);
+            }
             request.insert_header(CONTENT_LENGTH, request_body.len().to_string());
             request.set_body(request_body);
         } else {


sdk/security_keyvault/src/secrets/operations/set_secret.rs
```diff
@@ -21,7 +21,8 @@ impl SetSecretBuilder {
 
             let body = Some(Value::Object(request_body).to_string().into());
 
-            let headers = Headers::new();
+            let mut headers = Headers::new();
+            headers.insert(CONTENT_TYPE, "application/json");
```
bmc-msft, line 25: This appears to be a larger problem than just SetSecretBuilder.  Setting the content type was mistakenly dropped in https://github.com/Azure/azure-sdk-for-rust/pull/962.

Instead of updating just SetSecretBuilder, can you instead update `finalize_request`?

Perhaps something like this:
```diff
--- a/sdk/security_keyvault/src/clients/keyvault_client.rs
+++ b/sdk/security_keyvault/src/clients/keyvault_client.rs
@@ -4,7 +4,8 @@ use azure_core::{
     date,
     error::{Error, ErrorKind},
     headers::*,
-    Body, Context, Method, Pipeline, Request, Response,
+    prelude::*,
+    Body, Method, Pipeline, Request, Response,
 };
 use const_format::formatcp;
 use std::sync::Arc;
@@ -84,6 +85,9 @@ impl KeyvaultClient {
         request.insert_header(MS_DATE, time);

         if let Some(request_body) = request_body {
+            if request.headers().get_optional_str(&CONTENT_TYPE).is_none() {
+                request.insert_headers(&ContentType::APPLICATION_JSON);
+            }
             request.insert_header(CONTENT_LENGTH, request_body.len().to_string());
             request.set_body(request_body);
         } else {
```

bmc-msft (changes requested):

sdk/security_keyvault/src/secrets/operations/set_secret.rs
bmc-msft: Note, if you'd prefer, let me know and I can make the broader change.

Thanks!

pavelm10: I can do it, I was just thinking that the other operations (get, delete, get_versions) are working without the `Content-Type` header, hence putting it into the common `finalize_request` method seemed incorrect to me.

pavelm10: but your suggestion make sense that for request with body it should be added, in case in the future there are other operation which need to set the header. I will change it according to your suggestion.

bmc-msft: get & delete do not set request bodies, which makes this less of an issue.

bmc-msft (approved):

bmc-msft:
Thanks for the PR!

pavelm10:
No problem

Issue #1154, opened 2022-11-22
Add Service Bus Topic Functionality


pavelm10:
Currently, the SDK for `message_service` bus supports sending/receiving messages to/from a `queue`. Adding support for handling messages with a `topic` would be beneficial.

PR #1155, opened 2022-11-22
add service bus topic support

pavelm10:
Implements https://github.com/Azure/azure-sdk-for-rust/issues/1154


rylev (approved):
This seems reasonable to me. We need to switch the servicebus crate over to using the pipeline architecture that the other SDKs are using which would help with adding tests.

PR #1157, opened 2022-11-24
Add .bytes() fn to BlockId

juliusl:
Currently there is not a way to convert `BlockId` into `Bytes`. Constructing a `Bytes` from `AsRef<&[u8]>` requires an allocation (`Bytes::from(.as_ref().to_vec())`). This PR adds a function that returns a clone of `BlockId`'s inner `Bytes`, so that the zero-copy behavior of `Bytes` can be preserved.

### DIFF
--- a/sdk/storage_blobs/src/options/block_id.rs
+++ b/sdk/storage_blobs/src/options/block_id.rs
@@ -5,8 +5,16 @@ use bytes::Bytes;
 pub struct BlockId(Bytes);
 
 impl BlockId {
+    /// Returns a new block id,
+    ///
     pub fn new(block_id: impl Into<Bytes>) -> Self {
         Self(block_id.into())
+    }
+
+    /// Returns clone of bytes,
+    ///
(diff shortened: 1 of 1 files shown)


sdk/storage_blobs/src/options/block_id.rs
```diff
@@ -1,13 +1,25 @@
 use azure_core::AppendToUrlQuery;
 use bytes::Bytes;
 
+/// Struct wrapping the bytes of a block blob block-id,
+///
+/// A block id cannot exceed 64 bytes. In addition all block id's in a block list must be the same length when converted into a base64 string.
```
bmc-msft, line 6: Including these restrictions in the documentation is a good idea.  Can we link to a citation on learn.microsoft.com for this?

bmc-msft (approved):

sdk/storage_blobs/src/options/block_id.rs
juliusl: Added  a link to documentation

PR #1159, opened 2022-11-29
fix readme generation

bmc-msft:
This PR changes the following
* Fixes checking if individual crate README.md files are up-to-date
* Includes regenerated README.md crates
* Updates a handful of the lib.rs to include a basic example and updated text

### DIFF
--- a/sdk/core/README.md
+++ b/sdk/core/README.md
@@ -1,3 +1,13 @@
-# azure_core crate
+# azure_core
 
-The azure_core crate is an [Azure SDK for Rust](https://github.com/Azure/azure-sdk-for-rust) crate. It is a library that provides cross-cutting services to other client libraries. Please see the [general guidelines](https://azure.github.io/azure-sdk/general_azurecore.html). For changes in this version, see the [CHANGELOG.md](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/core/CHANGELOG.md). For open issues, filter with the [core label](https://github.com/Azure/azure-sdk-for-rust/issues?q=is%3Aopen+is%3Aissue+label%3Acore).
+Core types and traits for the Rust Azure SDK.
+
+This crate is part of the unofficial Azure SDK effort in Rust. For more
+information on the project, and an overview of other crates, please refer to
+[our GitHub repository](https://github.com/azure/azure-sdk-for-rust).
+
+It is a library that provides cross-cutting services to other client
+libraries.  Please see the [general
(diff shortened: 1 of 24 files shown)


sdk/core/README.md
```diff
@@ -1,3 +1,13 @@
-# azure_core crate
+# azure_core
 
-The azure_core crate is an [Azure SDK for Rust](https://github.com/Azure/azure-sdk-for-rust) crate. It is a library that provides cross-cutting services to other client libraries. Please see the [general guidelines](https://azure.github.io/azure-sdk/general_azurecore.html). For changes in this version, see the [CHANGELOG.md](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/core/CHANGELOG.md). For open issues, filter with the [core label](https://github.com/Azure/azure-sdk-for-rust/issues?q=is%3Aopen+is%3Aissue+label%3Acore).
+Core types and traits for the Rust Azure SDK.
+
+This crate is part of the unofficial Azure SDK effort in Rust. For more
+information on the project, and an overview of other crates, please refer to
```
heaths, line 6: Nit: unnecessary comma.
```suggestion
information on the project and an overview of other crates, please refer to
```

sdk/core/src/lib.rs
```diff
@@ -3,6 +3,10 @@
 //! This crate is part of the unofficial Azure SDK effort in Rust. For more
 //! information on the project, and an overview of other crates, please refer to
 //! [our GitHub repository](https://github.com/azure/azure-sdk-for-rust).
+//!
+//! It is a library that provides cross-cutting services to other client
+//! libraries.  Please see the [general
+//! guidelines](https://azure.github.io/azure-sdk/general_azurecore.html).
```
heaths, line 9: Why not just use `#![include_str!("../README.md")]`? Or is the assumption that the README may grow (for officialness, it mostly likely will have to) and we don't necessarily want this to grow the same?

heaths (approved):

sdk/core/src/lib.rs
bmc-msft: If we look at the README used in other SDKs, they include things we don't need as part of crate documentation but are entirely reasonable for the git repo, such release processes (how we define preview, ga, etc), contribution guidelines, security reporting guidelines, code of conduct for the repo, etc.

heaths (approved):
LGTM, so long as you got other such instances, if any, in other READMEs.

PR #1161, opened 2022-11-29
unify the code used to generate svc and mgmt crates

bmc-msft:


rylev (approved):
One small nit but otherwise looks good!

cataggar (approved):

PR #1162, opened 2022-11-29
generate service crates

bmc-msft:
generated from azure-rest-api-specs [commit from 2022-11-29](https://github.com/Azure/azure-rest-api-specs/d85953a9cb1b464824fae7c1fd300a99d086c8e4)


services/svc/cosmosdb/autorust.toml
```diff
@@ -0,0 +1,4 @@
+[tags]
+deny = [
+  "package-2019-02"
```
bmc-msft, line 3: Per discussion on teams, the current spec for this tag causes generation to fail with the error:

```
8 cosmos-db  package-2019-02Error: Error { context: Custom(Custom { kind: DataConversion, error: Error("data did not match any variant of untagged enum ReferenceOr", line: 1209, column: 5) }) }
```

cataggar (approved):

Issue #1156, opened 2022-11-23
Azure SDK should automatically set endpoints based on the cloud it is in


archoversight:
Hey,

We are using Vector (https://vector.dev/) for logging in Azure USGov Virginia and Azure USGov Arizona and unfortunately the Azure SDK for Rust makes many assumptions that is running entirely within the Public cloud. There are many endpoints and services that have a different URL and the defaults do not work. It would makes sense to either pick up the value automatically from `~/.azure/config` or the environment variable `AZURE_CLOUD_NAME`.

This way when creating clients for things like blob storage it would automatically pick the correct URL.

Please see:

https://learn.microsoft.com/en-us/azure/azure-government/compare-azure-government-global-azure#guidance-for-developers

for information on all the endpoints that change and the appropriate URL's to match them to.

Thanks,
Bert JW Regeer

juliusl:
Probably could add something for that here, https://github.com/Azure/azure-sdk-for-rust/blob/3de510ce055be3cdc65b3b46783d4f6d97db4fe7/sdk/storage/src/cloud_location.rs#L9 ?

rylev:
As @juliusl we should add support for this to the `CloudLocation` enum which then can be passed to the `BlobServiceClientBuilder` [here](https://github.com/Azure/azure-sdk-for-rust/blob/3de510ce055be3cdc65b3b46783d4f6d97db4fe7/sdk/storage_blobs/src/clients/blob_service_client.rs#L35).

juliusl:
@archoversight Where is `AZURE_CLOUD_NAME` documented?

archoversight:
@juliusl it is what I see most commonly used by various other companies when they are scripting support for alternate Azure locations.

I don't think it is explicitly documented anywhere by Microsoft. When it comes to AWS, we use `AWS_REGION` with similar effect.

For example `us-gov-west-1` and `us-east-2` are valid AWS regions that direct the API calls to the right locations.

archoversight:
Also adding the enum would not allow for automatic detection. Clients using the code would need to add detection for the appropriate cloud.

Vector for example requires:

- account name
- container name

When provided the account name it uses the default credential finder:

- EnvironmentCredential
- ManagedIdentityCredential
- AzureCliCredential

But those credentials are not searched for until after an attempt is made to use Object Storage, at which point the default location is used (AzureCloud) while the credentials are for AzureUSGovernment and the URL for object storage is wrong.

[...]

Issue #1127, opened 2022-10-17
converting Url to CloudLocation only works for Public & China clouds


bmc-msft:
In PR #1101, only China & Public clouds are supported when converting SAS URLs to CloudLocations

PR #1166, opened 2022-11-30
Auto-detect CloudLocation for Storage

juliusl:
Addresses #1156 and #1127

### DIFF
--- a/sdk/storage/src/cloud_location.rs
+++ b/sdk/storage/src/cloud_location.rs
@@ -227,7 +227,8 @@ impl CloudLocation {
 impl TryFrom<&Url> for CloudLocation {
     type Error = azure_core::Error;
 
-    // TODO: This only works for Public and China clouds.
+    // TODO: Only supports Public, China, USGov, and Emulator
+    // Is CustomURL required?
     // ref: https://github.com/Azure/azure-sdk-for-rust/issues/502
     fn try_from(url: &Url) -> azure_core::Result<Self> {
         let token = url.query().ok_or_else(|| {
@@ -269,11 +270,34 @@ impl TryFrom<&Url> for CloudLocation {
                 account,
                 credentials,
             }),
+            "core.usgovcloudapi.net" => Ok(CloudLocation::USGov {
+                account,
+                credentials,
+            }),
+            _ if url.path().trim_matches('/') == EMULATOR_ACCOUNT
+                && url.has_host()
+                && url.port().is_some() =>
+            {
+                if let Some(host) = url.host() {
+                    match host {
+                        url::Host::Ipv4(ip) => Ok(CloudLocation::Emulator {
+                            address: format!("{ip}"),
+                            port: url.port().expect("should have a port"),
+                        }),
+                        _ => Err(azure_core::Error::with_message(
+                            azure_core::error::ErrorKind::DataConversion,
+                            || format!("Unsupported emulator URL, expected ipv4: {}", host),
+                        )),
+                    }
+                } else {
+                    unreachable!()
+                }
+            }
             _ => Err(azure_core::Error::with_message(
                 azure_core::error::ErrorKind::DataConversion,
                 || {
                     format!(
-                        "URL refers to a domain that is not a Public or China domain: {}",
+                        "URL refers to a domain that is not a Emulator, Public, China, or USGov domain: {}",
                         host
                     )
                 },
@@ -328,6 +352,26 @@ mod tests {
         let cloud_location: CloudLocation = (&china_cloud).try_into()?;
         assert_eq!(
             china_cloud_without_token,
+            cloud_location.url(ServiceType::Blob)?
+        );
+
+        let us_gov_cloud = Url::parse("https://test.blob.core.usgovcloudapi.net/?token=1")?;
+        let us_gov_cloud_without_token = Url::parse("https://test.blob.core.usgovcloudapi.net")?;
+
+        let cloud_location: CloudLocation = (&us_gov_cloud).try_into()?;
+        assert_eq!(
+            us_gov_cloud_without_token,
+            cloud_location.url(ServiceType::Blob)?
+        );
+
+        let emulator =
+            Url::parse(format!("http://127.0.0.1:5555/{EMULATOR_ACCOUNT}/?token=1").as_str())?;
+        let emulator_without_token =
+            Url::parse(format!("http://127.0.0.1:5555/{EMULATOR_ACCOUNT}").as_str())?;
+
+        let cloud_location: CloudLocation = (&emulator).try_into()?;
+        assert_eq!(
+            emulator_without_token,
             cloud_location.url(ServiceType::Blob)?
         );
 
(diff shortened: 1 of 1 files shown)


sdk/storage/src/cloud_location.rs
```diff
@@ -44,20 +92,115 @@ impl CloudLocation {
...
+                            account: account.clone(),
+                            credentials: credentials.clone(),
+                        }
+                        .url(service_type),
+                        _ => {
+                            todo!()
+                        }
+                    };
+                } else {
+                    // Default to PROD
```
juliusl, line 147: Should we panic here instead to fail fast?

sdk/storage/src/cloud_location.rs
```diff
@@ -27,6 +55,26 @@ pub enum CloudLocation {
...
+    /// If either of these methods fail, this will default to public azure.
+    ///
+    /// Cloud names can be listed with, `az cloud list --output table`. Current public values are:
+    /// - AzureCloud - Public azure, in some cases may be AzurePublicCloud in env var form
+    /// - AzureChinaCloud
+    /// - AzureUSGovernment
+    /// - AzureGermanCloud - Might be deprecating based on public documentation but still shows up in `az cloud list`
+    ///
+    pub fn auto_detect(account: impl AsRef<str>, credentials: StorageCredentials) -> CloudLocation {
+        CloudLocation::AutoDetect {
```
juliusl, line 72: Alternatively, could add a fallback field, to handle test/custom clouds?

sdk/storage/src/cloud_location.rs
```diff
@@ -107,11 +270,34 @@ impl TryFrom<&Url> for CloudLocation {
...
+                && url.has_host()
+                && url.port().is_some() =>
+            {
+                if let Some(host) = url.host() {
+                    match host {
+                        url::Host::Ipv4(ip) => Ok(CloudLocation::Emulator {
+                            address: format!("{ip}"),
+                            port: url.port().expect("should have a port"),
+                        }),
+                        _ => Err(azure_core::Error::with_message(
```
juliusl, line 287: This is an assumption, but could as well just pass the other cases as well

sdk/storage/src/cloud_location.rs
```diff
@@ -44,28 +108,127 @@ impl CloudLocation {
...
+                            credentials: credentials.clone(),
+                        }
+                        .url(service_type),
+                        AZURE_GERMAN_CLOUD => CloudLocation::GermanCloud {
+                            account: account.clone(),
+                            credentials: credentials.clone(),
+                        }
+                        .url(service_type),
+                        _ => {
+                            todo!()
```
rylev, line 160: Return `Err`?

sdk/storage/src/cloud_location.rs
```diff
@@ -44,28 +108,127 @@ impl CloudLocation {
...
+                .join(".azure/config")
+                .canonicalize()
+                .ok()
+                .and_then(|config| File::open(config).ok())
+            {
+                let mut lines = BufReader::new(config).lines();
+
+                while let Some(Ok(line)) = lines.next() {
+                    // Alternatively, import serde_toml for parsing this file, but probably better off doing
+                    // that by creating a struct dedicated to managing the config file
```
rylev, line 208: we should either do this or remove this comment. I'm unsure if the dependency is worth it if we can keep things this short.

sdk/storage/src/cloud_location.rs
```diff
@@ -44,28 +108,127 @@ impl CloudLocation {
...
+                    if line.trim() == "[cloud]" {
+                        if let Some(Ok(name)) = lines.next() {
```
rylev, line 210: Is the next line allowed to be empty?

sdk/storage/src/cloud_location.rs
```diff
@@ -169,6 +358,184 @@ mod tests {
...
+                .url(ServiceType::Blob)
+                .expect("should return a url")
+                .as_str(),
+            "https://test_account.blob.core.cloudapi.de/"
+        );
+
+        std::env::remove_var("AZURE_CLOUD_NAME");
+
+        std::env::set_var("HOME", ".test");
+        std::fs::create_dir_all(".test/.azure").expect("should be able to make test dir");
```
rylev, line 450: Should we be writing these into a temp folder?

rylev (changes requested):

sdk/storage/src/cloud_location.rs
juliusl: I'm not sure, I think this file is either set by az cli, or when Azure Compute initializes the VM

juliusl:
@rylev Addressed all comments

sdk/storage/src/cloud_location.rs
juliusl: Ok I think I understand what you might be concerned about, added some more resiliency.

sdk/storage/src/cloud_location.rs
```diff
@@ -44,28 +88,123 @@ impl CloudLocation {
...
+            | CloudLocation::AutoDetect { credentials, .. } => credentials,
             CloudLocation::Emulator { .. } => &EMULATOR_CREDENTIALS,
-            CloudLocation::Custom { credentials, .. } => credentials,
+        }
+    }
+
+    /// Finds the cloud name, first by environment variable, then by parsing the current user's $HOME/.azure/config file
+    ///
+    fn find_cloud_name() -> Option<String> {
+        if let Ok(name) = std::env::var("AZURE_CLOUD_NAME") {
```
cataggar, line 167: Is `AZURE_CLOUD_NAME` used by any existing Azure SDKs? https://github.com/Azure/azure-sdk-for-net/search?q=AZURE_CLOUD_NAME comes up empty.

sdk/storage/src/cloud_location.rs
```diff
@@ -44,28 +88,123 @@ impl CloudLocation {
...
+            Some(name)
+        } else if let Ok(home_dir) = std::env::var("HOME") {
+            if let Some(config) = PathBuf::from(home_dir)
+                .join(".azure/config")
```
cataggar, line 171: I'm not sure the storage crate should be reading `$HOME/.azure/config`. Is something like this done in the other Azure SDKs?

cataggar (changes requested):

sdk/storage/src/cloud_location.rs
juliusl: I couldn't find anything either, but apparently, it's used by other libraries for cloud detection. I actually think this env variable might make more sense, https://github.com/Azure/azure-sdk-for-net/blob/eca8faf3c868bbb75ed8a104661bc77f5a976c9a/sdk/identity/Azure.Identity/src/EnvironmentVariables.cs#L30

sdk/storage/src/cloud_location.rs
juliusl: Can't find any other usage besides the cli repos. It appears the storage sdk for .net handles this by using an 'EndpointSuffix=' arg in the connection string. Is there a possible middle ground of moving this to a shared location or what are your thoughts?

cataggar:
I think we should follow the lead of the other official Azure SDKs here.

PR #1171, opened 2022-12-09
address concurrent issue with sleep future 

### LINKED ISSUE #1170: Managed Identity Timing Out
**Problem**
I am getting the following error when I run my rust functions app in the cloud. I am using the version of the identity crate on git as I am using a feature that hasn't been released yet #1168. 
```
Multiple errors were encountered while attempting to authenticate. 
error getting environment credential 
getting managed identity credential timed out
error getting token credential from Azure CLI
```
I noticed that in one of the recent pull request #1016 a timeout was added for managed identity when using the DefaultAzureCredentialBuilder. I tried changing the timeout from 1 second to 5 seconds, but it didn't help. 

**Workaround**
I fixed the issue by not using the DefaultAzureCredentialBuilder on the server. 

```rust
fn azure_identity() -> impl TokenCredential {
    #[cfg(debug_assertions)]
    return azure_identity::DefaultAzureCredential::default();

    #[cfg(not(debug_assertions))]
    return azure_identity::ImdsManagedIdentityCredential::default();
}
```

demoray:
As is, the Sleep future relies on the future not getting polled until the waker files, which isn't guaranteed.

This changes the implementation to check if the thread is done.  The waker still triggers the polling to occur.

fixes #1170

### DIFF
--- a/sdk/core/src/sleep.rs
+++ b/sdk/core/src/sleep.rs
@@ -21,7 +21,13 @@ impl Future for Sleep {
     type Output = ();
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        if self.thread.is_none() {
+        if let Some(thread) = &self.thread {
+            if thread.is_finished() {
+                Poll::Ready(())
+            } else {
+                Poll::Pending
+            }
+        } else {
             let waker = cx.waker().clone();
             let duration = self.duration;
             self.get_mut().thread = Some(thread::spawn(move || {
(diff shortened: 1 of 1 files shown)


cataggar (approved):
I'm not super familiar, but looks good to me. May be @rylev or @yoshuawuyts can review.

johnbatty:
Looks good to me too.

You can see that in the original code that the sleep timer is started on the first poll, and if a second poll is made before the timer has completed it always returned `Poll::Ready`, which is incorrect.  The fix correctly handles any polls after the timer has been started by checking for timer completion before returning `Poll::Ready`.

PR #1181, opened 2022-12-16
regenerate crates for 2022-12 release

demoray:
This is based on Azure/azure-rest-api-specs@fd296f4.  Note this is built upon #1180 .   To review this on it's own, check the second commit in this PR.


cataggar (approved):
Looks good. I see two new service packages.

PR #1183, opened 2022-12-19
Support hierarchical partition keys

thed24:
This method of extension should be backwards compatible with all existing usages of the trait whilst allowing users to opt into using the preview feature correctly.


thed24:
@microsoft-github-policy-service agree [company="no company"]

thed24:

thed24:

sdk/data_cosmos/src/cosmos_entity.rs
```diff
@@ -44,11 +44,20 @@ impl CosmosEntity for serde_json::Value {
...
+    // if pk is array, we dont need to wrap it in an array
+    // if it is not, we need to wrap it in an array
+
+    let serialized_string = serde_json::to_string(pk).context(
         azure_core::error::ErrorKind::DataConversion,
         "could not convert partition_key into String",
-    )
+    )?;
+
+    if serialized_string.starts_with('[') {
```
cataggar, line 56: Although this will work, can this be implemented without checking the serialization output?

cataggar (commented):
Can a test be added?

sdk/data_cosmos/src/cosmos_entity.rs
thed24: I tried changing some of the existing tests, although they relied on predefined `result` data that was asserted against so I left it be. Perhaps I can write some in-file tests?

The only way I could imagine achieving this without checking the serialized text is matching on the implementation of the generic type or something along those lines. This is method is more coupled, although safer as it works for all vector-like types or anything that serializes down to an array in JSON.

cataggar:
I opened #1410, but closing this as it is stale.

Issue #643, opened 2022-01-26
need AMQP 1.0 crate for Service Bus


cataggar:
There does not appear to be an AMQP 1.0 spec compliant crate. Crates like https://github.com/amqp-rs/lapin implement 0.9.1, but Service Bus requires 1.0. https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-amqp-overview

andrei-ionescu:
What about Dove?

- https://github.com/lulf/dove
- https://crates.io/crates/dove

cataggar:
There is also:
https://github.com/ntex-rs/ntex-amqp

minghuaw:
I have been working on a tokio and serde based AMQP 1.0 implementations called [fe2o3-amqp](https://crates.io/crates/fe2o3-amqp), and I am planning to start working on the qpid integration test soon. It might be worth a try, and any feedback would really apprecaited

minghuaw:
How much of the AMQP 1.0 spec is needed for Azure SDK? I have currently implemented the core spec as well as the web socket binding, and I am currently debating which extension spec to implement next.

minghuaw:
I have made two working examples showing how to use `fe2o3-amqp` with Azure Service Bus, one over TLS and the other one over WebSocket. Both have been tested with my Azure free trial.

1. https://github.com/minghuaw/fe2o3-amqp/tree/main/examples/service_bus
2. https://github.com/minghuaw/fe2o3-amqp/tree/main/examples/service_bus_over_websocket

I might start working on implementing the claim based security next. Would love to have some feedback :)

minghuaw:
Here is another quick update. I have added some working examples of sending to and receiving from Azure Event Hubs.

- https://github.com/minghuaw/fe2o3-amqp/tree/main/examples/event_hub

minghuaw:
I think I figured out the offset issue with the filter [example](https://github.com/minghuaw/fe2o3-amqp/blob/main/examples/event_hubs/src/bin/receiver_with_filter.rs)

minghuaw:
I am planning to work on an amqp client for servicebus and event hub. Is there any recommendations? Would mimicking the dotnet sdk be a good starting point?

cataggar:
[...]

PR #1185, opened 2022-12-28
AMQP 1.0 Support for Service Bus SDK

minghuaw:
This PR is a partial fullfillment of service bus SDK using the AMQP 1.0 protocol (#643). The overall design follows that of the dotnet SDK.

## What's added

The following key public types are added in this PR.

- `ServiceBusClient`
- `ServiceBusSender`
- `ServiceBusReceiver`
- `ServiceBusSessionReceiver`
- `ServiceBusRuleManager` 
- `ServiceBusMessageBatch`

This PR also changes the minimum required rust version to 1.65 due to the use of generic associate types.

## What's removed

The old `reqwest` based HTTP client (`QueueClient`, `SubscriptionReceiver`, `TopicClient`, and `TopicSender`) and the corresponding dependencies/examples/tests are removed.

## What's not implemented yet

- Transaction
- `ServiceBusProcessor`
- `ServiceBusSessionProcessor`
- `ServiceBusAdministrationClient`

## Testing

Mock testing (using the `mockall` crate) has been applied to some unit tests but not implemented for integration testing, and integration testing relies heavily on live testing. Because of this, the environment variables listed below must be set in order to run `cargo test`. For testing on a local machine, it is recommended to put these environment variables in a `.env` file that is placed at the root of this crate (ie. `azure-sdk-for-rust/sdk/messaging_servicebus/.env`).

- `SERVICE_BUS_CONNECTION_STRING` - full connection string to a service bus namespace. It should look like `"Endpoint=sb://<your-namespace>.servicebus.windows.net/;SharedAccessKeyName=<your-policy>;SharedAccessKey=<your-key>"`
- `SERVICE_BUS_NAMESPACE` - `"<your-namespace>.servicebus.windows.net"`
- `SERVICE_BUS_SAS_KEY_NAME` - shared access key name, for example `"RootManageSharedAccessKey"`
- `SERVICE_BUS_SAS_KEY` - shared access key value. This is essentially the field `"<your-key>"` in the full connection string
- `SERVICE_BUS_QUEUE` - a queue that does NOT have session enabled
- `SERVICE_BUS_SESSION_QUEUE` - a queue that has session enabled
- `SERVICE_BUS_TOPIC` - a topic whose subscriptions will be used for non-session topic/subscription testing
- `SERVICE_BUS_SUBSCRIPTION` - a subscription of `SERVICE_BUS_TOPIC`, and it does NOT have session enabled
- `SERVICE_BUS_SESSION_TOPIC` - a topic whose subscriptions will be used for session-enabled topic/subscription testing
- `SERVICE_BUS_SESSION_SUBSCRIPTION` - a subscription of `SERVICE_BUS_SESSION_TOPIC`, and it must have session enabled
- `SERVICE_BUS_RULE_FILTER_TEST_TOPIC` - a topic whose subscription will be used to test get/delete/create rule filters. It is recommended to have a separate topic that is different from `SERVICE_BUS_TOPIC` and `SERVICE_BUS_SESSION_TOPIC` to avoid accidentally affecting the subscription testing with the rule filters.
- `SERVICE_BUS_RULE_FILTER_TEST_SUBSCRIPTION` - a subscription of `SERVICE_BUS_RULE_FILTER_TEST_TOPIC` that will be used for get/delete/create rule filter testing. It is recommended to have a separate topic/subscription that is different from `SERVICE_BUS_TOPIC`/`SERVICE_BUS_SUBSCRIPTION` and `SERVICE_BUS_SESSION_TOPIC`/`SERVICE_BUS_SESSION_SUBSCRIPTION` to avoid accidentally affecting the subscription testing with the rule filters.

Most integration test cases assume an empty queue/subscription at the beginning of the test and should leave an empty queue/subscription at the end of the test. So it is important to make sure only one live test is running for the testing queue/topic/subscription.

There is one long test in the integration tests (`fn send_to_queue_every_minute_for_two_hour()`  in `messaging_servicebus/tests/long_tests.rs`) that is ignored in the usual `cargo test`. This test can be explicitly executed with `cargo test --test long_tests -- --ignored --exact --nocapture` which will also periodically print out message sent and message received. This test could also be used for testing recovery against network interruptions.

I can provide my testing namespace/queue/topic/subscription (actually a copy of my `.env` file) in an email if necessary.

## Documentation

The current implementation sets a rust doc flag `"docsrs"` to conditionally import types for intra-doc-links. To test the doc locally, the following command could be used. Please note that a nightly toolchain is needed (the nightly toolchain will used for the doc only, the crate itself doesn't need nightly at all)

`cargo +nightly doc --no-deps --open -Z unstable-options --config "build.rustdocflags=[\"--cfg\", \"docsrs\"]"`

Most of the documentations are ported from the dotnet SDK. More detailed explanation and/or examples might be necessary

## Limitations and discussions

I have not got the time to review whether the APIs are cancel safe, but the send and recv method provided by the upstream AMQP crate have been reviewed and should be cancel safe.

There are a few design decisions that, I think, could be discussed

1. Whether `AzureNamedKeyCredential` and `AzureSasCredential` should be moved to the `azure_core` crate (which is the way that is implemented in the dotnet sdk)?
2. The current retry policy trait and implementation (`ServiceBusRetryPolicy` and `BasicRetryPolicy`) essentially mimic what the dotnet sdk does. However, there is another retry policy trait in the `azure_core` crate (`azure_core::RetryPolicy`), which is not exactly the same. Should this implementation move to use `azure_core::RetryPolicy` instead?

There is a known problem in the current implementation. Sometimes, upon certian network interruptions (eg. switching from one wireless network to another), the sending operation may need to wait for `ServiceBusRetryOptions::delay` (default to 1 minute) before it starts to try to recover the connection and retry the sending operation. The `tests/long_tests.rs` has been used for testing network interruptions by manually disconnecting/reconnecting or switching the network.

Another problem is logging. Both `log` and `tracing` are popular in the rust ecosystem. Would it make sense to use a feature to allow user to choose which logging crate to use? (this will likely be in the next PR). Ths current use of `log` also definitely has room for improvement.


minghuaw:

minghuaw:
There are three failed checks. Two are caused by `wasm` not supported by `tokio` with `net` feature, and the other failed check is caused by the required environment variables not set.

I guess I can take a look at how `reqwest` achieves `wasm` support.

minghuaw:
It seems like `tokio_wasi` might be a drop-in wasm replacement for `tokio`. However, this is maintained by wasmedge, I might need to test a wasm build with different wasm runtime then

minghuaw:
Another option is probably keeping the `reqwest` based HTTP client for `wasm32` targets and using the AMQP client for non-wasm32 targets

ctaggart:
I do not think this initial pull request needs to support AMQP on Wasm or WASI. I do think that AMQP on WASI would be a fantastic addition for another pull request. I think it makes sense to use an AMQP implementation for the `azure_messaging_servicebus` crate. I am not sure if we want to rename or maintain the REST based crate. One idea would be to rename it to `azure_messaging_servicebus_rest`.

minghuaw:
I think supporting WASI would be straightforward once `tokio`'s full wasm or wasi support lands. Keeping the AMQP and rest client as two separate crates indeed sounds like a better solution

minghuaw:
I think another possibility for wasm support is to restrict the `TransportType` to `AmqpWebSockets`. It looks like there is a crate for websocket on wasm (https://crates.io/crates/ws_stream_wasm).

minghuaw:
While I have had some success with establishing websocket connection to service bus, the problem that the futures generated for wasm is not `Send` becomes the next big problem. I guess I will leave wasm support till the next PR then.

minghuaw:
> While I have had some success with establishing websocket connection to service bus in wasm, the problem that the futures generated for wasm is not `Send` becomes the next big problem. I guess I will leave wasm support till the next PR 

Some good news here. I managed to make most of the client API of the upstream AMQP crate work in wasm. And I was able to send to a test service bus queue in a browser tab. The big problem with the AMQP crate right now is that somehow closing the `Session`/`Connection` gets stuck forever. I will probably draft another PR once I get this problem solved.

cataggar:
@minghuaw, this is looking great! Is there a way to get CI passing? May be the crate can be removed from the wasm check.

minghuaw:
> @minghuaw, this is looking great! Is there a way to get CI passing? May be the crate can be removed from the wasm check.

What about making this PR using the original HTTP client for wasm32 target_arch? I have a branch for that, and the next PR will be using AMQP in both wasm32 and non-wasm32 targets.

cataggar:
My preference is to remove and not support the REST based API. It would be best to disable the wasm check for azure_messaging_servicebus.

minghuaw:
> My preference is to remove and not support the REST based API. It would be best to disable the wasm check for azure_messaging_servicebus.

Sounds good. I will put everything behind `#[cfg(not(target_arch = "wasm32"))]` later today.

minghuaw:
I have gated everything behind a `#[cfg(not(target_arch = "wasm32"))]`. I think the only thing left for the CI is setting those environment variables for the live tests.

minghuaw:
@cataggar Quick question. Would it be better to expose the choice of `rustls` or `native-tls` to the user through features? This PR currently only uses `rustls`, but the upstream AMQP crate supports choosing either using the corresponding feature

minghuaw:
I have exposed choice of TLS implementation through feature flags and updated crate level documentation to reflect this.

minghuaw:
There are two failed checks, one is `cargo test` in nightly and the other is `cargo` test in stable. Both are caused by not finding the environment variables for the live tests.

sdk/messaging_servicebus/Cargo.toml
```diff
@@ -9,27 +9,70 @@ repository = "https://github.com/azure/azure-sdk-for-rust"
...
+dotenv = "0.15"
 futures = "0.3"
-tokio = { version = "1.0", features = ["macros", "rt-multi-thread"] }
+tokio = {version = "1.0", features = ["macros", "rt-multi-thread"]}
+mockall = "0.11"
 env_logger = "0.10"
 
 [features]
-default = ["azure_core/enable_reqwest"]
+default = ["rustls"]
```
cataggar, line 70: Default should be `native-tls` like all our other crates. You must opt in to `rustls`. In Microsoft, we must use `native-tls`.

minghuaw: Fixed this in the new commit

minghuaw:
I managed to pass all checks in my own forks using github environment secrets, and the workflow file (build/yml) was modified to set those environment variables for the test environment. However, because `test-sdk` on stable and nightly, and integration test all run `cargo test`, the workflow file was also modified to run these three jobs sequentially. This makes the entire workflow take 45 minutes to finish.

I am wondering if there is a better strategy for live integration testing?

minghuaw:
I just realized that the integration tests (`sdk/messaging_servicebus/tests/service_bus.rs`) for the old rest-based client were not run but simply checked. The three jobs that invokes `cargo test` do not toggle the `test_e2e` feature, which disables the tests, and the test_e2e job simple calls `cargo check --tests --features test_e2e`, which again does not actually run the tests.

I guess I will follow the same setup for the sake of passing the CI for now.

cataggar:
Hi @minghuaw, we discussed this PR on our internal biweekly call with some project contributors. We love the work that you are doing!

Unfortunately, we do not have the bandwidth to accept and own this right now. It was recommended that you release it separately under your own crate name and we revisit bringing it on in 6 months.

minghuaw:
@cataggar Thank you for the feedback.

Would it make sense if I push future works on wasm support, transaction and processor directly to this PR before the revisit? I will probably convert this into a draft PR for those work.

minghuaw:
FYI , I have released this under the name `azservicebus` for now

minghuaw:
I have just added initial support for `wasm32`. An example that works in a browser tab can be found here (https://github.com/minghuaw/azure-sdk-for-rust/tree/servicebus_over_amqp/sdk/messaging_servicebus/examples/wasm32_in_browser).

mario-guerra:
Hi @minghuaw, are you still using the service bus test resource I set up for you?

minghuaw:
> Hi @minghuaw, are you still using the service bus test resource I set up for you?

No. I am currently only using the ones I created under my own SB namespace

mario-guerra:
Okay, thank you. I will delete the resource.

cataggar:
@minghuaw, thank you for continuing to update this. ❤️

cataggar:
We really like seeing this work, but it is not something those of us volunteering can take on. It is probably best to create a new git repo for `azservicebus`.

https://crates.io/crates/azservicebus

Issue #1189, opened 2023-01-03
Implement `IntoFuture` trait in services


ctaggart:
Add support like https://github.com/microsoft/azure-devops-rust-api/pull/89. When complete, it should be possible to remove `into_future()` from:

https://github.com/Azure/azure-sdk-for-rust/blob/356952d4e10a8d93e0cc782eb01a53f95604ec68/services/mgmt/compute/examples/list_free_linux_images.rs#L25-L32

If you try to remove it right now, you get:

```
PS C:\Users\cataggar\io\azure-sdk-for-rust\services\mgmt\compute> cargo check --examples
    Checking azure_mgmt_compute v0.9.0 (C:\Users\cataggar\io\azure-sdk-for-rust\services\mgmt\compute)
error[E0277]: `azure_mgmt_compute::virtual_machine_images::list_publishers::RequestBuilder` is not a future
  --> mgmt\compute\examples\list_free_linux_images.rs:26:54
   |
26 |           .list_publishers(&location, &subscription_id)
   |  ______________________________________________________^
27 | |         .await?
   | |______________^ `azure_mgmt_compute::virtual_machine_images::list_publishers::RequestBuilder` is not a future
   |
   = help: the trait `Future` is not implemented for `azure_mgmt_compute::virtual_machine_images::list_publishers::RequestBuilder`
   = note: azure_mgmt_compute::virtual_machine_images::list_publishers::RequestBuilder must be a future or must implement `IntoFuture` to be awaited
   = note: required because of the requirements on the impl of `IntoFuture` for `azure_mgmt_compute::virtual_machine_images::list_publishers::RequestBuilder`
help: remove the `.await`
   |
26 -         .list_publishers(&location, &subscription_id) 
26 +         .list_publishers(&location, &subscription_id)?
   |

For more information about this error, try `rustc --explain E0277`.
error: could not compile `azure_mgmt_compute` due to previous error
```

PR #1193, opened 2023-01-11
Implement IntoFuture trait for RequestBuilder in autorust for autogenerated services

johnbatty:
Implement support for the new [std::future::IntoFuture](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#enhancing-await-with-intofuture) trait on request builder objects.

This eliminates the need to call `into_future()` on request builders, as `.await` automatically invokes `into_future()` if the `IntoFuture` trait is implemented.

This change simplifies the API.  However, it is a breaking API change for existing users.

The change requires Rust 1.64.0 or higher. This is enforced via the `rust-version` field in `Cargo.toml` for each of the generated crates.

I also updated the documentation for `RequestBuilder` to provide more guidance on usage, and to help users locate the return type (which now appears on the `IntoFuture::into_future()` method in the traits section) .  See [this example](https://docs.rs/azure_devops_rust_api/latest/azure_devops_rust_api/git/repositories/list/struct.RequestBuilder.html) from my `azure_devops_rust_api` crate docs for how the final documentation looks.

Fixes: https://github.com/Azure/azure-sdk-for-rust/issues/1189


cataggar (approved):
Fantastic! ❤️

PR #1194, opened 2023-01-12
update base64 to 0.21

demoray:
Updating base64 included addressing breaking changes in the API.  This update follows the crate's [RELEASE-NOTES](https://github.com/marshallpierce/rust-base64/blob/master/RELEASE-NOTES.md#0210) for updating our usage.


heaths:
But do we otherwise need to? I was reading about this situation this morning, but do we actually have a need to update? Or is this just to make it easier to upgrade should we ever need to for good reason?

cataggar:
The build is failing for this one and it has merge conflicts.

heaths (approved):
Nice to remove that dependency and simplify the call sites.

PR #1201, opened 2023-02-08
Fix: BytesStream failing to set correct slice range

dmweis:
`BytesStream` calls [Bytes::slice](https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.slice) with incorrect range of `START..SIZE`. Range should be `START..END`.

This results in incorrect slicing or even an assert failure of `"range start must not be greater than end: ` coming from [the slice methnod](https://github.com/tokio-rs/bytes/blob/74b04c7aae5fd7e73f4283774eab0ef72a26a8a7/src/bytes.rs#L255).

I wanted to add a test for this but don't see any tests for this code. If there are some I am happy to add a test for this case.

Tested locally from my fork

### DIFF
--- a/sdk/core/src/bytes_stream.rs
+++ b/sdk/core/src/bytes_stream.rs
@@ -99,3 +99,54 @@ impl AsyncRead for BytesStream {
         }
     }
 }
+
+// Unit tests
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use futures::io::AsyncReadExt;
+    use futures::stream::StreamExt;
+
+    // Test BytesStream Stream
+    #[test]
+    fn test_bytes_stream() {
+        let bytes = Bytes::from("hello world");
+        let mut stream = BytesStream::new(bytes.clone());
+
+        let mut buf = Vec::new();
+        let mut bytes_read = 0;
+        while let Some(Ok(bytes)) = futures::executor::block_on(stream.next()) {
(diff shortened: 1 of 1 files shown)


dmweis:
Ah. Looks like the failures are because of new clippy version. Not because of my changes.

Should I create a separate PR to fix that?

demoray:
#1203 should have already addressed the clippy issues.  Can you rebase to the latest `main`?

cataggar:
I'm curious what the impact is and how we get unit tests on it.

johnbatty:
Fix looks good to me.

johnbatty:
@dmweis I've written a few unit tests to provide some coverage of these functions (and which fail with the unmodified code/work with your fix).  Are you happy for me to push these to your branch?

dmweis:
> @dmweis I've written a few unit tests to provide some coverage of these functions (and which fail with the unmodified code/work with your fix). Are you happy for me to push these to your branch?

Oh! That would be amazing! Please go ahead!

And my appologieze. I should have written tests. I forgot about this PR.

Happy to do any more work if there is anything else you'd like changed?

johnbatty:
@dmweis Ok, I've added a few unit tests.
@demoray Please could you take another look.

demoray (approved):

dmweis:

PR #1202, opened 2023-02-08
fix bare URLs in documentation

demoray:
This addresses a warning seen when building crate documentation

```
 --> sdk/storage_blobs/src/options/block_id.rs:7:16
  |
7 | /// Reference:
https://learn.microsoft.com/en-us/rest/api/storageservices/put-block#uri-parameters
  |
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: use an automatic link instead:
`<https://learn.microsoft.com/en-us/rest/api/storageservices/put-block#uri-parameters>`
  |
  = note: bare URLs are not automatically turned into clickable links
```

### DIFF
--- a/sdk/core/src/util.rs
+++ b/sdk/core/src/util.rs
@@ -17,7 +17,7 @@ where
 }
 
 /// Deserialize JSON null as default
-/// https://github.com/serde-rs/serde/issues/1098
+/// <https://github.com/serde-rs/serde/issues/1098>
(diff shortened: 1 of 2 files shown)


demoray:
This is included in #1203

Issue #1207, opened 2023-02-10
[DataTable] insert_or_replace and insert_or_merge fails with 404 inside transaction if entity does not already exist


mdahamiwal:
Example failing if entity does not already exist in the table with 404 response code
[[https://github.com/Azure/azure-sdk-for-rust/blob/873c80c99a84b96ebd8e5f98ba3b09da417024e3/sdk/data_tables/examples/table.rs
        .insert_or_replace(&entity5.surname, &entity5, IfMatchCondition::Any)?
        .insert_or_merge(&entity6.surname, &entity6, IfMatchCondition::Any)?](https://github.com/Azure/azure-sdk-for-rust/blob/873c80c99a84b96ebd8e5f98ba3b09da417024e3/sdk/data_tables/examples/table.rs#L82)](https://github.com/Azure/azure-sdk-for-rust/blob/873c80c99a84b96ebd8e5f98ba3b09da417024e3/sdk/data_tables/examples/table.rs#L82)

### Root cause

SDK is setting `if-match=*` in request header which makes the request conditional and enforces that etag should exist. https://learn.microsoft.com/en-us/rest/api/storageservices/insert-or-replace-entity 

### fix 
Stop sending if-match header for insert_or_replace and insert_or_merge operation. 

https://github.com/Azure/azure-sdk-for-rust/blob/d27ac14034848dbc7857bdf8711f7fffed5c5bf5/sdk/data_tables/src/operations/transaction.rs#L69

PR #1208, opened 2023-02-10
Fix: Avoid passing if-match header for insert_or_replace ops

mdahamiwal:
#1207 if-match header enforces that the entity already exists whiich breaks the insert new scenarios.


demoray (approved):

demoray:
The If-Match implementation is aligned with the [documentation](https://learn.microsoft.com/en-us/rest/api/storageservices/performing-entity-group-transactions) for update, delete, and merge.

PR #1209, opened 2023-02-12
Azure IMDS default endpoint does not use HTTPS

dpiche174:


dpiche174:

demoray:
@dpiche174 This PR updates generated code.  Unfortunately, directly updating the code will only correct this issue until the next time we regenerate the crate.  

The underlying specification, which this crate is based upon, states it supports both HTTP and HTTPS.

https://github.com/Azure/azure-rest-api-specs/blob/4802c0f73151b90bd207df82cb0207d70db7b445/specification/imds/data-plane/Microsoft.InstanceMetadataService/stable/2021-12-13/imds.json#L8-L13

I've submitted an issue to track updating the rest API specification ( https://github.com/Azure/azure-rest-api-specs/issues/22581), but we can probably work on a work-around to address using the correct scheme until this is addressed upstream.

demoray:
Note, in the meantime, you can use [ClientBuilder.endpoint](https://docs.rs/azure_svc_imds/latest/azure_svc_imds/package_2021_12_13/struct.ClientBuilder.html#method.endpoint) to specify an alternate endpoint of `http://169.254.169.254/metadata`.

demoray:
This will be fixed in #1211 (or another PR like it).

dpiche174:
> Note, in the meantime, you can use [ClientBuilder.endpoint](https://docs.rs/azure_svc_imds/latest/azure_svc_imds/package_2021_12_13/struct.ClientBuilder.html#method.endpoint) to specify an alternate endpoint of `http://169.254.169.254/metadata`.

Works perfectly. Thank you!

PR #1215, opened 2023-02-15
bump crate versions in prep of 2023-02 release

demoray:
Note: While this updates the generated crate template, this does not regenerate the crates.  That will be done in a follow-up commit.


sdk/core/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "azure_core"
-version = "0.8.0"
+version = "0.10.0"
```
cataggar, line 3: Shouldn't this be 0.9.0?

demoray:
Note: This bumps azure_core to 0.10 such that all of the shipping crate versions are the same.  This simplifies both our lives as maintainers and eases the burden for the user.

sdk/core/Cargo.toml
demoray: I commented on the PR on this.  I intentionally bumped azure_core to 0.10.0 to reduce the burden of having to keep track of different version numbers for different crates.

demoray:
To review _just_ the code in the sdk crates: https://github.com/Azure/azure-sdk-for-rust/pull/1215/commits/4a5ed10bbecf8e679a0a269245e1ea647a1b1e18

To review _just_ the code in the generated crates: https://github.com/Azure/azure-sdk-for-rust/pull/1215/commits/a05f07bbf233ddc678c1be9e26857daa5c0bcffe

cataggar (approved):

PR #1223, opened 2023-02-19
Allow parsing list blobs response without ServerEncrypted tag

rguerreiromsft:
Allow parsing list blobs response without ServerEncrypted tag, which is how Azurite responds when a blob is created by multi-block algorithm.

By default, it will assume this tag is false.

I wasn't able to test this using the e2e provided with my Azurite credentials. But I was able to test this issue independently, and I added a few tests to cover this situation. The payload in `deserde_azurite_without_server_encrypted` is a true payload that happened to me while listing blobs.

I'm not sure if it makes a difference, but this is how my container is setup:
`azurite --blobHost 0.0.0.0 --queueHost 0.0.0.0 --tableHost 0.0.0.0 --loose`


demoray (approved):

demoray:
Thanks for the contribution!

Issue #1225, opened 2023-02-21
AutoRefreshingTokenCredential token override issue


Nohac:
Hi.

I have a program that streams some data from an http request to azure storage blob as well as accessing some other api's that also needs to authenticate using the same `AutoRefreshingTokenCredential`.

This is the logic flow:
1. Stream all data to storage using `put_block` (wait to run `put_block_list` for now)
2. Access other api's and do some work (uses `get_token("<scope>")` on the same `AutoRefreshingTokenCredential` that the storage client uses)
3. If everything succeeds in step 2, run `put_block_list` to "finalize" the upload.

The problem I'm facing is that, after running `get_token` in step 2, step 3. ( `put_block_list`) returns the following error:
```
server returned error status which will not be retried: 401

Caused by:
    HttpError {  Status: 401,  Error Code: InvalidAuthenticationInfo,  Body: "b"\xef\xbb\xbf<?xml version=\"1.0\" encoding=\"utf-8\"?><Error><Code>InvalidAuthenticationInfo</Code><Message>Server failed to authenticate the request. Please refer to the information in the www-authenticate header.\nRequestId:<omitted>\nTime:2023-02-21T16:11:53.5576343Z</Message><AuthenticationErrorDetail>Issuer validation failed. Issuer did not match.</AuthenticationErrorDetail></Error>"",  Headers: [   date:Tue, 21 Feb 2023 16:11:53 GMT   content-type:application/xml   server:Microsoft-HTTPAPI/2.0   content-length:402   x-ms-error-code:InvalidAuthenticationInfo   www-authenticate:Bearer authorization_uri=https://login.microsoftonline.com/<omitted>/oauth2/authorize resource_id=https://storage.azure.com   x-ms-request-id:<omitted> ], } 
```

If I skip step 2. then step 3. works.

Another observation I made is that, if I don't use `AutoRefreshingTokenCredential`, all the steps work:
```rust
let creds = Arc::new(DefaultAzureCredential::with_sources(defaul_credentials));
// let creds = Arc::new(AutoRefreshingTokenCredential::new(creds));
return creds;
```

I'm not really sure what's going on here, but I suspect `AutoRefreshingTokenCredential` overwrites the current token when I run `get_token` to call the API's, then when it gets to `put_block_list`, it uses the token from step 2?

Another thing I'm not sure about is if I even need `AutoRefreshingTokenCredential`, the documentation is really sparse, all it says is:
```
[...]

Issue #1143, opened 2022-11-01
AutoRefreshingTokenCredential is always expired for AzureCliCredential


rguerreiromsft:
It seems that changing from chrono to time caused the `az account get-access-token --output json --resource "https://storage.azure.com/"` command date to be converted to an expired date.

I've ran tests connecting to Azure Storage Account Queues and it works without `AutoRefreshingTokenCredential`, but with it, it hangs in the refreshing loop.

I've created the following test, and it fails:

```
    #[test]
    fn test() -> azure_core::Result<()> {
        let tr = AzureCliCredential::get_access_token(Some("https://storage.azure.com/"))?;
        let token = TokenResponse::new(tr.access_token, tr.expires_on);

        println!("{}", token.expires_on);
        // Prints 2022-11-01 16:18:56.0 +00:00:00
        // But the correct time would be `2022-11-01 16:18:56.0 -07:00:00` Notice the TZ portion.
        // Typing `date` on terminal I get: Tue Nov  1 15:05:40 PDT 2022

        let expired = token.expires_on < OffsetDateTime::now_utc() + Duration::from_secs(20);

        assert!(!expired, "Brand new token should not be expired.");
        Ok(())
    }
```

demoray:
This is fixed in #1176

PR #1229, opened 2023-02-23
Implement a per-resource token cache for AutoRefreshingTokenCredential

johnbatty:
The current implementation of `AutoRefreshingTokenCredential` caches a single token.  This does not work if the same credential is used for multiple resources, as tokens are specific to a given resource.

This PR updates `AutoRefreshingTokenCredential` to cache tokens per-resource.  This is implemented using a HashMap indexed by the resource name.

Fixes: https://github.com/Azure/azure-sdk-for-rust/issues/1225

### DIFF
--- a/sdk/identity/src/token_credentials/auto_refreshing_credentials.rs
+++ b/sdk/identity/src/token_credentials/auto_refreshing_credentials.rs
@@ -15,7 +15,7 @@ fn is_expired(token: &TokenResponse) -> bool {
 pub struct AutoRefreshingTokenCredential {
     credential: Arc<dyn TokenCredential>,
     // Tokens are specific to a resource, so we cache tokens by resource.
-    token_cache: Arc<RwLock<HashMap<String, azure_core::Result<TokenResponse>>>>,
+    token_cache: Arc<RwLock<HashMap<String, TokenResponse>>>,
 }
 
 impl std::fmt::Debug for AutoRefreshingTokenCredential {
@@ -42,7 +42,7 @@ impl TokenCredential for AutoRefreshingTokenCredential {
     async fn get_token(&self, resource: &str) -> azure_core::Result<TokenResponse> {
         // if the current cached token for this resource is good, return it.
         let token_cache = self.token_cache.read().await;
-        if let Some(Ok(token)) = token_cache.get(resource) {
+        if let Some(token) = token_cache.get(resource) {
             if !is_expired(token) {
                 return Ok(token.clone());
             }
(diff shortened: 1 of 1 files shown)


sdk/identity/src/token_credentials/auto_refreshing_credentials.rs
```diff
@@ -10,10 +11,11 @@ fn is_expired(token: &TokenResponse) -> bool {
 }
 
 #[derive(Clone)]
-/// Wraps a TokenCredential and handles token refresh on token expiry
+/// Wraps a TokenCredential and handles token refresh on token expiry.
 pub struct AutoRefreshingTokenCredential {
     credential: Arc<dyn TokenCredential>,
-    current_token: Arc<RwLock<Option<azure_core::Result<TokenResponse>>>>,
+    // Tokens are specific to a resource, so we cache tokens by resource.
+    token_cache: Arc<RwLock<HashMap<String, azure_core::Result<TokenResponse>>>>,
```
demoray, line 18: I recognize this is existing behavior, but could we move to only caching the `TokenResponse`, not the `Result<TokenResponse>`?   We only use the value if it's `Ok`.

demoray (approved):

sdk/identity/src/token_credentials/auto_refreshing_credentials.rs
johnbatty: Yep - excellent idea.  Done.

sdk/identity/src/token_credentials/auto_refreshing_credentials.rs
```diff
@@ -38,38 +40,139 @@ impl AutoRefreshingTokenCredential {
...
         // waiting on the write lock
-        if let Some(Ok(token)) = guard.as_ref() {
+        if let Some(token) = token_cache.get(resource) {
             if !is_expired(token) {
                 return Ok(token.clone());
             }
         }
 
-        let res = self.credential.get_token(resource).await;
+        let result = self.credential.get_token(resource).await;
```
demoray, line 63: What do you think about bubbling out the error using `?`?  The match behavior was important before, when this cached the `Result`, but I think this might be more readable.


```suggestion
        let token = self.credential.get_token(resource).await?;
         // NOTE: we do not check to see if the token is expired here, as at
         // least one credential, `AzureCliCredential`, specifies the token is
         // immediately expired after it is returned, which indicates the token
         // should always be refreshed upon use.        
        token_cache.insert(resource.to_string(), token.clone());
        Ok(token)
```

Note: the GitHub UI makes making a multi-line suggestion that includes new, unchanged, and removed code behave weirdly, so this patch suggestion does not remove the match code later on.

demoray: The difference here is that this code bubbles out the underlying token error instead of wrapping in another error type.

johnbatty: Yep - I agree.  I have updated.

I'm also tempted to remove the `NOTE:...` comment, as I think it is confusing (why would you check a newly created token for expiry?) and possibly out of date.  I've just tested `az account get-access-token` in Linux and Windows and both return an access token with a 1 hour expiry.  Thoughts?

johnbatty: Hmm... I see that the `NOTE:` comment was added as part of the fix for this issue:
    https://github.com/Azure/azure-sdk-for-rust/issues/1143

I remember seeing this issue, and thinking that it must be related to the switch from `chrono` to `time`, and handling of timezones when parsing the returned expiry time.  I did look at the code and couldn't see anything obviously wrong with it, but might take another look.

demoray (approved):

sdk/identity/src/token_credentials/auto_refreshing_credentials.rs
demoray: LGTM

PR #1234, opened 2023-03-08
storage_blobs: Fix typo in VersionId definiton.

gorzell:
It looks like the part of the VersionId that is used for the key when adding it as a query arg has a typo `version_id`. The [API docs](https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob?tabs=azure-ad#uri-parameters) specifiy the parameter name as `versionid`.


demoray (approved):
This matches the REST API specs as well.

https://github.com/Azure/azure-rest-api-specs/blob/main/specification/storage/data-plane/Microsoft.BlobStorage/preview/2021-12-02/blob.json#L12084

Thanks for the contribution.

PR #1235, opened 2023-03-09
Key Vault Certificate Operation Get and Delete

daniel-larsen:
Adds Key Vault certificate operation get and delete.


sdk/security_keyvault/src/clients/certificate_client.rs
```diff
@@ -250,6 +279,38 @@ impl CertificateClient {
...
+    ///     let mut client = KeyvaultClient::new(
+    ///         &"KEYVAULT_URL",
+    ///         Arc::new(creds),
+    ///     ).unwrap().certificate_client();
+    ///     client.delete_operation("NAME").await.unwrap();
+    /// }
+    ///
+    /// Runtime::new().unwrap().block_on(example());
+    /// ```
+    pub fn delete_operation<N>(&self, name: N) -> DeleteCertificateOperationBuilder
```
heaths, line 307: To somewhat align with our other languages, this should be `delete_certificate`. In other lanugages we actually define this as a long-running operation (LRO), but only because DELETE returns immediately but may not take effect immediate. It's a pseudo-LRO and I think, at this point, it's not worth replicating here.

heaths (approved):

sdk/security_keyvault/src/clients/certificate_client.rs
daniel-larsen: Sorry for the confusion. I forgot to update the documentation above this function. This function deletes the creation operation for a specific certificate not the certificate itself. https://learn.microsoft.com/en-us/rest/api/keyvault/certificates/delete-certificate-operation/delete-certificate-operation?tabs=HTTP

heaths: I see. Eventually these might have to change to better align with our other languages' SDKs, but this works for now if that unblocks you.

heaths:
Just need to fix the code style issue and I can merge.

demoray:
The code style issues are newly an issue with the latest rust.  I'll have a PR open shortly.

heaths:
@demoray you want me to wait, or just merge it anyway (I can)?

daniel-larsen:
I will merge @demoray's changes.

PR #1240, opened 2023-03-14
[AppConfiguration] Adds initial support for Azure AppConfiguration and feature management

Aliaksie:
This PR is a partial fullfillment of appconfiguration and feature management  SDK.

### What's added

appconfiguration crate

### What's removed
nothing

### What's not implemented yet
 - reusable cache for appconfiguration items   
 - auto refresh via Azure Event Grid 
 - mock/e2e test 

### Testing
Tested manually by own azure appcofiguration instance, also added examples that covered main functionality    

### Limitations and discussions
Elaboration how to use differ version of generated services. 

@PavelKhatsimtsau  thanks for helping.


Aliaksie:

sdk/appconfiguration/src/feature_manager/feature_type.rs
```diff
@@ -0,0 +1,39 @@
...
+const FEATURE_TARGETING: &str = "Microsoft.Targeting";
+const FEATURE_PERCENTAGE: &str = "Microsoft.Percentage";
+const FEATURE_TIME_WINDOW: &str = "Microsoft.TimeWindow";
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum Feature {
+    Percentage(FeatureContext),
+    Targeting(FeatureContext),
+    TimeWindow(FeatureContext),
+    OnOff(bool),
```
demoray, line 14: I'm having trouble finding this in the service documentation or in other SDKs.  Is there reference documentation for this feature somewhere?

sdk/appconfiguration/src/configuration/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+            .into_stream();
+        while let Some(rs) = stream.next().await {
+            match rs {
+                Ok(rs) => {
+                    let items = rs
+                        .items
+                        .iter()
+                        .map(|it| match (&it.key, &it.value) {
+                            (Some(key), Some(value)) => (key.to_string(), value.to_string()),
+                            _ => (String::from(""), String::from("")),
```
demoray, line 60: Can you describe the use case for adding the key & value of empty strings here?

sdk/appconfiguration/src/configuration/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+            Some(value) => value,
+            None => String::from(""),
+        }
+    }
+
+    pub fn get_value(&self, key: String) -> String {
+        block_on(self.get_value_async(key))
+    }
+
+    pub async fn get_values_async(&self, label: String) -> HashMap<String, String> {
```
demoray, line 44: As the underlying call to azure_svc_appconfiguration's `get_key_values().label()` takes an `Into<String>`, this should as well.  This way we don't need to require the caller to pass an owned value.

sdk/appconfiguration/src/configuration/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+        let mut map = HashMap::new();
+        let mut stream = self
+            .client
+            .clone()
```
demoray, line 48: I don't believe the client needs to be cloned here.

sdk/appconfiguration/src/configuration/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+            Ok(rs) => rs.value,
+            Err(err) => {
+                log::debug!("*ERROR :  {:?}", err);
+                None
+            }
+        };
+
+        match value {
+            Some(value) => value,
+            None => String::from(""),
```
demoray, line 36: Instead of an empty string, have you considered returning `Option<String>` here?

demoray: If returning an empty string is appropriate here, this could use `value.unwrap_or_default()`

demoray (changes requested):

sdk/appconfiguration/src/configuration/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+
+#[derive(Clone)]
+pub struct ConfigurationExplorer {
+    client: azure_svc_appconfiguration::Client,
+}
+
+impl ConfigurationExplorer {
+    pub fn new(token_credential: Arc<dyn TokenCredential>) -> Self {
+        let name =
+            std::env::var("AZCONFIG_NAME").expect("Missing AZCONFIG_NAME environment variable.");
```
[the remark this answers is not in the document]
Aliaksie: agree, updated according you recommendation.

sdk/appconfiguration/src/configuration/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+            }
+        };
+
+        match value {
+            Some(value) => value,
+            None => String::from(""),
+        }
+    }
+
+    pub fn get_value(&self, key: String) -> String {
```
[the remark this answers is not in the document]
Aliaksie: agree, updated according you recommendation.

sdk/appconfiguration/src/feature_manager/feature_filter.rs
```diff
@@ -0,0 +1,98 @@
+use chrono::DateTime;
```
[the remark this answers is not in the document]
Aliaksie: agree, updated according you recommendation.

sdk/appconfiguration/src/feature_manager/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+            .finish()
+    }
+}
+
+impl FeatureExplorer {
+    pub fn new(
+        token_credential: Arc<dyn TokenCredential>,
+        context: Option<Arc<dyn ContextHolder>>,
+    ) -> Self {
+        let on_off: HashMap<String, Vec<Feature>> = if std::env::var("FEATURE_ON_OFF").is_ok() {
```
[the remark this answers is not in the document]
Aliaksie: agree, updated according you recommendation.

sdk/appconfiguration/src/configuration/mod.rs
Aliaksie: good catch. It is not required, I  replaced it to None.

sdk/appconfiguration/src/configuration/mod.rs
Aliaksie: agree, updated according you recommendation.

sdk/appconfiguration/src/feature_manager/feature_type.rs
Aliaksie: There two options for using such feature type : 

1.  In feature manager, has possibility to create feature toggle without any  `client_filters`, like simple boolean feature (on/off => "enabled": true|false)
```{
	"id": "on_off",
	"description": "",
	"enabled": true,
	"conditions": {
		"client_filters": []
	}
}```
2. I assume it would be great options to developers, has possibility to override remote value of any feature toggles types. So this OnOff stand for this, you could check how it used in example  dev_on_off_feature_manager

File: sdk/appconfiguration/src/configuration/mod.rs (hunk 4 above)
Comment (Aliaksie) [reply sdk/appconfiguration/src/configuration/mod.rs]: agree, updated according you recommendation.

File: sdk/appconfiguration/src/feature_manager/feature_type.rs (hunk 1 above)
Comment (demoray) [reply sdk/appconfiguration/src/feature_manager/feature_type.rs:14]: I recognize these are potentially useful, but this is opinionated functionality that is unlike any of the other SDKs.

Consider [Microsoft.FeatureManagement](https://github.com/microsoft/FeatureManagement-Dotnet), a dotnet library that builds on top of the Azure SDK for Net, and implements ideas such as what you're proposing.

@heaths do you have thoughts on this?

File: sdk/appconfiguration/src/feature_manager/mod.rs [hunk 10]
```diff
@@ -0,0 +1,172 @@
...
+#[derive(Clone)]
+pub struct FeatureExplorerBuider {
+    endpoint: Option<String>,
+    credential: Arc<dyn TokenCredential>,
+    context: Option<Arc<dyn ContextHolder>>,
+    on_off: HashMap<String, Vec<Feature>>,
+    retry: Option<RetryOptions>,
+}
+
+impl FeatureExplorerBuider {
```
demoray, line 35: I believe this is a typo, this should be FeatureExplorerBuilder?

sdk/appconfiguration/src/feature_manager/feature_type.rs
heaths: I'm afraid I don't really have much context on App Configuration, but In .NET, we expose a `string Value` property from the [base class](https://github.com/Azure/azure-sdk-for-net/blob/b2a2f3e281f56fb74298f7e975cef7fa8e3ec830/sdk/appconfiguration/Azure.Data.AppConfiguration/src/Models/ConfigurationSetting.cs#L65) and, in the derived class, effectively make sure we can convert as needed (even for complex objects via JSON) or throw otherwise): https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/appconfiguration/Azure.Data.AppConfiguration/src/Models/FeatureFlagConfigurationSetting.cs

That said, newer libraries use a slightly different approach since .NET nor C# has unions / enums like Rust e.g.,

* [DocumentField](https://github.com/Azure/azure-sdk-for-net/blob/b2a2f3e281f56fb74298f7e975cef7fa8e3ec830/sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs#L97)
* [DocumentFieldValue](https://github.com/Azure/azure-sdk-for-net/blob/b2a2f3e281f56fb74298f7e975cef7fa8e3ec830/sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentFieldValue.cs#L14)

This is an evolving pattern and has been used in a couple newer SDKs or features in older SDKs, like Key Vault.

That said, because Rust has enums like it does, a similar approach such as this makes more sense to me. What do you think, @tg-msft, @JeffreyRichter, @johanste, @bterlson? (Rust enums are effectively like unions with more context - TypeScript has similar enums.)

JeffreyRichter: Shouldn't this boil down to being an "extensible enum". We never discussed the best way to expose these in Rust; we should have that conversation and then I think this would be a usage of it.

heaths:
@Aliaksie thank you for the continued updates, but please do not force push and rewrite history. This makes it very hard to review what you have changed in response to our feedback and may mean we have to re-review the PR again in its entirety.

sdk/appconfiguration/src/feature_manager/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+        Self {
+            credential,
+            endpoint: None,
+            on_off: HashMap::new(),
+            context: None,
+            retry: None,
+        }
+    }
+
+    #[doc = "Set the endpoint."]
```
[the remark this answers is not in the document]
Aliaksie: done, thanks!

sdk/appconfiguration/src/feature_manager/mod.rs
Aliaksie: done, thanks!

CryoMyst:
What's happening with this?

cataggar:
I like to see this. However, I think it could live in its own repo outside of the azure-sdk-for-rust.

Pieeer1:
Why was this not included as part of the sdk? This is something I would expect in the azure-sdk-for-rust, especially considering this also contains keyvault.

Pieeer1:
> Why was this not included as part of the sdk? This is something I would expect in the azure-sdk-for-rust, especially considering this also contains keyvault.

until this is merged i made a rest api wrapper here.

https://github.com/Pieeer1/Azure-App-Config-Rs

PR #1243, opened 2023-03-16
azure_mgmt_containerservice is not built anymore using the current rest specs

demoray:


heaths (approved):

XavierGeerinck:
Why was this removed? @demoray what is the way to deploy AKS at the moment with the Rust SDK?

Issue #1252, opened 2023-03-29
azurite_workaround compile failure in azure_storage_blobs


jaimeva-msft:
We're working on a service which talks to Azure Blob Storage, and using azurite to mock out ABS for local testing. While debugging an issue with those tests, we came across the `azurite_workaround` feature flag.

However, a naive attempt to enable it fails to compile. Minimal repro:
```
$ mkdir temp && cd temp
$ cargo init --lib
     Created library package
$ cargo add azure_storage_blobs --features azurite_workaround
      Adding azure_storage_blobs v0.11.0 to dependencies.
             Features:
             + azurite_workaround
             + enable_reqwest
             - enable_reqwest_rustls
             - into_future
             - test_e2e
             - test_integration
$ cargo check --quiet
error[E0308]: mismatched types
  --> /opt/rust/registry/src/crates.io/azure_storage_blobs-0.11.0/src/blob/operations/delete_blob.rs:49:43
   |
49 |     delete_type_permanent_from_headers => delete_type_permanent: Option<bool>,
   |                                           ^^^^^^^^^^^^^^^^^^^^^ expected enum `Option`, found `bool`
   |
   = note: expected enum `std::option::Option<bool>`
              found type `bool`
help: try wrapping the expression in `Some`
   |
49 |     delete_type_permanent_from_headers => delete_type_permanent: Some(delete_type_permanent): Option<bool>,
   |                                           ++++++++++++++++++++++++++++                     +

For more information about this error, try `rustc --explain E0308`.
error: could not compile `azure_storage_blobs` due to previous error
```
 After some head-scratching and poking around, we discovered that the fix was to make sure `azure_core` has the same feature flag enabled:
```
$ cargo add azure_core --features azurite_workaround
      Adding azure_core v0.11.0 to dependencies.
             Features:
             + azurite_workaround
             - enable_reqwest
             - enable_reqwest_gzip
             - enable_reqwest_rustls
             - quick-xml
             - reqwest
             - test_e2e
             - xml
$ cargo check
    Finished dev [unoptimized + debuginfo] target(s) in 0.22s
```

The reason for this is that `azure_core` also uses this feature flag, and the function actually changes signature in the process:
https://github.com/Azure/azure-sdk-for-rust/blob/021079b9a270c18ef7d4654a1a67c7f3cf4c2d10/sdk/core/src/headers/utilities.rs#L72-L80

I believe there's two issues here:
[...]

PR #1254, opened 2023-03-31
chain azurite_workaround when in any one crate

demoray:


demoray:
This addresses one of the two issues raised in #1252 1252

demoray:
Addressing the other issue will simplify this moving forward.

PR #1260, opened 2023-04-09
Implementing Event Hubs SDK over AMQP 1.0

minghuaw:
This is the initial attempt at implementing the EventHubs sdk over AMQP 1.0 protocol in rust. The overall approach is very similar to the Service Bus over AMQP PR (#1185), which essentially follow what the dotnet SDK does. There is one minor distinction though, which is that all public methods that returns a `Result` try to use `azure_core::Error` instead of custom enums. This minor distinction will be adopted by #1185 later.

### What has been implemented in this PR

Not all features are implemented. Below is a list of the features that are included in this PR.

Feature checklist (this will be updated over time):

- [x] `EventHubConnection`
    - [x] Get Event Hub properties
    - [x] Get partition properties
    - [x] Automatic recovery
- [x] `EventHubProducerClient` 
    - [x] create from connection string
    - [x] (won't implement) ~~`send_idempotent`~~
           This is NOT implemented because (1) the underlying AMQP 1.0 sender already takes care of keeping track of unsettled deliveries (2) the current design take the ownership of event/message, making storing things like `pending_sequence_no` useless
    - [x] create and send `EventBatch`
    - [x] Get Event Hub properties
    - [x] Get partition properties
    - [x] Automatic recovery
- [x] `EventHubConsumerClient`
    - [x] create from connection string
    - [x] receive from partition
    - [x] receive from all partition
            Closing the AMQP receivers currently uses a temporary workaround which simply drops all receivers and their sessions. The workaround is acceptable because the internal states of AMQP `Sender` and `Session` (not `Sender` itself but its inner state) implements the `Drop` trait which performs proper exchange of `Detach`/`End` frames upon dropping.
    - [x] auto-recovery
    - [x] impl `Stream`
- [x] `PartitionReceiver`

### What will be implemented in future PRs?

- `EventHubBufferedProducerClient`
- `EventHubProcessor`
- `CheckpointStore` and `BlobCheckpointStore`
- Support for `wasm32` targets


minghuaw:
It seems like Event Hubs will force close a link if the link has been inactive for 30 minutes. Recovering the link may require recovering the connection and session as well.

minghuaw:
This PR has been temporarily published on crates.io as [`azeventhubs`](https://crates.io/crates/azeventhubs)

cataggar:
Hi @minghuaw,

Those of us volunteering on this project really like seeing this work. However, for now, I think it is best suited in an another repository.

https://crates.io/crates/azeventhubs

PR #1261, opened 2023-04-11
azure resource graph : fetch virtual machines

giridharmb:


services/mgmt/compute/examples/vm_list_using_resource_graph.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    // fetch first 10 records
+    let query_options = QueryRequestOptions {
+        skip_token: None,
+        top: Option::from(10),
+        skip: Option::from(0),
+        result_format: None,
+        allow_partial_scopes: None,
+        authorization_scope_filter: None,
+    };
```
demoray, line 35: QueryRequestOptions implements [std::default::Default](https://doc.rust-lang.org/std/default/trait.Default.html).

 As such, this could be simplified via:
```suggestion
    let query_options = QueryRequestOptions {
        top: Option::from(10),
        skip: Option::from(0),
        ..Default::default()
    };
```

giridharmb: Yep, that looks better :)

giridharmb (commented):
Yep, this makes sense !

giridharmb (commented):
Looks good !

services/mgmt/compute/examples/vm_list_using_resource_graph.rs
```diff
@@ -0,0 +1,85 @@
+use std::sync::Arc;
```
demoray, line 1: Thanks for the example.  As this is an example of how to use `azure_mgmt_resourcegraph`, can the example be moved to that crate instead of the `azure_mgmt_compute` crate?

services/mgmt/compute/examples/vm_list_using_resource_graph.rs
```diff
@@ -0,0 +1,85 @@
...
+    // in the below request, we are fetching resources of type (virtual machines)
+    let custom_query_request = QueryRequest {
+        subscriptions: vec![],
+        management_groups: vec![],
+        query: "Resources | where type == 'microsoft.compute/virtualmachines'".to_string(),
+        options: Option::from(query_options),
+        facets: vec![],
+    };
+
+    let query_request = QueryRequest::from(custom_query_request);
```
demoray, line 43: I don't believe this is needed, as `custom_query_request` is already a `QueryRequest` here.

services/mgmt/compute/examples/vm_list_using_resource_graph.rs
```diff
@@ -0,0 +1,85 @@
...
+    InvalidRequest,
+}
+
+#[derive(Debug)]
+struct CustomError {
+    err_type: GenericError,
+    err_msg: String
+}
+
+async fn fetch_azure_records() -> Result<Value, CustomError> {
```
demoray, line 21: Instead of using a custom error & Result, our current preference is to use azure_core::Result.

Something akin to:
```rust
async fn main() -> azure_core::Result<()> {
```

services/mgmt/compute/examples/vm_list_using_resource_graph.rs
```diff
@@ -0,0 +1,85 @@
...
+        }
+        Err(e) => {
+            let actual_error_message = format!("could not fetch data from azure resource graph : {:?}", e.to_string());
+            let custom_err = CustomError {
+                err_msg: actual_error_message,
+                err_type: GenericError::InvalidRequest
+            };
+            return Err(custom_err)
+        }
+    };
```
demoray, line 61: Instead of reserializing these to json for an example, perhaps this could be simplified to show a single entry for each record that is returned?  Something akin to this?

```
    let response = client.resources(query_request).await?;
    println!("total records: {}", response.total_records);
    if let Some(as_array) = response.data.as_array() {
        for entry in as_array {
            let resource_id = entry.get("id");
            println!("{resource_id:?}");
        }
    }
```

This would print something akin to:
```
total records: N
Some(String("/subscriptions/SUBSCRIPTION_ID/resourceGroups/RESOURCE_GROUP/providers/Microsoft.Compute/VirtualMachines/VM_NAME"))
Some(String("/subscriptions/SUBSCRIPTION_ID/resourceGroups/RESOURCE_GROUP/providers/Microsoft.Compute/VirtualMachines/VM_NAME"))
Some(String("/subscriptions/SUBSCRIPTION_ID/resourceGroups/RESOURCE_GROUP/providers/Microsoft.Compute/VirtualMachines/VM_NAME"))
```

services/mgmt/compute/examples/vm_list_using_resource_graph.rs
```diff
@@ -0,0 +1,85 @@
+use std::sync::Arc;
+use azure_identity::DefaultAzureCredential;
+use azure_mgmt_resourcegraph::models::{QueryRequest, QueryRequestOptions};
+use serde_json::{json, Value};
+use azure_mgmt_resourcegraph::Client as AzureResourceGraphClient;
+use azure_mgmt_resourcegraph::ClientBuilder as AzureResourceGraphClientBuilder;
```
demoray, line 6: As this only uses a single `Client`, the alias isn't important here.  Additionally, using `Client::builder` below does not require importing `ClientBuilder` here.  If this was importing clients from multiple crates, differentiating the clients like this would be valuable.

```suggestion
use azure_mgmt_resourcegraph::Client;
```

services/mgmt/compute/examples/vm_list_using_resource_graph.rs
```diff
@@ -0,0 +1,85 @@
...
+struct CustomError {
+    err_type: GenericError,
+    err_msg: String
+}
+
+async fn fetch_azure_records() -> Result<Value, CustomError> {
+
+    let azure_creds = Arc::new(DefaultAzureCredential::default());
+
+    let az_rg_client = AzureResourceGraphClient::builder(azure_creds).build();
```
demoray, line 25:
```suggestion
    let az_rg_client = Client::builder(azure_creds).build();
```

demoray (changes requested):

services/mgmt/compute/examples/vm_list_using_resource_graph.rs
giridharmb: Sure thing. I wasn't sure , hence I initially put it under azure_mgmt_compute, but yeah putting it here : azure_mgmt_resourcegraph makes more sense.

services/mgmt/compute/examples/vm_list_using_resource_graph.rs
giridharmb: Yep, you can make that change.

services/mgmt/compute/examples/vm_list_using_resource_graph.rs
giridharmb: This also works.

services/mgmt/compute/examples/vm_list_using_resource_graph.rs
giridharmb: I actually had multiple Clients in my code, and not just AzureResourceGraphClient. Hence, explicitly used it. But this is just one example, we can use Client as is.

giridharmb (commented):
Looks fine

giridharmb:
Is something pending ? 

Looks like this has not passed : 
Build and Test / Services Tests (pull_request)

giridharmb (commented):
I thought this was already reviewed ?

demoray (approved):

demoray:
@giridharmb thanks for the contribution!

PR #1271, opened 2023-04-26
key_client: supply GetRandomBytes API.

LykxSassinator:
Signed-off-by: Lucasliang <nkcs_lykx@hotmail.com>

In this pr:
* ~~The `ManagedHsmClient` has been supplied for accessing managed HSM resources.~~
* ~~Supply `GetRandomBytes` API to `ManagedHsmClient` for generating random values.~~
* Supply `GetRandomBytes` API to `KeyClient`.


heaths (changes requested):
This should be added to the `KeyClient` reflecting what we do in other languages. Don't validate REST methods apart from any URI path parameters, and let the service return a service error as implemented; thus, `KeyClient` would return an error reflecting the service's HTTP 400.

LykxSassinator:
> This should be added to the `KeyClient` reflecting what we do in other languages. 

Several things are a little bit wired to me, triggering me to split `GetRandomBytes` from `KeyClient`:
* The `GetRandomBytes` API belongs to ManagedHsm, from the official doc, but 
image
* Moreover, I've made the related tests and found that there exists errors that:
> ErrorDetails { code: Some("Unauthorized"), message: Some("Found token with aud=https://vault.azure.net, but expected aud=https://managedhsm.azure.net

These phenomenon showed me that `GetRandomBytes` is not an API belongings to `KeyClient`.

LykxSassinator:
@heaths Done for it. FYI, I accept to make it consistent with implementations in other languages.

heaths (changes requested):
Much better, thank you. One more change, though.

There are a number of APIs that don't work for both Key Vault and Managed HSM, but I'm one of the owners for the SDK and the design of both the service REST APIs and the SDKs are such that developers could test most functionality with the relatively cheap Key Vault service but deploy against a Managed HSM. Few key operations are not supported for both, including /rng.

sdk/security_keyvault/src/keys/models.rs
```diff
@@ -403,3 +403,9 @@ pub struct EncryptResult {
     )]
     pub result: Vec<u8>,
 }
+
+#[derive(Debug, Deserialize)]
+pub struct GetRandomBytesResult {
+    #[serde(rename = "value")]
+    pub result: String,
```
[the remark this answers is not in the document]
LykxSassinator: Got. Done for it, thx.

sdk/security_keyvault/src/keys/models.rs
```diff
@@ -406,6 +406,10 @@ pub struct EncryptResult {
 
 #[derive(Debug, Deserialize)]
 pub struct GetRandomBytesResult {
-    #[serde(rename = "value")]
-    pub result: String,
+    #[serde(
+        rename = "value",
+        serialize_with = "ser_base64",
```
heaths, line 411: It's actually base64url: https://github.com/Azure/azure-rest-api-specs/blob/be4adb7c2681e82c81c25902bb5e4964229a79a2/specification/keyvault/data-plane/Microsoft.KeyVault/stable/7.4/keys.json#L1230

heaths (changes requested):

sdk/security_keyvault/src/keys/models.rs
LykxSassinator: Got it. Thx for the reminder.

sdk/security_keyvault/src/keys/models.rs
```diff
@@ -406,10 +406,7 @@ pub struct EncryptResult {
 
 #[derive(Debug, Deserialize)]
 pub struct GetRandomBytesResult {
-    #[serde(
-        rename = "value",
-        serialize_with = "ser_base64",
-        deserialize_with = "deser_base64"
-    )]
+    /// `value` is encoded as a base64url string.
+    #[serde(rename = "value", deserialize_with = "deser_base64")]
```
heaths, line 410: Ah, I see we're already doing that with this function. I forgot it was locally implemented.

heaths (approved):
Thank you!

LykxSassinator:
@heaths If there was no other advices or comments, could u help me to merge it? thx.

demoray:
Thanks for the contributions!

PR #1273, opened 2023-05-03
Fix clippy lints and enforce in CI

caass:
A couple small tidy-up changes to fix clippy warnings and enforce clippy in CI


caass:

.github/workflows/build.yml
```diff
@@ -16,6 +16,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: Swatinem/rust-cache@v2
       - run: eng/scripts/code_style.sh
+      - run: eng/scripts/clippy_all.sh
```
demoray, line 19: This should be handled via `eng/scripts/code_style.sh`.

PR #1277, opened 2023-05-11
Fix SAS token generation for directory-scoped access

dmweis:
Directory scoped SAS tokens require `signedDirectoryDepth` field to work. This adds it as an optional element.

Docs say "The value of the sdd field must be a non-negative integer." so I went with a u32 assuming no one would need a higher directory depth than that. Is that a sensible assumption?

I also added some tests but they aren't particularly inspired so I am happy to change/remove them

Links:
About directory scoped access: https://learn.microsoft.com/en-us/rest/api/storageservices/create-service-sas#service-sas-support-for-directory-scoped-access
Directory depth field: https://learn.microsoft.com/en-us/rest/api/storageservices/create-service-sas#specify-the-directory-depth

### DIFF
--- a/sdk/storage/src/shared_access_signature/service_sas.rs
+++ b/sdk/storage/src/shared_access_signature/service_sas.rs
@@ -196,6 +196,7 @@ impl SasToken for BlobSharedAccessSignature {
 #[cfg(test)]
 mod test {
     use super::*;
+    use std::collections::HashSet;
     use time::Duration;
 
     const MOCK_SECRET_KEY: &str = "RZfi3m1W7eyQ5zD4ymSmGANVdJ2SDQmg4sE89SW104s=";
@@ -215,10 +216,17 @@ mod test {
             BlobSignedResource::Blob,
         )
         .token();
+
+        // splitting by & is only safe if & is not part of any fields
+        // if that changes in the future we might want to use a proper query string parser
+        let elements = signed_token.split('&').collect::<HashSet<_>>();
+
         // BlobSignedResource::Blob
-        assert!(signed_token.contains("sr=b"));
-        // this assert will be a bad idea if we ever add a new field that ends with "sdd"
-        assert!(!signed_token.contains("sdd="));
+        assert!(elements.contains("sr=b"));
+        // signed_directory_depth NOT set
+        assert!(!elements.iter().any(|element| element.starts_with("sdd=")));
+
+        assert_eq!(signed_token, "sv=2020-06-12&sp=r&sr=b&se=1970-01-08T00%3A00%3A00Z&sig=alEGfKjtiLs5LO%2FyrfPkzjQBHbk4Uda9XOezbRyKwEM%3D");
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


sdk/storage/src/shared_access_signature/service_sas.rs
```diff
@@ -100,6 +100,7 @@ pub struct BlobSharedAccessSignature {
     identifier: Option<String>,
     ip: Option<String>,
     protocol: Option<SasProtocol>,
+    signed_directory_depth: Option<u32>, // sdd
```
demoray, line 103:
```suggestion
    signed_directory_depth: Option<usize>, // sdd
```

sdk/storage/src/shared_access_signature/service_sas.rs
```diff
@@ -128,6 +130,7 @@ impl BlobSharedAccessSignature {
         identifier: String => Some(identifier),
         ip: String => Some(ip),
         protocol: SasProtocol => Some(protocol),
+        signed_directory_depth: u32 => Some(signed_directory_depth),
```
demoray, line 133:
```suggestion
        signed_directory_depth: usize => Some(signed_directory_depth),
```

sdk/storage/src/shared_access_signature/service_sas.rs
```diff
@@ -179,9 +182,63 @@ impl SasToken for BlobSharedAccessSignature {
...
+            ..Default::default()
+        };
+        let signed_token = BlobSharedAccessSignature::new(
+            String::from(MOCK_SECRET_KEY),
+            String::from(MOCK_CANONICALIZED_RESOURCE),
+            permissions,
+            OffsetDateTime::UNIX_EPOCH + Duration::days(7),
+            BlobSignedResource::Directory,
+        )
+        .signed_directory_depth(2_u32)
```
demoray, line 237:
```suggestion
        .signed_directory_depth(2_usize)
```

demoray (changes requested):
Thanks for the contribution!  This looks positive.  

My only requested change is to use `usize` rather than `u32`, in most of the counting results (think [Iterator::count](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.count) or [String::len](https://doc.rust-lang.org/std/string/struct.String.html#method.len) return usize.

sdk/storage/src/shared_access_signature/service_sas.rs
```diff
@@ -179,9 +182,63 @@ impl SasToken for BlobSharedAccessSignature {
...
+            String::from(MOCK_CANONICALIZED_RESOURCE),
+            permissions,
+            OffsetDateTime::UNIX_EPOCH + Duration::days(7),
+            BlobSignedResource::Blob,
+        )
+        .token();
+        // BlobSignedResource::Blob
+        assert!(signed_token.contains("sr=b"));
+        // this assert will be a bad idea if we ever add a new field that ends with "sdd"
+        assert!(!signed_token.contains("sdd="));
```
demoray, line 221: As this example uses a constant expire date of 1970-01-08, the signed token result should always be the same.  What are your thoughts for just having a const for the expected result?

Otherwise, to address your comment, this could also naively split on `&`, as the SasToken in the example should not result in a `&` except as a delim.

dmweis: Good idea!

I added both better splitting and an assert against a hardcoded string. This way you get more clear failures if things break but have a wider test too.

If people add more tests later or other things break the hardcoded tests they can remove one or the other

demoray (approved):

dmweis:
Thanks for the approve @demoray ! 
I assume that since it's approved it's just going to be merged later? Or is there something else I need to do?

Issue #1276, opened 2023-05-11
PutBlockBlobBuilder doesn’t support conditional updates


Porges:
Other builders support `if_match` and its ilk; `PutBlockBlobBuilder` does not. At the REST API level [this is supported](https://learn.microsoft.com/en-us/rest/api/storageservices/specifying-conditional-headers-for-blob-service-operations#Subheading2).

JWT95:
I'm hitting this too. I'll look to raise a PR

PR #1280, opened 2023-05-15
Add missing conditional headers to blob and container operations

JWT95:
## Overview

Fix https://github.com/Azure/azure-sdk-for-rust/issues/1276 and add the remaining conditional headers missing from blob and container operations as per [the REST API documentation](https://learn.microsoft.com/en-us/rest/api/storageservices/specifying-conditional-headers-for-blob-service-operations#Subheading2).

## Testing
I tested the use case I care about, support for preventing duplicate blob writes, with the following code.

```rust
use azure_core::request_options::IfMatchCondition;
use azure_storage::prelude::StorageCredentials;
use azure_storage_blobs::prelude::BlobServiceClient;

#[tokio::main]
async fn main() {
    let token_credential = azure_identity::AzureCliCredential::default();
    let storage_credentials =
        StorageCredentials::token_credential(std::sync::Arc::new(token_credential));

    let blob_client = BlobServiceClient::new("<redacted>".to_string(), storage_credentials)
        .container_client("<redacted>")
        .blob_client("<redacted>");

    // Write blob
    let file = "Message One".as_bytes();
    blob_client
        .put_block_blob(file)
        .if_match(IfMatchCondition::NotMatch("*".into()))
        .await
        .expect("Failed to write message one");

    // Try to overwrite blob
    let file = "Message Two".as_bytes();
    blob_client
        .put_block_blob(file)
        .if_match(IfMatchCondition::NotMatch("*".into()))
        .await
        .expect("Failed to write message two");
}

```

The output was as expected:
```
thread 'main' panicked at 'Failed to write message two: Error { context: Full(Custom { kind: HttpResponse { status: Conflict, error_code: Some("BlobAlreadyExists") }, error: HttpError { status: Conflict, details: ErrorDetails { code: Some("BlobAlreadyExists"), message: None }...
```

### DIFF
--- a/sdk/storage_blobs/src/blob/operations/acquire_lease.rs
+++ b/sdk/storage_blobs/src/blob/operations/acquire_lease.rs
@@ -7,7 +7,10 @@ operation! {
     client: BlobClient,
     lease_duration: LeaseDuration,
     ?lease_id: LeaseId,
-    ?proposed_lease_id: ProposedLeaseId
+    ?proposed_lease_id: ProposedLeaseId,
+    ?if_modified_since: IfModifiedSinceCondition,
+    ?if_match: IfMatchCondition,
+    ?if_tags: IfTags
 }
 
 impl AcquireLeaseBuilder {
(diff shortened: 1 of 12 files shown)


JWT95:

demoray (approved):

demoray:
Thank you for the contributions!

PR #1282, opened 2023-05-18
Add response headers to PeekLockResponse.

jltorresm:
Service Bus Queue messages can have custom properties set (additionally to the standard properties set by the broker). Those are [mapped to headers](https://learn.microsoft.com/en-us/rest/api/servicebus/introduction) in the rest API.

In order to get those values we need to include the Headers as part of the response.

### DIFF
--- a/sdk/messaging_servicebus/Cargo.toml
+++ b/sdk/messaging_servicebus/Cargo.toml
@@ -23,9 +23,8 @@ hmac = "0.12"
 sha2 = "0.10"
 ring = "0.16"
 bytes = "1.0"
-serde = "1.0.163"
-serde_json = "1.0.96"
-
+serde = "1.0"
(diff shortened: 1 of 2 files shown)


jltorresm:
I don't oppose it in principle, but that wouldn't cover the custom properties. In the final section of the docs, it mentions:

> In addition to these properties, you can specify custom properties. If a single message is sent or received, each custom property is placed in its own HTTP header

Since these can have arbitrary names, I don't see how we could contain them struct. I tested custom properties on a dev queue and they are indeed set in their own headers.

jltorresm:
Let me know your thoughts, I'm happy to rework this towards a better solution.

jltorresm:

demoray (approved):

demoray (approved):

PR #1285, opened 2023-05-22
re-generate services

demoray:
Ref: https://github.com/Azure/azure-rest-api-specs/commit/529205bb71049de32c1126c69b6016904261da32


services/mgmt/arcdata/README.md
```diff
@@ -6,12 +6,12 @@ https://github.com/Azure/azure-rest-api-specs/blob/main/specification/azurearcda
 
 To get started with these generated service crates, see the [examples](https://github.com/Azure/azure-sdk-for-rust/blob/main/services/README.md#examples).
 
-The default tag is `package-preview-2023-03`.
```
johnbatty, line 9: Scanning through a selection of the changes, I was a little surprised to see some of the latest versions switch to an older version.  However, it seems that this is just due (in this case at least) to the newer API being removed, e.g. for `ArcData`:

- https://github.com/Azure/azure-rest-api-specs/pull/23821/files

So this appears to be fine.

services/mgmt/security/README.md
```diff
@@ -6,12 +6,12 @@ https://github.com/Azure/azure-rest-api-specs/blob/main/specification/security/r
...
-- `package-preview-2022-08` has 6 operations from 1 API versions: `2022-08-01-preview`. Use crate feature `package-preview-2022-08` to enable. The operations will be in the `package_preview_2022_08` module.
-- `package-preview-2022-07` has 8 operations from 1 API versions: `2022-07-01-preview`. Use crate feature `package-preview-2022-07` to enable. The operations will be in the `package_preview_2022_07` module.
\ No newline at end of file
+- `package-dotnet-sdk` has 197 operations from 20 API versions: `2015-06-01-preview`, `2017-08-01`, `2017-08-01-preview`, `2019-01-01`, `2019-01-01-preview`, `2019-08-01`, `2020-01-01`, `2020-01-01-preview`, `2021-01-15-preview`, `2021-05-01-preview`, `2021-06-01`, `2021-07-01-preview`, `2021-10-01-preview`, `2022-01-01`, `2022-01-01-preview`, `2022-03-01`, `2022-05-01`, `2022-07-01-preview`, `2022-08-01-preview`, `2023-02-01-preview`. Use crate feature `package-dotnet-sdk` to enable. The operations will be in the `package_dotnet_sdk` module.
```
johnbatty, line 13: Worth noting that there are some significant changes to the security package names.  I'm presuming that this is fine - it is autogenerated and builds.

johnbatty (approved):
Just a couple of observations, but think it looks good.

services/mgmt/security/README.md
demoray: This is a little odd, but something we can address in later updates.

PR #1286, opened 2023-05-22
bump versions 0.12.0

demoray:
This is built upon #1285.

Review of this PR should either be focused on the second commit to this PR or wait until the aforementioned PR is merged.


sdk/iot_deviceupdate/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "azure_iot_deviceupdate"
-version = "0.1.0"
+version = "0.12.0"
```
johnbatty, line 3: It appears that `azure_iot_deviceupdate` has not previously been published/updated.
I see that there was an earlier PR to disable publishing: https://github.com/Azure/azure-sdk-for-rust/pull/1138
Have you made a subsequent change to re-enable updating and/or publishing (if it is now ok to be published)?

johnbatty (approved):
Just one observation.  Looks good.

sdk/iot_deviceupdate/Cargo.toml
demoray: This crate is still not ready to be published, but as we standardized on a single version across all of the crates, this has no negative impact when we move forward to release this crate eventually.

PR #1291, opened 2023-06-06
Log unrecongnized resource quota parts instead of returning an error

### LINKED ISSUE #1137: azure_data_cosmos error: resource quota has an unrecognized part
I was testing out the azure_data_cosmos crate with some simple code, lightly adapted from the examples:
```rust
let account = "...";
let key = "...";
let token = AuthorizationToken::primary_from_base64(key)?;
let client = CosmosClient::new(account, token);

let mut dbs = client.list_databases().into_stream();

while let Some(db) = dbs.try_next().await? {
    for d in db {
        println!("{}", d.id);
    }
}
```
Both versions 0.6.0 and 0.7.0 of the crate produce the following error when running:
```
Error: Error { context: Message { kind: DataConversion, message: "resource quota has an unrecognized part - part: \"roleDefinitions=100\" full string: \"databases=1000;collections=5000;users=500000;permissions=2000000;clientEncryptionKeys=20;interopUsers=2000;authPolicyElements=200000;roleDefinitions=100;roleAssignments=2000;\"" } }
```
That seems to point pretty obviously at https://github.com/Azure/azure-sdk-for-rust/blob/4c60222bf1d56d6af87e6e8a555f93fe5cc501e5/sdk/data_cosmos/src/resource_quota.rs#L86 as the source of the error. I pulled the code down locally and commented out the whole else block for the sake of quick testing. With that change my code above works as expected (i.e. I get a list of databases from the account).

tcallan:
Fixes #1137

An alternate solution (or possible addition to this change) would be to capture the resource quota parts that were triggering the error. The parts in question are not useful to me and I'm not sure if they are generally useful so I didn't make that change.

### DIFF
--- a/sdk/data_cosmos/src/resource_quota.rs
+++ b/sdk/data_cosmos/src/resource_quota.rs
@@ -80,12 +80,6 @@ pub(crate) fn resource_quotas_from_str(
             v.push(ResourceQuota::InteropUsers(parseu64(stripped)?));
         } else if let Some(stripped) = token.strip_prefix(AUTH_POLICY_ELEMENTS) {
             v.push(ResourceQuota::AuthPolicyElements(parseu64(stripped)?));
-        } else {
-            return Err(Error::with_message(ErrorKind::DataConversion, || {
-                format!(
-                    "resource quota has an unrecognized part - part: \"{token}\" full string: \"{full_string}\""
-                )
-            }));
         }
 
         trace!("v == {:#?}", v);


sdk/data_cosmos/src/resource_quota.rs
```diff
@@ -81,11 +81,7 @@ pub(crate) fn resource_quotas_from_str(
         } else if let Some(stripped) = token.strip_prefix(AUTH_POLICY_ELEMENTS) {
             v.push(ResourceQuota::AuthPolicyElements(parseu64(stripped)?));
         } else {
-            return Err(Error::with_message(ErrorKind::DataConversion, || {
```
heaths, line 84: I can't exactly tell from context, but is this quote string returned by the service or from input by the caller?

heaths (changes requested):

sdk/data_cosmos/src/resource_quota.rs
tcallan: Do you mean the quota string that's being parsed by this method? If so, it comes from Cosmos (apparently part of the response headers to certain requests based current usages of `resource_quotas_from_str`).

heaths: Thanks. In case I agree it shouldn't be causing an error for callers.

JeffreyRichter:
I'm missing the full context on this, but I think not erroring is correct. I also think that not warning may also be correct. What is the value of warning? Would you also warn if an undocumented response header came back from the service?

tcallan:
FWIW, the warning is there mostly out of excessive caution, just in case it happens to be useful to someone, because I'm relatively unfamiliar with this code. I have zero objections to removing it.

heaths (approved):

heaths:
BTW, we squash-merge so your fork's `main` will be different than the upstream `main`. I recommend always creating a topic branch in your fork and open a PR to merge that into main, such that your `main` is always in sync - if not behind, but fast-forwardable - with upstream `main`.

PR #1292, opened 2023-06-07
make md5 optional

demoray:


johnbatty (approved):
Looks great to me - particularly the fact that the md5 dependency can be totally excluded via a feature flag.

Just one observation:
- This obviously isn't backward compatible, but as we are 0.x then there are no compatibility guarantees and the README also says so.  Plus it is pretty easy to add in the `md5` feature if users want it.  The alternative would be to add `md5` into the default features, but personally I think it is better as an opt-in feature.

demoray:
> Looks great to me - particularly the fact that the md5 dependency can be totally excluded via a feature flag.
> 
> Just one observation:
> 
> * This obviously isn't backward compatible, but as we are 0.x then there are no compatibility guarantees and the README also says so.  Plus it is pretty easy to add in the `md5` feature if users want it.  The alternative would be to add `md5` into the default features, but personally I think it is better as an opt-in feature.

Yep, the change to make this an "opt-in required" feature was intentional primarily because we're in a 0.x version.

PR #1301, opened 2023-06-13
security_keyvault: Fix continuation requests

marzullo:
The `set_query` function was overriding the query string when listing enough secrets to be Pageable, so the next_link query string was being replaced. This caused an infinite loop where the `ListSecretsBuilder` stream repeatedly returned the first 25 objects, never completing the subsequent paginated requests.

Please let me know if there are any other changes/suggestions.

### DIFF
--- a/sdk/security_keyvault/Cargo.toml
+++ b/sdk/security_keyvault/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "azure_security_keyvault"
-version = "0.12.1"
+version = "0.12.0"
 authors = ["Microsoft Corp."]
 description = "Rust wrapper around Microsoft Azure REST APIs for Azure Key Vault"
 repository = "https://github.com/azure/azure-sdk-for-rust"
--- a/sdk/security_keyvault/Cargo.toml
+++ b/sdk/security_keyvault/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "azure_security_keyvault"
-version = "0.12.0"
+version = "0.12.1"
 authors = ["Microsoft Corp."]
 description = "Rust wrapper around Microsoft Azure REST APIs for Azure Key Vault"
 repository = "https://github.com/azure/azure-sdk-for-rust"
--- a/sdk/security_keyvault/src/clients/keyvault_client.rs
+++ b/sdk/security_keyvault/src/clients/keyvault_client.rs
@@ -75,7 +75,7 @@ impl KeyvaultClient {
         let dt = OffsetDateTime::now_utc();
         let time = date::to_rfc1123(&dt);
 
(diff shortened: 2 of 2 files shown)


marzullo:

heaths (changes requested):

sdk/security_keyvault/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "azure_security_keyvault"
-version = "0.12.0"
+version = "0.12.1"
```
demoray, line 3: For the foreseeable future, we prefer to increment the crate versions all at once to reduce difficulties with incompatibility issues for different `azure_core`. 

```suggestion
version = "0.12.0"
```

sdk/security_keyvault/src/clients/keyvault_client.rs
```diff
@@ -75,7 +75,9 @@ impl KeyvaultClient {
         let dt = OffsetDateTime::now_utc();
         let time = date::to_rfc1123(&dt);
 
-        url.set_query(Some(API_VERSION_PARAM));
+        if url.query().is_none() || !url.query().unwrap().contains(API_VERSION_PARAM) {
+            url.set_query(Some(API_VERSION_PARAM));
+        }
```
demoray, line 80: The current implementation _always_ overrides the `api-version`.

I believe the intent of this PR is to set the query param `api-version` if it doesn't already exist in the URL, but if it does, use that version.  I think this is more in line with what you're trying to accomplish:

```suggestion
        if !url
            .query_pairs()
            .into_iter()
            .any(|(key, _)| key == API_VERSION_PARAM_NAME)
        {
            url.set_query(Some(API_VERSION_PARAM))
        }
```

demoray: Note, this suggestion assumes this change at the top of the file

Replace:
```rust
const API_VERSION_PARAM: &str = formatcp!("api-version={}", API_VERSION);
```

With this:
```rust
const API_VERSION_PARAM_NAME: &str = "api-version";
const API_VERSION_PARAM: &str = formatcp!("{}={}", API_VERSION_PARAM_NAME, API_VERSION);
```

demoray (changes requested):

sdk/security_keyvault/src/clients/keyvault_client.rs
demoray: @heaths , I believe this suggestion would address the concerns you highlighted in your comment.  Thoughts?

heaths: Our REST API guidelines say to override the api-version with whatever the client was configured unconditionally.

demoray:
@marzullo thank you identifying the issue and for the contribution.  We've merged #1396 that was based on this PR.

PR #1304, opened 2023-06-16
add customer encryption key for upload and download blob request

Leavrth:
support `x-ms-encryption-algorithm`, `x-ms-encryption-key` and `x-ms-encryption-key-sha256` for upload/download blobs in headers.


sdk/storage_blobs/src/blob/operations/get_blob.rs
```diff
@@ -43,6 +44,9 @@ impl GetBlobBuilder {
                 }
 
                 headers.add(this.lease_id);
+                if let Some(cpk_info) = &this.encryption_key {
+                    headers.add_ref(cpk_info);
+                }
```
demoray, line 49: Can you discuss this choice, instead of using 
```suggestion
                headers.add(this.encryption_key);
```

This is made possible by the following:
https://github.com/Azure/azure-sdk-for-rust/blob/be63d1eb8bcd8f760376fb4b71ad8a140261947f/sdk/core/src/headers/mod.rs#L25-L27

Leavrth: Thank you for your review.

The struct `CPKInfo` contains 3 headers: `EncryptionKey`, `EncryptionKeySha256` and `EncryptionAlgorithm`. That's because once the `EncryptionKey` is specified, the other 2 headers must be provided. (from azure-sdk-for-go)

The `this.encryption_key` is wrapped by `Option`, for which I find there has been an implement:
```
// ./sdk/core/src/headers/mod.rs:L25
impl<T> AsHeaders for Option<T>
where
    T: Header,
{
    type Iter = std::option::IntoIter<(HeaderName, HeaderValue)>;

    fn as_headers(&self) -> Self::Iter {
        match self {
            Some(h) => h.as_headers(),
            None => None.into_iter(),
        }
    }
}
```

The T has a trait bound `Header`, which has the function `name` and `value`. But the trait `Header` can not be implemented by `CPKInfo`. So I cannot directly use `headers.add(this.encryption_key);`.

Leavrth: @demoray  PTAL

Leavrth: Maybe we can just specify the `AsHeaders::Iter` to expand the scope. See the code as followes:
```
impl<T> AsHeaders for Option<T>
where
    T: AsHeaders<Iter = std::option::IntoIter<(HeaderName, HeaderValue)>>,
{
    type Iter = T::Iter;

    fn as_headers(&self) -> Self::Iter {
        match self {
            Some(h) => h.as_headers(),
            None => None.into_iter(),
        }
    }
}

```

Leavrth:
In the commit [c18fbdd](https://github.com/Azure/azure-sdk-for-rust/pull/1304/commits/c18fbdd1b79b28c32e929e30f067344377520a4c), I modify the `std::option::IntoIter<(HeaderName, HeaderValue)>` to `std::vec::IntoIter<(HeaderName, HeaderValue)>` so that we can integrate multiple headers into one type.

demoray (approved):

sdk/storage_blobs/src/options/encryption_key.rs
```diff
@@ -0,0 +1,67 @@
+use azure_core::headers::{self, AsHeaders, HeaderName, HeaderValue};
+
+const DEFAULT_ENCRYPTION_ALGORITHM: &'static str = "AES256";
```
demoray, line 3:
```suggestion
const DEFAULT_ENCRYPTION_ALGORITHM: &'str = "AES256";
```

sdk/storage_blobs/src/options/encryption_key.rs
```diff
@@ -0,0 +1,67 @@
+use azure_core::headers::{self, AsHeaders, HeaderName, HeaderValue};
+
+const DEFAULT_ENCRYPTION_ALGORITHM: &'str = "AES256";
```
demoray, line 3:
```suggestion
const DEFAULT_ENCRYPTION_ALGORITHM: &str = "AES256";
```

PR #1316, opened 2023-07-06
fix BlobClient::put_block_list

### LINKED ISSUE #1314: put_block_list fails after removing the md5 feature.
I have dependencies on
azure_storage = "0.13.0"
azure_storage_blobs = {version = "0.13.0", features = ["md5"]}
azure_core = "0.13.0"

I have a working client which uploads blobs to Azure storage.

I want to remove the md5 feature (as now allowed by https://github.com/Azure/azure-sdk-for-rust/pull/1292), but when I do so, I see put_block_list requests fail with error "header not found content-md5". Note that put_block requests succeed.

I think I can see the bug:

API docs for put_block and put_block_list say that if I omit the content-md5 header from the request then it will also be omitted from the response.
- https://learn.microsoft.com/en-us/rest/api/storageservices/put-block-list?tabs=azure-ad
- https://learn.microsoft.com/en-us/rest/api/storageservices/put-block?tabs=azure-ad

But looking at the code I see:
- PutBlockResponse.from_headers() allows for content_md5 to be optional.
- PutBlockListResponse.from_headers() still requires content_md5 to be present. This is the bug.

demoray:
Fixes #1314

### DIFF
--- a/sdk/storage_blobs/src/blob/operations/put_block_list.rs
+++ b/sdk/storage_blobs/src/blob/operations/put_block_list.rs
@@ -2,6 +2,7 @@ use crate::prelude::*;
 #[cfg(feature = "md5")]
 use azure_core::base64;
 use azure_core::{headers::*, prelude::*, RequestId};
+#[cfg(feature = "md5")]
 use azure_storage::{headers::content_md5_from_headers, ConsistencyMD5};
 use bytes::Bytes;
 use time::OffsetDateTime;
@@ -79,6 +80,7 @@ impl PutBlockListBuilder {
 pub struct PutBlockListResponse {
     pub etag: String,
     pub last_modified: OffsetDateTime,
+    #[cfg(feature = "md5")]
     pub content_md5: ConsistencyMD5,
     pub request_id: RequestId,
     pub date: OffsetDateTime,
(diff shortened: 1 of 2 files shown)


sdk/storage_blobs/src/blob/operations/put_block_list.rs
```diff
@@ -79,6 +80,7 @@ impl PutBlockListBuilder {
 pub struct PutBlockListResponse {
     pub etag: String,
     pub last_modified: OffsetDateTime,
+    #[cfg(feature = "md5")]
```
heaths, line 83: Is it idiomatic to change the shape of a model based on a feature? Should this maybe just be an `Option<ConsistencyMD5>` and always be `None` otherwise?

demoray: Probably not, though my implementation was based on the service implementation.  If the md5 is not sent in the request, the service should not send it in a response.

Happy to adjust the PR though.

heaths: Generally, we want all our SDKs to have the same public APIs regardless of targets/features/whatever. In .NET, for example, our public APIs are the same regardless of which TFMs we target.

heaths (approved):

PR #1319, opened 2023-07-17
implement federated_credential_flow and workload_identity_creds and add them to the environment_credential

### LINKED ISSUE #1305: Feature: Support kubernetes workload identities
Current workaround is to use an init container with 

```bash
az login --service-principal -u $AZURE_CLIENT_ID --allow-no-subscriptions -t $AZURE_TENANT_ID --federated-token $(cat $AZURE_FEDERATED_TOKEN_FILE) &&
az account get-access-token --query 'accessToken' --resource https://storage.azure.com --output tsv > /tmp/az-access-token.json  &&
cat /tmp/az-access-token.json
```

Code in Application:

```rust
    let bearer_token = std::env::var("BEARER_TOKEN").expect("BEARER_TOKEN not set");
    let storage_credentials = StorageCredentials::BearerToken(bearer_token);
```

Would be realy nice if the DefaultCredential would support this login flow.
If someone guides me where and how to do it, i would like to contribute and create a PR.

BR Simon

simon-an:
fixes #1305

### DIFF
--- a/sdk/identity/examples/federated_credential.rs
+++ b/sdk/identity/examples/federated_credential.rs
@@ -1,0 +1,48 @@
+use azure_identity::{authority_hosts, federated_credentials_flow};
+use url::Url;
+
+use std::env;
+use std::error::Error;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let client_id = env::var("CLIENT_ID").expect("Missing CLIENT_ID environment variable.");
+    let token = env::var("FEDERATED_TOKEN").expect("Missing FEDERATED_TOKEN environment variable.");
+    let tenant_id = env::var("TENANT_ID").expect("Missing TENANT_ID environment variable.");
+
+    let vault_name = std::env::args()
+        .nth(1)
+        .expect("please specify the vault name as first command line parameter");
+
+    let http_client = azure_core::new_http_client();
+    // This will give you the final token to use in authorization.
+    let token = federated_credentials_flow::perform(
+        http_client.clone(),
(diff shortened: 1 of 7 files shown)


simon-an:

heaths (changes requested):
Why does the environment_credential not work? Can we not fix that? In our other languages for Azure SDKs, we have only the environment_credential that works for both user and app registrations.

simon-an:
Sorry i am not sure what you mean. If you look at the pr, it is implemented in the environment credential. I can rename the pr if that helps :)

heaths:
> Sorry i am not sure what you mean. If you look at the pr, it is implemented in the environment credential. I can rename the pr if that helps :) Edit: Since this is my first contribution to this repo, i tried to stick as close to the current code as possible.

You added a public (`WorkloadIdentityCredential`](https://github.com/Azure/azure-sdk-for-rust/pull/1319/files#diff-2ad4468fd1fd66aac39f9410601f0b67dd7d796bf9a36db12c5b6a87c0b99d35) which we don't need or have in any other language. All support should be internal (however you choose to implement it) to the `EnvironmentCredential`.

heaths (approved):

demoray:
First, thank you for the contribution.

While the Azure SDK for Rust is not yet an official SDK, we do our best to follow the [Azure SDK development guidelines](https://azure.github.io/azure-sdk/general_introduction.html).

One of the principles we take to heart for this SDK is consistency across all of the SDKs.  We would prefer future contributions to consider this and other principles set forth in the above guidelines.

PR #1320, opened 2023-07-20
remove rust-toolchain.toml

demoray:
Instead of specifying the toolchain via `rust-toolchain.toml`, our CI pipelines specify the intended toolchain.  For contributor's purposes, this is out of sync with what we verify against for PRs and has contributed to confusion over time.

For users of the SDK, this isn't applied.  For those, we specify the toolchain needed via setting MSRV in the `Cargo.toml` for the specific crate.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,0 @@
-[toolchain]
-channel = "1.65.0"
-components = [ "rustfmt", "clippy" ]


heaths:
Rather than keeping various sources up to date, wouldn't it make more sense just to standardize the entire e2e flow on _rust-toolchain_ (or _rust-toolchain.toml_ - I just know a popular GH Actions task, which we don't use, still hasn't added support for the _.toml_ file)? A single source of truth?

cataggar (approved):

Issue #1322, opened 2023-07-24
Bug -- Content-MD5 header is not optional in PutPageResponse


juliusl:
According to docs this header can be missing, 
https://learn.microsoft.com/en-us/rest/api/storageservices/put-page?tabs=azure-ad

<html>
<body>

Content-MD5 | This header is returned so that the client can check for message content integrity. The value of this header is computed by Blob Storage. It's not necessarily the same as the value that's specified in the request headers. For version 2019-02-02 and later, this header is returned only when the request has this header.
-- | --

</body>
</html>

PR #1323, opened 2023-07-24
fix. content-md5 header in response is optional

juliusl:
Addresses #1322


demoray (approved):

demoray:
In the result:

```
pub content_md5: ConsistencyMD5,
```
This needs to be:
```
    pub content_md5: Option<ConsistencyMD5>,
```

juliusl:
@demoray sorry about that, I was being impatient.

PR #1324, opened 2023-07-25
Add blob committed block count for append blobs.

jetuk:
Add `x-ms-blob-committed-block-count` response header for use with append blobs.


jetuk:
@microsoft-github-policy-service agree company="James Tomlinson Associates Ltd"

demoray (approved):

PR #1325, opened 2023-07-29
Update client_credentials_flow.rs

bgdev-guy:
Clarified that the scope is a variable dependent on what you're attempting to access. As written, "https://management.azure.com/" could be assumed to be a static value and a naive attempt at testing  would result in a response of  

HttpResponse { status: BadRequest, error_code: None }


bgdev-guy:

heaths (approved):
Thank you.

demoray:
Thank you for the contribution.

PR #1326, opened 2023-07-29
Make the login_response module public

bgdev-guy:
Although LoginResponse is returned correctly as part of the client_credentials_flow::perform method. I can not directly declare the type, nor store it in a struct because it's not directly accessible to be defined.

i.e. 
use azure_identity::client_credentials_flow::login_response::LoginResponse;

will not work since login_response is private.  This change makes the login_response module public

### DIFF
--- a/sdk/identity/src/client_credentials_flow/mod.rs
+++ b/sdk/identity/src/client_credentials_flow/mod.rs
@@ -35,7 +35,7 @@
 //!
 //! You can learn more about this authorization flow [here](https://docs.microsoft.com/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow).
 
-mod login_response;
+pub mod login_response;
 
 use azure_core::Method;
 use azure_core::{


heaths:
Details of the different credential types should not be public. Can you provide example code why you need this to be exported?

bgdev-guy:
I'm implementing a POC by migrating an existing azure function app to Rust. The function processes events and executes a series of calls to one or more API's. Ideal scenario is to cache the token with the expiration time and refresh when appropriate. That data is available in the returned LoginRespons. However utilizing the LoginResponse is complicated by the fact that I can't explicitly state that I'm using it. I can't pass it to a function to extract what I need or store it in a struct, since it can only be inferred and not explicitly defined.

The LoginResponse and it's data is already being returned. Making the module public doesn't reveal or expose any additional structs, functions, or values. All it does is make it easier for me to implement a solution that fits my needs.

heaths:
> Ideal scenario is to cache the token with the expiration time and refresh when appropriate. That data is available in the returned LoginResponse.

In other languages, we do that in a pipeline policy e.g., https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/Azure.Security.KeyVault.Shared/src/ChallengeBasedAuthenticationPolicy.cs (extends https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/BearerTokenAuthenticationPolicy.cs). This isn't something that should be the responsibility of the caller.

And regardless of refetching a near-expiring token, tokens should already be cached by credentials. Are you finding that, specifically, is not the case for subsequent calls? I'm not eager to open up implementation details, but perhaps we could find an easier workaround for now since refreshing tokens may yet not be implemented.

bgdev-guy:
I feel like I'm missing something. You're referencing a pipeline that doesn't exist in this language. You keep saying implementation details and I am genuinely curious as to what you mean by that. What exactly is being revealed by making this module public? But all in all, I've just decided to go ahead and use a different solution that I can work with.

Issue #803, opened 2022-06-14
migrate SDKs to use generated services


cataggar:
The goal is to have the SDKs use the generated services. Ideally, they are a thin layer on top.

I'll fill in this description more later. The list of crates is pretty well covered in #717.

cataggar:
Not yet, we were going to try to figure that out as part of #733.

mario-guerra:
@cataggar will the service bus crate be included in this effort?

ctaggart:
@mario-guerra, there is not a generated `azure_svc_servicebus` crate because the [Service Bus Runtime REST API](https://docs.microsoft.com/en-us/rest/api/servicebus/service-bus-runtime-rest) does not have a published specification https://github.com/Azure/azure-rest-api-specs/issues/16429. If it was documented, would we have a REST API implementation and an AMQP implementation of overlapping functionality? There has been some recent work geting an AMQP implemeantion to work with Service Bus in https://github.com/Azure/azure-sdk-for-rust/issues/643 by @minghuaw.

reenigneEsrever92:
Regarding the generated sources - I've found the following example code in svc/keyvault under examples:

```rust
#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let credential = Arc::new(AzureCliCredential::new());
    let keyvault_name = std::env::args().nth(1).expect("please specify an existing keyvault");
    let key_name = std::env::args().nth(2).expect("please specify the name of the key to create");
    let endpoint = format!("https://{keyvault_name}.vault.azure.net");
    let scopes = &["https://vault.azure.net"];
    let client = azure_svc_keyvault::Client::builder(credential)
        .endpoint(endpoint)
        .scopes(scopes)
        .build();
```

for the scope the it defines `https://vault.azure.net` yet this is the resource, ain't it? Is there a reason for it? Why wouldn't it define `.default` for the scope for example and set `https://vault.azure.net` for the (currently non existent) resource field?

PR #1327, opened 2023-07-30
Feat/keystore shim

reenigneEsrever92:
Maybe this could serve as a discussion basis for: https://github.com/Azure/azure-sdk-for-rust/issues/803


sdk/keyvault/src/keyvault_client.rs
```diff
@@ -0,0 +1,39 @@
...
+use azure_core::auth::TokenCredential;
+use azure_svc_keyvault::{
+    create_key::Response,
+    models::{
+        key_create_parameters::{Crv, Kty},
+        KeyAttributes, KeyCreateParameters, KeyReleasePolicy,
+    },
+};
+
+pub struct KeyvaultClient {
```
heaths, line 12: We already have Key Vault clients for keys, secrets, and certs. Why add another one that isn't scoped like all our other languages? We want our SDKs to be idiomatic, but still similarly designed.

heaths (changes requested):

sdk/keyvault/src/keyvault_client.rs
cataggar: This looks to be for #803, where we look to migrate our SDKs to be on top of the generated service crates.

heaths: Why is it under `sdk/keyvault` then? The structures of SDKs should remain idiomatic but consistent with all other languages' Azure SDKs.

cataggar:
Thanks for this prototype. Yes, we want to modifying the SDK `azure_security_keyvault` to use the generated `azure_svc_keyvault` crate.

PR #1330, opened 2023-08-01
Do not add timeout header if it has already been added

maxburke:
When an operation is retried, Azure SDK For Rust will re-add parameters to the query url -- this results in multiple values for the same parameter key and, at least in the case of the timeout value, will cause Azure to error out with an invalid value response (http error 400)

### DIFF
--- a/sdk/core/src/request_options/timeout.rs
+++ b/sdk/core/src/request_options/timeout.rs
@@ -12,6 +12,10 @@ impl Timeout {
 
 impl AppendToUrlQuery for Timeout {
     fn append_to_url_query(&self, url: &mut url::Url) {
+        if url.query_pairs().any(|(k, _)| k == "timeout") {
+            return;
+        }
+
         url.query_pairs_mut()
             .append_pair("timeout", &format!("{}", self.0.as_secs()));
     }


heaths:
> Azure SDK For Rust will re-add parameters to the query url

@maxburke do you see this is a general problem, or only for the timeout header? If the former, we should solve it generically if possible.

maxburke:
I don't know? I've only been seeing it with the timeout header.

sdk/core/src/request_options/timeout.rs
```diff
@@ -12,6 +12,10 @@ impl Timeout {
 
 impl AppendToUrlQuery for Timeout {
     fn append_to_url_query(&self, url: &mut url::Url) {
+        if url.query_pairs().any(|(k, _)| k == "timeout") {
+            return;
+        }
```
heaths, line 17: Wouldn't it make more sense to update the value if it is present? You could short-circuit string reallocation if it's the same value.

heaths (changes requested):

sdk/core/src/request_options/timeout.rs
maxburke: As far as I can tell from the API for URL, you can't update the query pairs, only add.

heaths (approved):

heaths (approved):
The build error is unrelated and needs to be fixed independently.

maxburke:
Thank you, @heaths !

PR #1331, opened 2023-08-09
Fix: Authorization Policy does applies SAS token twice on retry requests 

dmweis:
# Background 

I encountered a bug where `put_block` operations would sporadically fail with http status code `403 - Forbidden` when authenticating using a SAS token.
 
Detailed of the error were:
`"AuthenticationFailed"`
`"Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature."`
`<AuthenticationErrorDetail>Signature fields not well formed.</AuthenticationErrorDetail>`

Upon investigating I narrowed it down to the Authorization Policy applying SAS token params multiple times for retry requests.

Simple way to diagnose this issues when using `reqwest` as the url library is to run Env Logger filter with `"azure_core::http_client::reqwest=debug"` which will cause `reqwest` to print the full URL for all requests.

# Details

I used the same approach as the Access Key policy already uses and simply check the presence of `sig` param in the path. This doesn't feel like a very foolproof solution but it fixes the issue for now.

It seems that the python SDK suffered from a similar issues and they chose the same approach for resolving it https://github.com/Azure/azure-storage-python/pull/304

The tests I added feel a little awkward because I had to mock the policy setup. If there is an already existing mock that would allow me to test this easily I am happy to update the PR!

I don't believe this should be an issue for header based authentication since `insert_header` would override an existing header if it was present

### DIFF
--- a/sdk/storage/src/authorization/authorization_policy.rs
+++ b/sdk/storage/src/authorization/authorization_policy.rs
@@ -52,6 +52,7 @@ impl Policy for AuthorizationPolicy {
                 request
             }
             StorageCredentials::SASToken(query_pairs) => {
+                // Ensure the signature param is not already present
                 if !request.url().query_pairs().any(|(k, _)| &*k == "sig") {
                     request
                         .url_mut()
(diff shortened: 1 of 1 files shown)


heaths (approved):

dmweis:
I see there are CI failures but I think that's due to https://github.com/Azure/azure-sdk-for-rust/issues/1332

dmweis:
> @dmweis can you rebase from `main` to pull in fixes for #1332

Done!

PR #1335, opened 2023-08-16
unify management of MSRV

demoray:
This PR standardizes where all crates set the MSRV (azure_core), as well as updates the actions to build on a matrix of 'stable', 'nightly', and the rust-version set in `azure_core`.

This allows developers of the SDK to run each of the build tasks exactly as they are run in CICD with minimal configuration.


sdk/core/Cargo.toml
```diff
@@ -11,7 +11,7 @@ documentation = "https://docs.rs/azure_core"
 keywords = ["sdk", "azure", "rest", "iot", "cloud"]
 categories = ["api-bindings"]
 edition = "2021"
-rust-version = "1.64.0"
+rust-version = "1.67.0"
```
cataggar, line 14: Just one place. Loos like the right spot.

cataggar (approved):
Thanks for the infrastructure work.

PR #1344, opened 2023-08-24
address lints from clippy 1.72.0

demoray:


johnbatty (approved):
It's a shame we have to bump the minimum Rust version here - may make it harder for some people to take the update (although I'm fully in favour of keeping the Rust toolchain up to date, there will obviously be potential extra work/issues getting the toolchain updated).

I guess that's needed to be able to specify `allow(clippy::<new lint name>)`?

Is it possible to do something like this?
```
#[rustversion::since(1.70)]
#[allow(clippy::extra_unused_type_parameters)]
```
Then again, perhaps it is not worth this extra complication.

cataggar (approved):

demoray:
@johnbatty the combination of one of the prereqs (that I don't recall at this time) bumping their minimum version as well as the new lint names, this was easier.  I think for future clippy lints, we should consider your suggestion for future breaking lints.

PR #1353, opened 2023-09-05
use newer workspace resolver

demoray:
As we use `edition=2021` in a handful of places, we should align the resolver in the Cargo.toml for the workspace to use the same resolver.

Ref: https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html


heaths (approved):
Interesting. Thanks for the link. I was not aware of this resolver change (or even the old behavior, really), and it makes a lot of sense.

PR #1362, opened 2023-09-07
add support for listing system containers when listing containers

demoray:
From the documentation:
> -system: Version 2020-10-02 and later. Specifies if system containers are to be included in the response. Including this option will list system containers, such as $logs and $changefeed. Note that the specific system containers returned will vary, based on which service features are enabled on the storage account.

ref: https://learn.microsoft.com/en-us/rest/api/storageservices/list-containers2?tabs=azure-ad


demoray:
Note for whomever reviews this: As this required bumping the API version, the transactions used during mock testing had to be rebuilt.  The changes to the API are in the first commit, the regenerating of the transaction records is the second commit.

johnbatty (approved):

PR #1363, opened 2023-09-07
use access keys for storage blob integration tests

demoray:
The majority of the storage account examples use access keys, as do the ADL2 integration tests.  This aligns the blob storage integration tests to follow the same model.


johnbatty (approved):
Looks good

PR #1128, opened 2022-10-18
WIP: Prototype azure_core WASM support via reqwest

johnbatty:
In a recent hackathon I managed to get `azure_core` compiled to WASM and making requests from a browser.  This made use of `reqwest`'s WASM support but did require a couple of hacks to work around the fact that it doesn't yet support streaming responses (although there is an open PR which does appear to add this: https://github.com/seanmonstar/reqwest/pull/1576).

Not sure whether you want to take any part of this, but thought I'd make it available in case you are interested!

The main changes are:
- re-enable inclusion of `reqwest` and creation of a `reqwest` http client in the `http_client` module when the target is `wasm32`
- for `wasm32` target omit the `Send` trait when using `async_trait`
  - https://github.com/rustwasm/wasm-bindgen/issues/2409
- for `wasm32` target remove support for `Body::SeekableStream`
- when receiving a response read all the bytes and wrap it up in a "stream wrapper" (due to lack of a `bytes_stream(...)` function)
  - this is the hackiest part!  Any alternative suggestions welcomed!
- for `wasm32` target remove `Send + Sync` from the `PinnedStream` definition
- add a new feature `wasm-bindgen` that enables `time/wasm-bindgen`
  - if this `time` feature is not enabled then the retry policy code triggers a panic when querying the current time

PR #1367, opened 2023-09-11
wasm32 target support (v2)

johnbatty:
This PR replaces my earlier draft wasm PR: https://github.com/Azure/azure-sdk-for-rust/pull/1128

The main changes are as follows.

## azure_core

- enable inclusion of `reqwest` and creation of a reqwest http client in the `http_client` module when the target is wasm32
- for `wasm32` target omit the Send trait when using async_trait by specifying `async_trait(?Send)`
   - https://github.com/rustwasm/wasm-bindgen/issues/2409
- use newly added `reqwest` support (0.11.14) for `wasm` implementation of `Response::bytes_stream()`
- for `wasm32` target remove support for `Body::SeekableStream`
  - this currently can't be implemented as `reqwest::Body::wrap_stream()` is not implemented for WASM
  - I think we should accept this as a `wasm32` limitation for now
- for `wasm32` target remove Send + Sync from the `PinnedStream` definition
- add a new feature `wasm-bindgen` that enables `time/wasm-bindgen` to allow the UTC time to be queried via the browser API
  - if this time feature is not enabled then the retry policy code triggers a panic when querying the current time

## autorust
 
Update `autorust` to use `LocalBoxFuture` for `wasm32` target rather than `BoxFuture`

## Autogenerated services

Regenerate with modified `autorust`


sdk/core/src/http_client/reqwest.rs
```diff
@@ -2,6 +2,7 @@ use crate::error::{ErrorKind, ResultExt};
 use crate::{Body, HttpClient, PinnedStream};
 
 use async_trait::async_trait;
+//#[cfg(not(target_arch = "wasm32"))]
```
demoray, line 5: Did you intend to leave this comment?

sdk/core/src/http_client/reqwest.rs
```diff
@@ -24,9 +26,18 @@ impl HttpClient for ::reqwest::Client {
...
+            #[cfg(not(target_arch = "wasm32"))]
             Body::SeekableStream(seekable_stream) => req
                 .body(::reqwest::Body::wrap_stream(seekable_stream))
                 .build(),
+
+            // Cannot currently implement `Body::SeekableStream` for WASM
+            // because `reqwest::Body::wrap_stream()` is not implemented for WASM.
+            #[cfg(target_arch = "wasm32")]
+            Body::SeekableStream(_seekable_stream) => {
+                todo!("Body::SeekableStream is not currently supported for WASM")
```
demoray, line 39: Instead of using `todo!` here, can the `SeekableStream` variant of `Body` be only defined in via `not(target_arch = "wasm32")`?

Failing at compile-time is much less surprising than failing at runtime due to a missing platform feature.

demoray (approved):

sdk/core/Cargo.toml
```diff
@@ -54,3 +54,4 @@ enable_reqwest_rustls = ["reqwest/rustls-tls"]
 test_e2e = []
 azurite_workaround = []
 xml = ["quick-xml"]
+wasm-bindgen = ["time/wasm-bindgen"]
```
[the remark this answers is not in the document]
johnbatty: My original thinking was that the `wasm32` target was not necessarily tied to use of `wasm-bindgen` (e.g. if you want to use `wasi` rather than run in a browser).  However, for now this will be the case (`wasi` support seems quite a long way off..), so I've changed to automatically set the `time` crate `wasm-bindgen` feature on `wasm32` builds.  Unfortunately this is a little ugly - have to specify the time dependency twice with different feature sets (I couldn't find a better way to do it).

sdk/core/src/http_client/reqwest.rs
johnbatty: Nope - removed.

sdk/core/src/http_client/reqwest.rs
johnbatty: Yep - fixed.  I obviously had to add `wasm32` conditionals to all the places where `SeekableStream` was used.

sdk/core/Cargo.toml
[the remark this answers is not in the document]
demoray: What are your thoughts on the following?  Due to features being additive, you could leave the original `time` definition, then use this:
```toml
[target.'cfg(target_arch = "wasm32")'.dependencies]
time = { version = "0.3.10", features = ["wasm-bindgen"] }
```

cataggar (approved):
This is great!

sdk/core/Cargo.toml
```diff
@@ -32,13 +32,11 @@ url = "2.2"
...
 pin-project = "1.0"
 paste = "1.0"
-
-[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 time = { version = "0.3.10", features = ["serde-well-known", "macros", "local-offset"] }
 
 # When target is `wasm32`, include `getrandom` and enable `wasm-bindgen` feature in `time`.
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-time = { version = "0.3.10", features = ["serde-well-known", "macros", "local-offset", "wasm-bindgen"] }
+time = { version = "0.3.10", features = ["wasm-bindgen"] }
```
cataggar, line 40: This is easier to read.

PR #1368, opened 2023-09-11
Wrap DeviceCodePhaseOneResponse::stream response in `Pin<Box>>`

demoray:
In order to use the stream, it needs to be a Pinned box.

This change aligns the code with other places in the SDK that return Streams, where we handle pinning the stream for the caller.

### DIFF
--- a/sdk/identity/src/device_code_flow/mod.rs
+++ b/sdk/identity/src/device_code_flow/mod.rs
@@ -13,8 +13,7 @@ use azure_core::{
 pub use device_code_responses::*;
 use futures::stream::unfold;
 use serde::Deserialize;
-use std::time::Duration;
-use std::{borrow::Cow, sync::Arc};
+use std::{borrow::Cow, pin::Pin, sync::Arc, time::Duration};
 use url::{form_urlencoded, Url};
 
 /// Start the device authorization grant flow.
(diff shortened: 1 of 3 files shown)


sdk/identity/src/device_code_flow/mod.rs
```diff
@@ -90,76 +89,87 @@ impl<'a> DeviceCodePhaseOneResponse<'a> {
...
-                            let rsp_status = rsp.status();
-                            let rsp_body = match rsp.into_body().collect().await {
-                                Ok(b) => b,
-                                Err(e) => return Some((Err(e), NextState::Finish)),
-                            };
-                            if rsp_status.is_success() {
-                                match serde_json::from_slice::<DeviceCodeAuthorization>(&rsp_body) {
-                                    Ok(authorization) => {
-                                        Some((Ok(authorization), NextState::Finish))
+        Box::pin(unfold(
```
heaths, line 100: So the returned `Iterator` will just run forever on any caller running `next()` (apart from one error case I see). Shouldn't we guard against that, or is that based into futures' `unfold`? I admit I'm only somewhat familiar with `itertools::unfold`.

demoray: As it relates to this PR, the only change in this PR is wrapping existing call to `unfold` with `Box::pin()`.

That said, I believe this has a handful of cases where `next()` will stop.

1. If the service returns success, but the body can't be read
2. If the service returns success and the body doesn't parse as intended
3. If the service returns non-success and the body doesn't parse as intended
4. If the service returns non-success and the response doesn't include `authorization_pending`
5. If the request itself fails (below the HTTP request layer)

I think it would be reasonable to add additional loop termination (such as maximum time we'll poll the service), but given the iterating over the stream is up to the caller, is this something that should be done in the SDK?

Either way, I think it would be appropriate to make such a change in a future PR.

heaths (approved):

PR #1380, opened 2023-09-14
make NoopClient feature dependant

demoray:
Instead of always compiling in `NoopClient`, make it something that's only available if the `reqwest` features were not enabled.


heaths (approved):
The code looks fine, but I don't understand what point is there in using the SDK if you have a client that does nothing? What's the intended use case? More of a meta question, though, and shouldn't block this PR. We can discuss on Teams, or here on this GitHub PR/issue.

demoray:
The original intent was to enable working towards supporting wasm.  For reference, this work was originally done in #939.   With the changes in #1367, we likely can look to get rid of `NoopClient` soon.

Issue #1366, opened 2023-09-08
Blob Service Example fails with XML parse error


joverwey:
I am trying to run the this example: `services\svc\blobstorage\examples\list_containers.rs` but I am getting an error:
`page error: failed to deserialize the following xml into a azure_svc_blobstorage::package_2021_12::models::ListContainersSegmentResponse`
```

<?xml version="1.0" encoding="utf-8"?>
<EnumerationResults ServiceEndpoint="https:///XXXX.blob.core.windows.net/">
    <Containers>
        <Container>
            <Name>azureml</Name>
            <Properties>
                <Last-Modified>Fri, 08 Sep 2023 16:36:39 GMT</Last-Modified>
                <Etag>"XXXX"</Etag>
                <LeaseStatus>unlocked</LeaseStatus>
                <LeaseState>available</LeaseState>
                <DefaultEncryptionScope>$account-encryption-key</DefaultEncryptionScope>
                <DenyEncryptionScopeOverride>false</DenyEncryptionScopeOverride>
                <HasImmutabilityPolicy>false</HasImmutabilityPolicy>
                <HasLegalHold>false</HasLegalHold>
                <ImmutableStorageWithVersioningEnabled>false</ImmutableStorageWithVersioningEnabled>
            </Properties>
        </Container>
        <Container>
            <Name>azureml-blobstore-XXXX</Name>
            <Properties>
                <Last-Modified>Fri, 08 Sep 2023 16:36:39 GMT</Last-Modified>
                <Etag>"XXXX"</Etag>
                <LeaseStatus>unlocked</LeaseStatus>
                <LeaseState>available</LeaseState>
                <DefaultEncryptionScope>$account-encryption-key</DefaultEncryptionScope>
                <DenyEncryptionScopeOverride>false</DenyEncryptionScopeOverride>
                <HasImmutabilityPolicy>false</HasImmutabilityPolicy>
                <HasLegalHold>false</HasLegalHold>
                <ImmutableStorageWithVersioningEnabled>false</ImmutableStorageWithVersioningEnabled>
            </Properties>
        </Container>
    </Containers>
    <NextMarker />
</EnumerationResults>
```

I know that I could use the blob storage API instead, but I am getting the same error using the filestorage generated API, which is actually what I am interested in since it does not have an SDK equivalent.

demoray:
I can replicate your issue.  Investigating.

demoray:
[...]

PR #1381, opened 2023-09-15
address XML parsing issues in generated crates

demoray:
This addresses a handful of XML parsing issues in generated crates.
1. Deserializing enum variants with the value being a string content (such as `<field>Value</field>`) needs a serde helper.
2. Deserializing attribute values (such as `<a b="foo"/>`) needs to the attribute name to be prefixed with `@` to indicate it's an attribute
3. Continuation tokens that are strings need to be considered the same as not having a continuation token

Additionally, this PR includes the following:
1. Regenerating the `azure_svc_blobstorage` crate for review
2. A unit test for `azure_svc_blobstorage` that validates the XML parsing works as intended
3. Updates to the `azure_svc_blobstorage` example such that it no longer silently swallows parsing issues

Once this PR is reviewed, I will submit a follow-on PR that includes regenerating the rest of the crates.

Ref: <https://docs.rs/quick-xml/latest/quick_xml/serde_helpers/text_content/index.html>
Ref: <https://docs.rs/quick-xml/latest/quick_xml/de/index.html#mapping-xml-to-rust-types>

NOTE: While this fixes the parsing error indicated by #1366, the `Continuable` functionality used by `azure_svc_storageblob` is currently broken.  That will be addressed in a follow-on PR.

### DIFF
--- a/sdk/core/src/xml.rs
+++ b/sdk/core/src/xml.rs
@@ -1,4 +1,5 @@
 use crate::error::{ErrorKind, ResultExt};
+pub use quick_xml::serde_helpers::text_content;
 use quick_xml::{
     de::{from_reader, from_str},
     se::to_string,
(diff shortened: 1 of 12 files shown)


heaths (approved):

cataggar (approved):
I really love to see this!

Issue #225, opened 2021-04-09
Automatically box use of struct as a field in that struct


ctaggart:
As hinted in #84, we should (when possible) automatically box types when they are used as the type of their own field. 

For example, it should be possible to detect the following situation:

```rust
struct Foo {
   bar: Bar,
   foo: Foo
}
```

and turn it into:

```rust
struct Foo {
   bar: Bar,
   foo: Box<Foo>
}
```

cataggar:
Fixed by #1385.

PR #1385, opened 2023-09-15
start automatically boxing types for code generation

demoray:
This addresses members that are directly recursive.  It does not handle nested recursion.

Partial fix for #225


cataggar (approved):
This helps a lot of releases.

PR #1393, opened 2023-09-18
add `models` to autorust's reserved words

### LINKED ISSUE #1382: Missing latest package-2023-05 api version for CognitiveServices
Thanks for the hard work that is put into this repo!

I noticed that `DeploymentProperties` do not have the [rateLimits](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/stable/2023-05-01/cognitiveservices.json#L3543C10-L3543C20) field, and then I found that the `package-2023-05` version is not available (and [denied](https://github.com/Azure/azure-sdk-for-rust/blob/main/services/mgmt/cognitiveservices/autorust.toml)). Will this version be generated soon(ish)?

demoray:
Because we use `models` heavily throughout to describe where the generated structs are stored, services that use `models` as a identifier name will cause issues.

By adding `models` to the pre-existing `is_keyword` and using a something that doesn't check `is_keyword` for when we _actually_ want `models`, we can generate code for services that use the keyword `models`.

Fixes #1382 (once the services are re-generated after this PR is merged).

### DIFF
--- a/services/autorust/codegen/src/codegen.rs
+++ b/services/autorust/codegen/src/codegen.rs
@@ -1,4 +1,8 @@
-use crate::{identifier::parse_ident, spec::TypeName, CrateConfig, PropertyName, Spec};
+use crate::{
+    identifier::{parse_ident, raw_str_to_ident},
+    spec::TypeName,
+    CrateConfig, PropertyName, Spec,
+};
 use crate::{Error, Result};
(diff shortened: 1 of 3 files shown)


cataggar (approved):
perfect

Issue #1346, opened 2023-08-30
Azure storage: support Get User Delegation Key


KodyKantor:
When an Azure Storage account is accessed via Azure AD the client secret cannot be used to generate SAS tokens. Instead, the clients are expected to retrieve a short-lived user delegation key.

https://learn.microsoft.com/en-us/rest/api/storageservices/get-user-delegation-key

The user delegation key can later be used to create user delegation SAS tokens.

The user delegation key expires occasionally. Ideally the SDK would additionally refresh this token in the background as necessary.

KodyKantor:
Related to #1298.

CC @mario-guerra

demoray:
This is now available in 0.16.0

https://docs.rs/azure_storage_blobs/latest/azure_storage_blobs/prelude/struct.BlobServiceClient.html#method.get_user_deligation_key

PR #1412, opened 2023-09-20
add support to fetch user delegated keys

demoray:
A first step in creating user-delegated SAS signatures (#1346) is fetching a user delegation key.

This PR implements fetching the key from the service.  Note, this adds the `ISO-8601` format as used by Azure Storage, which is close but not exactly `RFC-3339`.

Ref: https://learn.microsoft.com/en-us/rest/api/storageservices/get-user-delegation-key


cataggar (approved):

demoray:
This is now available in 0.16.0.

https://docs.rs/azure_storage_blobs/latest/azure_storage_blobs/prelude/struct.BlobServiceClient.html#method.get_user_deligation_key

PR #1416, opened 2023-09-22
add parallelism to autorust using rayon

demoray:
This PR does two things:

1. Moves to iterating over the crates to be generated via `rayon` for multi-processing
2. Moves to formatting the rust code using rustfmt via stdin/stdout rather than `cargo fmt` at the end.

This speeds up generation by 390% on my workstation.


cataggar (approved):
On my Azure DevBox, here are the results.

Before on `main`:
cargo build --release
timeit cargo run -p azure-autorust --release
6min 20sec 970ms 842µs 400ns

With this PR:
cargo build --release
timeit cargo run -p azure-autorust --release
1min 42sec 565ms 421µs 900ns

This is wonderful!

PR #1418, opened 2023-09-25
mock_api code generation 2023-09

cataggar:
This merged in mock_api code generation from `mock_api` branch to the latest.

PR #1419, opened 2023-09-25
rename ParameterType to ParameterIn in autorust

cataggar:
This upstreams changes that I use to [generate some server side code](https://github.com/Azure/azure-sdk-for-rust/pull/1418).
- rename ParameterType to ParameterIn in autorust
- add TypeNameCode::reference
- remove builder pattern from TypeNameCode


services/autorust/codegen/src/codegen.rs
```diff
@@ -166,71 +168,102 @@ impl TypeNameCode {
         Ok(type_name_code)
     }
 
+    pub fn set_is_ref(&mut self, is_ref: bool) {
+        self.is_ref = is_ref;
+    }
```
demoray, line 173: What are your thoughts on using the builder-style pattern here?

```suggestion
    pub fn set_is_ref(mut self, is_ref: bool) -> Self {
        self.is_ref = is_ref;
        self
    }
```

This mirrors other functions used in the impl of this struct, such as this (which is a few lines below):
```rust
    pub fn optional(mut self, optional: bool) -> Self {
        self.optional = optional;
        self
    }
```

services/autorust/codegen/src/codegen.rs
```diff
@@ -130,6 +130,8 @@ pub fn parse_query_params(uri: &str) -> Result<HashSet<String>> {
 
 #[derive(Clone)]
 pub struct TypeNameCode {
+    /// Whether or not to pass a type as a reference.
+    is_ref: bool,
```
demoray, line 134: The other bools in this struct are not prefaced with `is_`.  Consider `optional`, which similarly could be named `is_optional`.    What are your thoughts on naming this `reference` to follow the convention, and the related function `pub fn reference(..., reference: bool)`?

services/autorust/codegen/src/codegen.rs
cataggar: For this type, I'd like to move away from the builder-style pattern. I don't think it is needed. Mixing the two is not great, so I'll proceed with removing it if that works.

demoray: I don't have strong feelings either way.  I probably should have labeled these "questions/nit".

demoray (approved):

PR #1414, opened 2023-09-21
add support for unions in generated service serialization

cataggar:
Fixes #1394. It adds new types with `Union` suffix added to represent the discriminated unions.

PR #1435, opened 2023-10-04
fix union types

cataggar:
This fixes a bug from #1414 which causes deserialization to fail.
- create the base type without the discriminator
- adds the description to the union type
- fixes union types for vector aliases
- fixes support for nested unions
- removes empty base types


services/mgmt/alertsmanagement/src/package_preview_2021_08/models.rs
```diff
@@ -5,56 +5,13 @@ use serde::{Deserialize, Serialize, Serializer};
 use std::str::FromStr;
 #[doc = "Action to be applied."]
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-pub struct Action {
-    #[doc = "Action that should be applied."]
-    #[serde(rename = "actionType")]
-    pub action_type: action::ActionType,
-}
+pub struct Action {}
```
demoray, line 8: This doesn't feel correct.  I've checked a handful of these, and found each I've checked look to generate empty structs.

services/autorust/codegen/src/codegen_models.rs
```diff
@@ -859,6 +906,14 @@ fn create_struct(
     needs_boxing.insert(struct_name.to_camel_case_ident()?.to_string());
 
     for schema in schema.all_of() {
+        // skip empty base types
+        let mut properties_len = schema.properties().len();
```
cataggar, line 910: This should be field length, which is allOf + properties.

services/mgmt/videoanalyzer/src/package_preview_2021_11/models.rs
```diff
@@ -1321,26 +1264,24 @@ impl NetworkAccessControl {
...
 pub enum NodeBaseUnion {
     #[serde(rename = "#Microsoft.VideoAnalyzer.ProcessorNodeBase")]
-    MicrosoftVideoAnalyzerProcessorNodeBase(ProcessorNodeBase),
+    MicrosoftVideoAnalyzerProcessorNodeBase(ProcessorNodeBaseUnion),
     #[serde(rename = "#Microsoft.VideoAnalyzer.SinkNodeBase")]
-    MicrosoftVideoAnalyzerSinkNodeBase(SinkNodeBase),
+    MicrosoftVideoAnalyzerSinkNodeBase(SinkNodeBaseUnion),
     #[serde(rename = "#Microsoft.VideoAnalyzer.SourceNodeBase")]
-    MicrosoftVideoAnalyzerSourceNodeBase(SourceNodeBase),
+    MicrosoftVideoAnalyzerSourceNodeBase(SourceNodeBaseUnion),
```
cataggar, line 1284: This one is a union of unions.

services/svc/purview/src/package_2023_02_15_preview/models.rs
```diff
@@ -462,23 +457,7 @@ impl PurviewShareErrorInfo {
         }
     }
 }
-#[doc = "A received share data transfer object."]
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-pub struct ReceivedShare {
-    #[serde(flatten)]
-    pub proxy_resource: ProxyResource,
```
cataggar, line 469: !ReceiveShare should still exist.

services/mgmt/datashare/src/package_2021_08_01/models.rs
```diff
@@ -1744,88 +1744,7 @@ pub mod consumer_source_data_set_properties {
         }
     }
 }
-#[doc = "A DataSet data transfer object."]
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-pub struct DataSet {
-    #[serde(flatten)]
-    pub proxy_dto: ProxyDto,
```
cataggar, line 1751: This should still exist.

services/mgmt/alertsmanagement/src/package_preview_2021_08/models.rs
cataggar: I agree. I've removed the empty structs. It uncovered two other minor issues that I had to fix
- union types for vector aliases
- support for nested unions

demoray:
I'm digging into this and found at least one issue that I don't think is right.

In `services/svc/agrifood/src/package_2023_07_01_preview/models.rs`, `AuthCredentialsKind` is one of the enums defined in the spec.  It's generated, but never used.   As is `AuthCredentialsUnion`.

From looking at the spec, these should be used in `ApiKeyAuthCredentials`, but that doesn't include it.

Ref: https://github.com/Azure/azure-rest-api-specs/blob/a26794fe26cdc68f4d18bc782e0b4f2f29e24b40/specification/agrifood/data-plane/Microsoft.AgFoodPlatform/preview/2023-07-01-preview/agfood.json#L13738-L13758

cataggar:
Thanks for digging in. The generated code for that example looks fine to me. Here is the Union generated from `#/definitions/AuthCredentials"`.

``` rust
#[doc = "CredentialTypeEnum."]
#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
#[serde(tag = "kind")]
pub enum AuthCredentialsUnion {
    ApiKeyAuthCredentials(ApiKeyAuthCredentials),
    OAuthClientCredentials(OAuthClientCredentials),
}


#[doc = "Api Key Auth Credentials class for API Key based Auth."]
#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
pub struct ApiKeyAuthCredentials {
    #[doc = "Properties of the key vault."]
    #[serde(rename = "apiKey", default, skip_serializing_if = "Option::is_none")]
    pub api_key: Option<KeyVaultProperties>,
}

#[doc = "OAuthClientCredentials for clientId clientSecret auth."]
#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
pub struct OAuthClientCredentials {
    #[doc = "ClientId associated with the provider."]
    #[serde(rename = "clientId", default, skip_serializing_if = "Option::is_none")]
    pub client_id: Option<String>,
    #[doc = "Properties of the key vault."]
    #[serde(rename = "clientSecret", default, skip_serializing_if = "Option::is_none")]
    pub client_secret: Option<KeyVaultProperties>,
}
```

It is strange that that definition is not referenced anywhere else in the spec. Most specs then define the string enumeration inline, but the spec as defined it externally

``` json
 "AuthCredentials": {
      "description": "AuthCredentials abstract base class for Auth Purpose.",
      "required": [
        "kind"
      ],
      "type": "object",
      "properties": {
        "kind": {
          "$ref": "#/definitions/AuthCredentialsKind"
        }
      },
      "discriminator": "kind"
    },
    "AuthCredentialsKind": {
      "description": "CredentialTypeEnum.",
      "enum": [
        "OAuthClientCredentials",
        "ApiKeyAuthCredentials"
      ],
      "type": "string",
      "x-ms-enum": {
        "name": "AuthCredentialsKind",
        "modelAsString": true
      }
    },
```

Which is why there is a enum generated:

``` rust
#[doc = "CredentialTypeEnum."]
#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
#[serde(remote = "AuthCredentialsKind")]
pub enum AuthCredentialsKind {
    OAuthClientCredentials,
    ApiKeyAuthCredentials,
    #[serde(skip_deserializing)]
    UnknownValue(String),
}
```

It is not used anywhere else in the spec, but it is not wrong to also generate it.

demoray (approved):

PR #1436, opened 2023-10-06
Fix #[serde(flatten)] annotation and add ResourceType value

CHarrod283:
Bug 1: When `storage_blobs::blob::BlobProperties` deserializes with an unknown field it errors.
    
Reason: Deserialization is iffy in regard to XMl. This is due to ambiguous typing. Deserialization will interpret the inner field as a map as opposed to a string, thus resulting in a failed conversion with the #[serde(flatten)] annotation and as such, an error.
    
Fix: Change the map to use serde_json::Value
    
Bug 2: List blobs response occasionally can contain a `ResourceType` field. This isn't deserialized to a public field.

Fix: Add the field


CHarrod283:

cataggar (approved):

PR #1437, opened 2023-10-09
add union test

cataggar:
~This allows a `tests.rs` file for a generated service.~

It now just adds a union test and adds `anyhow` as a dev-dependency for services.
This PR does not add the tests to CI. That would be next.

### DIFF
--- a/services/mgmt/vmware/src/tests.rs
+++ b/services/mgmt/vmware/src/tests.rs
@@ -1,0 +1,47 @@
+use super::*;
+
+#[cfg(feature = "package-2023-03-01")]
+#[test]
+fn test_list_dhcp_configurations_deserialization() -> anyhow::Result<()> {
+    use anyhow::{bail, ensure};
+    use package_2023_03_01::models::*;
+    // copied from specification\vmware\resource-manager\Microsoft.AVS\stable\2023-03-01\examples\WorkloadNetworks_ListDhcpConfigurations.json
+    let json = br#"
+    {
+        "value": [
+          {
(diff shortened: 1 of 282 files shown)


services/mgmt/vmware/src/tests.rs
```diff
@@ -0,0 +1,47 @@
...
+    // copied from specification\vmware\resource-manager\Microsoft.AVS\stable\2023-03-01\examples\WorkloadNetworks_ListDhcpConfigurations.json
+    let json = br#"
+    {
+        "value": [
+          {
+            "id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/group1/providers/Microsoft.AVS/privateClouds/cloud1/workloadNetworks/default/dhcpConfigurations/dhcpConfigurations1",
+            "name": "dhcp1",
+            "properties": {
+              "displayName": "dhcpConfigurations1",
+              "dhcpType": "SERVER",
```
cataggar, line 17: This is a union that now passes with #1435 fixed.

heaths (approved):
Seems fine, but just to double check my understanding of *why*: effectively, it's a file in an otherwise generated location, so you want to move it, regen, then move it back (as well as reference it in `lib.rs`),yeah?

services/mgmt/vmware/src/tests.rs
```diff
@@ -0,0 +1,47 @@
+use super::*;
+
+#[cfg(feature = "package-2023-03-01")]
```
demoray, line 3: Instead of src/tests.rs, why not have tests/foo.rs?  Based on my understanding of the rest of the code generation, this wouldn't need anything special to manage.

cataggar: That is a lot easlier.

PR #1442, opened 2023-10-23
regenerate services

demoray:
ref: https://github.com/Azure/azure-rest-api-specs/commit/7c906b15744aaf7b782dfb82ccefcbef05cbf20b


services/mgmt/servicefabricmanagedclusters/autorust.toml
demoray, line 1: The only manual configuration was to ignore a preview tag for defining an object twice.

heaths (commented):
I don't think this is a sustainable model. We don't build SDKs for all services. Each SDK that we do generate is done so when the service is updated. If we keep updating all sources like this, there will always be differences from generators thus clients won't be stable. Perhaps it's moot since we're not a stable version (and won't be until official, at the very least), but perhaps we need to rethink when/how we regenerate.

demoray:
> I don't think this is a sustainable model. We don't build SDKs for all services. Each SDK that we do generate is done so when the service is updated. If we keep updating all sources like this, there will always be differences from generators thus clients won't be stable. Perhaps it's moot since we're not a stable version (and won't be until official, at the very least), but perhaps we need to rethink when/how we regenerate.

I agree this isn't sustainable as is.  The Go SDK [used this model for some time](https://github.com/Azure/azure-sdk-for-go/pulls?q=is%3Apr+%22Release+refresh+all+package%22+sort%3Aupdated-desc) and has since moved to the "sdk-release-request" model, which aligns to the idea you commented on, generating based on service update.

IMO, this is one of the multitude of things we will need to adjust to align to the SDK release process as part of the "becoming official" process.

heaths (approved):

Issue #1447, opened 2023-10-25
The SequenceNumber for azure_messaging_servicebus::service_bus::BrokerProperties should be i64


Gaelik-git:
I got this message when trying to fetch message from a servicebus queue

```
ServiceBussError(Error { context: Full(Custom { kind: DataConversion, error: Error("invalid value: integer `32932572275146753`, expected i32", line: 1, column: 270) }, "unable to parse header 'HeaderName(\"brokerproperties\"): HeaderValue(\"{\\\"DeliveryCount\\\":3,\\\"EnqueuedTimeUtc\\\":\\\"Wed, 25 Oct 2023 20:57:39 GMT\\\",\\\"LockToken\\\":\\\"2d5cd34b-b7ff-4796-843a-551023940075\\\",\\\"LockedUntilUtc\\\":\\\"Wed, 25 Oct 2023 21:14:52 GMT\\\",\\\"MessageId\\\":\\\"65c5020980844d05abf3b7ace97818bc\\\",\\\"PartitionKey\\\":\\\"63\\\",\\\"SequenceNumber\\\":32932572275146753,\\\"State\\\":\\\"Active\\\",\\\"TimeToLive\\\":922337203685.47754}\")' into azure_messaging_servicebus::service_bus::BrokerProperties") })
```

When checking the documentation here : https://learn.microsoft.com/en-us/azure/service-bus-messaging/message-sequencing, it says
> The SequenceNumber value is a unique 64-bit integer

but it's defined as i32

https://github.com/Azure/azure-sdk-for-rust/blob/4849d7937edb8a95bbdfb85e1837c7606c83f57b/sdk/messaging_servicebus/src/service_bus/mod.rs#L278-L293

I am not sure when a `BrokerProperties` is provided to the `PeekLockResponse`, but is it possible to avoid this ?

Either way I think the code should be fixed

PR #1448, opened 2023-10-25
fix: sequence_number should be i64

Gaelik-git:
I tought I'd give it a try to check if would break in multiple place to fix this issue :
https://github.com/Azure/azure-sdk-for-rust/issues/1447

It's a breaking change for the api, but it's still better than having the runtime issue I met

Let me know, if you more information 

Thanks for the good work on all the crates

### DIFF
--- a/sdk/messaging_servicebus/src/service_bus/mod.rs
+++ b/sdk/messaging_servicebus/src/service_bus/mod.rs
@@ -289,7 +289,7 @@ pub struct BrokerProperties {
     pub message_id: String,
     pub sequence_number: i64,
     pub state: String,
-    pub time_to_live: i64,
+    pub time_to_live: f64,
 }
 
 impl BrokerProperties {
--- a/sdk/messaging_servicebus/src/service_bus/mod.rs
+++ b/sdk/messaging_servicebus/src/service_bus/mod.rs
@@ -287,7 +287,7 @@ pub struct BrokerProperties {
     #[serde(with = "time::serde::rfc2822")]
     pub locked_until_utc: OffsetDateTime,
     pub message_id: String,
-    pub sequence_number: i32,
+    pub sequence_number: i64,
     pub state: String,
     pub time_to_live: i64,
 }


heaths (approved):
This jives with the documentation: https://learn.microsoft.com/azure/service-bus-messaging/message-sequencing#sequence-number

Gaelik-git:
I think there is also an issue with the `TimeToLive`

> floating point `922337203685.4775`, expected i64

heaths:
I don't think that's correct. According to https://learn.microsoft.com/rest/api/servicebus/message-headers-and-properties,

> BrokerProperties {TimeToLive} is the number of seconds of the TimeSpan (double).

/cc @jsquire @JoshLove-msft

heaths (approved):

Gaelik-git:
I don't know about your release cycle, hopefully it can be merged and released soon.
In the mean time I can use my git branch as deps

Thanks

jsquire:
Careful with those docs, @heaths.   They seem to be focused on SBMP, which is a deprecated protocol.   You'll want to look at the [AMQP protocol guide for Service Bus](https://learn.microsoft.comazure/service-bus-messaging/service-bus-amqp-protocol-guide#message-annotations).

That said, I can confirm that the sequence number should be `i64`, as it is a message annotation specific to Azure and not the standard AMQP sequence number.   TimeToLive is a bit different.  We generally represent it as a time interval in the language representation.  The [AMQP format](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-header) is a "duration in milliseconds" which is `uint`, by [spec](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-milliseconds).   If Rust doesn't have an intrinsic time interval type, `i64` seems more correct than `f64`, given the spec.

heaths:
@Gaelik-git from where did you get the error,
> floating point 922337203685.4775, expected i64

That was due to a deserialization error from Service Bus? What version of the service?

Gaelik-git:
It was the same event as #1447 message once I modified my project

Gaelik-git:
```
unable to parse header 'HeaderName(\"brokerproperties\"): HeaderValue(\"{\\\"DeliveryCount\\\":3,\\\"EnqueuedTimeUtc\\\":\\\"Wed, 25 Oct 2023 20:57:39 GMT\\\",\\\"LockToken\\\":\\\"2d5cd34b-b7ff-4796-843a-551023940075\\\",\\\"LockedUntilUtc\\\":\\\"Wed, 25 Oct 2023 21:14:52 GMT\\\",\\\"MessageId\\\":\\\"65c5020980844d05abf3b7ace97818bc\\\",\\\"PartitionKey\\\":\\\"63\\\",\\\"SequenceNumber\\\":32932572275146753,\\\"State\\\":\\\"Active\\\",\\\"TimeToLive\\\":922337203685.47754}\")' into azure_messaging_servicebus::service_bus::BrokerProperties
```

I am not sure how to find the version I am using of the service. It's a servicebus queue that was just created

Gaelik-git:
>     * TimeToLive is a bit different.  We generally represent it as a time interval in the language representation.  The [AMQP format](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-header) is a "duration in milliseconds" which is `uint`, by [spec](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-milliseconds).   If Rust doesn't have an intrinsic time interval type, `i64` seems more correct than `f64`, given the spec.

On the php sdk the ttl is also documented as float
https://github.com/Azure/azure-sdk-for-php/blob/3528039537b6ba9dd1f033a3c47c299964ee0f63/src/ServiceBus/Models/BrokerProperties.php#L113

The servicebus was create using terraform, we are not setting the default_ttl_value so it's using the TimeSpan.MAX as value `P10675199DT2H48M5.4775807S`

https://github.com/hashicorp/terraform-provider-azurerm/blob/8cea9cf79df5f3c138d0d51719824daea85d7019/helpers/validate/time_test.go#L85-L87

PR #1459, opened 2023-11-09
fix partition key typo in cosmos readme

merlinio2000:
note: I'm unsure whether `a_number` was chosen to avoid confusion with reserved keywords in other languages but I decided to just make the comment consistent with the actual code.


demoray:
@merlinio2000 , the readme is generated from the rust code.   Can you make the same changes in `sdk/data_cosmos/src/lib.rs`?

merlinio2000:

merlinio2000:
@demoray Thanks for pointing that out! Changes are made as requested

Issue #1228, opened 2023-02-23
azure_identity Credential implementations do not cache/refresh tokens


johnbatty:
In `azure_identity` the `XXXCredential` implementations of get_token() always do the work to fetch a new token, even if the previous token has not expired. It does no caching or refreshing of previously obtained tokens. This will clearly impact performance if you are making many API calls.  The impact will vary according to the credential mechanism, but a particularly bad example is if `AzCliCredential` is used then an `az` process is spawned to get a credential on every REST API call!

There is an [`AutoRefreshingTokenCredential`](https://docs.rs/azure_identity/latest/azure_identity/struct.AutoRefreshingTokenCredential.html) wrapper that does provide caching and renewing on expiry.  However, the documentation/examples do not explain that this needs to be used in order to have token caching, or the implications of not using it.  This wrapper also has [a recently discovered issue](https://github.com/Azure/azure-sdk-for-rust/issues/1225) where a single credential cannot be used for multiple different resources.

Links to several `get_token()` implementations, demonstrating that they unconditionally fetch new tokens:
- [AzureCliCredential::get_token()](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/identity/src/token_credentials/azure_cli_credentials.rs#L125)
- [ClientSecretCredential::get_token()](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/identity/src/token_credentials/client_secret_credentials.rs#L103)
- [ImdsManagedIdentityCredential::get_token()](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs#L92)

At a minimum we should improve the `azure_identity` documentation to explain the behaviour.  However, I don't think that this is sufficient as it is too easy for developers to use a Credential provider in an inefficient way that will degrade performance in a way that is not obvious (it just makes your API calls slower).

The equivalent [Credential objects in the dotnet SDK do cache/refresh tokens](https://learn.microsoft.com/en-gb/dotnet/api/azure.identity.managedidentitycredential?view=azure-dotnet#methods), so I think that the ideal goal would be to rework the Rust SDK Credential implementations to implement token caching/refreshing to align with this behaviour.

demoray:
With at least one of the providers, IMDS, there are rate limits that will break due to rate limits unless the tokens are cached.

demoray:
Implemented in 1484.

PR #1464, opened 2023-11-16
Add support for authenticating using azureauth cli

demoray:
This PR adds support for calling out to the azureauth CLI.

https://github.com/AzureAD/microsoft-authentication-cli


sdk/identity/src/token_credentials/azureauth_cli_credentials.rs
```diff
@@ -0,0 +1,182 @@
...
+            std::io::ErrorKind::NotFound => {
+                Error::message(ErrorKind::Other, "azureauth CLI not installed")
+            }
+            error_kind => Error::with_message(ErrorKind::Other, || {
+                format!("Unknown error of kind: {error_kind:?}")
+            }),
+        })?;
+
+        if !output.status.success() {
+            let output = String::from_utf8_lossy(&output.stderr);
```
heaths, line 137: Does `Command` already deal with the Unicode->UTF8 conversion? Output on Windows is not UTF8 by default unless explicit set as the codepage - even for PowerShell (had to deal with this for another app where proper UTF8 encoding was critical).

heaths (dismissed):

sdk/identity/src/token_credentials/azureauth_cli_credentials.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+/// Authentication Mode
+///
+/// Note: While the azureauth CLI supports devicecode, users wishing to use
+/// devicecode should use `azure_identity::device_code_flow`
+#[derive(Debug, Clone, Copy)]
+pub enum AzureauthCliMode {
+    All,
+    IntegratedWindowsAuth,
```
kyle-rader-msft, line 53: The `IntegratedWindowsAuth` and `Broker` modes are only valid modes on Windows. Not sure if you want to expose that here with `cfg!`'s or not. We generally recommend leaving the mode alone as the default will be the best options for your platform :)

sdk/identity/src/token_credentials/azureauth_cli_credentials.rs
```diff
@@ -0,0 +1,182 @@
...
+            .unwrap_or(false)
+        {
+            "azureauth.exe"
+        } else {
+            "azureauth"
+        };
+
+        let mut cmd = Command::new(cmd_name);
+        cmd.args([
+            "--resource",
```
kyle-rader-msft, line 106: The `resource` flag is a bit of an unfortunate reminent. Really under the hood, this is just a shortcut for a `scope` of `{resource}/.default`. We hope to at some point in the future remove that flag. I recommend using the `--scope` flag which is repeatable.

sdk/identity/src/token_credentials/azureauth_cli_credentials.rs
demoray: I don't think it does.  I tested the output empirically to validate it worked as expected on error.  Additionally, this is how the az-cli integration works.

kyle-rader-msft: AzureAuth explicitly sets it's output encoding as UTF-8.

sdk/identity/src/token_credentials/mod.rs
```diff
@@ -17,6 +18,7 @@ mod workload_identity_credentials;
 
 pub use auto_refreshing_credentials::*;
 pub use azure_cli_credentials::*;
+pub use azureauth_cli_credentials::*;
```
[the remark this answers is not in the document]
kyle-rader-msft: Hi @heaths, https://aka.ms/azureauth is a purpose built generic azure credential provider, that my team builds with the blessing of the Identity group. Not all developer environments can reliably install Python. So use AzureAuth as a stand-alone cred provider in most of our boot-strapping scenarios. it's also widely used across several internal applications, hitting roughly 10K MAU so far.

kyle-rader-msft (changes requested):
Thank you so much @demoray for jumping on this! This will unlock easy usage across a lot of developer tools!

sdk/identity/src/token_credentials/mod.rs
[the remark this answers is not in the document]
heaths: What Identity group, exactly? Within Azure SDKs? /cc @schaabs @christothes. While unofficial, the theme is to remain congruent with the official Azure SDKs for if/when we can GA.

heaths:
After discussing with some members of the Azure SDK Identity team, they have no plans to support _azureauth_ at this time but are aware of it and have talked with the team; however, they were fine with the idea to make it an optional feature in rust - something our other languages don't enjoy. If/when we support _azureauth_ we can revisit but our understanding is that this is mostly (only) an internal-only tool at this point.

So can you instead put this behind an optional off-by-default feature?

sdk/identity/src/token_credentials/mod.rs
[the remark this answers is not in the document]
kyle-rader-msft: The MSAL team(s). This is why we build AzureAuth alongside the other MSAL libraries, as the CLI MSAL variant :)

kyle-rader-msft:
@heaths That is understandable, we very much appreciate the inclusion as an optional feature :) (ty, Rust).

AzureAuth is OSS and might be used outside the company, but we explicitly enable no telemetry inherently as we care about data privacy first and foremost.

demoray:
> After discussing with some members of the Azure SDK Identity team, they have no plans to support _azureauth_ at this time but are aware of it and have talked with the team; however, they were fine with the idea to make it an optional feature in rust - something our other languages don't enjoy. If/when we support _azureauth_ we can revisit but our understanding is that this is mostly (only) an internal-only tool at this point.
> 
> So can you instead put this behind an optional off-by-default feature?

Done.

sdk/identity/src/token_credentials/mod.rs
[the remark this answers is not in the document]
demoray: This is now gated behind the feature "azureauth-cli".

sdk/identity/src/token_credentials/azureauth_cli_credentials.rs
cataggar: @demoray, this looks like the last outstanding issue, now that it is behind a feature flag.

sdk/identity/src/token_credentials/azureauth_cli_credentials.rs
```diff
@@ -0,0 +1,185 @@
...
+/// Authentication Mode
+///
+/// Note: While the azureauth CLI supports devicecode, users wishing to use
+/// devicecode should use `azure_identity::device_code_flow`
+#[derive(Debug, Clone, Copy)]
+pub enum AzureauthCliMode {
+    All,
+    #[cfg(target_os = "windows")]
+    IntegratedWindowsAuth,
+    Broker,
```
kyle-rader-msft, line 55:
```suggestion
    #[cfg(target_os = "windows")]
    Broker,
```
Broker would also need it's own `cfg` atribute right?

sdk/identity/src/token_credentials/azureauth_cli_credentials.rs
```diff
@@ -0,0 +1,185 @@
...
+        self.modes.push(mode);
+        self
+    }
+
+    pub fn with_modes(mut self, modes: Vec<AzureauthCliMode>) -> Self {
+        self.modes = modes;
+        self
+    }
+
+    fn get_access_token(&self, resource: &str) -> azure_core::Result<CliTokenResponse> {
```
kyle-rader-msft, line 90: Could we take `scopes: &Vec<String>` or similar here instead of a single `resource`?

kyle-rader-msft (changes requested):
Thanks for the feature flag, 2 remaining small comments :)

kyle-rader-msft:
@demoray , One more flag we highly encourage passing when using AzureAuth is the `--prompt-hint` flag, which, if possible, when there is an auth prompt, will be added to the auth prompt title bar. This only works, at present for WAM dialogs on Windows, but it aides with telemetry (if using it) and in the future will hopefully have better display options on Mac/Linux.

demoray:
> @demoray , One more flag we highly encourage passing when using AzureAuth is the `--prompt-hint` flag, which, if possible, when there is an auth prompt, will be added to the auth prompt title bar. This only works, at present for WAM dialogs on Windows, but it aides with telemetry (if using it) and in the future will hopefully have better display options on Mac/Linux.
> 
> Perhaps to future proof, we can add a `with_args` method to take an arbitrary list of extra args one might want to add to their usage?

I've added `--prompt-hint`.

kyle-rader-msft:
@demoray While testing this internally, we quickly found (and confirmed by reading the source for `Credential`, that each header construction invokes the get token method. So, internally we've tried adding to the AzureAuthTokenCredential a `Mutex<Option<TokenCliResponse>>` to cache the token in memory, and only proceed with calling azureauth if the token we have is valid for <= 2 minutes (which could be another ctor arg).

Calling AzureAuth can be expensive (300-1000ms depending on what's going on under the hood). So caching this will be important for performance in developer facing scenarios, which is exactly where AzureAuth is designed to be used :).

johnbatty:
@kyle-rader-msft Regarding token caching, there's an existing wrapper for this - see [AutoRefreshingTokenCredential](https://docs.rs/azure_identity/latest/azure_identity/struct.AutoRefreshingTokenCredential.html).

See my [`azure_devops_rust_api` example code](https://github.com/microsoft/azure-devops-rust-api/blob/f7fc6e5baf6ceb2d0ed87cfeae9170b63f570d11/azure_devops_rust_api/examples/utils/mod.rs#L27) for typical usage.

This is clearly not a great situation, as it is too easy for people to use the existing credential providers with no caching.  Ideally we would integrate the token caching into the credential providers so that it is done by default (which the other Azure SDKs seem to do).

I raised an issue about this earlier this year, but have not yet had a chance to implement it:
https://github.com/Azure/azure-sdk-for-rust/issues/1228

One further snippet on token caching - we had an issue raised previously which observed that the token cache needs to be per-resource.  I fixed this in `AutoRefreshingTokenCredential`, but worth being aware of it:
https://github.com/Azure/azure-sdk-for-rust/issues/1225

sdk/identity/src/token_credentials/azureauth_cli_credentials.rs
kyle-rader-msft: @demoray ping.

kyle-rader-msft (changes requested):
I think we still need 1 `cfg` on the broker mode (I could also see just leaving out the concept of modes and instead taking an arbitrary list of extra args to pass, so that as the args evolve over time this wrapper need not chance).

The other question I have is on the "resource" passed and it actually being a space separated list of scopes.

kyle-rader-msft:
@johnbatty - Thank you for calling that out!!

demoray:
> I think we still need 1 `cfg` on the broker mode (I could also see just leaving out the concept of modes and instead taking an arbitrary list of extra args to pass, so that as the args evolve over time this wrapper need not chance).
> 
> The other question I have is on the "resource" passed and it actually being a space separated list of scopes.

kyle-rader-msft (approved):
👍 given impending follow up PR.

cataggar (approved):

PR #1470, opened 2023-11-21
Removed mutex from storage credentials

jorgecarleitao:
Mutexes should be avoided whenever possible, as they can result in slower code and deadlocks. It also requires more code to use it, as well as a higher cognitive load on the developer side to reason about potential deadlocks and how to run stuff in parallel.

This PR removes the `Mutex` in StorageCredentials's inner argument.

A corollary is that many function signatures no longer need to be `async` (since most were async for the async lock operation). This PR does not make such a change, as it has a larger blast radius.

A second corollary is that the crate no longer needs to depend on `futures`. This PR removes this dependency.

# Backward incompatible changes:

* the inner attribute of `StorageCredentials` was modified from `Arc<Mutex<StorageCredentialsInner>>` to `Arc<StorageCredentialsInner>`. Its usage no longer requires locking.
* `StorageCredentials::replace`'s signature changed from `replace(&self, other: Self)` to `replace(&mut self, other: Self)`

### DIFF
--- a/sdk/storage/src/authorization/mod.rs
+++ b/sdk/storage/src/authorization/mod.rs
@@ -6,12 +6,7 @@ use azure_core::{
     auth::TokenCredential,
     error::{ErrorKind, ResultExt},
 };
-use futures::lock::Mutex;
-use std::{
-    mem::replace,
-    ops::{Deref, DerefMut},
-    sync::Arc,
-};
+use std::sync::Arc;
 use url::Url;
 
 /// Credentials for accessing a storage account.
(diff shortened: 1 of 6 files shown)


jorgecarleitao:

sdk/storage/src/authorization/mod.rs
```diff
@@ -136,52 +131,35 @@ impl StorageCredentials {
     ///
     /// This method is useful for updating credentials that are used by multiple
     /// clients at once.
```
demoray, line 133: By making this take `&mut self`, you cannot update multiple clients at once.

Can you discuss how this would work for those cases?

jorgecarleitao: Thank you for the review. Could you exemplify what you mean with an example? I could not fund a test that would no longer work with this PR.

demoray: @jorgecarleitao , the underlying reason for mutex wasn't exposed appropriately.  

#1474 exposes the intended functionality.

demoray: @jorgecarleitao , with the change in #1477, the impact performance from using a Mutex should be reduced.

cataggar:
An example has been provided for why the types are separate. And performance improved by using a RwLock instead of Mutex.
- #1474
- #1477

PR #1471, opened 2023-11-21
Replaced `StorageCredentials` by `StorageCredentialsInner`

jorgecarleitao:
This PR:

* Removes `StorageCredentials` struct
* Renames `StorageCredentialsInner` to `StorageCredentials`

Thereby removing one level of indirection in the storage credentials.

# Background

The struct `StorageCredentials` was a wrapper to `pub Arc<Mutex<StorageCredentialsInner>>`. However, this is not really needed:

* #1470 removed the `Mutex`, thereby making `StorageCredentials(pub Arc<StorageCredentialsInner>)`
* The `Arc` in this context is not required, as `StorageCredentialsInner` is not a large enum (and if any of the members is large, we should optimize them instead, e.g. using `Arc<[String]>` instead of `Vec<String>`).

# This PR

As such, we can make life easier for our users by providing an easier API: remove the indirection provided by `StorageCredentials` by replacing it altogether with the current `StorageCredentialsInner`. This PR does this.

### DIFF
--- a/sdk/storage/Cargo.toml
+++ b/sdk/storage/Cargo.toml
@@ -16,7 +16,6 @@ edition = "2021"
 async-trait = "0.1"
 azure_core = { path = "../core", version = "0.17", features = ["xml"] }
 time = "0.3.10"
-futures = "0.3"
 log = "0.4"
 serde = "1.0"
 serde_derive = "1.0"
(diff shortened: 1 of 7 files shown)


cataggar:
An example has been provided for why the types are separate. And performance improved by using a RwLock instead of Mutex.
- #1474
- #1477

PR #1472, opened 2023-11-27
rename AccessToken to Secret and expand it's usage

demoray:
This does 3 things:
1. Renames `AccessToken` to `Secret`
2. Prevents `Debug` of the `AccessToken` from actually showing the secret
3. Starts expanding the use of `Secret` to other areas, such as client certificates

### DIFF
--- a/sdk/core/src/auth.rs
+++ b/sdk/core/src/auth.rs
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 use std::{borrow::Cow, fmt::Debug};
 use time::OffsetDateTime;
 
-#[derive(Clone, Deserialize, Serialize)]
+#[derive(Clone, Deserialize, Serialize, Eq)]
 pub struct Secret(Cow<'static, str>);
 
 impl Secret {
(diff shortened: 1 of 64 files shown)


sdk/core/src/auth.rs
```diff
@@ -20,26 +20,44 @@ impl AccessToken {
...
 
+impl From<String> for Secret {
+    fn from(access_token: String) -> Self {
+        Self::new(access_token)
+    }
+}
+
+impl Debug for Secret {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("Secret").field(&"<REDACTED>").finish()
```
cataggar, line 31: Nice

cataggar (commented):
I'm open to this. Is there precedence in any other Azure SDKs or libraries? .NET uses Azure.Core.AccessToken.
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/AccessToken.cs

demoray:
> I'm open to this. Is there precedence in any other Azure SDKs or libraries? .NET uses Azure.Core.AccessToken. https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/AccessToken.cs

I don't know about wrapping secrets in a fashion like we've done.

As it relates to your link, the AccessToken implementation used in DotNet is what we call TokenResponse, as it includes both the token and an expiration date.  It's my intention to rename TokenResponse to AccessToken once this PR lands, as that should impact different places.

cataggar (approved):

PR #1473, opened 2023-11-28
standardize hmac_sha256 implementation

demoray:
With #1467 , we now include an openssl based hmac 256 implementation.

This unifies the implementation across the crates that need it.  Future work would be to provide an implementation that makes use of schannel as well.


demoray:
Something I would like to do once #1472 is merged, a future PR should move to using `Secret` for the hmac keys.

cataggar (approved):
Wonderful

PR #1481, opened 2023-12-01
Support user delegated sas urls

demoray:


sdk/storage/src/shared_access_signature/account_sas.rs
```diff
@@ -190,7 +191,7 @@ impl AccountSharedAccessSignature {
                     self.version,
                 );
 
-                hmac_sha256(&string_to_sign, &self.key).unwrap()
```
cataggar, line 193: Good to see a Result used.

cataggar (approved):

PR #1490, opened 2023-12-05
Add support for cloud to device messages

jltorresm:
Adds support for cloud-to-device messages.

### DIFF
--- a/sdk/iot_hub/examples/c2d_message.rs
+++ b/sdk/iot_hub/examples/c2d_message.rs
@@ -1,0 +1,58 @@
+use azure_iot_hub::service::ServiceClient;
+use std::collections::HashMap;
+
+use std::error::Error;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error + Send + Sync>> {
+    let iot_hub_connection_string = std::env::var("IOTHUB_CONNECTION_STRING")
+        .expect("Set env variable IOTHUB_CONNECTION_STRING first!");
+
+    let device_id = std::env::args()
+        .nth(1)
+        .expect("Please pass the device id as the first parameter");
+
+    let message_body =
+        std::env::args()
+            .nth(2)
+            .and_then(|s| if s.trim().is_empty() { None } else { Some(s) });
+
+    // Properties should be passed as a comma separated list of key=value pairs.
+    // For example: key1=value1,key2=value2,key3=value3
+    let properties = std::env::args().nth(3).map(|s| {
+        s.split(',')
+            .map(|s| {
+                let mut split = s.split('=');
+                (
+                    split.next().expect("use the format key=value").to_string(),
+                    split.next().expect("use the format key=value").to_string(),
+                )
+            })
+            .collect::<HashMap<String, String>>()
+    });
+
+    let service_client = ServiceClient::new_connection_string(iot_hub_connection_string, 3600)?;
+
+    println!(
+        "Sending cloud-to-device message to {device_id} on: {}\
+        \n\tbody: {message_body:?}\n\tproperties: {properties:?}",
+        service_client.iot_hub_name
+    );
+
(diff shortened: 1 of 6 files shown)


demoray:
@heaths I would like your feedback on this from a future-project direction standpoint.

For other languages, separate libraries are maintained outside of the Azure SDK.
* https://github.com/Azure/azure-iot-sdk-csharp
* https://github.com/Azure/azure-iot-sdk-python
* https://github.com/Azure/azure-iot-sdk-java
* https://github.com/Azure/azure-iot-sdk-c

Thus far, we've had limited updates for the azure_iot_hub crate maintained in this repo.

How should we move forward with regards to this and future features for azure_iot_hub?

jltorresm:
Moving the IoTHub code to a separate repo is an interesting prospect. Would this mean more support for IoTHub related features? I would also be interested in keeping the HTTP based sdk but also offering an AMQP based alternative which would resemble more other language (and more official) sdks.

jltorresm:
Are there any updates on this topic? I would love to have support for cloud-to-device messages in the current crate. I don't oppose having it as a separate repo, but in any case, it would be nice to have this merged before that happens.

heaths:
Sorry for the delay. This dropped off my radar.

I believe the other languages, it was done that way because of legacy. The service team supports them themselves but still mostly follows the Azure SDK guidelines. For the Rust SDKs, we'll need to find out. I'll figure out appropriate contacts and add them to this PR to discuss.

sdk/iot_hub/src/service/mod.rs
```diff
@@ -816,6 +816,25 @@ impl ServiceClient {
         DeleteConfigurationBuilder::new(self.clone(), if_match.into(), configuration_id.into())
     }
 
+    /// Send a cloud-to-device message.
+    ///
+    /// ```
+    /// use azure_iot_hub::service::ServiceClient;
+    /// use std::collections::HashMap;
+    ///
+    /// # let connection_string = "HostName=cool-iot-hub.azure-devices.net;SharedAccessKeyName=iot_hubowner;SharedAccessKey=YSB2ZXJ5IHNlY3VyZSBrZXkgaXMgaW1wb3J0YW50Cg==";
```
heaths, line 825: This secret must be removed and you should rotate it immediately. Do not put secrets in the repo. Mark this code fence `no_run` and use environment variables in your example.

sdk/iot_hub/examples/c2d_message.rs
```diff
@@ -0,0 +1,58 @@
...
+        std::env::args()
+            .nth(2)
+            .and_then(|s| if s.trim().is_empty() { None } else { Some(s) });
+
+    // Properties should be passed as a comma separated list of key=value pairs.
+    // For example: key1=value1,key2=value2,key3=value3
+    let properties = std::env::args().nth(3).map(|s| {
+        s.split(',')
+            .map(|s| {
+                let mut split = s.split('=');
```
heaths, line 25: Nit: I'd use `splitn(2, '=')` here to get the rest as the value.

sdk/iot_hub/src/service/mod.rs
heaths: Use the same ones as your other example.

heaths (changes requested):
One nit, but the secrets must be scrubbed and you should rotate them if you still have that resource imediately.

sdk/iot_hub/src/service/mod.rs
jltorresm: It's not a real secret, if you base64 decode it you'll realize.

heaths: Our secret scanner may still flag it. It doesn't decode the values (most won't be ASCII anyway). Since the value won't work anyway, you could just leave it blank.

heaths:
`main` branch has been moved to `legacy`. Closing old PRs with no timely author response. Please reopen once you address feedback.

PR #1491, opened 2023-12-05
enable caching at the top level DefaultAzureCredential

demoray:
Right now, the underlying layers exposed by DefaultAzureCredential include caching.  However, in common use, if the user is using az-cli, it will require environment credential/IMDS to fail for every request before moving onto az-cli.

By caching at the DefaultAzureCredential layer and the underlying layer, most repeated calls will hit the first layer cache quickly.


cataggar (approved):
This is true. I would also be interested in failing faster for the other credentials. Credentials that do not fail fast, may be should not be enabled by default. See #1495.

Issue #1487, opened 2023-12-05
Add support for to other blob service API


Gerrystev:
Is there any possibilities to add support to other [blob service api](https://learn.microsoft.com/en-us/rest/api/storageservices/operations-on-the-account--blob-service-) such as [set blob properties](https://learn.microsoft.com/en-us/rest/api/storageservices/set-blob-service-properties?tabs=microsoft-entra-id)? because I need it to set CORS for the blob service.

heaths:
Thank you for your report. Since we are working on official, supported crates and have no plans to release further versions of the previously unsupported crates, we have decided to close this issue. Some older issues were migrated to the new crates if they seemed relevant. If you feel that this issue was relevant even to newer code that has been or will be rewritten, please reopen and comment as to why you consider it relevant.

PR #1492, opened 2023-12-05
Expose default tag as feature

demoray:
Right now, if you want to specify the http client implementation, you _also_ must specify the package tag which can change regularly during updates.

This moves to including a feature named `default_tag` which refers to the specific tag name, as well as renames `no-default-tag` to `without_tag_import` to do what was previously referred as `no-default-tag`.

This allows users to have a consistent Cargo.toml entry for the latest tag and a specified http client such as the following:

```
azure_svc_batch = {version="0.17", default-features=false, features=["enable_reqwest", "default_tag"]}
```

This PR is broken up into two commits.  The first is the change to the generator.  The second is the changes after regenerating the crates.


demoray:
Note, this is important as I am looking to address #1487 by making use of models from the generated crate, where we want to specify the HTTP client but not have to keep the default tag up-to-date.

cataggar (approved):

PR #1493, opened 2023-12-05
Move to azure identity scopes

demoray:
This models what is being done in the `azidentity` implementation from the azure-sdk-for-go.

This PR is made up of 3 commits:
1. Update the SDK crates
2. Update the generator to use the same model
3. Regenerate the crates to include the update.


sdk/core/src/auth.rs
```diff
@@ -4,6 +4,8 @@ use serde::{Deserialize, Serialize};
 use std::{borrow::Cow, fmt::Debug};
 use time::OffsetDateTime;
 
+pub static DEFALT_SCOPE_SUFFIX: &str = "/.default";
```
cataggar, line 7: DEFAULT

cataggar (approved):

PR #1504, opened 2023-12-07
fix handling application/xml in generated crates

demoray:
This includes two changes:
1. Align `to_xml` to `to_json` by returning Bytes instead of String
2. Updates generator to use `to_xml` when the content-type is set to `application/xml`


cataggar (approved):
❤️

Issue #1516, opened 2023-12-17
Server `retry-after` throttling headers are ignored


johnbatty:
Throttling of client requests is typically handled by a server returning a status of 429 (`TooManyRequests`), with a `retry-after` header indicating how long the client should delay before retrying the request.  The current `azure_core` request pipeline has various retry policies to handle retries for this status code, but none of them inspect or use the `Retry-After` header value.

It is worth noting that there are some services that provide alternative retry-after headers that specify the retry time with millisecond precision.  Any fix should also handle these headers if present:
- `retry-after-ms`
- `x-ms-retry-after-ms`
  - [the Cosmos API uses this](https://learn.microsoft.com/en-us/rest/api/cosmos-db/common-cosmosdb-rest-response-headers)

The [Javascript Azure SDK looks for the headers in this order](https://github.com/Azure/azure-sdk-for-js/blob/17de1a2b7f3ad61f34ff62876eced7d077c10d4b/sdk/core/core-rest-pipeline/src/retryStrategies/throttlingRetryStrategy.ts#L35):
- `retry-after-ms`
- `x-ms-retry-after-ms`
- `retry-after`

See [here for general Azure retry guidance](https://learn.microsoft.com/en-us/azure/architecture/best-practices/retry-service-specific#retry-usage-guidance).

PR #1517, opened 2023-12-17
Add throttling support using retry-after headers

johnbatty:
Fixes: https://github.com/Azure/azure-sdk-for-rust/issues/1516

Updated `azure_core` retry policy to implement throttling by looking for and using the values returned by a server via `retry-after`, `retry-after-ms` or `x-ms-retry-after-ms` headers when the server returns 429 (`TooManyRequests`) or 503 (`ServerUnavailable`).

Details:
- There was already some parsing of the `retry-after` headers in `lro.rs` (long-running operations).
  - I moved this into the `retry_policies` module, as I felt this was probably a more appropriate home for it.
  - The existing code had a bug where it was treating the "ms" variant values as seconds rather than milliseconds.  I've fixed this.
  - The existing code also looked for the headers in a different order to the Javascript SDK.  I've switched this around to match - now looking for the `<xxx>-ms` headers before the `retry-after` header.
- I modified the `RetryPolicy` trait `wait(...)` method to have an additional `retry_after` parameter, to enable passing in of retry-after duration extracted from the headers (if any):
    ```rust
    async fn wait(&self, _error: &Error, retry_count: u32, retry_after: Option<Duration>)
    ```
- I updated the common retry policy code to extract any `retry-after` header value when the server response is 429 or 503.  Then when calculating the sleep time for any retry policy in the `wait(...)` method it takes the maximum of the retry-after value (if any) and the calculated policy value.
- I added some unit tests for the new code.

### DIFF
--- a/sdk/core/src/lro.rs
+++ b/sdk/core/src/lro.rs
@@ -1,5 +1,6 @@
-use crate::headers::{Headers, RETRY_AFTER, RETRY_AFTER_MS, X_MS_RETRY_AFTER_MS};
+use crate::headers::Headers;
 use std::time::Duration;
+use time::OffsetDateTime;
 
 /// Default retry time for long running operations if no retry-after header is present
 ///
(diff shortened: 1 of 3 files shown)


johnbatty:
@reillysiemens This may be of interest, given your comments in https://github.com/Azure/azure-sdk-for-rust/pull/1510

demoray (approved):

demoray (approved):

Issue #1515, opened 2023-12-15
sleep method in azure-core has a race and causes deadlock


msabansal:
The [sleep](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/core/src/sleep.rs) method can be better written and it currently suffers from a race that causes a deadlock. 

### Minimal repro
  `
tracing::info!("Starting tasks");
    let mut join_set = JoinSet::default();
    for _i in 0..50 {
        join_set.spawn(async move {
            for _i in 0..10 {
                azure_core::sleep(Duration::from_millis(100)).await;
            }        
        });
    }    
    while let Some(_) = join_set.join_next().await {
    }
    tracing::info!("Tasks done");
`

PR #1523, opened 2023-12-21
Fixing the naive implementation of sleep

msabansal:
Address Issue highlighted in https://github.com/Azure/azure-sdk-for-rust/issues/1515 in the naive implementation.

Delegates to tokio's sleep if tokio feature is turned on

### DIFF
--- a/sdk/core/Cargo.toml
+++ b/sdk/core/Cargo.toml
@@ -49,7 +49,7 @@ rustc_version = "0.4"
 
 [dev-dependencies]
 env_logger = "0.10"
-tokio = { version = "1.0", features = ["default"] }
+tokio = { version = "1.0", features = ["default", "macros", "rt", "time"] }
 thiserror = "1.0"
 
 [features]
@@ -62,7 +62,8 @@ hmac_openssl = ["dep:openssl"]
 test_e2e = []
 azurite_workaround = []
 xml = ["quick-xml"]
-tokio-fs = ["tokio/fs", "tokio/io-util"]
+tokio-fs = ["tokio/fs", "tokio/sync", "tokio/io-util"]
+tokio-sleep = ["tokio"]
 
 [package.metadata.docs.rs]
 features = ["xml", "tokio-fs", "enable_reqwest", "enable_reqwest_gzip", "enable_reqwest_rustls", "hmac_rust", "hmac_openssl", "xml"]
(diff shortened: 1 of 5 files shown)


demoray:
Instead of creating the atomic during poll, why not create it when creating
the Sleep object?

msabansal:
> Instead of creating the atomic during poll, why not create it when creating the Sleep object?

Can do that and improve. Wanted to see if this approach works. I had the Issue opened to hope and have the discussion there but since it didnt move i thought i would just open  a PR

demoray:
> Can do that and improve. Wanted to see if this approach works. I had the
> Issue opened to hope and have the discussion there but since it didnt move
> i thought i would just open a PR.
>

Many (most) of us take much of December off, hence the delay.  I'm
responding via email as I'm away from work.

I would also like to split off the tokio part as it's own off-by-default
feature (and PR), and keep this pr to just fixing the bug you're
experiencing.

msabansal:
> Can do that and improve. Wanted to see if this approach works. I had the Issue opened to hope and have the discussion there but since it didnt move i thought i would just open a PR.
> Many (most) of us take much of December off, hence the delay. I'm responding via email as I'm away from work. I would also like to split off the tokio part as it's own off-by-default feature (and PR), and keep this pr to just fixing the bug you're experiencing.

Also out of office but getting to play around with things :). Thanks for replying and reviewing the PR. Will split it in that manner

msabansal:
Made the tokio sleep alternative opt-in disabled by default. The naive implementation does a let of work in the poll as it needs to grab the wake context to wake the blocked task.

sdk/core/Cargo.toml
```diff
@@ -62,7 +62,8 @@ hmac_openssl = ["dep:openssl"]
 test_e2e = []
 azurite_workaround = []
 xml = ["quick-xml"]
-tokio-fs = ["tokio/fs", "tokio/io-util"]
+tokio-fs = ["tokio/fs", "tokio/sync", "tokio/io-util"]
```
demoray, line 65: What changed to the tokio-fs feature that requires `tokio/sync`?

msabansal: It uses mutex which is behind sync feature flag. It just didn't compile. Should fix ci tests as well

demoray (approved):

Issue #1503, opened 2023-12-07
add SpecificAzureCredential


cataggar:
This is a proposed design to be able to specify the type of credential from an environment variable such as `AZURE_CREDENTIAL_TYPE`. If the environment variable was set it would pick the desired credential. The mapping from String to TokenCredential would be like:

- "virtualmachine" -> VirtualMachineManagedIdentityCredential
- "appservice" -> AppServiceManagedIdentityCredential 
- "azurecli" -> AzureCliCredential
- "clientsecret" -> ClientSecretCredential
- "clientcertificate" -> ClientCertificateCredential
- "workloadidentity" -> WorkloadIdentityCredential
- "environment" -> EnvironmentCredential
- "managedidentity" -> ManagedIdentityCredential #1536
- "servicefabric" -> ServiceFabricManagedIdentityCredential #1529
- "cloudshell" -> CloudShellManagedIdentityCredential #1531
- "azurearc" -> AzureArcManagedIdentityCredential #1530

The `SpecificAzureCredential` could then be added as the first option in `DefaultAzureCredential`.

The proposed implementation is #1532.

cataggar:
Here are the credentials used by the Azure SDK for .NET based on
- [DefaultAzureCredential.cs](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/Credentials/DefaultAzureCredential.cs)
- [EnvironmentCredential.cs](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/Credentials/EnvironmentCredential.cs)
- [ManagedIdentityClient.cs](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs)

``` mermaid
mindmap
  root((.NET DefaultAzureCredential))
    1 Environment
      1a Client Secret
      1b Client Certificate
      1c Username Password
    2 Workload Identity
    3 Managed Identity
      3a Service Fabric
      3b App Service
      3c Cloud Shell
      3d Azure Arc
      3e Virtual Machine
    4 Shared Token Cache
    5 Visual Studio
    6 Visual Studio Code
    7 Azure CLI
    8 Azure PowerShell
    9 Azure Developer CLI
    10 Interactive Browser
```

demoray:
As a total aside: TIL GitHub supports mindmaps!

cataggar:
[...]

Issue #1495, opened 2023-12-06
split up ImdsManagedIdentityCredential


cataggar:
In https://github.com/Azure/azure-sdk-for-rust/issues/423 was renamed from `ManagedIdentityCredential` to `ImdsManagedIdentityCredential`.

If we look in the .NET SDK, their managed identity is a combination of these:
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs#L80-L91
``` csharp
        private static ManagedIdentitySource SelectManagedIdentitySource(ManagedIdentityClientOptions options)
        {
            return
                ServiceFabricManagedIdentitySource.TryCreate(options) ??
                AppServiceV2019ManagedIdentitySource.TryCreate(options) ??
                AppServiceV2017ManagedIdentitySource.TryCreate(options) ??
                CloudShellManagedIdentitySource.TryCreate(options) ??
                AzureArcManagedIdentitySource.TryCreate(options) ??
                TokenExchangeManagedIdentitySource.TryCreate(options) ??
                new ImdsManagedIdentitySource(options);
        }
    }
```

I think our current `ImdsManagedIdentityCredential` is a combination:
- Service Fabric
- App Service 2019
- Azure Arc
- IMDS

All but IMDS has a `TryCreate` that checks environment variables and fails really fast. I'm not sure what the best way is to implement something similar to `TryCreate` in Rust for our credentials.

cataggar:
The Go SDK split up `managedIdentityClient` from `managedIdentityCredential`. The client supports these types:

https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/managed_identity_client.go#L49C1-L56C2

``` golang
const (
	msiTypeAppService msiType = iota
	msiTypeAzureArc
	msiTypeAzureML
	msiTypeCloudShell
	msiTypeIMDS
	msiTypeServiceFabric
)
```

cataggar:
The JS SDK supports these types:

https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/identity/identity/src/credentials/managedIdentityCredential/index.ts#L160-L168

    const MSIs = [
      arcMsi,
      fabricMsi,
      appServiceMsi2019,
      appServiceMsi2017,
      cloudShellMsi,
      tokenExchangeMsi(),
      imdsMsi,
    ];

PR #1532, opened 2024-01-01
add azure_identity::create_credential(), SpecificAzureCredential, AppServiceManagedIdentityCredential, VirtualMachineManagedIdentityCredential 

cataggar:
Happy new year 2024!

This adds `azure_identity::new_credential()`, `SpecificAzureCredential`, `AppServiceManagedIdentityCredential`, and `VirtualMachineManagedIdentityCredential`.
- Similar to `azure_core::new_http_client()`, `azure_identity::create_credential()` is a function that simplifies the most common use cases. 
- `SpecificAzureCredential` fixes https://github.com/Azure/azure-sdk-for-rust/issues/1503. It provides an easy and fast way to specify a specific credential type at runtime using an `AZURE_CREDENTIAL_KIND` environment variable.
- `AppServiceManagedIdentityCredential` and `VirtualMachineManagedIdentityCredential` are split off of and make use of `ImdsManagedIdentityCredential`. This fixes https://github.com/Azure/azure-sdk-for-rust/issues/1495.

Most all of the credentials have something similar to `create(options: impl into<TokenCredentialOptions>) -> Result<Self>` now. This is similar to how C# has `TryCreate` shown in https://github.com/Azure/azure-sdk-for-rust/issues/1495. They are able to check environment variables or files to see if they can be created, something milliseconds fast. The main exception is `VirtualMachineManagedIdentityCredential`, which does not have anything quick to check, so it is disabled by default. This is a breaking behavioral change that allows the default credentials to be faster by default. You can still opt in to using them by setting `AZURE_CREDENTIAL_TYPE` to `virtualmachine`. You can also include them using the `DefaultAzureCredentialBuilder`.

Manual functional tests of the `AppServiceManagedIdentityCredential` and `VirtualMachineManagedIdentityCredential` was completed.


sdk/identity/src/token_credentials/azure_cli_credentials.rs
```diff
@@ -119,6 +119,11 @@ impl Default for AzureCliCredential {
 }
 
 impl AzureCliCredential {
+    pub fn create() -> azure_core::Result<Self> {
+        // TODO check `az version` to see if it's installed
```
cataggar, line 123: I created https://github.com/Azure/azure-sdk-for-rust/issues/1533 to track this TODO.

sdk/core/Cargo.toml
```diff
@@ -59,6 +59,7 @@ enable_reqwest_gzip = ["reqwest/gzip"]
 enable_reqwest_rustls = ["reqwest/rustls-tls"]
 hmac_rust = ["dep:sha2", "dep:hmac"]
 hmac_openssl = ["dep:openssl"]
+test = []
```
demoray, line 62: Can this be named something more self-describing?

cataggar: Yes, open to suggestions. May be `unit_test`?

sdk/identity/src/token_credentials/default_credentials.rs
```diff
@@ -33,15 +40,45 @@ impl DefaultAzureCredentialBuilder {
...
+        self
+    }
+
+    /// Include using virtual machine managed identity credential
+    pub fn include_virtual_machine_managed_identity_credentials(&mut self) -> &mut Self {
+        self.include_virtual_machine_managed_identity_credential = true;
+        self
+    }
+
+    /// Inlucde using app service managed identity credential
```
cataggar, line 79: Spelling

sdk/core/Cargo.toml
cataggar: I renamed it `unit_test`. Were you suggesting something even more specific like `noop_client`?

sdk/identity/src/token_credentials/client_secret_credentials.rs
```diff
@@ -145,6 +104,44 @@ impl ClientSecretCredential {
...
     }
+
+    pub fn create(
+        options: impl Into<TokenCredentialOptions>,
+    ) -> azure_core::Result<ClientSecretCredential> {
+        let options = options.into();
+        let env = options.env();
+        let tenant_id =
+            env.var(AZURE_TENANT_ID_ENV_KEY)
+                .with_context(ErrorKind::Credential, || {
```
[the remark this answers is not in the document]
demoray: It looks like this exists already, with map_kind.

demoray (changes requested):
In general, this is a significant positive step forward.

sdk/identity/src/token_credentials/client_certificate_credentials.rs
```diff
@@ -93,27 +115,25 @@ impl ClientCertificateCredential {
...
     {
+        let options = options.into();
         ClientCertificateCredential {
             tenant_id,
             client_id,
             client_certificate: client_certificate.into(),
             client_certificate_pass: client_certificate_pass.into(),
-            http_client: new_http_client(),
-            options,
+            http_client: options.options().http_client().clone(),
```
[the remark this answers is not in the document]
cataggar: Yes, a little bit awkward. It still felt better than copying all the members.

sdk/identity/src/token_credentials/default_credentials.rs
```diff
@@ -51,38 +88,99 @@ impl DefaultAzureCredentialBuilder {
...
+        if self.include_specific_credential {
+            if let Ok(credential) = SpecificAzureCredential::create(self.options.clone()) {
+                return vec![DefaultAzureCredentialEnum::Specific(credential)];
+            }
         }
+
+        let mut sources = Vec::<DefaultAzureCredentialEnum>::with_capacity(included.len());
+        for source in included {
+            match source {
+                DefaultAzureCredentialType::Specific => {}
```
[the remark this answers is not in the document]
cataggar: It is not. It is an idea that I introduced in this SDK.

sdk/identity/src/token_credentials/client_certificate_credentials.rs
```diff
@@ -22,40 +21,62 @@ use url::{form_urlencoded, Url};
...
-    authority_host: Url,
+#[derive(Clone, Debug)]
+pub struct ClientCertificateCredentialOptions {
+    options: TokenCredentialOptions,
     send_certificate_chain: bool,
 }
 
-impl Default for CertificateCredentialOptions {
-    fn default() -> Self {
+// alias of CertificateCredentialOptions for backwards compatibility
```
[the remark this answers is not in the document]
cataggar: I know, but good time to begin practicing. I'm sure this will help somebody.

sdk/identity/src/token_credentials/default_credentials.rs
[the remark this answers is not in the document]
demoray: It's my understanding unless something makes us more rust-like, we should model what exists in other SDKs, of which this would implement a rather different experience.

CC: @heaths

sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs
```diff
@@ -23,104 +29,60 @@ const MSI_API_VERSION: &str = "2019-08-01";
...
 impl ImdsManagedIdentityCredential {
-    /// Creates a new `ImdsManagedIdentityCredential` using the given `HttpClient`.
-    pub fn new(http_client: Arc<dyn HttpClient>) -> Self {
+    pub fn new(
+        options: impl Into<TokenCredentialOptions>,
+        endpoint: Url,
+        api_version: &str,
+        secret_header: HeaderName,
+        secret_env: &str,
+        id: ImdsId,
```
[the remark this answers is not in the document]
cataggar: Yes, we could add a `ManagedIdentityCredential` that contains all of the managed identities, but it is essentially just a subset of `DefaultAzureCredential`. I'm not sure how useful it is. Same with `EnvironmentCredential` now. Still, probably good to check off the compatibility box.

I expect `ImdsManagedIdentityCredential` to primarily be for internal use. Feature like `AzureArcManagedIdentity` #1503 will build off of it. I left it public, primarily for others to be able to build off of it for features we have not implemented.

sdk/identity/src/token_credentials/default_credentials.rs
```diff
@@ -51,38 +88,99 @@ impl DefaultAzureCredentialBuilder {
...
         if self.include_azure_cli_credential {
-            sources.push(DefaultAzureCredentialEnum::AzureCli(
-                AzureCliCredential::new(),
-            ));
+            sources.push(DefaultAzureCredentialType::AzureCli);
+        }
+        sources
+    }
+
+    fn try_create(
```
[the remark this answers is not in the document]
cataggar: Renamed as `create_sources`.

sdk/identity/src/token_credentials/environment_credentials.rs
```diff
@@ -30,117 +28,76 @@ const AZURE_AUTHORITY_HOST: &str = "AZURE_AUTHORITY_HOST";
...
 }
 
 impl EnvironmentCredential {
-    /// Creates a new `EnvironmentCredential` with the given `TokenCredentialOptions`.
-    pub fn new(http_client: Arc<dyn HttpClient>, options: TokenCredentialOptions) -> Self {
-        Self {
-            http_client,
-            options,
-            cache: TokenCache::new(),
+    pub fn create(
```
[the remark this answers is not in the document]
cataggar: Where should this be documented?

sdk/core/Cargo.toml
cataggar: Renamed to `noop_client`.

sdk/identity/src/token_credentials/default_credentials.rs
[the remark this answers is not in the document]
cataggar: The experience is different only when you opt into setting `AZURE_CREDENTIAL_TYPE`. For example, deploying to App Service, I can set it to `appservice` and the corresponding managed identity will be used. For a VM deployment, `virtualmachine` may be set and the corresponding managed identity will be used. Locally for development, `azurecli` may be set and it is guaranteed to use it. There is not delay in checking other credential types. Performance and low level flexibility is more rust-like.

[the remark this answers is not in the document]
demoray: Part of the ideals for the SDKs is to align implementations with other SDKs except where it's making it more idiomatic for the language.  "Try" verses "Result" is an idiomatic difference.

This isn't a language-idiomatic difference, but a unique feature.  (hence my CCing heaths)

sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs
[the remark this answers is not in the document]
demoray: Adding `ManagedIdentityCredential` would be useful, as it aligns with the different SDK implementations.

sdk/identity/src/token_credentials/default_credentials.rs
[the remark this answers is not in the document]
cataggar: I started an [internal design discussion](https://teams.microsoft.com/l/message/19:f6d52ac6465c40ea80dc86b8be3825aa@thread.skype/1704231889101?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1704231889101&teamName=Azure%20SDK&channelName=General&createdTime=1704231889101).

sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs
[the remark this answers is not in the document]
cataggar: I agree and I will add it after this merges. I created #1536 to track.

sdk/identity/src/token_credentials/environment_credentials.rs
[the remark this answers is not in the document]
demoray: Perhaps here? https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/identity/CHANGELOG.md

sdk/identity/src/token_credentials/options.rs
```diff
@@ -0,0 +1,76 @@
...
+/// The authority host is taken from the `AZURE_AUTHORITY_HOST` environment variable if set.
+/// If not, the default authority host is `https://login.microsoftonline.com` for the Azure public cloud.
+impl Default for TokenCredentialOptions {
+    fn default() -> Self {
+        let env = Env::default();
+        let authority_host = env
+            .var(AZURE_AUTHORITY_HOST_ENV_KEY)
+            .map(|s| Url::parse(&s))
+            .unwrap_or_else(|_| Ok(AZURE_PUBLIC_CLOUD.to_owned()))
+            .unwrap_or_else(|_| AZURE_PUBLIC_CLOUD.to_owned());
```
demoray, line 27: I am concerned this will swallow errors where the user intends to set the AZURE_AUTHORITY_HOST but it doesn't parse correctly.

sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs
```diff
@@ -23,104 +29,60 @@ const MSI_API_VERSION: &str = "2019-08-01";
...
+    ) -> Self {
+        let options = options.into();
         Self {
-            http_client,
-            object_id: None,
-            client_id: None,
-            msi_res_id: None,
+            http_client: options.http_client(),
```
demoray, line 51: Should this consume the `Env` from the options and use it below?

sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs
```diff
@@ -23,104 +29,60 @@ const MSI_API_VERSION: &str = "2019-08-01";
...
-        )?;
+        let mut url = self.endpoint.clone();
+        url.query_pairs_mut().extend_pairs(query_items);
 
         let mut req = Request::new(url, Method::Get);
 
         req.insert_header("metadata", "true");
 
-        let msi_secret = std::env::var(MSI_SECRET_ENV_KEY);
+        let msi_secret = std::env::var(&self.secret_env);
```
demoray, line 83: This uses `std::env::var`.  Should this be updated to use env from the options?  (see comment from earlier in this file)

sdk/identity/src/token_credentials/workload_identity_credentials.rs
```diff
@@ -17,37 +23,87 @@ use time::OffsetDateTime;
...
+        let options = options.into();
+        let env = options.env();
+        let tenant_id =
+            env.var(AZURE_TENANT_ID_ENV_KEY)
+                .with_context(ErrorKind::Credential, || {
+                    format!(
+                        "working identity credential requires {} environment variable",
+                        AZURE_TENANT_ID_ENV_KEY
+                    )
+                })?;
```
demoray, line 67:
```suggestion
                .map_kind(ErrorKind::Credential)?;
```

sdk/identity/src/token_credentials/client_secret_credentials.rs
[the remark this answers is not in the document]
cataggar: Here is an example error message:
```
Error: Error { context: Full(Custom { kind: Credential, error: Error { context: Full(Custom { kind: Credential, error: Error { context: Message { kind: Io, message: "environment variable IDENTITY_ENDPOINT not set" } } }, "app service credential requires IDENTITY_ENDPOINT environment variable") } }, "unable to create AZURE_CREDENTIAL_TYPE of appservice") }
```
It is providing more error context. It is saying where the error came from.

sdk/identity/src/token_credentials/workload_identity_credentials.rs
```diff
@@ -17,37 +23,87 @@ use time::OffsetDateTime;
...
+        let client_id =
+            env.var(AZURE_CLIENT_ID_ENV_KEY)
+                .with_context(ErrorKind::Credential, || {
+                    format!(
+                        "working identity credential requires {} environment variable",
+                        AZURE_CLIENT_ID_ENV_KEY
+                    )
+                })?;
```
demoray, line 75:
```suggestion
                .map_kind(ErrorKind::Credential)?;
```

sdk/identity/src/token_credentials/workload_identity_credentials.rs
```diff
@@ -17,37 +23,87 @@ use time::OffsetDateTime;
...
+
+        if let Ok(token) = env
+            .var(AZURE_FEDERATED_TOKEN)
+            .map_kind(ErrorKind::Credential)
```
demoray, line 79: As the error content isn't used here, it doesn't need to be mapped first.

```suggestion
```

sdk/identity/src/token_credentials/workload_identity_credentials.rs
```diff
@@ -17,37 +23,87 @@ use time::OffsetDateTime;
...
+        {
+            return Ok(WorkloadIdentityCredential::new(
+                options, tenant_id, client_id, token,
+            ));
+        }
+
+        if let Ok(token_file) = env
+            .var(AZURE_FEDERATED_TOKEN_FILE)
+            .map_kind(ErrorKind::Credential)
```
demoray, line 88: Same as above
```suggestion
```

sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs
cataggar: I thought about making that change earlier, but decided it should be separate. The main benefit would be being able to unit test `get_token`.

sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs
cataggar: No, I think that would be better off in a separate PR that aimed at unit testing `get_token`. `Env` is used in the `Credential` `create` functions so that they could be unit tested.

sdk/identity/src/token_credentials/workload_identity_credentials.rs
cataggar: The error messages are indeed better with how it is. I provided an example [above](https://github.com/Azure/azure-sdk-for-rust/pull/1532#discussion_r1439922931).

sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs
demoray: If `Env` is only supposed to be used for unit testing get_token, can it be made visible only to the crate?

I'm concerned that it's only sometimes used, which will be a surprise if someone attempts to use it.

sdk/identity/src/token_credentials/options.rs
cataggar: Okay, will have another look at this.

sdk/identity/src/token_credentials/default_credentials.rs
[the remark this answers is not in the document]
cataggar: The sentiment during review was to not add `SpecificAzureCredential` to `DefaultAzureCredential`, so I will remove it here.

sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs
cataggar: Made is `pub(crate)`.

sdk/identity/src/token_credentials/options.rs
cataggar: I made it a `String` and `Url` parsing will fail in the `create`s.

sdk/core/src/http_client/noop.rs
```diff
@@ -3,6 +3,10 @@ use async_trait::async_trait;
 #[derive(Debug)]
 pub struct NoopClient;
 
+pub fn new_noop_client() -> std::sync::Arc<dyn crate::HttpClient> {
```
demoray, line 6: I checked out this PR, as I wanted to investigate how these changes might play out with some of my other projects.  I noticed for the unit test which this function is used, using `new_http_client` (the primary mechanism to expose HTTP clients in the SDK) does not result in a change in the tests.

Why not use that, rather than have a new public API we have to account for the compatibility the future?

sdk/identity/src/token_credentials/specific_azure_credential.rs
```diff
@@ -0,0 +1,316 @@
...
+    pub const APP_SERVICE: &str = "appservice";
+    pub const CLIENT_SECRET: &str = "clientsecret";
+    pub const WORKLOAD_IDENTITY: &str = "workloadidentity";
+    #[cfg(feature = "client_certificate")]
+    pub const CLIENT_CERTIFICATE: &str = "clientcertificate";
+}
+
+/// Creates a `DefaultAzureCredential` by default with default options.
+/// If `AZURE_CREDENTIAL_TYPE` environment variable is set, it creates a `SpecificAzureCredential` with default options.
+pub fn create_credential() -> azure_core::Result<Arc<dyn TokenCredential>> {
```
demoray, line 32: I am concerned with this new public API.

It's my understanding of the meeting we attended last week with the Azure Identity v-team, this doesn't align with their desires moving forward.  I'll try and wrap up what I took away from the meeting here:

1. The v-team "disliked" `DefaultAzureCredential` (DAC), but acknowledged DAC was going to stay around moving forward because it had the benefit to have a single, easy to cut and paste, authentication mechanism that "just worked" for examples and documentation.
2. DAC should behave similarly across all of the SDKs.
3. Adding a new undiscoverable public API to DAC should be avoided.  During the meeting, multiple people described using an environment variable as an undiscoverable public API.

As such, their recommendation was to _not_ add `SpecificAzureCredential` to DAC.

I acknowledge `create_credential` is not DAC by name, but it's DAC by intent. 
 It's _usually_ just DAC with a shorter name, and then put into all of the examples as it's easier to understand by name.

By adding `create_credential` as a public API, this effectively creates a rust-only DAC clone with the feature the v-team did not want included.

In general, the work in this PR includes multiple positive steps towards aligning the azure-identity in Rust with azure-identity in the other SDKs.  That said, I have concerns with `SpecificAzureCredential` and `create_credential` based on the feedback provided by the Azure Identity v-team.

demoray (changes requested):

sdk/core/src/http_client/noop.rs
cataggar: That is a good point. Generally, how can one expose test functionality to another crate is the same workspace? It does validate that no network calls are made in the create tests. I am okay with backing this out.

sdk/identity/src/token_credentials/default_credentials.rs
```diff
@@ -1,27 +1,32 @@
...
 
 impl Default for DefaultAzureCredentialBuilder {
     fn default() -> Self {
         Self {
+            options: TokenCredentialOptions::default(),
             include_environment_credential: true,
-            include_managed_identity_credential: true,
+            include_app_service_managed_identity_credential: true,
+            // Unable to quickly detect if running in Azure VM, so it is disabled by default.
+            include_virtual_machine_managed_identity_credential: false,
```
demoray, line 29: During the Azure Identity v-team, this came up. 

I believe the stated stance was thus:
1. The v-team acknowledged the potential frustration that IMDS authentication includes a delay due to the network connection
2. Authenticating generally happens up-front, only once, and then auth cached & refreshed as needed.  As such, the additional delay at initialization was minimal and acceptable.
3. Someone on the v-team was investigating if there was a faster way to tell if the app was running in a VM, which if found, all of the SDKs could built upon in the future.
4. The "Default Azure Credential" should therefore have this enabled by default.  This ensures continuity across languages.

sdk/core/src/http_client/noop.rs
demoray: Unfortunately, the only way to expose functionality from one crate to another is with a public API.  

That said, in `azure_core`, we use the module name `azure_core::__private` to re-export the `paste` macro to the other crates as it's used by the `operation!` implementation which we export as a public API.  

https://github.com/Azure/azure-sdk-for-rust/blob/9cd67babbfccb22f7c4e6ab989123b6145869ba9/sdk/core/src/lib.rs#L110-L114

In my opinion, we should be careful about adding additional private-by-name-only APIs.  As Windows backwards compatibility testing has repeatedly shown, even if something is named "dont_use_this_as_its_badly_broken", people will depend on it.  (obligatory XKCD ref: https://xkcd.com/1172/)

sdk/identity/src/token_credentials/specific_azure_credential.rs
cataggar: In the meeting, I proposed that `SpecificAzureCredential` functionality be added to `DefaultAzureCredential` across all Azure Identity SDKs. Their sentiment was to not add it to DAC. In commit 6c8edf71311f65cc31e592f5785fd17f099e9680, I removed `SpecificAzureCredential` from `DefaultAzureCredential` to align with that. It is a separate credential. Our DAC aligns with the other Azure SDKs.

This PR introduces 3 convenience factory method:
-  `create_default_credential` which creates an `Arc<DefaultAzureCredential>`
- `create_specific_credential` which creates an `Arc<SpecificAzureCredential>`
- `create_credential` which is a factory methods that will create one of the above two credentials

This aligns DAC with the other Azure SDKs and adds #1503.

sdk/identity/src/token_credentials/default_credentials.rs
cataggar: I think it is a bug to enable it by default in the other Azure SDKs based on the open issues. We do not need to add this bug.

sdk/core/src/http_client/noop.rs
cataggar: It has been backed out. Some API surfaces are more stable than others. I think naming the features and modules appropriately is good communication.

sdk/security_keyvault/examples/get_secret_versions.rs
```diff
@@ -1,21 +1,16 @@
...
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let keyvault_url =
         env::var("KEYVAULT_URL").expect("Missing KEYVAULT_URL environment variable.");
     let secret_name = env::var("SECRET_NAME").expect("Missing SECRET_NAME environment variable.");
 
-    let creds = Arc::new(
-        DefaultAzureCredentialBuilder::new()
-            .exclude_managed_identity_credential()
```
cataggar, line 14: This is an all-to-common pattern to avoid the delay cost the managed identity for a VM. By defaulting it to off, the cost is avoided. This aligns with the Rust performance theme.

Those deploying to virtual machines can do:
```
DefaultAzureCredentialBuilder::new()
    .include_virtual_machine_managed_identity_credential()
    .build()?;
```

Or they can set `AZURE_CREDENTIAL_TYPE` to `virtualmachine` and use `SpecificAzureCrednetial` via `azure_identity::create_specific_credential()?` or `azure_identity::create_credential()?`.

heaths: The DAC is less about speed and more about getting a user-friendly behavior by default. I appreciate the initial (as @demoray pointed out above) delay is less than optimal, but I think having different behavior with DAC is worse. If the delay happened every time, I'm right there with you. But it shouldn't.

sdk/identity/src/token_credentials/specific_azure_credential.rs
heaths: Why an `Arc<>` vs. a `Rc<>`, though? Protecting access to the whole thing seems unnecessary when it should support multiple readers, single writer. It only needs fully protecting when a token is refreshed.

sdk/identity/src/token_credentials/specific_azure_credential.rs
```diff
@@ -0,0 +1,340 @@
...
+
+/// Creates a new `SpecificAzureCredential` with the default options.
+pub fn create_specific_credential() -> azure_core::Result<Arc<dyn TokenCredential>> {
+    Ok(Arc::new(SpecificAzureCredential::create(
+        TokenCredentialOptions::default(),
+    )?))
+}
+
+#[derive(Debug)]
+pub enum SpecificAzureCredentialEnum {
```
heaths, line 59: Suggest `SpecificAzureCredentialKind`. Or is an "Enum" suffix idiomatic? I have to say I've not seen an established pattern in rust and, in the face of none, we prefer "Kind" as a differentiator.

sdk/identity/src/token_credentials/specific_azure_credential.rs
```diff
@@ -0,0 +1,340 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub struct SpecificAzureCredential {
+    source: SpecificAzureCredentialEnum,
+}
+
+impl SpecificAzureCredential {
+    pub fn create(options: TokenCredentialOptions) -> azure_core::Result<SpecificAzureCredential> {
```
heaths, line 127: Come to think of it, why `create` vs. `new`?

heaths (changes requested):

sdk/identity/src/token_credentials/specific_azure_credential.rs
cataggar: Arc<dyn TokenCredential> is what all the clients are created with. [search](https://github.com/search?q=repo%3AAzure%2Fazure-sdk-for-rust%20Arc%3Cdyn%20TokenCredential%3E&type=code)

sdk/identity/src/token_credentials/default_credentials.rs
```diff
@@ -183,10 +284,11 @@ impl DefaultAzureCredential {
...
 
-impl Default for DefaultAzureCredential {
-    fn default() -> Self {
-        DefaultAzureCredentialBuilder::new().build()
-    }
+/// Creates a new `DefaultAzureCredential` with the default options.
+pub fn create_default_credential() -> azure_core::Result<Arc<dyn TokenCredential>> {
+    DefaultAzureCredentialBuilder::default()
+        .build()
+        .map(|cred| Arc::new(cred) as Arc<dyn TokenCredential>)
```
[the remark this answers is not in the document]
cataggar: So you would prefer `azure_identity::DefaultAzureCredential::create()?` instead of `azure_identity::create_default_credential()?`? Works for me.

sdk/identity/src/token_credentials/environment_credentials.rs
```diff
@@ -1,22 +1,20 @@
...
-const AZURE_CLIENT_ID_ENV_KEY: &str = "AZURE_CLIENT_ID";
-const AZURE_CLIENT_SECRET_ENV_KEY: &str = "AZURE_CLIENT_SECRET";
-const AZURE_USERNAME_ENV_KEY: &str = "AZURE_USERNAME";
-const AZURE_PASSWORD_ENV_KEY: &str = "AZURE_PASSWORD";
-const AZURE_CLIENT_CERTIFICATE_PATH_ENV_KEY: &str = "AZURE_CLIENT_CERTIFICATE_PATH";
-const AZURE_FEDERATED_TOKEN_FILE: &str = "AZURE_FEDERATED_TOKEN_FILE";
-const AZURE_FEDERATED_TOKEN: &str = "AZURE_FEDERATED_TOKEN";
-const AZURE_AUTHORITY_HOST: &str = "AZURE_AUTHORITY_HOST";
+#[derive(Debug)]
+pub enum EnvironmentCredentialEnum {
```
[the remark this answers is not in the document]
cataggar: I think I followed `DefaultAzureCredentialEnum`. I look at changing both. Would you leave the environment variable as `AZURE_CREDENTIAL_TYPE` or would you suggest calling it `AZURE_CREDENTIAL_KIND` or something else?

sdk/identity/src/token_credentials/specific_azure_credential.rs
cataggar: Works for me.

sdk/identity/src/token_credentials/specific_azure_credential.rs
cataggar: For the credentials, `new` is being used when you are passing everything in. `create` is being used  you are trying to create the credential from environment variables or via a filesystem check or whatnot in TokenCredentialOptions. It is similar to `TryCreate` that .NET is using.

sdk/identity/README.md
```diff
@@ -1,38 +1,39 @@
...
-        .text()
-        .await?;
-
-    println!("{:?}", response);
-    Ok(())
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+   let subscription_id =
+       std::env::var("AZURE_SUBSCRIPTION_ID").expect("AZURE_SUBSCRIPTION_ID required");
+
+   let credential = azure_identity::create_credential()?;
```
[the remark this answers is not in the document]
cataggar: This PR introduces it. It is similar to `azure_core::new_http_client()`. It does not belong to either of the two types it can create.

sdk/security_keyvault/examples/get_secret_versions.rs
cataggar: As an example, a CLI app using DAC that is relying on Azure CLI authentication would hit it every time. Basically, every example we have.

sdk/identity/src/token_credentials/azure_cli_credentials.rs
```diff
@@ -155,6 +155,11 @@ impl Default for AzureCliCredential {
 }
 
 impl AzureCliCredential {
+    pub fn create() -> azure_core::Result<Self> {
```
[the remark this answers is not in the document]
cataggar: I added `DefaultAzureCredential::create(options)` that is like the others.

sdk/identity/src/token_credentials/default_credentials.rs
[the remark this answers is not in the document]
cataggar: I added azure_identity::DefaultAzureCredential::create(options)?, but it is takes a `TokenCredentialOptions` and does not return an `Arc`. `azure_identity::create_default_credential()?` is a bit different and useful for getting started.

sdk/identity/src/token_credentials/environment_credentials.rs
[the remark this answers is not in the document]
cataggar: All renamed to `Kind`. Most now `pub(crate)`.

sdk/identity/src/token_credentials/specific_azure_credential.rs
```diff
@@ -0,0 +1,340 @@
...
+    Error,
+};
+use std::sync::Arc;
+
+pub const AZURE_CREDENTIAL_KIND: &str = "AZURE_CREDENTIAL_KIND";
+
+pub mod azure_credential_kinds {
+    pub const ENVIRONMENT: &str = "environment";
+    #[cfg(not(target_arch = "wasm32"))]
+    pub const AZURE_CLI: &str = "azurecli";
```
kyle-rader-msft, line 22: I've not read this whole PR yet, but wanted to call out that the azureauth_cli_credential has been left out here, and since it's behind a feature flag and isn't building was broken in version 0.18, I'm in the process of opening an issue + PR, but want o make sure I take in all the context from this PR first :).

cataggar: It can be added, but I'm leaving that as a follow-up by us.

sdk/identity/src/token_credentials/default_credentials.rs
cataggar: I put it back.

sdk/identity/src/token_credentials/specific_azure_credential.rs
heaths: I get that. I guess this is more of a question overarching this discussion. I've opened #1571 to track and will not hold up this PR for that.

sdk/identity/src/token_credentials/environment_credentials.rs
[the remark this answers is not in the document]
heaths: `AZURE_CREDENTIAL_TYPE` is new, right? Pretty sure, but wanted to check.

If so, that's actually a good question. The main reason we avoid "type" in SDKs is because it's overloaded and can be confusing, at least in some languages. But "type" for an env var name seems more human-friendly, IMO. @JeffreyRichter, any thoughts here?

If nothing else is blocking, I would just stick with "TYPE" for the env var for now. It feels right to me given what I said above.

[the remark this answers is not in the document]
JeffreyRichter: I think kind for the enum & envvar is better than type but we can leave it for now. If we get funding, we'll do a review of the entire API surface area and I'm sure we'll clean up lots of things.

cataggar:
Testin on a Azure Virtual Machine is complete. I created a VM, `scp`d the same `get_secret` binary and the `VirtualMachineManagedIdentityCredential` worked with the system assigned identity. Some testing notes:

``` nushell
az vm create -g $vm_name -n $vm_name --image $image --admin-username $env.USER --ssh-key-values $'($env.HOME)/.ssh/id_rsa.pub'
```

I used the portal to add the identity, but it can be done via the CLI too:
https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/qs-configure-cli-windows-vm

sdk/identity/README.md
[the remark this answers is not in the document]
demoray: We should undo changing all of the examples (and the readme) to use create_credential, and go back to using `DefaultAzureCredential`.

All other SDKs use `DefaultAzureCredential` in their documentation and examples.

ref: https://azure.github.io/azure-sdk/general_introduction.html#consistent
> * The Azure SDK for each target language feels like a single product developed by a single team.
> * There should be feature parity across target languages. This is more important than feature parity with the service.

demoray (dismissed):

sdk/identity/README.md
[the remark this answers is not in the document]
cataggar: It results in a DefaultAzureCredential being created by default. Let's discuss in the monthly call, today, if you can.

cataggar:
As we just discussed, I created #1581 as a placeholder for updating the README. This PR is ready to merge.

heaths (approved):

heaths:
Opened a couple tracking issues for subsequent discussion / work, but nothing to hold up this PR.

sdk/identity/src/token_credentials/default_credentials.rs
```diff
@@ -148,15 +248,19 @@ impl TokenCredential for DefaultAzureCredentialEnum {
...
+        DefaultAzureCredentialBuilder::default()
+            .with_options(options)
+            .build()
+    }
+
     /// Creates a `DefaultAzureCredential` with specified sources.
-    ///
-    /// These sources will be tried in the order provided in the `TokenCredential` authentication flow.
-    pub fn with_sources(sources: Vec<DefaultAzureCredentialEnum>) -> Self {
+    fn with_sources(sources: Vec<DefaultAzureCredentialKind>) -> Self {
```
cataggar, line 263: I removed `pub` here. It could be made public, but I'd like to know the use case.

sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs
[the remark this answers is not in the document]
cataggar: 👆  Relevant for #1659 @johnbatty.

Issue #1533, opened 2024-01-01
implement `az version` check in `AzureCliCredential::create`


cataggar:
This is a follow up to #1532 to check `az version` in the `AzureCliCredential::create` function.

The version check could also be used in combination with a feature to determine if the `expires_on` can be used instead of `expireOn`. It was fixed in Azure CLI 2.54.0 https://github.com/Azure/azure-cli/pull/27476#issuecomment-1835432354.

cataggar:
```
~/ms/azure-sdk-for-rust> az version --output json
{
  "azure-cli": "2.54.0",
  "azure-cli-core": "2.54.0",
  "azure-cli-telemetry": "1.1.0",
  "extensions": {
    "vmware": "5.0.2"
  }
}
~/ms/azure-sdk-for-rust> az account get-access-token --output json
{
  "accessToken": "MuchLonger_NotTheRealOne_Sv8Orn0Wq0OaXuQEg",
  "expiresOn": "2024-01-01 19:23:16.000000",
  "expires_on": 1704158596,
  "subscription": "33b83be5-faf7-42ea-a712-320a5f9dd111",
  "tenant": "065e9f5e-870d-4ed1-af2b-1b58092353f3",
  "tokenType": "Bearer"
}
```

demoray:
Instead of parsing the az-cli version, can we just check if expires_on exists and use that, otherwise use expiresOn?

Based on the az-cli PR linked above, this looks potentially feasible.

chlowell:
This was effectively closed by #2443--AzureCliCredential will simply require CLI version 2.54.0+

PR #1534, opened 2024-01-02
use expire_on from Azure CLI 2.54.0 if it exists

cataggar:
Related to #1533. Now that https://github.com/Azure/azure-cli/issues/19700 was released with Azure CLI 2.54.0 in November, the `expires_on` timestamp can be used instead of `expiresOn` that is a local date. It will be used if present. It will continue to fallback to `expiresOn` if the `old_azure_cli` feature is enabled. I added it to default features. May be in 2025, we can remove it from default features.

### DIFF
--- a/sdk/identity/src/token_credentials/azure_cli_credentials.rs
+++ b/sdk/identity/src/token_credentials/azure_cli_credentials.rs
@@ -108,7 +108,7 @@ struct CliTokenResponse {
     pub local_expires_on: OffsetDateTime,
     #[serde(rename = "expires_on")]
     /// The token's expiry time in seconds since the epoch, a unix timestamp.
-    /// Available in Azure CLI 2.54.0 or newer
+    /// Available in Azure CLI 2.54.0 or newer.
     pub expires_on: Option<i64>,
     pub subscription: String,
     pub tenant: String,
@@ -333,7 +333,7 @@ mod tests {
             token_response.tenant,
             "065e9f5e-870d-4ed1-af2b-1b58092353f3"
         );
-        assert_eq!(token_response.expires_on_timestamp, Some(1704158596));
+        assert_eq!(token_response.expires_on, Some(1704158596));
         assert_eq!(token_response.expires_on()?.unix_timestamp(), 1704158596);
         Ok(())
     }
--- a/sdk/identity/src/token_credentials/azure_cli_credentials.rs
+++ b/sdk/identity/src/token_credentials/azure_cli_credentials.rs
@@ -1,13 +1,14 @@
 use crate::token_credentials::cache::TokenCache;
 use azure_core::{
     auth::{AccessToken, Secret, TokenCredential},
-    error::{Error, ErrorKind},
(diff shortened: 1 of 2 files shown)


sdk/identity/src/token_credentials/azure_cli_credentials.rs
```diff
@@ -94,18 +95,47 @@ mod az_cli_date_format {
     }
 }
 
+/// The respnose from `az account get-access-token --output json`.
```
cataggar, line 98: spelling

sdk/identity/src/token_credentials/azure_cli_credentials.rs
```diff
@@ -94,18 +95,47 @@ mod az_cli_date_format {
...
 #[derive(Debug, Clone, Deserialize)]
-#[serde(rename_all = "camelCase")]
 struct CliTokenResponse {
+    #[serde(rename = "accessToken")]
     pub access_token: Secret,
-    #[serde(with = "az_cli_date_format")]
+    #[cfg(feature = "old_azure_cli")]
```
demoray, line 103: Instead of a feature, can we just check if it exists in the response and use it?

sdk/identity/src/token_credentials/azure_cli_credentials.rs
```diff
@@ -94,18 +95,47 @@ mod az_cli_date_format {
...
+    #[serde(rename = "expiresOn", with = "az_cli_date_format")]
     pub expires_on: OffsetDateTime,
+    #[serde(rename = "expires_on")]
+    pub expires_on_timestamp: Option<i64>,
```
demoray, line 107: Given `expires_on` is what we want to use moving forward, can this use the name `expires_on` and rename the value that comes from `expiresOn` to something else?

demoray (changes requested):
(see above comments)

sdk/identity/src/token_credentials/azure_cli_credentials.rs
cataggar: Yes, if it is in the response, it uses it.

The feature `old_azure_cli` affects when `expire_on` is not in the response. Additional dependencies are needed for `assume_local`. If you remove the `old_azure_cli` feature, it removes those dependencies, but only works with AZ CLI 2.54.0 and above.

sdk/identity/src/token_credentials/azure_cli_credentials.rs
cataggar: May be `local_expires_on` and `expires_on`?

sdk/identity/src/token_credentials/azure_cli_credentials.rs
cataggar: This PR has two changes.
1. use `expires_on` if it exists
2. the ability to opt out of the additional local timezone dependencies by not including `old_azure_cli`

demoray (approved):

PR #1541, opened 2024-01-03
fix token scopes for azure_storage and azure_iot_hub

demoray:


jorendorff:
Thanks for fixing this! IIUC this issue causes storage not to work at all with default credentials in 0.18. Is there a workaround? Or, any chance of a quick `azure_storage 0.18.1` with this fix in it?

demoray:
There isn't an easy work-around.  I'm working through kicking off a new release and hope to get that out the door soon.

Issue #1549, opened 2024-01-05
Hyper client hang in container


tierriminator:
I've been hit with random client hangs recently when deploying an application, which uses azure blob storage, to a kubernetes cluster. I've tracked it down to hyperium/hyper#2312, which affects reqwest and therefore also this repo.
The workaround is setting `pool_max_idle_per_host(0)` for the client.

PR #1550, opened 2024-01-05
work around hang issue in hyper

demoray:
As indicated in #1549, there is an issue with hyper (the underlying layer used by reqwest) that hangs in some cases on connection pools. 

This PR uses a commonly discussed workaround of setting `pool_max_idle_per_host` to 0.

Ref: https://github.com/hyperium/hyper/issues/2312

Mitigates #1549

### DIFF
--- a/sdk/core/src/http_client/reqwest.rs
+++ b/sdk/core/src/http_client/reqwest.rs
@@ -1,14 +1,24 @@
-use crate::error::{ErrorKind, ResultExt};
-use crate::{Body, HttpClient, PinnedStream};
-
+use crate::{
+    error::{ErrorKind, ResultExt},
+    Body, HttpClient, PinnedStream,
+};
 use async_trait::async_trait;
 use futures::TryStreamExt;
-use std::{collections::HashMap, str::FromStr};
+use std::{collections::HashMap, str::FromStr, sync::Arc};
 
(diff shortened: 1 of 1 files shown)


sdk/core/src/http_client/reqwest.rs
```diff
@@ -1,14 +1,24 @@
...
+pub fn new_reqwest_client() -> Arc<dyn HttpClient> {
     log::debug!("instantiating an http client using the reqwest backend");
-    std::sync::Arc::new(::reqwest::Client::new())
+
+    // set `pool_max_idle_per_host` to `0` to avoid an issue in the underlying
+    // `hyper` library that causes the `reqwest` client to hang in some cases.
+    //
+    // See <https://github.com/hyperium/hyper/issues/2312> for more details.
+    let client = ::reqwest::ClientBuilder::new()
+        .pool_max_idle_per_host(0)
```
ctaggart, line 18: Are there any adverse side effects of setting this?

demoray: It effectively disables connection pools.  This has a performance impact but mitigates a client hang.

In addition to the original reporter, I've experienced this issue as well.  It's been an open issue in the hyper repo since October 2020.

cataggar (approved):

PR #1551, opened 2024-01-05
Prep 0.19.0 release

demoray:
This is broken up into multiple commits

1. Setting the version for the idiomatic crates
2. Setting the version for the generated crates


demoray:
Note, this will conflict with #1545.  Which ever gets merged first, the other will need to be updated to include the same changes.

cataggar (approved):

Issue #1535, opened 2024-01-02
Docs have failed to publish for some crates for several releases


krishanjmistry:
This isn't an exhaustive list, but a few examples where I've noticed failures.

The docs for the generated crate `azure_mgmt_network` have failed to publish with the message _recursive without indirection_. Links to docs.rs follow:
- https://docs.rs/crate/azure_mgmt_network/0.16.0
- https://docs.rs/crate/azure_mgmt_network/0.17.0
- https://docs.rs/crate/azure_mgmt_network/0.18.0

There's been some work in the past to counteract this - notably the `properties.boxed` section of the [`autorust.toml`](https://github.com/Azure/azure-sdk-for-rust/blob/main/services/mgmt/network/autorust.toml) file.  However, the source of failure here is introduction of new tags. The tags are taken from the README in the spec: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/network/resource-manager/readme.md

The tag which is currently failing is `profile_hybrid_2020_09_01` - there's a question here as to whether this tag needs to be generated at all.
However, it's also worth noting that the latest tag now is `package-2023-09` - which was released last week. When the Rust SDK crates are regenerated with the latest specs, this will be another faulty tag.

Another example I found whilst running `cargo clippy --all-features` locally was on the `azure_mgmt_hardwaresecuritymodules` crate. This time, docs have failed to publish since version 0.12 due to multiple definitions of models within the preview tags.
- https://docs.rs/crate/azure_mgmt_hardwaresecuritymodules/0.18.0

It's worth noting that something like this would have been spotted by `cargo clippy --all-features` or `cargo check  - something the GitHub actions which currently run don't do.

cataggar:
Good idea. We have a Geneva Action that you can run from the Actions tab in any fork for any branch named `check-all-services`. May be we could add one named `clippy-all-services`.

- https://github.com/Azure/azure-sdk-for-rust/actions/workflows/check-all-services.yml
- https://github.com/Azure/azure-sdk-for-rust/blob/main/.github/workflows/check-all-services.yml
- https://github.com/Azure/azure-sdk-for-rust/blob/main/services/autorust/codegen/templates/check-all-services.yml.jinja

It would be good to run after updating the services. After new versions are published, you can monitor the failures at https://docs.rs/releases/recent-failures .

demoray:
Do you mean github-action?

demoray:
[...]

PR #1552, opened 2024-01-05
Fix docs.rs build for iot_hub, messaging_eventgrid, message_servicebus

krishanjmistry:
See https://github.com/Azure/azure-sdk-for-rust/issues/1535#issuecomment-1879164384.

Not sure how this could be tested without spinning up a docs.rs local build container:
https://docs.rs/about/builds#testing-builds-locally


demoray:
Something I'm looking into is using dtolnay's `cargo-docs-rs`.  I hope to have a PR that enables that soon.

ref: https://crates.io/crates/cargo-docs-rs

demoray (approved):

PR #1560, opened 2024-01-08
use async-process for executing commands

demoray:
As is, our use of Command to execute az-cli or azureauth-cli executes ina blocking contest.  This moves to async-process.

### DIFF
--- a/sdk/identity/src/token_credentials/azure_cli_credentials.rs
+++ b/sdk/identity/src/token_credentials/azure_cli_credentials.rs
@@ -1,11 +1,12 @@
 use crate::token_credentials::cache::TokenCache;
+use async_process::Command;
 use azure_core::{
     auth::{AccessToken, Secret, TokenCredential},
     error::{Error, ErrorKind, ResultExt},
     from_json,
 };
 use serde::Deserialize;
-use std::{process::Command, str};
+use std::str;
 use time::OffsetDateTime;
 
 #[cfg(feature = "old_azure_cli")]
@@ -162,7 +163,7 @@ impl AzureCliCredential {
     }
 
     /// Get an access token for an optional resource
-    fn get_access_token(scopes: Option<&[&str]>) -> azure_core::Result<CliTokenResponse> {
+    async fn get_access_token(scopes: Option<&[&str]>) -> azure_core::Result<CliTokenResponse> {
         // on window az is a cmd and it should be called like this
         // see https://doc.rust-lang.org/nightly/std/process/struct.Command.html
         let program = if cfg!(target_os = "windows") {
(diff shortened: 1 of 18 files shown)


sdk/identity/src/token_credentials/azure_cli_credentials.rs
```diff
@@ -162,7 +163,7 @@ impl AzureCliCredential {
     }
 
     /// Get an access token for an optional resource
-    fn get_access_token(scopes: Option<&[&str]>) -> azure_core::Result<CliTokenResponse> {
+    async fn get_access_token(scopes: Option<&[&str]>) -> azure_core::Result<CliTokenResponse> {
```
heaths, line 166: Should this necessarily always require an async state machine to call? E.g., in .NET we have both sync and async calls for everything, but partly because it took forever to implement the TPL and merge then in as "async await" to .NET so people just grew accustomed to sync. Unfortunately, it's still frighteningly prevalent today. But it seems similar for rust currently.

heaths (changes requested):
Overall it looks good, but I have one overarching question.

sdk/identity/src/token_credentials/azure_cli_credentials.rs
demoray: As is, nearly all of the IO in the SDK is async-based.  What isn't is either test code or likely a mistake.

There is an often-discussed issue regarding the sync/async differences in the APIs.  

Some of the more popular libraries in the ecosystem handle this by providing a copy of basically every wrapped call in `block_on`.  This is how the `reqwest` and `postgres` crate provide async/blocking versions.   There are some rough areas doing this, and some of the procedural macro-based solutions end up requiring a far bit of work.

This is something the Rust Async working group acknowledges is a challenge, and IIRC is still something they'd like to address in the future.  Pinging @yoshuawuyts, as they are a member of said working group and has written about this and related subjects a fair bit.

heaths: Sounds good enough for me. Just given the history of .NET / C#, this concerned me but, as you point out, practically all over prod code is async. I'm willing to take it as is, but if you want to wait for @yoshuawuyts to chime in, feel free. I'll sign off still.

heaths (approved):

sdk/identity/src/token_credentials/azure_cli_credentials.rs
demoray: As it relates to this PR, the parent that calls get_access_token is async already. 

That said, this PR changes 2 public APIs exposed by the crate (the `get_subscription` and `get_tenant` methods of the Az CLI wrapper) to be async.

I tagged Yosh as they are far more familiar with the future of async in Rust than I am, and therefore might have insight/recommendations for how we handle blocking/async moving forward.

For this PR however, I think it's fine as is.

PR #1566, opened 2024-01-09
address `clippy::match_same_arms`

demoray:


cataggar (commented):
Is this duplicate of #1561?

PR #1579, opened 2024-01-16
move to using `tracing` instead of `log`

demoray:
In order to move towards supporting OpenTelemetry for distributed tracing, we need to move the underlying logging implementation to use `tracing`.

This does _not_ start exposing spans or tracing state yet.

ref: https://azure.github.io/azure-sdk/general_implementation.html

### DIFF
--- a/sdk/data_cosmos/src/resources/permission/permission_token.rs
+++ b/sdk/data_cosmos/src/resources/permission/permission_token.rs
@@ -1,5 +1,6 @@
 use super::AuthorizationToken;
 use azure_core::{auth::Secret, base64};
+use tracing::trace;
 
 const PERMISSION_TYPE_PREFIX: &str = "type=";
 const VERSION_PREFIX: &str = "ver=";
--- a/sdk/data_tables/src/lib.rs
+++ b/sdk/data_tables/src/lib.rs
@@ -19,7 +19,7 @@ struct MyEntity {
 
 #[tokio::main]
 async fn main() -> azure_core::Result<()> {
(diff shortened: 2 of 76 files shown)


sdk/data_cosmos/src/resources/permission/permission_token.rs
```diff
@@ -1,5 +1,6 @@
 use super::AuthorizationToken;
 use azure_core::{auth::Secret, base64};
+use tracing::trace;
```
heaths, line 3: Is this even used that you could just add a `use` and not remove one to, say, `log::trace`?

heaths (approved):
I saw quite a few files where a `use tracing::trace` was added but no obvious indication of use. If a `use log::trace` were removed (and in some places it was), that makes sense; however, it's puzzling to see one without the other. Were `log::trace` instances left behind? I expanded a few files and didn't see any, but it was a small sample. I'd just ask that you double check if not sure. I suppose as long as the `log` crate was removed from all manifests it shouldn't be possible to use.

sdk/data_tables/src/lib.rs
```diff
@@ -63,8 +63,6 @@ async fn main() -> azure_core::Result<()> {
 
 */
 
-#[macro_use]
-extern crate log;
```
demoray, line 67: @heaths, this is what required the inclusion of `tracing::trace` and the like in a handful of places.

Rather than import _all_ of the macros from the tracing crate implicitly, this PR move to explicitly using the intended macros in the intended locations.

cataggar (approved):

PR #1611, opened 2024-02-20
fix: naive_redirect_server TCP Port should be a u16

NTmatter:
The Naive Redirect Server function accepts a u32 for what should be a 16-bit unsigned integer.

This change changes the function to accept a u16 to avoid out-of-bounds input.


heaths (approved):
Thank you. This does jive not only with `SocketAddrV4` but makes sense for ports.

JeffreyRichter (approved):

PR #1619, opened 2024-02-29
Remove default features of http-types

omid:


heaths:
Can you explain why?

heaths (changes requested):
Please provide a reason as to why this change is necessary or recommended.

omid:
@heaths I'm not familiar with the source code that much, so my change can break the lib, but at least the tests I run passes.

I use azure sdk in a project and saw useless (to me!) and old dependencies which are attached to the not-actively-maintained `http-types`. And this change is to avoid including them, if I understood correctly.

omid:
OK, now I investigated more :/

`http-types` has some 3-year-old changes on `master` which are not yet released! So my change here is meaningless and wrong!

Rust ecosystem should think about this old `http-types` package with ~10M downloads!

svix-jplatte:
Related: #1644

Issue #1284, opened 2023-05-20
Can't filter blob by multiple tags due to url encoding using `+` for spaces instead of `%20`


rafamerlin:
I'm using azure_storage_blobs = "0.11.0"
Using ServiceClient.filter_blobs_by_tags.

It works fine with 1 single tag but if you need to add the ` AND ` in the search expression you start getting some odd errors, the library returns authentication errors but if you hit the URL directly you get different ones.

Example, I've created a blob container, created a couple of files and add a few tags in each, then using azure storage explorer I got one a query-string access signature that has Read, Write, Tag and Filter enabled.

If I call this URL:

https://redacted.blob.core.windows.net/?comp=blobs&where=%22Tag1%22%3D%27Hello%27+AND+%22Tag2%22%3D%27Goodbye%27&sv=the_sv_I_generated

I get this response:
```
<Error>
<Code>InvalidQueryParameterValue</Code>
<Message>
Error parsing query at or near character position 15: unexpected '1' RequestId:4cb3fb70-101e-001f-3709-8bc088000000 Time:2023-05-20T10:53:55.1754118Z
</Message>
<QueryParameterName>where</QueryParameterName>
<QueryParameterValue>"Tag1"='Hello'+AND+"Tag2"='Goodbye'</QueryParameterValue>
<Reason>This query parameter value is invalid.</Reason>
</Error>
```

If I simply change the `+` with a `%20` in the URL I get the proper response.

The azure-sdk-for-rust is using form_urlencoded to generate the url key/value pairs and this library uses `+` instead of `%20`.

It seems to be done in here:
https://github.com/servo/rust-url/blob/master/form_urlencoded/src/lib.rs#L140-L147

I'm unsure where this needs to be fixed as the form_urlencoded seems to be following the URL standards properly.

johnbatty:
Thanks for the detailed report.

These `url` crate issues describe the same problem, and show there is quite a bit of confusion on this!
- https://github.com/servo/rust-url/issues/695
- https://github.com/servo/rust-url/issues/482
- https://github.com/servo/rust-url/issues/416

It's not yet clear to me what the correct code fix is.

heaths:
[...]

PR #1622, opened 2024-03-11
Fix find blob by tag

rafamerlin:
There were 2 issues in here, one is that the API version got updated recently and that caused the response to be different than the one expected so querying blobs by tag wasn't working. I believe the version that the current `main` code has is the response for when we query with `x-ms-version: 2019-12-12`, however, we're now sending our requests with `x-ms-version: 2022-11-02` which caused this to completely break.

Another item was when we add multiple tags to the where expression, that was failing because of the way the URL was being encoded and that doesn't seem to be accepted by Azure REST API.

I've been using a workaround branch that was still using the old API for quite a while without any issues with this URL fix, related to this issue I raised a while ago https://github.com/Azure/azure-sdk-for-rust/issues/1284 

The error seems to have changed on the new API version as without the `make_url_compatible_with_api` if we query anything with more than 1 tag in the filter we get this error: 
```
<Error>
    <Code>AuthenticationFailed</Code>
    <Message>Server failed to authenticate the request. Make sure the value of Authorization header
        is formed correctly including the
        signature.RequestId:602fdf47-a01e-0010-5d99-732f4c000000Time:2024-03-11T09:50:11.4905786Z</Message>
    <AuthenticationErrorDetail>
        The MAC signature found in the HTTP request
        &apos;{{redacted}}&apos; is not the same as any computed
        signature. Server used following string to sign: &apos;GETx-ms-date:Mon, 11 Mar 2024
        09:50:06
        GMTx-ms-version:2022-11-02/{{redacted}}/comp:blobswhere:@container=&apos;unit-tests-with-index&apos;+AND+tagis=&apos;a&apos;&apos;.</AuthenticationErrorDetail>
</Error>
```

One question I have regarding the extra structs I had to add to have the response matching the new format of the XML for the tag content, do we need that at all? I was wondering if we could just drop the tag content as it doesn't seem like it's being used anywhere. Happy to keep it though in case there's any intention of using it in the future or if we want it for information purposes.

### DIFF
--- a/sdk/storage_blobs/src/service/operations/find_blobs_by_tags.rs
+++ b/sdk/storage_blobs/src/service/operations/find_blobs_by_tags.rs
@@ -1,7 +1,7 @@
 use crate::prelude::BlobServiceClient;
 use azure_core::{prelude::*, Response as HttpResponse};
 use azure_storage::headers::CommonStorageResponseHeaders;
-use std::str::FromStr;
+use url::Url;
 
 operation! {
     #[stream]
@@ -26,7 +26,7 @@ impl FindBlobsByTagsBuilder {
                 }
                 url.query_pairs_mut().append_pair("where", &this.expression);
 
-                let url = make_url_compatible_with_api(url)?;
+                make_url_compatible_with_api(&mut url);
 
                 let mut request = BlobServiceClient::finalize_request(
                     url,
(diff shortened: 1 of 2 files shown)


heaths:
@seanmcc-msft is this a known breaking change to the format?

rafamerlin:
I took the chance to hook up the MaxResults into FindBlobsByTagsBuilder as it wasn't doing anything, removed the next_marker on the input and change it to Marker to match the `ListBlobsBuilder` and made the next_marker public on the response so we can actually use it.

heaths:
@seanmcc-msft is there anyone on your team that can review this? It's mostly service-related code.

seanmcc-msft:
@jaschrep-msft @vincenttran-msft @jalauzon-msft please take a look.

sdk/storage_blobs/src/service/operations/find_blobs_by_tags.rs
```diff
@@ -40,13 +45,20 @@ impl FindBlobsByTagsBuilder {
     }
 }
 
+/// ` AND ` in the spec can be converted to `+AND+`, however, azure rest api won't accept it unless we do a `%20AND%20`
+fn make_url_compatible_with_api(url: &mut Url) {
+    let compatible_query = url.query().map(|q| q.replace("+", "%20"));
+
+    url.set_query(compatible_query.as_deref());
+}
```
jaschrep-msft, line 53: General note, requiring `%20` instead of `+` as the proper encoding for a space character is service-wide. Any URL encoding done by the library for storage requests needs to do this. Probably beyond the scope of this PR, but good future reference.

heaths:
`main` branch has been moved to `legacy`. Retargeting some old PRs.

PR #1631, opened 2024-03-28
Support xml error respsonse

### LINKED ISSUE #1275: `HttpError::error_message()` is unexpectedly `None`
When using Azurite or the azure public cloud blob service, I am seeing `HttpError`s that have the `error_code()` filled in, but `error_message()` is None, despite the body of the response containing a Message.

I think the problem is that `http_error::get_error_message_from_body()` is assuming that the body is Json:
```rust
let json = serde_json::from_slice::<serde_json::Value>(body).ok()?;
```
However, in my experience with the Blob APIs, the body is actually XML (possibly with a BOM ([Byte Order Mark](https://en.wikipedia.org/wiki/Byte_order_mark))).  A [Microsoft employee explains](https://learn.microsoft.com/en-us/answers/questions/158427/does-blobstorage-api-returns-errors-in-json-format) that XML is returned even when the client requests Json.  Examples:

From Azurite (no BOM):
`StatusCode { status: 400, body: "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?>\n<Error>\n  <Code>InvalidQueryParameterValue</Code>\n  <Message>Value for one of the query parameters specified in the request URI is invalid.\nRequestId:dee0c303-ede6-4dab-95cd-eaaaabda595d\nTime:2023-05-11T17:12:46.001Z</Message>\n</Error>" }`
From public cloud (with BOM):
`HTTP error status (status: 404, body: "\u{feff}<?xml version=\"1.0\" encoding=\"utf-8\"?><Error><Code>BlobNotFound</Code><Message>The specified blob does not exist.\nRequestId:92ed15c1-901e-007b-2f90-82f3b3000000\nTime:2023-05-09T16:07:20.0241138Z</Message></Error>")`

In a previous version of the SDK, I was able to retrieve the body of the error and then parse it myself.  However, in 0.11.0, I don't see a way to get the body, except via `HttpError.to_string()`, which includes a bunch of other output and is not really intended for parsing.

FWIW, this is the code that I used to parse the body.  It would probably need to be changed to use quick-xml instead of xml-rs:
```rust
[derive(Debug, Deserialize)]
struct XmlRequestError {
    #[serde(rename = "Code")]
    code: String,
    #[serde(rename = "Message")]
    message: String,
}

                // The Byte-Order-Mark (or BOM), is a special marker added at the very beginning
                // of an Unicode file encoded in UTF-8, UTF-16 or UTF-32. It is used to indicate
                // whether the file uses the big-endian or little-endian byte order. Azure-Blob
                // returns xml with a BOM prefix. Since serde_xml_rs does not deal with it, it
                // needs to be trimmed first.
                let body = std::str::from_utf8(response.body()).unwrap();
                debug!("response body: {body}");
                if let Some(index) = body.find("<?xml") {
                    if let Ok(error) = serde_xml_rs::from_str::<XmlRequestError>(&body[index..]) {
                        return format!("{}: {}", error.code, error.message);
                    }
                }
```

I can probably open a PR to fix this, if there's agreement on the desired semantics.  Are there some responses that use Json, or do all of them use XML?  Should we try decoding as XML and if that fails then try Json?  Should we add a function to retrieve the raw body?

merlinio2000:
closes #1275 

As mentioned in the issue for those places that only need the error code it would've been easier to get it from the headers but I'm unsure whether there are any case where this information is only provided through the body.
So to keep the current behavior I decided to keep parsing it from the body (now based on `content-type` though)

If you think it would be alright to rely solely on the headers than I'd be happy to change that, it would remove a bit of code and prevent having to parse the body in some cases.

EDIT: Thanks to the all_tests script I discovered the issue of `quick-xml` being behind the `xml` feature flag.
I see two options but don't know which one is better:
- dont feature gate`quick-xml` / remove the `xml` feature
- create a non-xml version of the body parsing and gate the current one behind the `xml` feature (needs documentation)

To quick fix for CI I will remove `quick-xml` from the `xml` feature for now. I'm thankful for any further guidance here

### DIFF
--- a/sdk/core/src/error/http_error.rs
+++ b/sdk/core/src/error/http_error.rs
@@ -11,7 +11,7 @@ use serde::Deserialize;
 pub struct HttpError {
     status: StatusCode,
     details: ErrorDetails,
-    headers: Headers,
+    headers: impl Into<std::collections::HashMap<String, String>>,
     body: Bytes,
 }
 
--- a/sdk/core/src/error/http_error.rs
+++ b/sdk/core/src/error/http_error.rs
@@ -1,7 +1,6 @@
 use crate::{
     content_type, from_json,
     headers::{self, Headers},
-    xml::read_xml,
     Response, StatusCode,
 };
 use bytes::Bytes;
@@ -145,7 +144,7 @@ impl ErrorBody {
 }
 
 /// Gets the error code and message from the body based on the specified content_type
-///
+/// Support for xml decoding is dependent on the 'xml' feature flag
 ///
 /// Assumes JSON if unspecified/inconclusive to maintain old behaviour
 /// [#1275](https://github.com/Azure/azure-sdk-for-rust/issues/1275)
@@ -154,13 +153,22 @@ pub(crate) fn get_error_code_message_from_body(
     body: &[u8],
     content_type: Option<&str>,
 ) -> (Option<String>, Option<String>) {
-    let err_body: Option<ErrorBody> =
-        if content_type.is_some_and(|ctype| ctype == content_type::APPLICATION_XML.as_str()) {
-            read_xml(body).ok()
-        } else {
-            // keep old default of assuming JSON
-            from_json(body).ok()
-        };
+    let err_body: Option<ErrorBody> = if content_type
+        .is_some_and(|ctype| ctype == content_type::APPLICATION_XML.as_str())
+    {
+        #[cfg(feature = "xml")]
+        {
+            crate::xml::read_xml(body).ok()
+        }
+        #[cfg(not(feature = "xml"))]
+        {
+            tracing::warn!("encountered XML response but the 'xml' feature flag was not specified");
+            None
+        }
+    } else {
+        // keep old default of assuming JSON
+        from_json(body).ok()
+    };
 
     err_body
         .map(ErrorBody::into_code_message)
--- a/sdk/core/src/error/http_error.rs
+++ b/sdk/core/src/error/http_error.rs
@@ -11,7 +11,7 @@ use serde::Deserialize;
 pub struct HttpError {
     status: StatusCode,
     details: ErrorDetails,
-    headers: impl Into<std::collections::HashMap<String, String>>,
+    headers: Headers,
     body: Bytes,
 }
 
--- a/sdk/core/src/error/http_error.rs
+++ b/sdk/core/src/error/http_error.rs
@@ -114,14 +114,20 @@ fn get_error_code_from_header(headers: &Headers) -> Option<String> {
 
 #[derive(Deserialize)]
 struct NestedError {
+    #[serde(alias = "Message")]
     message: Option<String>,
+    #[serde(alias = "Code")]
     code: Option<String>,
 }
 
+/// Error from a response body, aliases are set because XML responses follow different case-ing
 #[derive(Deserialize)]
 struct ErrorBody {
+    #[serde(alias = "Error")]
     error: Option<NestedError>,
+    #[serde(alias = "Message")]
     message: Option<String>,
+    #[serde(alias = "Code")]
     code: Option<String>,
 }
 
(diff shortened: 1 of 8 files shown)


sdk/core/Cargo.toml
```diff
@@ -27,16 +27,16 @@ reqwest = { version = "0.12.0", features = [
 ], default-features = false, optional = true }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-quick-xml = { version = "0.31", optional = true, features = ["serialize", "serde-types"] }
+quick-xml = { version = "0.31", features = ["serialize", "serde-types"] }
```
heaths, line 30: `azure_core` is used by all crates, so it would be nice if `xml` was still an opt-in feature and you could condition XML support on the feature, which identity could default opt into if needed.

Still, if it's available in the headers, that not only seems more straight forward but faster at runtime.

merlinio2000: Thanks! I implemented it as suggested, didn't consider that the individual crates can require the optional features of the core.

I don't think adding the `xml` feature from the identity sdk is necessary as to my knowledge only the storage API returns XML responses, please correct me if I'm wrong.

sdk/core/src/error/http_error.rs
```diff
@@ -1,14 +1,17 @@
...
 use serde::Deserialize;
-use std::collections::HashMap;
 
 /// An unsuccessful HTTP response
 #[derive(Debug)]
 pub struct HttpError {
     status: StatusCode,
     details: ErrorDetails,
-    headers: std::collections::HashMap<String, String>,
+    headers: Headers,
```
heaths, line 14: Perhaps more flexible since `Headers` implements it:
```suggestion
    headers: impl Into<std::collections::HashMap<String, String>>,
```

sdk/core/src/error/http_error.rs
```diff
@@ -89,46 +90,87 @@ struct ErrorDetails {
...
+pub(crate) fn get_error_code_from_header(headers: &Headers) -> Option<String> {
+    headers.get_optional_string(&headers::ERROR_CODE)
 }
 
 #[derive(Deserialize)]
 struct NestedError {
+    #[serde(alias = "Message")]
     message: Option<String>,
+    #[serde(alias = "Code")]
     code: Option<String>,
```
heaths, line 119: [REST API guidelines](https://aka.ms/azapi/guidelines) actually recommend camelCase. If you're seeing these PascalCase, that's an anomaly that shouldn't be accounted for here. None of our other languages even deserialize error details and this will eventually be removed anyway. If there's a service that has details you need, I suggest deserialize your own models or - for now, anyway - you limit it to the crate that is different i.e., the one using PascalCase.

If that's only for XML, this would hamper JSON deserialization as well. Some other way may be necessary even to the point, if possible (though I don't think it is for stock serde) to ignore case.

heaths (changes requested):
Looking much better, but a few changes would be good.

sdk/core/src/error/http_error.rs
merlinio2000: It is only for XML which is not included into the REST API guidelines and for XML PascalCase is the norm.
Would you prefer to have a separate `struct XmlErrorBody`? I did it using `serde(alias =)` because it required the least amount of code and thought the internal serde overhead would be negligible

heaths: No, this is fine. My apologies. For some reason I was thinking of `rename` instead of `alias`. I think this is fine.

sdk/core/src/error/http_error.rs
```diff
@@ -11,7 +11,7 @@ use serde::Deserialize;
 pub struct HttpError {
     status: StatusCode,
     details: ErrorDetails,
-    headers: Headers,
+    headers: impl Into<std::collections::HashMap<String, String>>,
```
heaths, line 14: This is a breaking change. Please keep as-is.

heaths (changes requested):
Please revert the `headers` change and it looks good to merge. We want to keep it as that type because callers can always convert, but `Headers` may provide additional APIs a simple `HashMap` may not.

sdk/core/src/error/http_error.rs
merlinio2000: Thanks a lot, makes sense.
Changed as suggested

heaths (approved):

PR #1636, opened 2024-04-08
chore: updated iothub device update api version to 2022-10-01

JanZachmann:


sdk/iot_deviceupdate/src/client.rs
```diff
@@ -45,19 +44,17 @@ impl DeviceUpdateClient {
...
-            endpoint,
             token_credential,
         };
         Ok(client)
     }
 
     async fn get_token(&self) -> azure_core::Result<AccessToken> {
         self.token_credential
-            .get_token(&[&self.endpoint])
+            .get_token(&["https://api.adu.microsoft.com/.default"])
```
heaths, line 57: Endpoints should not be hardcoded.

heaths (changes requested):
Did you generate this via autorust, or hand edit? I think it's great that you're upgrading to the GA version, but OpenAPI definitions can change between preview and release (I see it all the time since I'm on the review board). I see you dropped some endpoints so maybe that is the case, but I wanted to double check.

JanZachmann:
> Did you generate this via autorust, or hand edit? I think it's great that you're upgrading to the GA version, but OpenAPI definitions can change between preview and release (I see it all the time since I'm on the review board). I see you dropped some endpoints so maybe that is the case, but I wanted to double check.

It is hand edited. I just tried to fix the current incompatibility with current main. There are two issues:

1. `const API_VERSION: &str = "2021-06-01-preview";` isn't supported anymore
2. endpoint changed

sdk/iot_deviceupdate/src/client.rs
JanZachmann: done

sdk/iot_deviceupdate/src/client.rs
```diff
@@ -7,8 +7,9 @@ use const_format::formatcp;
 use serde::de::DeserializeOwned;
 use std::sync::Arc;
 
-pub(crate) const API_VERSION: &str = "2021-06-01-preview";
+pub(crate) const API_VERSION: &str = "2022-10-01";
 pub(crate) const API_VERSION_PARAM: &str = formatcp!("api-version={}", API_VERSION);
+pub(crate) const ENDPOINT: &str = "https://api.adu.microsoft.com/.default";
```
heaths, line 12: This isn't what I meant. The endpoint should not be hardcoded at all. It was a parameter before. Why did you remove it? If it's not correct - which I assume it's not - strip the URL down to its hostname and append "/.default". That's what we do in all our other languages most often.

heaths (changes requested):

sdk/iot_deviceupdate/src/client.rs
JanZachmann: hopefully done the way you expect?

heaths (approved):

Issue #1637, opened 2024-04-12
TopicClient always receive 401


Gaelik-git:
I created a Service Bus Namespace with: 
- One Queue : `dev`
- One Topic : `test-topic`
  - One Subscription : `test-subscription`

I also have a `Sharred access Policies` on the Namespace level named `common`
But the topic client always receives a 401, when my QueueClient receives 204 

Can you help me with this ? Is my code incorrect ? 
Can it be linked to the ordering when creating things ?
I am using the latest 0.19 version of the sdk with rust 1.77.1

This is the result of this test program
```
INFO peeked message queue val=204
INFO send message to topic ko=Error { context: Simple(HttpResponse { status: Unauthorized, error_code: None }) }
INFO peeked message from subscription val=401
```

```rs
use azure_messaging_servicebus::service_bus::{QueueClient, TopicClient};
use reqwest::Client;
use std::sync::Arc;
use std::time::Duration;
use tracing::{info, subscriber::set_global_default};

#[tokio::main]
async fn main() {
    // common conf
    let namespace = "REDACTED";
    let policy_name = "common";
    let policy_key = "REDACTED";

    // topic conf
    let topic_name = "test-topic";
    let subscription_name = "test-subscription";

    // queue conf
    let queue_name = "dev";

    set_global_default(tracing_subscriber::fmt::Subscriber::new())
        .expect("set tracing default subscriber");

    let client = QueueClient::new(
        Arc::new(Client::new()),
        namespace,
        queue_name,
        policy_name,
        policy_key,
    )
    .unwrap();
    let val = client
        .peek_lock_message2(Some(Duration::from_micros(1)))
        .await
        .unwrap();

    info!(val = %val.status(), "peeked message queue");
    let client = TopicClient::new(
        Arc::new(Client::new()),
        namespace,
        topic_name,
        policy_name,
        policy_key,
    )
    .unwrap();

    let val = client.topic_sender().send_message("msg").await;

    match val {
        Ok(_) => info!(ok = ?val.is_ok(), "send message to topic"),
        Err(_) => info!(ko = ?val.unwrap_err(), "send message to topic"),
    }

    let val = client
        .subscription_receiver(subscription_name)
        .peek_lock_message2(Some(Duration::from_micros(1)))
        .await
        .unwrap();

    info!(val = %val.status(), "peeked message from subscription");
}
```

Gaelik-git:
I think it might a bug introduced in v0.18

[...]

PR #1638, opened 2024-04-12
(fix) Signing key for TopicClient not correctly encoded

Gaelik-git:
Fixes this issue https://github.com/Azure/azure-sdk-for-rust/issues/1637
It's a copy of the fix make for QueueClient in this commit https://github.com/Azure/azure-sdk-for-rust/commit/c999ef3df4c8c65edf458e7948235e31bd44dc61

### DIFF
--- a/sdk/messaging_servicebus/src/service_bus/topic_client.rs
+++ b/sdk/messaging_servicebus/src/service_bus/topic_client.rs
@@ -47,6 +47,9 @@ impl TopicClient {
         P: Into<String>,
         K: Into<Secret>,
     {
+        // NOTE: This is to account for the azure_core::auth::hmac_sha256 assumption
+        // that the key needs to be base64 decoded.
+        let signing_key = azure_core::base64::encode(signing_key.into().secret());
         Ok(Self {
             http_client,
             namespace: namespace.into(),


sdk/messaging_servicebus/src/service_bus/topic_client.rs
```diff
@@ -47,6 +47,9 @@ impl TopicClient {
         P: Into<String>,
         K: Into<Secret>,
     {
+        // NOTE: This is to account for the azure_core::auth::hmac_sha256 assumption
+        // that the key needs to be base64 decoded.
+        let signing_key = azure_core::base64::encode(signing_key.into().secret());
```
heaths, line 52: How feasible is it just to change the `TopicClient` to accept the raw string? Seems a waste to encode it just to have it decoded.

heaths (changes requested):

sdk/messaging_servicebus/src/service_bus/topic_client.rs
Gaelik-git: I'm not sure it's feasible, but I think it's more important to correct this regression first.
We can think about a redesign afterwards.
Even if it's not ideal, doing this base64 encoding/decoding is probably not significant in terms of performance compared to the http calls made by the topic client. And the base64 thing is only done once when the client is created.

heaths (approved):

PR #1641, opened 2024-04-18
Remove all but core, identity

heaths:


heaths:
This is the start for our "track2" work. We'll just have core and identity initially. I wanted to leave in Key Vault, but it relies on generated code. In a future PR, I can write a simple `SecretClient` with which we can test various things like I have been in my prototype repos.

heaths:
There are a lot of files, yes, but it's almost all deletes. The brunt of the changes is to `git rm -rf services` and then all the `sdk/*` crates except `core` and `identity`. A few minor changes to a few other files e.g., tests so we don't try to test anything under `servces/`/.

jhendrixMSFT (approved):

PR #1658, opened 2024-05-05
Add if-match condition to PutAppendBlob

ishai42:


sdk/storage_blobs/src/blob/operations/put_append_blob.rs
```diff
@@ -31,6 +32,7 @@ impl PutAppendBlobBuilder {
                     headers.add(m);
                 }
             }
+            headers.add(self.if_match);
```
heaths, line 35: We tend to do `If-Match` and `If-None-Match` headers in a pair. Could you add support for `If-None-Match` as well?

ishai42: I believe these are handled by the match condition meaning the following fragment from the test generates if-none-match header:

```
    let result = blob
        .put_append_blob()
        .content_type("text/plain")
        .if_match(IfMatchCondition::NotMatch("*".into()))
        .await
        .unwrap_err();
```

ishai42: I guess you mean if-modified-since.   Added.

heaths: I meant that header, but didn't realize that's how it was being handled. If that's the case, then this code looked fine.

heaths (approved):
The added condition doesn't hurt, though.

heaths:
I'm working on fixing the new, breaking nightly clippy rule violations. At least they have fixes available.

heaths:
Blocked on #1667; I will rebase after that merges.

PR #1660, opened 2024-05-14
make ImdsManagedIdentityCredential public again

### LINKED ISSUE #1659: Support user-assigned managed identity
Using `azure_identity` prior to 0.20.0 I was able to use specific User Assigned Managed Identity credentials like this:
```rust
ImdsManagedIdentityCredential::default()
                    .with_object_id(&managed_identity_object_id)
```

For 0.20.0 there was an [overhaul of how credentials are created](https://github.com/Azure/azure-sdk-for-rust/pull/1532), and it no longer appears possible to create an ImdsManagedIdentity with an object_id.

This is a major issue for my project (and I imagine others) - we rely on User Assigned Managed Identities so can't upgrade `azure_identity` without a fix to restore this capability.

I do note there is an outstanding issue for creating a `ManagedIdentityCredential`: https://github.com/Azure/azure-sdk-for-rust/issues/1536

I'm happy to make a fix, but need to agree what the API should look like.

- In 0.19.0 the implementation of `ImdsManagedIdentityCredential` had [methods to allow you to set one of `object_id`, `client_id` or `identity` (resource id)](https://docs.rs/azure_identity/0.19.0/azure_identity/struct.ImdsManagedIdentityCredential.html).
  - Ideally we should be able to use any of these methods
- The current 0.20.0 `ImdsManagedIdentityCredential` implementation does have an enum defined that includes all the different types of `ImdsManagedIdentityCredential`, so we just need to provide a way to create instances with an id of each of these types.
    ```rust
    pub(crate) enum ImdsId {
        SystemAssigned,
        ClientId(String),
        ObjectId(String),
        MsiResId(String),
    }
    ```

The previous method of creating an `ImdsManagedIdentityCredential` and then calling one of the other methods to set a value felt a bit hacky.  Might be simpler to expose the above `ImdsId` enum in the public API, and then allow it to be passed in a new constructor. 
 Although possibly a bit odd passing in `ImdsId` to a `ManagedIdentityCredential` (without the `Imds` prefix).  Could rename `ImdsId` in the API, perhaps:

```rust
ManagedIdentityCredential::new(id: ImdsId, options: impl Into<TokenCredentialOptions>)
```

For comparison, the .NET SDK `ManagedIdentityCredential` has multiple overloaded constructors for creating the different variants:
- https://learn.microsoft.com/en-us/dotnet/api/azure.identity.managedidentitycredential?view=azure-dotnet#constructors

Any thoughts/suggestions appreciated.

cataggar:
Fixes #1659. As it points out, I think it was a bit too early to make `ImdsManagedIdentityCredential` private in #1532. I think it is the right direction after the credentials that use it like `VirtualMachineManagedIdentityCredential` and `AppServiceManagedIdentityCredential` add support for configurating user managed identities.

### DIFF
--- a/sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs
+++ b/sdk/identity/src/token_credentials/imds_managed_identity_credentials.rs
@@ -14,7 +14,7 @@ use std::{str, sync::Arc};
 use time::OffsetDateTime;
 
 #[derive(Debug)]
-pub(crate) enum ImdsId {
+pub enum ImdsId {
     SystemAssigned,
     #[allow(dead_code)]
     ClientId(String),
@@ -30,7 +30,7 @@ pub(crate) enum ImdsId {
 ///
 /// Built up from docs at [https://docs.microsoft.com/azure/app-service/overview-managed-identity#using-the-rest-protocol](https://docs.microsoft.com/azure/app-service/overview-managed-identity#using-the-rest-protocol)
 #[derive(Debug)]
-pub(crate) struct ImdsManagedIdentityCredential {
+pub struct ImdsManagedIdentityCredential {
     http_client: Arc<dyn HttpClient>,
     endpoint: Url,
     api_version: String,


heaths (approved):

RobJellinghaus:
@cataggar is there some reason this never went in? Still seems worth landing? I am looking at doing UAMI auth and this would help.

cataggar:
> @cataggar is there some reason this never went in? Still seems worth landing? I am looking at doing UAMI auth and this would help.

I just haven’t been active in this project lately. At the time, there was an unrelated CI failure. I’ll have a look.

cataggar:
I targeted new sdk branch in #1799.

PR #1661, opened 2024-05-14
cargo fmt sdk/iot_deviceupdate/src/client.rs

cataggar:
Other PR was failing without this file formatted.


heaths (approved):

heaths:
Seems no longer necessary as we've done a number of full builds without seeing an issue.

PR #1662, opened 2024-05-15
Move from http-types to http

### LINKED ISSUE #1644: Consider replacing http-types crate with http crate

svix-jplatte:
Closes #1644.

For the second commit, I used sed as I couldn't figure out the right rev of `azure-rest-api-specs` to check out to get identical output (aside from the intended changes). When I had it checked out at the rev referenced in #1496 (the latest such PR I could find), it still added and removed a bunch of files.

I split the PR into two commits to make review easier, but this should likely be squash-merged as the first commit does not compile on its own.


svix-jplatte:
@microsoft-github-policy-service agree company="Svix"

heaths (changes requested):
Blocked on discussion in #1644. Also, this may be too big of a change for unofficial sources given near-term plans.

svix-jplatte:
Re. the size of the change, would it help if you just re-did the second commit on your end? It's really just the sed command I put into its description, which needs to be run for the following replacements one-by-one in addition the `Get` one already provided:

- `s/azure_core::Method::Head/azure_core::Method::HEAD/g`
- `s/azure_core::Method::Post/azure_core::Method::POST/g`
- `s/azure_core::Method::Put/azure_core::Method::PUT/g`
- `s/azure_core::Method::Delete/azure_core::Method::DELETE/g`
- `s/azure_core::StatusCode::Ok/azure_core::StatusCode::OK/g`
- `s/azure_core::StatusCode::Accepted/azure_core::StatusCode::ACCEPTED/g`
- `s/azure_core::StatusCode::Created/azure_core::StatusCode::CREATED/g`
- `s/azure_core::StatusCode::NoContent/azure_core::StatusCode::NO_CONTENT/g`

heaths:
`main` branch has been moved to `legacy`. Closing this PR since `legacy` is unsupported, but we'll consider for the newer crates we'll be working on in `main`.

PR #1663, opened 2024-05-15
Update azure-autorust to append path and parameters via URL lib

andras-pinter:
When returning a URL from Client::endpoint the URL contains a tailing slash in any case. This is handled by the URL lib which will append the tailing slash to the end upon calling Display::fmt. Therefore formatting the returned URL from Client::endpoint will result in a duplicated slash, which could cause some (not all) Azure endpoints to have some trouble.

For example:
Base URL: https://some-app-conf.azconfig.io/
AppConfigUrl: https://some-app-conf.azconfig.io//keys?api-version=2023-10-01 This resulted in a 404 response.

Removing the tailing slash from the base endpoint's URL manually does not solve the problem, due to the reason mentioned above.

This patch applies a fix to append (set) the formatted path via rust URL lib, which will handle the duplicated slash (or any other problem).

Although I tried to locally generate `mgmt` and `svc` crates, I encountered some problems, and `autorust` removed and added a bunch of new files and modules.
So I'd like to ask somebody to help me out with the generation.

### DIFF
--- a/services/autorust/codegen/src/codegen_operations/request_builder_send.rs
+++ b/services/autorust/codegen/src/codegen_operations/request_builder_send.rs
@@ -38,8 +38,7 @@ impl ToTokens for RequestBuilderSendCode {
             quote! { &self.#url_arg }
         });
         let url_str_args = quote! { #(#url_args),* };
-
-        let fpath = format!("{{}}{}", &format_path(&new_request_code.path));
+        let fpath = PARAM_RE.replace_all(&new_request_code.path, "{}");
 
         let mut match_status = TokenStream::new();
         for status_response in &self.response_code.status_responses {
@@ -58,7 +57,8 @@ impl ToTokens for RequestBuilderSendCode {
             let api_version = &self.request_builder.api_version;
             quote! {
                 fn url(&self) -> azure_core::Result<azure_core::Url> {
-                    let mut url = azure_core::Url::parse(&format!(#fpath, self.client.endpoint(), #url_str_args))?;
+                    let mut url = self.client.endpoint().clone();
+                    url.set_path(&format!(#fpath, #url_str_args));
 
                     let has_api_version_already = url.query_pairs().any(|(k, _)| k == azure_core::query_param::API_VERSION);
                     if !has_api_version_already {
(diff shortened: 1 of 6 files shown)


andras-pinter:
Hey @heaths,

Yes, there is a particular error, what I recently encountered. Please see:
https://github.com/Azure/azure-sdk-for-rust/issues/1664

heaths:
@ctaggart could you review this? If you're good with it, I can regenerate the service crates.

andras-pinter:
My goal was to let `azure_core::Url` handle URL creation :)

services/autorust/codegen/src/codegen_operations/request_builder_send.rs
```diff
@@ -58,7 +57,8 @@ impl ToTokens for RequestBuilderSendCode {
             let api_version = &self.request_builder.api_version;
             quote! {
                 fn url(&self) -> azure_core::Result<azure_core::Url> {
-                    let mut url = azure_core::Url::parse(&format!(#fpath, self.client.endpoint(), #url_str_args))?;
+                    let mut url = self.client.endpoint().clone();
+                    url.set_path(&format!(#fpath, #url_str_args));
```
cataggar, line 61: This change looks good to me.

cataggar (approved):

cataggar:
> @ctaggart could you review this? If you're good with it, I can regenerate the service crates.

Sounds good. Please go ahead.

kingofthehill444:
I'm facing this exact issue when trying to use the azure web pub sub crate. 

I'm happy to help out but it seems like the change just needs to be merged? Can someone help push this through please?

andras-pinter:
I made this as a PR instead of a draft. Still facing some issues, if somebody please help me finish this would be really nice! Thanks!

heaths:
Most of the issues were with nightly tests failing due to evolving lints. Checking if a rebase on `main` fixes them first.

heaths (approved):

andras-pinter:

andras-pinter:
🎉  Thanks everyone!

PR #1667, opened 2024-05-24
Fix new nightly clippy rules

heaths:


heaths:
clippy keeps adding new lints to nightly and breaking the build. This is another batch of fixes.

RickWinter (approved):

PR #1670, opened 2024-05-28
No nightly clippy except weekly

heaths:
Also removes dependabot.yml since defaults should be sufficient. This was causing problems with the Azure/azure-sdk-for-rust-pr repo.


heaths:
@weshaggard does this look right to you - to run a command against a specific toolchain weekly on Sunday night?

.github/workflows/build.yml
```diff
@@ -113,12 +115,16 @@ jobs:
...
+            clippy: false
             experimental: true
+          - clippy: true
     steps:
       - uses: actions/checkout@v4
       - uses: Swatinem/rust-cache@v2
         with:
           workspaces: services
+      - run: echo "NO_CLIPPY=true" >> $GITHUB_ENV
+        if: github.event_name != 'schedule' && ! matrix.clippy
```
weshaggard, line 127:
```suggestion
        if: github.event_name != 'schedule' && !matrix.clippy
```

weshaggard (approved):

.github/workflows/build.yml
heaths: While this is probably fine, all the samples on docs.github.com show a space so I'd prefer to leave it. Troubleshooting pipelines is too time-consuming.

PR #1671, opened 2024-05-28
Fixes scheduled pipeline

heaths:
Fixes #1670


weshaggard (approved):

heaths:
I'm setting auto-merge now because the changes appear to be working, unlike #1670 that didn't work right but auto-merged because none of the affected checks were required. 🤦

PR #1684, opened 2024-06-21
Support hierarchical Context

### LINKED ISSUE #1655: Need to design a context that doesn't always flow to callees
We need a way to start over with a new `Context` - optionally keeping telemetry spans - when calling into other crates' methods like `azure_identity`. We don't want, for example, `Retry` options for Key Vault clients to pass through to `DefaultAzureCredential`. See Go for more context and solution concept ideas.

heaths:
Conceptually similar to Go, which was also similarly copied for C++. Instead of being hierarchical, however, which introduces complexity into `len`, `is_empty`, etc. - though we might only want those for testing by GA - this uses a `Cow` to copy the `HashMap` only when necessary.

Closes #1655


sdk/core/src/context.rs
```diff
@@ -1,66 +1,110 @@
 use std::any::{Any, TypeId};
+use std::borrow::{Borrow, Cow};
 use std::collections::HashMap;
 use std::sync::Arc;
 
 /// Pipeline execution context.
 #[derive(Clone, Debug)]
-pub struct Context {
-    type_map: HashMap<TypeId, Arc<dyn Any + Send + Sync>>,
+pub struct Context<'a> {
```
JeffreyRichter, line 8: I feel that this is a bit over-complicated.
My mental model is that there is a root Context with children who have grandchildren, etc. 
A Context be canceled (explicitly or via time) and this affects all children.
A Context can have a key/value pair set on it which also affects all children (if they explicitly look for the key to get its value).
And, at some point in the tree, a child can be handed to another thread.
The only mutable field in a context is the cancellation time (if a Context is canceled explicitly) which has to be updated in a thread-safe way to NOW.
Once a new Context is created with key/value pairs, these key/value pairs are immutable; no need to update or remove. If this is required for some reason, then code can always go back to the parent Context and derive a new Context from that.
Returning up a call stack, implicitly removes leaf-node Contexts from the tree.

If you have a different mental model in mind, we need to discuss it.
If you think this accomplishes the above mental model, then I question the existence of functions that mutate/remove the key/value pairs (which I saw below).

LarryOsterman: QQ: In Rust, is cancellation via Context common? Or are operations typically cancelled by canceling the future returned by the operation?

JeffreyRichter: Oh yes, you're right. I completely forgot. In Rust, cancelation is done by dropping the future. So, for Rust, the Context is just about the key/value pairs. Therefore, I think a Context can/should be completely immutable but it can copy the key/values from a parent context with some new key/values. I'll still maintain that there is no need to be able to update or remove a key/value from a context as a new one can be created from the same parent.

(Thanks for reminding me Larry)

heaths: I don't want to copy unless necessary, hence the `Cow`. And even to take a *reference* to the parent requires a lifetime and we always have to look up the stack so the UX will be no different.

Yes, context is not and will not be used for cancellation. That's not idiomatic. As previously discussed, this was really only for things clients need to pass through pipelines or, as an "advanced" user scenarios for now, we agreed I can add "extension methods" like configuring per-call retries.

heaths: That said, I did consider over the weekend removing `remove`, but I want to keep `len` for testing. I can attribute it with `#[test]` for that purpose.

LarryOsterman (approved):

JeffreyRichter (approved):

PR #1692, opened 2024-07-01
Implement RequestContent

heaths:
Used in method parameters that take a body.

### DIFF
--- a/sdk/core/src/request.rs
+++ b/sdk/core/src/request.rs
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
 #[cfg(not(target_arch = "wasm32"))]
 use crate::SeekableStream;
 use crate::{
@@ -7,15 +10,18 @@ use crate::{
 };
 use bytes::Bytes;
 use serde::Serialize;
-use std::fmt::Debug;
+use std::{fmt::Debug, marker::PhantomData, str::FromStr};
 
 /// An HTTP Body.
 #[derive(Debug, Clone)]
 pub enum Body {
     /// A body of a known size.
     Bytes(bytes::Bytes),
+
     /// A streaming body.
+    ///
     /// This is not currently supported on WASM targets.
+
     // We cannot currently implement `Body::SeekableStream` for WASM
     // because `reqwest::Body::wrap_stream()` is not implemented for WASM.
     #[cfg(not(target_arch = "wasm32"))]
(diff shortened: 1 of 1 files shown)


heaths:
Sorry @RickWinter , @LarryOsterman. I was targeting the wrong base again. Hopefully this PR looks more palatable...and reviewable :wink:

LarryOsterman (approved):

JeffreyRichter (approved):

sdk/core/src/request.rs
```diff
@@ -60,6 +66,18 @@ impl From<Box<dyn SeekableStream>> for Body {
     }
 }
 
+#[cfg(test)]
+impl PartialEq for Body {
```
RickWinter, line 70: What does the `Eq` mean in `PartialEq`?

sdk/core/src/request.rs
```diff
@@ -148,3 +166,140 @@ impl Request {
...
+    fn from_str(body: &str) -> Result<Self, Self::Err> {
+        let body: Bytes = Bytes::copy_from_slice(body.as_bytes());
+        Ok(Self {
+            body: Body::Bytes(body),
+            phantom: PhantomData,
+        })
+    }
+}
+
+#[cfg(test)]
```
RickWinter, line 247: Food for thought: 
Tests in the file is common and an idiomatic pattern. I have also seen patterns for putting tests in *_test.rs files. It might be cleaner to have that model.

RickWinter (approved):

sdk/core/src/request.rs
LarryOsterman: FWIW, there are strong technical reasons for tests in the same source file - specifically, tests in the same source file can access private fields in structures.

Typically, I've seen the _test.rs files used for integration or non-unit tests, and the `#[cfg(test)]` in source files used for unit tests.

RickWinter: Within the same folder you can get the same via use of `super`.

heaths: Tests in separate files are generally integration tests: https://doc.rust-lang.org/book/ch11-03-test-organization.html

If you put them in separate files, they are separate modules and forces your non-test module to make everything public at least to the crate, which is not always what you want. Submodules have access to ancestors' privates.

I think we should remain consistent.

FWIW, you don't even need the ol'

```rust
#[cfg(test)]
mod tests {
  use super::*;
}
```

It's an easy way to condition things, but it's not uncommon to see tests interspersed with functions, since the `#[test]` procedural macro automatically conditions on `test`:

```rust
fn foo() -> bool {
  true
}

#[test]
fn test_foo() {
  assert!(foo());
}

fn bar() -> &'static str {
  "hello, world!"
}

#[test]
fn test_bar() {
  assert_eq!("hello, world!", bar());
}
```

In function-heavy files, this can actually be handy.

PR #1693, opened 2024-07-02
Adding BrokerProperties when sending message to queue or topic

Gaelik-git:
Following this documentation : https://learn.microsoft.com/en-us/rest/api/servicebus/message-headers-and-properties#message-headers

I would like the possibility to set an ScheduledEnqueueTimeUtc (other option could be added in the future)

I updated the existing send_message method, making to modification a breaking change. Do you think it's ok like this ?
Should I create a new method for sending messages with options set ?

Thanks


sdk/messaging_servicebus/src/service_bus/mod.rs
```diff
@@ -89,17 +89,25 @@ async fn send_message(
     policy_name: &str,
     signing_key: &Secret,
     msg: &str,
+    broker_properties: Option<SettableBrokerProperties>,
```
heaths, line 92: I'd be fine with the breaking change, but if we're going to break it once - since it's all unsupported and prerelease anyway - lets keep it at once. Instead, it would be better to have an options bag following our current guidelines in part, using the method name as the root + "Options" e.g., `SendMessageOptions` or, rather, `Option<SendMessageOptions>`. That should have member(s) including this `broker_properties` header value. It might also be worth it to implement `AsHeader` like all our other headers in `azure_core`, even if only for this crate since it's special-purposed to Service Bus.

heaths (changes requested):
Thanks for improving Service Bus SDK! One blocking comment, though. I'm okay with the break in this case, but let's not special-case too much when we do that. With an options bag, we can add more in the future without breaking.

sdk/messaging_servicebus/src/service_bus/mod.rs
Gaelik-git: Thanks for the feedback !
I just updated the PR, I also added all the other options in BrokerProperties (I did not test all of them)

I just not sure about the `unwrap` in the `AsHeaders` implementation

sdk/messaging_servicebus/src/service_bus/mod.rs
```diff
@@ -21,6 +25,7 @@ use url::form_urlencoded::{self, Serializer};
 
 /// Default duration for the SAS token in days — We might want to make this configurable at some point
 const DEFAULT_SAS_DURATION: u64 = 3_600; // seconds = 1 hour
+pub const BROKER_PROPERTIES: HeaderName = HeaderName::from_static("brokerproperties");
```
heaths, line 28: Nit: why is this public? Better to expose less and open it up as needed that expose it prematurely and have to support it forever (appreciating that this isn't supported anyway, but in general).

heaths (approved):
Thank you! I just have one question/nit but otherwise LGTM. Once I get this merged, I'll publish new versions.

sdk/messaging_servicebus/src/service_bus/mod.rs
Gaelik-git: you are write it probably does not have to be

heaths (approved):

heaths:
Thank you for making the requested changes!

PR #1694, opened 2024-07-02
Add Microsoft Fabric cloud location

kevinzwang:
Microsoft Fabric uses a different endpoint from the Azure cloud, this PR adds support for this endpoint.

### DIFF
--- a/sdk/storage/src/cloud_location.rs
+++ b/sdk/storage/src/cloud_location.rs
@@ -7,6 +7,8 @@ use azure_core::Url;
 pub enum CloudLocation {
     /// Azure public cloud
     Public { account: String },
+    /// Microsoft Fabric
+    Fabric { account: String },
     /// Azure China cloud
     China { account: String },
     /// Use the well-known emulator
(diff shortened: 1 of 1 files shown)


kevinzwang:

heaths (changes requested):
@seanmcc-msft does this change seem appropriate for Storage, or any service? We don't have a Service Fabric endpoint for any other language, and I'm surprised to see a fixed domain but maybe that is valid for Storage at least?

kevinzwang:
I'm not sure how this fits into other services, but it's just something I noticed was missing from the storage API compared to the [`object_store`](https://crates.io/crates/object_store) crate: https://github.com/apache/arrow-rs/blob/035b5899f3198cbbdddc772f64c214332e6323fe/object_store/src/azure/builder.rs#L846-L849

heaths:
`main` branch has been moved to `legacy`. Closing old PRs that don't seem to meet the criticality bar. If you disagree, please reopen or switch to official, supported crates from `main` once available.

PR #1695, opened 2024-07-02
Add "root" crate for TypeSpec

heaths:
Serves as a package-as-a-namespace for RFC 3243:
https://rust-lang.github.io/rfcs/3243-packages-as-optional-namespaces.html.

We may put common types between possible client- and service-side code
generators in here as well.


heaths:
If approved, I plan to publish this immediately.

sdk/typespec/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "typespec"
+version = "0.1.0"
+authors.workspace = true
+edition.workspace = true
+description = "Project root for all TypeSpec-related crates"
+homepage = "https://aka.ms/typespec"
```
timotheeguerin, line 7:
```suggestion
homepage = "https://typespec.io"
```

Maybe that aka.ms link should be updated but it links to the telemetry app.

heaths: Good idea!

heaths:
Adding @ronniegeraghty. We can fill out this README more (which will also appear on crates.io) as we do more with it.

LarryOsterman (approved):

PR #1697, opened 2024-07-08
Adds Cold as AccessTier option

gurkanindibay:
Azure Storage has 4 AccessTier option. 
* Hot
* Cold
* Cool
* Archive
here in AccessTier enum , Cold does not exist which causes Unsupported type error

### DIFF
--- a/sdk/storage_blobs/src/options/access_tier.rs
+++ b/sdk/storage_blobs/src/options/access_tier.rs
@@ -3,6 +3,7 @@ use azure_core::headers::{self, Header};
 create_enum!(
     AccessTier,
     (Hot, "Hot"),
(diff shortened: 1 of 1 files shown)


heaths (approved):

heaths:
Thank you for the contribution! @jaschrep-msft you good with this change to our unofficial sources until we have official track 2 replacements? Aligns with the docs but would like to get your's or @seanmcc-msft's approval on concept at least.

jaschrep-msft (approved):
My only concern would have been if this library supports the service version necessary for this access tier, but according to the REST docs `The list of supported tiers is not restricted by the request version, and new tiers may be added in the future.`

PR #1699, opened 2024-07-10
Redact the authorization header from Debug impl

arpad-m:
Right now, if you use the azure SDK and enable debug logging, the authorization header is printed in the debug logs through [this](https://github.com/Azure/azure-sdk-for-rust/blob/8c4caa251c3903d5eae848b41bb1d02a4d65231c/sdk/core/src/policies/transport.rs#L32) place. Printing the authorization token is dangerous however as it might get into the logging infrastructure, and become available to people with read access to the logs.

Therefore, this commit adjusts the `Debug` impl of the `Headers` struct to not print the content of the authorization header.

### DIFF
--- a/sdk/core/src/headers/mod.rs
+++ b/sdk/core/src/headers/mod.rs
@@ -51,7 +51,7 @@ pub trait Header {
 }
 
 /// A collection of headers
-#[derive(Clone, Debug, PartialEq, Eq, Default)]
+#[derive(Clone, PartialEq, Eq, Default)]
 pub struct Headers(std::collections::HashMap<HeaderName, HeaderValue>);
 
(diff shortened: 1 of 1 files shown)


arpad-m:
@microsoft-github-policy-service agree company="Neon, Inc."

heaths:
Thanks. This is a good temp workaround for this problem. I've opened an issue to track a larger effort and whether we should even do that for `Debug` impls. Probably, since it'd be easy for devs to just use it in logging and tracing, for example. /cc @JeffreyRichter @RickWinter

heaths (approved):

arpad-m:
Rebased due to CI failure on nightly.

PR #1705, opened 2024-07-16
Move crates into subdirs under service dirs

heaths:
The exception in this commit is `typespec`, which will likely be a root project for RFC 3243 (packages-as-namespaces) with packages thereunder.


RickWinter:
This is against main, shouldn't it be against track2?

heaths:
> This is against main, shouldn't it be against track2?

Yes, and I'm too used to just running `gh pr create`. I need to work on that feature request I opened - and was invited to work on - to discover the closest branch. Unfortunately, that's actually quite a hard problem to solve reliably. It'd really be nice (as I told the gh CLI team) if GH had an endpoint / GraphQL query to do that based on their centralized info rather than what's cloned locally. For example, I tend to prune branch specs from the upstream repo to avoid lengthy pulls in repos with a lot of contributors.

heaths:
@RickWinter so you're good with moving all crates to subdirs under service dirs? See internal Teams thread for some additional thoughts / questions. Figured more people would see that. Open to feedback here too, of course.

PR #1708, opened 2024-07-23
Implement client options, method options builders

heaths:


sdk/core/azure_core/src/options/retry.rs
```diff
@@ -0,0 +1,208 @@
...
+///
+/// The default is an exponential retry policy using the default `ExponentialRetryOptions`.
+#[derive(Clone, Debug, Default)]
+pub struct RetryOptions {
+    /// The algorithm to use for calculating retry delays.
+    mode: RetryMode,
+}
+
+impl RetryOptions {
+    /// A retry strategy where attempts happen at intervals that get exponentially longer with each retry.
```
LarryOsterman, line 56: Does filling in the rest of the doccomment (parameter description for instance) make sense now?

heaths: I don't understand. Could you clarify what you mean?

LarryOsterman (approved):

sdk/core/azure_core/src/options/retry.rs
LarryOsterman: It was a generic comment related to a lot of the parts of this file. For instance, the to_policy method has no doccomment, there are methods with a 1 line description and parameters without descriptions, etc.

I assume that all this will need to be filled in at some point or other, but I was wondering if now was the right time, or if it was going to be handled in a later PR.

heaths: Thanks for clarifying offline. Retries will be changing a bit when I refactor and I was planning on doing it then to see what survives as-is (or close). I opened #1709 to track.

PR #1710, opened 2024-07-23
Refactor general implementation into TypeSpec crates

heaths:
This is part of the unbranding work across languages. For now we have decided to brand around the "TypeSpec runtime" as some other Azure SDK languages have considered.


sdk/identity/azure_identity/src/refresh_token.rs
```diff
@@ -10,6 +10,7 @@ use azure_core::{json::from_json, Method};
 use serde::Deserialize;
 use std::fmt;
 use std::sync::Arc;
+use typespec_client_core::error::http_response_from_body;
```
RickWinter, line 13: Do we want references to TypeSpec core showing up everywhere or should these be exposed from azure_core? This is a general design pattern

//cc @bterlson

RickWinter (approved):

sdk/identity/azure_identity/src/refresh_token.rs
heaths: Probably re-export them from `azure_core`. I did that with all the others but was reluctant to ship these. That said, this is effectively an "internal" method so customers wouldn't be expected to use this in their code. It's for us only. In those cases, I don't really have a strong opinion on re-exporting or not from `azure_core`.

heaths:
Sorry, @LarryOsterman. I meant to delete my pending comment (just wanted to submit a reply in the Review tab), but deleted yours by mistake. It's resolved here: #1730

sdk/identity/azure_identity/src/refresh_token.rs
heaths: Re-exported

Issue #1707, opened 2024-07-17
Redact Debug printing by default


heaths:
Re: #1699

After some discussion, we have decide to keep the `SafeDebug` trait but change how it works so that we're secure by default but devs can opt into more, potentially insecure (if they're not careful) "logging".

The main concern here is that, by implementing `Debug` on headers, models, etc., anyone - us in tracing, callers in logging or "printf debugging" - could leak PII. So `SafeDebug` only logs the type. With a [TypeSpec decorator](https://github.com/Azure/typespec-azure/issues/2540) we could drive this in authoring, but lacking that in the immediate future we need a solution.

What @jhendrixMSFT came up with is a feature - off by default - that opts the existing `SafeDebug` derive macro into just deriving `Debug` if enabled. By default still, then, we'd show just the type name and ".." if there are any members (tuple or otherwise).

We also need to make sure that, w.r.t, #682, we never use the `Debug` trait in any way.

For documentation, we'd document this feature in `azure_core`'s `README.md` (as well as in `typespec_client_core` where we'll ultimately have to support this) and then each higher-level client library's `README.md`'s "Troubleshooting" section would make reference to it.

The main section should discuss how it works by default, how to enable more, and why devs should be careful if they do enable more.

heaths:
For headers, one idea is to attach data to mark which known headers are safe. Any known header which is not attributed with such data or is unknown is redacted. This is similar to what we did in .NET/C# and what, IIRC, most other SDK languages do.

E.g.,

```rust
use std::{borrow::Cow, fmt};

#[derive(Debug)]
struct HeaderName {
    name: Cow<'static, str>,
    safe: bool,
}

impl HeaderName {
    pub const fn new(name: &'static str, safe: bool) -> Self {
        Self {
            name: Cow::Borrowed(name),
            safe,
        }
    }

    pub fn safe(&self) -> bool {
        self.safe
    }
}

impl fmt::Display for HeaderName {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.write_str(&self.name)
    }
}

impl From<&'static str> for HeaderName {
    fn from(name: &'static str) -> Self {
        Self {
            name: name.into(),
            safe: false,
        }
    }
}

impl PartialEq for HeaderName {
    fn eq(&self, other: &HeaderName) -> bool {
        PartialEq::<str>::eq(self, &other.name)
    }
}

impl PartialEq<str> for HeaderName {
[...]

PR #1711, opened 2024-07-24
Refactor RawResponse and Response<T> into a single Response<T> that can contain a ResponseBody

analogrelay:
In starting up on the Cosmos Rust SDK, I found that the `Response<T>` type requires that the caller know the serialization format used, in order to call `.json` or `.xml`. Since this is the type that client methods are [expected to return](https://azure.github.io/azure-sdk/rust_introduction.html#rust-client-methods-return) it seems like the layering isn't quite right there.

So, what I've got here is an idea for a little bit of a different approach. Currently there is `RawResponse`, which carries the actual data (status code, headers, body stream) and `Response<T>` which wraps the `RawResponse` and allows you to read it as a `T` (using either `.json` or `.xml` methods).

My proposal here is a little different. Instead, there is a single `Response<T>` type, which contains the actual `T` as a field (instead of using a phantom data field). The HTTP pipeline is then expected to return `Response<ResponseBody>` (using the existing `ResponseBody` type that wraps a "pinned stream"). This type has the `.json` and `.xml` properties on it (and any other future serialization formats we choose to add). Those methods, rather than just parsing the body, _transform_ a `Response<ResponseBody>` **into** a `Response<T>` (for a given `T: DeserializeOwned`) by deserializing the body.

This means that a client method is now implemented something like this:

```rust
pub async fn read_database(&self) -> azure_core::Result<azure_core::Response<DatabaseResponse>> {
    let req = Request::new("...", azure_core::Method::Get);
    let ctx = Context::new();
    let resp = self.pipeline.send(&ctx, &mut req).await?; // Returns a Response<ResponseBody>
    // ... early-return with an error if the response indicates failure ...
    let parsed_resp = resp.json().await?; // Returns a Response<DatabaseResponse> (inferred from the function return type)
    Ok(parsed_resp)
}
```

A few quirks here:

* I changed `Response<T>` to implement `Debug` only when `T: Debug`, to allow for `Debug` rendering of `Response<SomeClientSpecificObject>`. The `ResponseBody` type implements `Debug` using a placeholder value so the effect is mostly the same. I think it's unlikely someone will want to `Debug`-render and `Response<T>` where `T` is not `Debug`
* A few existing `azure_identity` APIs needed to be updated to return `Response<T>`. This also means that some of the places that _used_ those methods have to call `into_body` to unwrap the Response into the body type.

@heaths I'd appreciate your thoughts here. It should unblock me to be able to at least start an initial Cosmos DB SDK that can call _an_ API. We're still missing things like `Pager` in order to get into the core APIs, but this should allow us to put the initial package in place.

### DIFF
--- a/eng/test/mock_transport/src/mock_response.rs
+++ b/eng/test/mock_transport/src/mock_response.rs
@@ -33,18 +33,18 @@ impl MockResponse {
 
     pub(crate) async fn duplicate(response: Response<ResponseBody>) -> error::Result<(Response<ResponseBody>, Self)> {
         use error::ResultExt;
-        let (status, headers, body) = response.deconstruct();
+        let (status_code, header_map, body) = response.deconstruct();
         let response_bytes = body.collect().await.context(
             error::ErrorKind::Io,
             "an error occurred fetching the next part of the byte stream",
         )?;
 
         let response = Response::new(
-            status.clone(),
-            headers.clone(),
+            status_code,
+            header_map.clone(),
             ResponseBody::new(Box::pin(BytesStream::new(response_bytes.clone()))),
         );
-        let mock_response = MockResponse::new(status, headers, response_bytes);
+        let mock_response = MockResponse::new(status_code, header_map, response_bytes);
 
         Ok((response, mock_response))
     }
--- a/eng/test/mock_transport/src/mock_response.rs
+++ b/eng/test/mock_transport/src/mock_response.rs
@@ -1,8 +1,4 @@
-use azure_core::{
-    base64, error,
-    headers::{HeaderName, HeaderValue, Headers},
-    BytesStream, RawResponse, StatusCode,
-};
+use azure_core::{base64, error, headers::{HeaderName, HeaderValue, Headers}, BytesStream, StatusCode, Response, ResponseBody};
 use bytes::Bytes;
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
@@ -14,14 +10,14 @@ pub(crate) struct MockResponse {
     body: Bytes,
 }
 
-impl From<MockResponse> for RawResponse {
+impl From<MockResponse> for Response<ResponseBody> {
     fn from(mock_response: MockResponse) -> Self {
         let bytes_stream: azure_core::BytesStream = mock_response.body.into();
 
         Self::new(
             mock_response.status,
             mock_response.headers,
-            Box::pin(bytes_stream),
+            ResponseBody::new(Box::pin(bytes_stream)),
         )
     }
 }
@@ -35,20 +31,20 @@ impl MockResponse {
         }
     }
 
-    pub(crate) async fn duplicate(response: RawResponse) -> error::Result<(RawResponse, Self)> {
+    pub(crate) async fn duplicate(response: Response<ResponseBody>) -> error::Result<(Response<ResponseBody>, Self)> {
         use error::ResultExt;
-        let (status_code, header_map, body) = response.deconstruct();
+        let (status, headers, body) = response.deconstruct();
         let response_bytes = body.collect().await.context(
             error::ErrorKind::Io,
             "an error occurred fetching the next part of the byte stream",
         )?;
 
-        let response = RawResponse::new(
-            status_code,
-            header_map.clone(),
-            Box::pin(BytesStream::new(response_bytes.clone())),
+        let response = Response::new(
+            status.clone(),
+            headers.clone(),
+            ResponseBody::new(Box::pin(BytesStream::new(response_bytes.clone()))),
         );
-        let mock_response = MockResponse::new(status_code, header_map, response_bytes);
+        let mock_response = MockResponse::new(status, headers, response_bytes);
 
         Ok((response, mock_response))
     }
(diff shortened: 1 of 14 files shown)


RickWinter:
Added Larry and Joel to the discussion.

heaths:
> I changed Response<T> to implement Debug only when T: Debug, to allow for Debug rendering of Response<SomeClientSpecificObject>. The ResponseBody type implements Debug using a placeholder value so the effect is mostly the same. I think it's unlikely someone will want to Debug-render and Response<T> where T is not Debug

We'll want to be careful here with deriving or implementing `Debug` in general: #1707. It's great for debugging, but I worry about people that would blindly `println!("{:#?}", model)` or trace it. We can redact headers easily enough by attaching "safe" markers to known headers and redact any unknowns, but I don't know how to solve this for models yet. /cc @RickWinter @JeffreyRichter

heaths:
> We're still missing things like Pager in order to get into the core APIs, but this should allow us to put the initial package in place.

@analogrelay It's there currently as [`azure_core::Pageable`](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/core/src/pageable.rs). This will be getting refactored soon. For the most part, only the name changes.

heaths:
After an initial glance, I think this makes sense. I'd love @demoray's thoughts as well and will take a look at this more on Monday.

The main goal with some of the changes I already made to `Response<T>` and related types was deserializing late (which it already did, to be fair) but also retaining the model type. The unsupported "track 1" code returned bespoke response types from which you could get a model `T` but this proliferated the number of types in the module(s) that I wasn't crazy about. Most languages don't do that and I think it "litters" the public API with so many types it makes it hard to find what devs are looking for. That's also why the guidelines state to only put types most devs will need to potentially type explicitly in the crate root e.g., clients and parameter types.

analogrelay:
> We'll want to be careful here with deriving or implementing Debug in general

Yep, I'm on the same page. This PR only implements `Debug` for `Response`s where the content is also `Debug`, so it remains up to the service clients to select which models to impl `Debug` on, and how to do it (for example, _most_ Cosmos-related models should be safe to render with `Debug`). I was hoping to be able to hack the type system enough to have an implementation for `!Debug` that rendered a placeholder for the body, but alas [impl specialization](https://rust-lang.github.io/rfcs/1210-impl-specialization.html) remains in limbo. I'd be OK with going back to a `Debug` impl that doesn't render the body at all, if there are concerns here, but I think it's reasonable to leave it up to the service clients.

> It's there currently as [azure_core::Pageable](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/core/src/pageable.rs). This will be getting refactored soon. For the most part, only the name changes.

Yeah, I noticed it but wasn't sure where that is in the refactor stages. In looking deeper I agree it looks like it's mostly got what we need, so we should be able to move forward.

> The main goal with some of the changes I already made to Response<T> and related types was deserializing late (which it already did, to be fair) but also retaining the model type.

I do like this approach if possible, and it fits with Rust's zero-cost model, since it allows you to omit parsing the body if you don't care about it. For example, when creating a database you may not care about the response if you get a 200 back. The main concern I had was around the `.json` and `.xml` methods which required the caller to know how the type was serialized. I could pivot a bit and retain the lazy-parsing, but have a single `async fn parse(self) -> T` that is serialization-format agnostic. I'm fairly confident I can finagle things so that the service client can specify the serialization format when constructing the type, rather than actually deserializing it inline. Let me play with that a bit today and see what I get.

> The unsupported "track 1" code returned bespoke response types from which you could get a model T but this proliferated the number of types in the module(s) that I wasn't crazy about.

💯 agreed. I'd like to keep the bespoke types to a minimum and focus them primarily on the actual data models.

analogrelay:
> I could pivot a bit and retain the lazy-parsing, but have a single async fn parse(self) -> T that is serialization-format agnostic.

I poked around with this option as well and it's doable, but probably requires that the `Response<T>` carry a pointer to a 'deserialization' function, or use a marker trait of some kind (like `FromJson` or` FromXml`) identify _what_ kind of deserialization logic to apply.

Saw you sent out a meeting request to chat, I think that's a great plan to get on the same page here.

heaths:
After thinking about this more, I remember that the intent is that the caller doesn't have to call any JSON or XML (or any other format) functions themselves, but the method body does e.g., https://github.com/heaths/azure-sdk-for-rust-proto/pull/8/files#diff-7c797d6db5dbc36b7d1fedf1d10e15e63dc29010082b6af7e36517454b6c494bR17.

The point of returning a `Response<T>` where `T` is a model or unit type `()` if no model is to 1) provide a "hint" to the caller as to the intended model response (or request model in the case of `RequestContent<T>`), and 2) allow the generated or hand-written body to have to do less, delegating more to the `azure_core` implementation.

Before the guidelines were ratified, an early form of this I prototyped is https://gist.github.com/heaths/39fcc4ad40e53c2db55b97bc7398ff8d. Ideally I'd like to elide the second `await` but not sure if that's possible since the initial `await` is to send the request and wait for the first response, and the subsequent `await` is to wait for all content to come through the pipeline.

heaths:
Converting to draft since we're focusing on #1724. In fact, you can close this one if you're sure this isn't the direction you want to go, @analogrelay. We can always resurrect it, of course.

heaths:
@analogrelay we can close this now, right?

analogrelay:
Yep!

PR #1714, opened 2024-07-25
Add lifetime annotation to ClientMethodOptionsBuilder trait

jhendrixMSFT:
This allows implementors of the trait to ensure that the lifetime of context lives at least as long as the trait's implementor.

This lets me write the following.
```rust
impl<'a> ClientMethodOptionsBuilder<'a> for ExtensibleStringGetKnownValueOptionsBuilder<'a> {
    fn with_context<P>(mut self, context: &'a Context) -> Self {
        self.options.method_options.set_context(context);
        self
    }
}
```
Without this change, the compiler complains that `context` might not live long enough.

I tried many permutations until I arrived at this solution. Please let me know if there's a better way and I can update as required.

### DIFF
--- a/sdk/core/azure_core/src/options/builders.rs
+++ b/sdk/core/azure_core/src/options/builders.rs
@@ -57,9 +57,9 @@ pub trait ClientOptionsBuilder {
 }
 
 /// Methods to set general method options for client-specific [`ClientMethodOptions`].
-pub trait ClientMethodOptionsBuilder {
+pub trait ClientMethodOptionsBuilder<'a> {
     /// Set optional [`Context`] for each client method call.
-    fn with_context<P>(mut self, context: &Context<'_>) -> Self
+    fn with_context<P>(mut self, context: &'a Context) -> Self
     where
         Self: Sized,
     {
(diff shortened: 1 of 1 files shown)


sdk/core/azure_core/src/options/builders.rs
```diff
@@ -57,9 +57,9 @@ pub trait ClientOptionsBuilder {
 }
 
 /// Methods to set general method options for client-specific [`ClientMethodOptions`].
-pub trait ClientMethodOptionsBuilder {
+pub trait ClientMethodOptionsBuilder<'a> {
     /// Set optional [`Context`] for each client method call.
-    fn with_context<P>(mut self, context: &Context<'_>) -> Self
+    fn with_context<P>(mut self, context: &'a Context) -> Self
```
heaths, line 62: This isn't actually accurate, though. `Context` needs to outlive the options that get built; though, that's a clone and this applies to the internal `options: ClientMethodOptions`. And, really, that only needs to outlive the parent `Context`.

Maybe this is fine for now and I should open a tracking issue to consider if this is even worth it. If we end up using the `Context` for tracing, it's likely every derivative `Context` will copy on write anyway so the `Cow` necessitating the lifetime argument would be pointless.

heaths, line 62:
```suggestion
    fn with_context(mut self, context: &'a Context) -> Self
```

While you're in there, this was probably just old copy pasta.

heaths (approved):

RickWinter (approved):

sdk/core/azure_core/src/options/builders.rs
jhendrixMSFT: I was wondering about that, removed.

PR #1715, opened 2024-07-26
First cut at a Rust Eventhubs client.

LarryOsterman:


LarryOsterman:
I would appreciate any feedback on this PR - there are a couple of places I moved outside the amqp/eventhubs codebase, mostly to fix some wasm32 build issues related to tokio.

.vscode/settings.json
```diff
@@ -1,4 +1,7 @@
 {
     "editor.formatOnSave": true,
-    "rust-analyzer.cargo.features": "all"
-}
+    "rust-analyzer.cargo.features": "all",
+    "vscode-nmake-tools.workspaceBuildDirectories": [
+        "."
+    ]
+}
```
RickWinter, line 7:
```suggestion
}

```

sdk/eventhubs/azure_messaging_eventhubs/tests/round_trip.rs
```diff
@@ -0,0 +1,169 @@
...
+    const EVENTHUB_PARTITION: &str = "1";
+    common::setup();
+    let host = env::var("EVENTHUBS_HOST").unwrap();
+    let eventhub = env::var("EVENTHUB_NAME").unwrap();
+    let producer = ProducerClient::new(
+        host.clone(),
+        eventhub.clone(),
+        DefaultAzureCredential::create(TokenCredentialOptions::default()).unwrap(),
+        ProducerClientOptions::builder()
+            .with_application_id("test_new")
```
RickWinter, line 34: Does this value need to be unique for each test to avoid resource collisions?

sdk/eventhubs/azure_messaging_eventhubs/tests/round_trip.rs
```diff
@@ -0,0 +1,169 @@
...
+    assert!(producer.submit_batch(&batch).await.is_ok());
+
+    let consumer = ConsumerClient::new(
+        host,
+        eventhub,
+        None,
+        DefaultAzureCredential::create(TokenCredentialOptions::default()).unwrap(),
+        Some(
+            ConsumerClientOptions::builder()
+                .with_application_id("test_new")
```
RickWinter, line 117: Share the application Id in a variable so that both producer and consumer change the value in one place

RickWinter (approved):

sdk/eventhubs/azure_messaging_eventhubs/tests/round_trip.rs
LarryOsterman: It's used to identify the connection to the service, which shows up in diagnostics, it does not need to be unique.

sdk/eventhubs/azure_messaging_eventhubs/tests/round_trip.rs
LarryOsterman: Makes sense.

.vscode/settings.json
LarryOsterman: That change shouldn't have snuck in, removing.

sdk/core/azure_core/Cargo.toml
```diff
@@ -49,7 +49,7 @@ rustc_version = { workspace = true }
 
 [dev-dependencies]
 tracing-subscriber = { workspace = true }
-tokio = { workspace = true }
+tokio = { workspace = true, features = ["rt-multi-thread"] }
```
heaths, line 52: We don't want to do this in every crate. Instead, leave it as it was but in your crates do:

```toml
tokio = { workspace = true, default-features = false, features = ["whatever"] }
```

you may have to add `default-features = false` to the root `Cargo.toml`, but I already did to some other ones and list out what features I want by default. That's fine.

sdk/core/azure_core_amqp/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+# AMQP Stack for consumption by packages in the Azure SDK.
+[package]
+name = "azure_core_amqp"
+version = "0.19.0"
+description = "Rust client Azure SDK"
+readme = "README.md"
+authors = ["Microsoft Corp."]
+license = "MIT"
+repository = "https://github.com/azure/azure-sdk-for-rust"
+homepage = "https://github.com/azure/azure-sdk-for-rust"
```
heaths, line 12: Nit: we should send them to this crates' subdir.

Some other info in here will be changed, but not necessary now.

sdk/core/azure_core_amqp/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+documentation = "https://docs.rs/azure_core_amqp"
+
+keywords = ["sdk", "azure", "rest", "iot", "cloud"]
+categories = ["api-bindings"]
+
+edition = "2021"
+
+[dependencies]
+async-std = { workspace = true }
+azure_core = { path = "../../core/azure_core" }
```
heaths, line 22:
```suggestion
azure_core = { workspace = true }
```

sdk/core/azure_core_amqp/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+time = { workspace = true }
+tracing = { workspace = true }
+url = { workspace = true }
+uuid = { workspace = true }
+fe2o3-amqp = { version = "0.11", features = [
+  "native-tls",
+  "tracing",
+  "uuid",
+], optional = true }
```
heaths, line 31: All dependencies need to be centrally managed in the root `Cargo.toml`'s `[workspace.dependencies]` section. This is or will be validated.

sdk/core/azure_core_amqp/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+serde_bytes = { version = "0.11", optional = true }
+
+[build-dependencies]
+rustc_version = "0.4"
+
+[dev-dependencies]
+tracing-subscriber = { workspace = true, features = ["env-filter"] }
+
+[features]
+default = ["enable-fe2o3-amqp"]
```
heaths, line 46: FWIW, I'm doing to remove "enable-" from the features for core. I don't see it anywhere else, and it's enough just to mention the project name of the default e.g., `reqwest` or `fe2o3`. This is what I've seen other projects do as well. Many crates, for example, mention `serde` by name as a feature. And just that: `serde`.

sdk/core/azure_core_amqp/build.rs
```diff
@@ -0,0 +1,12 @@
...
+
+// cspell: words rustc AZSDK
+use rustc_version::version;
+
+fn main() {
+    let version = match version() {
+        Ok(version) => version.to_string(),
+        Err(_) => "unknown".to_owned(),
+    };
+    println!("cargo:rustc-env=AZSDK_RUSTC_VERSION={version}");
```
heaths, line 11: Is this needed? We specify an MSRV in `azure_core` already (will probably move to `typespec`). If anything, we should have a general build function every crate can/should use with just a single call. I'd open a bug if you think it important to keep this information, but with an MSRV I really don't think it is.

sdk/core/azure_core_amqp/src/lib.rs
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) Microsoft Corp. All Rights Reserved.
```
heaths, line 1: Nit: full header should go here. See files under `sdk/typespec` in the `feature/track2` branch now. Eventually we'll enforce this with linting like some other languages do. Be sure to install the `EditorConfig` extension in VSCode. We do have a `.editorconfig` file that should help with this (not an error currently, though).

Cargo.toml
```diff
@@ -28,16 +30,23 @@ default-features = false
...
+
 [workspace.dependencies.azure_identity]
 path = "sdk/identity/azure_identity"
 
 [workspace.dependencies.azure_storage_common]
 version = "0.1.0"
 path = "sdk/storage"
 
+[workspace.dependencies.azure_messaging_eventhubs]
+path = "sdk/eventhubs/azure_messaging_eventhubs"
```
heaths, line 44: Only add crates that will be dependencies here e.g., `azure_core_amqp`.

sdk/core/azure_core_amqp/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+pub mod session;
+pub mod value;
+
+use std::fmt::Debug;
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum SenderSettleMode {
+    Unsettled = 0,
+    Settled = 1,
+    Mixed = 2,
```
heaths, line 25: Do these need values? This isn't normal. Really only useful if you're marshaling to/from another ABI e.g., C, or have to de/serialize these values. And in the former case, you should mark it as `#[repr(transparent)]` or `#[repr(C)]` - there's a subtle different I forget off hand but you can look it up if needed.

sdk/core/azure_core_amqp/src/error.rs
```diff
@@ -0,0 +1,64 @@
...
+// cspell: words amqp
+
+#[cfg(all(feature = "enable-fe2o3-amqp", not(target_arch = "wasm32")))]
+use crate::fe2o3::error::Fe2o3AmqpError;
+
+pub enum ErrorKind {
+    #[cfg(all(feature = "enable-fe2o3-amqp", not(target_arch = "wasm32")))]
+    IronOxideError { source: Fe2o3AmqpError },
+    #[cfg(any(not(feature = "enable-fe2o3-amqp"), target_arch = "wasm32"))]
+    NoopError,
```
heaths, line 11: I think this is unwieldy to use and maintain. The underlying implementation shouldn't be exposed like this, especially since it's varied on features. Be more generic. Can you uplift info from the fe2o3 crate if it's even useful? You can move or even ref data from underlying errors into a more generic name. The source itself is often a `Box<std::error::Error>` so anything implementing `Error` can do there.

sdk/core/azure_core_amqp/src/error.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+pub struct AmqpError {
```
heaths, line 14: It's not common - though not necessarily a bad thing, @JeffreyRichter? - to use anything other than `Error` since it's already namespaced to your crate. Users might use `azure_core_amqp::Error` or `use azure_core_amqp::Error as AmqpError` if they really need to differentiate.

Also, can we not just use what is in `typespec::error::Error` for this? Your factory functions below can be just that: functions. I did the same for HTTP errors so I could keep HTTP support conditional.

sdk/core/azure_core_amqp/src/error.rs
```diff
@@ -0,0 +1,64 @@
...
+    #[cfg(all(feature = "enable-fe2o3-amqp", not(target_arch = "wasm32")))]
+    pub fn new_iron_oxide_error(source: Fe2o3AmqpError) -> Self {
+        Self {
+            kind: ErrorKind::IronOxideError { source },
+        }
+    }
+}
+impl std::error::Error for AmqpError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        None
```
heaths, line 28: I assume this is just temporary since you actually stored the underlying error as `source`?

sdk/core/azure_core_amqp/src/connection.rs
```diff
@@ -0,0 +1,315 @@
...
+    pub(crate) buffer_size: Option<usize>,
+}
+
+impl AmqpConnectionOptions {
+    pub fn builder() -> builders::AmqpConnectionOptionsBuilder {
+        builders::AmqpConnectionOptionsBuilder::new()
+    }
+}
+
+pub trait AmqpConnectionTrait {
```
heaths, line 28: Not common to have "Trait" in a trait. It's obvious what it is.

sdk/core/azure_core_amqp/src/connection.rs
```diff
@@ -0,0 +1,315 @@
...
+    #[allow(unused_variables)]
+    fn open(
+        &self,
+        name: impl Into<String>,
+        url: Url,
+        options: Option<AmqpConnectionOptions>,
+    ) -> impl std::future::Future<Output = Result<()>> {
+        async {
+            unimplemented!();
```
heaths, line 37: Was this necessary? You typically just declare the function without a body if it must be implemented. You only have to do this when returning an unsized thing like `Self` since, in a trait, `Self` doesn't implemented `Sized`. rust-analyzer and VSCode will easily add all "abstract" functions then, too. It won't this way.

sdk/core/azure_core_amqp/src/connection.rs
```diff
@@ -0,0 +1,315 @@
...
+{
+    pub(crate) fn new(connection: T) -> Self {
+        Self(connection)
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct AmqpConnection(pub(crate) AmqpConnectionImpl<ConnectionImplementation>);
+
+impl AmqpConnectionTrait for AmqpConnection {
```
heaths, line 68: I don't understand the point of this trait. Could you explain? There's a lot of traits here, some of which I hope aren't necessary. .NET likes to use lots of interfaces. Rust does not (well, traits, but similar).

sdk/core/azure_core_amqp/src/connection.rs
```diff
@@ -0,0 +1,315 @@
...
+pub mod builders {
+    use super::*;
+    pub struct AmqpConnectionOptionsBuilder {
+        options: AmqpConnectionOptions,
+    }
+
+    impl AmqpConnectionOptionsBuilder {
+        pub(super) fn new() -> Self {
+            Self {
+                options: AmqpConnectionOptions {
```
heaths, line 99: Tip: derive `Default` on `AmqpConnectionOptions` - since everything is an `Option<T>` it'll work - and just do: `options: Default::default(),` here.

heaths: It'll be more generally useful anyway if it impls `Default`.

sdk/core/azure_core_amqp/src/connection.rs
```diff
@@ -0,0 +1,315 @@
...
+            self
+        }
+        #[allow(dead_code)]
+        pub fn with_desired_capabilities(mut self, desired_capabilities: Vec<AmqpSymbol>) -> Self {
+            self.options.desired_capabilities = Some(desired_capabilities);
+            self
+        }
+        pub fn with_properties(
+            mut self,
+            properties: Vec<(impl Into<AmqpSymbol>, impl Into<AmqpValue>)>,
```
heaths, line 152: Is a tuple an expected form that the user will pass? Just wondering if all this type coercion is actually necessary, especially if it's a corner case. We can always add helpers eventually but should probably keep it lower level for now.

sdk/core/azure_core_amqp/src/cbs.rs
```diff
@@ -0,0 +1,94 @@
...
+    ///
+    /// # Returns
+    ///
+    /// A `Result` which is:
+    /// - `Ok(())` on successful attachment of the CBS node.
+    /// - `Err(e)` where `e` is an error from the `azure_core::error::Result` indicating the failure reason.
+    ///
+    fn attach(&self) -> impl std::future::Future<Output = Result<()>> {
+        async { unimplemented!() }
+    }
```
heaths, line 29: Async traits are supported now, and our MSRV of 1.76 asserts that:
```suggestion
    async fn attach(&self) -> Result<()>;
```

sdk/core/azure_core_amqp/src/cbs.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+#[derive(Debug)]
+struct AmqpClaimsBasedSecurityImpl<T>(T);
+
+impl<T> AmqpClaimsBasedSecurityImpl<T>
+where
+    T: AmqpClaimsBasedSecurityTrait,
+{
+    pub(crate) fn new(cbs: T) -> Self {
```
heaths, line 65: Tip: unless you end up exposing a type - or are worried it might be - outside of `src/lib.rs`, `pub(crate)` is overkill. Ultimately, you could make everything `pub` but if it's not `pub mod` or `pub use` in `src/lib.rs` it won't matter. This doesn't hurt, hence the "tip" prefix.

sdk/core/azure_core_amqp/src/value.rs
```diff
@@ -0,0 +1,735 @@
...
+                        AmqpValue::$field(v) => self == v,
+                        _ => false,
+                    }
+                }
+            }
+        )*
+    }
+}
+
+conversions_for_amqp_types!(
```
heaths, line 225: Using function macros for ease: nice! But I'd consider something more like a tuple for input just to reinforce that there's a pairing here. This feels fragile. And, most likely, having the macro take a tuple argument matcher would actually make the impl easier as well (I haven't looked at it yet, mind you).

heaths: For an example, see `create_enum` in `macros.rs` under `azure_core` in `main`.

heaths (commented):
I went through a lot of the code and hopefully gave enough feedback for now to chew on that should permeate throughout the code. Not all is blocking, and - as usual - I've tried to preface such things with "tip" or "nit". It's late on a Friday so we can talk some more Monday morning. Overall, it's looking pretty good. Mainly just some questions about (unnecessary?) complexity and a few non-idiomatic things. I wouldn't necessarily block on those if some issue(s) were filed to keep track of changes to make later.

sdk/core/azure_core_amqp/Cargo.toml
LarryOsterman: I was wondering about that - if we should include optional dependencies for just one package in the global set of packages in cargo.toml.

sdk/core/azure_core_amqp/src/connection.rs
LarryOsterman: This was a suggestion from the rustacians teams channel - they suggested providing a default implementation on these trait methods so that they don't need to be implemented in every new trait implementation.

I'm not wedded to this pattern at all.

sdk/core/azure_core_amqp/Cargo.toml
LarryOsterman: Annoyingly, the feature name cannot match a crate name - cargo rejects it :(.

sdk/core/azure_core_amqp/build.rs
LarryOsterman: No - it came in as a copy from eventhubs.

And it is needed in eventhubs to capture the compiler version for the user-agent string (eventhubs has to generate the user-agent privately).

sdk/core/azure_core_amqp/src/lib.rs
LarryOsterman: Those values are defined by the [AMQP protocol](https://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-sender-settle-mode). They are initialized to ensure that the internal representation matches the protocol.

I'm not comfortable using `#[repr(transparent)]` or `#[repr(C)]` because these are protocol elements - I would be initializing them in C/C++ as well.

LarryOsterman: I've defined manifest constants for these values and then included a link to the protocol documentation which defines them.

Btw, I'm going to end up moving them out of the main crate, they really don't belong there.

sdk/core/azure_core_amqp/src/error.rs
LarryOsterman: I'll move it to Error.

sdk/core/azure_core_amqp/src/error.rs
LarryOsterman: That was ... unpleasant, but I think I am happy with the results.

sdk/core/azure_core_amqp/src/connection.rs
LarryOsterman: Oooh, nice. Love that one.

sdk/core/azure_core_amqp/src/error.rs
LarryOsterman: No, it was a cut&paste thingy I picked up from somewhere, not even sure where (could have been copilot halucinating, tbh).

Its fixed now.

LarryOsterman: Fixed in both AMQP and Fe2o3.

sdk/core/azure_core_amqp/src/connection.rs
LarryOsterman: Yeah, I know - the problem here is that I have three related types:
1) The type `AmqpConnection`
1) The implementation of the type `AmqpConnectionImpl`
1) The trait which defines the API surface for the type `AmqpConnectionTrait`.

I have been struggling with what to call these three - I want the concrete type to be named `AmqpConnection`. I don't care about #2 and #3, but they each need their own name. And thus my dilemma.

sdk/core/azure_core_amqp/src/cbs.rs
LarryOsterman: Async traits do not appear to be supported:

```
  --> sdk\core\azure_core_amqp\src\connection.rs:32:5
   |
32 |     async fn open(
   |     ^^^^^
   |
   = note: you can suppress this lint if you plan to use the trait only in your own code, or do not care about auto traits like `Send` on the `Future`
   = note: `#[warn(async_fn_in_trait)]` on by default
help: you can alternatively desugar to a normal `fn` that returns `impl Future` and add any desired bounds such as `Send`, but these cannot be relaxed without a breaking API change
   |
32 ~     fn open(
33 |         &self,
 ...
36 |         options: Option<AmqpConnectionOptions>,
37 ~     ) -> impl std::future::Future<Output = Result<()>> + Send {async {
38 |         unimplemented!();
39 ~     } }
   |
```

sdk/core/azure_core_amqp/Cargo.toml
heaths: Yeah, just to keep everything in one place. .NET et. al. do it as well.

sdk/core/azure_core_amqp/Cargo.toml
heaths: Use "dep:{crate}" to work around that.

sdk/core/azure_core_amqp/src/lib.rs
heaths: See <https://docs.opensource.microsoft.com/releasing/general-guidance/copyright-headers/>.

/cc @RickWinter @ronniegeraghty as an FYI.

sdk/core/azure_core_amqp/src/connection.rs
heaths: Is a trait even necessary and does it need to be exposed publicly?

We can run with it for now, though. Curious if @demoray, @eholk, @yoshuawuyts, or others know of some precedent or have ideas here. Not blocking, though.

sdk/core/azure_core_amqp/src/connection.rs
heaths: Oh, so these are *required* implementations? Than why err at all? Just to reinforce develop intent?

sdk/core/azure_core_amqp/src/cbs.rs
heaths: Ah, yes: <https://blog.rust-lang.org/2023/12/28/Rust-1.75.0.html#async-fn-and-return-position-impl-trait-in-traits>

Perhaps we should do its second suggestion for now:

> you can alternatively desugar to a normal `fn` that returns `impl Future` and add any desired bounds such as `Send`, but these cannot be relaxed without a breaking API change

Curious what @yoshuawuyts thinks. We do declare our `rust-version` as 1.76 currently, so I don't think this is a real issue. A bit surprised the lint isn't taking that into consideration, though maybe it only does that for the current crate and not scan the entire dependency tree (`azure_core` declares the MSRV currently, though I'll move that to `typespec`).

sdk/core/azure_core_amqp/src/connection.rs
LarryOsterman: I *think* the trait is necessary, and it unfortunately does have to be public - I've tried having it be private and I get a bunch of compiler errors indicating that the compiler can't generate correct code without visibility into the connection :(.

sdk/core/azure_core_amqp/src/connection.rs
LarryOsterman: There's basically 1 trait per construct, and 3 implementations of the trait. One implementation is a thin forwarder to the actual implementation (that's the one you see in this file). The two implementations are in noop.rs (that's the "NoOp implementation which fails everything"), and the one in the fe2o3 module (which is the "real" iron oxide client).

It's a trait because I need all 3 implementations to have the same API surface. And I don't want to expose any of the feature complexity outside the AMQP package - that's why I have two newtypes - the concrete one in AmqpConnection and the generic one in AmqpConnectionImpl which is specialized in AmqpConnection.

sdk/core/azure_core_amqp/src/cbs.rs
LarryOsterman: Ah - declare the "new" function as "pub" because the containing struct is pub(crate) and thus pub(crate) is irrelevant on the new function.

I was worried about a "function `new` has greater visibility than containing struct" error, but that apparently doesn't happen.

sdk/core/azure_core_amqp/src/connection.rs
LarryOsterman: Actually, this one should be "impl Into<AmqpOrderedMap<AmqpSymbol, AmqpValue>>" which is more accurate.

sdk/core/azure_core_amqp/src/connection.rs
LarryOsterman: Ok, now a trait and 3 implementations of the trait, removing both of the newtypes.

sdk/core/azure_core_amqp/src/cbs.rs
LarryOsterman: I'm pretty sure that the "desugar to a normal fn that returns impl Future" is what I'm currently doing.

If it's not what I'm doing, please let me know how I misinterpreted the suggestion.

PR #1721, opened 2024-07-29
[Rust] [Storage] Initial Storage Scaffolding (azure_storage_blob)

vincenttran-msft:
This PR introduces the scaffolding for upcoming Rust SDK storage work. The structure is as follows:

- sdk/storage _# This will contain common code like boilerplate authentication stuff and any other "base client" type of processing like URL processing_
- sdk/storage/azure_storage_blob _# This is where Blob storage will live_
- sdk/storage/azure_storage_blob/src/clients _# This is where all of the Blob client's actual user-spun up client code will live
Which scaffolds into:_
  - /blob_client.rs
  - /blob_service_client.rs
  - /blob_container_client.rs

In broad strokes, are we happy with the naming? Is there any conventions we are not following here or patterns we want to establish? i.e. in track1 it is "storage_blob**s**"


sdk/storage/azure_storage_blob/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+version = "0.1.0"
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
```
heaths, line 12: Not necessary, but I'd add the following as part of your scaffolding since you know you'll need it, and it makes a good example:

```suggestion
[dependencies]
azure_storage_common = { workspace = true }
```

heaths (approved):

sdk/storage/azure_storage_blob/Cargo.toml
vincenttran-msft: Good callout, took this feedback! Thanks for the review Heath I appreciate the help 😄

heaths (approved):

vincenttran-msft:
@ronniegeraghty Hi Ronnie, sorry for the random ping, I pushed some changes up regarding the feedback left by Heath, but this is good to go!

PR #1724, opened 2024-07-30
Take 2: Refactoring Response<T> to collapse json/xml methods

analogrelay:
Ok, here's another approach for the `Response<T>` refactor that's primarily motivated by the fact that our latest attempt was stymied by the fact that Rust does not permit default values on type parameters for functions.

As a reminder, here are our goals:

* Allow the user to specify their own model type to deserialize in to
* Allow the user to to read status code and headers without requiring they wait for the body
* Retain a smooth and discoverable experience for the "default" use case of deserializing into the service-provided model.
* Allow us to model large streamed responses, like downloading a blob (being forced to load the entire response body at once is undesirable). 

So, given that I've made some refactorings:

1. Removed `RawResponse`, there is only `Response<T>`. A `RawResponse` is represented by `Response<()>`
2. As in the current `feature/track2` branch, the `Response<T>` does not _contain_ a `T`, it indicates that a `T` can be deserialized from the body of the response. A `Response<T>` contains the raw `PinnedStream` representing a stream of incoming bytes.
3. Since the `T` in `Response<T>` just indicates how the body can be read, it has a default of `()` (so `Response<()>` is what pipeline elements return since they don't know the intended return type).
4. The target type `T` can be changed using `map_body`: `let resp: Response<MyOtherType> = resp.map_body()`. This doesn't parse anything, it just changes the target type.
5. The body type `T` can be retrieved from a `Response<T>` using `read_body`. This is an `async` function that relies on `T` implementing the `FromResponseBody` trait.
6. We use `FromResponseBody` instead of `TryFrom` or `From` because we need an async version. I don't want to claim a name like `AsyncTryFrom` so it felt best to just limit the trait to our specific need.
7. A service client has flexibility in how it implements `FromResponseBody` because it doesn't have to wait for the entire body to be downloaded.
8. To simplify writing client models, there are two macros: `json_serializable!` and `xml_serializable!` which generate an automatic implementation of `FromResponseBody` for any type that implements `Deserialize`, using our `from_json` and `read_xml` helpers.

Here's an example client method, taken from my coming PR to implement the "read database metadata" API for Cosmos DB:

```rust
async fn read(
    &self,
    options: Option<ReadDatabaseOptions>,
) -> azure_core::Result<azure_core::Response<DatabaseProperties>> {
    let mut req = Request::new(self.base_url.clone(), azure_core::Method::Get);
    let ctx = Context::new().with_value(ResourceType::Databases);
    self.root_client.pipeline.send(&ctx, &mut req).await
}
```

The `Pipeline::send` method is generic, which allows for this fairly clean code that infers the `T` in `Response<T>` from the return type of the `read` function.

A user application would call this API like so:

```rust
let response = db_client.read(None).await?.read_body().await?;
```

The double-`await` is here but somewhat unavoidable without even more type machinations. The first one is for sending the request and receiving the headers. The second one is for loading the entire body. However, you could see some benefit from this in say, a fictious API like this:

```rust
blob_client.get(None).await?.save_to_file("/foo.txt").await?;
```

The `save_to_file` method has the ability to stream bytes directly from the `ResponseBody` to disk and doesn't require that you load it all into memory first.
 
A user can _also_ choose to use their own type, like so:

```rust
#[derive(Deserialize, Debug)]
struct MyDatabase {
    #[serde(rename = "id")]
    pub yon_id: String,
}
azure_core::json_serializable!(MyDatabase);
let response: MyDatabase = db_client.read(None).await?.map_body().read_body().await?;
```

Note the additional `map_body` call. It would be nice if `map_body` wasn't required, but we end up hitting exactly the same problem as we had before. Making the `read_body` method generic would require that the user **always** specify the target type, even when using the default (since it can't have a default type).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,15 +1,15 @@
 [workspace]
 resolver = "2"
 members = [
-  "sdk/typespec",
-  "sdk/typespec/typespec_client_core",
-  "sdk/core/azure_core",
-  "sdk/core/azure_core_amqp",
-  "sdk/identity/azure_identity",
-  "sdk/eventhubs/azure_messaging_eventhubs",
-  "eng/test/mock_transport",
-  "sdk/storage",
-  "sdk/storage/azure_storage_blob",
+    "sdk/typespec",
+    "sdk/typespec/typespec_client_core",
+    "sdk/core/azure_core",
+    "sdk/core/azure_core_amqp",
+    "sdk/identity/azure_identity",
+    "sdk/eventhubs/azure_messaging_eventhubs",
+    "eng/test/mock_transport",
+    "sdk/storage",
+    "sdk/storage/azure_storage_blob",
 ]
 
 [workspace.package]
@@ -67,8 +67,8 @@ pin-project = "1.0"
 quick-xml = { version = "0.31", features = ["serialize", "serde-types"] }
 rand = "0.8"
 reqwest = { version = "0.12", features = [
-  "json",
-  "stream",
+    "json",
+    "stream",
 ], default-features = false }
 rustc_version = "0.4"
 serde = { version = "1.0", features = ["derive"] }
@@ -80,15 +80,14 @@ serial_test = "3.0"
 sha2 = { version = "0.10" }
 thiserror = "1.0"
 time = { version = "0.3.10", features = [
-  "serde-well-known",
-  "macros",
-  "wasm-bindgen",
+    "serde-well-known",
+    "macros",
+    "wasm-bindgen",
 ] }
 tokio = { version = "1.0", default-features = false, features = [
-  "macros",
-  "time",
+    "macros",
+    "time",
 ] }
-tokio-test = "0.4.4"
 tracing = "0.1.40"
 tracing-subscriber = "0.3"
 tz-rs = { version = "0.6" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,15 +1,15 @@
 [workspace]
 resolver = "2"
 members = [
-    "sdk/typespec",
-    "sdk/typespec/typespec_client_core",
-    "sdk/core/azure_core",
-    "sdk/core/azure_core_amqp",
-    "sdk/identity/azure_identity",
-    "sdk/eventhubs/azure_messaging_eventhubs",
-    "eng/test/mock_transport",
-    "sdk/storage",
-    "sdk/storage/azure_storage_blob",
+  "sdk/typespec",
+  "sdk/typespec/typespec_client_core",
+  "sdk/core/azure_core",
+  "sdk/core/azure_core_amqp",
+  "sdk/identity/azure_identity",
+  "sdk/eventhubs/azure_messaging_eventhubs",
+  "eng/test/mock_transport",
+  "sdk/storage",
+  "sdk/storage/azure_storage_blob",
 ]
 
 [workspace.package]
@@ -67,8 +67,8 @@ pin-project = "1.0"
 quick-xml = { version = "0.31", features = ["serialize", "serde-types"] }
 rand = "0.8"
 reqwest = { version = "0.12", features = [
-    "json",
-    "stream",
+  "json",
+  "stream",
 ], default-features = false }
 rustc_version = "0.4"
 serde = { version = "1.0", features = ["derive"] }
@@ -80,13 +80,13 @@ serial_test = "3.0"
 sha2 = { version = "0.10" }
 thiserror = "1.0"
 time = { version = "0.3.10", features = [
-    "serde-well-known",
-    "macros",
-    "wasm-bindgen",
+  "serde-well-known",
+  "macros",
+  "wasm-bindgen",
 ] }
 tokio = { version = "1.0", default-features = false, features = [
-    "macros",
-    "time",
+  "macros",
+  "time",
 ] }
 tokio-test = "0.4.4"
 tracing = "0.1.40"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,15 +1,15 @@
 [workspace]
 resolver = "2"
 members = [
-  "sdk/typespec",
-  "sdk/typespec/typespec_client_core",
-  "sdk/core/azure_core",
-  "sdk/core/azure_core_amqp",
-  "sdk/identity/azure_identity",
-  "sdk/eventhubs/azure_messaging_eventhubs",
-  "eng/test/mock_transport",
-  "sdk/storage",
-  "sdk/storage/azure_storage_blob",
+    "sdk/typespec",
+    "sdk/typespec/typespec_client_core",
+    "sdk/core/azure_core",
+    "sdk/core/azure_core_amqp",
+    "sdk/identity/azure_identity",
+    "sdk/eventhubs/azure_messaging_eventhubs",
+    "eng/test/mock_transport",
+    "sdk/storage",
+    "sdk/storage/azure_storage_blob",
 ]
 
 [workspace.package]
@@ -67,8 +67,8 @@ pin-project = "1.0"
 quick-xml = { version = "0.31", features = ["serialize", "serde-types"] }
 rand = "0.8"
 reqwest = { version = "0.12", features = [
-  "json",
-  "stream",
+    "json",
+    "stream",
 ], default-features = false }
 rustc_version = "0.4"
 serde = { version = "1.0", features = ["derive"] }
@@ -80,14 +80,15 @@ serial_test = "3.0"
 sha2 = { version = "0.10" }
 thiserror = "1.0"
 time = { version = "0.3.10", features = [
-  "serde-well-known",
-  "macros",
-  "wasm-bindgen",
+    "serde-well-known",
+    "macros",
+    "wasm-bindgen",
 ] }
 tokio = { version = "1.0", default-features = false, features = [
-  "macros",
-  "time",
+    "macros",
+    "time",
 ] }
+tokio-test = "0.4.4"
 tracing = "0.1.40"
 tracing-subscriber = "0.3"
 tz-rs = { version = "0.6" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,15 +1,15 @@
 [workspace]
 resolver = "2"
 members = [
-    "sdk/typespec",
-    "sdk/typespec/typespec_client_core",
-    "sdk/core/azure_core",
-    "sdk/core/azure_core_amqp",
-    "sdk/identity/azure_identity",
-    "sdk/eventhubs/azure_messaging_eventhubs",
-    "eng/test/mock_transport",
-    "sdk/storage",
-    "sdk/storage/azure_storage_blob",
+  "sdk/typespec",
+  "sdk/typespec/typespec_client_core",
+  "sdk/core/azure_core",
+  "sdk/core/azure_core_amqp",
+  "sdk/identity/azure_identity",
+  "sdk/eventhubs/azure_messaging_eventhubs",
+  "eng/test/mock_transport",
+  "sdk/storage",
+  "sdk/storage/azure_storage_blob",
 ]
 
 [workspace.package]
@@ -67,8 +67,8 @@ pin-project = "1.0"
 quick-xml = { version = "0.31", features = ["serialize", "serde-types"] }
 rand = "0.8"
 reqwest = { version = "0.12", features = [
-    "json",
-    "stream",
+  "json",
+  "stream",
 ], default-features = false }
 rustc_version = "0.4"
 serde = { version = "1.0", features = ["derive"] }
@@ -80,13 +80,13 @@ serial_test = "3.0"
 sha2 = { version = "0.10" }
 thiserror = "1.0"
 time = { version = "0.3.10", features = [
-    "serde-well-known",
-    "macros",
-    "wasm-bindgen",
+  "serde-well-known",
+  "macros",
+  "wasm-bindgen",
 ] }
 tokio = { version = "1.0", default-features = false, features = [
-    "macros",
-    "time",
+  "macros",
+  "time",
 ] }
 tracing = "0.1.40"
 tracing-subscriber = "0.3"
--- a/sdk/identity/azure_identity/src/refresh_token.rs
+++ b/sdk/identity/azure_identity/src/refresh_token.rs
@@ -49,7 +49,7 @@ pub async fn exchange(
     let rsp_status = rsp.status();
 
     if rsp_status.is_success() {
-        rsp.read_body().await.map_kind(ErrorKind::Credential)
+        rsp.deserialize_body().await.map_kind(ErrorKind::Credential)
     } else {
         let rsp_body = rsp.into_body().collect().await?;
         let token_error: RefreshTokenError =
--- a/sdk/identity/azure_identity/src/refresh_token.rs
+++ b/sdk/identity/azure_identity/src/refresh_token.rs
@@ -45,11 +45,11 @@ pub async fn exchange(
     );
     req.set_body(encoded);
 
-    let rsp = http_client.execute_request(&req).await?;
+    let rsp = http_client.execute_request(&req).await?.map();
     let rsp_status = rsp.status();
 
     if rsp_status.is_success() {
-        rsp.json().await.map_kind(ErrorKind::Credential)
+        rsp.read_body().await.map_kind(ErrorKind::Credential)
     } else {
         let rsp_body = rsp.into_body().collect().await?;
         let token_error: RefreshTokenError =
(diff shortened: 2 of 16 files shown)


heaths:
Overall, I like the concept. I haven't dug into the code just yet, but wanted to make some time to at least read through your proposal.

> To simplify writing client models, there are two macros: json_serializable! and xml_serializable! which generate an automatic implementation of FromResponseBody for any type that implements Deserialize, using our from_json and read_xml helpers.

I like this but would prefer a derive macro. I plan on adding some for other reasons including client options and client method options builders, and could do this eventually. Having functional macros for now would be fine. I wouldn't want to block on this, but this is a case perfect for derive macros or an attribute-like procedural macro e.g.,

```rust
#[derive(JsonModel, Clone, Default, Debug)]
pub struct ModelA;

#[derive(Clone, Default, Debug)]
#[model(json)]
pub struct ModelB;
```

A few nits on name to be more consistent with `std`, IMO (certainly up for discussion):

1. `read_body` -> `deserialize`? "read" I see more often on methods that gets you a `&[u8]` or something like that - raw reads, basically.
2. `save_to_file` -> `write_to_file` a la `std::io::Write`.
3. Could we just have `map_body` return a `T` on success? Again, I haven't looked at the code, but calling `map_body().read_body()` feels extraneous.

Having typed that last line, I suppose `read_body` has a nice symmetry with `map_body`, but still not crazy about "read". Maybe `deserialize_body`? Everything ties around "body". I don't hate "read", but think it still feels like a "raw" operation.

analogrelay (commented):
Ah dang, I forgot to submit these comments earlier. But I left some comments in-line explaining a few things.

analogrelay:
> I like this but would prefer a derive macro.

Absolutely agreed. I was planning on this being a derive or attribute (or both) macro but didn't want to do that right now. The inline comments I wrote (but only just published because I forgot) indicate that too.

> 1. read_body -> deserialize? "read" I see more often on methods that gets you a &[u8] or something like that - raw reads, basically.
>
> Having typed that last line, I suppose read_body has a nice symmetry with map_body, but still not crazy about "read". Maybe deserialize_body? Everything ties around "body". I don't hate "read", but think it still feels like a "raw" operation.

I like `deserialize_body` a lot better, good call. I totally agree on `read_` being something I expect to yield bytes.

> 2. save_to_file -> write_to_file a la std::io::Write.

This was just a fictious API to show how you could avoid loading the response into memory, but I definitely agree this would be a `write_to_file` method.

> 3. Could we just have map_body return a T on success? Again, I haven't looked at the code, but calling map_body().read_body() feels extraneous.

The reason that doesn't work is that the intent is for `map_body` to be called _inside_ the service client method. If it unwrapped the `Response<T>` into a `T`, the user code would lose access to the headers and status code. However, this does bring up an alternate strategy, if we're OK with two deserialize methods: `deserialize_body` and `deserialize_body_into` (feel free to suggest alternate suffixes):

```rust
impl<T: DeserializeOwned> Response<T> {
    // Deserializes a `Response<T>` body into the `T` represented in its type parameter
    pub fn deserialize_body(self) -> azure_core::Result<T> { ... }
}

impl<T> Response<T> {
    // Deserializes a `Response<T>` body into a provided `U`, which may be any deserializable type.
    pub fn deserialize_body_into<U: DeserializeOwned>(self) -> azure_core::Result<U> { ... }
}
```

So, instead of `map_body::<U>().deserialize_body()` you can use `deserialize_body_into::<U>()` to deserialize into some other type `U`. The only time someone needs to convert a `Response<T>` into a `Response<U>` is when taking the `Response<()>` that the underlying HTTP pipeline produces into a `Response<SomeModelType>`. We could probably provide some kind of method that's relatively hidden (hopefully not using `#[doc(hidden)]` but that may be one option) and intended for use in SDK crates only to do that conversion.

analogrelay:
Updated to address the rename to `deserialize_body` and to replace `map_body` as much as possible. It's still needed when converting the `Response<()>` that the underlying HTTP Pipeline provides (representing an "untyped" response) into a `Response<T>` but I was able to scope it to that. We could _probably_ `#[doc(hidden)]` that method if we want, since it should only be used in the `TransportOptions::send` method, but we can't `pub(crate)` it because it's in `typespec_client_core` but used in `azure_core`

sdk/core/azure_core/src/response.rs
```diff
@@ -7,27 +7,74 @@ use crate::{
...
+    ///
+    /// A [`ResponseBody`] represents a stream of bytes coming from the server.
+    /// The server may still be sending data, so it's up to implementors whether they want to wait for the entire body to be received or not.
+    /// For example, a type representing a simple REST API response will want to wait for the entire body to be received and then parse the body.
+    /// However, a type representing the download of a large file, may not want to do that and instead prepare to stream the body to a file or other destination.
+    fn from_response_body(body: ResponseBody) -> impl Future<Output = crate::Result<Self>>;
+}
+
+#[macro_export]
+macro_rules! json_serializable {
```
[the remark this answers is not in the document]
heaths: We're on the same page there. See my comments from yesterday. 😉

sdk/core/azure_core/src/response.rs
```diff
@@ -46,189 +93,88 @@ impl RawResponse {
...
+            .field("status", &self.status)
+            .field("headers", &self.headers)
             .field("body", &"(body)")
             .finish()
     }
 }
 
-/// An HTTP response with the body collected as bytes.
-#[derive(Debug, Clone)]
-pub struct CollectedResponse<T> {
```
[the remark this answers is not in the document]
heaths: The `main` branch was using it, though off hand I can't remember why. I think it was primarily for testing and/or pageables.

analogrelay:
@heaths I'd love to see if we can get this, or something similar that refactors `Response<T>` to avoid users having to know about json/xml, merged soon so I can proceed on a few different dependent workstreams: Integrating an initial Cosmos DB crate, and preparing a derive macro to replace `json_serializable`/`xml_serializable` (something I wanted to avoid putting in this PR to keep it simpler but could add here if you'd prefer)

heaths:
We need to also make sure multipart-form-data is supported, or maybe even easy to use as JSON or XML would. There are some services that make use of it.

analogrelay:
> We need to also make sure multipart-form-data is supported

Yep, that should be doable via updating our derive macro, once that lands. I'm working on the derive macro now (which will replace `json_serializable!` and `xml_serializable!`, but would prefer to put it in a separate PR after this one lands since it's a significant chunk of logic (not a ton of code, but could use its own dedicated review)

heaths:
The derive macro I was planning on putting in a crate called `typespec_derive`, which will contain things for both `typespec` and `typespec_client_core`, and conceivably a future `typespec_server_core`. If we need one for anything Azure-specific, I was thinking of `azure_derive`.

analogrelay:
@heaths Yep, that's along the same lines that I was thinking. I've got a branch where I've got it mostly working, but need to adjust it a bit to align with your suggested crates.

Aside from resolving conflicts (which I'm doing today), do you see any blockers for getting this merged soon? I'd like to get the initial CosmosDB crate in soon, if possible.

sdk/identity/azure_identity/src/refresh_token.rs
```diff
@@ -44,11 +44,11 @@ pub async fn exchange(
     );
     req.set_body(encoded);
 
-    let rsp = http_client.execute_request(&req).await?;
+    let rsp = http_client.execute_request(&req).await?.map_body();
```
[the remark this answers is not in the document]
analogrelay: Actually, this can use `deserialize_body_into` now 🎉

sdk/core/azure_core/src/response.rs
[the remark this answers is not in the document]
analogrelay: Do you feel a need to have it back? I'd like to avoid having a separate type to represent this, if possible. If/when we need to represent a response using collected bytes (instead of a stream) we can revisit this.

sdk/core/azure_core/src/response.rs
```diff
@@ -7,27 +7,74 @@ use crate::{
...
+use std::future::Future;
 use std::{fmt, marker::PhantomData, pin::Pin};
 
 #[cfg(not(target_arch = "wasm32"))]
 pub(crate) type PinnedStream = Pin<Box<dyn Stream<Item = crate::Result<Bytes>> + Send + Sync>>;
 #[cfg(target_arch = "wasm32")]
 pub(crate) type PinnedStream = Pin<Box<dyn Stream<Item = crate::Result<Bytes>>>>;
 
+/// Trait that represents types that can be deserialized from an HTTP response body.
+pub trait FromResponseBody: Sized {
```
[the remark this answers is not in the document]
analogrelay: I'm going to rename this to Model, since that's what I used in the prototype derive macro I'm building.

sdk/typespec/typespec_client_core/src/http/response.rs
```diff
@@ -45,193 +99,125 @@ impl RawResponse {
...
-
-impl<T> From<RawResponse> for Response<T> {
-    fn from(response: RawResponse) -> Self {
-        Self {
-            response,
+impl Response<()> {
+    /// Changes the type of the response body.
+    ///
+    /// Used to set the "type" of an untyped `Response<()>`, transforming it into a `Response<T>`.
+    pub fn map_body<T>(self) -> Response<T> {
```
heaths, line 142: I don't want to block the PR on this - but there needs to be another commit for the formatting issues, so I leave it up to you whether you want to change this now...if we can agree. I think "map" feels right, but "body" doesn't given it maps a Response<()> into a `Response<T>`. What about just implementing `Into` in this case?

```rust
use std::marker::PhantomData;

#[derive(serde::Deserialize)]
struct Response<'a, T: serde::Deserialize<'a>> {
    bytes: Vec<u8>,
    phantom: PhantomData<&'a T>,
}

impl<'a, T> Into<Result<Response<'a, T>, ()>> for Response<'a, ()>
where
    T: serde::Deserialize<'a>,
{
    fn into(self) -> Result<Response<'a, T>, ()> {
        serde_json::from_slice(&self.bytes).map_err(|_| ())
    }
}
```

sdk/typespec/typespec_client_core/src/http/response.rs
```diff
@@ -45,193 +99,125 @@ impl RawResponse {
...
+    /// #      typespec_client_core::http::headers::Headers::new(),
+    /// #      "{\"name\":\"database_password\",\"value\":\"hunter2\"}",
+    /// #    )
+    /// #  }
+    /// # }
+    /// # pub fn create_secret_client() -> SecretClient {
+    /// #   SecretClient { }
+    /// # }
+    ///
+    /// # tokio_test::block_on(async {
```
heaths, line 179: Alternatively, instead of adding yet another dependency (though, it seems we already had it, but I want to reduce them as we go), what about just hiding an async main e.g.?

```rust
/// # #[tokio::main]
/// # async fn main() {
///    // clipped
/// # }
```

heaths (changes requested):
A few thoughts - those could get cleaned up in another PR - but wanted to make sure `.editorconfig` was honored so we don't get into "git formatting wars".

Cargo.toml
```diff
@@ -88,6 +88,7 @@ tokio = { version = "1.0", default-features = false, features = [
   "macros",
   "time",
 ] }
+tokio-test = "0.4.4"
```
heaths, line 91: Ah, so this was added. See my comment re: this in my previous PR feedback. I'd rather avoid more dependencies where we can.

analogrelay: Yep, agreed. I'll yank it and use an async main as you suggested.

Cargo.toml
```diff
@@ -1,15 +1,15 @@
...
-  "sdk/typespec",
-  "sdk/typespec/typespec_client_core",
-  "sdk/core/azure_core",
-  "sdk/core/azure_core_amqp",
-  "sdk/identity/azure_identity",
-  "sdk/eventhubs/azure_messaging_eventhubs",
-  "eng/test/mock_transport",
-  "sdk/storage",
-  "sdk/storage/azure_storage_blob",
+    "sdk/typespec",
```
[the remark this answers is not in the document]
analogrelay: This should be resolved now. Might be worth seeing if we can add a linter to check TOML and similar files against editor config 🤔 . I'm very much in favor of just making _any_ call on formatting and then enforcing it and ending the nearly-endless debate

sdk/typespec/typespec_client_core/src/http/response.rs
analogrelay: Unfortunately no, we can't use `Into` or `From` here. Both of `impl<T> From<Response<()> for Response<T>` **and** `impl<T> Into<Response<T>> for Response<()>` conflict with existing impls (`impl<T> From<T> for T`).

However, let me see if I can make this `pub(crate)`. I think I can now, and doing that means the weirdness is limited to `typespect_client_core`. A service client is going to call `Pipeline::send<T>` which already returns a `Response<T>` (because it calls `map_body` within it). For the rare service client that needs to take the "raw" `Response<()>` returned by a policy or HTTP client and deserialize it into a custom model (i.e. Azure Identity), they can use `deserialize_body_into`, which exists on `Response<()>`.

sdk/typespec/typespec_client_core/src/http/pipeline.rs
```diff
@@ -136,7 +138,7 @@ mod tests {
             .send(&Context::default(), &mut request)
             .await
             .unwrap()
-            .json()
+            .deserialize_body()
```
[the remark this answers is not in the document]
analogrelay: We chatted offline, but just to include context here, `map_body` isn't something I expect users to call, and now it's `pub(crate)` (see below). We have just `deserialize_body` and `deserialize_body_into`, which I think are good-enough names for now.

analogrelay:
Pushed another iteration, I believe I've addressed feedback now!

Cargo.toml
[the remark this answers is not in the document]
heaths: Interestingly, `dotnet format` [purports](https://learn.microsoft.com/community/content/how-to-enforce-dotnet-format-using-editorconfig-github-actions) to enforce this. Seems `rustfmt` and friends don't, and there's not any seemingly high-quality tools out there that do. I only took a quick look, though.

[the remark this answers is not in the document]
heaths: Opened #1768 to track.

sdk/typespec/typespec_client_core/src/http/response.rs
heaths: This works for now, though the `set_` prefix seems odd for something that isn't a setter. I'll let you decide whether to act now, later, or not at all (given we'll probably rename a bunch of this later anyway, depending on how #1766 goes) before merging.

Cargo.toml
```diff
@@ -67,8 +67,8 @@ pin-project = "1.0"
 quick-xml = { version = "0.31", features = ["serialize", "serde-types"] }
 rand = "0.8"
 reqwest = { version = "0.12", features = [
-  "json",
-  "stream",
+    "json",
```
heaths, line 70: These should be gone - just 2 spaces now, accordingly to `.editorconfig`.

heaths (approved):

sdk/typespec/typespec_client_core/src/http/response.rs
analogrelay: That's true. How about `with_`?

heaths: Sounds better, sure.

Cargo.toml
analogrelay: 🤦🏻‍♀️

analogrelay: Ok, I believe _now_ I've got RustRover configured right for EditorConfig... *sigh*

sdk/typespec/typespec_client_core/src/http/response.rs
```diff
@@ -13,20 +14,73 @@ pub type PinnedStream = Pin<Box<dyn Stream<Item = crate::Result<Bytes>> + Send +
 #[cfg(target_arch = "wasm32")]
 pub type PinnedStream = Pin<Box<dyn Stream<Item = crate::Result<Bytes>>>>;
 
-/// An HTTP response containing a stream of bytes.
-pub struct RawResponse {
+/// Trait that represents types that can be deserialized from an HTTP response body.
+pub trait Model: Sized {
```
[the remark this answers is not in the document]
analogrelay: I don't think this has to restrict our ability to support "allocation-free" deserialization, presuming what you mean is not 100% allocation free but rather not having to allocate the memory for the entire body. This constraint forces the `Model` type itself to be Sized, but the `LazyBody` test below illustrates a model that just takes the existing body (a Boxed Pinned pointer to an async stream of Bytes) and wraps it up in a "smart-pointer" of sorts. It's not allocation-free, but the `LazyBody` type is essentially the same size as the already-allocated `ResponseBody` that gets moved into it.

In general though, I'm totally on board with enabling allocation-free and low-allocation options. I looked in to supporting `Deserialize` rather than `DeserializeOwned` and it's probably doable, but would need some further investigation to get it working.

PR #1726, opened 2024-07-31
Add `BearerTokenCredentialPolicy` to `azure_core`

vincenttran-msft:
This is the first-pass rough implementation that we were using in our [playground](https://github.com/wastore/azure-sdk-for-rust-pr/blob/feature/track2/sdk/core/src/policies/storage_bearer_token.rs)

Fully functioning (in simple cases, does not have some intelligent caching / refreshing when the auth expires), but I do have some questions:

1. Are the lifetime annotations here valid? To be quite honest lifetimes are still a challenge for me, but this is what we came up with to get it in a working state
2. Is there any recommended patterns / ideas regarding the auto-refreshing token implementation in Rust? Or should we just side-step this and revisit it later. But from my cursory digging into other SDKs:
    - .NET has its own [class ](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/BearerTokenAuthenticationPolicy.cs#L176)`AccessTokenCache` which has some logic to make it able to refresh when necessary
    - Python added an `on_request` that is called on every request before sending, and[ has some logic](https://github.com/vincenttran-msft/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py#L100) checking for a currently `None` token or checking `self._need_new_token` (but not 100% sure what maintains the state of `_need_new_token`)
    - EDIT: Doi, maybe I should look at the code deeper than a scroll-through 😆 `_need_new_token` is a property that literally checks a similar condition as you stated (in Python's case, 300 seconds or 5 mins)
   
So we can either iterate on this PR to try to work on the auto-caching, or we can leave it on the backlog and revisit it when necessary 😄


heaths:
> Is there any recommended patterns / ideas regarding the auto-refreshing token implementation in Rust?

.NET and Python are actually doing the same thing. I was involved in the development of .NET's. It's still determining if the token needs refreshed on each request via the pipeline. It needs a trigger: there's no thread running concurrently doing this. The logic is basically: if the token for the current call is within 2 minutes of expiring or already expired, fetch a new one. The original request waits on that to complete. From what you described, sounds like Python is conceptually doing the same and I can't imagine what any language would do differently since we wouldn't want to burn a thread (green or otherwise) just to refresh tokens when they may not be needed for long after a token expires.

sdk/core/azure_core/src/policies/mod.rs
```diff
@@ -1,10 +1,12 @@
 mod custom_headers_policy;
 mod retry_policies;
+mod storage_bearer_token_policy;
```
jalauzon-msft, line 3: We should not include `storage` anywhere for this since it is going into Core and will be common for all packages. `bearer_token_policy` is probably fine for the module.

Though I am curious on what this is called in other languages. Isn't it something like TokenCredentialPolicy?

sdk/core/azure_core/src/policies/storage_bearer_token_policy.rs
```diff
@@ -0,0 +1,39 @@
...
+#[derive(Debug, Clone)]
+pub struct BearerTokenCredentialPolicy<'a> {
+    credential: Arc<dyn TokenCredential>,
+    scopes: &'a [&'a str],
+}
+
+impl<'a> BearerTokenCredentialPolicy<'a> {
+    pub fn new(credential: Arc<dyn TokenCredential>, scopes: &'a [&'a str]) -> Self {
+        Self { credential, scopes }
+    }
```
heaths, line 19: If you stored a `Vec<String>`, you'd want to take a slice of `Into<String>`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=548f8e639f4e9216eec24c5d53f83923

sdk/core/azure_core/src/policies/storage_bearer_token_policy.rs
```diff
@@ -0,0 +1,39 @@
...
+        &self,
+        ctx: &Context,
+        request: &mut Request,
+        next: &[Arc<dyn Policy>],
+    ) -> PolicyResult {
+        let access_token = self.credential.get_token(self.scopes).await?;
+        let token = access_token.token.secret();
+
+        request.insert_header("authorization", format!("Bearer {token}"));
+        debug!("the following request will be passed to the transport policy: {request:#?}");
```
heaths, line 35: Credential leak. We'll have a common facility for secure tracing and logging as with all our other languages.

sdk/core/azure_core/src/policies/storage_bearer_token_policy.rs
```diff
@@ -0,0 +1,39 @@
...
+    async fn send(
+        &self,
+        ctx: &Context,
+        request: &mut Request,
+        next: &[Arc<dyn Policy>],
+    ) -> PolicyResult {
+        let access_token = self.credential.get_token(self.scopes).await?;
+        let token = access_token.token.secret();
+
+        request.insert_header("authorization", format!("Bearer {token}"));
```
heaths, line 34: Import `azure_core::headers::AUTHORIZATION` and do something like:

```rust
request.insert_header(AUTHORIZATION, format!("Bearer {token}"));
```

Easier to keep track of usage.

heaths (changes requested):
One major blocker to clean up, but some API changes requested as well.

sdk/core/azure_core/src/policies/mod.rs
vincenttran-msft: https://github.com/Azure/azure-sdk-for-net/blob/ede7138b0150024daa440819cc0008cd51f46bc1/sdk/core/Azure.Core/src/Pipeline/BearerTokenAuthenticationPolicy.cs#L17

.NET uses `BearerTokenAuthenticationPolicy`

sdk/core/azure_core/src/policies/storage_bearer_token_policy.rs
vincenttran-msft: I can take a shot at this implementation. Thanks for the helpful playground link, what a nifty map!

sdk/core/azure_core/src/policies/bearer_token_policy.rs
```diff
@@ -0,0 +1,50 @@
...
+        ctx: &Context,
+        request: &mut Request,
+        next: &[Arc<dyn Policy>],
+    ) -> PolicyResult {
+        let scopes = self
+            .scopes
+            .iter()
+            .map(|s| s.as_str())
+            .collect::<Vec<&str>>();
+        let access_token = self.credential.get_token(&scopes).await?;
```
vincenttran-msft, line 43: Does this now seem slightly odd / clunky? The crux of the issue is that `self.credential.get_token` expects type `&[&str]` but since we are storing `scopes` now as `Vec<String>`, there is no zero-cost conversion (to my knowledge) since these iirc look completely different in memory. 

Does this mean `Vec<String>` may not be the best choice here? Another option is to explore whether `get_token` should accept `&[&str]`.

sdk/core/azure_core/src/policies/storage_bearer_token_policy.rs
```diff
@@ -0,0 +1,39 @@
...
+    Context, Request,
+};
+use async_trait::async_trait;
+use std::sync::Arc;
+use tracing::debug;
+
+#[derive(Debug, Clone)]
+pub struct BearerTokenCredentialPolicy<'a> {
+    credential: Arc<dyn TokenCredential>,
+    scopes: &'a [&'a str],
```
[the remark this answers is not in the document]
vincenttran-msft: I took this change as well, but have left a follow-up comment below that I am curious to hear your thoughts on 😄

sdk/core/azure_core/src/policies/bearer_token_policy.rs
heaths: You can simplify it slightly: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=71c2cfc4468771da12936a187a4ea5c3

That said, we should probably change `get_token` to be more accepting as well. You can file an issue if you like and think this might be common enough to basically run this routine in the credential impl itself rather than callers.

sdk/core/azure_core/src/policies/bearer_token_policy.rs
```diff
@@ -0,0 +1,76 @@
...
+        if OffsetDateTime::now_utc().unix_timestamp() - self.last_refresh_time
+            > DEFAULT_REFRESH_TIME
+        {
+            self.need_new_token().await;
+        }
+        let token = self.access_token.clone().unwrap();
+        let token = token.token;
+        let token = token.secret();
+        let token = String::from(token);
+        request.insert_header(AUTHORIZATION, format!("Bearer {token}"));
```
vincenttran-msft, line 68: This will all go away, I mostly have this broken out so I can see the types changing

sdk/core/azure_core/src/policies/bearer_token_policy.rs
vincenttran-msft: I have taken this feedback- but we can hold off on filing an issue until maybe we see other scenarios that reveal that running this routine in the credential makes more sense than the callers as we do at the moment 😄

vincenttran-msft:
Alrighty, so at this point the code is not compiling, and here is my compiler errors:

```
**error[E0596]: cannot borrow `*__self` as mutable, as it is behind a `&` reference
  --> sdk\core\azure_core\src\policies\bearer_token_policy.rs:62:13
   |
62 |             self.need_new_token().await;
   |             ^^^^ `__self` is a `&` reference, so the data it refers to cannot be borrowed as mutable
   |
help: you can `clone` the `BearerTokenCredentialPolicy` value and consume it, but this might not be your desired behavior
  --> sdk\core\azure_core\src\policies\bearer_token_policy.rs:54:10
   |
54 |         &self,
   |          ^^^^
help: consider specifying this binding's type
   |
54 |         &self: &mut BearerTokenCredentialPolicy,
   |              ++++++++++++++++++++++++++++++++++

For more information about this error, try `rustc --explain E0596`.
error: could not compile `azure_core` (lib) due to 1 previous error**
```

After fighting with it for about an hour, I took a step back to think deeply about what is going wrong, and now an implementation question:

Currently, the way that I am trying to make the retry mechanism work is:

- Maintain a `last_refresh_time` and `access_token` on the policy itself that we can use to either check to refresh or to use the token itself
- Use a function `need_new_token` that gets called if the condition is met (>120 secs since last refresh), and this gets a new access token, and should update `self.last_refresh_time` and `self.access_token`.

But the issue I am running into is that because we are doing `impl Policy for BearerTokenCredentialPolicy`, this means we have to match the signature on `send`, and `send` only has a `&self` (non-mutable).
With that being said, is there any way given our current requirements, that we can implement this "caching" policy? I cannot wrap my head around how to store and update some values that live on `self` when we only have a non-mutable reference (something tells me-- you shouldn't be able to, that's why it's a `&self` and not a `&mut self`). But in that case, how do we go about implementing some caching?

Curious to hear your thoughts @heaths and apologies for the wall of text, but I wanted to be as thorough as possible. Thanks!

sdk/core/azure_core/src/policies/bearer_token_policy.rs
```diff
@@ -22,8 +28,23 @@ impl BearerTokenCredentialPolicy {
...
+        self.scopes
+            .iter()
+            .map(String::as_str)
+            .collect::<Vec<&str>>()
+    }
+
+    async fn need_new_token(&mut self) {
+        self.last_refresh_time = OffsetDateTime::now_utc().unix_timestamp();
+        let access_token = self.credential.get_token(&self.scopes()).await;
+        self.access_token = Some(access_token.unwrap());
```
heaths, line 46: Don't call `unwrap()` in production code unless you know for sure it'll never fail. This can.

1. This function is probably better as `refresh()`.
2. This function should return an `azure_core::Result<()>`.
3. Then on line 45, just do `... .await?;`

heaths (changes requested):

heaths:
@vincenttran-msft are you going to be able to finish this, or would you like me to prioritize this in my queue?

vincenttran-msft:
> @vincenttran-msft are you going to be able to finish this, or would you like me to prioritize this in my queue?

I am going to try to get the interior mutability working with Mutex by EOD, otherwise I can revert this back to non-caching and we can review + prioritize this to merge without the caching capabilities.
Thanks!

vincenttran-msft:
@heaths Hi Heath, I have gone ahead and just reverted back to the non-caching implementation. I did notice that the other policies have been moved out to `typespec_client_core`, so please let me know if we need to move this policy as well. 

We can look to get this reviewed and merged in this state as I don't want to be blocking any other developments any longer on this item. Sorry for the delay!

heaths:
Making the title more generic since this isn't just for Storage. Also, we have labels we generally like for this stuff because that triggers other events sometimes.

heaths:
> @heaths Hi Heath, I have gone ahead and just reverted back to the non-caching implementation. I did notice that the other policies have been moved out to `typespec_client_core`, so please let me know if we need to move this policy as well.
> 
> We can look to get this reviewed and merged in this state as I don't want to be blocking any other developments any longer on this item. Sorry for the delay!

When do you think you might have the caching in, or is this something you want help with? I helped a bit on <https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/BearerTokenAuthenticationPolicy.cs>, which was refactored out of <https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/Azure.Security.KeyVault.Shared/src/ChallengeBasedAuthenticationPolicy.cs> and we'll need to accommodate for Key Vault for Rust here as well. In fact, Key Vault will be our first service crate we end up releasing, most likely.

/cc @jhendrixMSFT

heaths:
Also, yes, we can leave this in `azure_core`. I don't think it's Azure-specific, but seems other languages are leaving it in `azure_core` instead of their unbranded packages - at least, .NET and Go left it in Azure core.

vincenttran-msft:
> > @heaths Hi Heath, I have gone ahead and just reverted back to the non-caching implementation. I did notice that the other policies have been moved out to `typespec_client_core`, so please let me know if we need to move this policy as well.
> > We can look to get this reviewed and merged in this state as I don't want to be blocking any other developments any longer on this item. Sorry for the delay!
> 
> When do you think you might have the caching in, or is this something you want help with? I helped a bit on https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/BearerTokenAuthenticationPolicy.cs, which was refactored out of https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/Azure.Security.KeyVault.Shared/src/ChallengeBasedAuthenticationPolicy.cs and we'll need to accommodate for Key Vault for Rust here as well. In fact, Key Vault will be our first service crate we end up releasing, most likely.
> 
> /cc @jhendrixMSFT

I think the caching is something I may need help with. I've gotten close, and I've scoured the internet for resources regarding interior mutability in a thread-safe manner (using RwLock and Mutex), but I just _cannot_ for the life of me ever get it completely satisfying all requirements. One of the main issues with using Mutex was that we have: `#[derive(Clone)]`, which cannot easily be auto-generated for our Mutex (I get kicked back by the compiler).

I know that you can then provide a hand-written implementation, but I think this is all still a little outside of my scope of knowledge. So I think the next best step would be if you could implement the caching, and I will probably learn more from seeing your implementation and understanding how it works rather than sinking any more time (which has been a great learning opportunity, but I also am working on some Python SDK commitments as well which are taking longer than expected)

heaths (approved):
Good as an MVP, but we'll definitely want to cache the access token. What we did in .NET was, if the token is already expired or within 2 minutes of expiring, is fetch it again. It should be indexed based on the authority, as you can see in <https://github.com/Azure/azure-sdk-for-net/blob/125bfbb68c585771c6c60f32e0542d45bfc20cfe/sdk/keyvault/Azure.Security.KeyVault.Shared/src/ChallengeBasedAuthenticationPolicy.cs#L48>. This is Key Vault's modifications to our more generic <https://github.com/Azure/azure-sdk-for-net/blob/125bfbb68c585771c6c60f32e0542d45bfc20cfe/sdk/core/Azure.Core/src/Pipeline/BearerTokenAuthenticationPolicy.cs>.

PR #1730, opened 2024-08-01
Configure and fix spelling issues

heaths:


eng/dict/crates.txt
```diff
@@ -0,0 +1,36 @@
+async-lock
```
RickWinter, line 1: Should the two eng/dict files be in eng/cspell since that is what they are used for?

RickWinter (approved):

LarryOsterman (approved):

eng/dict/crates.txt
heaths: They are fairly generic word lists, so I'd rather not. We can use those with other spell checkers if we need to e.g., Rust Rover, if they have some support because I don't think cspell has a plugin for them.

PR #1736, opened 2024-08-02
Refactor json and xml mods into typespec crate

heaths:


heaths:
Actually, given something @JeffreyRichter said in a Go review the other day, I should pull these back up to `typespec_client_core` and we can move them down the dependency stack later if and when needed. We can't pull them up post-GA.

jhendrixMSFT (approved):

Issue #1737, opened 2024-08-02
Replace AsClientOptions, AsClientMethodOptions with derive macro-defined functions


heaths:
Once the derive macro is complete for client and client method options, we should remove the `AsClientOptions` and `AsClientMethodOptions` added for #1734 as a stopgap solution. They'd serve no purpose at that point and just end up being yet another import devs will have to add; though, rust-analyzer should make short work of that.

heaths:
No longer relevant since we're using PODO

PR #1738, opened 2024-08-02
Add traits with getters for ClientOptions, ClientMethodOptions

### LINKED ISSUE #1734: Add public getters to ClientOptions and ClientMethodOptions
Forgot to migrate public getters from https://github.com/heaths/azure-sdk-for-rust-proto/blob/e1a0df015120c0b29083bd6483a94720a32691df/sdk/core/src/options/mod.rs#L33. See https://github.com/Azure/autorest.rust/pull/67/files#r1700855374 for context.

heaths:
Resolves #1734, at least temporarily.

### DIFF
--- a/sdk/core/azure_core/src/options/mod.rs
+++ b/sdk/core/azure_core/src/options/mod.rs
@@ -66,7 +66,7 @@ impl ClientOptions {
         self.retry = Some(retry.into());
     }
 
-    /// Set [`TelemetryOptions`] used by the client.
+    /// Set the [`TelemetryOptions`] used by the client.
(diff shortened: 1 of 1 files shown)


heaths:
-

heaths:
-

heaths:
This provides at least a temporary workaround. I created #1737 as a follow up that, once we have the derive macro, we should probably remove these and will just generate the getters. IIRC, that's what I had intended before I opened #1734 but it's been lower priority.

We'll also want to rebase this on `feature/track2` once #1736 is merged.

jhendrixMSFT (approved):

PR #1740, opened 2024-08-06
Added content to the readme for Rust Event Hubs

LarryOsterman:
Lots more to come


sdk/eventhubs/azure_messaging_eventhubs/README.md
```diff
@@ -1,7 +1,302 @@
-# azure_messaging_servicebus
+<!-- cspell:words eventhubs  -->
```
heaths, line 1: Probably used enough just to add to the root `.vscode/cspell.json`

sdk/eventhubs/azure_messaging_eventhubs/README.md
```diff
@@ -1,7 +1,302 @@
...
+# Azure Event Hubs Client Package for Rust
 
-This crate is part of a collection of crates: for more information please refer to [https://github.com/azure/azure-sdk-for-rust](https://github.com/azure/azure-sdk-for-rust).
+Azure Event Hubs crate for the Microsoft Azure SDK for Rust.
 
 License: MIT
+
+[Azure Event Hubs](https://azure.microsoft.com/services/event-hubs/) is a big data streaming platform and event ingestion service from Microsoft. For more information about Event Hubs see: [link](https://docs.microsoft.com/azure/event-hubs/event-hubs-about).
+
+Use the client library `` in your application to:
```
heaths, line 11: Forgot to fill in between the backticks?

sdk/eventhubs/azure_messaging_eventhubs/README.md
```diff
@@ -1,7 +1,302 @@
...
+
+## Getting started
+
+### Install the package
+
+Add the Azure Event Hubs client package for rust to your `cargo.toml` file:
+
+```toml
+[dependencies]
+azure_messaging_eventhubs = { version="*" }
```
heaths, line 31: This is not good advice. It should be explicit, or at least a major and/or minor range. Alternatively, just tell them to run `cargo add azure_messaging_eventhubs`...which will work someday once we publish, which what you have won't work now anyway.

sdk/eventhubs/azure_messaging_eventhubs/rust-toolchain.toml
```diff
@@ -1,2 +0,0 @@
-[toolchain]
-channel = "1.78.0"
```
heaths, line 2: Did you have a specific need for 1.78? Our current MSRV is 1.76.

sdk/eventhubs/azure_messaging_eventhubs/README.md
```diff
@@ -1,7 +1,302 @@
...
+            }
+        }
+    }
+```
+
+# Troubleshooting
+
+## Logging
+
+The EventHubs SDK client uses the [tracing](https://docs.rs/tracing/latest/tracing/) package to
```
heaths, line 251: Use "Event Hubs" or "EventHubs" consistently. It had a space more often above that I remember.

sdk/eventhubs/azure_messaging_eventhubs/README.md
```diff
@@ -1,7 +1,302 @@
-# azure_messaging_servicebus
+<!-- cspell:words eventhubs  -->
 
-Azure Event Hubs crate for the Microsoft Azure SDK for Rust.
+# Azure Event Hubs Client Package for Rust
 
-This crate is part of a collection of crates: for more information please refer to [https://github.com/azure/azure-sdk-for-rust](https://github.com/azure/azure-sdk-for-rust).
+Azure Event Hubs crate for the Microsoft Azure SDK for Rust.
 
 License: MIT
```
heaths, line 7: License is already at the bottom with the usual block. I don't recall seeing it this high or this basic in any other language. Would just remove.

heaths (approved):

sdk/eventhubs/azure_messaging_eventhubs/README.md
LarryOsterman: I'll change it to cargo add.

sdk/eventhubs/azure_messaging_eventhubs/README.md
```diff
@@ -296,7 +295,7 @@ Azure SDK for C++ is licensed under the [MIT](https://github.com/Azure/azure-sdk
 [source]: https://github.com/Azure/azure-sdk-for-rust/tree/feature/track2/sdk/eventhubs/azure_messaging_eventhubs
 [azure_identity_pkg]: https://docs.rs/azure_identity/latest/azure_identity/
 [default_azure_credential]: https://docs.rs/azure_identity/latest/azure_identity/struct.DefaultAzureCredential.html
-[cppdoc]: https://docs.rs/azure_messaging_eventhubs/latest/azure_messaging_eventhubs
-[cppdoc_examples]: https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/eventhubs/azure-messaging-eventhubs/samples
+[rustdoc]: https://docs.rs/azure_messaging_eventhubs/latest/azure_messaging_eventhubs
+[rustdoc_examples]: https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/eventhubs/azure-messaging-eventhubs/samples
```
heaths, line 299: Once set up, the link checker will 404 and fail this. Probably better to use the `feature/track2` branch instead of `main`. We'll have a lot to flip later anyway.

sdk/eventhubs/azure_messaging_eventhubs/docs/eh_consumer.md
heaths, line 1: You should `include_str!()` this as I showed in my demo the other day: https://github.com/heaths/samples-rs/blob/main/src/docs/index.rs

sdk/eventhubs/azure_messaging_eventhubs/README.md
LarryOsterman: Just checked - it's officially `Event Hubs`. I believe I've updated all of the instances.

sdk/eventhubs/azure_messaging_eventhubs/docs/eh_consumer.md
LarryOsterman: I do - see `src\lib.rs`

LarryOsterman: And I'm doing that explicitly because you called it out in the demo the other day :). It's WAY much cleaner than having all that crud in the sources.

sdk/eventhubs/azure_messaging_eventhubs/rust-toolchain.toml
LarryOsterman: I did - that's the version that was used for iron oxide. But it works just fine with 1.76 so I removed the file (it was accidentally added).

sdk/eventhubs/azure_messaging_eventhubs/docs/eh_consumer.md
heaths: Oh,

https://github.com/Azure/azure-sdk-for-rust/pull/1740/files/2f14562cc5acd2e1f189f1a084f76953f087959a..0607142f3cf4b1731f3de52cd6dcb5c58568b2cb#diff-767c3b3e1a00d0087d7cac040e2ee4c3aa387d198ad50df98a5a34a5ed3d643cR35-R36

So, like, a README for the module? That works. What I demoed was so that lines match up for pure tests. This isn't a bad idea, though. So it's not a "samples" doc as much as a module README, yeah?

sdk/core/azure_core_amqp/README.md
```diff
@@ -1,4 +1,4 @@
-<!-- cspell: words EVENTHUB eventhubs -->
+<!-- cspell: words  -->
```
heaths, line 1: Nit: just get rid of this now and delete the initial empty lines so the H1 is first.

sdk/eventhubs/azure_messaging_eventhubs/README.md
LarryOsterman: The rustdoc links aren't going to work until we actually publish the package event if we say feature/track2 branch.

And I think I do use feature/track2 in my link to the repo.

heaths: The markdown-rendered links will, though; but, yes, you make a good point. It's still going to fail the build if @hallipr beats you to setting up the link checker vs. publishing track 2 betas to crates.io. I still think it's worth fixing for that reason.

sdk/core/azure_core_amqp/README.md
LarryOsterman: Yes, the eh_consumer.md file is essentially a README for the module. 

Somehow I convinced GH copilot to create this and I honestly decided I really liked it (because the module level documentation is incredibly sparse). 

So rather than cut and paste the entire goo as a rustdoc comment, I used include_str to include it into the module documentation.

I'm rather happy with how it renders :).

heaths: I like the idea, but to standardize around something, what if we make the module a directory with its own README, to keep everything nice and tight together i.e.,

```
src/
  consumer/
    README.md
    mod.rs
  lib.rs
```

That also has the benefit of rendering in the GitHub web UI.

PR #1741, opened 2024-08-06
Add components to rust-toolchain.toml

heaths:
Makes sure these components are always installed when running
    `rustup show` or other tools. Adding `rust-analyzer` is questionable,
    but enough dev UX uses it's probaby fine and not too slow to acquire on
    CI build machines.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,10 @@
 [toolchain]
 channel = "1.76.0"
+components = [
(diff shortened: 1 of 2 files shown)


heaths:
Fixed the base. One of these times I'll remember to pass `--base` to `gh pr create`. Hopefully we land cli/cli#8979.

weshaggard (approved):

PR #1742, opened 2024-08-07
Moved consumer readme to consumer directory

LarryOsterman:


heaths (changes requested):
I would do it like described in <https://github.com/Azure/azure-sdk/pull/7847/files>: basically, in `cosumer/mod.rs` add toward the top after the copyright notice:

```rust
#![doc = include_str!("README.md")]

// ...
```

Keeps everything together nice and neat.

LarryOsterman:
> I would do it like described in https://github.com/Azure/azure-sdk/pull/7847/files: basically, in `cosumer/mod.rs` add toward the top after the copyright notice:
> 
> ```rust
> #![doc = include_str!("README.md")]
> 
> // ...
> ```
> 
> Keeps everything together nice and neat.

Done.

LarryOsterman:
@heaths, can you please take a look at the current changes?

sdk/eventhubs/azure_messaging_eventhubs/src/consumer/mod.rs
```diff
@@ -3,6 +3,7 @@
 
 //cspell: words amqp mgmt amqps
 
+#![doc = include_str!("./README.md")]
```
heaths, line 6: Interesting. Was the PWD reference needed? I'm guessing so. I suppose it jives with lib lookup rules on non-Windows platforms.

(Not blocking - just a Q)

heaths (approved):

sdk/eventhubs/azure_messaging_eventhubs/src/consumer/mod.rs
LarryOsterman: The PWD reference was *not* needed. I just included it because, well, I pretty much always include it (I have a trust issue with tools that aren't passed in either absolute or relative paths - it's almost always safer to add a relative path in a filename).

PR #1743, opened 2024-08-09
chore: use Deref for single struct wrapper

ouertani:


ouertani:

heaths:
Thank you for the contribution. Can you please explain how this is beneficial? Since this is unofficial source and we are busy working on a new codebase, whether to merge it against main and/or our future feature branch needs a clear explanation of why.

heaths (changes requested):
Per comment, need an explanation of how this is beneficial.

ouertani:
> Per comment, need an explanation of how this is beneficial.

`Deref` can streamline code by providing direct access to the inner type of a container. For example, in the struct `TokenCache(RwLock<HashMap<Vec<String>, AccessToken>>)`, the **Deref** trait allows treating _TokenCache_ as an `RwLock<HashMap<Vec<String>, AccessToken>>` without explicitly using `.0` which simplifies code and enhances readability.

heaths:
I appreciate what it's for, but in this case it's a `pub(crate)` type but otherwise exposes the field publicly, which isn't always desirable. Should its visibility change to `pub` in the future, we are incidentally over-exposing. It's better to open up as needed than to reduce visibility later - which we can't, because that'd be a breaking change.

Thanks for your consideration, but we prefer to open things as needed and I don't see a strong need here. If there are other reasons you have in mind, please reply here and we can reconsider.

PR #1748, opened 2024-08-13
Added some AMQP types that were missed in the initial conversion, but…

LarryOsterman:
… are needed for C++ interoperability

### DIFF
--- a/sdk/core/azure_core_amqp/src/fe2o3/value.rs
+++ b/sdk/core/azure_core_amqp/src/fe2o3/value.rs
@@ -162,6 +162,7 @@ impl From<AmqpValue> for fe2o3_amqp_types::primitives::Value {
             AmqpValue::Array(a) => fe2o3_amqp_types::primitives::Value::Array(
                 a.into_iter().map(|v| v.into()).collect(),
             ),
+            AmqpValue::Composite(_) => panic!("Composite values are not supported in Fe2o3"),
             AmqpValue::Described(d) => fe2o3_amqp_types::primitives::Value::Described(Box::new(
                 serde_amqp::described::Described {
                     descriptor: d.descriptor.clone().into(),
@@ -326,6 +327,8 @@ impl PartialEq<AmqpValue> for fe2o3_amqp_types::primitives::Value {
                 fe2o3_amqp_types::primitives::Value::Described(a) => **d == **a,
                 _ => false,
             },
+
+            AmqpValue::Composite(_) => panic!("Composite values are not supported in Fe2o3"),
 
             AmqpValue::Unknown => todo!(),
         }
--- a/sdk/core/azure_core_amqp/src/value.rs
+++ b/sdk/core/azure_core_amqp/src/value.rs
@@ -144,7 +144,6 @@ impl AmqpComposite {
     }
 }
 
-
 #[derive(Debug, PartialEq, Clone, Default)]
 pub enum AmqpValue {
     #[default]
--- a/sdk/core/azure_core_amqp/src/value.rs
+++ b/sdk/core/azure_core_amqp/src/value.rs
@@ -36,7 +36,7 @@ impl AmqpList {
         Self(Vec::new())
     }
 
-    pub fn new_with_size(size: usize) -> Self {
+    pub fn new_with_capacity(size: usize) -> Self {
         Self(Vec::with_capacity(size))
     }
 
--- a/sdk/core/azure_core_amqp/src/value.rs
+++ b/sdk/core/azure_core_amqp/src/value.rs
@@ -207,7 +207,7 @@ where
     }
 
     pub fn is_empty(&self) -> bool {
-        self.0.is_empty()
+        self.inner.is_empty()
     }
 
     pub fn remove(&mut self, key: &K) -> Option<V> {
--- a/sdk/core/azure_core_amqp/src/value.rs
+++ b/sdk/core/azure_core_amqp/src/value.rs
@@ -36,6 +36,14 @@ impl AmqpList {
         Self(Vec::new())
     }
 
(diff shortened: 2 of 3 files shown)


sdk/core/azure_core_amqp/src/fe2o3/value.rs
```diff
@@ -162,6 +162,7 @@ impl From<AmqpValue> for fe2o3_amqp_types::primitives::Value {
             AmqpValue::Array(a) => fe2o3_amqp_types::primitives::Value::Array(
                 a.into_iter().map(|v| v.into()).collect(),
             ),
+            AmqpValue::Composite(_) => panic!("Composite values are not supported in Fe2o3"),
```
heaths, line 165: Is panicking truly the right thing here? It'll bring down the app. Is it low likelihood at least? Can we document this? You can put docs on trait impls to override the trait docs if that would help.

sdk/core/azure_core_amqp/src/value.rs
```diff
@@ -36,6 +36,14 @@ impl AmqpList {
         Self(Vec::new())
     }
 
+    pub fn new_with_size(size: usize) -> Self {
```
heaths, line 39: Nit: `with_capacity` is idiomatic, just like you call in the line below.

sdk/core/azure_core_amqp/src/value.rs
```diff
@@ -103,6 +111,37 @@ impl AmqpDescribed {
...
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct AmqpComposite {
+    pub descriptor: AmqpDescriptor,
+    pub value: AmqpList,
+}
+
+impl AmqpComposite {
+    pub fn new(descriptor: impl Into<AmqpDescriptor>, value: impl Into<AmqpList>) -> Self {
```
heaths, line 131: Is the `impl Into<T>` overkill here? That's mainly meant for common types like `String`.

heaths (approved):
A few questions/nits, but nothing blocking.

sdk/core/azure_core_amqp/src/fe2o3/value.rs
LarryOsterman: I don't see a way of returning an error in a From trait implementation.

The Fe2o3 AMQP implementation doesn't directly support the AMQP "Composite" type - instead it treats AMQP "Composite" types as a serialization artifact - basically each serializable construct has a set of composition attributes which trigger the creation of an AMQP Composite when serializing and deserializing the type. TBH, this is actually rather clever. But uAMQP *does* have Composite as a first class primitive, so for interoperability, I need to have one in the AMQP wrapper.

That works great until you try to convert an AMQP type to an Fe2o3 type. If you do that, I * want* the app to be terminated, because it's never going to work.

sdk/core/azure_core_amqp/src/value.rs
LarryOsterman: Makes sense, fixing.

sdk/core/azure_core_amqp/src/value.rs
LarryOsterman: I've started using `impl Into<T>` pretty much everywhere because it means that I can do things like passing in a either a string or an integer to `AmqpComposite::new`, and I can also do things like passing in a `Vec<AmqpValue>` to the constructor without having to force the caller to type `.into()` all over the place.

Basically it moves the work of calling `.into()` to the implementation rather than forcing the consumer of the type to call `.into()`. And tbh, I hate typing `.into()` all over the place :).

If moving `.into()` calls into the implementation of a function is not considered idiomatic, then so be it.

sdk/core/azure_core_amqp/src/value.rs
```diff
@@ -36,7 +36,7 @@ impl AmqpList {
         Self(Vec::new())
     }
 
-    pub fn new_with_size(size: usize) -> Self {
+    pub fn new_with_capacity(size: usize) -> Self {
```
heaths, line 39: Anything wrong with just `with_capacity`? It's not idiomatic to preface that with "new". `new` typically stands alone. See the hordes of `with_capacity` even in `std` for prior art.

LarryOsterman: Nothing at all - I already fixed it after your first comment.

sdk/core/azure_core_amqp/src/value.rs
LarryOsterman: FWIW, one of the other really *nice* aspects of doing the `Into<EventData>` thingy is that you can do:

```rust
message_producer.send_event("This is a test message", None);
message_producer.send_event(vec![1,2,3,4], None);
```

which IMHO is a nicer developer experience from:
```rust
message_producer.send_event("This is a test message".into(), None);
message_producer.send_event(vec![1,2,3,4].into(), None);
```

that's because the `.into()` doesn't contribute to the developer experience, in fact it detracts from it because you try the thing you want to try (just adding the message body), then you get an error about a type mismatch, then you have to add the `.into()` to resolve it.

heaths: No, that's fine to leave it as-is. I just wasn't expecting concrete types to often-enough have `Into<AmqpComposite>` impls, but if enough to make it worth it, great!

sdk/core/azure_core_amqp/src/fe2o3/value.rs
heaths: There is `TryFrom`, but not sure if that makes things harder for C++ and/or the Rust UX. Even in that case, `From` could only panic.

heaths (approved):

PR #1752, opened 2024-08-14
move is_expired to be a method on AccessToken

demoray:
Checking if an access token is expired is needed by consumers.  rather than have users of the SDK recreate this, this makes the internal function accessible off of AccessToken.

### DIFF
--- a/sdk/core/azure_core/src/auth.rs
+++ b/sdk/core/azure_core/src/auth.rs
@@ -83,7 +83,7 @@ impl AccessToken {
     ///
     /// If no duration is provided, then the default duration of 30 seconds is used.
     pub fn is_expired(&self, window: Option<Duration>) -> bool {
-        self.expires_on < OffsetDateTime::now_utc() + skew.unwrap_or(Duration::from_secs(30))
+        self.expires_on < OffsetDateTime::now_utc() + window.unwrap_or(Duration::from_secs(30))
     }
 }
 
--- a/sdk/core/azure_core/src/auth.rs
+++ b/sdk/core/azure_core/src/auth.rs
@@ -1,7 +1,7 @@
 //! Azure authentication and authorization.
 
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


sdk/core/azure_core/src/auth.rs
```diff
@@ -78,6 +78,14 @@ impl AccessToken {
             expires_on,
         }
     }
+
+    /// Check if the token is expired
+    ///
+    /// This checks if the token is expiring within the next 30 seconds to
+    /// address clock skew
```
heaths, line 85: Nit: punctuation looks more professional:

```suggestion
    /// Check if the token is expired.
    ///
    /// This checks if the token is expiring within the next 30 seconds to
    /// address clock skew.
```

Also, should we make the time configurable? Or not worth it until someone asks for it? Without overloads, we'd need a new method name, though, when that happens. Don't care too much, though: just putting out the question to see what you think. Could take an `Option<Duration>` and `unwrap_or()` to 30s.

sdk/identity/azure_identity/src/token_credentials/cache.rs
```diff
@@ -32,7 +27,7 @@ impl TokenCache {
         let token_cache = self.0.read().await;
         let scopes = scopes.iter().map(ToString::to_string).collect::<Vec<_>>();
         if let Some(token) = token_cache.get(&scopes) {
-            if !is_expired(token) {
+            if token.is_expired() {
```
heaths, line 30: Should keep the `!` logic here and below.

heaths (approved):

sdk/core/azure_core/src/auth.rs
demoray: addressed

heaths: If you're fine with that. It wasn't a "hard suggestion". Really just wondering out loud. :)

heaths (approved):

PR #1759, opened 2024-08-20
Caching Rough Draft

vincenttran-msft:


sdk/core/azure_core/src/policies/bearer_token_policy.rs
```diff
@@ -5,14 +5,20 @@ use crate::{
...
 #[derive(Debug, Clone)]
 pub struct BearerTokenCredentialPolicy {
     credential: Arc<dyn TokenCredential>,
     scopes: Vec<String>,
+    access_token: Arc<Mutex<String>>,
+    last_refresh_time: Arc<Mutex<i64>>,
 }
 
+/// Default timeout in seconds before refreshing a new token.
+const DEFAULT_REFRESH_TIME: i64 = 120;
```
heaths, line 20: Should be stored as a `Duration`. Type safety avoids common pitfalls. What if someone changes this to a negative value accidentally?

sdk/core/azure_core/src/policies/bearer_token_policy.rs
```diff
@@ -22,8 +28,31 @@ impl BearerTokenCredentialPolicy {
         Self {
             credential,
             scopes: scopes.into_iter().map(|s| s.into()).collect(),
+            access_token: Arc::new(Mutex::new("UNSET".to_string())), // Need a better default here or use Option
```
heaths, line 31: `Option<T>` is how you make things "nullable" in Rust.

sdk/core/azure_core/src/policies/bearer_token_policy.rs
```diff
@@ -22,8 +28,31 @@ impl BearerTokenCredentialPolicy {
...
+    fn update_data(&self, new_access_token: &str) {
+        let mut my_access_token = self.access_token.lock().unwrap();
+        *my_access_token = new_access_token.to_string();
+        let mut my_last_refresh_time = self.last_refresh_time.lock().unwrap();
+        *my_last_refresh_time = OffsetDateTime::now_utc().unix_timestamp();
+    }
+
+    fn should_refresh(&self) -> bool {
+        let current_time = OffsetDateTime::now_utc().unix_timestamp();
+        let my_last_refresh_time = self.last_refresh_time.lock().unwrap();
```
heaths, line 45: Never `unwrap()` in production code. This should have a better default e.g., `unwrap_or_default()` or the method needs to return a `std::result::Result<T, E>`.

sdk/core/azure_core/src/policies/bearer_token_policy.rs
```diff
@@ -22,6 +27,35 @@ impl BearerTokenCredentialPolicy {
...
+
+    fn scopes(&self) -> Vec<&str> {
+        self.scopes
+            .iter()
+            .map(String::as_str)
+            .collect::<Vec<&str>>()
+    }
+
+    fn refresh_token(&self, new_access_token: &AccessToken) {
+        let mut access_token = self.access_token.write().unwrap();
```
vincenttran-msft, line 42: Going to start a comment here but this serves as discussion for all uses of `unwrap()` that still exists here.
I know we have alternatives such as `unwrap_or_default()` but I couldn't quite wrap my head around how we would use it in these instances.

One thing is that these `unwrap()` should (from my understanding) never really raise since calling `.write()` or `.read()` before blocks the current thread until the lock can be acquired, and `unwrap()` should only raise if it's poisoned (a writer thread panics while holding an exclusive lock) which given the current code shouldn't ever occur? 

As always, open to feedbacks on how to make this more rigorous and robust! Thanks 😄

sdk/core/azure_core/src/policies/bearer_token_policy.rs
```diff
@@ -28,30 +28,44 @@ impl BearerTokenCredentialPolicy {
...
-        *my_last_refresh_time = OffsetDateTime::now_utc().unix_timestamp();
+    fn scopes(&self) -> Vec<&str> {
+        self.scopes
+            .iter()
+            .map(String::as_str)
+            .collect::<Vec<&str>>()
+    }
+
+    fn update_token(&self, new_access_token: &AccessToken) {
+        let mut access_token = self.access_token.lock().unwrap();
```
heaths: Never `unwrap()` in production. It'll crash the app. Functions that can fail and when that failure should bubble up to the caller should return a `Result<T, E>`. We wrap all ours in `azure_core::Result<T>`, for which you should use an appropriate `azure_core::Error` variant or `Other`.

sdk/core/azure_core/src/policies/bearer_token_policy.rs
```diff
@@ -28,30 +28,44 @@ impl BearerTokenCredentialPolicy {
...
+            }
+            None => true,
         }
-        return false;
     }
 
-    fn access_token(&self) -> String {
-        let my_access_token = self.access_token.lock().unwrap();
-        (*my_access_token.clone()).to_string()
+    fn access_token(&self) -> Result<String, &'static str> {
```
heaths, line 54: Only return `azure_core::Result<T>` from our APIs.

heaths: ...unless, in some cases, `io::Result`, `fmt::Result`, or other is more appropriate. In this case it's not.

heaths (changes requested):
Better, but never call `unwrap()` or `expect(&str)` in production. Functions like `unwrap_or_default(...)` or `unwrap_or_else(...)` are fine but not appropriate here. If the error should be bubbled up to the caller, your function should return a `Result<T>`. In this case, it may just be pertinent to skip. What do other languages' bearer token policies do in this case? We generally don't want to throw either.

sdk/core/azure_core/src/policies/bearer_token_policy.rs
```diff
@@ -69,17 +61,23 @@ impl Policy for BearerTokenCredentialPolicy {
...
         next: &[Arc<dyn Policy>],
     ) -> PolicyResult {
-        if self.is_token_expired() {
-            let access_token = self.credential.get_token(&self.scopes()).await?;
-            self.refresh_token(&access_token);
+        let access_token = self.access_token.as_ref().read().await;
+
+        if let Some(token) = &(*access_token) {
+            if token.is_expired(Some(DEFAULT_REFRESH_TIME)) {
+                drop(access_token);
```
heaths, line 68: Don't call this explicitly. It will be dropped when it gets shadowed in the line below. Same with the else below. This isn't idiomatic or necessary.

sdk/core/azure_core/src/policies/bearer_token_policy.rs
```diff
@@ -38,24 +40,14 @@ impl BearerTokenCredentialPolicy {
...
-            None => true,
-        }
-    }
-
-    fn access_token(&self) -> Result<String, &'static str> {
-        let access_token = self.access_token.read().unwrap().clone();
-        match access_token {
-            Some(access_token) => Ok(String::from(access_token.token.secret())),
-            None => Err("access_token is None."),
+    async fn access_token(&self) -> Result<String> {
```
heaths, line 43: Why is this returning a `Result` instead of an `Option`? Can the user do anything with the info? There's a lot of complication on here converting to and from different formats just to return a `String` from a `String`.

Could just be as simple as:

```rust
pub struct BearerTokenCredentialPolicy {
    access_token: RwLock<Option<String>>,
}

impl BearerTokenCredentialPolicy {
    pub fn access_token(&self) -> Option<String> {
        let access_token = self.access_token.read().ok()?;
        access_token.as_ref().map(|s| s.clone())
    }
}
```

Also not sure why this needs to be an `Arc` since it's internal. The policy itself will be in an `Arc` to be potentially shared in multiple pipelines, but that will be a shared object. Its fields are naturally shared.

heaths (changes requested):

sdk/core/azure_core/src/policies/bearer_token_policy.rs
vincenttran-msft: Hi Heath, thanks for the feedback! I chose to return `Result` here so that we could use the `?` operator in `send`, in case we do in fact fail to fetch the access token we can bubble up the error to the caller.

I can try to take a refactor to return `Option` here, but then in that case we would have to "unwrap" the option in `send` and handle the error case, which is all possible but might bloat up `send` when we could instead do the error handling in this separate function `access_token`?

In response to why it is `Arc`, that is because `RwLock` does not derive the `Clone` trait. 
sample output from compiler when using `RwLock<Option<AccessToken>>`:
```
15 |     access_token: RwLock<Option<AccessToken>>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `RwLock<std::option::Option<AccessToken>>`

```

heaths: Then the underlying error should be returned, not that the `Option` is `None`. That's not an error, really. Either the error needs to be retained (not ideal), or returning an `Option` is better. The fact an internal option is `None` is not actionable and, therefore, useless to callers.

There are lots of functions to map `Result` <==> `Option`. Search for "Result" and "Option" in <https://doc.rs/std>.

sdk/core/azure_core/src/policies/bearer_token_policy.rs
```diff
@@ -38,15 +58,22 @@ impl Policy for BearerTokenCredentialPolicy {
...
+            }
+        } else {
+            let mut access_token = self.access_token.write().await;
+            *access_token = Some(self.credential.get_token(&self.scopes()).await?);
+        }
 
-        request.insert_header(AUTHORIZATION, format!("Bearer {token}"));
+        request.insert_header(
+            AUTHORIZATION,
+            format!("Bearer {}", self.access_token().await.unwrap()),
```
vincenttran-msft, line 75: @heaths 
After taking your suggestion to refactor `access_token` to now return an `Option<String>`, this is the last "gotcha" that I am stuck on. What is the suitable way to handle the `Option<String>` unwrapping in the `send` method here? I know that it is definitely NOT `unwrap()` 😄
Thanks!

heaths: See [`Result::ok`](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.ok).

sdk/core/azure_core/src/policies/bearer_token_policy.rs
```diff
@@ -2,20 +2,27 @@
...
 
 use crate::{
-    credentials::TokenCredential,
+    credentials::{AccessToken, TokenCredential},
+    error::{Error, ErrorKind},
     headers::AUTHORIZATION,
     policies::{Policy, PolicyResult},
     Context, Request,
 };
+use async_lock::RwLock;
```
vincenttran-msft, line 11: Is this the desired RwLock? I modeled this off of the caching implementation in identity:
https://github.com/Azure/azure-sdk-for-rust/blob/feature/track2/sdk/identity/azure_identity/src/credentials/cache.rs#L4

sdk/core/azure_core/src/policies/bearer_token_policy.rs
```diff
@@ -38,15 +58,27 @@ impl Policy for BearerTokenCredentialPolicy {
...
-            .iter()
-            .map(|s| s.as_str())
-            .collect::<Vec<&str>>();
-        let access_token = self.credential.get_token(&scopes).await?;
-        let token = access_token.token.secret();
+        let access_token = self.access_token.read().await;
+
+        if let Some(token) = &(*access_token) {
+            if token.is_expired(Some(DEFAULT_REFRESH_TIME)) {
+                drop(access_token);
```
vincenttran-msft, line 65: I know previously we had a comment about this being unnecessary, but I found in my testing that the lock was getting hung up (from the read lock being acquired in L61 above). I was under the impression that we are able to have multiple readers **and** only one writer, but it seems like it is actually an or according to the [docs](https://docs.rs/async-lock/latest/async_lock/struct.RwLock.html#)
```
running 1 test
test blob_client::tests::test_download_blob has been running for over 60 seconds

sdk/core/azure_core/src/policies/bearer_token_policy.rs
heaths: We don't want to depend on any particular async runtime, so you can't use `tokio` directly. This should be fine.

sdk/core/azure_core/src/policies/bearer_token_policy.rs
heaths: It's about scope. You established the scope in L61 so it will live until the outer scope is closed. But yes, this is functionally correct.

jalauzon-msft (approved):

heaths (approved):

PR #1760, opened 2024-08-20
Test PR: DO NOT MERGE

hallipr:
Validation PR for the rust-pr pipeline.


heaths (changes requested):
Why? And could we add a bit more description as to why to the commit title and/or description (or to the PR, which will become the merged commit) for diagnostics?

PR #1761, opened 2024-08-20
make scripts more portable across linux distros

analogrelay:
For my previous PRs, I've been working on my Windows box, but I noticed that all the scripts for running tests and checks are POSIX scripts (which means I wasn't able to run them directly), so I'm presuming a Linux environment is preferred for development, which absolutely works for me :). However, my usual WSL environment runs [NixOS](https://nixos.org/) which doesn't have `bash` installed at `/bin/bash`. While most distros do put `bash` in `/bin/bash`, not all do, and the shebang `#!/bin/bash` can cause issues with portability across those distros.

This PR switches the shebang for scripts to `#!/usr/bin/env bash`, which is generally portable across distros, as it uses the `env` tool to resolve the path to `bash` using the `PATH` environment variable.

### DIFF
--- a/eng/common/testproxy/apply-dev-cert.sh
+++ b/eng/common/testproxy/apply-dev-cert.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 TMP_PATH=$CERT_FOLDER
 PFXFILE=$CERT_FOLDER/dotnet-devcert.pfx
 CRTFILE=$CERT_FOLDER/dotnet-devcert.crt
--- a/eng/scripts/all_tests.sh
+++ b/eng/scripts/all_tests.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -eux -o pipefail
 cd $(dirname ${BASH_SOURCE[0]})/../../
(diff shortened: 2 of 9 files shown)


heaths:
Thanks. I may have to pick your brain on nixOS. I've been learning about it for helix-editor/helix development but having various issues. Each time I have to remember how to work around them, and nixOS doesn't seem to jive with a docker container since it is, in effect, a container itself.

heaths (approved):
FYI: Most if not all of these scripts are being removed. Almost none of them do anything useful and will be replicated in Azure DevOps templates anyway. /cc @hallipr

analogrelay:
Fair enough. I do think there's quite a bit of value in having locally-runnable scripts that, where possible, replicate CI checks (so devs can run those checks easily).

heaths:
But the scripts themselves don't really do anything special. I 100% agree about CIs doing what devs can do. At best they just make sure to pass e.g., `--no-deps` or `--all-features`. While I personally think devs should do much more before submitted to a CI, there's disagreement in the team and some devs like the CI to do "all the stuff", so they can "move faster" (which seems counterproductive since then they have to wait for results instead of a local check, but I digress).

PR #1762, opened 2024-08-21
Add ADO pipeline scaffolding

hallipr:
- Add archetype template, jobs and steps
- Add languageSettings functions to support PR diff based builds


eng/pipelines/pr.yml
```diff
@@ -0,0 +1,19 @@
...
+    include:
+      - feature/track2
+
+parameters:
+  - name: Service
+    type: string
+    default: auto
+
+extends:
+  template: ./templates/stages/archetype-sdk-client.yml
```
weshaggard, line 14:
```suggestion
  template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
```

eng/pipelines/templates/jobs/ci.yml
```diff
@@ -0,0 +1,113 @@
+parameters:
+  - name: ServiceDirectory
+    type: string
+    default: ""
+  - name: BuildTargetingString
+    type: string
+    default: "azure-*"
```
weshaggard, line 7: This looks to be copied from python. Do the crates follow a similar scheme?

eng/pipelines/templates/jobs/ci.yml
```diff
@@ -0,0 +1,113 @@
...
+                parameters:
+                  Paths: [/*]
+
+              - template: /eng/pipelines/templates/steps/test-packages.yml
+                parameters:
+                  ServiceDirectory: ${{ parameters.ServiceDirectory }}
+                  BuildTargetingString: ${{ parameters.BuildTargetingString }}
+                  RustToolchain: ${{ RustToolchain }}
+                  ArtifactSuffix: ${{ matrix.pool.os }}_${{ RustToolchain }}
+                  PublishArtifacts: ${{ and(eq(matrix.pool.os, 'linux'), eq(RustToolchain, 'stable')) }}
```
heaths, line 70: This is the sort of conditioning that makes me nervous. Could instead add a boolean parameter like `publish: true` on that matrix slice and then check here: `${{ eq(matrix.publish, true) }}` (or maybe we don't need to use `eq` if empty strings/absent values are falsey already).

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -0,0 +1,22 @@
...
+      Toolchain: ${{ parameters.RustToolchain }}
+
+  - pwsh: |
+      . ./eng/common/scripts/Helpers/CommandInvocation-Helpers.ps1
+
+      $env:RUSTDOCFLAGS = "-D warnings"
+      $env:RUSTFLAGS = "-Dwarnings"
+
+      Invoke-LoggedCommand "cargo +$(RustToolchain) check -p azure_core --no-default-features"
+      Invoke-LoggedCommand "cargo +$(RustToolchain) install cargo-readme"
```
heaths, line 18: This isn't even being used. Let's get rid of this before. If/when we add third-party dependencies, I want to re-review now that it's in official source.

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -0,0 +1,22 @@
...
+      Invoke-LoggedCommand "cargo +$(RustToolchain) fmt --all -- --check"
+      Invoke-LoggedCommand "cargo +$(RustToolchain) clippy --all"
+      Invoke-LoggedCommand "cargo +$(RustToolchain) doc --all --no-deps"
```
heaths, line 21: You should also run `eng/scripts/verify-dependencies.rs` (source-as-script). You'd either need to run it as any command in bash, or pass the shebang command at the top of the file in pwsh.

eng/pipelines/templates/steps/use-rust.yml
```diff
@@ -0,0 +1,19 @@
...
+
+steps:
+  - pwsh: |
+      $toolchain = '${{ parameters.Toolchain }}'
+      if ($toolchain -eq 'msrv') {
+          Write-Host "Reading MSRV from azure_core"
+          $toolchain = cargo metadata --manifest-path ./sdk/core/azure_core/Cargo.toml --format-version 1 | convertfrom-json | select -expand packages | where { $_.name -eq 'azure_core' } | select -expand rust_version
+      }
+
+      Write-Host "Setting RustToolchain variable to $toolchain"
```
heaths, line 14: You changed the parameter name to `Toolchain` after our f2f conversation so this string is wrong, but I would change it to `Toolchain` everywhere. "Rust" is redundant since this is already a rust-specific file.

eng/scripts/Language-Settings.ps1
```diff
@@ -0,0 +1,92 @@
...
+    $allPackageProps = @()
+    Push-Location $RepoRoot
+    try {
+        $searchPath = 'sdk'
+
+        if ($ServiceDirectory -and $ServiceDirectory -ne 'auto') {
+            $searchPath = Join-Path 'sdk' $ServiceDirectory
+        }
+
+        $cargoTomlFiles = Get-ChildItem -Path $searchPath -Recurse -Filter 'Cargo.toml' -ErrorAction SilentlyContinue
```
heaths, line 19: I think this is too much of an assumption. Cargo doesn't yet support nested workspaces, but it's being discussed and would solve a number of problems, so when it is supported I'd love to adopt it then this would fail. Instead, if you want to get all packages, perhaps `cargo metadata` would be better.

```pwsh
$packages = cargo metadata --format-version 1 | convertfrom-json | select -expand packages | where { $_.manifest_path.StartsWith((join-path $PWD 'sdk')) }
foreach ($package in $packages) {
    $manifest = cargo read-manifest --manifest-path $package.manifest_path
    # ...
}
```

heaths (changes requested):

eng/pipelines/templates/jobs/ci.yml
hallipr: I can expand it, but the matrix gets really verbose:

```yaml
  - name: BuildMatrix
    type: object
    default:
      - pool:
          os: linux
          name: $(LINUXPOOL)
          image: $(LINUXVMIMAGE)
        Toolchains:
          - name: stable
            publishArtifacts: true
          - name: msrv
          - name: nightly
      - pool:
          os: windows
          name: $(WINDOWSPOOL)
          image: $(WINDOWSVMIMAGE)
        Toolchains:
          - name: stable
          - name: msrv
          - name: nightly
      - pool:
          os: macOS
          name: $(MACPOOL)
          vmImage: $(MACVMIMAGE)
        Toolchains:
          - name: stable
          - name: msrv
          - name: nightly
```

eng/pipelines/templates/steps/use-rust.yml
hallipr: The variable was still called RustToolchain.  I'll replace RustToolchain with Toolchain across the repo

eng/pipelines/templates/jobs/ci.yml
heaths: Would it be possible to set a variable then - something more obvious to change? GitHub Actions seems more capable in this instance, unfortunately.

eng/pipelines/templates/jobs/ci.yml
hallipr: @heaths, it's used for package targeting in the build.   It get overridden in PRs with a package list that's based on the discovery in Language-Settings.ps1 and the pr diff.  In a manual or CI run where there's no changeset driving the package selection, we can still use this to filter packages

hallipr: I've changed the default to `*`

eng/pipelines/templates/jobs/ci.yml
```diff
@@ -0,0 +1,113 @@
...
+    type: boolean
+    default: false
+  - name: BuildMatrix
+    type: object
+    default:
+      - pool:
+          os: linux
+          name: $(LINUXPOOL)
+          image: $(LINUXVMIMAGE)
+        RustToolchains: [stable, msrv, nightly]
```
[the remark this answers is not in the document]
hallipr: Replaced everywhere

eng/pipelines/templates/jobs/ci.yml
hallipr: We can't make variable based as we use it in conditional yaml templating, so we're stuck with compile time parameters and expressions.

I just expanded the matrix and changed the parameter to:
```yaml
 PublishArtifacts: ${{ toolchain.publish }}
 ```

File: eng/pipelines/templates/steps/analyze.yml (hunk 4 above)
Comment (hallipr) [reply eng/pipelines/templates/steps/analyze.yml]: done

File: eng/pipelines/templates/steps/analyze.yml (hunk 5 above)
Comment (hallipr) [reply eng/pipelines/templates/steps/analyze.yml]: added

File: eng/scripts/Language-Settings.ps1 (hunk 7 above)
Comment (hallipr) [reply eng/scripts/Language-Settings.ps1]: I updated to use `cargo metadata`

File: eng/pipelines/templates/jobs/ci.yml (hunk 2 above)
Comment (hallipr) [reply eng/pipelines/templates/jobs/ci.yml]: After talking to @weshaggard, I've removed all of the Targeting string related code from the pipelines.   The pipelines will use the contents of the package info folder to decide what to build.

File: eng/pipelines/templates/steps/analyze.yml (hunk 5 above)
Comment (hallipr) [reply eng/pipelines/templates/steps/analyze.yml]: Added to test-packages.ps1 as it needs to be run in each affected package

File: eng/pipelines/templates/stages/1es-redirect.yml [hunk 9]
```diff
@@ -0,0 +1,60 @@
...
+      customBuildTags:
+        - 1ES.PT.Tag-refs/tags/canary
+    settings:
+      skipBuildTagsForGitHubPullRequests: true
+    sdl:
+      sourceAnalysisPool:
+        name: azsdk-pool-mms-win-2022-general
+        image: azsdk-pool-mms-win-2022-1espt
+        os: windows
+      eslint:
```
weshaggard, line 44: You might want to add this setting https://github.com/Azure/azure-sdk-for-net/pull/45568 to disable warnings

eng/pipelines/templates/steps/download-package-artifacts.yml
```diff
@@ -0,0 +1,18 @@
+parameters:
+  TargetPath: $(Build.ArtifactStagingDirectory)
+
+steps:
+    - task: DownloadPipelineArtifact@2
+      inputs:
+        artifactName: 'packages_windows'
```
weshaggard, line 7: Do we need OS specific artifacts for rust? If not then you shouldn't need this like python does.

eng/pipelines/templates/steps/test-packages.yml
```diff
@@ -0,0 +1,111 @@
...
+          Toolchain: nightly
+
+  - template: /eng/pipelines/templates/steps/use-rust.yml@self
+    parameters:
+      Toolchain: ${{ parameters.Toolchain }}
+
+  - template: /eng/common/pipelines/templates/steps/set-default-branch.yml@self
+
+  - script: |
+      echo "##vso[build.addbuildtag]Scheduled"
```
weshaggard, line 38: Do you know what we use this for?

eng/pipelines/templates/steps/test-packages.yml
```diff
@@ -0,0 +1,111 @@
...
+    displayName: "Tag scheduled builds"
+    condition: and(eq(variables['Build.SourceBranchName'], variables['DefaultBranch']), eq(variables['Build.Reason'],'Schedule'))
+
+  # now we need to call Save-Package-Properties so that we can filter on it
+  - ${{ if eq(variables['Build.Reason'], 'PullRequest') }}:
```
weshaggard, line 43: @scbedd @hallipr we might want to consider adding this logic to eng/common.

eng/pipelines/templates/stages/1es-redirect.yml
hallipr: added

eng/scripts/Language-Settings.ps1
```diff
@@ -0,0 +1,93 @@
+$Language = "rust"
+$LanguageDisplayName = "Rust"
+$PackageRepository = "crates.io"
+$packagePattern = "Cargo.toml"
+$MetadataUri = "https://raw.githubusercontent.com/Azure/azure-sdk/main/_data/releases/latest/rust-packages.csv"
```
weshaggard, line 5: This doesn't exist yet but probably won't be needed until we start working on docs.

eng/scripts/Language-Settings.ps1
```diff
@@ -0,0 +1,93 @@
...
+$GithubUri = "https://github.com/Azure/azure-sdk-for-rust"
+$PackageRepositoryUri = "https://crates.io/crates"
+
+function Get-AllPackageInfoFromRepo ([string] $ServiceDirectory) {
```
weshaggard, line 9:
```suggestion
function Get-AllPackageInfoFromRepo ([string] $ServiceDirectory) 
{
```
nit: I generally like `{` on a newline for larger blocks.

eng/scripts/Language-Settings.ps1
```diff
@@ -0,0 +1,93 @@
...
+    $allPackageProps = @()
+    Push-Location $RepoRoot
+    try {
```
weshaggard, line 12: Majority of our PS scripts are 2 space indentions, I'd suggest that here as well.

eng/scripts/Language-Settings.ps1
```diff
@@ -0,0 +1,93 @@
...
+        foreach ($package in $packages) {
+            if ($package.manifest_path -replace '\\', '/' -match '/sdk/([^/]+)/') {
+                $package.ServiceDirectoryName = $Matches[1]
+            }
+            else {
+                # ignore manifests that are not in a service directory
+                continue
+            }
+
+            $package.RelativePath = (Split-Path $package.manifest_path -Parent).Replace($RepoRoot, "").SubString(1)
```
weshaggard, line 34: Why the Substring(1)? What is the first char?

eng/pipelines/templates/steps/test-packages.yml
hallipr: Agreed.   It's a big chunk of reusable yaml

eng/scripts/Language-Settings.ps1
hallipr: @weshaggard, should I drop that line or make an empty file?

weshaggard: I would at least comment this out until a time it is consumed.

eng/scripts/Language-Settings.ps1
hallipr: it's a path separator

`"/mnt/vss/_work/1/s/sdk/identity/azure_identity".Replace("/mnt/vss/_work/1/s", "")` => `"/sdk/identity/azure_identity"`

We could also use `.TrimStart("\/")`

eng/scripts/Language-Settings.ps1
hallipr: I've updated .editorconfig to use 2 spaces for powershell

eng/scripts/Language-Settings.ps1
hallipr: With autoformat on save turned on, this gets reverted every time the file's edited.   The options for braces aren't granular enough to have it use "next line" for function definitions and "same line" for other blocks

eng/pipelines/pr.yml
```diff
@@ -0,0 +1,12 @@
+trigger: none
+
+pr:
+  branches:
+    include:
+      - feature/track2
```
weshaggard, line 6: We should probably go ahead and add the other normal branch patterns here as well. They shouldn't trigger on main until this is in main so I think it should be safe.

eng/pipelines/pr.yml
```diff
@@ -0,0 +1,12 @@
...
+
+extends:
+  template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
+  parameters:
+    TestProxy: true
```
weshaggard, line 11: Perhaps we should set these as the default in this repo and not need to set them here.

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -0,0 +1,21 @@
...
+    type: string
+    default: stable
+
+steps:
+  - template: /eng/pipelines/templates/steps/use-rust.yml@self
+    parameters:
+      Toolchain: ${{ parameters.Toolchain }}
+
+  - pwsh: |
+      . ./eng/common/scripts/Helpers/CommandInvocation-Helpers.ps1
```
weshaggard, line 12: We should probably put these into a script instead of inline so folks can run it easier locally.

eng/scripts/Language-Settings.ps1
```diff
@@ -0,0 +1,93 @@
...
+      if ($package.manifest_path -replace '\\', '/' -match '/sdk/([^/]+)/') {
+        $package.ServiceDirectoryName = $Matches[1]
+      }
+      else {
+        # ignore manifests that are not in a service directory
+        continue
+      }
+
+      $package.RelativePath = (Split-Path $package.manifest_path -Parent).Replace($RepoRoot, "").SubString(1)
+      $package.DependentPackages = @()
```
weshaggard, line 35: Just keep in mind you will need to update this once Scott's eng/common changes goes through.  https://github.com/Azure/azure-sdk-tools/pull/8878

weshaggard (approved):
Looks good to try and get things in.

eng/scripts/Language-Settings.ps1
heaths: I prefer consistency as well. Don't fight the formatter. It's a losing battle.

eng/pipelines/templates/steps/test-packages.yml
hallipr: I've started a draft pr:
https://github.com/Azure/azure-sdk-tools/pull/8900

eng/scripts/Language-Settings.ps1
hallipr: commented

eng/pipelines/pr.yml
hallipr: Added as:
```yaml
pr:
  branches:
    include:
      - main
      - feature/*
      - hotfix/*
      - release/*
```

eng/pipelines/pr.yml
hallipr: Set `TestProxy: true` as default.   I'm not sure we need the 2 hour test timeout.  Using the existing default of 60 minutes

eng/pipelines/templates/steps/analyze.yml
hallipr: @heaths would prefer devs run the cargo commands directly instead of a script

eng/scripts/Language-Settings.ps1
hallipr: I'll have to manually pull in eng/common.   It'll give me an opportunity to uptake the new property name and callback then

eng/pipelines/templates/steps/analyze.yml
heaths: Yes. It's not helpful to onboard to a project and have to learn unique commands for each repo. We have enough controls over some defaults in the most common cases, but I should just be able to run `cargo test` in any Rust repo and expect it to work for most cases. And if there are optional features, I should expect `cargo test --all-features` to just work. It's the responsibility of a PR build to run all the things we want to check. We hope - and, sadly, it seems not to happen too often - devs are doing some due diligence before pushing commits, but that's the role of the PR build: make sure everything works.

I'd like to see most scripts removed, actually. Most just run common commands with some extra flags that our PR should do anyway, and I'd rather see do inline for the same reason: these are supported, non-breaking command lines that should just work in any repo.

Our other languages don't require scripts for the basics like building and testing. Maybe code generation, doc validation, etc.. Sure, those are all custom things anyway. But tools/frameworks like `dotnet`, `go`, `cargo`, `nppm`, etc. are all opinionated and should be consistent for the most part.

.vscode/settings.json
```diff
@@ -1,5 +1,11 @@
 {
   "cSpell.enabled": true,
   "editor.formatOnSave": true,
-  "rust-analyzer.cargo.features": "all"
+  "rust-analyzer.cargo.features": "all",
+  "[powershell]": {
+    "editor.defaultFormatter": "ms-vscode.powershell",
+  },
+  // the yaml formatter will use complex key notation to wrap long lines
+  // allow really long lines to prevent that.
```
heaths, line 9: I appreciate VSCode supports comments, but GitHub does not and this immediately raises red flags that turn out to be red herrings in code reviews:

image

If the comments don't add any real substance I'd recommend removing them.

eng/pipelines/templates/steps/test-packages.yml
```diff
@@ -0,0 +1,111 @@
...
+  - name: TestTimeoutInMinutes
+    type: number
+    default: 60
+
+steps:
+  # using -Zscript with verify-dependencies.rs requires nightly be installed
+  - ${{ if ne(parameters.Toolchain, 'nightly') }}:
+      - template: /eng/pipelines/templates/steps/use-rust.yml@self
+        parameters:
+          Toolchain: nightly
```
heaths, line 29: Is the result of this that we'll install nightly toolchain into every build? I don't think that's necessary. Just running this script *in* nightly on one agent should be sufficient. It just makes sure that all dependencies are centrally managed in the root `Cargo.toml` much like .NET's `eng/Packages.Data.props`.

heaths (approved):

eng/pipelines/templates/steps/test-packages.yml
hallipr: I made the call in test-packages dependent on `$toolchain -eq 'nightly'` rather than installing nightly on every leg

.vscode/settings.json
hallipr: comment removed

PR #1764, opened 2024-08-21
Add back missing public exports

heaths:
@jhendrixMSFT I compared public docs and think I have them all. I also ended up removing a few types exported directly from `azure_core` that weren't before. Everything should still be in their respective modules, though.


jhendrixMSFT (approved):

jhendrixMSFT:
Can you please also add `Continuable` to the appropriate place? Presumably next to where `NextMarker` is exported.

PR #1767, opened 2024-08-22
AMQP changes to support message headers and properties

LarryOsterman:
Also:
AMQP change to support type serialization; 
Integrated a coupleof fe2o3 bug fixes.

### DIFF
--- a/sdk/core/azure_core_amqp/src/messaging.rs
+++ b/sdk/core/azure_core_amqp/src/messaging.rs
@@ -12,6 +12,16 @@ pub enum TerminusDurability {
     UnsettledState,
 }
 
+impl From<TerminusDurability> for AmqpSymbol {
+    fn from(durability: TerminusDurability) -> Self {
+        match durability {
+            TerminusDurability::None => "none".into(),
+            TerminusDurability::Configuration => "configuration".into(),
+            TerminusDurability::UnsettledState => "unsettled-state".into(),
+        }
+    }
+}
+
 #[derive(Debug, Clone, PartialEq)]
 pub enum TerminusExpiryPolicy {
     LinkDetach,
@@ -20,10 +30,30 @@ pub enum TerminusExpiryPolicy {
     Never,
 }
 
+impl From<TerminusExpiryPolicy> for AmqpSymbol {
+    fn from(policy: TerminusExpiryPolicy) -> Self {
+        match policy {
+            TerminusExpiryPolicy::LinkDetach => "link-detach".into(),
+            TerminusExpiryPolicy::SessionEnd => "session-end".into(),
+            TerminusExpiryPolicy::ConnectionClose => "connection-close".into(),
+            TerminusExpiryPolicy::Never => "never".into(),
+        }
+    }
+}
+
 #[derive(Debug, Clone, PartialEq)]
 pub enum DistributionMode {
     Move,
     Copy,
+}
+
+impl From<DistributionMode> for AmqpSymbol {
+    fn from(mode: DistributionMode) -> Self {
+        match mode {
+            DistributionMode::Move => "move".into(),
+            DistributionMode::Copy => "copy".into(),
+        }
+    }
 }
 
 #[derive(Debug, Clone, PartialEq)]
@@ -92,6 +122,17 @@ impl From<Vec<u8>> for AmqpMessageId {
 impl From<u64> for AmqpMessageId {
     fn from(ulong: u64) -> Self {
         AmqpMessageId::Ulong(ulong)
+    }
+}
+
+impl From<AmqpMessageId> for AmqpValue {
+    fn from(message_id: AmqpMessageId) -> Self {
+        match message_id {
+            AmqpMessageId::String(string) => AmqpValue::String(string),
+            AmqpMessageId::Uuid(uuid) => AmqpValue::Uuid(uuid),
+            AmqpMessageId::Binary(binary) => AmqpValue::Binary(binary),
+            AmqpMessageId::Ulong(ulong) => AmqpValue::ULong(ulong),
+        }
     }
 }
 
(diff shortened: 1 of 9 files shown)


sdk/core/azure_core_amqp/src/messaging.rs
```diff
@@ -188,6 +190,121 @@ impl AmqpMessageHeader {
     pub fn delivery_count(&self) -> Option<&u32> {
         self.delivery_count.as_ref()
     }
+
+    pub fn set_durable(&mut self, durable: bool) {
```
heaths, line 194: Why do you have setters if you also have a builder? We should have one way of doing things.

sdk/core/azure_core_amqp/src/messaging.rs
```diff
@@ -188,6 +190,121 @@ impl AmqpMessageHeader {
...
+            list.push(AmqpValue::Null);
+        }
+        if let Some(first_acquirer) = header.first_acquirer {
+            list.push(AmqpValue::Boolean(first_acquirer));
+        } else if header.delivery_count.is_some() {
+            list.push(AmqpValue::Null);
+        }
+        if let Some(delivery_count) = header.delivery_count {
+            list.push(AmqpValue::UInt(delivery_count));
+        }
```
heaths, line 305: There's a lot of duplication here. Might be cleaner to do something like:

```rust
let list = vec![AmqpValue::Null; 5];
if let Some(durable) = header.durable {
    list[0] = durable;
}
if let Some(priority) = header.priority {
    list[1] = priority;
}
// etc.
```
Also, if `durable` is `None`, don't you need to also push `Null` or is that not how the protocol works?

Working e.g.,
```rust
#[allow(dead_code)]
#[derive(Clone, Debug)]
enum AmqpValue {
    Null,
    Boolean(bool),
    U32(u32),
}

struct Options {
    foo: Option<bool>,
    bar: Option<u32>,
}

fn main() {
    let opts = Options {
        foo: Some(true),
        bar: None,
    };

    let mut list = vec![AmqpValue::Null; 2];
    if let Some(foo) = opts.foo {
        list[0] = AmqpValue::Boolean(foo);
    }
    if let Some(bar) = opts.bar {
        list[1] = AmqpValue::U32(bar);
    }

    println!("{list:#?}");
}
```

heaths: Same - even more so, actually - below.

heaths (changes requested):
I don't understand why there are public setters on options when options exposes a builder. We should have one pattern.

sdk/core/azure_core_amqp/src/messaging.rs
LarryOsterman: If durable is None, you need to push Null onto the array - . And you need to trim all the trailing null fields.

The array is mutable though. I can resize the array to remove all the trailing nulls.

sdk/core/azure_core_amqp/src/messaging.rs
heaths: If the builder pattern doesn't make sense in this case, we can relax requirements. Even as I was reading this, I was wondering why we need the builder pattern here when these aren't client options. The only reason we were favoring client options builders is to avoid the deep nesting we get by, say, `BlobClientOptions` containing ` StorageClientOptions` containing an `azure_core::ClientOptions` containing a `typespec_client_core::ClientOptions`, which looks hideous to type out. It's idiomatic in rust for simple data structures to just make the fields public.

sdk/core/azure_core_amqp/src/messaging.rs
heaths: Ah, so `Null` is terminal? Shouldn't there be a return then to short-circuit the method? Or did I misunderstand?

sdk/core/azure_core_amqp/src/messaging.rs
LarryOsterman: Builder makes sense, but it's slightly annoying - I need to keep a builder and a header both in my Header object (and the other objects) - the builder for the "set" APIs and the header for the "get" APIs. 

Also the builder needs to have a mutable self, and that in turn means I need to have a .clone() in the builder.build() method - again, slightly annoying.

But I'm now wondering about your suggestion of having wondering about the idea of having a "cpp_apis" feature which enables the C++ setters instead. 

While Header has only 5 fields, Properties has like 12 of them so the builder pattern makes sense for it.

sdk/core/azure_core_amqp/src/messaging.rs
LarryOsterman: No, `Null` is explicitly *not* terminal. But each array slot has a particular semantic - for instance, `first_acquirer` is always at index 3 in the array. If `first_acquirer` is the only field, the array will have 4 entries, 3 of them are `Null`, the 4th will be a `boolean` value. The last entry in the array needs to be the last entry with a value.

sdk/core/azure_core_amqp/src/messaging.rs
LarryOsterman: Ok, upon thinking about it, I have a better way of handling this case. I'll have to play some games with the internal semantics of the C++ wrapper but I can make it work (the wrapper APIs mimic the corresponding uAMQP APIs so I'll have some weird variable names but I can make it all work.

sdk/core/azure_core_amqp/src/messaging.rs
LarryOsterman: And I have a much better pattern for this - serialize the whole thing to an array, then scan backwards to the first non Null value and resize the array to remove them.

It makes the conditionals dramatically simpler.

sdk/core/azure_core_amqp/src/messaging.rs
```diff
@@ -586,77 +522,27 @@ impl From<AmqpMessageProperties> for AmqpList {
...
-            || properties.correlation_id.is_some()
-            || properties.content_type.is_some()
-            || properties.content_encoding.is_some()
-            || properties.absolute_expiry_time.is_some()
-            || properties.creation_time.is_some()
-            || properties.group_id.is_some()
-            || properties.group_sequence.is_some()
-            || properties.reply_to_group_id.is_some()
-        {
+        } else {
```
heaths, line 525: Tip: you could still simplify a lot of this conditional goo e.g.,

```rust
#[allow(dead_code)]
#[derive(Clone, Debug)]
enum AmqpValue {
    Null,
    String(String),
    UInt32(u32),
}

#[derive(Debug, Default)]
struct Options {
    pub s: Option<String>,
    pub u: Option<u32>,
}

fn main() {
    let options = Options {
        s: Some("string".to_string()),
        ..Default::default()
    };

    let mut values = vec![AmqpValue::Null; 2];
    values[0] = options
        .s
        .as_ref()
        .map_or_else(|| AmqpValue::Null, |v| AmqpValue::String(v.to_string()));
    values[1] = options
        .u
        .map_or_else(|| AmqpValue::Null, |v| AmqpValue::UInt32(v));

    println!("{options:?}");
}
```

heaths (approved):
A lot of this code could be further simplified, but otherwise LGTM.

heaths (approved):

sdk/core/azure_core_amqp/src/messaging.rs
LarryOsterman: Oh, that is so nide.

One minor refinement:

```rust
values[1] = options.u.map_or(AmqpValue::Null, AmqpValue::Int32);
```

courtesy of GH copilot - testing this version right now, but reading the code, there's no reason it shouldn't work.

heaths: Makes sense. Much like .NET delegates, if the signature matches you just pass in a ~~function~~ method.

LarryOsterman: It's interesting. I think of `AmqpValue::UInt32(v)` as an initializer for `AmqpValue`, but apparently the actual semantics are that it's a function call to the `UInt32` function on `AmqpValue`.

Which is weird, but I guess it works.

sdk/core/azure_core_amqp/src/messaging.rs
```diff
@@ -12,6 +12,16 @@ pub enum TerminusDurability {
     UnsettledState,
 }
 
+impl From<TerminusDurability> for AmqpSymbol {
+    fn from(durability: TerminusDurability) -> Self {
+        match durability {
+            TerminusDurability::None => "none".into(),
```
heaths, line 18: Nit: I'm assuming this then calls into code that converts a `&str` into an `AmqpSymbol::String` or something like that? I would skip the automagical conversion here and be explicit. One would hope the compiler or linker would be smart enough to elide the extra hops, but seems more efficient to go straight into the appropriate variant.

heaths (approved):

sdk/core/azure_core_amqp/src/messaging.rs
LarryOsterman: That makes a ton of sense. Fixing.

heaths:
Be sure to rebase on `feature/track2` and that should fix the build errors (without having looked, I admit) you're hitting.

PR #1771, opened 2024-08-28
add a shell.nix to ease building on nixos

analogrelay:
@heaths we've had some conversations about NixOS and assorted development environment needs. I'm still totally game for keeping things super simple, but I have run into a snag. We use `openssl-sys` transitively (through `ring`, I believe). That gives us a build-time dependency on openssl **and** requires that it be available through `pkg-config`. In NixOS, the way to do that is through `nix-shell`, as even globally installing `openssl` does not add it to `pkg-config`. So, anyone building this repo on NixOS will have to run `nix-shell -p pkg-config openssl` in order to build the repo. We can simplify that a bit by adding our own `shell.nix` that specifies the native dependencies, allowing a user to run `nix-shell` alone to automatically enter the recommended shell environment for our repo.

In addition, I've added an `.envrc` which allows a user who uses [direnv](https://github.com/direnv/direnv) to automatically enter `nix-shell` when they change directory into the repo.

This is _purely optional_, it doesn't require other developers in this repo to use Nix, but it helps any developer who _does_ use Nix by having a clear specification of our native dependencies.

I'm OK if you'd rather not have this in the repo. I fully recognize I have become a 🤓 Full NixOS Nerd 🤓 and have to live with the consequences of my actions 😅. However, I do think there's some value on specifying our dependencies in a declarative and programmatic way. If it were just `rustup`, I wouldn't feel a need for this, but as long as we have dependencies _beyond_ the main language toolchain, I think this has value.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,2 @@
+# Automatically enter 'nix-shell' when using direnv, and nix.
+use nix
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ Cargo.lock
 *.swo
 *.swp
 *.vim
+
+# Ignore .direnv, which is where the direnv environment is built, if a user is using direnv.
+.direnv/
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,12 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+pkgs.mkShell {
+    nativeBuildInputs = [
+        pkgs.pkg-config
+    ];
+
+    buildInputs = [
+        pkgs.rustup
+        pkgs.openssl
+    ];
+}


heaths:
I'm leaning toward not taking this because of the friction to onboarding, but I do appreciate it helps pin dependencies on native libraries. I will discuss with @RickWinter, @JeffreyRichter, @weshaggard, @hallipr, et. al. My own experience with it - just jumping into helix-editor/helix - has been a pain and forced me to have to learn a bunch of stuff just to contribute what would've otherwise been pretty simple. I get it's optional, but you get different results whether you use it or not. And, yes, I appreciate the irony that using NixOS is to reduce/eliminate those discrepancies. 😉

heaths:
I'd also be curious how this companies to devcontainers, which we do use/support in some other repos already. Conceptually, they are the same but it seems you get more by default (depending on base image, of course) in most devcontainers. If they get used heavily, we can also regularly build and cache devcontainers so they are "hot" and ready to use by contributors.

analogrelay:
> because of the friction to onboarding

I do want to be clear that the intent of _this_ PR is **not** to require that contributors and developers use Nix and `nix-shell`. It is only to provide that _option_ to people who do want to use Nix to prepare any native dependencies automatically. If we have native dependencies required to build (and whether we should is definitely a valid question to ask), developers need to install them anyway.

As it is, this repo requires installing a C compiler, `pkg-config`, and `libssl-dev` on a regular Ubuntu WSL instance (as-is, from the Microsoft Store). I'm totally fine having a README.md section describing these pre-requisites, a dev container with them preinstalled, etc. I just figured it made sense to provide an option for anyone using Nix (which is not entirely uncommon in the Rust ecosystem, it's certainly something I've seen in Rust repos more often than other languages).

I don't really have a strong desire for Nix to be the sole mechanism we use to provide development environments though, it's definitely powerful but the learning curve is too steep for us to require it.

heaths:
Oh, I get you don't want to require it. No worries there. But we've also been investing some into devcontainers, which then means we have duplicate forms of configuration-as-a-vm to maintain. Devcontainers support GitHub Codespaces and you can even drive your CIs from them. Seems nix would require some outbound channel to monitor progress should we use it for build, though maybe that's not required.

analogrelay:
Going to try a different approach here and use a codespace instead.

PR #1772, opened 2024-08-28
Derive Macro for `typespec_client_core::http::Model`

analogrelay:
This PR adds a new `typespec_derive` crate. This crate provides a derive macro for the `Model` type that replaces the `json_model` and `xml_model` declarative macros I added in #1724 . The generated code is essentially the same. Instead of two separate macros to choose formats, you can use `#[typespec(format = "xml")]` to change the format to `xml` (and other formats in the future. Primarily to support testing inside the crate, I also added support for an attribute to set the crate that contains `Model`, [based on serde's equivalent](https://serde.rs/container-attrs.html#crate).

I was going to have the `azure_core` crate re-exports this macro, like it re-exports the `Model` trait. However, there's a problem. You need to reference `typespec_client_core` in order to get the `Model` trait in scope. We re-export that trait from `azure_core` but the generated code _won't use that_, it'll use `typespec_client_core`. That can be overridden with a `#[typespec(crate = "azure_core")]` attribute, but that would require clients that don't reference `typespec_client_core` directly to put that attribute on every model type. 

Instead, I'm re-exporting it from `typespec_client_core`. This means that service client crates now need to add a reference to `typespec_client_core` directly, but I think that's reasonable. It's not like crates can avoid being aware of typespec, or it's related libraries.

There are some alternative approaches we could take, but I'm not a huge fan of those. For example, we could make the macro reference `Model` from `azure_core` and move it to an `azure_derive` crate, but that means it's no longer available in the unbranded libraries. Or we could do something even funkier and find a way to "re-wrap" the proc_macro defined in `typespec_derive` in a new macro in `azure_derive` that sets the crate. That feels really clunky and difficult to achieve. So, I've stuck with this approach for now.


heaths:
Do we even need to re-export it from `typespec_client_core`? I do that with `Error` and friends at https://github.com/Azure/azure-sdk-for-rust/blob/c1065e80b04e831d84956cffdf77404c87dca26b/sdk/typespec/typespec_client_core/src/error/mod.rs#L11 but I've been rethinking that. Is there any advantage? `typespec` is a dependency and, conceptually, a package deal with `typespec_client_core` already. The idea for re-exporting from `azure_core` was trying to hide the "typespec"iness, though rust-analyzer seems to know better when suggesting to import types.

@analogrelay, @RickWinter, @JeffreyRichter, @LarryOsterman thoughts about re-exporting `typespec` types from `typespec_client_core`? I'm leaning against it since you're already in "typespec" land and it's just more to maintain. I still think - for now, at least - we re-export types moved from `azure_core` *from* `azure_core`, but I at least want to move them around like with `typespec` and `typespec_client_core` so we better segment protocols and have fewer submodules people have to hunt through.

analogrelay:
> Do we even need to re-export it from typespec_client_core?

We don't need to, but it does provide one advantage. I've noticed that _generally_ when you see a `derive`able trait at path `foo::bar::SomeTrait`, the derive macro is _also_ placed at the same path. This is valid because the derive macro is actually called something like `model_derive` and the ability to call it as `#[derive(Model)]` is syntactic sugar. This makes it so you only need to import the one "symbol" and you get both the trait **and** the derive macro in scope:

```rust
use serde::Deserialize;
use typespec_client_core::http::Model;

#[derive(Model, Deserialize)]
pub struct MyModel {}
```

If we don't re-export the derive macro _next_ to the trait (which is actually where I started in an earlier iteration of this branch), then you have to import them separately. Now, if you're just deriving the trait, you don't actually have to import the trait, so you can do this:

```rust
use serde::Deserialize
use typespec_derive::Model;

#[derive(Model, Deserialize)]
pub struct MyModel {}
```

If you later want to use `Model` as a trait, you end up with some duplication (which the compiler should handle fine but looks odd):

```rust
use serde::Deserialize
use typespec_derive::Model;
use typespec_client_core::http::Model;

#[derive(Model, Deserialize)]
pub struct MyModel {}

fn do_something(m: impl Model) {}
```

The trick is that the symbol `Model` in `#[derive(Model)]` needs to resolve to the _derive macro_, **not** the trait.

> thoughts about re-exporting typespec types from typespec_client_core?

In general, I'm against re-exporting unless there are specific reasons to do it. But I've also lived a life forever tainted by having to work on NuGet dependency algorithms, so I like to avoid hiding dependencies as much as possible 😅 .

I will say, the layering felt backwards to me when I saw it. I initially assumed that `typespec` was a **superset** of `typespec_client_core` and contained and re-exported all the types from that (and a future `typespec_server_core`?). So, there's definitely some potential for confusion there already.

My inclination is to start with as few re-exports as possible and see what customers think. It's generally an additive change to re-export a type elsewhere after you've shipped it, but a breaking change to stop re-exporting a type.

heaths:
`typespec` is the "core" because:

1. It will serve as the root namespace when <https://rust-lang.github.io/rfcs/3243-packages-as-optional-namespaces.html> is fully implemented.
2. We may one day have a `typespec_server_core` for writing services.

As for re-exports, I may just end up removing most. I still want to keep them in `azure_core` for now, though, even if rearranged after the first beta (no point in breaking things now). I'm still holding onto hope that, to @RickWinter's pleasure, I can figure out someway to get rust-analyzer to recommend, e.g., using `azure_core::Result` instead of `typespec::error::Result` or even `typespec::Result`.

It does make me want to revisit preludes. We opted against them since they seem to be falling out of favor. Probably won't change our minds, though.

analogrelay:
My personal preference here would be to just drop the re-exports here and have client code reference `typespec_client::Response`, `typespec_client::Model`, etc. but I'm open to whatever. I do think re-exporting the **derive macro** is valuable though, for the reasons I described above. Procedural Macros are **required** to be in a separate crate, so we're forced to put it in `typespec_derive`, but as I noted above I think it's very helpful to have the macro made available at the same path as the trait.

sdk/typespec/typespec_derive/src/lib.rs
```diff
@@ -0,0 +1,205 @@
+extern crate proc_macro;
+
+use proc_macro2::{Span, TokenStream};
+use syn::parse::ParseStream;
+use syn::spanned::Spanned;
+use syn::{Attribute, DeriveInput, LitStr, Meta, Path};
+
+/// Derive macro for implementing `Model` trait.
+/// Deriving this trait allows a type to be deserialized from an HTTP response body.
```
heaths, line 9: Nit: I'd put a blank line in here or it all gets mushed together and harder to read.

There's a lint coming soon to nightly that will complain about long "first lines" as well.

sdk/typespec/typespec_derive/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+/// By default, the type must also implement `serde::Deserialize`, or the generated code will not compile.
```
heaths, line 10: My idea for a derive was to actually do this, or do we want to push any customization of `#[serde(...)]` attributes to explicit decisions? I was going to allow it anyway, but maybe making devs explicitly derive `Deserialize` (and/or `Serialize`, which you may also want to mention) makes it more obvious of what's expected. Thoughts?

sdk/typespec/typespec_derive/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+/// # use typespec_derive::Model;
+/// # use serde::Deserialize;
+/// extern crate typespec_client_core as my_typespec;
+///
+/// #[derive(Model, Deserialize)]
+/// #[typespec(crate = "my_typespec")]
+/// struct MyModel {
+///   value: String
+/// }
+/// ```
```
heaths, line 30: Nit: I expect this will get used less. I imagine you're alphabetizing - I know I like to most times :smile: - but I'd mention the `format` first since it's likely going to get used more...and still probably not that much.

sdk/typespec/typespec_derive/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+///
+/// **NOTE:** Using formats other than JSON may require enabling additional features in `typespec_client_core`.
+#[proc_macro_derive(Model, attributes(typespec))]
+pub fn model_derive(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
+    let ast: DeriveInput = syn::parse(input).unwrap();
+    let body = generate_body(ast);
```
heaths, line 51: I eventually want to add more derive macros (lower pri right now, but like for extensible enums that have a lot of boiler plate), so I'd use a function name indicative of the derive macro e.g., `generate_model` or `model_derive_impl`. I also recommend considering putting it in a separate module since, when we add more macros, this `lib.rs` will become monolithic. Makes it easier to segment specific derives. I know serde et. al. do that...granted, they have much larger impls as well.

sdk/typespec/typespec_derive/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+            body.xml().await
+        },
+    };
+
+    // If the standard path is used, we need to add 'extern crate', because it's possible the calling code
+    // depends on typespec_client_core transitively, which means it's not in scope by default.
+    // That's not necessary when using a custom path because we assume the user has done that work.
+    let typespec_import = match attrs.typespec_path {
+        Some(path) => quote::quote! {
+            use #path as _typespec;
```
heaths, line 93: Because `typespec` is actually a separate crate, should this maybe be `_typespec_client_core` to disambiguate in the future?

heaths: Same below and elsewhere as needed.

sdk/typespec/typespec_derive/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+}
+
+enum Format {
+    Json,
+    Xml,
+}
+
+struct Context {
```
heaths, line 118: Nit: only because this came up today in a discussion, using "Context" - even though private - might cause confusion with the public `azure_core::Context` (well, as re-exported). Honestly, I don't mind this, but having just merged a change for #1778 it might be worth considering a less ambiguous name e.g., `DeriveContext` (if reusable for other derive macros, which it seems it would be).

sdk/typespec/typespec_derive/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+        let token_stream = quote::quote_spanned!(span=> compile_error!(#message));
+        self.errors.push(token_stream);
+    }
+
+    pub fn errors(&self) -> &[TokenStream] {
+        &self.errors
+    }
+}
+
+struct Attrs {
```
heaths, line 137: Same here. These should be `Model`-specific, but moving them to a submodule fixes that.

sdk/typespec/typespec_derive/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+            };
+
+            meta_list.parse_nested_meta(|meta| {
+                let Some(ident) = meta.path.get_ident() else {
+                    ctx.emit_error(meta.path.span(), "Invalid typespec attribute, expected attribute in form #[typespec(key = value)]");
+                    return Ok(());
+                };
+                match ident.to_string().as_str() {
+                    "crate" => {
+                        let value = meta.value().unwrap();
```
heaths, line 159: How likely is this to panic? Some unwraps in macros isn't rare, but seems a better error message is more informative if this fails less than rarely. My experience writing macros hasn't given me enough evidence to base any informed opinion here, so merely raising the question.

sdk/typespec/typespec_derive/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+            }).unwrap();
+        }
+        Attrs {
+            typespec_path,
+            format,
+        }
+    }
+}
+
+fn parse_literal_string(value: ParseStream) -> Result<LitStr, ()> {
```
heaths, line 196: FWIW, helpers like this can and should stay in `lib.rs`. Same for `Context` above.

sdk/typespec/typespec_client_core/Cargo.toml
```diff
@@ -49,6 +51,7 @@ default = [
   "reqwest_rustls",
   "tokio_sleep",
 ]
+derive = ["typespec_derive"]
```
heaths, line 54:
```suggestion
derive = ["dep:typespec_derive"]
```

Just to make sure no implicit `typespec_derive` feature is created, or I may be misunderstanding the docs about this. Still, I've been trying to be consistent to avoid confusion.

sdk/typespec/typespec_client_core/src/http/response.rs
```diff
@@ -114,21 +91,31 @@ impl<T> Response<T> {
...
     ///    value: String,
     /// }
-    /// typespec_client_core::json_model!(MySecretResponse); // Mark the type as deserializable from JSON.
     ///
     /// async fn parse_response(response: Response<GetSecretResponse>) {
-    ///   // Calling `map` will parse the body into `MySecretResponse` instead of `GetSecretResponse`.
+    ///   // Calling `deserialize_body_into` will parse the body into `MySecretResponse` instead of `GetSecretResponse`.
     ///   let my_struct: MySecretResponse = response.deserialize_body_into().await.unwrap();
-    ///   println!("value: {}", my_struct.value);
+    ///   assert_eq!("hunter2", my_struct.value);
```
heaths, line 107: Why are you checking in a string with "******"?! 😆

sdk/typespec/typespec_client_core/src/http/response.rs
```diff
@@ -156,12 +143,14 @@ impl<T: Model> Response<T> {
     ///
     /// # Example
     /// ```rust
-    /// # #[derive(serde::Deserialize)]
+    /// # use serde::Deserialize;
+    /// # use typespec_client_core::http::Model;
+    /// # use typespec_derive::Model;
```
heaths, line 148: Shouldn't we make importing the first automatically import the second like serde et. al.? This would make it much easier to consume.

sdk/identity/azure_identity/src/refresh_token.rs
```diff
@@ -71,7 +72,6 @@ pub struct RefreshTokenResponse {
     access_token: Secret,
     refresh_token: Secret,
 }
-azure_core::json_model!(RefreshTokenResponse);
```
heaths, line 74: Rebase on `feature/track2`. There's at least one other one that needs to be updated.

heaths: Come to think of it, should this PR just remove the old macros entirely?

heaths (changes requested):
This is looking great but while I see nothing truly blocking, I expect enough changes I want to look at this again.

sdk/typespec/typespec_derive/src/lib.rs
analogrelay: Yep, I agree. I wasn't even conciously alphabetizing, I just had to implement `crate` first to get tests to compile before I even started on `format`. I agree that `format` should come first.

sdk/typespec/typespec_derive/src/lib.rs
analogrelay: Done. The proc macro function is required to live in the root of the crate, but I can organize it into an inner function in a trait-specific module that actually does all the work.

sdk/typespec/typespec_derive/src/lib.rs
analogrelay: Yeah, I actually meant to refactor to remove this (replacing it with returning `Result<TokenStream, Vec<Error>>` so that things were more "functional". I've done that and Context is no more.

sdk/identity/azure_identity/src/refresh_token.rs
analogrelay: It does, but this usage was floating around still.

sdk/typespec/typespec_derive/src/lib.rs
heaths: Yeah, and that's what serde does as well. Their `src/lib.rs` is basically just calling all the functions in separate modules. Nice and neat.

sdk/typespec/typespec_client_core/src/http/response.rs
analogrelay: Yep, this predates re-exporting the macro from `typespec_client_core`. Fixed now.

analogrelay:
I believe all the feedback has been resolved. In addition, I've added tests that validate compilation errors. I was originally looking at [compiletest_rs](https://github.com/Manishearth/compiletest-rs) for this (which is an extraction of the [compiletest](https://rustc-dev-guide.rust-lang.org/tests/compiletest.html) utility used in rustc itself to test errors) but ran into some issues getting dependencies right. I realized it was actually fairly easy to just have a Cargo project that produced all the expected errors and then write a standard integration test that ran `cargo build` and validated the results against baseline files we check in.

heaths:
There are some build failures. Rebase on feature/track2 as well, which should fix one of them that @hallipr fixed and I merged just now. You can also update `.vscode/cspell.json` from within VSCode or easy enough manually. Or inline it with `// cspell:ignore <word>` if not universal. VSCode has a cSpell extension I highly recommend installing, especially if you write documentation. learn.microsoft.com requires it, for example.

Cargo.toml
```diff
@@ -92,6 +96,7 @@ tokio = { version = "1.0", default-features = false, features = [
 tracing = "0.1.40"
 tracing-subscriber = "0.3"
 tz-rs = { version = "0.6" }
+typespec_derive = { path = "sdk/typespec/typespec_derive" }
```
heaths, line 99: Let's add this up around line 30 so it's with the other first-party (in-repo) dependencies for easy reference.

sdk/typespec/typespec_derive/compilation-tests/src/model/bad_attributes.rs
heaths, line 1: I think this is fine and not advocating you necessarily get rid of it, but any reason to not just use unit tests against the actual implementing function? I've done that and basically take for granted - which I think is perfectly acceptable in this case - that the actual derive function has the right signature and calls the right function. Apart from that, it doesn't really do anything. So we can still test without getting creative like this. See <https://github.com/Azure/azure-sdk-for-rust/pull/1690/files#diff-e6b4f89ebef847a34eb56db461e3c6868b73dc4be117fa1a58bedb9407344393R101> as an example. This works with normal `cargo test`.

sdk/typespec/typespec_derive/src/lib.rs
```diff
@@ -0,0 +1,82 @@
+use syn::{parse::ParseStream, parse_macro_input, spanned::Spanned, DeriveInput, Error, LitStr};
```
heaths, line 1: Copyright headers on all sources per corporate policy. `.editorconfig` has something, but seems we don't enforce it (it's non-standard but common `.editorconfig`).

heaths (approved):

sdk/typespec/typespec_derive/compilation-tests/src/model/bad_attributes.rs
analogrelay: I did look at this, but it can be very finicky to assert `TokenStream` results, particularly for errors. Producing an error from a proc macro is a little bit of a hack. We can convert a `TokenStream` to a string and assert that, but the [result is not guaranteed to be stable](https://doc.rust-lang.org/stable/proc_macro/struct.TokenStream.html#impl-Display-for-TokenStream).

Plus, the way we handle compilation errors can be a little tricky to test. There's no officially-supported way to generate errors spanning certain source text in proc macros. Instead, `syn` uses [a bit of a hack](https://docs.rs/syn/latest/syn/struct.Error.html#error-reporting-in-proc-macros) that works well but again it's difficult to assert the `Span` associated with a `TokenStream`.

Most guidance I found for testing proc macros indicated that the most reliable way to test error results is to run the compiler and assert the results so that's what I went with here. I'm open to re-evaluating that though.

sdk/typespec/typespec_derive/src/lib.rs
analogrelay: Hmm... wonder if there's a VS Code extension that could do that automatically 🤔

sdk/typespec/typespec_derive/compilation-tests/src/model/bad_attributes.rs
analogrelay: To be fair, the main problem I was having with unit testing was around asserting the `Span`s, and I had to simplify my current approach down to just asserting line numbers because of some compiler version inconsistencies. So maybe it's reasonable to go back to the unit testing approach 🤔 .

sdk/typespec/typespec_client_core/src/http/response.rs
analogrelay: Ah, I remember. This needs to be here because we don't build doctests with the `derive` feature. Here, it's hidden from the docs. I can change the `deserialize_body_into` doc comment to hide that derive and "fake" using the `derive` feature.

sdk/typespec/typespec_derive/src/lib.rs
analogrelay: Excellent catch, this fails when `#[typespec(whatever` is not followed by an `=` token. So, for example, `#[typespec(format("json"))]` (a potential misunderstanding of our attribute syntax) could cause it and we can provide a better error. I've updated this code and added an error test. The `value` is needed in both arms of the match, so I've gone ahead and promoted it above the `match`. That does change the behavior if you have both a malformed attribute, **and** reference an unknown attribute name (for example `#[typespec(not_something_we_support("bar"))]` would previously have errored with `unknown typespec attribute` and will now error with `invalid typespec attribute`). That seems reasonable to me though.

sdk/typespec/typespec_derive/compilation-tests/src/model/bad_attributes.rs
analogrelay: Leaving this as-is given some of the constraints I listed above. Happy to revisit this later! I would _much_ prefer to be able to call the proc macro func directly from a unit test, if we can figure out a good way to validate the result.

sdk/typespec/typespec_derive/tests/compilation-tests.rs
```diff
@@ -0,0 +1,176 @@
...
+#[derive(Serialize)]
+#[serde(transparent)]
+struct FileResult {
+    pub messages: Vec<MessageExpectation>,
+}
+
+#[test]
+pub fn compilation_tests() {
+    let test_root = {
+        let mut cwd = std::env::current_dir().expect("failed to get current directory");
```
heaths, line 32: Is this documented - stable - that the PWD during `cargo test` is the parent directory? Otherwise, you might consider using the `file!()` macro here instead.

As for the "compilation-tests" directory, to avoid confusion that it's an actual crate, could we put it in "tests/data"? We have a fairly common pattern across many languages of this and could add blanket exemptions for directories of that pattern for various things.

analogrelay:
I'm hitting a strange difference between the GitHub and AzDO builds. The GitHub build uses `cargo test --all` to run tests (as a reminder, `--all` is **not** the same as `--all-features`, it is equivalent to `--workspace` and builds all packages in the workspace, from here on I'll use `--workspace` to make that distinction clear). The AzDO build iterates through each package and runs `cargo test --doc` on each.

I don't understand why, but there's a difference in enabled features between these two options when running `rustdoc`. When running doctests through `cargo test --workspace` the rustdoc invocation looks like this (irrelevant options removed):

```
   Doc-tests typespec_client_core
     Running `/home/ashleyst/.rustup/toolchains/1.76.0-x86_64-unknown-linux-gnu/bin/rustdoc ... --cfg 'feature="default"' --cfg 'feature="derive"' --cfg 'feature="http"' --cfg 'feature="json"' --cfg 'feature="reqwest"' --cfg 'feature="reqwest_gzip"' --cfg 'feature="reqwest_rustls"' --cfg 'feature="tokio"' --cfg 'feature="tokio_sleep"' --cfg 'feature="xml"' --error-format human`
```

Note that `--cfg 'feature="derive"'` is **present**

When building the package directly, without `--workspace`, the rustdoc invocation looks like this (again, irrelevant options removed):

```
   Doc-tests typespec_client_core
     Running `/home/ashleyst/.rustup/toolchains/1.76.0-x86_64-unknown-linux-gnu/bin/rustdoc --cfg 'feature="default"' --cfg 'feature="http"' --cfg 'feature="json"' --cfg 'feature="reqwest"' --cfg 'feature="reqwest_gzip"' --cfg 'feature="reqwest_rustls"' --cfg 'feature="tokio"' --cfg 'feature="tokio_sleep"' --error-format human`
```

Note that this time `--cfg 'feature="derive"'` is **absent**.

This impacts the doctests for `Response::deserialize_body` and `Response::deserialize_body_into`. Those tests use the `Model` trait **and** the derive macro. When built **with** the `derive` feature, they can import `use typespec_client_core::http::Model` and get both the trait and the derive macro. When built **without** the `derive` feature, they have to **also** import `use typespec_derive::Model` to get the derive macro.

Unfortunately, if built **with** the `derive` feature, they **cannot** import the derive macro because it creates an ambiguity (the derive macro is imported from both `typespec_client_core` **and** `typespec_derive`).

For now, what I've done is just to fix the doctest. I can conditionally import `typespec_derive::Model` **only** when the `derive` feature is not present, which fixes the build in both cases. But I'm still unclear as to why the builds have different sets of features.

sdk/typespec/typespec_derive/tests/compilation-tests.rs
analogrelay: I changed this to use `CARGO_MANIFEST_DIR`, which is more stable and gives us what we want.

PR #1773, opened 2024-08-28
Initial CosmosDB crate

analogrelay:
This PR creates a new crate for the Azure Cosmos DB client in the track2 branch. It includes only a single operation, the [Get Database](https://learn.microsoft.com/en-us/rest/api/cosmos-db/get-a-database) operation which returns the "properties" of a database, primarily its ID and other "system" properties (like `_etag`, `_self`, `_rid`, and `_ts`).

There's a single sample, `cosmosdb_connect` (following the [convention for prefixing examples with the service name](https://learn.microsoft.com/en-us/rest/api/cosmos-db/get-a-database)). It can be run with the following command in the root of the repo (hence why the service name prefix is useful, example names are a "global" scope for the repo):

```bash
$ az login
... do your Azure Login ...
$ cargo run --example cosmosdb_connect -- https://someaccount.documents.azure.com:443/ SomeDB
   Compiling azure_data_cosmos v0.19.0 (/home/ashleyst/code/Azure/azure-sdk-for-rust/sdk/cosmosdb/azure_data_cosmos)
    Finished dev [unoptimized + debuginfo] target(s) in 3.99s
     Running `target/debug/examples/cosmosdb_connect 'https://ashleyst-e2e-sql3.documents.azure.com:443/' SampleDB`
DatabaseProperties { id: "SomeDB", system_properties: SystemProperties { etag: Some(Etag("\"0000fe01-0000-4d00-0000-66b64a2d0000\"")), self_link: Some("dbs/aaaMAA==/"), resource_id: Some("aaaMAA=="), last_modified: Some(CosmosTimestamp(1723222573)) } }
```

This is a **draft** PR, because there are some things I intend to do before this is fully ready for review. However, there are some core elements I want to call out and get initial feedback on. I've made inline comments to point those out.

## TODO

These are the things I already intend to do, so if you notice they're missing, hang tight for the next iteration ;).

* [x] Improve doc comments. I want to try and keep on top of those, so I need to add a full set of doc comments to `CosmosClient` and friends.
* [x] Rebase on `feature/track2` when #1772 is merged, and replace use of `json_model!` with the derive macro.


.github/CODEOWNERS
```diff
@@ -31,6 +31,12 @@
 # PRLabel: %Storage
 /sdk/storage/           @heaths @RickWinter @JeffreyRichter
 
+# AzureSDKOwners: @heaths
+# ServiceOwner: @Pilchie
+# ServiceLabel: %Cosmos
+# PRLabel: %Cosmos
+/sdk/cosmosdb/          @heaths @RickWinter @analogrelay @Pilchie
```
analogrelay, line 38: @Pilchie is there anyone else from our side to include here?

@heaths let me know if this is the right set of folks for you and if I have any other issues with the various tags (like `AzureSDKOwners`).

sdk/cosmosdb/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -0,0 +1,81 @@
...
+pub struct CosmosClient {
+    endpoint: Url,
+    pub(crate) pipeline: Pipeline,
+
+    #[allow(dead_code)]
+    options: CosmosClientOptions,
+}
+
+pub trait CosmosClientMethods {
+    fn database(&self, name: impl Into<String>) -> DatabaseClient;
```
analogrelay, line 21: Using an `impl` type here makes `CosmosClientMethods` **object-unsafe**. As long as we expect the trait to always be used as a generic constraint or in an `impl` type, that's OK, but it would prevent having a `Box<dyn CosmosClientMethods>` (or `Arc<dyn CosmosClientMethods>`, `&dyn CosmosClientMethods`, etc.).

The guidelines here are fairly clear though, [using an `impl` type is desirable](https://azure.github.io/azure-sdk/rust_introduction.html#rust-parameters) because it allows calling this like `client.database("foo")` rather than having to call `into()` yourself as a caller.

For the most part, we have this trait to support mocking. I think most applications will use `CosmosClient` directly, and those that care about mocking can use `impl CosmosClientMethods` or generic constraints (for example `Arc<T>` where `T: CosmosClientMethods` will work fine with this trait as is, it's just using `dyn` pointers that won't)

@heaths Is object-safety something that was considered when writing this guideline? As long as it's something we consider and consciously decide on, I'm alright with this trait being object-unsafe.

sdk/cosmosdb/azure_data_cosmos/src/clients/database_client.rs
```diff
@@ -0,0 +1,51 @@
...
+use azure_core::{Context, Request};
+use url::Url;
+
+pub trait DatabaseClientMethods {
+    fn read(
+        &self,
+        options: Option<ReadDatabaseOptions>,
+    ) -> impl std::future::Future<
+        Output = azure_core::Result<azure_core::Response<DatabaseProperties>>,
+    > + Send;
```
analogrelay, line 13: Our guidelines recommend using `async fn` in traits, but that runs counter to the guidance in the `async_fn_in_trait` warning:

```
warning: use of `async fn` in public traits is discouraged as auto trait bounds cannot be specified
 --> sdk/cosmosdb/azure_data_cosmos/src/clients/database_client.rs:8:5
  |
8 |     async fn read(
  |     ^^^^^
  |
  = note: you can suppress this lint if you plan to use the trait only in your own code, or do not care about auto traits like `Send` on the `Future`
  = note: `#[warn(async_fn_in_trait)]` on by default
help: you can alternatively desugar to a normal `fn` that returns `impl Future` and add any desired bounds such as `Send`, but these cannot be relaxed without a breaking API change
  |
8 ~     fn read(
9 |         &self,
10|         options: Option<ReadDatabaseOptions>,
11~     ) -> impl std::future::Future<Output = azure_core::Result<azure_core::Response<DatabaseProperties>>> + Send;
  |

```

.github/CODEOWNERS
Pilchie: Can we make a team to mention here? I think this is fine for now, but we will likely want to include others before long.

Pilchie:
Looking forward to taking a look at this, but it will probably be a couple of days.  Don't block on my review please.

.github/CODEOWNERS
analogrelay: Yep, I've created a team and will update this to reference it.

analogrelay: I've updated it to a team, but that team will need to be granted write access to the repo in order for it to be usable as CODEOWNERS. I think @heaths or someone on the Azure SDK team would need to do this. I noticed we do use a team on the Python and Java SDKs (our .NET SDK is in its own repo), so I think there's reasonable precedent here to use a team. But I can revert it to individual mentions if necessary.

sdk/cosmosdb/azure_data_cosmos/src/clients/database_client.rs
analogrelay: I don't think we _need_ this to be `Send` or `Sync`, but the point made by the warning is still valid. By using `async fn`, it make not be clear that adding one of those auto-trait bounds would be breaking.

analogrelay: It's also important to note that if we desugar the trait method as recommended by the warning, the `impl` **can still be an `async fn`** so we still retain the ability to use `async`/`await` in implementations of the trait.

analogrelay: In fact, this warning is causing a build failure since we fail on all lints in CI at the moment.

sdk/cosmosdb/azure_data_cosmos/src/options/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+
+impl ReadDatabaseOptions {}
+
+pub mod builders {
+    use crate::{CosmosClientOptions, ReadDatabaseOptions};
+
+    #[derive(Default)]
+    pub struct CosmosClientOptionsBuilder(CosmosClientOptions);
+
+    impl CosmosClientOptionsBuilder {
```
analogrelay, line 25: Currently, this PR doesn't contain any way to _set_ the `client_options`. That's because we need some sort of pattern to be established in `azure_core`. I'm looking at options for that (derive macros, traits, whatnot) but for now I'm just leaving it out with the intent that it will come later via ✨ New Cool Magic ✨ .

Pilchie (commented):
Not an expert, but LGTM

analogrelay:
Marking this ready for final review. There are a few outstanding things I could use some assistance with:

* I created a team to represents the Cosmos DB SDK team, `@Azure/azure-cosmos-rust-sdk`, but it needs to have write access to this repo to be a CODEOWNER. The [docs](https://github.com/Azure/azure-sdk/blob/main/docs/policies/opensource.md#codeowners) indicate that's done by adding it as a member of `azure-sdk-write` but I don't have any access to do that (@heaths or @RickWinter , do you?). We can also go back to individual mentions if using a team is problematic, but it would be best if we can add/remove reviewers without having to edit the CODEOWNERS file each time.

* CSpell is failing when analyzing the CODEOWNERS file because it's finding all the usernames. For now, I excluded the file entirely (based on [precedent in azure-sdk-for-python](https://github.com/Azure/azure-sdk-for-python/blob/4e8ee5a8e2d2fc86ccd8c07d80f32e83f1511f14/.vscode/cspell.json)) since CODEOWNERS is likely to contain spelling errors (due to usernames) and has fairly minimal text that needs spell checking.

.github/CODEOWNERS
heaths: The teams are scoped to the `Azure` org so you should be able to use it here as well.

I don't think Rick and I need to be on there, but you can always add us. At some point we'll wire up the APIView workflow...though, I suppose until then maybe we should be on there for that reason. 🤔

RickWinter: > I've updated it to a team, but that team will need to be granted write access to the repo in order for it to be usable as CODEOWNERS. I think @heaths or someone on the Azure SDK team would need to do this. I noticed we do use a team on the Python and Java SDKs (our .NET SDK is in its own repo), so I think there's reasonable precedent here to use a team. But I can revert it to individual mentions if necessary.

@kurtzeborn, @weshaggard Can you share best pratices on setting up github teams for write permissions to the SDKs

sdk/cosmosdb/azure_data_cosmos/src/clients/cosmos_client.rs
heaths: No. I admit object-safety was not considered when writing these guidelines. The user experience was, and being able to test services is paramount. That said, if there are other considerations and we can be both object-safe and flexible, I'm open to it. As you noted, the main reason for preferring `Into<String>` (or `AsRef<str>` - whichever makes sense in context) over `&str` is flexibility but it's not like `&str` is that difficult. Callers can pass `"foo"` or `some_string.as_ref()` easily enough. I'm not thinking much about more abstract types that can be converted into a string e.g. integers.

Given we [can't declare `async fn`](https://doc.rust-lang.org/reference/items/traits.html#object-safety) for object-safe traits, I don't think it's worth considering `&str` in lieu of `Into<String>` or `AsRef<str>`, though.

/cc @RickWinter @JeffreyRichter

sdk/cosmosdb/azure_data_cosmos/src/clients/database_client.rs
heaths: #1796

sdk/cosmosdb/azure_data_cosmos/src/options/mod.rs
heaths: #1737 will define a derive macro, but we don't want users to set `client_options` directly (the [guidelines are a little ahead on this](https://azure.github.io/azure-sdk/rust_introduction.html#rust-client-configuration)). You need to implement the `ClientOptionsBuilder` for now as generated code does: <https://github.com/Azure/autorest.rust/blob/a27439cd9817c41aa801c89534ec4c873757c851/packages/typespec-rust/test/sdk/keyvault_secrets/src/generated/clients/key_vault_client.rs#L249>.

heaths:
Up to @kurtzeborn and @weshaggard but I don't think service teams should be language-specific e.g., `@Azure/azure-cosmos-rust-sdk`. Lots of management overhead in many cases, I imagine, unless there are dedicated language resources within each partner team.

.vscode/cspell.json
```diff
@@ -92,6 +101,14 @@
         "azure",
         "azurite"
       ]
+    },
+    {
+      "filename": "sdk/cosmosdb/azure_data_cosmos/**",
```
heaths, line 106: Any reason not just `sdk/cosmosdb/**`? Precedent is also that we inline words instead of a separate dictionary file, but I'm open to it if @weshaggard or @hallipr don't have any problem with this. Still, I would suggest just `.dict.txt` or something so it doesn't show up by default in common dev shells. It's not something most people need to worry about, but the cSpell VSCode plug-in should still give them a choice to add words to it.

sdk/cosmosdb/azure_data_cosmos/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+description = "Rust wrappers around Microsoft Azure REST APIs - Azure Cosmos DB"
+readme = "README.md"
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+homepage = "https://github.com/azure/azure-sdk-for-rust"
+documentation = "https://docs.rs/azure_data_cosmos"
+keywords = ["sdk", "azure", "rest", "cloud", "cosmosdb", "database"]
```
heaths, line 13: Also just "cosmos" maybe?

sdk/cosmosdb/azure_data_cosmos/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+async-trait.workspace = true
+azure_core.workspace = true
+typespec_client_core = { workspace = true, features = ["derive"] }
+time.workspace = true
+tracing.workspace = true
+url.workspace = true
```
heaths, line 24: We re-export this, and should for `time` as well. Any reason not to reuse those? Ideally I was hoping to "hide" most dependencies such that `azure_core` is, I'd hope, the only one callers would need (apart from `serde` and a few others, of course).

sdk/cosmosdb/azure_data_cosmos/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+clap.workspace = true
+
+[lints]
+workspace = true
+
+[features]
+default = ["hmac_rust"]
+hmac_rust = ["azure_core/hmac_rust"]
+hmac_openssl = ["azure_core/hmac_openssl"]
+key-auth = [] # Enables support for key-based authentication (Primary Keys and Resource Tokens)
```
heaths, line 40: Use underscores for feature names. New guideline I added yesterday so everything from crate names to features to module names are uniformly underscores.

sdk/cosmosdb/azure_data_cosmos/examples/cosmosdb_connect.rs
```diff
@@ -0,0 +1,45 @@
...
+pub struct Args {
+    /// The CosmosDB endpoint to connect to.
+    endpoint: String,
+
+    /// The database to fetch information for.
+    database: String,
+
+    /// An authentication key to use when connecting to the Cosmos DB account. If omitted, the connection will use Entra ID.
+    #[clap(long)]
+    #[cfg(feature = "key-auth")]
```
heaths, line 15: Seems a bit complicated to add feature support to an example. Since you have to check for `Some` anyway, maybe just require the `key_auth` (nee `key-auth`) feature? Granted, that's not a great UX either since callers have to explicitly pass it (which `[[example]]` sections could just define what features to enable, but alas).

sdk/cosmosdb/azure_data_cosmos/examples/cosmosdb_connect.rs
```diff
@@ -0,0 +1,45 @@
...
+    let db_client = client.database(&args.database);
+    let response = db_client.read(None).await?.deserialize_body().await?;
+    println!("{:?}", response);
+    Ok(())
+}
+
+#[cfg(feature = "key-auth")]
+fn create_client(args: &Args) -> CosmosClient {
+    if let Some(key) = args.key.as_ref() {
+        CosmosClient::with_key(&args.endpoint, key.clone(), None).unwrap()
```
heaths, line 34: How married to shared key auth are you? I don't remember if we discussed it with your team, but did discuss with Storage about not support shared key auth at all. Or was it deemed too ubiquitous for Cosmos?

sdk/cosmosdb/azure_data_cosmos/src/authorization_policy.rs
```diff
@@ -0,0 +1,455 @@
...
+}
+
+#[derive(Debug, Clone)]
+enum Credential {
+    /// The credential is an Entra ID token.
+    Token(Arc<dyn TokenCredential>),
+
+    /// The credential is a key to be used to sign the HTTP request (a shared key)
+    #[cfg(feature = "key-auth")]
+    PrimaryKey(Secret),
```
heaths, line 46: Oh, good, you found this!

sdk/cosmosdb/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    /// Gets the endpoint of the database account this client is connected to.
+    pub fn endpoint(&self) -> &Url {
+        &self.endpoint
+    }
+}
+
+impl CosmosClientMethods for CosmosClient {
+    /// Gets a [`DatabaseClient`] that can be used to access the database with the specified ID.
+    fn database(&self, id: impl Into<String>) -> DatabaseClient {
```
heaths, line 100: [Current guideline](https://azure.github.io/azure-sdk/rust_introduction.html#rust-subclients-suffix) states this should be named `database_client`. We can discuss, but this was done to avoid confusion about what it returns since you rarely declare a variable type in Rust.

sdk/cosmosdb/azure_data_cosmos/src/clients/database_client.rs
```diff
@@ -0,0 +1,83 @@
...
+/// A client for working with a specific database in a Cosmos account.
+///
+/// You can get a `DatabaseClient` by calling [`CosmosClient::database()`](CosmosClient::database).
+pub struct DatabaseClient {
+    base_url: Url,
+    root_client: CosmosClient,
+}
+
+impl DatabaseClient {
+    pub(crate) fn new(root_client: CosmosClient, database_id: String) -> Self {
```
heaths, line 51: You only need a `&str` per line 59.

sdk/cosmosdb/azure_data_cosmos/src/clients/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod cosmos_client;
+mod database_client;
+
+pub use cosmos_client::{CosmosClient, CosmosClientMethods};
+pub use database_client::{DatabaseClient, DatabaseClientMethods};
```
heaths, line 5: Didn't consider this. I don't think we should export sub-clients from the crate root since they are not creatable directly and only "pollute" the namespace. @RickWinter @JeffreyRichter thoughts?

heaths (changes requested):
Overall, I think this looks great! Nice job! There are a few open questions, and a few minor changes I listed so far like using underscores for feature names as well. I'd love to hear feedback on that as well, but I added that to guidelines yesterday to use underscores uniformly. Personally, I hate having to guess at whether to use `-` or `_` and get errors 50% of the time. That's USB-A. I'm shooting for USB-C. 😉

sdk/cosmosdb/azure_data_cosmos/examples/cosmosdb_connect.rs
analogrelay: I don't think we'll be able to remove it completely. Certain control-plane operations, namely managing Databases and Containers, can be performed using the Data Plane REST APIs (for example, see the [Create Database](https://learn.microsoft.com/en-us/rest/api/cosmos-db/create-a-database) REST API) **but only** when authenticating with a shared key (if you want to perform those operations using Entra ID, you need to use the Control Plane APIs).

As a result, it's difficult to justify removing shared key auth _entirely_. My hope here is that by putting it behind a feature, we can walk a bit of a line here of providing access to those APIs while still making the "standard" experience support Entra ID only.

.vscode/cspell.json
analogrelay: The main reason I went with a custom dict here is that these are terms that will likely appear with some frequency in the Cosmos DB crate **and** are quite unique to Cosmos. For example `pkranges`, which refers to the Partition Key Ranges API. I'm certainly open to other options here though.

> Any reason not just sdk/cosmosdb/**

Nope, I'm happy to expand it up.

sdk/cosmosdb/azure_data_cosmos/src/clients/mod.rs
JeffreyRichter: Yes, I would agree. The crate root is the main stuff that all/most customers need and sub crates are for the minority of customers or are gettable from something in the root create so as not to pollute the root crate (as Heath says).

sdk/cosmosdb/azure_data_cosmos/src/clients/cosmos_client.rs
JeffreyRichter: I don't know all the ramifications around object-safety, so I think I'm not qualified to weigh in on this. I'm fine with approving this now and later having a discussion regarding the ramifications to make the best decision. As always, I'm hoping we get customer feedback which should help us either build confidence in what we've done or being flexible to change our decisions/designs.

JeffreyRichter (approved):

sdk/cosmosdb/azure_data_cosmos/src/clients/cosmos_client.rs
analogrelay: That makes sense. I generally agree that we don't need to design for object-safety here. The user experience is really much clearer if we can use `impl` types in the method names. I mostly just wanted to make sure that constraint was called out.

sdk/cosmosdb/azure_data_cosmos/src/clients/mod.rs
analogrelay: I think that makes sense. My main concern was that `DatabaseClient` (and later `ContainerClient`) are types that users will likely want to name in their apps. For example, when a user is building a helper to fetch an item that takes `ContainerClient` as an argument. However, I think I'm sufficiently convinced that will be rare, and the type will still be exported publicly, so it can be used, it just requires referencing the submodule via `use` or a full path.

sdk/cosmosdb/azure_data_cosmos/src/clients/database_client.rs
analogrelay: I left a fairly detailed comment there with some of my understanding of the problem and proposal for how to move forward. For now, I'll do as suggested and use `async fn` and `expect` the lint to suppress it.

analogrelay: I'll `allow` it actually. The `expect` attribute was only just stabilized in `1.81` and appears it's not possible to condition attributes on compiler version in Stable rust either.

sdk/cosmosdb/azure_data_cosmos/Cargo.toml
analogrelay: We need `form_urlencoded` from the `url` crate, which isn't re-exported (and I'd prefer we don't get too deep down the rabbit hole of re-exporting all the functionality from dependent crates out of `typespec_client_core`.

analogrelay: I _can_ use the re-export of `OffsetDateTime` though.

sdk/cosmosdb/azure_data_cosmos/src/clients/mod.rs
analogrelay: The approach I'm going with here is to continue to export `DatabaseClient(Methods)` from _this_ module. But at `lib.rs` I'm exporting only `CosmosClient` and `CosmosClientMethods` from the top-level. That way the database client types are still nameable with `azure_data_cosmos::clients::DatabaseClient` but only `CosmosClient(Methods)` is in the top-level namespace.

heaths: > it just requires referencing the submodule via `use` or a full path

Fortunately, rust-analyzer makes simple work of that. Too simple, in fact. In some cases I wish it wouldn't suggest some things because I think it leads to confusion but can't have my cake and eat it too.

.vscode/cspell.json
heaths: Oh, and by inlining I didn't mean in the source itself but in the section you added here i.e., just a `"words": []` array instead of a dictionary. Again, though, I'm open to allowing for dictionary files within service directories if @weshaggard is cool with it. It provides isolation to just the service directory which shouldn't trigger such broad pipelines for modifying more generally-scoped files like we do for the same case in the specs repos.

.github/CODEOWNERS
weshaggard: Create a github team and add it "azure-sdk-write" as the parent of that team and let Scott or I know and we can approve the parenting.

.vscode/cspell.json
weshaggard: https://cspell.org/configuration/imports/

I've been starting to think we should just have service teams create a config in their directory and import the common shared one to avoid conflicts and allow customization.

weshaggard: That said I'm also OK with just having a dictionary file in the service directory as well.

analogrelay: Yeah, that's a good point. I'll stick with a `words` list here and we can add a dictionary file later if it gets to be a long list.

.github/CODEOWNERS
analogrelay: I'll work on this separately. I've already create a team, and I'm not sure I can re-parent it, but I can work something out ;). For now, the individual mentions will be fine.

.vscode/cspell.json
```diff
@@ -56,8 +56,8 @@
       "noSuggest": true
     },
     {
-      "name": "cosmosdb",
-      "path": "../sdk/cosmosdb/cosmosdb.dict.txt",
+      "name": "cosmos",
+      "path": "../sdk/cosmos/.dict.txt",
```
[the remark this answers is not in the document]
analogrelay: Oop, just accidentally left the dict path in. But if you're good with the `.dict.txt` file, I'll do that!

sdk/cosmos/azure_data_cosmos/examples/cosmos_connect.rs
```diff
@@ -1,18 +1,18 @@
-use azure_data_cosmos::{CosmosClient, CosmosClientMethods, DatabaseClientMethods};
+use azure_data_cosmos::{clients::DatabaseClientMethods, CosmosClient, CosmosClientMethods};
 use clap::Parser;
 
 /// A simple example to show connecting to a Cosmos Account and retrieving the properties of a database.
 #[derive(Parser)]
 pub struct Args {
-    /// The CosmosDB endpoint to connect to.
+    /// The Cosmos endpoint to connect to.
```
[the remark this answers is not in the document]
analogrelay: Yeah, you're right, it's "Cosmos DB". I'll update this.

PR #1780, opened 2024-08-29
Rename private Context to ErrorContext

### LINKED ISSUE #1778: Rename `error::Context` to `ErrorContext`
...to avoid confusion

heaths:
To avoid confusion with our HTTP client `Context`, even if private so it doesn't show up in symbol searches. Resolves #1778

### DIFF
--- a/sdk/typespec/src/error/mod.rs
+++ b/sdk/typespec/src/error/mod.rs
@@ -37,7 +37,7 @@ impl ErrorKind {
     /// Consumes the `ErrorKind` and converts to an [`Error`].
     pub fn into_error(self) -> Error {
         Error {
-            context: Context::Simple(self),
+            context: ErrorContext::Simple(self),
         }
     }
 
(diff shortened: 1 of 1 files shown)


LarryOsterman (approved):

sdk/typespec/src/error/mod.rs
```diff
@@ -71,7 +71,7 @@ impl Display for ErrorKind {
 /// An error encountered when communicating with the service.
 #[derive(Debug)]
 pub struct Error {
-    context: Context,
+    context: ErrorContext,
```
RickWinter, line 74: Part of me wonders if the Azure::Core::Context is the type in Rust that has a poor name choice. Afterall the field on Error is already context.

We should user study this naming choice before GA

RickWinter (approved):

sdk/typespec/src/error/mod.rs
heaths: This isn't even exposed. `Context` in the public API is fine and matches other contexts in other languages' HTTP stacks.

PR #1781, opened 2024-08-29
Move verify-dependencies to analyze job

hallipr:


eng/pipelines/templates/steps/analyze.yml
```diff
@@ -4,18 +4,60 @@ parameters:
     default: stable
 
 steps:
+  - ${{ if ne(parameters.Toolchain, 'nightly') }}:
+      - template: /eng/pipelines/templates/steps/use-rust.yml@self
+        parameters:
+          Toolchain: nightly
+
```
heaths, line 11: Why did you put this back after removing it from the other PR? We don't need to run this on every toolchain - just nightly where it'll work until stabilized. Even then, we only need to run it on, say, stable.

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -4,18 +4,60 @@ parameters:
...
+      Write-Host "`n`n"
       Invoke-LoggedCommand "cargo +$(Toolchain) doc --all --no-deps"
+
+      # Verify package dependencies
+
+      $verifyDependenciesScript = Join-Path $RepoRoot '/eng/scripts/verify-dependencies.rs' -Resolve
+      $packageInfoPath = '$(Build.ArtifactStagingDirectory)/PackageInfo'
+
+      if ('$(NoPackagesChanged)' -ne 'true') {
```
heaths, line 41: What is going on here? In PowerShell, this won't resolve. Single quotes are a literal, and even if it were double quotes `$(NoPackagesChanged)` wouldn't resolve a variable (`${NoPackagesChanged}`) would. `NoPackagesChanged` would then have to be a command.

Even after all that, why is it returning a boolean as a string? This makes truthy comparisons difficult and non-idiomatic to PowerShell.

All said, I don't think this will work the way you intend. This looks more like transpiled AzDO Pipelines grammar.

heaths (changes requested):

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -4,18 +4,60 @@ parameters:
...
+          Toolchain: nightly
+
   - template: /eng/pipelines/templates/steps/use-rust.yml@self
     parameters:
       Toolchain: ${{ parameters.Toolchain }}
 
+  - template: /eng/common/pipelines/templates/steps/save-package-properties.yml
+
   - pwsh: |
+      . ./eng/common/scripts/common.ps1
```
weshaggard, line 19: We should go ahead and move this into a script. It was already getting large but this pushes it over for me.

eng/pipelines/templates/steps/analyze.yml
hallipr: We only run the Analyze job with stable, but we have to use nightly to run the rust script.  This bakes out to a single invocation of the "Run source analysis" step with toolchain == stable, but with the invocations:

```
/> cargo +stable check -p azure_core --no-default-features
/> cargo +stable fmt --all -- --check
/> cargo +stable clippy --all
/> cargo +stable doc --all --no-deps

/sdk/typespec> cargo +nightly -Zscript /mnt/vss/_work/1/s/eng/scripts/verify-dependencies.rs
/sdk/core> cargo +nightly -Zscript /mnt/vss/_work/1/s/eng/scripts/verify-dependencies.rs
...
```

eng/pipelines/templates/steps/analyze.yml
hallipr: It's because we're using PowerShell in YAML.   This is checking the string content of the pipeline variable `NoPackagesChanged`

hallipr: It'll compile to `if ('' -ne 'true')` or `if ('true' -ne 'true')` before invocation.

hallipr: This script is getting almost complex enough to put into a .ps1 file instead of inline.

eng/pipelines/templates/steps/analyze.yml
hallipr: moved to script file

eng/pipelines/templates/steps/analyze.yml
heaths: That's fine, but let's make sure it's clear it's not for devs' inner loop. We don't need scripts for people to learn when typical `cargo` commands should work just fine.

eng/scripts/Analyze-Code.ps1
```diff
@@ -0,0 +1,54 @@
...
+
+$verifyDependenciesScript = Join-Path $RepoRoot 'eng' 'scripts' 'verify-dependencies.rs' -Resolve
+
+if (!$SkipPackageAnalysis) {
+  if (!(Test-Path $PackageInfoPath)) {
+    Write-Error "Package info path '$PackageInfoPath' does not exist."
+    exit 1
+  }
+
+  $pacakgesToTest = Get-ChildItem $PackageInfoPath -Filter "*.json" -Recurse
```
heaths, line 39:
```suggestion
  $packagesToTest = Get-ChildItem $PackageInfoPath -Filter "*.json" -Recurse
```

Misspelled below. At least you were consistent. :smile:

heaths (approved):

eng/scripts/Analyze-Code.ps1
hallipr: It was 6 places in 2 files 😳

PR #1782, opened 2024-08-29
Fix some all-features build issues

heaths:


analogrelay (approved):
Good catch. I only just realized that `--all` is **not** actually shorthand for `--all-features`, as I'd previously assumed 🤦🏻‍♀️ .

PR #1786, opened 2024-09-03
Created cplusplus feature to isolate C++ only elements

LarryOsterman:
Isolate AMQP "composite" type to C++ only because it has no fe2o3 equivalent (the AMQP "composite" type is handled using macros in fe2o3).

### DIFF
--- a/sdk/core/azure_core_amqp/src/fe2o3/value.rs
+++ b/sdk/core/azure_core_amqp/src/fe2o3/value.rs
@@ -185,6 +185,7 @@ impl From<AmqpValue> for fe2o3_amqp_types::primitives::Value {
             AmqpValue::Array(a) => fe2o3_amqp_types::primitives::Value::Array(
                 a.into_iter().map(|v| v.into()).collect(),
             ),
+            #[cfg(feature = "cplusplus")]
             AmqpValue::Composite(d) => fe2o3_amqp_types::primitives::Value::Described(Box::new(
                 serde_amqp::described::Described {
                     descriptor: d.descriptor.clone().into(),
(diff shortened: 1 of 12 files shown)


sdk/core/azure_core_amqp/src/fe2o3/value.rs
heaths, line 1: I don't understand why these are C++-only features. If `Compositive` isn't supported, depending on the direction of marshaling, seems you should do something (though not panic). I'll take your word for it since I don't have the context you do, but it does smell a little fishy.

heaths (approved):

heaths:
You can probably "cplusplus" to the global dictionary.

sdk/core/azure_core_amqp/src/fe2o3/value.rs
LarryOsterman: The iron oxide version of AMQP handles AMQP Composite types differently from C++ and UAMQP.

In order to use a newtype implementation of AMQP Values in the C++ wrapper, there needs to be a 1:1 type mapping between the Rust AmqpValue and the RustWrapper (C++ projection) AmqpValue. Otherwise you need to do an additional level of indirection just for the AmqpComposite type (which exists only in C++).

When we're asked to express an AmqpComposite as an fe2o3 type, we can (and should) represent it as an AmqpDescribed type.

I've changed the PartialEq code to always fail on a AmqpComposite type instead of panicking.


There are a bunch of serialization functions in AMQP which also only exist to support the C++ experience, so those are also tagged with the cplusplus feature.

LarryOsterman: Similarly, the serialize to list functions are only there to handle the case of C++ serialize/deserialize, which isn't needed for eventhubs, but is a part of the C++ API surface.

PR #1790, opened 2024-09-03
Move nightly toolchain scripts to xtasks

heaths:
This is an emerging pattern many other projects are using and allows us to develop them with current rust-analyzer and friends, which rust scripts - requiring the nightly toolchain - do not.


heaths:
Despite .editorconfig files using a TOML-like syntax, seems I won't be able to use the `toml` crate because it strictly enforces some rules like "valid" section names and quoted strings. I don't want to pull in a lot of dependencies because `cargo xtask`s should be quick to build. Perhaps I should go the route of making each a separate EXE to minimize dependencies for some tasks at least.

analogrelay:
> Despite .editorconfig files using a TOML-like syntax, seems I won't be able to use the toml crate because it strictly enforces some rules like "valid" section names and quoted strings. I don't want to pull in a lot of dependencies because cargo xtasks should be quick to build.

Looks like [`configparser`](https://crates.io/crates/configparser) would work. Agreed in general about keeping dependencies minimal, but dependency builds _should_ be cached for the user (even with xtasks) so I'm not _that_ worried about adding one more if it avoids custom parsing stuff.

> This is an emerging pattern many other projects are using and allows us to develop them with current rust-analyzer and friends, which rust scripts - requiring the nightly toolchain - do not.

I like this pattern, particularly as it avoids nightly rust as much as possible. In general, I prefer writing scripts using the toolchain if possible anyway, since they can be cross-platform that way.

heaths:
I'm not sure about this approach either. As our repo grows, putting more and more into the workspace - especially with dependencies not generally needed by other crates (which we want to reduce anyway) - is going to make things slower and slower. Rust scripts will stabilize eventually and editor support should come along with that, so this may not be worth it.

heaths:
> Looks like [`configparser`](https://crates.io/crates/configparser) would work. Agreed in general about keeping dependencies minimal, but dependency builds _should_ be cached for the user (even with xtasks) so I'm not _that_ worried about adding one more if it avoids custom parsing stuff.

Looks like it's active as well. I did find a crate for helping assert files against an `.editorconfig` but it hasn't been touched for 3 years. Unfortunately, this is all too common. I may also just look into EditorConfig-recomended JS libraries we can run at build time and see if I can instead lint a header. I started work on one a while back but shelved it for a time. Based on dylint, which I think we'd want long-term anyway just like Azure SDKs for .NET add Roslyn analyzers as needed.

analogrelay:
Another option for linting is [super-linter](https://github.com/super-linter/super-linter). I've worked with some projects that use it. It's essentially a bundle of linters. It can be a bit of a heavy option but has the advantage of just setting a style baseline for everything.

heaths:
Not sure we want to do this, but I also don't want them to be part of the toolchain. We can still add them to `.cargo/config.toml` so they are easy to execute, but lets put them in a separate workspace if we do move to xtasks. We'll already have enough workspace members to worry about that rust-analyzer will take a while.

PR #1791, opened 2024-09-05
Refactor some models into typespec_client_core

heaths:


heaths:
I'm going to try to do more, smaller PRs for refactoring to make them easier to review and easier to see when/where I might've missed re-exporting some type(s) exactly as they were defined before in `azure_core`.

LarryOsterman (approved):

sdk/typespec/typespec_client_core/src/macros.rs
```diff
@@ -1,6 +1,100 @@
...
 // Licensed under the MIT License.
 
+/// The following macro invocation:
+/// ```
+/// # #[macro_use] extern crate typespec_client_core;
+/// create_enum!(Words, (Chicken, "Chicken"), (White, "White"), (Yellow, "Yellow"));
+/// ```
+/// Turns into a struct where each variant can be turned into and construct from the corresponding string.
+#[macro_export]
+macro_rules! create_enum {
```
analogrelay, line 11: nit: `create_enum` feels like a very generic name to me. I'd prefer something a little more descriptive like `create_string_enum`.

I know you're just moving an existing macro though, so I'm totally fine with you leaving this as-is or deferring that change to a separate PR.

sdk/typespec/typespec_client_core/src/macros.rs
```diff
@@ -1,6 +1,100 @@
...
+                }
+            }
+        }
+
+        impl serde::Serialize for $name {
+            fn serialize<S>(&self, s: S) -> ::core::result::Result<S::Ok, S::Error>
+            where S: serde::Serializer {
+                return s.serialize_str(&self.to_string())
+            }
+        }
```
analogrelay, line 74: nit: I think we could do this without implementing `Deserialize` and `Serialize` directly. My understanding is that applying [the `#[serde(rename)]` attribute](https://serde.rs/variant-attrs.html#rename) to each variant would do the same thing.

Again, I know you're just moving code so this is non-blocking.

analogrelay (approved):
Minor nits that are mostly just me reviewing this moved code that I've never seen before ;). Nothing blocking from me.

sdk/typespec/typespec_client_core/src/macros.rs
heaths: I plan to replace this with a derive macro at some point anyway. This is more so for creating "normal" and "extensible" enums for service methods, to which I've provided guidelines for Rust and C# as well (back in the day).

sdk/typespec/typespec_client_core/src/macros.rs
heaths: Yeah, I've never been sure why the author did this originally. Maybe to product faster code than `serde/derive`? I can't imagine it being much different. In my derive macro plans, I do plan to utilize `#[serde(rename)]`, though.

PR #1793, opened 2024-09-09
General client core refactoring

heaths:


heaths:
I should have one more after this that will mostly be just moves.

analogrelay (approved):

analogrelay:
I merged #1772 which caused the conflicts. They look quite simple to resolve.

PR #1794, opened 2024-09-09
Refactor request options and some related types

heaths:
Moves common request options e.g., `ContentLength` into `typespec_client_core` while reducing the number of seldom-used (specified directly) types exported from `typespec_client_core::http`.


heaths:
This should hopefully be the last major refactor. Honestly, though, I'm not sure this is worth it. Looking at some other languages' unbranding, it seems they don't do much with headers (or query params), but this seems doable and at least serves as an example in `typespec_client_core`. @RickWinter, @JeffreyRichter, thoughts?

JeffreyRichter (approved):

heaths:
> Seems like a fine thing to try and get feedback on. I always worry when we make value judgments (which options are "common") on behalf of customers. Some customers may find that one thing we buried is common for them and something we make more visible is not common for them or vice versa. Also, as we add more things, we then have to make arbitrary decisions as to whether it is common or not and this can breed contradictions between team members.
> 
> That being said, I still agree that this is a fine thing to try and get feedback on.

Fair point, though by "common" I probably should've said "standard" e.g., "Content-Length" or "User-Agent" are standard headers, but any query string parameter or "x-ms-"-prefaced HTTP headers would, of course, not be standard.

PR #1798, opened 2024-09-11
Better centralize and remove unused dependencies

heaths:


Cargo.toml
```diff
@@ -65,6 +65,7 @@ futures = "0.3"
 getrandom = { version = "0.2", features = ["js"] }
 hmac = { version = "0.12" }
 http-types = { version = "2.12", default-features = false }
```
LarryOsterman, line 67: Log instead of tracing?

LarryOsterman (approved):

JeffreyRichter (approved):

Cargo.toml
heaths: This is only cleaning up dependencies but not changing code. We can normalize around logging, tracing, etc. later. We have a lot of work to do for otel anyway as it is.

PR #1799, opened 2024-09-12
support user-assigned identities in VirtualMachineManagedIdentityCredential

cataggar:
Same as #1660, but targeting new sdk branch.

Fixes https://github.com/Azure/azure-sdk-for-rust/issues/1659. As it points out, I think it was a bit too early to make ImdsManagedIdentityCredential private in https://github.com/Azure/azure-sdk-for-rust/pull/1532. I think it is the right direction after the credentials that use it like VirtualMachineManagedIdentityCredential and AppServiceManagedIdentityCredential add support for configurating user managed identities.

### DIFF
--- a/sdk/identity/azure_identity/src/token_credentials/imds_managed_identity_credentials.rs
+++ b/sdk/identity/azure_identity/src/token_credentials/imds_managed_identity_credentials.rs
@@ -16,14 +16,18 @@ use serde::{
 use std::{str, sync::Arc};
 use time::OffsetDateTime;
 
+/// An identifier for the Azure Instance Metadata Service (IMDS).
+/// IMDS provides information about currently running virtual machine instances.
+/// https://learn.microsoft.com/azure/virtual-machines/instance-metadata-service
+/// https://learn.microsoft.com/azure/app-service/overview-managed-identity?tabs=portal%2Chttp#rest-endpoint-reference
 #[derive(Debug)]
 pub enum ImdsId {
     SystemAssigned,
-    #[allow(dead_code)]
+    /// The client ID of the user-assigned identity to be used.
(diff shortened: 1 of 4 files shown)


cataggar:
Thanks @johnbatty! Wonderful feedback as always. Changes made.

heaths:
Ultimately, this credential wouldn't be exposed directly and will align with all other Azure SDK languages' implementations, which expose a `ManagedIdentityCredential` and any required inputs will align as well e.g., <https://learn.microsoft.com/dotnet/api/azure.identity.managedidentitycredential>.

sdk/identity/azure_identity/src/token_credentials/imds_managed_identity_credentials.rs
```diff
@@ -16,14 +16,18 @@ use serde::{
 use std::{str, sync::Arc};
 use time::OffsetDateTime;
 
+/// An identifier for the Azure Instance Metadata Service (IMDS).
+/// IMDS provides information about currently running virtual machine instances.
+/// https://learn.microsoft.com/azure/virtual-machines/instance-metadata-service
+/// https://learn.microsoft.com/azure/app-service/overview-managed-identity?tabs=portal%2Chttp#rest-endpoint-reference
```
heaths, line 22: Nit:
```suggestion
/// An identifier for the Azure Instance Metadata Service (IMDS).
///
/// IMDS provides information about currently running virtual machine instances. For more information, see:
/// * https://learn.microsoft.com/azure/virtual-machines/instance-metadata-service
/// * https://learn.microsoft.com/azure/app-service/overview-managed-identity#rest-endpoint-reference
```

heaths (approved):

PR #1800, opened 2024-09-12
Add codespaces/devcontainer config files

analogrelay:
In my search for a fairly consistent and isolated development environment, we've agreed that [Nix ain't it](https://github.com/Azure/azure-sdk-for-rust/pull/1771). Honestly, while I like NixOS, I've drunk the kool-aid hard and I'm the first to admit it isn't even remotely accomodating to newcomers.

So, instead, how about Codespaces? I've noticed several other SDK repos use them and I've had great success working with Codespaces before.

This PR adds devcontainer config that should make building this repo relatively straightforward, and we can add/remove/update it as we go. The current config:

1. Is based on the devcontainer image using the [latest 1.x Rust official docker image](https://hub.docker.com/_/rust/) at the time you build it (so `1.80.1` as of today)
2. Includes the Azure CLI, GitHub CLI, and SSHD (to enable the `gh cs ssh` command from your local machine to SSH to a running codespace)
3. Has our MSRV toolchain (whatever is in `rust-toolchain.toml`), the latest stable (in case it's ahead of the docker image) and the latest nightly rust installed
4. Has all our dependencies built against all three toolchains (MSRV, latest stable, latest nightly) cached, to make your first build 
🚤 fast 🚤 .

I noticed that the VS Code "Add Devcontainer.json" config automatically added a dependabot.yml to keep the devcontainer up to date. I'm fine removing that if we don't want it.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,27 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/rust
+{
+  "name": "Rust",
+  "image": "mcr.microsoft.com/devcontainers/rust:1-1-bullseye",
+  "onCreateCommand": ".devcontainer/oncreate",
+  "features": {
+    "ghcr.io/devcontainers/features/azure-cli:1": {
+      "version": "latest"
+    },
+    "ghcr.io/devcontainers/features/github-cli:1": {
+      "installDirectlyFromGitHubRelease": true,
+      "version": "latest"
+    },
+    "ghcr.io/devcontainers/features/sshd:1": {
+      "version": "latest"
+    }
+  },
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "rust-lang.rust-analyzer",
+        "EditorConfig.EditorConfig"
+      ]
+    }
+  }
+}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -5,22 +5,16 @@
   "image": "mcr.microsoft.com/devcontainers/rust:1-1-bullseye",
   "onCreateCommand": ".devcontainer/oncreate",
   "features": {
-    "ghcr.io/devcontainers/features/azure-cli:1": {
-      "version": "latest"
-    },
-    "ghcr.io/devcontainers/features/github-cli:1": {
-      "installDirectlyFromGitHubRelease": true,
-      "version": "latest"
-    },
-    "ghcr.io/devcontainers/features/sshd:1": {
-      "version": "latest"
-    }
+    "ghcr.io/devcontainers/features/azure-cli:1": {},
+    "ghcr.io/devcontainers/features/github-cli:1": {},
+    "ghcr.io/devcontainers/features/sshd:1": {}
   },
   "customizations": {
     "vscode": {
       "extensions": [
         "rust-lang.rust-analyzer",
-        "EditorConfig.EditorConfig"
+        "EditorConfig.EditorConfig",
+        "streetsidesoftware.code-spell-checker"
       ]
     }
   }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,7 @@
 // For format details, see https://aka.ms/devcontainer.json. For config options, see the
 // README at: https://github.com/devcontainers/templates/tree/main/src/rust
 {
-  "name": "Rust",
+  "name": "Azure SDK for Rust",
   "image": "mcr.microsoft.com/devcontainers/rust:1-1-bullseye",
   "onCreateCommand": ".devcontainer/oncreate",
   "features": {
@@ -12,9 +12,11 @@
   "customizations": {
     "vscode": {
       "extensions": [
+        "tamasfe.even-better-toml",
+        "editorconfig.editorconfig",
         "rust-lang.rust-analyzer",
-        "EditorConfig.EditorConfig",
-        "streetsidesoftware.code-spell-checker"
+        "streetsidesoftware.code-spell-checker",
+        "vadimcn.vscode-lldb"
       ]
     }
   }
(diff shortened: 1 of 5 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,21 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/rust
+{
+  "name": "Rust",
```
heaths, line 4: Given where this typically shows up, "Azure SDK for Rust" is probably better.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,21 @@
...
+  "image": "mcr.microsoft.com/devcontainers/rust:1-1-bullseye",
+  "onCreateCommand": ".devcontainer/oncreate",
+  "features": {
+    "ghcr.io/devcontainers/features/azure-cli:1": {},
+    "ghcr.io/devcontainers/features/github-cli:1": {},
+    "ghcr.io/devcontainers/features/sshd:1": {}
```
heaths, line 10: Why? This opens a can of worms I think we should avoid given all the ways you can connect to a devcontainer currently.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,21 @@
...
+  },
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "rust-lang.rust-analyzer",
+        "EditorConfig.EditorConfig",
+        "streetsidesoftware.code-spell-checker"
+      ]
```
heaths, line 18: This should match what's in `.vscode/extensions.json` - which can certain be updated as well. I think we should also add the LLDB debugger like I do in all mine e.g., <https://github.com/heaths/template-rustlang/blob/main/.devcontainer/devcontainer.json>.

.devcontainer/oncreate
```diff
@@ -0,0 +1,31 @@
...
+    stable \
+    nightly \
+    --component "rustc" \
+    --component "rustfmt" \
+    --component "rust-std" \
+    --component "cargo" \
+    --component "clippy" \
+    --component "rust-analyzer"
+
+echo "Performing an initial build to cache dependencies"
```
heaths, line 21: Overkill and harder to maintain. Just run `rustup show` with the PWD set to our repo and our `rust-toolchain.toml` will do all this automatically.

.devcontainer/oncreate
```diff
@@ -0,0 +1,31 @@
...
+
+for toolchain in "stable" "nightly" "$MSRV"; do
+    echo "Building for Rust $toolchain ..."
+    # It's not possible to tell cargo to build _only_ dependencies, so we just build the repo and ignore failures.
+    # When building the codespace against main (such as in prebuilds) the repo _should_ build successfully, but we don't care if this build fails.
+    # It's just to prime dependencies and make user compiles faster.
+    if ! cargo +$toolchain build --all; then
+        echo "Warning: Toolchain $toolchain failed to build. Continuing precreate script"
+    fi
+done
```
heaths, line 31: This adds a lot of time to starting up the devcontainer, and most devs probably won't care about any toolchain outside what our `rust-toolchain.toml` specifies. I'd nix (no pun intended) this block.

heaths (changes requested):

.devcontainer/devcontainer.json
analogrelay: [It's part of the requirements in order to be able to SSH to a Codespace](https://docs.github.com/en/codespaces/developing-in-a-codespace/using-github-codespaces-with-github-cli#ssh-into-a-codespace). I'm OK removing it if you don't feel that's necessary.

.devcontainer/devcontainer.json
analogrelay: Strange that it prompted me to add extensions here when `.vscode/extensions.json` exists. I'd kinda hope that Codespaces could just install the extensions from there... I'll remove this list for now and see if `.vscode/extensions.json` is sufficient, and I'll add the LLDB debugger and spell checker to that file.

analogrelay: Well, that doesn't work. Users can obviously still install them, but there's no way that I can find to tell the codespace config to _automatically_ install extensions from `.vscode/extensions.json`. So, for now, I'll mirror the lists, I guess.

.devcontainer/devcontainer.json
analogrelay: Clarifying a bit.

I do often SSH to codespaces, to get a full-screen terminal or to run TUI apps like [lazygit](https://github.com/jesseduffield/lazygit) (which I'll admit I use more than most, as a 🤓 Big Nerd 🤓 ). So I do prefer having it available, but if you're concerned about other aspects of it, I'm OK removing it. 

The SSHD feature works through the existing secure port forwarding supported by Codespaces, and the `gh` CLI has native support for it, so we're not talking about Codespaces with port `22` open to the internet.

heaths: Thanks for clarifying. I'm fine keeping it in, then. I was afraid it was basically just stock sshd which - surprisingly - still has PasswordAuthentication enabled - at least in Ubuntu - last I checked.

.devcontainer/devcontainer.json
heaths: Yeah, I brought this up with a team internally as well but never heard back. I also didn't follow up.

heaths (approved):

.vscode/cspell.json
```diff
@@ -7,6 +7,7 @@
     "**/test-resources.json",
     ".devcontainer/devcontainer.json",
     ".devcontainer/oncreate",
+    ".github/CODEOWNERS",
     ".github/dependabot.yml",
     ".vscode/cspell.json",
     ".vscode/extensions.json",
```
heaths, line 13: @weshaggard there are enough of "misspellings" in these files, should we just ignore these directories entirely?

(I wouldn't hold up this PR for it, though.)

analogrelay: I've got to resolve this to merge, so I'm going to do that and we could move this conversation elsewhere. My personal thought is that it's not a big deal to exclude new files as they come for now.

weshaggard: Yes I'm OK excluding the entire directory.

PR #1801, opened 2024-09-12
AMQP changes for C++ Connection support

LarryOsterman:


LarryOsterman:
Note that the new close_with_error pretty much hacks around the semantics of closing with an error. 

I need to revisit how AMQP expresses errors at some point.

sdk/core/azure_core_amqp/Cargo.toml
```diff
@@ -46,6 +47,7 @@ fe2o3-amqp = [
 ]
 cplusplus = []
 test_e2e = []
+cpp-repo = ["uuid"]
```
heaths, line 50: New guideline: all features must use underscores. Also, to disambiguate, I've been prefacing dependencies with "dep":

```suggestion
cpp_repo = ["dep:uuid"]
```

sdk/core/azure_core_amqp/src/connection.rs
```diff
@@ -81,45 +124,54 @@ pub mod builders {
                 options: Default::default(),
             }
         }
-        pub fn build(self) -> AmqpConnectionOptions {
-            self.options
+        pub fn build(&mut self) -> AmqpConnectionOptions {
+            self.options.clone()
         }
         #[allow(dead_code)]
```
heaths, line 130: Are all these annotations needed now? I'm guessing they are left-overs.

sdk/core/azure_core_amqp/src/fe2o3/messaging/message_fields.rs
```diff
@@ -50,7 +50,11 @@ impl From<AmqpMessageId> for fe2o3_amqp_types::messaging::MessageId {
 
 #[test]
 fn test_message_id_conversion() {
+    #[cfg(not(feature = "cpp-repo"))]
     use azure_core::Uuid;
+    #[cfg(feature = "cpp-repo")]
+    use uuid::Uuid;
```
heaths, line 56: This shouldn't be necessary. `azure_core::Uuid` is `uuid::Uuid` (but leaves open the possibility of replacing it, which I am considering along with `Url` to reduce 3P dependencies). I don't understand how this is supposed to fix your issues when vendoring into the azure-sdk-for-cpp repo. We should sync up offline. This feels like a pretty gross and unnecessary hack.

heaths (changes requested):

sdk/core/azure_core_amqp/src/fe2o3/messaging/message_fields.rs
LarryOsterman: It *is* a gross hack.

The core problem here is that when I'm building in the C++ repo (when the cpp_repo feature is enabled), the Uuid type needs to come from the `uuid` crate (because the Uuid type is private to the `azure_core` crate in the track1 Rust SDK). But when I'm building in the Rust SDK, the Uuid type can come from the azure_core crate.

LarryOsterman: Here's the diagnostic without the hack:

```
error[E0603]: struct `Uuid` is private
  --> sdk\core\azure-core-amqp\rust_amqp\azure_core_amqp\src\fe2o3\messaging\message_fields.rs:54:21
   |
54 |     use azure_core::Uuid;
   |                     ^^^^ private struct
   |
note: the struct `Uuid` is defined here
  --> C:\Users\larryo.REDMOND\.cargo\registry\src\index.crates.io-6f17d22bba15001f\azure_core-0.18.0\src\lib.rs:43:5
   |
43 | use uuid::Uuid;
   |     ^^^^^^^^^^
help: import `Uuid` directly
   |
54 |     use uuid::Uuid;
   |         ~~~~~~~~~~

```

sdk/core/azure_core_amqp/src/connection.rs
LarryOsterman: Removed. When I initially built this, Rust was upset with them, but it doesn't seem to be as annoyed any more.

PR #1802, opened 2024-09-13
fix azure_data_cosmos build with key_auth feature

analogrelay:
Fixes the build for `azure_data_cosmos` when the `key_auth` feature is enabled. I missed a `use` statement, and I actually do need `time::Duration` which we don't re-export, so I added it as a _dev-dependency_ for `azure_data_cosmos` (since we only need it in a test)

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/Cargo.toml
+++ b/sdk/cosmos/azure_data_cosmos/Cargo.toml
@@ -28,6 +28,7 @@ tokio.workspace = true
 serde_json.workspace = true
 azure_identity.workspace = true
 clap.workspace = true
+time.workspace = true
 
 [lints]
 workspace = true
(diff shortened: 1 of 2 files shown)


sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -28,6 +28,7 @@ tokio.workspace = true
 serde_json.workspace = true
 azure_identity.workspace = true
 clap.workspace = true
+time.workspace = true
```
analogrelay, line 31: Just a note that this is a dev-dependency (hard to tell from the diff). It's only used by a test in `azure_data_cosmos` (to create an `AccessToken` response for testing our custom authorization policy)

Pilchie (approved):

PR #1807, opened 2024-09-17
Removed unsupported or undesirable credentials

### LINKED ISSUE #1805: Remove credential flows from public API surface, if not entirely

### LINKED ISSUE #1806: Replace azureauth_cli_credential with AzurePipelinesCredential
`AzurePipelinesCredential` replaces this. See <https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/azure_pipelines_credential.go> for an example.

heaths:
Fixes #1805 by removing unused flows (we'll use the remaining flows internally) and fixes #1806 by removing support for `azureauth`.


Cargo.toml
```diff
@@ -65,6 +65,7 @@ futures = "0.3"
 getrandom = { version = "0.2", features = ["js"] }
 hmac = { version = "0.12" }
 http-types = { version = "2.12", default-features = false }
+log = "0.4"
```
LarryOsterman, line 68: Do we want to use log or tracing or just standardize on one or the other? Tracing has the advantage that it has an opentelemetry Subscriber, which means you get otel basically for free.

log seems to be more common though.

heaths: This actually should've have been there. Seems I forgot to `rebase --onto` or something. Cleaning this up...

analogrelay (approved):
LGTM

sdk/core/azure_core/src/hmac.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-use crate::auth::Secret;
+use crate::authentication::Secret;
```
heaths, line 4: @JeffreyRichter what do you think about this? @LarryOsterman and I were talking and perhaps we should just go back to "auth", but I think all those years of .NET development and an anti-abbreviation stance worse on me. That said, Azure SDK for C++ uses namespace `Azure::Core::Credentials` so we could use "credentials" instead since everything therein support credentials.

*Naming is hard example 2,349,208*

heaths: /cc @analogrelay

heaths: At least with "auth", it's "deliciously ambiguous", as @LarryOsterman put it, in that it can mean "authentication" or "authorization". Still, if we want to avoid abbreviations I think "credentials" accomplishes the same.

analogrelay: I'm generally anti-abbreviation unless absolutely necessary, but it's not an immovable stance ;).

> At least with "auth", it's "deliciously ambiguous", as @LarryOsterman put it, in that it can mean "authentication" or "authorization". Still, if we want to avoid abbreviations I think "credentials" accomplishes the same.

Is there any authorization going on here? I'd presume not since this is a client. Authorization is the server's responsibility and if we have APIs to manage AuthZ in the future they would live in a separate module anyway.

If we do anticipate a need to handle authorization components here, then I'd lean towards a more general (if still ambiguous) term like "Security" rather than **intentionally** create ambiguity between two specific concepts.

RickWinter: I like authentication. I agree with Ashley on being anti-abbreviation (except if I'm writing Go).

RickWinter (approved):
As a seperate work item, I'd prefer we get rid of the SpecificAzureCredential and instead use the ChainedTokenCredential. 

IMO the DefaultAzureCredential is the quick and easy getting started credential and we shouldn't allow customization of that work flow. Once you get into the bucket of customized credential checking flow, you are squarely in the ChainedTokenCredential territory.

sdk/core/azure_core/src/hmac.rs
heaths: Technically, a service authenticates and authorizes (often separate services) so a client really does neither. We can't just say who we are with any authority. 😉 And the HTTP header is "Authorization" in requests or "WWW-Authenticate" in responses, and in with some policies we use both.

I'm starting to lean toward "credentials" like C++ if we agree on no abbreviations. We're idiomatic, but I still believe in consistency across languages if there's no other reason not to: it helps foster familiarity for polyglot projects.

heaths:
> As a seperate work item, I'd prefer we get rid of the SpecificAzureCredential and instead use the ChainedTokenCredential.

We have a meeting about that today. I could write up `SpecificAzureCredential` as an example - even use it in examples - but I still think showing `DefaultAzureCredential` in examples is counterproductive when we don't want people to use it. We know people copy and paste our samples, which is why many of us spent so much time making sure they are compiled and executed as tests - especially in toolchains like `dotnet` that didn't natively support such a thing.

While we may not want people to actually use a `SpecificAzureCredential` - since it seems we shouldn't even define one - it seems pretty clear that we want people to effectively write that themselves in code e.g.,

```text
let credential = if is_dev_environment() {
  DefaultAzureCredential
} else {
  ManagedIdentityCredential
}
```

/cc @christothes @scottaddie

sdk/core/azure_core/src/hmac.rs
JeffreyRichter: The crate defines credential types, right? If so, credentials seems the best to me.

christothes:
>  I still think showing DefaultAzureCredential in examples is counterproductive when we don't want people to use it. 

I don't think this is necessarily true. It's still the best for getting started and samples can't really be nuanced enough to provide more specific guidance outside the context of the Identity readme and API docs.

sdk/core/azure_core/src/hmac.rs
heaths: We're talking about a submodule of `azure_core` but, yes, it defines credential-*associated* types including `trait TokenCredential`, `struct AccessToken`, etc.

analogrelay: > Technically, a service authenticates and authorizes (often separate services) so a client really does neither. We can't just say who we are with any authority. 😉

image

> The crate defines credential types, right? If so, credentials seems the best to me.

I like this. `credentials` feels right to me.

ronniegeraghty:
> I don't think this is necessarily true. It's still the best for getting started and samples can't really be nuanced enough to provide more specific guidance outside the context of the Identity readme and API docs.

Agreeing with @christothes here. Cross SDK languages we are working to replace samples that don't use DefaultAzureCred with ones that do. This has been the guidance from the Identity SDK V team for helping customers get started in an easy and secure way.

sdk/core/azure_core/src/hmac.rs
heaths: Renamed to `azure_core::credentials`.

Issue #1811, opened 2024-09-18
[Cosmos] Item Create/Update/Delete


analogrelay:
#1809 will add querying for Cosmos Items. We need to add the ability to create/update/delete items. The API would be something like this, based on the Go SDK:

```rust
pub trait ContainerClientMethods {
    // ... other methods ...
    // Will fail if an item with the same ID already exists (POST with Upsert header false)
    fn create_item(&self, partition_key: impl Into<PartitionKey>, item: impl Serialize, options: Option<CreateItemOptions>);
    // Will replace an item with the same ID if one exists (POST with Upsert header true)
    fn upsert_item(&self, partition_key: impl Into<PartitionKey>, item: impl Serialize, options: Option<UpsertItemOptions>);
    // Will fail if an item with the same ID does NOT exist (PUT)
    fn replace_item(&self, partition_key: impl Into<PartitionKey>, item: impl Serialize, options: Option<ReplaceItemOptions>);
    // Will fail if an item with the specified ID does not exist (DELETE)
    fn delete_item(&self, partition_key: impl Into<PartitionKey>, item_id: impl Into<String>, options: Option<DeleteItemOptions>);
}
```

Notable exclusions: PATCHing documents (requires modelling the patch operations and will be handled separately)

Open Questions:
* Does each operation need it's own options? The Go SDK uses a single shared `ItemOptions`. We can do the same, but if we ever need operation-specific options, it's a breaking change. I'll check if the other SDKs generally accept that risk and use a shared options type. If they do, I think we can here as well.

analogrelay:
This is done!

PR #1814, opened 2024-09-19
[Cosmos] Add APIs to perform single-partition queries against a container

### LINKED ISSUE #1809: [Cosmos] Support for single-partition item queries
Add support for single-partition item queries, including:

* Specifying Strings, Numbers, or Null for Partition Key
* Specifying query parameters, supporting any JSON-serializable value

Notable exclusions here:

* Support for query options (though the `QueryOptions` type will be added, it will be empty)
* Unit tests for the query_items method itself (I'm opening a separate issue to start catch-up on tests using the mock transport)

analogrelay:
Closes #1809 

This adds the initial rudimentary APIs to perform a single-partition query against items in a container.

Some example code (see the `cosmos_query` example for a larger example):

```rust
let cosmos_client = CosmosClient::new(...);
let db_client = client.database_client("SampleDB");
let container_client = db_client.container_client("SampleContainer");
let mut items_pager =
    container_client.query_items::<serde_json::Value>("SELECT * FROM c", "some_partition_key", None)?;

// ".next()" requires importing futures::StreamExt
while let Some(page) = items_pager.next().await {
    for item in page?.items {
        println!("    * {:?}", item);
    }
}
```

A few pieces of note:

1. A query can be parameterized. To do that, instead of passing the raw string in, you'd call `Query::from("...")` and then call `.with_parameter("@name", "value")` on the resulting `Query` to build up the query and parameters

2. We only support single-partition queries. This is aligned with the support we currently provide in other SDKs that only support Gateway mode. Over time we can expand this to cross-partition queries, but there are significant limitations to those.

3. For serializing Partition Keys we have to do something a little custom and can't just use `serde_json` as-is. The partition key for a query is encoded as an HTTP header, which means we have to escape any non-ASCII characters because HTTP headers may not contain non-ASCII characters. Serde's default JSON serialization does not perform that escaping, it allows non-ASCII UTF-8 characters to pass through unescaped. Even Rust's built-in escaping is unsuitable because it uses the format `\u{XXXX}`, when JSON requires `\uXXXX` (sans-`{}`). So we do some custom stuff. Importantly, we still use the standard library's logic to encode characters as UTF-16, which handles surrogate pairs properly.

4. **UPDATE: I reverted this change and decided to leave `Pageable` as-is** I changed `Pageable` to include a lifetime parameter. Previously, it expected the provided closure to return a future with the `'static` lifetime, which would require the Future to own all necessary data. However, the implementation I used for `query_items` required that the future borrowed `self.pipeline`, so it could use the same HTTP pipeline to make the requests. That isn't _totally_ necessary. I could clone the pipeline and give the clone to the closure, but that seemed heavier to me, so I hesitated to do that. Though, after the conversations we've had about lifetimes, I'm waffling on that and considering rolling that change back and requiring that the `Pageable` hold its own HTTP pipeline and avoid being coupled to the lifetime of the client that created it. I'll continue to look into that.

There are also fairly substantial API docs in place. We don't have a great way to preview those docs (maybe we should? publish updated docs to an Azure Static Web App on PR build?) but you can run `cargo doc --package azure_data_cosmos --open`, if you're on a local machine or RDP, or use #1808, if on a codespace, to view them.

### DIFF
--- a/eng/dict/rust-custom.txt
+++ b/eng/dict/rust-custom.txt
@@ -3,3 +3,4 @@ newtype
 repr
 rustc
 rustls
+turbofish
--- a/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
@@ -108,7 +108,7 @@ pub trait ContainerClientMethods {
     /// # }
     /// ```
     ///
-    /// See [`PartitionKey`] for more information on how to specify a partition key, and [`Query`] for more information on how to specify a query.
+    /// See [`PartitionKey`](crate::PartitionKey) for more information on how to specify a partition key, and [`Query`] for more information on how to specify a query.
     fn query_items<T: DeserializeOwned + Send>(
         &self,
         query: impl Into<Query>,
--- a/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
@@ -1,9 +1,10 @@
 use crate::{
     constants,
     models::{ContainerProperties, QueryResults},
+    options::{QueryOptions, ReadContainerOptions},
     pipeline::ResourceType,
     utils::WithAddedPathSegments,
-    CosmosClient, PartitionKey, Query, QueryItemsOptions, ReadContainerOptions,
+    CosmosClient, PartitionKey, Query,
 };
 
 use azure_core::{headers::HeaderValue, Context, Request};
@@ -57,13 +58,13 @@ pub trait ContainerClientMethods {
         &self,
         query: impl Into<Query>,
         partition_key: impl Into<PartitionKey>,
-        options: Option<QueryItemsOptions>,
+        options: Option<QueryOptions>,
     ) -> azure_core::Result<azure_core::Pageable<QueryResults<T>, azure_core::Error>>;
 }
 
 /// A client for working with a specific container in a Cosmos DB account.
 ///
-/// You can get a `Container` by calling [`DatabaseClient::container_client()`](DatabaseClient::container_client()).
+/// You can get a `Container` by calling [`DatabaseClient::container_client()`](crate::clients::DatabaseClient::container_client()).
 pub struct ContainerClient {
     base_url: Url,
     root_client: CosmosClient,
@@ -102,7 +103,7 @@ impl ContainerClientMethods for ContainerClient {
 
         #[allow(unused_variables)]
         // This is a documented public API so prefixing with '_' is undesirable.
-        options: Option<QueryItemsOptions>,
+        options: Option<QueryOptions>,
     ) -> azure_core::Result<azure_core::Pageable<QueryResults<T>, azure_core::Error>> {
         // Represents the raw response model from the server.
         // We'll use this to deserialize the response body and then convert it to a more user-friendly model.
--- a/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
@@ -6,13 +6,12 @@ use crate::{
     models::{ContainerProperties, QueryResults},
     options::{QueryOptions, ReadContainerOptions},
     pipeline::{CosmosPipeline, ResourceType},
-    utils::WithAddedPathSegments,
+    utils::AppendPathSegments,
     Query, QueryPartitionStrategy,
 };
 
 use azure_core::{headers::HeaderValue, Context, Request};
 use serde::{de::DeserializeOwned, Deserialize};
-use typespec_client_core::http::AppendPathSegments;
 use url::Url;
 
 #[cfg(doc)]
@@ -129,7 +128,7 @@ pub struct ContainerClient {
 impl ContainerClient {
     pub(crate) fn new(pipeline: CosmosPipeline, database_url: &Url, container_name: &str) -> Self {
         let mut container_url = database_url.clone();
-        container_url.append_path_segments(&["colls", container_name]);
+        container_url.append_path_segments(["colls", container_name]);
 
         Self {
             container_url,
@@ -179,7 +178,8 @@ impl ContainerClientMethods for ContainerClient {
             }
         }
 
-        let url = self.container_url.with_added_path_segments(vec!["docs"]);
+        let mut url = self.container_url.clone();
+        url.append_path_segments(["docs"]);
         let mut base_req = Request::new(url, azure_core::Method::Post);
 
         base_req.insert_header(constants::QUERY, "True");
--- a/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
@@ -32,11 +32,8 @@ pub trait ContainerClientMethods {
     ///
     /// ```rust,no_run
     /// # async fn doc() {
-    /// # use azure_data_cosmos::{CosmosClient, CosmosClientMethods, clients::DatabaseClientMethods, clients::ContainerClientMethods};
-    /// # let credential = azure_identity::create_default_credential().unwrap();
-    /// # let client = CosmosClient::new("https://myaccount.documents.azure.com/", credential, None).unwrap();
-    /// # let db_client = client.database_client("my_database");
-    /// # let container_client = db_client.container_client("my_container");
+    /// # use azure_data_cosmos::clients::{ContainerClient, ContainerClientMethods};
+    /// # let container_client: ContainerClient = panic!("this is a non-running example");
     /// let response = container_client.read(None)
     ///     .await.unwrap()
     ///     .deserialize_body()
@@ -71,11 +68,8 @@ pub trait ContainerClientMethods {
     ///
     /// ```rust,no_run
     /// # async fn doc() {
-    /// # use azure_data_cosmos::{CosmosClient, CosmosClientMethods, clients::DatabaseClientMethods, clients::ContainerClientMethods};
-    /// # let credential = azure_identity::create_default_credential().unwrap();
-    /// # let client = CosmosClient::new("https://myaccount.documents.azure.com/", credential, None).unwrap();
-    /// # let db_client = client.database_client("my_database");
-    /// # let container_client = db_client.container_client("my_container");
+    /// # use azure_data_cosmos::clients::{ContainerClient, ContainerClientMethods};
+    /// # let container_client: ContainerClient = panic!("this is a non-running example");
     /// #[derive(serde::Deserialize)]
     /// struct Customer {
     ///     id: u64,
@@ -92,11 +86,8 @@ pub trait ContainerClientMethods {
     ///
     /// ```rust,no_run
     /// # async fn doc() {
-    /// # use azure_data_cosmos::{CosmosClient, CosmosClientMethods, clients::DatabaseClientMethods, clients::ContainerClientMethods, Query};
-    /// # let credential = azure_identity::create_default_credential().unwrap();
-    /// # let client = CosmosClient::new("https://myaccount.documents.azure.com/", credential, None).unwrap();
-    /// # let db_client = client.database_client("my_database");
-    /// # let container_client = db_client.container_client("my_container");
+    /// # use azure_data_cosmos::{Query, clients::{ContainerClient, ContainerClientMethods}};
+    /// # let container_client: ContainerClient = panic!("this is a non-running example");
     /// #[derive(serde::Deserialize)]
     /// struct Customer {
     ///     id: u64,
--- a/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
@@ -187,16 +187,21 @@ impl ContainerClientMethods for ContainerClient {
         );
         base_req.set_json(&query.into())?;
 
+        // We have to double-clone here.
+        // First we clone the pipeline to pass it in to the closure
+        let pipeline = self.root_client.pipeline.clone();
         Ok(azure_core::Pageable::new(move |continuation| {
+            // Then we have to clone it again to pass it in to the async block.
+            // This is because Pageable can't borrow any data, it has to own it all.
+            // That's probably good, because it means a Pageable can outlive the client that produced it, but it requires some extra cloning.
+            let pipeline = pipeline.clone();
             let mut req = base_req.clone();
             async move {
                 if let Some(continuation) = continuation {
                     req.insert_header(constants::CONTINUATION, continuation);
                 }
 
-                let resp = self
-                    .root_client
-                    .pipeline
+                let resp = pipeline
                     .send(Context::new(), &mut req, ResourceType::Items)
                     .await?;
 
--- a/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
@@ -1,8 +1,13 @@
 use crate::{
-    models::ContainerProperties, pipeline::ResourceType, CosmosClient, ReadContainerOptions,
+    constants,
+    models::{ContainerProperties, QueryResults},
+    pipeline::ResourceType,
+    utils::WithAddedPathSegments,
+    CosmosClient, PartitionKey, Query, QueryItemsOptions, ReadContainerOptions,
 };
 
-use azure_core::{Context, Request};
+use azure_core::{headers::HeaderValue, Context, Request};
+use serde::{de::DeserializeOwned, Deserialize};
 use url::Url;
 
 #[cfg(doc)]
@@ -28,7 +33,7 @@ pub trait ContainerClientMethods {
     /// let credential = azure_identity::create_default_credential().unwrap();
     /// let client = CosmosClient::new("https://myaccount.documents.azure.com/", credential, None).unwrap();
     /// let db_client = client.database_client("my_database");
-    /// let container_client = client.container_client("my_container");
+    /// let container_client = db_client.container_client("my_container");
     /// let response = container_client.read(None)
     ///     .await.unwrap()
     ///     .deserialize_body()
@@ -40,6 +45,20 @@ pub trait ContainerClientMethods {
         &self,
         options: Option<ReadContainerOptions>,
     ) -> azure_core::Result<azure_core::Response<ContainerProperties>>;
+
+    /// Executes a single-partition query against items in the container.
+    ///
+    /// # Arguments
+    ///
+    /// * `query` - The query to execute.
+    /// * `partition_key` - The partition key to scope the query on.
+    /// * `options` - Optional parameters for the request.
+    fn query_items<T: DeserializeOwned + Send>(
+        &self,
+        query: impl Into<Query>,
+        partition_key: impl Into<PartitionKey>,
+        options: Option<QueryItemsOptions>,
+    ) -> azure_core::Result<azure_core::Pageable<QueryResults<T>, azure_core::Error>>;
 }
 
 /// A client for working with a specific container in a Cosmos DB account.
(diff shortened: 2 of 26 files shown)


analogrelay:
> I'm waffling on that and considering rolling that change back and requiring that the Pageable hold its own HTTP pipeline and avoid being coupled to the lifetime of the client that created it. I'll continue to look into that.

I was just reminded why I went with the lifetime approach when I looked at this earlier in the week. The problem is that if we continue to require that `Pageable` hold only `'static` lifetime values, we have to clone the pipeline _in each invocation_. So you end up having to write this:

```rust
let pipeline = self.pipeline.clone(); // A clone so we can move into the closure.
Pageable::new(move |continuation| {
  // Another clone so we can move in to the `async` block.
  async move {
    pipeline.send(...).await...
  }
});
```

Currently [`async` closures are unstable](https://github.com/rust-lang/rust/issues/62290), which forces us to use an `async` block inside a regular closure. I think that may be part of the requirement for the clone.

Another option might be to allow a `Pageable` to hold on to an `azure_core::Pipeline` itself (perhaps using an `Option` so that it's up to the caller if they want to do that). I don't mind cloning the pipeline to pass it in to the `Pageable`, but I'm concerned about doing that on every request.

Having said that, [it looks like maybe that's what the existing unofficial SDK code does](https://github.com/Azure/azure-sdk-for-rust/blob/b16e8790a1d0ebe608584db675b4042796b35caf/sdk/data_cosmos/src/operations/query_documents.rs#L46)? Maybe I'm fussing about a non-issue?

Would be interested to hear any thoughts you have here @heaths. Putting a lifetime parameter on `Pageable` allows it to borrow the client itself, in effect you'd write client methods like this:

```rust
fn query_items<'a>(&'a self, query: &str) -> Pageable<'a, ...>;
```

In practice, the lifetimes are all elided, so you don't see them. The main impacts of putting a lifetime parameter on `Pageable` that I can see are:

1. A `Pageable` cannot outlive the client that generated it. But I can't really think of a viable scenario where you'd _need_ it to...
2. Writing out the `Pageable` type **can** sometimes be more verbose (requires lifetime parameters), though as I said above, lifetime elision [often has you covered there](https://doc.rust-lang.org/reference/lifetime-elision.html).

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -0,0 +1,230 @@
...
+    ///     .with_parameter("@customer_id", 42).unwrap();
+    /// let items = container_client.query_items::<Customer>(query, "some_partition_key", None).unwrap();
+    /// # }
+    /// ```
+    ///
+    /// See [`PartitionKey`] for more information on how to specify a partition key, and [`Query`] for more information on how to specify a query.
+    fn query_items<T: DeserializeOwned + Send>(
+        &self,
+        query: impl Into<Query>,
+        partition_key: impl Into<PartitionKey>,
```
Pilchie, line 113: Why did we choose to elevate `partition_key` to be a top level parameter?  In .NET's `Container.GetItemQueryIterator`(https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.container.getitemqueryiterator?view=azure-dotnet) it's part of the `QueryOptions`.

sdk/cosmos/azure_data_cosmos/src/query.rs
```diff
@@ -0,0 +1,157 @@
...
+/// This includes arrays and objects, if they implement [`serde::Serialize`]:
+///
+/// ```rust
+/// # use azure_data_cosmos::{Query, NullValue};
+/// #[derive(serde::Serialize)]
+/// struct CustomerInfo {
+///     id: u64,
+///     name: String
+/// }
+/// let query = Query::from("SELECT * FROM c WHERE ...")
```
Pilchie, line 42: Should we fill in `...` here?

sdk/cosmos/azure_data_cosmos/src/query.rs
```diff
@@ -0,0 +1,157 @@
...
+#[derive(Debug, Serialize)]
+pub struct Query {
+    query: String,
+
+    #[serde(skip_serializing_if = "Vec::is_empty")] // Don't serialize an empty array.
+    parameters: Vec<QueryParameter>,
+}
+
+impl Query {
+    /// Creates a new [`Query`] with the same text and the same parameters, with one additional parameter added.
```
Pilchie, line 54: Technically not a new Query, right?  Same query that has been mutated?

Pilchie (approved):

heaths:
Cloning the pipeline is designed to be cheap. It's a `Vec<Arc<dyn Policy>>`. For this reason, we already have a guideline about cloning it for subclients. I don't see why pageables - subclients, really - shouldn't do the same. Introducing lifetimes isn't out of question, but can introduce problems for callers. What if they want to return a pageable to complete later? It's infectious.

For now until we get substantial feedback otherwise, I think we should just clone the pipeline.

heaths:
> For serializing Partition Keys we have to do something a little custom and can't just use serde_json as-is. The partition key for a query is encoded as an HTTP header, which means we have to escape any non-ASCII characters because HTTP headers may not contain non-ASCII characters. Serde's default JSON serialization does not perform that escaping, it allows non-ASCII UTF-8 characters to pass through unescaped. Even Rust's built-in escaping is unsuitable because it uses the format \u{XXXX}, when JSON requires \uXXXX (sans-{}). So we do some custom stuff. Importantly, we still use the standard library's logic to encode characters as UTF-16, which handles surrogate pairs properly.

IIRC, we have some helper functions already that might be close or even do that. IF they don't, I suggest actually adding some globals. You definitely won't be the only ones writing "params" as headers. That's another reason why I'm pretty sure we already have serde de/serialization helpers for that.

sdk/cosmos/azure_data_cosmos/src/options/cosmos_client_options.rs
```diff
@@ -0,0 +1,78 @@
...
+/// Obtain a [`CosmosClientOptionsBuilder`] by calling [`CosmosClientOptions::builder()`]
+#[derive(Default)]
+pub struct CosmosClientOptionsBuilder(CosmosClientOptions);
+
+impl CosmosClientOptionsBuilder {
+    /// Builds a [`CosmosClientOptions`] from the builder.
+    ///
+    /// This does not consume the builder, and can be called multiple times.
+    pub fn build(&self) -> CosmosClientOptions {
+        self.0.clone()
```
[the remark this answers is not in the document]
analogrelay: `.0`/`.1`/etc is the syntax for accessing items in a tuple. So `("a", "b", "c").1 == "b"`, for example.

The `CosmosClientOptionsBuilder` is a ["tuple-struct"](https://doc.rust-lang.org/book/ch05-01-defining-structs.html#using-tuple-structs-without-named-fields-to-create-different-types), which is a common pattern to use when you have a struct with only one member (though you can have multiple tuple members, i.e. `pub struct Foo(u32, String, ...)` which would be addressed as `self.0`, `self.1`, etc.)

So, this is accessing the `CosmosClientOptions` that is the only field within the `CosmosClientOptionsBuilder`, and then cloning it to return to the user.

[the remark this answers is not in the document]
Pilchie: Thanks for explaining!  I had used tuples in rust, but hadn't done "tuple-structs".

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
analogrelay: The main reason is that we _require_ a partition key. Currently, we are not supporting cross-partition queries, similar to the Go SDK. Therefore, we require a `partition_key`. Also, I modelled this API on [the similar API in the Go SDK](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/azcosmos/cosmos_container.go#L491). which requires the partition key be provided.

If we want to support cross-partition queries later, we could probably still use this API. Rust doesn't support overloading, so we have to keep the arguments the same, or introduce a new method. The `PartitionKey` struct this expects contains a `Vec<PartitionKeyValue>` which is expected to contain 1 or more keys (to support hierarchical keys). If you're doing a cross-partition query, we could update `PartitionKey` to allow 0 keys to be provided, and add an impl of `From<()> for PartitionKey` that would allow you to write this:

```rust
container.query_items("SELECT * FROM c", (), None); // The "()" indicates no partition key filter
```

eng/dict/rust-custom.txt
```diff
@@ -3,3 +3,4 @@ newtype
 repr
 rustc
 rustls
+turbofish
```
heaths, line 6: Is this established enough? It is listed [here](https://doc.rust-lang.org/book/appendix-02-operators.html?highlight=turbofish#:~:text=generic%20type%20parameters.-,Table%20B%2D4%3A%20Generics,-Symbol) so maybe.

sdk/cosmos/azure_data_cosmos/src/query.rs
analogrelay: Technically, you are correct ;). However, I think it's best to keep this as-is. The method consumes `self` by value, so the original query is moved into this method and no longer available to the caller. The fact that what we _actually_ do is take that object and add a parameter to it is mostly an implementation detail at that point. We could just as easily be creating a brand new one.

I think there's probably a rephrase here that could be clearer though, since the nuances there aren't super important to the caller.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -0,0 +1,230 @@
...
+/// Defines the methods provided by a [`ContainerClient`]
+///
+/// This trait is intended to allow you to mock out the `ContainerClient` when testing your application.
+/// Rather than depending on `ContainerClient`, you can depend on a generic parameter constrained by this trait, or an `impl ContainerClientMethods` type.
+pub trait ContainerClientMethods {
+    /// Reads the properties of the container.
+    ///
+    /// # Arguments
+    ///
+    /// * `options` - Optional parameters for the request.
```
heaths, line 29: Even I forgot I wrote those. Glad you caught it! 🎉

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
heaths: If it's required - which given the param type, it is - it should be a parameter. If it's optional it should go into parameters.

.NET also establishes 6 as the max number of parameters (can't remember if inclusive or exclusive) of when all parameters should go into an options bag, though the name of that bag has changed over the years. For Rust, I'd prefer we make the method options required but still called `<MethodName>Options` when that happens. You good with that @JeffreyRichter, @RickWinter?

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -0,0 +1,230 @@
...
+        options: Option<QueryOptions>,
+    ) -> azure_core::Result<azure_core::Pageable<QueryResults<T>, azure_core::Error>>;
+}
+
+/// A client for working with a specific container in a Cosmos DB account.
+///
+/// You can get a `Container` by calling [`DatabaseClient::container_client()`](crate::clients::DatabaseClient::container_client()).
+pub struct ContainerClient {
+    base_url: Url,
+    root_client: CosmosClient,
```
heaths, line 123: This might end up being more expensive. Why not just clone the pipeline, which is cheap? If you have strings on the client e.g., the `endpoint`, you presumably just cloned that as well.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -0,0 +1,230 @@
...
+}
+
+impl ContainerClient {
+    pub(crate) fn new(root_client: CosmosClient, database_url: &Url, container_name: &str) -> Self {
```
heaths, line 127: You're taking ownership of the `CosmosClient`? So you can't get multiple `ContainerClient`s from a single root client?

sdk/cosmos/azure_data_cosmos/src/lib.rs
```diff
@@ -7,12 +10,35 @@
...
+    /// Serializes the [`NullValue`].
+    ///
+    /// This will always produce the value `null` (or equivalent in whatever format the type is being serialized to).
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_none()
+    }
+}
```
heaths, line 44: Would this be useful in `typespec_client_core`? We could always move it later, though. Just putting it out there.

heaths: We do - or did? - have a serialization helper that serialized explicit nulls. Still in `main` at least.

sdk/cosmos/azure_data_cosmos/src/options/cosmos_client_options.rs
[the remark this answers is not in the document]
heaths: Personally, I dislike tuple-structs when the contained number of values is > 1 because remembering which is 0, 1, etc., it's a headache. That said, I feel no reason to have a guideline about this for Azure SDKs. Yet. If people start abusing it (say, 3 or more and often) we might have to because it gets hard to review and maintain.

sdk/typespec/typespec_client_core/src/http/response.rs
```diff
@@ -25,6 +25,12 @@ pub trait Model: Sized {
     /// The server may still be sending data, so it's up to implementors whether they want to wait for the entire body to be received or not.
     /// For example, a type representing a simple REST API response will want to wait for the entire body to be received and then parse the body.
     /// However, a type representing the download of a large file, may not want to do that and instead prepare to stream the body to a file or other destination.
+    #[cfg(not(target_arch = "wasm32"))]
+    fn from_response_body(
+        body: ResponseBody,
+    ) -> impl Future<Output = crate::Result<Self>> + Send + Sync;
+
+    #[cfg(target_arch = "wasm32")]
```
heaths, line 33: Good catch.

sdk/cosmos/azure_data_cosmos/src/query.rs
```diff
@@ -0,0 +1,157 @@
...
+        }
+        let obj_param = ObjectParameter {
+            name: "foo".into(),
+            value: "bar".into(),
+        };
+        let null_option: Option<&str> = None;
+
+        let query = Query::from("SELECT * FROM c")
+            .with_parameter("string_param", "value1")
+            .unwrap()
```
heaths, line 141: Nit: would it be worth it just returning a result with a boxed error to avoid calling all these unwraps? This reads like a spoiled kid on Christmas.

heaths (approved):
Overall, I love it! A few nits and questions, but nothing blocking.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
analogrelay: The code that calls this does the cloning, but I may refactor that given your suggestion to just clone the pipeline

sdk/cosmos/azure_data_cosmos/src/lib.rs
analogrelay: I'm not seeing what you're referring to in `main`, but I'm definitely open to promoting this to `typespec_client_core` if you think it's useful there.

sdk/cosmos/azure_data_cosmos/src/query.rs
analogrelay: TIL that Rust tests can return `Result<T, E>` 😅 .

eng/dict/rust-custom.txt
analogrelay: It's certainly official enough (IMO) to be in our cspell dict, so we can use it freely in internal comments. I'm actually using it in a doc comment, which also explains the term:

```rust
    /// We recommend using "turbofish" syntax (`query_items::<SomeTargetType>(...)`) to specify the target type, as it makes type inference easier.
```

I think that's a very reasonable doc comment and it's entirely appropriate to reference a common community term for the syntax while also explaining it. In fact, I'll put a link to your reference in there as an "official" citation ;).

analogrelay:
> **Update:** Never mind. I'm conflating https://github.com/Azure/azure-sdk-for-rust/blob/feature/track2/sdk/typespec/typespec_client_core/src/date/mod.rs but the idea could be the same, along with implementing `AsHeader` if appropriate.

I initially started by implementing `AsHeader`, but `AsHeader` is infallible and serializing a partition key to a header is fallible (serializing to JSON can fail). I could refactor later, but this is largely internal code (converting `PartitionKey` to a header is internal, that is).

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
analogrelay: Ok, the next iteration will have a slightly modified approach here. Combining this with my realization that serde treats `()` as `null`, so we can use that to represent null partition keys (which means it can't represent "cross-partition" queries here), I've adjusted my approach a bit.

Instead of `Into<PartitionKey>`, we accept `Into<QueryPartitionStrategy>`. There is only one strategy available today, `SinglePartition`. Any value that is `Into<PartitionKey>` is automatically `Into<QueryPartitionStrategy>` and results in `QueryPartitionStrategy::SinglePartition(...)`.

I did a little bit of API evolution "testing" and adding a second variant, `QueryPartitionStrategy::CrossPartition` would then allow us to call `query_items` like this:

```rust
container.query_items("SELECT * FROM c", QueryPartitionStrategy::CrossPartition, None);
```

While still preserving all these other ways of calling it:

```rust
container.query_items("SELECT * FROM c", (), None); // single "null" partition key, i.e. `[null]`
container.query_items("SELECT * FROM c", "my_key", None); // single partition key, i.e. `["my_key"]`
container.query_items("SELECT * FROM c", ("parent", "child"), None); // hierarchical partition key, i.e. `["parent", "child"]`

let pk = PartitionKey::from(("parent", "child"));
container.query_items("SELECT * FROM c", pk, None); // explicit partition key object
container.query_items("SELECT * FROM c", QueryPartitionStrategy::SinglePartition(pk), None); // explicit partition strategy
```

analogrelay:
The next iteration is live, I think I've covered all the feedback. I'm going to try and land this soon so that I can move on to item CRUD (https://github.com/Azure/azure-sdk-for-rust/issues/1811)

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
Pilchie: I'm okay with this approach, but I feel like the default approach is for queries to be cross-partition, and this will be a little bit of boiler-plate on all of those.

@kirankumarkolli or @jcocchi - any thoughts on the shape of this API?

Pilchie (approved):

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
analogrelay: I'd also be open to enabling support for cross-partition queries now, **limited** to the support the Gateway provides for cross-partition queries (so no aggregations, etc), by just setting the `x-ms-documentdb-query-enablecrosspartition` header. The previous unofficial client [supported that](https://docs.rs/azure_data_cosmos/latest/azure_data_cosmos/prelude/struct.QueryDocumentsBuilder.html#method.query_cross_partition).

If we did that, I'd be OK with moving `partition_key` to the Options and "defaulting" to cross-partition if it's not specified. That's actually what the previous unofficial client versions did

heaths (approved):

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
jcocchi: The problem with allowing *some* cross partition queries now is validation. IIRC the Gateway doesn't reject query shapes it can't handle, it just returns results that may be incorrect because they require extra client-side processing. We didn't open up limited cross partition query support on Go because we chose query correctness over adding functionality users may not read all the limitations of. However, Go allows users to explicitly set this header themselves to unblock if they need to (presumably after understanding the limitations).

analogrelay: I'm going to go ahead with this as-is for now. We will need to do some API review for this later, and can have another opportunity to dive into it there.

Issue #1810, opened 2024-09-18
Cosmos: Client testing using test proxy


analogrelay:
So far, I haven't added tests for Client methods using the test proxy. I'd like to catch up on that soon so that we can keep up with it iteratively and avoid big debt.

analogrelay:
We have test proxy tests now!

PR #1815, opened 2024-09-23
Refactor `mock_transport` into `azure_testing` and add initial Cosmos record/replay tests

analogrelay:
Part of #1810 

This PR has two commits, in reverse order:

1. A test for the current "read database metadata" operation on CosmosClient that uses the record/replay "mock" transport
2. Refactors to the `mock_transport` crate, including:
    * Rename to `azure_testing`, since I expect we'll want to add other test utilities here. Certainly, open to other names, or just not renaming.
    * Support for more readable (and thus reviewable) body encoding. For recorded responses, we can encode them as JSON. For recorded requests it's more difficult to do that because we need to be able to compare bytes during replay mode, but we can record them as UTF-8.
    * A little more sanitization. We still redact tokens and the base URL properly, but this refactoring adds a single place to add more headers to be "sanitized" (including the `content-location` header, which Cosmos returns)
    * A `context!` macro intended to be used to capture the current test "context" (the crate path, the name of the module) to allow for placement of the `tests/data/transactions` directory that contains recorded transactions
    * Related to that: We place test transactions in a directory _relative to the crate containing the test_ instead of at the workspace root (which is what the previous version did)
    * A mock "credential" that ensures that we don't try to authenticate against AAD when in replay mode, only when in record mode.

The easiest way to see an example test is to look at the `sdk/cosmos/azure_data_cosmos/tests` directory. With some of these adjustments I think we've got a pretty smooth looking record/replay experience. I'm a fan of the record/replay model here, since it allows us to have tests that don't depend on a real service, but **can** be validated against real service responses.

I'd also be interested in seeing if we can add a mechanism to actually run these tests against a real Cosmos DB instance and/or emulator at some interval (say a daily, or even less frequent, CI run), but that wouldn't be necessary for day-to-day dev, just to run occasionally to catch cases where the service behavior changes from our expectations (which would be Very Bad ™️ and should therefore be Quite Rare ™️ I hope). I don't have a specific suggestion there yet though.


Pilchie (approved):
Took an overall look, but you should get approval from core team.

heaths:
I'm hesitant to take these. `mock_transport` is not our final recording solution. There's no point is supporting or updating them unless you really need it now.

heaths (changes requested):
Not ready. Need to discuss since we're not sticking with `mock_transport` for recording. We probably will have *a* mock transport of some kind for one-offs like we do in .NET, but the Test Proxy work will underscore recording and playback like it does for all our languages because it provides a lot of functionality and security out of the box.

analogrelay:
Fair enough, I'm happy to withdraw this. My main concern is getting some kind of integration testing in place. I may look at adding some opt-in tests in the Cosmos crate that run against a real account so that I can get something I can run manually to have confidence the core scenarios are still working, but I'll be happy to transition those to something more consistent with the other SDKs.

heaths:
We have an `e2e_test` feature that was used before and I believe @LarryOsterman used for Event Hubs. That said, I want to get rid of that eventually. Conversely, you could write examples to run manually that take actual env vars and - at the end of the day - they will serve as actual, useful examples.

I hope to get to test proxy work soon. We're close to generating Key Vault, and test proxy is a pretty simple TypeSpec. That said, there's still quite a bit of work to properly redact everything by default and Rust makes it harder with `Display` and `Debug` both emitting everything by default. More on that in a subsequent issue...

analogrelay:
Closing this for now. I'll follow up separately on plans for integration testing. I'd like to try and stay on top of that so we don't build a big debt and if there's some infra work there that I can help out on, I'm happy to do it.

PR #1818, opened 2024-09-24
Give CODEOWNERS file an owner in CODEOWNERS

JimSuplizio:
Give the CODEOWNERS file owners. 

@RickWinter as per your instructions you and @ronniegeraghty, along with azure-sdk-eng are added to the owners.

When the linter pipeline and github actions have been added to the repository, those owners will also need to be added as part of the same block.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -35,3 +35,4 @@
 # Eng Sys
 ###########
 /eng/                    @weshaggard @heaths @RickWinter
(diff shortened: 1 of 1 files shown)


heaths (changes requested):
This needs to target the `feature/track2` branch.

JimSuplizio:
Closing out this one to open a new one targeting the correct branch.

PR #1822, opened 2024-09-25
complete session implementation; C++ related changes

LarryOsterman:
This pull request includes several changes to the `AmqpSessionApis` implementation and related components to improve functionality and code maintainability. The key changes involve refactoring how session options are accessed, implementing the `end` method, and modifying the `AmqpSessionOptions` builder.

### Improvements to session options access:
* Changed `AmqpSessionOptions` to use getter methods instead of directly accessing fields. (`sdk/core/azure_core_amqp/src/fe2o3/session.rs`, [[1]](diffhunk://#diff-786c4908e70ea4001d6ff44171a9e39bef40cf7c3b49303cc318602a03242d5fL55-R78) [[2]](diffhunk://#diff-786c4908e70ea4001d6ff44171a9e39bef40cf7c3b49303cc318602a03242d5fL88-R90)
* Updated `AmqpSessionOptions` to include getter methods for all fields. (`sdk/core/azure_core_amqp/src/session.rs`, [sdk/core/azure_core_amqp/src/session.rsL18-R65](diffhunk://#diff-00d41b21de5b22a65122dfb158fb613100a7729fe98102145d2eeef0be58be70L18-R65))

### Implementation of `end` method:
* Implemented the `end` method in `AmqpSessionApis` to properly end the session. (`sdk/core/azure_core_amqp/src/fe2o3/session.rs`, [sdk/core/azure_core_amqp/src/fe2o3/session.rsL101-R111](diffhunk://#diff-786c4908e70ea4001d6ff44171a9e39bef40cf7c3b49303cc318602a03242d5fL101-R111))

### Modifications to `AmqpSessionOptions` builder:
* Refactored the `AmqpSessionOptionsBuilder` to use mutable references and clone options for the `build` method. (`sdk/core/azure_core_amqp/src/session.rs`, [sdk/core/azure_core_amqp/src/session.rsL86-R167](diffhunk://#diff-00d41b21de5b22a65122dfb158fb613100a7729fe98102145d2eeef0be58be70L86-R167))
* Updated the `test_amqp_session_options_builder` test to reflect changes in the builder pattern. (`sdk/core/azure_core_amqp/src/session.rs`, [sdk/core/azure_core_amqp/src/session.rsL141-R180](diffhunk://#diff-00d41b21de5b22a65122dfb158fb613100a7729fe98102145d2eeef0be58be70L141-R180))

### Other changes:
* Changed the `Uuid` import to a type alias. (`sdk/core/azure_core_amqp/src/lib.rs`, [sdk/core/azure_core_amqp/src/lib.rsL19-R19](diffhunk://#diff-78b3c81b4d51cf0fa2b3cfa74f93bd2868269a62dea18469a016980bb1a212d0L19-R19))

### DIFF
--- a/sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
+++ b/sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
@@ -46,7 +46,12 @@ impl AmqpClaimsBasedSecurityApis for Fe2o3ClaimsBasedSecurity {
             .attach(session.borrow_mut())
             .await
             .map_err(AmqpManagementAttach::from)?;
-        self.cbs.set(Mutex::new(cbs_client)).unwrap();
+        self.cbs
+            .set(Mutex::new(cbs_client))
+            .map_err(azure_core::Error::message(
+                azure_core::error::ErrorKind::Other,
+                "Claims Based Security is already set.",
+            ))?;
         Ok(())
     }
 
@@ -69,12 +74,22 @@ impl AmqpClaimsBasedSecurityApis for Fe2o3ClaimsBasedSecurity {
                     .to_offset(time::UtcOffset::UTC)
                     .unix_timestamp()
                     .checked_mul(1_000)
-                    .unwrap(),
+                    .ok_or_else(|| {
+                        azure_core::Error::message(
+                            azure_core::error::ErrorKind::Other,
+                            "Unable to convert time to unix timestamp.",
+                        )
+                    })?,
             )),
         );
         self.cbs
             .get()
-            .unwrap()
+            .ok_or_else(|| {
+                azure_core::Error::message(
+                    azure_core::error::ErrorKind::Other,
+                    "Claims Based Security was not set.",
+                )
+            })?
             .lock()
             .await
             .borrow_mut()
--- a/sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
+++ b/sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
@@ -22,11 +22,11 @@ pub(crate) struct Fe2o3ClaimsBasedSecurity {
 }
 
 impl Fe2o3ClaimsBasedSecurity {
-    pub fn new(session: AmqpSession) -> Self {
-        Self {
+    pub fn new(session: AmqpSession) -> Result<Self> {
+        Ok(Self {
             cbs: OnceLock::new(),
-            session: session.implementation.get(),
-        }
+            session: session.implementation.get()?,
+        })
     }
 }
 
@@ -46,12 +46,12 @@ impl AmqpClaimsBasedSecurityApis for Fe2o3ClaimsBasedSecurity {
             .attach(session.borrow_mut())
             .await
             .map_err(AmqpManagementAttach::from)?;
-        self.cbs
-            .set(Mutex::new(cbs_client))
-            .map_err(azure_core::Error::message(
+        self.cbs.set(Mutex::new(cbs_client)).map_err(|_| {
+            azure_core::Error::message(
                 azure_core::error::ErrorKind::Other,
                 "Claims Based Security is already set.",
-            ))?;
+            )
+        })?;
         Ok(())
     }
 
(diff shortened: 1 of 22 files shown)


sdk/core/azure_core_amqp/src/fe2o3/session.rs
```diff
@@ -98,6 +100,14 @@ impl AmqpSessionApis for Fe2o3AmqpSession {
     }
 
     async fn end(&self) -> Result<()> {
-        todo!()
+        self.session
+            .get()
+            .unwrap()
```
heaths, line 105: Don't unwrap in production code. You're already returning a `Result` so just map the error as appropriate e.g.,

```rust
self.session
    .get()
    .map(|err| Error::new(ErrorKind::Io, err)?
    .lock()
    // ...
}

**Review by heaths — APPROVED** [review]
I see nothing blocking.

File: sdk/core/azure_core_amqp/src/fe2o3/session.rs (hunk 1 above)
Comment (LarryOsterman) [reply sdk/core/azure_core_amqp/src/fe2o3/session.rs:105]: oh, *THAT'S* how to do it (FWIW, it's `ok_or_else`, not `map`).

Comment (LarryOsterman) [reply sdk/core/azure_core_amqp/src/fe2o3/session.rs]: Updated almost all the .unwrap calls.

Any suggestions about how to deal with .unwrap in Into<> and From<>? 

Do I need to use TryFrom instead?

Comment (heaths) [reply sdk/core/azure_core_amqp/src/fe2o3/session.rs]: As discussed offline, yeah, use `TryFrom` and friends.

File: sdk/core/azure_core_amqp/src/fe2o3/cbs.rs [hunk 2]
```diff
@@ -69,12 +74,22 @@ impl AmqpClaimsBasedSecurityApis for Fe2o3ClaimsBasedSecurity {
                     .to_offset(time::UtcOffset::UTC)
                     .unix_timestamp()
                     .checked_mul(1_000)
-                    .unwrap(),
+                    .ok_or_else(|| {
+                        azure_core::Error::message(
+                            azure_core::error::ErrorKind::Other,
```
heaths, line 79: Not necessarily for this PR, but I think this'll happen enough where I'm wondering if we should add a new `ErrorKind` or - at least in this case - if `DataConversion` (for which you can still pass a custom message or `Error` is more appropriate.

sdk/core/azure_core_amqp/src/fe2o3/connection.rs
```diff
@@ -52,10 +52,24 @@ impl AmqpConnectionApis for Fe2o3AmqpConnection {
...
+                        "Connection options are not set.",
+                    )
+                })?;
+
+                // options.max_frame_size().inspect(move |s| {
+                //     builder.max_frame_size(*s);
+                // });
                 if options.max_frame_size.is_some() {
                     builder = builder.max_frame_size(options.max_frame_size.unwrap());
                 }
```
heaths, line 67: As discussed offline:
```suggestion
                if let Some(max_frame_size) = options.max_frame_size {
                    builder = builder.max_frame_size(max_frame_size);
                }
```

Use `as_ref()` for non-`Copy`ables.

heaths (approved):

sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
LarryOsterman: I'm currently rethinking my entire "how I handle errors" philosophy for both AMQP and Eventhubs based on comments you've made in the past.

I'd love to chat a bit about brainstorming how we reconcile service specific errors with azure_core::Error - which doesn't really have a mechanism to express specific errors.

This shows up with Eventhubs a bunch, where there are a number of state related errors that can occur (for instance, if you attempt to get properties on a `ProducerClient` before you call `open` on the client) but they end up getting glommed into `azure_core::error::ErrorKind::Other` because there isn't a better `ErrorKind` enumerator that represents service specific errors.

PR #1829, opened 2024-09-27
Improve support for typed HTTP headers

### LINKED ISSUE #1826: How should we provide "typed" access to response headers?
Cosmos DB provides a lot of information in HTTP headers. For example, when modifying items, the `x-ms-session-token` header returns a token that can be provided in future requests to maintain "session consistency" (so future requests "see" modifications made earlier in the session). We also return metadata like the number of RUs consumed by a request (the "request charge"), the etag for a newly-created document, the Activity ID for tracing, etc. _Most of our language SDKs provide access to these values using typed responses of some kind._

Other clients handle this using inheritance or embedding. For example, [the Go client returns an `azcosmos.Response` which embeds the Azure Core response and adds first-class properties for these](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos#Response). The [.NET client returns a response type that inherits from the Azure Core response](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.itemresponse-1?view=azure-dotnet).

What I'm seeking here is to figure out the ideal way to represent this in the Rust SDK. There are a few options:

## Option 1: Do nothing.

Users can retrieve these values from `Response<T>` using the header names. 

I don't really like this option because it's a significant loss of functionality compared to the other SDKs. However, it's certainly one of the simplest options.

For the cosmos example, if you wanted to access the session token or request charge while still deserializing the body, you'd write something like this:

```rust
let response = client.create_item(...).await.unwrap();
println!("Request Charge: {}", response.headers().get_as::<f32>(REQUEST_CHARGE).unwrap());
println!("Session Token: {}", response.headers().get_str(SESSION_TOKEN).unwrap());
println!("Created Item: {:?}", response.deserialize_body().await.unwrap());
```

## Option 2: Encode these values in the Model type

In this option, instead of returning `azure_core::Response<T>`, we'd return `azure_core::Response<ItemResponse<T>>`, where `ItemResponse` contains properties for things like the session token.

This is the most complicated option and requires some signficant changes. We'd need `azure_core::Model::from_response_body` to take the full `Response` instead, and we'd need to write our own custom logic to pull the headers and then deserialize the body. If a user changed the deserialize type, using `deserialize_body_into`, they'd lose this data.

**NOTE:** This is what I did for `azure_data_cosmos::clients::ContainerClient::query_items` but I don't think it works outside of paged responses. It worked there because we return `Pageable<QueryResults<T>>`, which doesn't provide access to the underlying responses (unless we return `Pageable<Response<T>>`, I suppose).

For the cosmos example, if you wanted to access the session token or request charge while still deserializing the body, you'd write something like this:

```rust
let response = client.create_item(...).await.unwrap();
let model = response.deserialize_body().await.unwrap();
println!("Request Charge: {}", model.request_charge);
println!("Session Token: {}", model.session_token);
println!("Created Item: {:?}", model.item);
```

## Option 3: Create custom `Response` types

Instead of `azure_core::Response`, methods that want to provide typed access to response headers would return their own struct, which wraps `azure_core::Response<T>` (for example, `azure_data_cosmos::ItemResponse<T>`). To be ergonomic, I think these structs would have to implement `Deref<Target = azure_core::Response>` so that you can call methods like `deserialize_body()`.

I don't mind this option. It's the most similar to how the Go SDK does it. It feels like inheritance, which is not very idiomatic in Rust, but it has the least impact on service client methods that _don't_ need to return custom data from headers.

For the cosmos example, if you wanted to access the session token or request charge while still deserializing the body, you'd write something like this:

```rust
let response = client.create_item(...).await.unwrap();
println!("Request Charge: {}", response.request_charge());
println!("Session Token: {}", response.session_token());

// If we provide a Deref impl OR reimplement some methods from `Response` manually.
println!("Created Item: {:?}", response.deserialize_body().await.unwrap());

// If we don't provide a Deref impl
println!("Created Item: {:?}", response.raw_response().deserialize_body().await.unwrap());
```

## Option 4: Add a "Detail" value to `Response<T>

This would add a new type parameter and field to `Response<T>`. The new type, `Response<T, D = ()>`, would be able to store some kind of service-specific details derived from the response headers. This detail would be provided by the service client, which would parse the headers. The pipeline would return `Response<T, ()>`, and we'd define a `Response<T, ()>::with_detail<D>(detail D) -> Response<T, D>` method that "sets" the details for a response.

`Response<T, ()>` would have a nonsense `detail()` method that returns `()`, which is harmless but could cause confusion in rust-analyzer completion prompts.

For the cosmos example, if you wanted to access the session token or request charge while still deserializing the body, you'd write something like this:

```rust
let response = client.create_item(...).await.unwrap();
println!("Request Charge: {}", response.detail().request_charge());
println!("Session Token: {}", response.detail().session_token());
println!("Created Item: {:?}", response.deserialize_body().await.unwrap());
```

---

Of these options, I'm leaning strongest towards **Option 3** at this point, though I've waffled back and forth between that and **Option 4** as I've thought through this and typed out this issue. If we do Option 3, implementing `Deref` or reimplementing some methods from `Response<T>` feels more ergonomic, but using Deref in this way is [often considered an **anti-pattern**](https://rust-unofficial.github.io/patterns/anti_patterns/deref.html).

analogrelay:
Closes #1826 

This has two changes to improve support for typed HTTP headers:

1. The `AsHeaders` trait, used to convert a type into a set of HTTP headers, is now fallible. Types which can be infallibly converted to headers should set their error type to `std::convert::Infallible` (which will eventually be an alias for the `never` type `!`).
2. Added a new `FromHeaders` trait, which allows for reading a type _out_ of a set of HTTP headers. It's fallible **and** allows for returning `None` if the headers are simply not present. There are now two more methods on `Headers`: `get` and `get_optional` which allow for reading a required or optional value out of the headers.

I've put a few notes inline that I wanted to specifically call attention to. I also updated `Range` (the only "built-in" type that implements `AsHeaders`) and switched Cosmos' `PartitionKey` to implement `AsHeaders` now that there's a way to return an error.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/partition_key.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/partition_key.rs
@@ -1,6 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+use azure_core::headers::{AsHeaders, HeaderName, HeaderValue};
+
+use crate::constants;
+
 /// Describes the partition strategy that will be used when querying.
 ///
 /// Currently, the only supported strategy is [`QueryPartitionStrategy::SinglePartition`], which executes the query against a single partition, specified by the [`PartitionKey`] provided.
@@ -90,8 +94,11 @@ impl<T: Into<PartitionKey>> From<T> for QueryPartitionStrategy {
 #[derive(Debug, Clone)]
 pub struct PartitionKey(Vec<PartitionKeyValue>);
 
-impl PartitionKey {
-    pub(crate) fn into_header_value(self) -> azure_core::Result<String> {
+impl AsHeaders for PartitionKey {
+    type Error = azure_core::Error;
+    type Iter = std::iter::Once<(HeaderName, HeaderValue)>;
+
+    fn as_headers(&self) -> Result<std::iter::Once<(HeaderName, HeaderValue)>, Self::Error> {
         // We have to do some manual JSON serialization here.
         // The partition key is sent in an HTTP header, when used to set the partition key for a query.
         // It's not safe to use non-ASCII characters in HTTP headers, and serde_json will not escape non-ASCII characters if they are otherwise valid as UTF-8.
@@ -100,10 +107,10 @@ impl PartitionKey {
         let mut json = String::new();
         let mut utf_buf = [0; 2]; // A buffer for encoding UTF-16 characters.
         json.push('[');
-        for key in self.0 {
+        for key in &self.0 {
             match key.0 {
                 InnerPartitionKeyValue::Null => json.push_str("null"),
-                InnerPartitionKeyValue::String(string_key) => {
+                InnerPartitionKeyValue::String(ref string_key) => {
                     json.push('"');
                     for char in string_key.chars() {
                         match char {
@@ -125,8 +132,10 @@ impl PartitionKey {
                     }
                     json.push('"');
                 }
-                InnerPartitionKeyValue::Number(num) => {
-                    json.push_str(serde_json::to_string(&serde_json::Value::Number(num))?.as_str());
+                InnerPartitionKeyValue::Number(ref num) => {
+                    json.push_str(
+                        serde_json::to_string(&serde_json::Value::Number(num.clone()))?.as_str(),
+                    );
                 }
             }
 
@@ -137,7 +146,10 @@ impl PartitionKey {
         json.pop();
         json.push(']');
 
-        Ok(json)
+        Ok(std::iter::once((
+            constants::PARTITION_KEY,
+            HeaderValue::from_cow(json),
+        )))
     }
 }
 
--- a/sdk/cosmos/azure_data_cosmos/src/partition_key.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/partition_key.rs
@@ -98,7 +98,7 @@ impl AsHeaders for PartitionKey {
     type Error = azure_core::Error;
     type Iter = std::iter::Once<(HeaderName, HeaderValue)>;
 
-    fn as_headers(&self) -> Result<std::iter::Once<(HeaderName, HeaderValue)>, Self::Error> {
+    fn as_headers(&self) -> Result<Self::Iter, Self::Error> {
         // We have to do some manual JSON serialization here.
         // The partition key is sent in an HTTP header, when used to set the partition key for a query.
         // It's not safe to use non-ASCII characters in HTTP headers, and serde_json will not escape non-ASCII characters if they are otherwise valid as UTF-8.
--- a/sdk/cosmos/azure_data_cosmos/src/partition_key.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/partition_key.rs
@@ -277,19 +277,22 @@ impl_from_tuple!(0 A 1 B 2 C);
 
 #[cfg(test)]
 mod tests {
-    use crate::PartitionKey;
-
-    use super::QueryPartitionStrategy;
+    use crate::{constants, PartitionKey, QueryPartitionStrategy};
+    use typespec_client_core::http::headers::AsHeaders;
 
     fn key_to_string(v: impl Into<PartitionKey>) -> String {
-        v.into().into_header_value().unwrap()
+        let key = v.into();
+        let mut headers_iter = key.as_headers().unwrap();
+        let (name, value) = headers_iter.next().unwrap();
+        assert_eq!(constants::PARTITION_KEY, name);
+        value.as_str().into()
     }
 
     /// Validates that a given value is `impl Into<QueryPartitionStrategy>` and works as-expected.
     fn key_to_single_partition_strategy_string(v: impl Into<QueryPartitionStrategy>) -> String {
         let strategy = v.into();
         let QueryPartitionStrategy::SinglePartition(key) = strategy;
-        key.into_header_value().unwrap()
+        key_to_string(key)
     }
 
     #[test]
@@ -339,10 +342,7 @@ mod tests {
     #[test]
     pub fn non_ascii_string() {
         let key = PartitionKey::from("smile 😀");
-        assert_eq!(
-            key.into_header_value().unwrap().as_str(),
-            r#"["smile \ud83d\ude00"]"#
-        );
+        assert_eq!(key_to_string(key), r#"["smile \ud83d\ude00"]"#);
     }
 
     #[test]
--- a/sdk/typespec/typespec_client_core/src/http/headers/mod.rs
+++ b/sdk/typespec/typespec_client_core/src/http/headers/mod.rs
@@ -14,7 +14,7 @@ use typespec::error::{Error, ErrorKind, ResultExt};
 
 /// A trait for converting a type into request headers.
 pub trait AsHeaders {
-    type Error;
+    type Error: std::error::Error + Send + Sync + 'static;
     type Iter: Iterator<Item = (HeaderName, HeaderValue)>;
 
     fn as_headers(&self) -> Result<Self::Iter, Self::Error>;
@@ -55,10 +55,10 @@ where
 pub trait FromHeaders: Sized {
     type Error: std::error::Error + Send + Sync + 'static;
 
-    /// Holds a list of the header names that [`FromHeaders::from_headers`] expects.
+    /// Gets a list of the header names that [`FromHeaders::from_headers`] expects.
     ///
     /// Used by [`Headers::get()`] to generate an error if the headers are not present.
-    const HEADER_NAMES: &'static [&'static str] = &[];
+    fn header_names() -> &'static [&'static str];
 
     /// Extracts the value from the provided [`Headers`] collection.
     ///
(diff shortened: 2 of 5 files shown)


sdk/typespec/typespec_client_core/src/http/headers/mod.rs
```diff
@@ -9,42 +9,66 @@ mod microsoft;
...
+    /// Used by [`Headers::get()`] to generate an error if the headers are not present.
+    const HEADER_NAMES: &'static [&'static str] = &[];
+
+    /// Extracts the value from the provided [`Headers`] collection.
+    ///
+    /// This method returns one of the following three values:
+    /// * `Ok(Some(...))` if the relevant headers are present and could be parsed into the value.
+    /// * `Ok(None)` if the relevant headers are not present, so no attempt to parse them can be made.
+    /// * `Err(...)` if an error occurred when trying to parse the headers. This likely indicates that the headers are present but invalid.
+    fn from_headers(headers: &Headers) -> Result<Option<Self>, Self::Error>;
```
analogrelay, line 69: The age-old question of how to represent something that is optional **and** fallible. I'm open to other Options that get us to the same Result (puns absolutely intended).

If we're going to have nested Result/Option, I think this is the best ordering, because it allows you to write `if let Some(h) = from_headers(...)?`. I think this is the most ergonomic option.

We do have `Headers::get()` which coerces errors in to `azure_core::Error` and produces an `ErrorKind::DataConversion` error if the required headers are missing, which can be used for reading mandatory header.

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
```diff
@@ -90,8 +94,11 @@ impl<T: Into<PartitionKey>> From<T> for QueryPartitionStrategy {
 #[derive(Debug, Clone)]
 pub struct PartitionKey(Vec<PartitionKeyValue>);
 
-impl PartitionKey {
-    pub(crate) fn into_header_value(self) -> azure_core::Result<String> {
+impl AsHeaders for PartitionKey {
+    type Error = azure_core::Error;
+    type Iter = std::iter::Once<(HeaderName, HeaderValue)>;
+
+    fn as_headers(&self) -> Result<std::iter::Once<(HeaderName, HeaderValue)>, Self::Error> {
```
Pilchie, line 101: Curious, could/should this be:

```suggestion
    fn as_headers(&self) -> Result<Self::Iter, Self::Error> {
```

Pilchie (approved):

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
analogrelay: Yep, it should be. The two are identical, but it's generally preferred to keep the actual type in one place and just reference it (because duplication yields inconsistency ;)).

I _wanted_ to use an `impl Trait` type to avoid having to name the actual type, but support for `impl Trait` in associated types is unstable 😞

sdk/typespec/typespec_client_core/src/http/headers/mod.rs
```diff
@@ -69,6 +93,28 @@ impl Headers {
         Self::default()
     }
 
+    /// Gets the headers represented by `H`, or return an error if the header is not found.
+    pub fn get<H: FromHeaders>(&self) -> crate::Result<H> {
+        match H::from_headers(self) {
+            Ok(Some(x)) => Ok(x),
+            Ok(None) => Err(crate::Error::with_message(
+                ErrorKind::DataConversion,
+                || format!("required headers not found: {}", H::HEADER_NAMES.join(", ")),
```
heaths, line 102: Nit: most impls will likely just have one header, so maybe:
```suggestion
                || format!("required header(s) not found: {}", H::HEADER_NAMES.join(", ")),
```

sdk/typespec/typespec_client_core/src/http/headers/mod.rs
```diff
@@ -146,13 +192,19 @@ impl Headers {
     }
 
     /// Add headers to the headers collection.
-    pub fn add<H>(&mut self, header: H)
+    ///
+    /// ## Errors
+    ///
+    /// The error this returns depends on the type `H`.
+    /// Many header types are infallible, and do not return
```
heaths, line 199: Nit: for public docs, we should strive for good grammar.
```suggestion
    /// Many header types are infallible, and do not return.
```

sdk/typespec/typespec_client_core/src/http/headers/mod.rs
```diff
@@ -273,3 +325,85 @@ impl From<&String> for HeaderValue {
         s.clone().into()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use typespec::error::ErrorKind;
+    use url::Url;
```
heaths, line 332: Nit: I recommend using our re-export in case we end up defining this ourselves, which I'm leaning toward to reduce dependencies. Same for `http_types` and some others as well.

sdk/typespec/typespec_client_core/src/http/headers/mod.rs
```diff
@@ -273,3 +325,85 @@ impl From<&String> for HeaderValue {
...
+    use super::{FromHeaders, HeaderName, Headers};
+
+    // Just in case we add a ContentLocation struct later, this one is named "ForTest" to indicate it's just here for this test.
+    #[derive(Debug)]
+    struct ContentLocationForTest(Url);
+
+    impl FromHeaders for ContentLocationForTest {
+        type Error = url::ParseError;
+
+        const HEADER_NAMES: &'static [&'static str] = &["content-location"];
```
heaths, line 343: I think this'll lint: consts don't need to declare `'static` as its implied.

heaths (approved):
I love this! I see nothing blocking, but had a few suggestions.

sdk/typespec/typespec_client_core/src/http/headers/mod.rs
```diff
@@ -9,42 +9,66 @@ mod microsoft;
...
+/// Extract a value from the [`Headers`] collection.
+///
+/// The [`FromHeaders::from_headers()`] method is usually used implicitly, through [`Headers::get()`] or [`Headers::get_optional()`].
+pub trait FromHeaders: Sized {
+    type Error: std::error::Error + Send + Sync + 'static;
+
+    /// Holds a list of the header names that [`FromHeaders::from_headers`] expects.
+    ///
+    /// Used by [`Headers::get()`] to generate an error if the headers are not present.
+    const HEADER_NAMES: &'static [&'static str] = &[];
```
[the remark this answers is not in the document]
analogrelay: That's actually what I started with but then I was thinking that it would usually be a `const` so I made it a `const`. I hadn't considered object-safety though, so I appreciate that check! I'll go back to using a method.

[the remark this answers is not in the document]
analogrelay: Actually, I don't think being object-safe is useful here. There are no methods on `FromHeaders` that receive `self` as a parameter, since it's entirely for constructing objects. If you had a `&dyn FromHeaders`, there's nothing you can call on it. In order to _make_ this object-safe, I'd have to apply `where Self: Sized` constraints to all the methods (and remove the `: Sized` constraint from the trait itself). Doing that excludes those methods from being called on trait objects and there's nothing left to call.

Having said that, I _do_ think a function is a more flexible option here.

sdk/typespec/typespec_client_core/src/http/headers/mod.rs
analogrelay: I'll just do the `if` to do plural right. I'm not a fan of using `(s)` for this kind of thing.

sdk/typespec/typespec_client_core/src/http/headers/mod.rs
analogrelay: I actually just forgot to write the rest of that section 😅

sdk/typespec/typespec_client_core/src/http/headers/mod.rs
analogrelay: No longer relevant with the refactor to a function.

sdk/typespec/typespec_client_core/src/http/headers/mod.rs
```diff
@@ -9,42 +9,66 @@ mod microsoft;
...
+            None => Ok(vec![].into_iter()),
         }
     }
 }
 
+/// Extract a value from the [`Headers`] collection.
+///
+/// The [`FromHeaders::from_headers()`] method is usually used implicitly, through [`Headers::get()`] or [`Headers::get_optional()`].
+pub trait FromHeaders: Sized {
+    type Error: std::error::Error + Send + Sync + 'static;
```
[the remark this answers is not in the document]
analogrelay: Yeah, I'll do that. It was necessary for `FromHeaders` because `Headers::get()` wraps the arbitrary error type up in `azure_core::Error` and that requires these constraints. We don't do that with `AsHeaders` anywhere, but adding those constraints later would be a breaking change, so I'd rather do it now.

sdk/typespec/typespec_client_core/src/http/headers/mod.rs
heaths: But, on the other hand, it makes a consistent message to search for in logs, traces, stack traces, etc. In normal UI messages I don't like cop-outs like that either, but this is basically a debug/dev error. Consistency is king.

E.g., some parts of Windows Installer error messages are localized while others aren't. Those in the know search for "value 3" because it's the only part of a typical error message not localized when looking for fatal errors. You don't want that for our customers, do you?

analogrelay: Good point, I'll revert to your suggestion.

heaths (approved):

sdk/typespec/typespec_client_core/src/http/headers/mod.rs
```diff
@@ -9,42 +9,66 @@ mod microsoft;
...
+
+    /// Gets a list of the header names that [`FromHeaders::from_headers`] expects.
+    ///
+    /// Used by [`Headers::get()`] to generate an error if the headers are not present.
+    fn header_names() -> &'static [&'static str];
```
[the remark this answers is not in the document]
analogrelay: A lifetime is needed because there are no input lifetimes to use to infer the lifetime of the output.

heaths (approved):

Issue #1803, opened 2024-09-16
azure_core::Model derive macro doesn't infer `Deserialize` bounds in `impl`


analogrelay:
The derive macro for `Model` doesn't work for generic arguments. For example, this doesn't work:

```rust
#[derive(Model, Deserialize)]
pub struct Wrapper<T> {
    items: Vec<T>,
}
```

This is because our logic generates an `impl` like this:

```
impl<T> Model for Wrapper<T> {
}
```

But, serde's `Deserialize` derive macro generates this (effectively):

```
impl<T: Deserialize> Deserialize for Wrapper<T> {
}
```

Since our `Model` impl depends on the impl of `Deserialize`, we end up with an error:

```
the trait bound `T: Deserialize<'_>` is not satisfied
required for `container_client::Wrapper<T>` to implement `DeserializeOwned`
```

What we'd need to do is add logic to our derive macro that matches [serde's logic to guess appropriate type parameter bounds for the impl](https://serde.rs/attr-bound.html). This isn't really blocking, since one can manually implement `Model` without too much trouble in this case. For example:

```rust
impl<M: DeserializeOwned> azure_core::Model for QueryResponseModel<M> {
    async fn from_response_body(
        body: azure_core::ResponseBody,
    ) -> typespec_client_core::Result<Self> {
        body.json().await
    }
}
```

analogrelay:
Logging this because I hit it in implementing a query API for Cosmos. I'm working around it with the work-around posted above, but I wanted to track this issue for future reference (and to hopefully fix it later)

PR #1830, opened 2024-09-30
[Cosmos] Item Create/Read/Replace/Upsert/Delete

analogrelay:
Adds support for the standard item "CRUD" operations, which in our case is:

* Create - Creates a new item, failing with 409 Conflict if an item with the same id already exists
* Read - Gets an item given it's partition key and item id
* Delete - Deletes an item given it's partition key and item id
* Replace - Replaces an item given it's partition key and item id, failing with 404 if the item does not exist. 
* Upsert - Creates a new item or replaces an existing item (depending on if the item already exists)

For our other clients, these are all unique methods and I think it makes sense to do that here as well, even though Upsert/Create is the same API but varies by the header provided. I've also held off on internal refactoring for now, though there is definitely a fair bit of repetition here.

A couple additional pieces:

* I refactored our existing two examples (`cosmos_metadata`, which retrieves metadata about a DB or Container, and `cosmos_query` which executes queries) with a single example that provides subcommands for lots of different scenarios. This pattern does make the example more complicated but it allows for much better sharing of the client-creation logic.
* I'm rebasing this on top of #1829 when it lands, which will allow us to use a more ergonomic syntax to insert the Partition Key-related headers.
* I realized we could eliminate the `CosmosTimestamp` type by providing serde with a function to parse Cosmos's timestamp format and return an `OffsetDateTime` (which is the type we've standardized on for a timestamp like this).

Finally, there's a bit of an annoying quirk that I'm doing some thinking on. All our APIs that return a user's own type (like `read_item`) currently return some kind of wrapper around that. For queries, we return `QueryResults<T>` representing a page of results, and for CRUD operations we return `Item<T>` which is a very basic "newtype"-style wrapper around the item. We do this because `azure_core::Response<T>` requires that `T` implement `azure_core::Model`, which handles picking the right serialization format to deserialize the object (JSON, XML, etc.). I didn't want user-provided types to have to implement `azure_core::Model`, so instead of returning `azure_core::Response<T>`, we return `azure_core::Response<Item<T>>`, where our `Item<T>` implements `azure_core::Model` for any given `T: Deserialize`.

I don't really like that. The `Item<T>` wrapper is a mostly-useless newtype. I could implement `Deref<Target = T>` on `Item<T>` to make it **appear** as a `T`, but it's not _really_ a smart-pointer. The other thing I'm considering is a refactor to remove `azure_core::Model` and replace it with some kind of field on `Response<T>` that tells the Core SDK how to deserialize the value **without** having to associate that with the _type_ itself. I'll open a separate issue to discuss this problem in detail.


sdk/cosmos/azure_data_cosmos/src/models/mod.rs
```diff
@@ -3,37 +3,74 @@
...
+        body.json()
+    }
+
+    #[cfg(target_arch = "wasm32")]
+    fn from_response_body(
+        body: azure_core::ResponseBody,
+    ) -> impl std::future::Future<Output = typespec_client_core::Result<Self>> {
+        body.json()
+    }
+}
```
analogrelay, line 50: We can actually use the derive macro, I'll remove this and use that instead. I had incorrectly assumed that the derive macro would require that `T` implement `Model` too, but unlike `Deserialize`, we don't need to be recursive in handling `Model`.

analogrelay: Ah never mind, I forgot that this is necessary because of #1803

sdk/typespec/typespec_client_core/src/http/response.rs
```diff
@@ -18,6 +18,10 @@ pub type PinnedStream = Pin<Box<dyn Stream<Item = crate::Result<Bytes>> + Send +
 pub type PinnedStream = Pin<Box<dyn Stream<Item = crate::Result<Bytes>>>>;
 
 /// Trait that represents types that can be deserialized from an HTTP response body.
+///
+/// [`Response<T>`] is designed to work with values that may be serialized in various formats (JSON, XML, etc.).
+/// In order to support that, the `T` provided must implement [`Model`], which provides the [`Model::from_response_body`]
+/// method to deserialize the type from an arbitrary
```
Pilchie, line 24: From an arbitrary what?

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -142,13 +340,125 @@ impl ContainerClientMethods for ContainerClient {
...
+        req.insert_header(
+            constants::PARTITION_KEY,
+            HeaderValue::from_cow(partition_key.into().into_header_value()?),
+        );
+        self.pipeline
+            .send(Context::new(), &mut req, ResourceType::Items)
+            .await
+    }
+
+    #[allow(async_fn_in_trait)] // REASON: See https://github.com/Azure/azure-sdk-for-rust/issues/1796 for detailed justification
```
Pilchie, line 432: This was copied from the trait and can be deleted, right?

Pilchie (approved):

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
analogrelay: Correct!

sdk/typespec/typespec_client_core/src/http/response.rs
analogrelay: Fixed this!

Pilchie (approved):

analogrelay:
Aside from addressing feedback, I made one additional change that warrants reviewing: The `Item<T>` wrapper type we use for returning items is now an Option-like enum [^1]. The other SDKs support setting a header when writing items to control if the server will return the content of an item when you perform a write. The default for this is `true`, so right now the Rust SDK will always return an item. However, we'll be adding logic soon to handle these options. Rust has no concept of null or nil, so we **must** account for the potential for a missing item in the API surface.

[^1]: Why introduce a new Option-like? There are some comments inline, but to repeat the general idea: We need to implement `azure_core::Model` and it's not reasonable to do so on `Option<T>`

Pilchie (approved):

sdk/typespec/src/error/mod.rs
```diff
@@ -169,6 +169,14 @@ impl Error {
         }
     }
 
+    #[cfg(feature = "http")]
+    pub fn http_status(&self) -> Option<StatusCode> {
```
heaths, line 173: I think for now this is fine, but for posterity I question whether this is too specific at least for HTTP status code. I could imagine returning more since we're already declaring it - e.g., a tuple of the status code and status message - but have no basis to suggest any changes now. We'll watch feedback.

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
```diff
@@ -74,7 +73,7 @@ impl DatabaseClientMethods for DatabaseClient {
         &self,
 
         #[allow(unused_variables)]
-        // This is a documented public API so prefixing with '_' is undesirable.
+        // REASON: This is a documented public API so prefixing with '_' is undesirable.
```
heaths, line 76: Perhaps you're only seeing this lint because there's nothing public that actually uses it though, like an example. Should it be demonstrated instead? I'm not opposed to adding this, just wanted to point out that nothing public is touching it if you're seeing this.

Actually, that said, I thought as long as the trait is public, this shouldn't be raised. *So many* public APIs would trigger this lint otherwise. Maybe double check. Again, not opposed to the `#[allow]` attribute here.

sdk/cosmos/azure_data_cosmos/src/models/item.rs
```diff
@@ -0,0 +1,125 @@
...
+
+use azure_core::Model;
+use serde::de::DeserializeOwned;
+
+/// Returned by Cosmos DB APIs that return a single item.
+///
+/// In some circumstances, an API that _can_ return an item will **not** return an item.
+/// For example, you can use [`ItemOptions`](crate::options::ItemOptions) to configure APIs
+/// that write new or updated items to avoid returning the updated item.
+// Marking this non-exhaustive would make things very difficult for users.
```
heaths, line 12: Nit: will the compiler complain if you put a new line here just to visually separate the public comment from internal comment? I did a double take initially. If not, no biggie. This is semantically correct.

sdk/cosmos/azure_data_cosmos/src/models/item.rs
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+
+    /// Returns a boolean indicating if the [`Item<T>`] is present.
+    pub fn is_present(&self) -> bool {
+        match self {
+            Item::Present(_) => true,
+            Item::Omitted => false,
+        }
+    }
```
heaths, line 45: Are these more useful in some scenarios that just a match as with `Option<T>` e.g.,

```rust
let Present(thing) = item else {
    // return or whatever
}
```

If you think these are generally useful without creating an anti-pattern as I've been seeing with newer rust devs in related projects (checking `Option<T>::is_some()`, for example, keep them. We just generally don't want to add public APIs for "might be useful" scenarios vs. "most likely or definitely useful" so we don't end up owning them forever. Too many APIs can also lead to confusion.

sdk/cosmos/azure_data_cosmos/src/models/item.rs
```diff
@@ -0,0 +1,125 @@
...
+    async fn from_response_body(
+        body: azure_core::ResponseBody,
+    ) -> typespec_client_core::Result<Self> {
+        let bytes = body.collect().await?;
+        if bytes.is_empty() {
+            Ok(Item::Omitted)
+        } else {
+            let item = azure_core::json::from_json(bytes)?;
+            Ok(Item::Present(item))
+        }
```
heaths, line 81: To my comment above, this could just as easily be `map_or_else`.

heaths (approved):
I have a few thoughts, but nothing blocking.

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
analogrelay: Y'know, it's probably that I put this here before adding the trait and then it became copy-pasta. I'll try removing it.

sdk/cosmos/azure_data_cosmos/src/models/item.rs
analogrelay: I typed it out with a newline, and `cargo fmt` removed it 😭 . Let me see if I can make it work.

sdk/typespec/src/error/mod.rs
analogrelay: The main reason I added this was to be able to `match` against an HTTP error here:

https://github.com/Azure/azure-sdk-for-rust/blob/1f2cef7c74cc0c8fb423986544cfad472cd517da/sdk/cosmos/azure_data_cosmos/examples/cosmos/read.rs#L33-L44

Being able to create concise match arms against HTTP status codes is pretty useful as a consumer of our APIs. I'm not 100% happy with this yet, but it's a start

sdk/cosmos/azure_data_cosmos/src/models/item.rs
heaths: Naw, don't worry about it. Not a big deal.

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
analogrelay: I was able to get rid of all the others, but not this one... Honestly, I don't understand why this one needs it and the others don't. This method is used in examples, doctests, etc. just as much as the ones in `container_client`. I was able to actually remove those though. Only this one continues to report a warning.

sdk/cosmos/azure_data_cosmos/src/models/item.rs
analogrelay: I didn't see a straightforward way to have rustfmt leave these alone, so I'll just leave it.

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
Pilchie: I didn't think it was use of the method, I thought it was use of the parameter within the method. In this case you never touch the options parameter, hence the lint warning.

analogrelay: Correct. But that's true of the `options` parameter in the methods on `ContainerClient`. The `options` parameters aren't currently used at all, they're there just so we can add fields to them later.

PR #1834, opened 2024-10-03
Add custom future type, ResponseFuture, for HTTP responses

### LINKED ISSUE #1831: Rethinking the `Model` trait
In working on the Cosmos DB APIs, I've identified something annoying about the currently approach we're taking to deserialization. Right now, to make `Response<T>` deserializable in either JSON or XML depending on what the _service-client_ wants, we use a trait: `azure_core::Model` which looks like this:

```rust
pub trait Model: Sized {
    fn from_response_body(
        body: ResponseBody,
    ) -> impl Future<Output = crate::Result<Self>> + Send + Sync;
}
```

We couple this with a derive macro that allows you to set the "format" to use:

```rust
#[derive(Model)]
pub struct Secret {
    name: String,
    value: String,
}
```

Or, to use XML:

```rust
#[derive(Model)]
#[typespec(format = "xml")]
pub struct Secret {
    name: String,
    value: String,
}
```

In effect, we tie the serialization format to the _type_ being deserialized. This is fine when we control all the types being deserialized (for example, when deserializing service API responses). We can derive this trait and users don't need to see anything related to the serialization format.

The problem arises when we need to support deserializing entirely user-authored types. We support this in two major scenarios:

1. Using a `Response::deserialize_body_into()` helper, we allow a user to deserialize _any_ service response into their own type. To do this, a user must derive `azure_core::Model` themselves. But this isn't too onerous a requirement if the user is providing a custom model to deserialize Azure responses.

2. Some APIs, like Cosmos, provide REST APIs that return data structures that are entirely user-controlled. For example, reading an item from a Cosmos container, the API returns the user's JSON document as-is. We don't really want users to have to implement `azure_core::Model` in this case. They'd have to add a direct dependency on `typespec_client_core` to do that, and it generally just seems undesirable to force the user to implement our own trait when they already implement `serde::Deserialize`. Today, in #1830 , we work around this by returning `azure_data_cosmos::models::Item<T>`, where `Item` is a simple newtype that handles implementing `azure_core::Model` for the user:

```rust
/// Wrapper type used to hold items stored in a Cosmos DB Container.
#[derive(Model, Deserialize, Debug, Clone, Default)]
#[serde(transparent)]
pub struct Item<T>(
    // This wrapper type is necessary because Response<T> needs T to implement azure_core::Model.
    // However, we don't want to require user types to have to implement azure_core::Model.
    // So this type does that, and handles deserializing the payload from JSON.
    T,
);
```

This type is largely useless to the user. It provides no extra data and serves only to "tag" the type with JSON deserialization logic. Meanwhile, the user has to unwrap this type to get at their user-specific type, making an already-long chain of calls and `.await`s even longer.

I've got a couple thoughts on what we could do here:

1. Nothing. We accept that we need to use "newtypes" for user-defined types, or users have to derive `azure_core::Model`

2. We use the `Content-Type` response header to drive deserialization, perhaps with some kind of override from the service client. This isn't my preferred approach, as it depends on services always returning known content types (not guaranteed) and matching against a string to parse (not ideal)

3. We use a fixed `SerializationFormat` enum to specify the format for a `Response<T>`. The service client _sets_ this value somehow (it could be a parameter to `typespec_client_core::http::Pipeline::send`, or something they add to the response after sending. The deserialize logic in `Response<T>` would look at this value. If we want to add new formats, we'd have to add to this enum, but it seems like the list of supported formats is very stable.

4. We add a new type parameter to `Response`: `Response<T, F>`, where `T: Deserialize` and `F: SerializationFormat`. `SerializationFormat` would be a trait we create to define a serialization format. **This** type controls deserialization, but the "body" type `T` is now free to just be a standard `Deserialize`. This is a fairly tidy approach except that it can pollute the type names quite a bit. We can use defaults to make that less impactful, but it still means the serialization "format" would appear in the type name for some APIs (technically it's always there just not named):

```rust
pub struct Response<T = (), F = JsonFormat>
{
}

fn call_some_json_api() -> Response<Secret> {
    todo!();
}

fn call_some_xml_api() -> Response<Secret, XmlFormat> {
    todo!();
}
```

5. The same as 4 but we use dynamic dispatch and store an `Arc<dyn SerializationFormat>` (or even `&'static dyn SerializationFormat`, since each format would only need one instance) on the `Response<T>`. This is basically the same as option 3, just using a trait object instead of a fixed enum, which could allow for new formats to be added more easily.

Of these options, honestly, I think 3 is a pretty reasonable choice. If we make the enum `#[non_exhaustive]` then we can safely add new ones later. The only place that should be `match`ing against it is our `Response<T>::deserialize_body` code anyway. Adding an extra `match` to `deserialize_body` doesn't seem like a performance risk worth digging in to since we're dealing with network I/O here anyway. I'd propose **either** having the `typespec_client_core::http::Pipeline::send` method accept it as a parameter, **or** have the pipeline accept it as a constructor argument. Or perhaps a hybrid of that where the pipeline has it as a constructor argument to set a default and there's a `send_with_format` "override" that allows you to use a different format than the default (which would be rare).

analogrelay:
Closes #1831 

*This is a long one, but I really wanted to try and lay out all the details of the change*

This PR introduces a new type, `ResponseFuture`, which is intended to be returned by service client APIs. For example:

```rust
impl ContainerClient {
    fn read(
        &self,
        options: Option<ReadContainerOptions>,
    ) -> azure_core::ResponseFuture<ContainerProperties>;
}
```

The `Response<T>` type has been simplified dramatically. It is now a plain-ol data object containing the status code, headers, and "body" (which is of type `T`). The default value of `T` is `ResponseBody`, which is our existing type that wraps a "Pinned Stream" representing an asynchronous stream of bytes coming from the transport.

## How is `ResponseFuture` different to the user?

Our current service client methods return `impl Future<Output = Result<Response<T>, Error>>`. However, in the current code, `Response<T>` is _itself_ a promise as well. It does not contain a fully downloaded and deserialized HTTP response body, only the promise of one, when `deserialize_body` is called. By default, a `ResponseFuture<T>`, when directly `await`ed, will execute the request, receive the status code and headers, **download the body, and deserialize it into a `T`**. I believe this represents the "majority" of use cases, which means we can turn our current long chain of method calls:

```rust
let response = secret_client.get_secret("secret_password")
    .await?.
    .deserialize_body()
    .await?;
println!("Secret Value", response.into_body().value);
```

Into just one await:

```rust
let response = secret_client.get_secret("secret_password").await?;
println!("Secret Value", response.into_body().value);
```

In a situation where most users _want_ to read the body (if there is one) into the default model provided by the service client, this is much more ergonomic. It also removes the need for the `Model` trait, since the service client gave the `ResponseFuture` a closure it could use to perform the deserialization.

## What about more advanced scenarios?

While I believe eagerly reading and deserializing the body is the majority case, it is not a good approach in all cases. Consider, for example, that Cosmos DB often returns the full JSON of a newly-created item. Often, the user doesn't need this since they just created the item (in fact, we even provide a request header to stop the server from sending the new item in the response). Or consider an API that returns a large payload, but also includes some response headers that the user may want to use to decide if it even _wants_ to download that whole payload. Or any other case where the user wants to download the response headers but defer downloading the body until later. Until the `ResponseFuture` is  `await`ed, it only represents the promise of a future response, and the user has the ability to modify it's behavior using methods on `ResponseFuture`.

Calling `.lazy()` will change the `ResponseFuture` into an `impl Future<Output = LazyResponse<T>>` which executes the requests and processes the response headers, but does not process the body until `LazyResponse::into_body` is called.

Calling `.raw()` will change the `ResponseFuture` into an `impl Future<Output = Response<ResponseBody>>`, in effect *removing the deserialization logic* so that the returned value will be a `Response<ResponseBody>`. That gives the user the ability to download the body and deserialize however they want[^1], including deserializing it to a custom type:

```rust
#[derive(Debug, Deserialize)]
struct MyCustomSecret {
    #[serde(rename = "name")]
    pub my_custom_name: String,
    #[serde(rename = "value")]
    pub my_custom_value: String,
}
let response = secret_client.get_secret("secret_password").raw().await?;
let raw_body = response.into_body();
let secret: MyCustomSecret = raw_body.json().await?;
```

## How does a service client create a `ResponseFuture`?

Rather than being `async fn`s that return `Result<Response<T>>`, service client methods should be **non-async** `fn`s that return `ResponseFuture<T>`. The main pipeline's `send` method returns a `ResponseFuture<T>`, so for most client methods that are assembling a request, sending it, and returning the resulting `ResponseFuture<T>`, there's no need for the client method itself to be `async`. In the less common event that a client needs to perform follow-up work after a response is received that **cannot be achieved using a custom `Policy`**, we have `ResponseFuture::new` which allows a caller to synthesize a `ResponseFuture` out of any other future that returns a `Result<Response>`.

`ResponseFuture<T>` represents the _promise_ of a `Response<ResponseBody>` (a raw HTTP response) to come at a future time **and** the ability to deserialize that raw body to a `T` when it does arrive. The `send` method on the pipeline returns `ResponseFuture<ResponseBody>` (i.e. a promise for a future `Response<ResponseBody>` representing a raw, unparsed, HTTP body, with no deserialization logic). From there, service client methods can call `ResponseFuture::json` and `ResponseFuture::xml` to convert that "raw" `ResponseFuture` into one that deserializes the body into their target type. For example:

```rust
impl ContainerClientMethods for ContainerClient {
    fn read(
        &self,
        options: Option<ReadContainerOptions>,
    ) -> azure_core::ResponseFuture<ContainerProperties> {
        let req = Request::new(self.container_url.clone(), azure_core::Method::Get);
        self.pipeline
            .send(Context::new(), req, ResourceType::Containers)
            .json() // Convert `ResponseFuture<ResponseBody>` to `ResponseFuture<ContainerProperties>` using JSON
    }
}
```

## Boxes. Boxes Everywhere.

The main downside is the liberal use of boxing and dynamic dispatch. We `Box` the future that performs the request, we `Box` a closure to deserialize the body, which itself returns another `Box`ed future representing the downloading and deserialization of the body. I did this because I wanted to avoid having to have per-operation types. If we want to optimize out the boxing, we have another option.

We would turn `ResponseFuture<T>` into a trait, intended to be used via `impl Trait` syntax, and use macros to have each service client return _its own custom struct_ that implements `ResponseFuture<T>`. Those client methods would look something like this:

```rust
impl ContainerClientMethods for ContainerClient {
    fn read(
        &self,
        options: Option<ReadContainerOptions>,
    ) -> impl azure_core::ResponseFuture<ContainerProperties> {
        // This macro expects async code that returns a raw response
        // The "actual" deserialization to JSON can't be done inside the code block
        // because we need to be able to defer it when "raw" is called
        response_future!(Json, {
            let req = Request::new(self.container_url.clone(), azure_core::Method::Get);
            self.pipeline
                .send(Context::new(), req, ResourceType::Containers)
                .await
        });
    }
}
```

In that approach, the type the macro generates would able to be fully static and avoid boxing altogether. I'm open to that idea, and returning custom futures **is** what the unofficial client does so there's some prior art there. It's just a *lot* of custom types. Having said that, **these types wouldn't have to be public**, they could be defined **in the method itself** and returned as `impl Trait` types so that the user doesn't actually see them.

## Ownership of `Context` and `Request`

Another call I made here was to change the `Pipeline::send` methods to take the context and request **by-value** instead of taking references. That isn't **totally** necessary. But it allows service clients to just call `Pipeline::send` and give it all the necessary data that needs to be stored in the closure used by `ResponseFuture`. Without taking those values by-value, the service client method would have to create an async closure manually, using `ResponseFuture::new`.

It felt reasonable to me for the pipeline to take those values. I recognize why **Policies** don't receive those by value (since policies may be executed multiple times on the same Context/Request pair) but it seemed from my reading that Context and Request were not intended to be shared across multiple separate requests.

[^1]: In this case, the user is expected to call `ResponseBody::json()` or `ResponseBody::xml()` manually. I wasn't a fan of that when it was in the primary use-case, but I'm OK with it when the user has explicitly decided to handle processing the response themselves.


sdk/core/azure_core/Cargo.toml
```diff
@@ -62,5 +62,4 @@ features = [
   "reqwest_rustls",
   "hmac_rust",
   "hmac_openssl",
-  "xml",
```
analogrelay, line 65: This was just duplicated, you can see it further up if you expand the context

sdk/identity/azure_identity/src/credentials/client_certificate_credentials.rs
```diff
@@ -255,7 +254,7 @@ impl ClientCertificateCredential {
             return Err(http_response_from_body(rsp_status, &rsp_body).into_error());
         }
 
-        let response: AadTokenResponse = rsp.deserialize_body_into().await?;
+        let response: AadTokenResponse = rsp.into_body().json().await?;
```
analogrelay, line 257: Azure Identity doesn't use the full pipeline, it calls the HTTP client itself. It's always used the "Raw Response" and manually deserialized.

sdk/typespec/typespec_client_core/src/http/mod.rs
```diff
@@ -22,12 +24,23 @@ pub use options::*;
...
 pub use url::Url;
 
+use bytes::Bytes;
+use futures::Stream;
+use std::pin::Pin;
+
+#[cfg(not(target_arch = "wasm32"))]
+pub type PinnedStream = Pin<Box<dyn Stream<Item = crate::Result<Bytes>> + Send + Sync>>;
+#[cfg(target_arch = "wasm32")]
+pub type PinnedStream = Pin<Box<dyn Stream<Item = crate::Result<Bytes>>>>;
```
analogrelay, line 42: Moved these types here because they are useful across a few different modules in `http`. This does mean it's been promoted from `typespec_client_core::http::response::PinnedStream` to `typespec_client_core::http::PinnedStream`

sdk/typespec/typespec_client_core/src/http/response_future.rs
```diff
@@ -0,0 +1,520 @@
...
+    Box<dyn FnOnce(ResponseBody) -> BoxedFuture<'a, Result<T, Error>> + Send + 'a>;
+
+/// Utility shorthand for a Boxed Future that takes a ResponseBody and "deserializes" the body, returning a BoxedFuture with the result.
+#[cfg(target_arch = "wasm32")]
+pub type Deserializer<'a, T> =
+    Box<dyn FnOnce(ResponseBody) -> BoxedFuture<'a, Result<T, Error>> + 'a>;
+
+/// Represents a response where the body has not been downloaded yet, but can be downloaded and deserialized into a value of type `T`.
+///
+/// You get a [`LazyResponse`] by calling [`ResponseFuture::lazy()`] on a [`ResponseFuture`] you received from a service API.
```
LarryOsterman, line 34: I'm being utterly pedantic here, but there's something about the term "Lazy" that is bothering me because it feels somewhat pejorative. 

My problem is that I'm not coming up with a better description of how to name it. Lazy has the advantage that it is succinct, whereas other descriptive terms (two step result?) are just way too wordy.

RickWinter: If it would help put you at ease, Lazy is used throughout the language
https://doc.rust-lang.org/std/cell/struct.LazyCell.html

analogrelay: Yeah, I largely agree @LarryOsterman , but it's definitely become a fairly industry-standard term for "A thing that isn't quite there yet". `DeferredResponse` could be another option, but I feel like `lazy()`/`LazyResponse<T>` is going to be the "expected" name, especially given @RickWinter 's point about its existing use in Rust.

heaths: I'd rather we stick with `Lazy` as well. It's all over not only in Rust but other languages as well. That, I'm just to lazy to think of a better term.

LarryOsterman: So be it. I'll put my pedantry back in its box.

analogrelay:
Re-drafting this. Hitting some snags around dealing with requests with bodies. With the current changes, because the functions all return `ResponseFuture<T>` instead of being async functions returning `Result<...>`, it's difficult to do asynchronous or fallible things before sending the request, because there's no support in Rust for custom "Future-likes" or "Result-likes" at this time. It's still completely possible to write the necessary code, but you have to wrap it in `ResponseFuture::new` manually and things get messy with lifetimes fast. I'll keep poking at it though. This isn't blocking, but I think the ideal of getting the 80% scenario to a single `.await?` is worth pursuing if we can.

heaths:
We should be able to do `Future`-likes, I thought, right? We just have to implement `IntoFuture<>`. As for custom `Result`-likes, I've been watching `try_trait_v2` for a couple years and it seems like maintainers are just bike shedding over names when the community seems largely unified on "just go with `FromResidual` like you have now and release it". 😔

analogrelay:
> We should be able to do Future-likes, I thought, right?

Not quite. Anything that implements `IntoFuture` can be `.await`ed. The problem is constructing the future out of an `async fn`. From what I can see, there's no way to turn an arbitrary `async fn` into a custom `Future`-like. There's no equivalent to [AsyncMethodBuilder](https://learn.microsoft.com/en-us/dotnet/csharp/asynchronous-programming/async-return-types#generalized-async-return-types-and-valuetasktresult) in C#, that I can find.

> I've been watching try_trait_v2 for a couple years

Same, and it's frustrating to see it languish so long.

---

Basically, the problem we have here is that creating a `ResponseFuture` requires two things:

1. A boxed async closure to perform the actual request (including any "pre-flight" requests[^1], if your API requires that) and get the final response
2. A boxed async closure to read and deserialize the response body, which has to be separate so that `.lazy()` can resolve the first future while deferring the deserialization.

I'm going to keep poking and see if I can find a clean way to do this. I'm not a huge fan of every service client having to look like this:

```rust
async fn replace_item<T: Serialize>(
    &self,
    item_id: impl AsRef<str>,
    item: T,
    options: Option<ItemOptions>,
) -> azure_core::ResponseFuture<T> {
    // Create the response future to return the raw response body
    // ResponseFuture::new expects an async closure returning Result<Response>, so fallible and async operations can be done here
    let raw_response_future = ResponseFuture::new(async {
        let url = self
            .container_url
            .with_path_segments(["docs", item_id.as_ref()]);
        let mut req = Request::new(url, azure_core::Method::Put);
        req.insert_headers(&partition_key.into())?;
        req.set_json(&item)?;
        self.pipeline
            .send(Context::new(), &mut req)
            .await
    });

    // Attach the JSON deserializer and return it.
    raw_response_future.json()
}
```

But if we're OK with that, I think we can work with something. Though when I did that, I started hitting lifetime issues around capturing the function parameters that might start to proliferate named lifetimes, which I'd really prefer not to do 😵.

I'll keep working a bit on it. It's not blocking my Cosmos work, it's just motivated by my desire to get the 80% scenario down to a single `.await?`.

Issue #1668, opened 2024-05-24
Make Pageable's stream item type be Send + Sync again


benny-n:
Hi, it looks like the changes introduced in #784, which were meant to address #720, were [reverted](https://github.com/Azure/azure-sdk-for-rust/pull/939/files#diff-7265da286ebc60fd2ec3d1bf4c4bc54dabb63cc223079290589549fbbf6f9c25L28) sometime later.

Without the changes in #784, I am having the exact same issue as in #720.
Was the revert intentional?
If it was, is there another way to resolve this issue? Specifically, how would one save a `Pageable` in a struct that must be `Sync`?
If not, is it possible to add these changes again?

Thanks in advance!

heaths:
Thank you for your report. Since we are working on official, supported crates and have no plans to release further versions of the previously unsupported crates, we have decided to close this issue. Some older issues were migrated to the new crates if they seemed relevant. If you feel that this issue was relevant even to newer code that has been or will be rewritten, please reopen and comment as to why you consider it relevant.

Issue #1796, opened 2024-09-10
Client methods traits should desugar async functions to return `impl Future`


heaths:
To avoid a valid lint, we should desugar `FooClientMethods` methods as `async fn foo() -> Result<Response<T>>` into `fn foo() -> impl Future<Result<Response<T>>>` instead.

See <https://github.com/Azure/azure-sdk-for-rust/pull/1773#discussion_r1735372332> for context.

analogrelay:
The main concern here is the `Send` bound (and to a lesser-degree, the `Sync` bound). In order to spawn a task in `tokio` and `async-std`, it must only capture `Send` data. This is required in order to support work-stealing, since that would require `Send`ing the data for a task from one thread to another.

If you use `async fn`, you have a trait like this:

```rust
trait DoAsyncThing {
    async fn do_async_thing(&mut self) -> bool;
}
```

Which is desugared to

```rust
trait DoAsyncThing {
    fn do_async_thing(&mut self) -> impl Future<Output = bool>;
}
```

Looking _only_ at the trait, it is not possible to verify that `do_async_thing` returns a `Future` that is also `Send`. Which means code like this fails (see [Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=226e677f9df343b2927c47d7a2c5e916)):

```rust
async fn spawn_async_thing(mut thing: impl DoAsyncThing + Send + 'static)
{
    tokio::task::spawn(async move {
        if !thing.do_async_thing().await {
            println!("nope!")
        }
    });
}
```

It fails because `tokio::spawn` requires that `thing.do_async_thing()` return a value that is `Future + Send`, but the trait does not guarantee that.

```
error: future cannot be sent between threads safely
   --> src/lib.rs:15:5
    |
15  | /     tokio::task::spawn(async move {
16  | |         if !thing.do_async_thing().await {
17  | |             println!("nope!")
18  | |         }
19  | |     });
    | |______^ future created by async block is not `Send`
```

However, I think we can move forward without being concerned about that *for now*, given a few mitigating circumstances:

1. This restriction only applies when using the trait through a generic parameter. For example, if there was a type `MyAsyncThing` that implemented `AsyncThing` and it's `do_async_thing()` **was** `Send`, then this code would work fine (see [Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=4076d3d0a2240cf58c2dd551140f99fc)):

```rust
[...]

PR #1836, opened 2024-10-07
Pager rename and refactor

analogrelay:
In working on pageable APIs, I found that we needed to return `Response<T>` values instead of just `T` values. Rather than explode the generic type (`Result<Pager<Response<T>>>`) it felt reasonable to have `Pager<T>` yield a stream of `Result<Response<T>>` itself. **In addition**, `Pager<T>` depends on `T` implementing a trait, `Continuable`, to extract the continuation value. If we want to return `Response<T>` (so that users can read response headers for each page), that gets very difficult and likely requires newtypes since both `Response<T>` and `Continuable` would be foreign types to service client crates and you can't implement a foreign trait on a foreign type.

So, here are a few refactors to `Pageable` that I think will make it easier to use:

1. Renamed to `Pager`, to align with the guidelines
2. The callback expects a `Result<(Response<T>, Option<C>)>` (where `C` is any `Send`able owned value). The expectation here is that the callback will return `Ok((response, Some(continuation)))` to indicate a continuation and `Ok((response, None))` to indicate the end of the pagination. This avoids the need for the `Continuable` trait entirely.
3. The `Pager` yields `Result<Response<T>>` values. This is really just shorthand. If we don't want to limit ourselves to HTTP here, `Pager<T>` could become `Pager<Response<T>>` without any issues, except that writing the type gets more verbose.

There's a full example in the doc comments to see how you would call this. Also `azure_data_cosmos::clients::ContainerClient::query_items` has been rewritten to use it.
 
I tried to see if we could get rid of the boxing by turning `Pager` into a trait and returning a gnarly (but largely inferred) generic type. That didn't work out, but we had boxing here before anyway and are network bound so I'm not too worried about it.

In addition, I was able to remove the `declare!` macro by using conditional compilation to include/exclude the `Send` bound. It makes for a slightly gnarly function definition (where clauses and individual trait bounds cannot be conditionally compiled, but `<T: Bound>` syntax **can**). I also removed the inner module trick since it doesn't seem like I'm getting the dead-code error on `project_ref` (which makes sense since it's _private_ to the struct, so I'm not sure why it was happening at all)


heaths:
Note: should see if this would close #1668

analogrelay:
So, I believe #1668 is tied to #1796. Requiring that `Pager<T>` be `Send + Sync` also means requiring that the futures it captures be `Send + Sync`, which requires that any futures _they_ capture and await be `Send + Sync`. Because of #1796, our trait `async fn`s don't have clear bounds for that. At least that how I _think_ things are playing out when I try to add `Send + Sync` bounds.

sdk/cosmos/azure_data_cosmos/src/constants.rs
```diff
@@ -9,8 +9,6 @@ use azure_core::{headers::HeaderName, request_options::ContentType};
 pub const QUERY: HeaderName = HeaderName::from_static("x-ms-documentdb-query");
 pub const PARTITION_KEY: HeaderName = HeaderName::from_static("x-ms-documentdb-partitionkey");
 pub const CONTINUATION: HeaderName = HeaderName::from_static("x-ms-continuation");
-pub const INDEX_METRICS: HeaderName = HeaderName::from_static("x-ms-cosmos-index-utilization");
-pub const QUERY_METRICS: HeaderName = HeaderName::from_static("x-ms-documentdb-query-metrics");
```
analogrelay, line 13: I'll replace these with typed headers (#1829) later.

heaths:
> Renamed to Pager, to align with the guidelines

Cool. That's what most other languages use anyway, I believe. I know Go and JS do off hand. The "pageable" was certainly inspired by my time on the Azure SDK for .NET team.

> The callback expects a Result<(Response<T>, Option<C>)> (where C is any Sendable owned value). The expectation here is that the callback will return Ok((response, Some(continuation))) to indicate a continuation and Ok((response, None)) to indicate the end of the pagination. This avoids the need for the Continuable trait entirely.

Could we use an opaque type if public instead? It's open to change if needed - why the REST API board doesn't allow top-level JSON array responses - and more obvious what the fields are.

> The Pager yields Result<Response<T>> values. This is really just shorthand. If we don't want to limit ourselves to HTTP here, Pager<T> could become Pager<Response<T>> without any issues, except that writing the type gets more verbose.

Our paging is already pretty HTTP-specific across most - if not all - languages, so I think this is fine.

sdk/typespec/typespec_client_core/src/http/pager.rs
```diff
@@ -0,0 +1,256 @@
...
+    /// });
+    /// ```
+    pub fn from_callback<
+        // This is a bit gnarly, but the only thing that differs between the WASM/non-WASM configs is the presence of Send bounds.
+        #[cfg(not(target_arch = "wasm32"))] C: Send + 'static,
+        #[cfg(target_arch = "wasm32")] C: 'static,
+        #[cfg(not(target_arch = "wasm32"))] F: Fn(Option<C>) -> Fut + Send + 'static,
+        #[cfg(target_arch = "wasm32")] F: Fn(Option<C>) -> Fut + 'static,
+        #[cfg(not(target_arch = "wasm32"))] Fut: Future<Output = Result<(Response<T>, Option<C>), typespec::Error>> + Send + 'static,
+        #[cfg(target_arch = "wasm32")] Fut: Future<Output = Result<(Response<T>, Option<C>), typespec::Error>> + 'static,
```
heaths, line 67: Could we use `cfg_if` instead? It would halve the number of lines for improved readability.

sdk/typespec/typespec_client_core/src/http/pager.rs
```diff
@@ -0,0 +1,256 @@
...
+    #[pin]
+    #[cfg(target_arch = "wasm32")]
+    stream: Pin<Box<dyn Stream<Item = Result<Response<T>, Error>>>>,
+}
+
+impl<T> Pager<T> {
+    /// Creates a [`Pager<T>`] from a callback that will be called repeatedly to request each page.
+    ///
+    /// This method expect a callback that accepts a single `Option<C>` parameter, and returns a `(Response<T>, Option<C>)` tuple, asynchronously.
+    /// The `C` type parameter is the type of the continuation. It may be any [`Send`]able type.
```
heaths, line 27: What was the point of getting rid of the `Continuable`? It was succinct, but was it creating more hassles? For an SDK, I'd rather we have clear names and intent, and unnamed (rather, unnamable) tuples erode that. We could use a concrete struct, but a trait like used before is flexible. Maybe unnecessarily so, though, which is why I'd love to understand more about why you replaced it with the tuple containing an `Option<C>`.

sdk/cosmos/azure_data_cosmos/src/constants.rs
heaths: Granted, but any point in deleting them now? Or maybe open a tracking issue?

heaths (changes requested):
Overall it looks great and the thoughts and design here are welcome! As noted, I really just have one outstanding question. We could discuss over Teams if it'd be easier or lengthy, or whatever.

sdk/typespec/typespec_client_core/src/http/pager.rs
analogrelay: Ooh, possibly! I'll look in to that

sdk/cosmos/azure_data_cosmos/src/constants.rs
analogrelay: They're unused, because currently the constants module is `pub(crate)`. I suppose that module can be public too though.

sdk/typespec/typespec_client_core/src/http/pager.rs
analogrelay: The problem is: What type do you implement it on? The `Response<T>` has the headers and in Cosmos the continuation token is stored in a response header, so you'd want to do something like this, inside the Cosmos crate:

```rust
impl<T> Continuable for Response<QueryResults<T>> {
    type Continuation: String;

    fn continuation(&self) -> Option<Self::Continuation> {
        self.get_optional_string("x-ms-continuation")
    }
}
```

But that's illegal, because both `Continuable` and `Response<T>` are foreign types:

image

You could do this:

```rust
impl<T> Continuable for QueryResults<T> {
    type Continuation: String;

    fn continuation(&self) -> Option<Self::Continuation> {
    }
}
```

But `QueryResults<T>` is the model type, and has no access to the headers.

Since the callback body can trivially extract the continuation, having it return the continuation value was the simplest option.

> It was succinct, but was it creating more hassles? For an SDK, I'd rather we have clear names and intent, and unnamed (rather, unnamable) tuples erode that.

I agree there. We have a few other options here. We could return a named struct of some kind instead of a tuple. Or we could use our own enum with named variants:

```rust
pub enum PaginationResult<T, C> { // Name TBD
    Continue { response: Response<T>, continuation: C },
    Finish { response: Response<T> },
}
```

I'm not totally sure those are much better, since they make the callback more verbose, but I'm game.

sdk/typespec/typespec_client_core/src/http/pager.rs
analogrelay: Alas, no. You can't invoke a macro within a generic arguments block or in where clauses. There are basically four ways we can do this:

1. Two completely separate `from_callback` functions, with duplicated bodies, marked with `#[cfg]` attributes and appropriate type bounds
2. This layout, with `#[cfg]` attributes on the trait bounds themselves.
3. Wrapping the entire declaration in a macro, invoked separately for each `#[cfg]` variant (this is what the old code did)
4. Possibly: A custom attribute macro we could write (I didn't see one on crates.io) to do this for us (something like `#[bound_if(not(target_arch = "wasm32"), Send)] C: 'static`)

I wasn't a fan of 3 in the original since it made error messages and F12 go-to-definition a little wonky, but I could live with it. What I like about option 2 is that while it's a little gnarly, it's fairly clear what's going on and you can largely ignore it. I did rearrange it a bit to this:

```rust
    pub fn from_callback<
        // This is a bit gnarly, but the only thing that differs between the WASM/non-WASM configs is the presence of Send bounds.
        #[cfg(not(target_arch = "wasm32"))] C: Send + 'static,
        #[cfg(not(target_arch = "wasm32"))] F: Fn(Option<C>) -> Fut + Send + 'static,
        #[cfg(not(target_arch = "wasm32"))] Fut: Future<Output = Result<(Response<T>, Option<C>), typespec::Error>> + Send + 'static,

        #[cfg(target_arch = "wasm32")] C: 'static,
        #[cfg(target_arch = "wasm32")] F: Fn(Option<C>) -> Fut + 'static,
        #[cfg(target_arch = "wasm32")] Fut: Future<Output = Result<(Response<T>, Option<C>), typespec::Error>> + 'static,
    >(
        make_request: F,
    ) -> Self {
```

sdk/typespec/typespec_client_core/src/http/pager.rs
analogrelay: Another note: Most of the time, your continuation token is _already an `Option<C>`_ (for whatever `C` you're using as a continuation value). For example, in Cosmos the code currently looks like this:

```rust
Pager::from_callback(move |continuation| {
    let mut req = base_req.clone();
    async move {
        if let Some(continuation) = continuation {
            req.insert_header(constants::CONTINUATION, continuation);
        }

        let resp = pipeline
            .send(Context::new(), &mut req, ResourceType::Items)
            .await?;
        let continuation_token =
            resp.headers().get_optional_string(&constants::CONTINUATION);

        Ok((resp, continuation_token))
    }
})
```

I don't know that trying to avoid the tuple gets us something clearer, but I'm still open to it if you disagree.

heaths:
> > Renamed to Pager, to align with the guidelines
> 
> Cool. That's what most other languages use anyway, I believe. I know Go and JS do off hand. The "pageable" was certainly inspired by my time on the Azure SDK for .NET team.

Actually, I remembered this morning they use `Poller`, but I think since we're doing the polling internally, `Pager` makes more sense. It's doing what a pager does as opposed to making the consumer poll the poller in a loop, which isn't idiomatic. I think if range-over-funcs existed earlier, the Go SDK would've been using that.

sdk/typespec/typespec_client_core/src/http/pager.rs
heaths: That layout had crossed my mind as well, but off hand wasn't sure how it'd look. Actually looks nice - more readable, certainly.

sdk/typespec/typespec_client_core/src/http/pager.rs
heaths: I like something like `PagerResult<T, C>` where it's coercible to a `Result<T>` and you can get an `Option<C>` from the `Continue` variant. Seems `None` would be equally fine for either variant. That wouldn't look to bad, would it?

Thanks for the context about the reason for change. Makes sense.

analogrelay:
> Actually, I remembered this morning they use Poller, but I think since we're doing the polling internally, Pager makes more sense.

I think Poller might be something else. The [Rust guidelines](https://azure.github.io/azure-sdk/rust_introduction.html#rust-client-methods-return) refer to `Pager` and `Poller` separately. The [Go SDK certainly has a Pager type](https://github.com/Azure/azure-sdk-for-go/blob/e15752b96b1cd695369fe56874f9b6e3c49821ab/sdk/azcore/runtime/pager.go#L35)

sdk/typespec/typespec_client_core/src/http/pager.rs
heaths: @JeffreyRichter do you think accepting and returning a continuation token will work in all paging cases? Seems to me it would, but you might have some insights on other paging workflows where that may not work.

In this design, we can get a continuation token from headers or body. But what about those services that still use `top` and `skip`? Do we manifest a continuation token that we use to keep track of the sliding window of results e.g., doing something akin got `skip += top`?

analogrelay: It's important to note that because the return type is `Option<C>` (instead of something concrete like `Option<String>`), `top/skip` continuation would absolutely work here. The callback could extract the number that need to be skipped for the next call and return that as their continuation value.

I can look at something using a concrete type. I'm a little concerned we'll be adding unnecessary verbosity to APIs that are used _predominantly_ by service client authors (as they're the main audience who are _creating_ `Pager<T>` instances) but I can spend some more cycles on it.

heaths:
You're right. I'm getting my wires cross. `Poller` is for LROs. Too many balls in the air right now. 🤹

sdk/typespec/typespec_client_core/src/http/pager.rs
```diff
@@ -6,7 +6,33 @@ use std::{future::Future, pin::Pin};
...
 
-use crate::http::Response;
+use crate::http::{headers::HeaderName, Response};
+
+pub enum PagerResult<T, C> {
+    Continue {
+        response: Response<T>,
+        continuation: C,
+    },
+    Finish {
```
heaths, line 16: Nit (but I could pick it up in another PR): most languages use "complete" for something like this.

heaths (approved):
Just a nit about naming, but otherwise LG(reat)TM!

analogrelay:
Just need a code review from some Cosmos owners to merge now, since part of this PR updates the Cosmos client to use the refactored Pager (ping @FabianMeiswinkel @Pilchie @kirankumarkolli @kundadebdatta @tvaron3 , let me know if you have any questions! Happy to go over anything)

Pilchie (approved):
I looked at it before, and it looked fine to me, I just didn't want to sign-off before the broader discussion happened.

RickWinter:
[heart]         Rick Winter reacted to your message:
________________________________

PR #1837, opened 2024-10-07
Migrate Rust emitter sources

jhendrixMSFT:
Taken from https://github.com/Azure/autorest.rust/commit/cfe50fb with changes to account for directory structure and package naming.


heaths (changes requested):
Why did we move it wholesale here? I thought we were going to make the old repo public until its final home is established by EngSys? I'd rather not have to mix TS and rust dev environments here. There's already enough maintenance with just rust because they keep changing the nightly toolchain against which we test.

/cc @RickWinter

RickWinter:
> Why did we move it wholesale here? I thought we were going to make the old repo public until its final home is established by EngSys? I'd rather not have to mix TS and rust dev environments here. There's already enough maintenance with just rust because they keep changing the nightly toolchain against which we test.
> 
> /cc @RickWinter

This is the model for all languages moving forward.

/cc @lmazuel

jhendrixMSFT:
Remaining work items.
- don't run SDK CI for changes to `/eng/packages/http-client-rust`
- create pipeline for emitter
- add pipeline for publishing cadl-ranch results to the dashboard
- ensure we have a pipeline for publishing dev/official emitter packages

jhendrixMSFT:
Current POR will keep emitter sources in a separate repo.

PR #1843, opened 2024-10-10
[Cosmos] Query Databases, Query Containers, and expand ContainerProperties to include all relevant properties

analogrelay:
This PR adds the following new APIs:

* `CosmosClient::query_databases` to query across the databases in an account
* `DatabaseClient::query_containers` to query across the containers in an account

To do this, I was able to make a common `send_query_request` to our internal `CosmosPipeline` wrapper, which makes the individual query APIs fairly compact.

In addition, I added missing properties to `ContainerProperties`, based on the Go SDK. I also added Vector Indexing properties, which aren't currently in the Go SDK.

I took another attempt at removing the `#[allow(unused_variables)]` from our `options` parameters. As far as I can tell, there is no escaping it. If an impl is in the same crate as the trait it's implementing, the fact that the trait requires a parameter is not sufficient to suppress the `unused_variables` lint (see [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e02381c4d1333ed9ab7ce48ee2516fad)). Annoying, but appears to be by-design, though I may follow up on the rust-lang users discourse board.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/examples/cosmos/metadata.rs
+++ b/sdk/cosmos/azure_data_cosmos/examples/cosmos/metadata.rs
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+use std::error::Error;
+
 use azure_data_cosmos::{
     clients::{ContainerClientMethods, DatabaseClientMethods},
     CosmosClient, CosmosClientMethods,
@@ -19,7 +21,7 @@ pub struct MetadataCommand {
 }
 
 impl MetadataCommand {
-    pub async fn run(self, client: CosmosClient) -> Result<(), Box<dyn std::error::Error>> {
+    pub async fn run(self, client: CosmosClient) -> Result<(), Box<dyn Error>> {
         let db_client = client.database_client(&self.database);
         if let Some(container_name) = &self.container {
             let container_client = db_client.container_client(container_name);
--- a/sdk/cosmos/azure_data_cosmos/examples/cosmos/metadata.rs
+++ b/sdk/cosmos/azure_data_cosmos/examples/cosmos/metadata.rs
@@ -19,7 +19,7 @@ pub struct MetadataCommand {
 }
 
 impl MetadataCommand {
-    pub async fn run(&self, client: CosmosClient) -> Result<(), Box<dyn std::error::Error>> {
+    pub async fn run(self, client: CosmosClient) -> Result<(), Box<dyn std::error::Error>> {
         let db_client = client.database_client(&self.database);
         if let Some(container_name) = &self.container {
             let container_client = db_client.container_client(container_name);
--- a/sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 use crate::clients::DatabaseClient;
-use crate::models::{DatabaseProperties, QueryResults};
+use crate::models::DatabaseQueryResults;
 use crate::pipeline::{AuthorizationPolicy, CosmosPipeline, ResourceType};
 use crate::utils::AppendPathSegments;
 use crate::{CosmosClientOptions, Query, QueryDatabasesOptions};
@@ -64,7 +64,7 @@ pub trait CosmosClientMethods {
         &self,
         query: impl Into<Query>,
         options: Option<QueryDatabasesOptions>,
-    ) -> azure_core::Result<azure_core::Pager<QueryResults<DatabaseProperties>>>;
+    ) -> azure_core::Result<azure_core::Pager<DatabaseQueryResults>>;
 }
 
 impl CosmosClient {
@@ -155,7 +155,7 @@ impl CosmosClientMethods for CosmosClient {
         #[allow(unused_variables)]
         // REASON: This is a documented public API so prefixing with '_' is undesirable.
         options: Option<QueryDatabasesOptions>,
-    ) -> azure_core::Result<azure_core::Pager<QueryResults<DatabaseProperties>>> {
+    ) -> azure_core::Result<azure_core::Pager<DatabaseQueryResults>> {
         let mut url = self.endpoint.clone();
         url.append_path_segments(["dbs"]);
         let base_request = Request::new(url, azure_core::Method::Post);
--- a/sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
@@ -2,10 +2,12 @@
 // Licensed under the MIT License.
 
 use crate::clients::DatabaseClient;
-use crate::pipeline::{AuthorizationPolicy, CosmosPipeline};
-use crate::CosmosClientOptions;
+use crate::models::{DatabaseProperties, QueryResults};
+use crate::pipeline::{AuthorizationPolicy, CosmosPipeline, ResourceType};
+use crate::utils::AppendPathSegments;
+use crate::{CosmosClientOptions, Query, QueryDatabasesOptions};
 use azure_core::credentials::TokenCredential;
-use azure_core::Url;
+use azure_core::{Request, Url};
 use std::sync::Arc;
 
 #[cfg(feature = "key_auth")]
@@ -31,6 +33,15 @@ pub trait CosmosClientMethods {
     /// # Arguments
     /// * `id` - The ID of the database.
     fn database_client(&self, id: impl AsRef<str>) -> DatabaseClient;
+
+    /// Gets the URL of the Cosmos DB Account this client is connected to.
+    fn endpoint(&self) -> &Url;
+
+    fn query_databases(
+        &self,
+        query: impl Into<Query>,
+        options: Option<QueryDatabasesOptions>,
+    ) -> azure_core::Result<azure_core::Pager<QueryResults<DatabaseProperties>>>;
 }
 
 impl CosmosClient {
@@ -98,11 +109,6 @@ impl CosmosClient {
             options,
         })
     }
-
-    /// Gets the endpoint of the database account this client is connected to.
-    pub fn endpoint(&self) -> &Url {
-        &self.endpoint
-    }
 }
 
 impl CosmosClientMethods for CosmosClient {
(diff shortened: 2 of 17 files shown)


sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -113,4 +142,25 @@ impl CosmosClientMethods for CosmosClient {
...
+        &self.endpoint
+    }
+
+    fn query_databases(
+        &self,
+        query: impl Into<Query>,
+
+        #[allow(unused_variables)]
+        // REASON: This is a documented public API so prefixing with '_' is undesirable.
+        options: Option<QueryDatabasesOptions>,
```
Pilchie, line 157: Can you have a different parameter name in the trait impl than in the trait itself?  Maybe like in C, where you put the parameter name in the .h file, but you omit it if it's unused in the .c file, you'd rename it to have the `_` here, but leave the original name in the Trait?

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
```diff
@@ -0,0 +1,402 @@
...
+    Ok(Option::<u64>::deserialize(deserializer)?.map(Duration::from_secs))
+}
+
+/// Properties of a Cosmos DB container.
+///
+/// Returned by [`ContainerClient::read()`](crate::clients::ContainerClient::read()).
+#[non_exhaustive]
+#[derive(Model, Clone, Debug, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct ContainerProperties {
```
Pilchie, line 24: You could consider adding the [`vectorEmbeddingPolicy`](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/4f515bb56f6687b61e780bc77f306b24b09fde7b/Microsoft.Azure.Cosmos/src/Resource/Settings/VectorEmbeddingPolicy.cs#L21) stuff here.  It's preview only in .NET right now, but already in some of the other SDKs.

Pilchie (approved):

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
```diff
@@ -40,28 +41,63 @@ pub trait DatabaseClientMethods {
...
+    ///
+    /// * `query` - The query to execute.
+    /// * `options` - Optional parameters for the request.
+    ///
+    /// # Examples
+    ///
+    /// The `query` parameter accepts anything that can be transformed [`Into`] a [`Query`].
+    /// This allows simple queries without parameters to be expressed easily:
+    ///
+    /// ```rust,no_run
```
heaths, line 67: Nit: no reason to state the `rust` block itself. It's always assumed, just ` ```no_run ` is fine...unless you're putting it into markdown and importing like I'm doing for READMEs.

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
```diff
@@ -0,0 +1,402 @@
...
+/// Represents the partition key definition for a container.
+#[non_exhaustive]
+#[derive(Clone, Debug, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct PartitionKeyDefinition {
+    /// The list of partition keys paths.
+    pub paths: Vec<String>,
+
+    /// The version of the partition key hash in use.
+    pub version: Option<usize>,
```
heaths, line 68: Is version truly unbounded by Cosmos itself? Though, perhaps this doesn't matter since it's only for deserialization but would also seem to put more burden on clients since [numbers default to i32](https://doc.rust-lang.org/book/ch03-02-data-types.html#:~:text=integer%20types%20default%20to%20i32) if no other signal is provided.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
analogrelay: Parameter names don't need to match between trait and impl, so this can absolutely just be `_options`, which suppresses the lint. The problem is that shows up that way in the docs, which I'd rather not do (I'll grab a screenshot and reply with it in a sec). I suppose it's not _harmful_ as all it does is hint to the user that the implementation doesn't actually **use** the parameter, but it feels... well... clunky. That's the intent of this "REASON" comment (which would eventually be the `reason` attribute of the `#[allow]` attribute, once we update our minimum rustc version to or beyond 1.81), to explain why we're doing an `#[allow]` here instead of prefixing with `_`.

Another option I've seen before, which I didn't like as much, is to "use" the parameter in a statement that will be compiled away. That looks something like this:

```rust
async fn create_item<T: Serialize + DeserializeOwned>(
    &self,
    item: T,
    options: Option<ItemOptions>,
) -> Result<Response<Item<T>>> {
    let _ = options;
    // .. the rest of the function
}
```

My personal thought was that I found that less clear than specifying exactly what lint we were addressing, but I'm open to it.

sdk/cosmos/azure_data_cosmos/src/models/mod.rs
```diff
@@ -39,7 +41,9 @@ where
 #[non_exhaustive]
 #[derive(Clone, Default, Debug, Deserialize)]
 pub struct QueryResults<T> {
-    #[serde(rename = "Documents")]
+    #[serde(alias = "Documents")]
+    #[serde(alias = "Databases")]
+    #[serde(alias = "DocumentCollections")]
```
heaths, line 46: I'm a little concerned these could diverge after a time. [REST API guidelines](https://aka.ms/azapi/guidelines) dictate services never return a top-level array because it's inflexible. You're not doing that here, which is good; however, it implies because you can add fields in the future. So, could a Documents vs. Databases vs. DocumentCollections result ever add differing fields which would necessitate a different type or at least using an enum here instead?

heaths (approved):
A few thoughts, but I see nothing blocking.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
Pilchie: Yeah, not necessarily suggesting we change it to that, just speculating on the reasoning behind it.

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
analogrelay: I didn't realize that, but it makes sense! I've put the `rust` tag in everywhere else, so if I do remove it, I'll just do a separate pass.

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
analogrelay: It's not really unbounded, [the .NET SDK does use an enum for this](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.partitionkeydefinitionversion?view=azure-dotnet). I was using a number primarily to avoid breaking deserialization later when a new version is returned (though we have other enums that will probably break when new values are used) and because it is intended to be a simple integer version number. The Go SDK uses a bare integer, but Go doesn't really have enums anyway.

I'll make this an enum for now. It does raise the idea of having a way to represent "Unknown" values in all of our enums, but that's a larger issue across the rest of Cosmos (and likely other SDKs)

analogrelay: After digging a bit in to serde's support for enums-as-numbers, I think I'm going to keep this a number for now. Serde doesn't directly support enums-as-numbers, it requires using the [`serde_repr`](https://crates.io/crates/serde_repr) crate, or a custom `Deserialize` implementation. Adding the dependency isn't a dealbreaker, but given the desire to potentially support new versions without updating the client, I think it's reasonable to leave this a number for now. I did change it to `i32` to match with Rust's default for numbers.

sdk/cosmos/azure_data_cosmos/examples/cosmos/metadata.rs
```diff
@@ -19,7 +19,7 @@ pub struct MetadataCommand {
 }
 
 impl MetadataCommand {
-    pub async fn run(&self, client: CosmosClient) -> Result<(), Box<dyn std::error::Error>> {
+    pub async fn run(self, client: CosmosClient) -> Result<(), Box<dyn std::error::Error>> {
```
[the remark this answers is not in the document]
analogrelay: Honestly, inconsistency 😅 , I'll fix that up.

Usually, I do have a preference for fully-qualifying a type like `std::error::Error` when it's highly likely to conflict with a crate-local `Error` type.

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
analogrelay: I'm going to hold on that for now, I don't yet have an account set up where I can verify that this data is properly being deserialized from the server's response, so I'll look into that and then add support for it.

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
heaths: I'd rather not take more dependencies as well. My main concern was about the integer type itself, but if you and your team think that most comparisons would be done in a way where rustc can infer a number as a `usize` because it's only used in a comparison or something for this use case, that's fine.

Pilchie: I don't think we need to future proof this too much.  We're only on hash v2 and we're 10 years into this thing ;)

analogrelay: So you're saying we're due ;P

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
Pilchie: No problem - though to set up the account, it's just a toggle on the features blade: https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/vector-search#enroll-in-the-vector-search-preview-feature

Pilchie: (But that can defnitely come later)

analogrelay: Ah yeah, I actually have messed with that before in a different account. I can probably add it in quick.

analogrelay: It was indeed pretty straightforward. I've added it now:

```
vector_embedding_policy: Some(
        VectorEmbeddingPolicy {
            embeddings: [
                VectorEmbedding {
                    path: "/vector1",
                    data_type: Float32,
                    dimensions: 10,
                    distance_function: Euclidean,
                },
                VectorEmbedding {
                    path: "/vector2",
                    data_type: Uint8,
                    dimensions: 10,
                    distance_function: Cosine,
                },
                VectorEmbedding {
                    path: "/vector3",
                    data_type: Int8,
                    dimensions: 4096,
                    distance_function: DotProduct,
                },
            ],
        },
    ),
```

analogrelay:
Added vector embedding stuff. While I don't _need_ a reapproval, I wouldn't mind a quick second look @Pilchie . I'll probably just merge this in later today though, and we can always do a little tidy up afterwards if I missed something. I've verified the vector embedding deserialization against a test account.

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
```diff
@@ -0,0 +1,468 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct VectorEmbedding {
+    /// The path to the property containing the vector.
+    pub path: String,
+
+    /// The data type of the elements stored in the vector.
+    pub data_type: VectorDataType,
+
+    /// The number of dimensions in the vector.
+    pub dimensions: u64,
```
Pilchie, line 84: We're in the process of changing this to `u32` elsewhere.  It was a C++ long -> C#/Java long that caused initial confusion.

Pilchie (commented):
Looks good except the note about the fact that we're changing the size of the dimensions property to be 32 bit.

PR #1844, opened 2024-10-10
Repo setup to support tsp-client generation

catalinaperalta:
Setting the configs necessary to run tsp-client in this repo.


heaths:
This needs to target the `feature/track2` branch.

PR #1845, opened 2024-10-10
Setup repo to support client generation with tsp-client

catalinaperalta:
Adding configs necessary to generate client libraries with tsp-client.

Example of how to update emitter-package.json in the future:
```
PS C:\repos\azure-sdk-for-rust> .\eng\common\scripts\typespec\New-EmitterPackageJson.ps1 -PackageJsonPath <path to>\typespec-rust\packages\typespec-rust\package.json -OutputDirectory .\eng\
```

**_There will be a tsp-client specific command coming soon!_**

To sync the lock file with the emitter-package.json file run: `tsp-client generate-lock-file`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ Cargo.lock
 *.swo
 *.swp
 *.vim
+
+# temporary folder to refresh SDK with TypeSpec
+TempTypeSpecFiles/
--- a/eng/emitter-package-lock.json
+++ b/eng/emitter-package-lock.json
@@ -1,0 +1,2624 @@
+{
+    "name": "TempTypeSpecFiles",
+    "lockfileVersion": 3,
+    "requires": true,
+    "packages": {
+        "": {
+            "dependencies": {
+                "@azure-tools/typespec-rust": "0.1.0"
+            },
+            "devDependencies": {
+                "@azure-tools/typespec-client-generator-core": ">=0.46.2 <1.0.0",
+                "@typespec/compiler": ">=0.60.1 <1.0.0",
+                "@typespec/xml": ">=0.60.1 <1.0.0"
+            }
+        },
+        "node_modules/@azure-tools/async-io": {
+            "version": "3.0.254",
+            "resolved": "https://registry.npmjs.org/@azure-tools/async-io/-/async-io-3.0.254.tgz",
+            "integrity": "sha512-X1C7XdyCuo50ch9FzKtTvmK18FgDxxf1Bbt3cSoknQqeDaRegHSSCO+zByq2YA4NvUzKXeZ1engh29IDxZXgpQ==",
+            "license": "MIT",
+            "dependencies": {
+                "@azure-tools/tasks": "~3.0.255",
+                "proper-lockfile": "~2.0.1"
+            },
+            "engines": {
+                "node": ">=10.12.0"
+            }
+        },
+        "node_modules/@azure-tools/codegen": {
+            "version": "2.9.2",
+            "resolved": "https://registry.npmjs.org/@azure-tools/codegen/-/codegen-2.9.2.tgz",
+            "integrity": "sha512-brVLyffOtPiEijYYBYgV+4q7IyAfqXIec7XbdEqvv7As6SeEdq5WtbtN9N0LdGVHDWtEfc+JArwIx9aYGFdMUg==",
+            "license": "MIT",
+            "dependencies": {
+                "@azure-tools/async-io": "~3.0.0",
+                "js-yaml": "~4.0.0",
+                "semver": "^7.3.5"
+            },
+            "engines": {
+                "node": ">=12.0.0"
+            }
+        },
+        "node_modules/@azure-tools/codegen/node_modules/js-yaml": {
+            "version": "4.0.0",
+            "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.0.0.tgz",
+            "integrity": "sha512-pqon0s+4ScYUvX30wxQi3PogGFAlUyH0awepWvwkj4jD4v+ova3RiYw8bmA6x2rDrEaj8i/oWKoRxpVNW+Re8Q==",
+            "license": "MIT",
+            "dependencies": {
+                "argparse": "^2.0.1"
+            },
+            "bin": {
+                "js-yaml": "bin/js-yaml.js"
+            }
+        },
+        "node_modules/@azure-tools/linq": {
+            "version": "3.1.263",
+            "resolved": "https://registry.npmjs.org/@azure-tools/linq/-/linq-3.1.263.tgz",
+            "integrity": "sha512-r104pkF96Sk8MW+PgxWqwiwNXhPbJSkMDcmShu9EDEJem04SM7DnHWmDo8JZR08vHYaecZA126OoGQFe7WikSA==",
+            "license": "MIT",
+            "engines": {
+                "node": ">=10.12.0"
+            }
+        },
+        "node_modules/@azure-tools/tasks": {
+            "version": "3.0.255",
+            "resolved": "https://registry.npmjs.org/@azure-tools/tasks/-/tasks-3.0.255.tgz",
+            "integrity": "sha512-GjALNLz7kWMEdRVbaN5g0cJHNAr3XVTbP0611Mv2UzMgGL6FOhNZJK+oPHJKLDR8EEDZNnkwPlyi7B+INXUSQA==",
+            "license": "MIT",
+            "engines": {
+                "node": ">=10.12.0"
+            }
+        },
+        "node_modules/@azure-tools/typespec-azure-core": {
+            "version": "0.47.0",
+            "resolved": "https://registry.npmjs.org/@azure-tools/typespec-azure-core/-/typespec-azure-core-0.47.0.tgz",
+            "integrity": "sha512-RcBC5+dE1BVXTrUkkKULTImGxzM/ea3P3IL2kr9pk7r1uqF7D4CGqEKHFTg5L6QUtqc1f+zgTgQTNn6t4gI92w==",
+            "license": "MIT",
+            "peer": true,
+            "engines": {
+                "node": ">=18.0.0"
+            },
+            "peerDependencies": {
+                "@typespec/compiler": "~0.61.0",
+                "@typespec/http": "~0.61.0",
+                "@typespec/rest": "~0.61.0"
+            }
+        },
+        "node_modules/@azure-tools/typespec-client-generator-core": {
+            "version": "0.47.1",
+            "resolved": "https://registry.npmjs.org/@azure-tools/typespec-client-generator-core/-/typespec-client-generator-core-0.47.1.tgz",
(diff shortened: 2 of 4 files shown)


eng/emitter-package.json
```diff
@@ -0,0 +1,11 @@
+{
+    "main": "dist/src/index.js",
+    "dependencies": {
+        "@azure-tools/typespec-rust": "0.1.0"
+    },
+    "devDependencies": {
+        "@azure-tools/typespec-client-generator-core": ">=0.46.2 <1.0.0",
+        "@typespec/compiler": ">=0.60.1 <1.0.0",
+        "@typespec/xml": ">=0.60.1 <1.0.0"
```
catalinaperalta, line 9: NOTE: I manually added this for storage, however we can remove if it's not needed for now. This should be added in the peerDependencies of the typespec-rust package for it to get carried over to emitter-package.json whenever you use the script to sync dependencies.

jhendrixMSFT: Do all tsp packages need to be listed as peer dependencies in an emitter's `package.json`? I thought it was only the ones the emitter directly uses, but I could be wrong.

catalinaperalta: That is the convention the emitters have been following, this allows emitters to pin supported peer dependency ranges. We used to just manually pin these deps in emitter-package.json but I believe @hallipr worked on shifting to using the peer deps in the emitter's package.json so he could share more. From a tsp-client perspective we simply install whatever deps are defined in emitter-package.json. Examples:
[typespec-python](https://github.com/Azure/autorest.python/blob/f3f6491ca1d653b584b092b07f7907237cc00c4e/packages/typespec-python/package.json#L47-L58)
[typespec-csharp](https://github.com/Azure/autorest.csharp/blob/b11bb94bc90e29694ab222e3b30cb43f167cb949/src/TypeSpec.Extension/Emitter.Csharp/package.json#L41-L53)

heaths:
Can we pass a URL instead of local file path? Currently, we have no plans to colocate the emitter source in this repo. Can we also specify it via config?

.gitignore
```diff
@@ -12,3 +12,6 @@ Cargo.lock
 *.swo
 *.swp
 *.vim
+
+# temporary folder to refresh SDK with TypeSpec
+TempTypeSpecFiles/
```
heaths, line 17: Why is this in the repo at all? We should use a subdirectory under `$TEMP`, which is universal across platforms.

eng/emitter-package-lock.json
heaths, line 1: We should add this file to `.gitattributes` like we have `Cargo.lock` files already:

```gitattributes
*package-lock.json merge=binary
```

heaths (changes requested):

catalinaperalta:
> Can we pass a URL instead of local file path? Currently, we have no plans to colocate the emitter source in this repo. Can we also specify it via config?

@hallipr can you help answer this? On the tsp-client side, I can take this into consideration for the new command that will replace the pwsh script.

eng/emitter-package-lock.json
catalinaperalta: This wont block maintainers from updating the file?

.gitignore
catalinaperalta: this is current tsp-client behavior, if you really want a change for this behavior please open an issue for tsp-client here and we can discuss: https://github.com/Azure/azure-sdk-tools/issues

I dont think this should block this PR from merging though if you want to start using tsp-client to generate client libraries, it's an implementation detail of the tool.

eng/emitter-package-lock.json
heaths: No. It just means it can't be merged. It's treated *in PRs only* as a binary file. It's still stored in git as text so storage is efficient. But it basically doesn't allow merges because this file should *never* be merged. Same as any lock file, which is why we already have `Cargo.lock` in there like this (serves the same purpose as `packages-lock.json`).

eng/emitter-package.json
hallipr: These shouldn't be specified as range dependencies.  We started pinning any floating peer dependencies with dev dependencies to reduce the risk of supply chain vulnerabilities caused by picking up a version in CI that we haven't yet verified on a dev machine.  Even though the emitter-package-lock.json will pin us to a specific version, we can't control which version we pin to without specifying the version in devDependencies.

`npm install` will automatically satisfy deep peer dependencies. This means that locally, your package.json will make no mention of the version to pick but your package-lock.json will pin to the latest available version that satisfies the different required ranges. 

Because we generate emitter-package.json and its lock file automatically in emitter CI builds, we can't use lock file pinning alone, so we came up with the heuristic of "any dependency mentioned as both a peer dependency and dev dependency must be doing peer dependency pinning, so we'll forward that pin on to emitter-package.json"

This is codified in: [eng/common/scripts/typespec/New-EmitterPackageJson.ps1
](https://github.com/Azure/azure-sdk-for-rust/blob/feature/track2/eng/common/scripts/typespec/New-EmitterPackageJson.ps1#L33)

package.json:
```json
{
  "name": "@azure-tools/typespec-foo",
  "version": "0.35.1",
  "peerDependencies": {
    "@azure-tools/typespec-azure-core": ">=0.46.0 <1.0.0",
    "@azure-tools/typespec-azure-resource-manager": ">=0.46.0 <1.0.0",
    "@typespec/compiler": ">=0.60.0 <1.0.0",
    "another-package": "2.3.4"
  },
  "devDependencies": {
    "@azure-tools/typespec-azure-core": "~0.46.0",
    "@azure-tools/typespec-azure-resource-manager": "~0.46.0",
    "@typespec/compiler": "~0.60.1",
    "something-else": "1.2.3"
  }
}
```

emitter-package.json
```json
{
  "main": "dist/src/index.js",
  "dependencies": {
    "@azure-tools/typespec-foo": "0.35.1"
  },
  "devDependencies": {
    "@azure-tools/typespec-azure-core": "~0.46.0",
    "@azure-tools/typespec-azure-resource-manager": "~0.46.0",
    "@typespec/compiler": "~0.60.1",
  }
}
```

As far as which packages should be included, you can run `npm ls -a`.  If you have any open peer dependencies, it will fail and list them.

hallipr: see: https://github.com/Azure/azure-sdk-for-python/blob/main/eng/emitter-package.json

hallipr (changes requested):

eng/emitter-package.json
hallipr: I have a PR out that will make emitter-package.json peer pinning more explicit in the emitter's package.json file:
https://github.com/Azure/azure-sdk-tools/pull/9147

Rather than specifying both peerDependencies and devDependencies in the emitter's package.json you pin the dev dependency and add the package name to a new section:

```json
{
  "name": "@azure-tools/typespec-python",
  "version": "0.35.1",
  "devDependencies": {
    "@azure-tools/typespec-azure-core": "~0.46.0",
    "@azure-tools/typespec-azure-resource-manager": "~0.46.0",
    "@typespec/compiler": "~0.60.1",
    "something-else": "1.2.3"
  },
  "azure-sdk/emitter-package-json-pinning": [
    "@azure-tools/typespec-azure-core",
    "@azure-tools/typespec-azure-resource-manager",
    "@typespec/compiler"
  ]
}
```

This avoids redeclaring transitive peer dependencies, it's less heuristic and it's more declarative.

catalinaperalta: @hallipr so is this PR blocked until the emitter's package.json is updated to this format?

catalinaperalta: @jhendrixMSFT I see several packages that also have ranges in devDependencies: https://github.com/Azure/typespec-rust/blob/a8c4021148d50dc3f54a44f7fd3489568c1e9dba/packages/typespec-rust/package.json#L64-L68 I think we should be pinning to a specific version according to what @hallipr mentioned.

jhendrixMSFT: I thought the comment was for `emitter-package.json`? Other emitters use the same range which is desirable in order to test emitters against dev versions.

@hallipr can we get clarification on this?

catalinaperalta: @hallipr if this is a new requirement, can we merge as is and then update the emitter-package.json once your new script is updated?

hallipr: `emitter-package.json` should pin all transitive peer dependency ranges as dev dependencies to the same version that was used when testing the emitter.  The only way to reliably get the same versions in package-lock.json and emitter-package-lock.json is to be specific in package.json's devDependencies and automate the creation of emitter-package.json.

If specific version pinning in devDependencies is too much of a headache, we could consider pulling the specific version from `package-lock.json` rather than `package.json`'s devDependencies when creating `emitter-package.json`

catalinaperalta: @jhendrixMSFT thoughts on pinning specific versions in devDeps?

I see c# and python pinning to specific versions already: 
C#: https://github.com/Azure/autorest.csharp/blob/d3615842dd87f222debf00bb1ca3beac0e953942/src/TypeSpec.Extension/Emitter.Csharp/package.json#L54-L81
Python: https://github.com/Azure/autorest.python/blob/333bc8e4b352f487ea24159fde23474fc3679e20/packages/typespec-python/package.json#L67-L92

catalinaperalta: @hallipr have you already updated you script? since last time I ran it, it still just pulled in the peerDeps...otherwise, we'll have to manually update emitter-package.json to unblock this pr

.gitignore
heaths: https://github.com/Azure/azure-sdk-tools/issues/9248

eng/emitter-package.json
hallipr: I would manually update the emitter-package.json, but use static pinning instead of ranged.

catalinaperalta: Done! Will dismiss the request for changes

jhendrixMSFT (approved):
I don't know enough about how this should be structured but the values for the Rust emitter look correct.

catalinaperalta:
> I don't know enough about how this should be structured but the values for the Rust emitter look correct.

This is the final shape. The only action for you is to pin your devDependencies in the typespec-rust emitter to a specific version instead of a range (otherwise you'll have to do it manually in your emitter-package.json updates). In the future you'll follow these steps to update this file with new rust emitter releases:

1. To sync emitter-package.json with the typespec-rust package.json, run:
```
PS C:\repos\azure-sdk-for-rust> .\eng\common\scripts\typespec\New-EmitterPackageJson.ps1 -PackageJsonPath <path to>\typespec-rust\packages\typespec-rust\package.json -OutputDirectory .\eng\
```

> NOTE: I'll let you know when tsp-client has a command for this so you can stop using the pwsh script

2. To sync your emitter-package-lock.json, run:
```
tsp-client generate-lock-file
```

And of course feel free to reach out if you have any questions!

RickWinter (approved):

eng/emitter-package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "main": "dist/src/index.js",
+  "dependencies": {
+    "@azure-tools/typespec-rust": "0.1.0"
+  },
+  "devDependencies": {
+    "@azure-tools/typespec-client-generator-core": "0.47.1",
```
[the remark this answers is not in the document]
catalinaperalta: Yes if 0.47.1 worked with 0.1.0 we should be fine for this PR. Resolving!

weshaggard (approved):

PR #1850, opened 2024-10-14
[page-blob] introduce IfMatch header to put request

goksgie:
The Azure Docs do not mention, but we can prevent accidental overwrites to Azure Page Blobs when two parties issue a create request. This can be prevented through IfMatch headers, as this commit introduces.

As of today, the only alternative option to prevent overwrite would be managing an internal blob service, or checking whether it exists before issuing a create request (which introduces an additional round-trip).

For C# equivalent, please see: https://github.com/Azure/azure-sdk-for-net/blob/fb0b80d4c8fc99e9e964862f2a374d2bc28096f7/sdk/storage/Azure.Storage.Blobs/src/PageBlobClient.cs#L844

### DIFF
--- a/sdk/storage_blobs/src/blob/operations/put_page_blob.rs
+++ b/sdk/storage_blobs/src/blob/operations/put_page_blob.rs
@@ -15,6 +15,7 @@ operation! {
     ?metadata: Metadata,
     ?tags: Tags,
     ?lease_id: LeaseId,
+    ?if_match: IfMatchCondition,
     ?sequence_number: SequenceNumber
 }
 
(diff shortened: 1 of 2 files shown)


goksgie:
@microsoft-github-policy-service agree company="Microsoft"

heaths (changes requested):
The `main` branch is unsupported and should not be used in production. You might want to check if this is already implemented in the `feature/track2` branch or retarget there if not.

goksgie:
hey @heaths , thanks for quick response. I don't see any code under storage directory in the branch you've suggested. I assume that is still under development.

But, please redirect me otherwise. Thanks!

heaths:
The storage team is working on a new crate. Keep in mind what's under `main` right now is unsupported.

heaths:
`main` branch has been moved to `legacy`. Retargeting some old PRs.

goksgie:
@heaths , would it be possible to merge this to "legacy" branch? We'll possibly use cargo syntax to consume this specific change to simplify the code until we acquire more clarity around official Azure Storage SDK.

Issue #1846, opened 2024-10-11
Remove create_enum and create_extensible_enum macros


heaths:
LSP refactoring tools and LLMs have problems with the `create_enum` and `create_extensible_enum` macros because they aren't defined as enums e.g., Copilot will grep for "enum EnumName" and won't find it.

While the macros are handy, proc macros are just code generators. And we already will generate most of our code via Azure/typespect-rust. Thus, we should just expand the enums we have now in core and other hand-written files.

The emitter already has an issue to remove use of them: Azure/typespec-rust#611

And rather than create a proc macro as the original description here describes, as the discussion below points out: with all the "wrapping" of serde attributes we could do but probably shouldn't, we end up having to type 90% or more of the enum anyway e.g., the type declaration, the variants, serde attributes, other derives e.g., `SafeDebug`, etc. We could even mark the non-unknown variants as `#[safe(true)]` because they are defined values, but leave the whole type as unsafe so that the `UnknownValue` isn't traced by default.

And given the complexity of writing, debugging, and maintaining a proc macro, it's just not worth it for the last 10% or less when we already use a code emitter for more code.

heaths:
In the meantime, I created a `create_extensible_enum!()` macro in #1852 as a stopgap solution for @jhendrixMSFT for the TypeSpec emitter.

analogrelay:
I like `#[variant(other)]` and had a similar thought that we might want a way to support unknown values from the server.

> Should we try to duplicate necessary serde attributes like rename_all? Off hand, that may be all we need.

If we're hand-implementing the `serde` traits, I think we would _have_ to duplicate those attributes, since the `#[serde]` attribute is only in scope when using their derive macro. I believe `rename_all` would be sufficient, though we might need `rename` as well to handle irregular cases.

---

Since we're not actually deriving a trait, it might also be reasonable to use a general [attribute macro](https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros), rather than a `#[derive]` macro. I don't _mind_ using a `#[derive]` one if that's what feels easiest, but it might be a little confusing to someone reading the code.

heaths:
> Since we're not actually deriving a trait, it might also be reasonable to use a general [attribute macro](https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros), rather than a `#[derive]` macro. I don't _mind_ using a `#[derive]` one if that's what feels easiest, but it might be a little confusing to someone reading the code.

[...]

PR #1852, opened 2024-10-14
Add create_extensible_enum macro

heaths:


heaths:
This uses a little "trick" of characters really only usable in macros to effectively make a match arm "internal". We can use this for shared code for DRY. I probably could've gone a little further with some creative match arms, but I think that'd diminish readability and these macros are hopefully temporary anyway (see #1846).

sdk/typespec/typespec_client_core/src/macros.rs
```diff
@@ -189,27 +290,63 @@ mod test {
     }
 
     #[test]
-    fn test_color_parse_1() {
+    fn color_parse_1() {
```
heaths, line 293: Note: I don't like "test_" prefixes when in a "test(s)" module. It's redundant. "test_" is great when tests are in the same module e.g.,

```rust
fn foo() {}

#[test]
fn test_foo() {}
```

jhendrixMSFT (approved):

analogrelay (approved):

LarryOsterman (approved):
I'm annoyed by the redundancy associated with having to say `Black, "Black"` and the potential for subtle bugs, but absent adding the ability for rust to introduce a tokenizing operator, I don't see a way of avoiding it.

heaths:
> I'm annoyed by the redundancy associated with having to say `Black, "Black"` and the potential for subtle bugs, but absent adding the ability for rust to introduce a tokenizing operator, I don't see a way of avoiding it.

It actually wouldn't be that hard and I had considered it, but it's of dubious value because 1) these are temporary, and 2) they wouldn't likely be used in emitters because it'd be harder - and rather pointless - to vary the emitted code. I suppose, also, 3) it's not that much more to type in what few hand-coded cases there should be but, again, see point 1.

PR #1853, opened 2024-10-14
[Cosmos] Control-Plane Over Data-Plane APIs (Database/Container CRUD)

analogrelay:
This PR adds the "control-plane over data-plane" APIs (basically Database and Container CRUD) to the Cosmos Rust SDK.

These APIs _require_ that you authenticate with a key. Using Entra ID to call these APIs will result in an error (instead, you're supposed to use ARM management APIs if you want to use Entra ID). So, in order to clarify that, I've placed a warning in the API docs:

image

This warning comes from a single shared markdown file, so we can edit it in one place.

In addition, I've placed these APIs behind a feature flag, `control_plane`. Enabling this feature flag _automatically_ enables the `key_auth` feature flag, because you have to use key-based auth and it wouldn't make sense to have one without the other.

If a user were referencing our crate from their app, they'd enable these APIs using this syntax in the `Cargo.toml` file:

```toml
[dependencies]
azure_data_cosmos = { version = "...", features = [ "control_plane" ] }
```

We're hoping to guide users towards Entra ID as much as possible, hence putting key-based auth behind a feature flag. Removing these feature flags later should be a non-breaking change, so I'd rather do it now and seek customer feedback rather than trying to remove/limit these APIs later.

You can test these APIs out with new commands on the `cosmos` "meta-example" (we provide a single example with multiple subcommands rather than separate examples. The `create` and `delete` commands are broken up in to `item`, `database` and `container` subcommands. However, because of the feature flags, to see the `database` and `container` ones, you'll need to run the example like this:

```bash
$ cargo run --example cosmos --features control_plane -- [endpoint] delete database [db name]
# etc..
```

There's no way, that I can see, to "enable" a feature flag in an example, so you have to do it manually when running the example. Not ideal, but I think the only alternative is to move the example to its own crate.

I've tested these APIs manually but we're waiting on getting Test Proxy integrated into this repo before we can build fully automated integration tests. Hopefully that's coming soon!


Cargo.toml
```diff
@@ -56,6 +56,7 @@ base64 = "0.22"
 bytes = "1.0"
 cargo_metadata = "0.18.1"
 clap = { version = "4.4.16", features = ["derive"] }
+cfg-if = "1.0.0"
```
analogrelay, line 59: I used [`cfg-if`](https://docs.rs/cfg-if/latest/cfg_if/) to simplify some conditional compilation in `azure_data_cosmos`. It's a relatively low-impact crate, but I'm open to removing it and going back to a series of `#[cfg(...)]`-attributed definitions if that's preferred. (cc @heaths )

sdk/cosmos/azure_data_cosmos/examples/cosmos/create.rs
```diff
@@ -4,42 +4,129 @@ use azure_data_cosmos::{
     clients::{ContainerClientMethods, DatabaseClientMethods},
     CosmosClient, CosmosClientMethods, PartitionKey,
 };
-use clap::Args;
+use clap::{Args, Subcommand};
 
-/// Creates a new item.
+#[cfg(feature = "control_plane")]
```
analogrelay, line 9: Annoyingly, it sees we need to conditionally-compile our `use` statements as well to avoid 'unused import' lints. I'd prefer it if clippy considered all features (maybe we just need to run it with `--all-features` 🤔 ) since extraneous imports don't impact the functionality, but alas.

sdk/cosmos/azure_data_cosmos/src/models/mod.rs
```diff
@@ -54,23 +57,20 @@ impl<T: DeserializeOwned> azure_core::Model for QueryResults<T> {
 }
 
 /// A page of results from [`CosmosClient::query_databases`](crate::clients::CosmosClient::query_databases())
-#[non_exhaustive]
```
analogrelay, line 57: Per ongoing internal discussions, we're removing `#[non_exhaustive]` from any types used as **request** models, since `#[non_exhaustive]` types _cannot be directly constructed outside the defining crate_ (which means we'd have to generate a whole bunch of constructors, setters, and builders). However, I removed them from all types here. I'm going back through and restoring them on the Response-only models in the next iteration. I'll wait to resolve this comment until that's done.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -9,6 +9,19 @@ mod query_options;
 mod read_container_options;
 mod read_database_options;
 
+cfg_if::cfg_if! {
```
analogrelay, line 12: This is the usage of `cfg_if`. Without it, we'd have to repeat the `#[cfg]` attribute on each individual `mod` and `pub use` line.

sdk/cosmos/azure_data_cosmos/src/pipeline/authorization_policy.rs
```diff
@@ -21,7 +21,7 @@ use url::form_urlencoded;
 #[cfg(feature = "key_auth")]
 use azure_core::{credentials::Secret, hmac::hmac_sha256};
 
-const AZURE_VERSION: &str = "2018-12-31";
+const AZURE_VERSION: &str = "2020-07-15";
```
analogrelay, line 24: Creating containers with hierarchical partition keys requires this version. I grabbed the string from the JS SDK.

sdk/cosmos/azure_data_cosmos/docs/control-plane-warning.md
```diff
@@ -0,0 +1,5 @@
+<div class="warning">
+
+This is a control-plane API and requires that you authenticate using a key. To use Entra ID to perform this operation, you must use the [Azure Resource Manager APIs](https://learn.microsoft.com/en-us/azure/templates/microsoft.documentdb/databaseaccounts).
```
Pilchie, line 3: I think this PR does the right thing for the point in time we're at, but just pointing out that I believe there is a plan to support these operations via Entra ID at some point in 2025, and also to integrate the dataplane RBAC with normal Azure RBAC.

sdk/cosmos/azure_data_cosmos/docs/control-plane-warning.md
```diff
@@ -0,0 +1,5 @@
+<div class="warning">
```
Pilchie, line 1: Are divs part of markdown?

sdk/cosmos/azure_data_cosmos/src/options/delete_container_options.rs
```diff
@@ -0,0 +1,37 @@
...
+
+#[cfg(doc)]
+use crate::clients::ContainerClientMethods;
+
+/// Options to be passed to [`ContainerClient::delete()`](crate::clients::ContainerClient::delete()).
+#[derive(Clone, Debug, Default)]
+pub struct DeleteContainerOptions {}
+
+impl DeleteContainerOptions {
+    /// Deletes a new [`DeleteContainerOptionsBuilder`](DeleteContainerOptionsBuilder) that can be used to construct a [`DeleteContainerOptions`].
```
Pilchie, line 12: A little overzealous find and replace here?

sdk/cosmos/azure_data_cosmos/src/options/delete_database_options.rs
```diff
@@ -0,0 +1,37 @@
...
+
+#[cfg(doc)]
+use crate::clients::DatabaseClientMethods;
+
+/// Options to be passed to [`DatabaseClient::delete()`](crate::clients::DatabaseClient::delete()).
+#[derive(Clone, Debug, Default)]
+pub struct DeleteDatabaseOptions {}
+
+impl DeleteDatabaseOptions {
+    /// Deletes a new [`DeleteDatabaseOptionsBuilder`](DeleteDatabaseOptionsBuilder) that can be used to construct a [`DeleteDatabaseOptions`].
```
Pilchie, line 12:
```suggestion
    /// Creates a new [`DeleteDatabaseOptionsBuilder`](DeleteDatabaseOptionsBuilder) that can be used to construct a [`DeleteDatabaseOptions`].
```

sdk/cosmos/azure_data_cosmos/src/options/delete_container_options.rs
Pilchie:
```suggestion
    /// Creates a new [`DeleteContainerOptionsBuilder`](DeleteContainerOptionsBuilder) that can be used to construct a [`DeleteContainerOptions`].
```

sdk/cosmos/azure_data_cosmos/src/models/container_definition.rs
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under the MIT License.
+
+use serde::Serialize;
+
+use crate::models::{IndexingPolicy, PartitionKeyDefinition};
+
+/// Describes the properties used to create a new container.
+#[derive(Clone, Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ContainerDefinition {
```
Pilchie, line 11: Aren't there more things here - throughput options, VectorEmbeddingPolicy, etc?

Ok to start with this and add more if we can do it non-breaking though.

sdk/cosmos/azure_data_cosmos/src/options/delete_container_options.rs
analogrelay: Yep. Definitely looking forward to having some macros or something to reduce duplication here 😅

sdk/cosmos/azure_data_cosmos/docs/control-plane-warning.md
analogrelay: Yep, the intent is that this feature can be "removed" later. It will have to remain present, and continue to imply `key_auth` to avoid breaking users, but the `#[cfg]` attributes referencing it would go away and the methods related to it would just become available to everyone. At that point users would be able to remove the feature reference from their Cargo.toml (possibly replacing with `key_auth` if they still need that support)

sdk/cosmos/azure_data_cosmos/docs/control-plane-warning.md
analogrelay: Markdown generally allows any HTML as well. More importantly, this is part of the rustdoc set of supported markdown: https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#adding-a-warning-block

sdk/cosmos/azure_data_cosmos/src/models/container_definition.rs
analogrelay: I've added the remaining things. Basically, I removed `ContainerDefinition` and am using `ContainerProperties` for both the Create request body and the Read response body. This mirrors what we do in Go.

The `SystemProperties` struct we use to contain things like `_etag` is still marked `#[non_exhaustive]` though, which means it's not constructable using struct syntax (`SystemProperties {...}` will produce an error outside the defining crate). However, the `SystemProperties` crate implements `Default`, so `SystemProperties::default()` can be used to construct an "empty" value. 

I think this benefits us, since we don't want users to try to _set_ those values anyway (even if they're ignored by the server). So, to construct a `ContainerProperties`, you have to use this syntax:

```rust
db_client.create_container(
    ContainerProperties {
        id: "MyContainer".to_string(),
        partition_key: "/partitionKey".into(),
        ..Default::default() // Set all remaining fields to their default
    }, None);
```

Using `..Default::default()` is allowed, because it uses `SystemProperties::default()` (provided by it's own implementation of `Default`) to construct that object, which results in all the properties being `None`. We **already** want users to use this syntax to create _all_ of our request models, so that we can safely add new properties later. So, this should work out fine.

I added a test which should serve to make sure we don't add non-optional properties to `ContainerProperties` in the future, and retain the same JSON serialization whenever a user uses the basic syntax above.

analogrelay: > throughput options

These are actually specified as headers and appear on the `options` type in other SDKs like Go, so they'll come later.

Pilchie: This looks great.  One thing that occurred to me is that we might need to do something extra work with deserialization of `ContainerProperties` and its nested types.  To support older SDKs talking to newer versions of the service, we might need to be able to roundtrip `ContainerProperties` for things we don't know about in the SDK. The scenario is that I want to modify some property, so I query the current properties, make some change, and then set the properties.  If we don't searialize everything we get back, we might lose options.

AFAIK, .NET handles this with [`[JsonExtensionData]`](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/f88605c207801bf305e9dc890c026e208cb4d041/Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerProperties.cs#L87C1-L92C88)

analogrelay: I believe serde does support a similar mechanism, I'll look in to it (but might merge this first, as you suggest in your review comment)

analogrelay: Well, unfortunately I'm wrong. Serde doesn't provide a built-in way to do this. We do have some options to implement it (dropping to manually implementing `Serialize` and `Deserialize` if necessary) but it doesn't look like there's a built-in feature for this.

I don't believe all the other SDKs do this, I know .NET has `[JsonExtendedData]` but I don't see similar things for Go, so I think it's reasonable for us to consider it later on.

analogrelay: Ok, actually it looks like `#[serde(flatten)]` can be used for this purpose: https://serde.rs/attr-flatten.html#capture-additional-fields

Still, going to hold off for now, but at least I have an idea of how we could do it.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
heaths, line 12: Why can't you just use the built-in `cfg`? If it's a scoping thing, it only takes a couple extra lines:

```rust
#[cfg(feature = "control_plane")]
mod control_plane {
  mod create_container_options;
  // ...
  pub use create_container_options::CreateContainerOptions;
  // ...
}
#[cfg(feature = "control_plane")]
pub use control_plane::*;
```

Separately, why is there a feature called "control_plane" in a data plane SDK? If you need it, does it matter that you use a control plane API? Just wondering why it's optional.

analogrelay: > Why can't you just use the built-in cfg? If it's a scoping thing, it only takes a couple extra lines

I can rework it to that.

> Separately, why is there a feature called "control_plane" in a data plane SDK?

Operations to create, update, and delete Databases and Containers are considered Control Plane operations, and they can be performed via ARM. However, the Cosmos DB **Data Plane** _also_ provides these APIs (which I've referred to here as "Control Plane Over Data Plane" operations). This SDK is calling those **Data Plane** operations. Hence the name of the feature flag being `control_plane`, though I appreciate it might not be the clearest name.

> Just wondering why it's optional.

The reason it's optional is something we raised earlier on in our discussions about this SDK. These "Control Plane Over Data Plane" operations _require_ key-based auth. They _cannot_ be performed with an Entra ID token. There is ongoing work to change this, but it's not committed to happening before this SDK GAs.

We _could_ just put these in the same `key_auth` feature flag as the rest of our key-based authentication, and now that I think through that, I might prefer that. I originally thought granularity might be good, but these APIs wouldn't be behind a feature flag _except_ for the fact that they depend on the `key_auth` feature, so there isn't really a need for a separate flag.

analogrelay:
An update, I've removed `cfg-if` dependency, but instead of using @heaths 's suggestion, I actually just removed the conditional compilation altogether.

I decided **not** to make the "Control Plane Over Data Plane" APIs use conditional compilation after all. I realized that if we add support for Entra ID tokens to the server-side APIs at a later date, it would be helpful if we didn't **require** users to update the SDK to a version which didn't conditionally-compile out these APIs. In the spirit of the SDKs not over-validating things that the server controls, I decided it would be better to leave the APIs, include the warning in the docs, and let them "just work" (if possible) when the server is updated. If the client ends up needing to be updated to support Entra ID for these APIs, so be it, but I'd rather not impose that restriction now.

If you attempt to use these APIs while authenticated with Entra ID, the error message is _fairly_ clear and searchable:

> Request blocked by Auth ashley-cosmos-testing : The given request [POST /dbs] cannot be authorized by AAD token in data plane. Learn more: https://aka.ms/cosmos-native-rbac

It's also notable that the conditional compilation didn't prevent you from seeing this error. Even when you compile-in the feature, you could still have authenticated with Entra ID and gotten this error.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
heaths: I remembered that conversation, which is why I wondered why it was optional but, given `key_auth` also was (that conversation I clearly forgot), I agree it makes sense to unify them. I just didn't see much point in making those CP-over-DP features optional.

PR #1855, opened 2024-10-15
Update enum macros according to guidelines

heaths:


sdk/typespec/typespec_client_core/src/macros.rs
```diff
@@ -26,7 +26,8 @@
 macro_rules! create_enum {
     ($(#[$type_meta:meta])* $name:ident, $($(#[$value_meta:meta])* ($variant:ident, $value:expr)), *) => (
         $(#[$type_meta])*
-        #[derive(Debug, PartialEq, Eq, PartialOrd, Clone, Copy)]
+        #[derive(Debug, PartialEq, Eq, Clone, Copy)]
+        #[non_exhaustive]
```
jhendrixMSFT, line 30: Is this really applicable for non-extensible enums?

heaths: Is what really applicable? `PartialEq` and `Eq`? Yeah, so they are comparable.

jhendrixMSFT: Is `#[non_exhaustive]` applicable for non-extensible enums. Which are, by definition, a closed set of values that will never change (sans a breaking change). If we want to keep this here for versioning resiliency then it makes sense.

heaths: Yes, because the service team can still add values later on. This isn't about "extensible enums" but about changes in future versions.

jhendrixMSFT (approved):

RickWinter (approved):

PR #1859, opened 2024-10-17
Fix code fence in AMQP README

heaths:


heaths:
@larryosterman just a little issue I noticed when looking through the current state of docs.

LarryOsterman (approved):

PR #1861, opened 2024-10-21
[Cosmos] Refactoring how the resource type and link are passed down to the auth policy

analogrelay:
In working on some stuff related to "offers" and throughput, I noticed that the way we were inferring the resource "link" when generating the payload to be signed for key-based auth wasn't going to work for offers since they use a specific format of link. So, I decided to do some refactoring here to make it so the values for generating signatures are _passed in_ to the policy (through the context) rather than inferred. The resource type was already being passed in, so I added a new type to represent the resource link (it'll need some special case stuff for offers, but this way that can be centralized there) and updated the authorization policy to use that value when generating the signature.

The value we put in the context is `ResourceLink`, which wraps up:

* The parent resource link
* The resource type
* The child item id, if any (a missing ID implies that the resource in question is a "feed")

The `ResourceLink` type also handles URL encoding to ensure that URL-unsafe characters are encoded before the URL and signature are generated. I've included APIs to allow you to build child feed/item links out of parent links.


sdk/cosmos/azure_data_cosmos/src/resource_context.rs
```diff
@@ -0,0 +1,233 @@
...
+impl ResourceLink {
+    pub fn root(resource_type: ResourceType) -> Self {
+        Self {
+            parent: None,
+            resource_type,
+            item_id: None,
+        }
+    }
+
+    pub fn feed(&self, resource_type: ResourceType) -> Self {
```
heaths, line 60: If this is exported and customers are meant to construct these (or can), builder setters should use `with_feed`. It's also recommended in guidelines to take `mut self` in builder methods since references can cause issues in some cases e.g., you can't just "stop" in the middle until you call the final `build()`.

heaths (approved):
Based on context it doesn't look like your builder-like struct is being publicly exported, but I'd still recommend following conventions for uniformity throughout the code base. If that is correct, I see nothing blocking.

sdk/cosmos/azure_data_cosmos/src/resource_context.rs
analogrelay: These aren't quite builder APIs in the same was as usual. It's more of an immutable value that allows you to quickly create a clone "child" value. The idea is the `DatabaseClient` holds a `ResourceLink` for the database and then uses these APIs to create the child link to give to a `ContainerClient`.

The APIs take references because they are intended to leave the original `ResourceLink` alone and create a completely new `ResourceLink` that is "descended" from `self`. Also, there's no final `build` step, every returned `ResourceLink` is complete enough to be used.

Pilchie (approved):

PR #1863, opened 2024-10-22
[Cosmos] remove Methods traits

analogrelay:
After some discussion with the SDK team, we've decided to remove the `...Methods` traits that were the initial mechanism by which we provided "mockability". Traits in Rust aren't quite as straightforward to use as interfaces in .NET/Java and came with a few downsides.

So, for now, we're removing them, since adding them later (or taking one of a few other alternative paths to "mockability") can be done without introducing a breaking change, plus we're not GA yet and could take a potential breaking change.

This PR removes all the `...Methods` traits from `azure_data_cosmos` and converts the methods that _were_ implemented through those traits to be directly implemented on the client types themselves. No functional changes to the implementations themselves, just code re-organization.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/examples/cosmos/create.rs
+++ b/sdk/cosmos/azure_data_cosmos/examples/cosmos/create.rs
@@ -1,9 +1,8 @@
 use std::error::Error;
 
 use azure_data_cosmos::{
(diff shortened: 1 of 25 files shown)


heaths (approved):
Nice! Much cleaner.

Pilchie (approved):

PR #1864, opened 2024-10-22
Remove duplicate re-export of Model macro

heaths:
Also re-exports `Model` from `azure_core` along with some other types that were exported from the crate root in track 1.


LarryOsterman (approved):

RickWinter (commented):
Nice clean up, thanks for fixing

heaths:
For posterity, this may clean up some incidental compile-time errors where the macro was imported from `typespec_client_core::Model` but everything else in `typespec_client_core::http` was out of scope.

PR #1865, opened 2024-10-23
Add binary request example

heaths:
Also refactors `FileStream` from `azure_core` to `typespec_client_core`, plus some trait implementations on necessary request structures to make it all work.

### DIFF
--- a/sdk/typespec/typespec_client_core/examples/binary_data_request.rs
+++ b/sdk/typespec/typespec_client_core/examples/binary_data_request.rs
@@ -1,0 +1,84 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+use tokio::fs;
+use tracing::level_filters::LevelFilter;
+use tracing_subscriber::fmt::format::FmtSpan;
+use typespec_client_core::{
+    fs::FileStreamBuilder,
+    http::{Body, RequestContent},
+};
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Log traces to stdout.
+    let _log = tracing_subscriber::fmt()
+        .with_span_events(FmtSpan::ENTER | FmtSpan::EXIT)
+        .with_max_level(LevelFilter::DEBUG)
+        .init();
+
+    // Asynchronously read the whole file into memory.
+    let body = RequestContent::from(fs::read(file!()).await?);
+    client::put_binary_data(body).await?;
+
+    // Asynchronously stream the file with the service client request.
+    #[cfg(not(target_arch = "wasm32"))]
+    {
+        let file = fs::File::open(file!()).await?;
+        let file = FileStreamBuilder::new(file)
+            // Simulate a slow, chunky request.
+            .buffer_size(256usize)
+            .build()
+            .await?;
+        let body: Body = file.into();
+        client::put_binary_data(body.into()).await?;
+    }
+
+    Ok(())
+}
+
+mod client {
+    use futures::StreamExt;
+    use tracing::debug;
+    use typespec_client_core::{
+        http::{headers::Headers, Body, RequestContent, Response, StatusCode},
+        stream::BytesStream,
+    };
+
+    #[tracing::instrument(skip(body))]
+    pub async fn put_binary_data(
+        body: RequestContent<()>,
+    ) -> typespec_client_core::Result<Response<()>> {
+        let body: RequestContent<()> = body.into();
+        let body: Body = body.into();
+
+        let content = match body {
+            Body::Bytes(ref bytes) => {
+                debug!("received bytes");
+                bytes.to_owned()
+            }
+            Body::SeekableStream(mut stream) => {
+                debug!("received stream");
+                let stream = stream.as_mut();
+
+                let mut bytes = Vec::new();
+                while let Some(buf) = stream.next().await {
+                    debug!("reading partial request...");
+                    bytes.extend(buf?);
+                }
+
+                bytes.into()
+            }
+        };
+
+        // Assume bytes are a string in this example.
+        let content = String::from_utf8(content.into())?;
+        println!("{content}");
+
+        Ok(Response::new(
+            StatusCode::NoContent,
+            Headers::new(),
+            Box::pin(BytesStream::new_empty()),
+        ))
+    }
+}
--- a/sdk/typespec/typespec_client_core/examples/binary_data_request.rs
+++ b/sdk/typespec/typespec_client_core/examples/binary_data_request.rs
@@ -4,10 +4,7 @@
 use tokio::fs;
 use tracing::level_filters::LevelFilter;
 use tracing_subscriber::fmt::format::FmtSpan;
-use typespec_client_core::{
-    fs::FileStreamBuilder,
-    http::{Body, RequestContent},
-};
+use typespec_client_core::{fs::FileStreamBuilder, http::RequestContent};
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -30,8 +27,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             .buffer_size(256usize)
             .build()
             .await?;
-        let body: Body = file.into();
-        client::put_binary_data(body.into()).await?;
+        client::put_binary_data(file.into()).await?;
     }
 
     Ok(())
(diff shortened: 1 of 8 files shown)


heaths:
I need to play around with this some. Track 1 didn't have a fully asynchronous answer either. I'm thinking the answer is to convert from a `Body` to `RequestContent<T>`, and add support for `Stream` into `Body`. We have a `tokio`-based `FileStream` already that I need to migrate to `typespec_client_core` (if necessary; might be outdated, actually).

Basically, it'd go:

```rust
let file_stream = FileStream::new(tokio::fs::File::open("data.bin")?);
let body: Body = file_stream.into();
client.put_data(&ctx, body, None).await?;
```

/cc @analogrelay @demoray @cataggar

sdk/typespec/typespec_client_core/examples/binary_data_request.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[cfg(not(target_arch = "wasm32"))]
+    {
+        let file = fs::File::open(file!()).await?;
+        let file = FileStreamBuilder::new(file)
+            // Simulate a slow, chunky request.
+            .buffer_size(256usize)
+            .build()
+            .await?;
+        let body: Body = file.into();
+        client::put_binary_data(body.into()).await?;
```
heaths, line 34: Alternatively, maybe we need a simpler (less hops) way to get a stream into a body. We already `impl From<Box<dyn SeekableStream>> for Body` but it's not the most intuitive to make the extra hop. Maybe we support going straight to a `RequestContent<T>`?

Note: ignore the `buffer_size()` call. It's to simulate a slow (small buffer as a proxy) client request only.

sdk/typespec/typespec_client_core/src/http/request/mod.rs
```diff
@@ -217,6 +217,26 @@ impl<T> From<Body> for RequestContent<T> {
...
+#[cfg(all(feature = "tokio_fs", not(target_arch = "wasm32")))]
+impl<T> From<&crate::fs::FileStream> for RequestContent<T> {
+    fn from(body: &crate::fs::FileStream) -> Self {
+        Self {
+            body: body.into(),
+            phantom: PhantomData,
+        }
+    }
+}
+
```
heaths, line 239: @RickWinter @JeffreyRichter @analogrelay @demoray @cataggar what do you think? Think this may be common enough for request data streaming to add first-tier support for streaming a file directly into `RequestContent`? I'm leaning that way, but maybe 60/40 or so. I generally agree with "expose things as you need them" and we certainly have several ways to get binary data - even streaming (well, one way) - into a request, but I think this greatly simplifies the API for a common case *for some* services e.g., Storage (or anything that can take a file upload directly even if it ends up in Storage).

JeffreyRichter: I agree with you on the "expose things as you need them" and so I'm hesitant to add this now.
In the early Rust days, we had streaming, and we took it out - I remember Yosh was very happy when we did this because it is complicated when you also consider retries and having to rewind/seek the stream back to the beginning.

sdk/typespec/typespec_client_core/src/http/request/mod.rs
```diff
@@ -208,6 +208,35 @@ impl<T> From<RequestContent<T>> for Body {
     }
 }
 
+impl<T> From<Body> for RequestContent<T> {
```
analogrelay, line 211: So, is the intent here is to have APIs be able to take `RequestContent<T>` and for those APIs be able to take in a `T` to be serialized **or** raw bytes/streams (which I presume would be pre-serialized) to be sent? I'm trying to understand the `T` in this context. The example you show elsewhere takes `ResponseContent<()>`, and I get the case for supporting an arbitrary file stream there, but I'm not clear where we need support for `ResponseContent<T>`.

For the Cosmos DB APIs I've implemented so far, the request body is accepted as just a plain `T` parameter (like `properties: ContainerProperties` when creating a Cosmos DB Container), which gets serialized and added to the body. Of course, that's just one small set of a large Azure SDK surface area, so I'm sure there a number of other use cases I'm not clear on, just trying to understand the purpose of `RequestContent<T>`.

sdk/typespec/typespec_client_core/examples/binary_data_request.rs
```diff
@@ -0,0 +1,80 @@
...
+    use futures::StreamExt;
+    use tracing::debug;
+    use typespec_client_core::{
+        http::{headers::Headers, Body, RequestContent, Response, StatusCode},
+        stream::BytesStream,
+    };
+
+    #[tracing::instrument(skip(body))]
+    pub async fn put_binary_data(
+        body: RequestContent<()>,
```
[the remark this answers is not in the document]
heaths: No, body should be just `RequestContent<T>` since that has all the stuff to go `From` other types; besides, I'm leaning toward taking concrete parameter types. See the internal thread I CC'd you et. al. on.

sdk/typespec/typespec_client_core/src/http/request/mod.rs
heaths: But because of the orphan rule, crates that do need it - say, `azure_storage_blobs` - couldn't add this themselves if they want to support streaming a large file in a request.

JeffreyRichter: What we did in other languages is take a generic stream (file, network or something else). The stream must be readable and seekable (for retries). Is there a generic stream in Rust standard library we could use?

heaths: In `std`, no. In fact, there's hardly anything in `std`. But we do declare our own `SeekableStream`. `Body` already has a `From<Box<dyn SeekableStream>>` so we could add one here. It still eliminates a line but callers will have to box the stream. It's not so elegant:

```rust
    let file = FileStream::open("foo.txt")?;
    let bytes = read(Box::new(file))?;

    println!("read {} bytes", bytes.len());

    Ok(())
}
```

Still, not so bad either.

heaths: Actually, this is even uglier:

```rust
        let file = fs::File::open(file!()).await?;
        let file: Box<dyn SeekableStream> = Box::new(
            FileStreamBuilder::new(file)
                // Simulate a slow, chunky request.
                .buffer_size(256usize)
                .build()
                .await?,
        );
        client::put_binary_data(file.into()).await?;
```

I think we should keep the explicit `From<FileStream>` and `From<&FileStream>` but can colocate them into `tokio.fs` like the ones for `Body` are now.

sdk/typespec/typespec_client_core/examples/binary_data_request.rs
```diff
@@ -0,0 +1,80 @@
...
+    ) -> typespec_client_core::Result<Response<()>> {
+        let body: RequestContent<()> = body.into();
+        let body: Body = body.into();
+
+        let content = match body {
+            Body::Bytes(ref bytes) => {
+                debug!("received bytes");
```
RickWinter, line 53: Is it helpful to put a bit more debug output into this statement? Maybe the number of bytes received?

sdk/typespec/typespec_client_core/examples/binary_data_request.rs
```diff
@@ -0,0 +1,80 @@
...
+                    bytes.extend(buf?);
+                }
+
+                bytes.into()
+            }
+        };
+
+        // Assume bytes are a string in this example.
+        let content = String::from_utf8(content.into())?;
+        println!("{content}");
```
RickWinter, line 72: Any risk here of dumping PII? Should there be a note surrounding this println about contents could be senstive in nature?

sdk/typespec/typespec_client_core/examples/stream_response.rs
```diff
@@ -2,11 +2,19 @@
...
+use tracing::level_filters::LevelFilter;
+use tracing_subscriber::fmt::format::FmtSpan;
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Log traces to stdout.
+    let _log = tracing_subscriber::fmt()
+        .with_span_events(FmtSpan::ENTER | FmtSpan::EXIT)
+        .with_max_level(LevelFilter::DEBUG)
+        .init();
```
RickWinter, line 14: Do we get sanitization for free? If not we should be careful to not dump access tokens

RickWinter (approved):

sdk/typespec/typespec_client_core/examples/binary_data_request.rs
heaths: I could, but it's only a sample. Typically what it emits isn't as important as how it works.

sdk/typespec/typespec_client_core/examples/binary_data_request.rs
heaths: There's no service client here. It's literally this source file. But, if you're comment is more about warning for real-world use cases I could.

sdk/typespec/typespec_client_core/examples/stream_response.rs
heaths: We're dumping this exact source file, so no. The real tracing implementation will not dump PII.

PR #1870, opened 2024-10-24
Add event-processor.config

hallipr:


JimSuplizio (commented):
I was going to approve this but it merged before I could. I'm approving this after the fact!

PR #1880, opened 2024-10-29
[Cosmos] Remove options builders and (most) `impl` parameters

analogrelay:
This realizes some recent changes to guidelines regarding options and service method parameters. Specifically:

* Options types should now be PODs (plain old data) with public fields and avoid using builders unless absolutely necessary.
* Avoid the use of `impl Into` and `impl AsRef` parameters

The options changes were easy and deleted a bunch of unnecessary boilerplate, my favorite! Avoiding the use of `impl`s led to the following exceptions (which I mentioned when we were chatting about the guidelines earlier but will detail again here):

* `impl Into<Query>` is used for APIs that accept a query. Cosmos queries are made up of a query text, and a parameter. Other SDKs, such as the Go SDK, encode the parameters in the options, but I don't feel like that's an accurate modelling. Query parameters appear in the body, and having to merge a parameter with an option to create the body doesn't feel right. We want to support calling `container_client.query_items("SELECT * FROM c", ...)` **and** `container_client.query_items(Query::from("SELECT * FROM c WHERE c.id = @id").with_parameter("@id", id), ...)`, so we use `impl Into<Query>` here, and implement `From<&'static str>` and `From<String>` on `Query`.

* `impl Into<PartitionKey>` is used on APIs for item CRUD, which require a partition key be specified. Cosmos allows multiple types as partition keys (strings, numbers, booleans) and allows up to 3 levels of hierarchical keys. We want to be able to call `container_client.read_item(item_id, "partition1", ...)`, and `container_client.read_item(item_id, ("parent_partition", 42, "grandchild_partition"), ...)`. So, we accept `impl Into<PartitionKey>` and have several implementations (in `sdk/data/cosmos/azure_data_cosmos/partition_key.rs`) of `From` for `PartitionKey`

* Finally, `impl Into<QueryPartitionStrategy>` is used on APIs that query items. Cosmos supports single-partition queries, where the partition key **must** be specified independent of the query, and cross-partition queries (where no partition key is specified). The Cosmos Rust SDK **does not support cross-partition queries at this time** but may add that support later. So, we use `QueryPartitionStrategy` as an enum that currently only has a single variant (`QueryPartitionStrategy::SinglePartition(PartitionKey)`) and have `impl<T: Into<PartitionKey>> From<T> for QueryPartitionStrategy` which allows the API to accept the same arguments as the CRUD APIs, **and** a future `QueryPartitionStrategy::CrossPartition` marker to indicate a cross-partition query.

I left all those cases above alone, though we can absolutely continue discussions there.

cc @heaths @JeffreyRichter @RickWinter (as FYI for the exceptions pointed out above, though I'm sure we'll do some more formal API reviewing later)


sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -77,28 +73,27 @@ impl CosmosClient {
     /// ```
     #[cfg(feature = "key_auth")]
     pub fn with_key(
-        endpoint: impl AsRef<str>,
-        key: impl Into<Secret>,
+        endpoint: &str,
+        key: Secret,
```
analogrelay, line 77: This one hurt a teeny tiny bit. It means you now have to call `.into()` manually when passing a key. For example:

```rust
CosmosClient::with_key(&args.endpoint, key.into(), None)
```

I'm inclined to be fine with that, particularly since we want to discourage key auth (the API isn't even present unless you enable the `key_auth` feature in the Cosmos crate)

heaths: Making it "harder" seems appropriate in this case.

sdk/cosmos/azure_data_cosmos/examples/cosmos/main.rs
```diff
@@ -74,11 +74,11 @@ pub async fn main() -> Result<(), Box<dyn Error>> {
     }
 }
 
-fn create_client(args: &SharedArgs) -> Result<CosmosClient, Box<dyn Error>> {
-    if let Some(key) = args.key.as_ref() {
+fn create_client(args: &mut SharedArgs) -> Result<CosmosClient, Box<dyn Error>> {
+    if let Some(key) = args.key.take() {
```
heaths, line 78: Couldn't this make the API harder to use? Is there any reason to `take()`?

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -1,44 +1,50 @@
...
-    pub use super::read_database_options::ReadDatabaseOptionsBuilder;
+use azure_core::ClientOptions;
+
+/// Options used when creating a [`CosmosClient`](crate::CosmosClient).
+#[derive(Clone, Debug, Default)]
+pub struct CosmosClientOptions {
+    pub client_options: ClientOptions,
 }
+
+/// Options to be passed to [`DatabaseClient::create_container()`](crate::clients::DatabaseClient::create_container()).
```
heaths, line 12: Method options should have a `method_options: typespec_client_core::http::ClientMethodOptions` member.

heaths (approved):
A change or two I'd like to see, but otherwise nothing blocking.

sdk/cosmos/azure_data_cosmos/examples/cosmos/main.rs
analogrelay: This is the only time the sample needs to read this argument, but see what you're saying. There's not really a specific _reason_ other than I kinda wanted to consume the key here. I can swap it back.

analogrelay: Oh, I remember. There's no `impl From<&String> for Secret`. So I have to clone, which feels kinda unnecessary. But I'm optimizing a sample that's _really_ just designed for testing, so I should probably stop worrying about it 😅

analogrelay: (To be fair, I was cloning before too)

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
analogrelay: I'm going to take that in a second PR, because I don't currently _use_ the `Context` that provides, and I want to make sure we do that when I add it. I'll start on that right now though.

analogrelay: Actually, I'll add it on to this PR, it won't take much. Going to flip this back to draft until that's done.

analogrelay:
Drafting while I add `method_options` to our options

Pilchie (commented):
I know it's draft right now, but this looks fine to me in it's current state (in case I don't have a chance to re-review before going on vacation).

analogrelay:
Re-un-drafting. I've added a `method_options: azure_core::ClientMethodOptions` field to every method options type. Since that type carries an optional ``Context`, I've also changed all our methods to use that `Context` as the base context we add our `ResourceLink` value to.

sdk/typespec/typespec_client_core/src/http/options/mod.rs
```diff
@@ -125,7 +125,7 @@ impl AsClientOptions for ClientOptions {
 /// Method options allow customization of client method calls.
 #[derive(Clone, Debug, Default)]
 pub struct ClientMethodOptions<'a> {
-    pub(crate) context: Context<'a>,
+    pub context: Context<'a>,
```
analogrelay, line 128: I didn't remove `AsClientMethodOptions`, but I did go ahead with making this properly `pub`, to align with our new guidance on options. I haven't gone through all the options and made changes, this one was required for us to use `ClientMethodOptions` in our Cosmos-specific method options.

sdk/typespec/typespec_client_core/src/http/context.rs
```diff
@@ -104,6 +104,20 @@ impl<'a> Context<'a> {
     pub fn len(&self) -> usize {
         self.type_map.len()
     }
+
+    /// Transforms this [`Context`] into a new [`Context`] that owns the underlying data, cloning it if necessary.
+    ///
+    /// If this [`Context`] already owns the underlying data, that data is moved into the new [`Context`] as-is.
+    /// If this [`Context`] is borrowing it's underlying data, that data is cloned and returned as a new [`Context`].
+    pub fn into_owned(self) -> Context<'static> {
```
analogrelay, line 112: When you have a `Context<'a>` and want to use it inside a `Pager` (which is `'static`), you need a way to convert that `Context` into a `Context<'static>` (unless we decide to add a lifetime parameter to `Pager`). This method performs that conversion.

Pilchie (approved):

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -60,15 +60,16 @@ impl ContainerClient {
...
-        // REASON: This is a documented public API so prefixing with '_' is undesirable.
-        options: Option<ReadContainerOptions>,
+        options: Option<ReadContainerOptions<'_>>,
     ) -> azure_core::Result<Response<ContainerProperties>> {
         let url = self.pipeline.url(&self.link);
         let mut req = Request::new(url, Method::Get);
         self.pipeline
-            .send(Context::new(), &mut req, self.link.clone())
+            .send(
+                options.map(|o| o.method_options.context),
```
heaths, line 69: This will work for now, but once we figure out the final tracing story, you may have to change it to something like https://gist.github.com/heaths/3aea54ccac2e404d22e110b2400e9200

sdk/typespec/typespec_client_core/src/http/options/mod.rs
heaths: I'll do a clean up soon of all this. Want to make sure stakeholders are good with the guidelines changes.

heaths (approved):

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
analogrelay: Yeah. I was also considering using `impl AsClientMethodOptions` in the `self.pipeline.send` helper (which is internal, so I don't mind using an `impl` type there) to reduce boilerplate, but I decided to wait on that and see what we need later. This is all internal code, so refactoring is straightforward.

sdk/typespec/typespec_client_core/src/http/options/mod.rs
analogrelay: Exactly what I was thinking :)

PR #1886, opened 2024-10-31
Do not log PII in Debug implementations

heaths:
Related to #1707 but does not resolve it. Still need a solution to redact headers as needed.

### DIFF
--- a/sdk/typespec/typespec_client_core/src/fmt.rs
+++ b/sdk/typespec/typespec_client_core/src/fmt.rs
@@ -1,0 +1,55 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+//! Formatting helpers.
+
+use std::borrow::Cow;
+
+/// Converts ASCII characters in `value` to lowercase if required; otherwise, returns the original slice.
+///
+/// # Examples
+///
+/// Returns the original slice:
+///
+/// ```
+/// # use std::borrow::Cow;
+/// # use typespec_client_core::fmt::to_ascii_lowercase;
+/// let actual = to_ascii_lowercase("hello, world!");
+/// assert!(matches!(actual, Cow::Borrowed("hello, world!")));
+/// ```
+///
+/// Returns a clone converted to lowercase ASCII character.
+///
+/// ```
+/// # use std::borrow::Cow;
+/// # use typespec_client_core::fmt::to_ascii_lowercase;
+/// let actual = to_ascii_lowercase("hello, World!");
+/// assert!(matches!(
+///     actual,
+///     Cow::Owned(expected) if expected == "hello, world!"
+/// ));
+/// ```
+pub fn to_ascii_lowercase<'a>(value: &'a str) -> Cow<'a, str> {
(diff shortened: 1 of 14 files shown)


sdk/typespec/typespec_client_core/src/fmt.rs
```diff
@@ -0,0 +1,55 @@
...
+/// ```
+pub fn to_ascii_lowercase(value: &str) -> Cow<'_, str> {
+    for (i, c) in value.chars().enumerate() {
+        if c.is_ascii_uppercase() {
+            let mut s = value.to_owned();
+            s[i..].make_ascii_lowercase();
+
+            return Cow::Owned(s);
+        }
+    }
```
analogrelay, line 40: This feels like a case where the savings over just calling `to_lowercase()` isn't much, given that you have to scan the string anyway, but I'm fine with it if there's a reason for it.

sdk/typespec/typespec_client_core/src/http/response.rs
```diff
@@ -214,9 +214,8 @@ impl<T> fmt::Debug for Response<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("Response")
             .field("status", &self.status)
-            .field("headers", &self.headers)
-            .field("body", &"(body)")
```
analogrelay, line 218: You could keep the `body` field, right? It's already not actually displaying the body.

analogrelay (commented):
Nothing blocking, just some small thoughts

sdk/typespec/typespec_client_core/src/fmt.rs
heaths: Figured no reason to rescan the whole string, or did you mean the purpose for this call anyway? That's fair. My plan was to use it (and I realized I'm not yet, so maybe I should just take it out for now) to normalize header names as I compare against a safe list. My main concern wasn't speed, though, it was allocations. Because we require - at build time - all headers to be lowercase and are encouraging Azure services to make them lowercase, my hope is that most headers we'd scan (think sanitization) are already lowercase and there'd be no need to allocate a new `String`.

/cc @JeffreyRichter

heaths: Also, that's why this is ASCII-specific: HTTP headers are limited to ASCII. Any non-ASCII character we'll leave as-is and let code elsewhere err or not as appropriate. But since HTTP headers are ASCII, that shouldn't be a practical concern.

sdk/typespec/typespec_client_core/src/http/response.rs
heaths: Yes, but I felt "(body)" isn't any more useful so the `finish_non_exhaustive()` that "prints" the string ".." seemed appropriate.

heaths: I'll probably do the same with `Display` as well, since we need to implement `Display` to satisfy the `Error` trait constraint.

LarryOsterman (approved):

PR #1892, opened 2024-11-04
Removed builders from eventhubs; removed impl Into<String> from eventhubs and AMQP; Fixed time conversion issue with 1-1-0001

LarryOsterman:
A lot of churn, basically boiling down to 4 changes, related to recent API guideline changes:

1. Removed most builders from Eventhubs API surface (EventData builder is still there because EventData is a fairly complicated structure).
2. Removed most impl Into<String> references from both EventHubs and AMQP.
3. Fixed a time conversion with the 1-1-0001 sentinel value EventHubs uses to represent times that aren't available yet.
4. Added the ability to specify a ServiceBus token to the AMQP layer to allow C++ to continue to use Connection Strings.

## CoPilot summary
This pull request includes several changes to the `sdk/core/azure_core_amqp` module, focusing on improving type consistency and handling optional values. The most important changes involve updating method signatures to use concrete types instead of generics and adding support for an optional token type in authorization methods.

### Type Consistency Improvements:

* Updated method signatures in `AmqpClaimsBasedSecurityApis`, `AmqpConnectionApis`, `AmqpManagementApis`, and `AmqpSenderApis` to use concrete `String` types instead of generics like `impl Into<String>`. (`sdk/core/azure_core_amqp/src/cbs.rs`, `sdk/core/azure_core_amqp/src/connection.rs`, `sdk/core/azure_core_amqp/src/fe2o3/cbs.rs`, `sdk/core/azure_core_amqp/src/fe2o3/connection.rs`, `sdk/core/azure_core_amqp/src/fe2o3/management.rs`, `sdk/core/azure_core_amqp/src/fe2o3/sender.rs`, `sdk/core/azure_core_amqp/src/management.rs`) [[1]](diffhunk://#diff-1ece8d44dc0336f6c37f7fbf8db9cf06ed4fde257cd679c7b9c007bc80c5c03aL52-R55) [[2]](diffhunk://#diff-3b1cde9b879b09dbdc03797027041992fe34df702acdf708bed1df18f358c695L65-R65) [[3]](diffhunk://#diff-3b1cde9b879b09dbdc03797027041992fe34df702acdf708bed1df18f358c695L86-R86) [[4]](diffhunk://#diff-c569aa674a24402ed7f7942e3678f39323f51e9959097893621d26da589be33bL70-R72) [[5]](diffhunk://#diff-c569aa674a24402ed7f7942e3678f39323f51e9959097893621d26da589be33bL80-R85) [[6]](diffhunk://#diff-c569aa674a24402ed7f7942e3678f39323f51e9959097893621d26da589be33bL106-R110) [[7]](diffhunk://#diff-3df1e5d2a77c38245c74ae9a38af5bcb7744b99d894b2e6936df1a4a6cc5905dL42-R42) [[8]](diffhunk://#diff-f3f68245298c5cd5e5d0ce7ddd3d484777acb642e1f48c230b72868b7f76d4faL43-R50) [[9]](diffhunk://#diff-f3f68245298c5cd5e5d0ce7ddd3d484777acb642e1f48c230b72868b7f76d4faL88-R87) [[10]](diffhunk://#diff-f3f68245298c5cd5e5d0ce7ddd3d484777acb642e1f48c230b72868b7f76d4faL125-R129) [[11]](diffhunk://#diff-76ff607b3937fe075233653a8f2f57295185066fff238ea7ad5832931b2ba9fcL27-R27) [[12]](diffhunk://#diff-76ff607b3937fe075233653a8f2f57295185066fff238ea7ad5832931b2ba9fcL66-R66) [[13]](diffhunk://#diff-d2e024aff1c9a3e898bca74e3fb57a3a90b894762c3de406099ce9029807f04aL25-R25) [[14]](diffhunk://#diff-d2e024aff1c9a3e898bca74e3fb57a3a90b894762c3de406099ce9029807f04aL44-R44)

### Optional Token Type Support:

* Added an optional `token_type` parameter to the `authorize_path` method in `AmqpClaimsBasedSecurityApis` to allow specifying the type of token used for authorization. If not supplied, "jwt" is assumed. (`sdk/core/azure_core_amqp/src/cbs.rs`, `sdk/core/azure_core_amqp/src/fe2o3/cbs.rs`) [[1]](diffhunk://#diff-1ece8d44dc0336f6c37f7fbf8db9cf06ed4fde257cd679c7b9c007bc80c5c03aR41) [[2]](diffhunk://#diff-c569aa674a24402ed7f7942e3678f39323f51e9959097893621d26da589be33bL80-R85)

### Code Simplification:

* Simplified the conversion logic in `fe2o3_amqp_types` by using direct `String` values and handling optional values more explicitly. (`sdk/core/azure_core_amqp/src/fe2o3/messaging/message_fields.rs`, `sdk/core/azure_core_amqp/src/fe2o3/messaging/message_source.rs`, `sdk/core/azure_core_amqp/src/fe2o3/value.rs`) [[1]](diffhunk://#diff-047d0b47f04cb197a02578cb0226d9eaddc10d50ee925a16d1a80a51301e7badL130-R130) [[2]](diffhunk://#diff-047d0b47f04cb197a02578cb0226d9eaddc10d50ee925a16d1a80a51301e7badL281-R281) [[3]](diffhunk://#diff-047d0b47f04cb197a02578cb0226d9eaddc10d50ee925a16d1a80a51301e7badL487-R493) [[4]](diffhunk://#diff-11ebc6b62e95377caef2bfe00db743fb8c09b39a39c49d235eabf8ae7c6503d2L225-R225) [[5]](diffhunk://#diff-4675476a7f4c4b0648db55fbbb9d4af283147cc178b97537b12e059dc6076220R39-R63) [[6]](diffhunk://#diff-4675476a7f4c4b0648db55fbbb9d4af283147cc178b97537b12e059dc6076220L242-R252) [[7]](diffhunk://#diff-4675476a7f4c4b0648db55fbbb9d4af283147cc178b97537b12e059dc6076220L332-R352) [[8]](diffhunk://#diff-4675476a7f4c4b0648db55fbbb9d4af283147cc178b97537b12e059dc6076220L389-R407) [[9]](diffhunk://#diff-4675476a7f4c4b0648db55fbbb9d4af283147cc178b97537b12e059dc6076220L420-R430)

### Comment Updates:

* Updated comments to reflect the changes, including the addition of `sastoken` to the list of spell-checked words. (`sdk/core/azure_core_amqp/src/cbs.rs`)

These changes collectively enhance the codebase by making it more consistent and easier to understand, while also adding flexibility for handling different token types in authorization scenarios.

### DIFF
--- a/sdk/core/azure_core_amqp/src/connection.rs
+++ b/sdk/core/azure_core_amqp/src/connection.rs
@@ -69,7 +69,7 @@ pub trait AmqpConnectionApis {
     fn close(&self) -> impl std::future::Future<Output = Result<()>>;
     fn close_with_error(
         &self,
-        condition: impl Into<AmqpSymbol>,
+        condition: AmqpSymbol,
         description: Option<String>,
         info: Option<AmqpOrderedMap<AmqpSymbol, AmqpValue>>,
     ) -> impl std::future::Future<Output = Result<()>>;
@@ -94,7 +94,7 @@ impl AmqpConnectionApis for AmqpConnection {
     }
     fn close_with_error(
         &self,
-        condition: impl Into<AmqpSymbol>,
+        condition: AmqpSymbol,
         description: Option<String>,
         info: Option<AmqpOrderedMap<AmqpSymbol, AmqpValue>>,
     ) -> impl std::future::Future<Output = Result<()>> {
--- a/sdk/core/azure_core_amqp/src/connection.rs
+++ b/sdk/core/azure_core_amqp/src/connection.rs
@@ -30,11 +30,11 @@ impl AmqpConnectionOptions {
     pub fn builder() -> builders::AmqpConnectionOptionsBuilder {
         builders::AmqpConnectionOptionsBuilder::new()
     }
-    pub fn max_frame_size(&self) -> Option<&u32> {
-        self.max_frame_size.as_ref()
+    pub fn max_frame_size(&self) -> Option<u32> {
+        self.max_frame_size
     }
-    pub fn channel_max(&self) -> Option<&u16> {
-        self.channel_max.as_ref()
+    pub fn channel_max(&self) -> Option<u16> {
+        self.channel_max
     }
     pub fn idle_timeout(&self) -> Option<&Duration> {
         self.idle_timeout.as_ref()
@@ -54,8 +54,8 @@ impl AmqpConnectionOptions {
     pub fn properties(&self) -> Option<&AmqpOrderedMap<AmqpSymbol, AmqpValue>> {
         self.properties.as_ref()
     }
-    pub fn buffer_size(&self) -> Option<&usize> {
-        self.buffer_size.as_ref()
+    pub fn buffer_size(&self) -> Option<usize> {
+        self.buffer_size
     }
 }
 
--- a/sdk/core/azure_core_amqp/src/connection.rs
+++ b/sdk/core/azure_core_amqp/src/connection.rs
@@ -75,7 +75,7 @@ pub trait AmqpConnectionApis {
     ) -> impl std::future::Future<Output = Result<()>>;
 }
 
-#[derive(Debug, Default)]
+#[derive(Default)]
 pub struct AmqpConnection {
     pub(crate) implementation: ConnectionImplementation,
 }
--- a/sdk/core/azure_core_amqp/src/connection.rs
+++ b/sdk/core/azure_core_amqp/src/connection.rs
@@ -62,7 +62,7 @@ impl AmqpConnectionOptions {
 pub trait AmqpConnectionApis {
     fn open(
         &self,
-        name: impl Into<String>,
+        name: String,
         url: Url,
         options: Option<AmqpConnectionOptions>,
     ) -> impl std::future::Future<Output = Result<()>>;
@@ -83,7 +83,7 @@ pub struct AmqpConnection {
 impl AmqpConnectionApis for AmqpConnection {
     fn open(
         &self,
-        name: impl Into<String>,
+        name: String,
         url: Url,
         options: Option<AmqpConnectionOptions>,
     ) -> impl std::future::Future<Output = Result<()>> {
--- a/sdk/core/azure_core_amqp/src/connection.rs
+++ b/sdk/core/azure_core_amqp/src/connection.rs
@@ -30,32 +30,32 @@ impl AmqpConnectionOptions {
     pub fn builder() -> builders::AmqpConnectionOptionsBuilder {
         builders::AmqpConnectionOptionsBuilder::new()
     }
-    pub fn max_frame_size(&self) -> Option<u32> {
-        self.max_frame_size
+    pub fn max_frame_size(&self) -> Option<&u32> {
+        self.max_frame_size.as_ref()
     }
-    pub fn channel_max(&self) -> Option<u16> {
-        self.channel_max
+    pub fn channel_max(&self) -> Option<&u16> {
+        self.channel_max.as_ref()
     }
-    pub fn idle_timeout(&self) -> Option<Duration> {
-        self.idle_timeout
+    pub fn idle_timeout(&self) -> Option<&Duration> {
+        self.idle_timeout.as_ref()
     }
-    pub fn outgoing_locales(&self) -> Option<Vec<String>> {
-        self.outgoing_locales.clone()
+    pub fn outgoing_locales(&self) -> Option<&Vec<String>> {
+        self.outgoing_locales.as_ref()
     }
-    pub fn incoming_locales(&self) -> Option<Vec<String>> {
-        self.incoming_locales.clone()
+    pub fn incoming_locales(&self) -> Option<&Vec<String>> {
+        self.incoming_locales.as_ref()
     }
-    pub fn offered_capabilities(&self) -> Option<Vec<AmqpSymbol>> {
-        self.offered_capabilities.clone()
+    pub fn offered_capabilities(&self) -> Option<&Vec<AmqpSymbol>> {
+        self.offered_capabilities.as_ref()
     }
-    pub fn desired_capabilities(&self) -> Option<Vec<AmqpSymbol>> {
-        self.desired_capabilities.clone()
+    pub fn desired_capabilities(&self) -> Option<&Vec<AmqpSymbol>> {
+        self.desired_capabilities.as_ref()
     }
-    pub fn properties(&self) -> Option<AmqpOrderedMap<AmqpSymbol, AmqpValue>> {
-        self.properties.clone()
+    pub fn properties(&self) -> Option<&AmqpOrderedMap<AmqpSymbol, AmqpValue>> {
+        self.properties.as_ref()
     }
-    pub fn buffer_size(&self) -> Option<usize> {
-        self.buffer_size
+    pub fn buffer_size(&self) -> Option<&usize> {
+        self.buffer_size.as_ref()
     }
 }
 
@@ -156,8 +156,8 @@ pub mod builders {
         }
         pub fn with_properties<K, V>(mut self, properties: impl Into<AmqpOrderedMap<K, V>>) -> Self
         where
-            K: Into<AmqpSymbol> + Debug + Default + PartialEq,
-            V: Into<AmqpValue> + Debug + Default,
+            K: Into<AmqpSymbol> + Debug + Clone + PartialEq,
+            V: Into<AmqpValue> + Debug + Clone,
         {
             let properties_map: AmqpOrderedMap<K, V> = properties.into();
             let properties_map = properties_map
--- a/sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
+++ b/sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
@@ -67,8 +67,9 @@ impl AmqpClaimsBasedSecurityApis for Fe2o3ClaimsBasedSecurity<'_> {
 
     async fn authorize_path(
         &self,
-        path: impl Into<String> + Debug,
-        secret: impl Into<String>,
+        path: String,
+        token_type: Option<String>,
+        secret: String,
         expires_at: time::OffsetDateTime,
     ) -> Result<()> {
         trace!(
@@ -77,8 +78,11 @@ impl AmqpClaimsBasedSecurityApis for Fe2o3ClaimsBasedSecurity<'_> {
             expires_at
         );
         let cbs_token = CbsToken::new(
-            secret.into(),
-            "jwt",
+            secret,
+            match token_type {
+                Some(token_type) => token_type,
+                None => "jwt".to_string(),
+            },
             Some(Timestamp::from(
                 expires_at
                     .to_offset(time::UtcOffset::UTC)
@@ -103,7 +107,7 @@ impl AmqpClaimsBasedSecurityApis for Fe2o3ClaimsBasedSecurity<'_> {
             .lock()
             .await
             .borrow_mut()
-            .put_token(path.into(), cbs_token)
+            .put_token(path, cbs_token)
             .await
             .map_err(AmqpManagement::from)?;
         Ok(())
--- a/sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
+++ b/sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
@@ -9,10 +9,9 @@ use azure_core::error::Result;
 use fe2o3_amqp_cbs::token::CbsToken;
 use fe2o3_amqp_types::primitives::Timestamp;
 use std::borrow::BorrowMut;
-use std::{fmt::Debug, sync::OnceLock};
+use std::sync::OnceLock;
 use tracing::{debug, trace};
 
-#[derive(Debug)]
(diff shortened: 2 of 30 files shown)


sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
```diff
@@ -77,8 +78,11 @@ impl<'a> AmqpClaimsBasedSecurityApis for Fe2o3ClaimsBasedSecurity<'a> {
             expires_at
         );
         let cbs_token = CbsToken::new(
-            secret.into(),
-            "jwt",
+            secret,
+            match token_type {
+                Some(token_type) => token_type,
+                None => "jwt".to_string(),
+            },
```
heaths, line 85:
```suggestion
            token_type.unwrap_or("jwt".to_string()),
```

I'll only comment on this one for brevity. Please fix other instances.

sdk/core/azure_core_amqp/src/fe2o3/value.rs
```diff
@@ -36,23 +36,31 @@ impl From<AmqpValue> for fe2o3_amqp_types::primitives::Symbol {
...
+// Number of milliseconds between the Unix epoch (1/1/1970) and year 1 CE.
+// This is the lowest value that can be represented by an AMQP timestamp.
+const CE_ZERO_MILLISECONDS: i64 = -62_135_596_800_000;
+
 impl From<fe2o3_amqp_types::primitives::Timestamp> for AmqpTimestamp {
     fn from(timestamp: fe2o3_amqp_types::primitives::Timestamp) -> Self {
         AmqpTimestamp(
-            std::time::UNIX_EPOCH
-                + std::time::Duration::from_millis(timestamp.milliseconds() as u64),
+            std::time::UNIX_EPOCH.checked_add(std::time::Duration::from_millis(
```
heaths, line 46: I believe this will fail for `wasm32-unknown-unknown` so we'll need to condition it out or, more likely, fail with a good error message.

sdk/core/azure_core_amqp/src/value.rs
```diff
@@ -76,12 +76,17 @@ impl From<Vec<AmqpValue>> for AmqpList {
     }
 }
 
+/// AMQP Timestamp. Number of milliseconds since UNIX_EPOCH.
```
heaths, line 79: Might want to clarify what the "UNIX epoch" instead of just mentioning its symbol (and not even in backticks.

heaths: And as in some of the text below, rustdoc renders markdown so I'd reference symbols / expressions in backticks. Reads cleaner.

sdk/eventhubs/azure_messaging_eventhubs/README.md
```diff
@@ -207,23 +208,24 @@ async fn receive_events() {
...
-                azure_messaging_eventhubs::consumer::ReceiveOptions::builder()
-                    .with_start_position(azure_messaging_eventhubs::consumer::StartPosition::builder().with_earliest_location().build())
-                    .build(),
-            ),
-        )
+                azure_messaging_eventhubs::consumer::ReceiveOptions{
+                    start_position: Some(azure_messaging_eventhubs::consumer::StartPosition{
+                        location: azure_messaging_eventhubs::consumer::StartLocation::Earliest,
+                        ..Default::default()}),
+                    ..Default::default()},
```
heaths, line 227: Nit: unless the formatter did this, I would put the struct termination on separate lines e.g.,
```suggestion
                        ..Default::default()
                    }),
                    ..Default::default()
                },
```

Must cleaner to read. The "trailing commas" threw me for a sec since you can do that with the "and the rest" operators until I noticed the `}` in there.

sdk/eventhubs/azure_messaging_eventhubs/src/common/mod.rs
```diff
@@ -78,63 +94,63 @@ impl ManagementInstance {
...
         let response = self
             .management
-            .call("com.microsoft:partition", application_properties)
+            .call(PARTITION_ENTITY_TYPE.to_string(), application_properties)
             .await?;
 
         // Look for the required response properties
-        if !response.contains_key("type")
-            || !response.contains_key("last_enqueued_sequence_number_epoch")
+        if !response.contains_key(EVENTHUB_PARTITION_PROPERTIES_TYPE.to_string())
```
heaths, line 110: Why does `contains_key` need an owned `String` as opposed to a borrowed `&str`? Same with `get()` below. These seems like the wrong API choice, or do we not own the callee (hard to tell from this PR)?

sdk/eventhubs/azure_messaging_eventhubs/src/consumer/mod.rs
```diff
@@ -389,21 +386,18 @@ impl ConsumerClient {
     /// }
     /// ```
     #[tracing::instrument]
```
heaths, line 388: Let's either remove these for now (we need to finalize the tracing story, and I've gotten some pushback on `tracing`) or change all instances to `#[tracing::instrument(skip_all, err)]`. It could leak secrets otherwise. Also, the client shouldn't implement `Debug` or `Display` - the former of which you'd have to have done to satisfy this tracing attribute as-is.

sdk/eventhubs/azure_messaging_eventhubs/src/producer/mod.rs
```diff
@@ -37,20 +38,20 @@ pub mod batch;
 const DEFAULT_EVENTHUBS_APPLICATION: &str = "DefaultApplicationName";
 
 /// Options used when creating an Event Hubs ProducerClient.
-#[derive(Default)]
+#[derive(Default, Debug, Clone)]
```
heaths, line 41: Don't implement `Debug` on clients. See previous comment.

heaths (changes requested):

sdk/eventhubs/azure_messaging_eventhubs/src/common/mod.rs
LarryOsterman: Can I do `impl Into<&K>`? Or would this be `impl AsRef<Into<K>>`?

sdk/core/azure_core_amqp/src/fe2o3/value.rs
LarryOsterman: FWIW, it seems to work. No idea exactly why though.

sdk/eventhubs/azure_messaging_eventhubs/src/common/mod.rs
LarryOsterman: This is an interesting question. 

Currently `contains_key` is defined as `contains_key(key: K)`, I can change it to `contains__key(key:&K)` but that still won't allow you to pass in a string slice.

Instead this forces you to write:

`&AmqpSymbol::from("key")`

sdk/core/azure_core_amqp/src/fe2o3/value.rs
heaths: Ah, I think I see why: we're not testing against `wasm32-unknown-unknown` in our AzDO tests it seems. Can you try that on the command line directly for now? @hallipr we'll have to make sure we testing against WASM as well.

```bash
cargo build --all-features --workspace --target wasm32-unknown-unknown
```

LarryOsterman: Getting that command to work is a nightmare. Tons of openssl related bugs due to c_uint and c_ulong not being found.

sdk/core/azure_core_amqp/src/connection.rs
```diff
@@ -30,32 +30,32 @@ impl AmqpConnectionOptions {
     pub fn builder() -> builders::AmqpConnectionOptionsBuilder {
         builders::AmqpConnectionOptionsBuilder::new()
     }
-    pub fn max_frame_size(&self) -> Option<u32> {
-        self.max_frame_size
+    pub fn max_frame_size(&self) -> Option<&u32> {
+        self.max_frame_size.as_ref()
```
heaths, line 34: Why all these changes? Typically a reference (pointer) to a primitive `Copy`able is more expensive than just copying it out. It's immutable anyway.

sdk/core/azure_core_amqp/src/session.rs
```diff
@@ -144,8 +144,8 @@ pub mod builders {
         }
         pub fn with_properties<K, V>(mut self, properties: impl Into<AmqpOrderedMap<K, V>>) -> Self
         where
-            K: Into<AmqpSymbol> + PartialEq + Default + Debug,
```
heaths, line 147: Why is `Debug` actually needed here? Typically only include constraints needed in the impl. Seems nothing should require a `Debug` impl, or does collecting into an `AmqpOrderedMap` actually?

sdk/eventhubs/azure_messaging_eventhubs/src/common/mod.rs
```diff
@@ -53,17 +52,17 @@ impl ManagementInstance {
             .call(EVENTHUB_ENTITY_TYPE.to_string(), application_properties)
             .await?;
 
-        if !response.contains_key(EVENTHUB_PROPERTY_PARTITION_COUNT.to_string()) {
+        if !response.contains_key(&EVENTHUB_PROPERTY_PARTITION_COUNT.to_string()) {
```
heaths, line 55: Aren't these already defined as `&'static str`? Why allocate a string just to take an address? Just pass the symbol:

```suggestion
        if !response.contains_key(EVENTHUB_PROPERTY_PARTITION_COUNT) {
```

Clippy should've warned you about this. Weird.

heaths (changes requested):
Better - almost there - but a couple outstanding questions and requested changes.

sdk/eventhubs/azure_messaging_eventhubs/src/common/mod.rs
LarryOsterman: Oh - last night in my sleep, I found what I believe is the perfect answer. 

```rust
   pub fn contains_key(&self, key: impl PartialEq<K>) -> bool {
```

The negative is that it uses an `impl Trait` for the parameter, the positive is that it works for ANY type which implements PartialEq, which means it works with `String`, `&str`, `AmqpSymbol`, whatever.

sdk/core/azure_core_amqp/src/connection.rs
LarryOsterman: I removed the references for all the POD types, and left them in for the complex types.

sdk/eventhubs/azure_messaging_eventhubs/src/common/mod.rs
LarryOsterman: This is weird. On my view of the source, line 55/56 doesn't contain the & and the to_string - it's just your suggested change.

sdk/core/azure_core_amqp/src/session.rs
LarryOsterman: It's not, removed.

sdk/core/azure_core_amqp/src/value.rs
```diff
@@ -18,11 +20,40 @@ use azure_core::Result;
 #[derive(Debug, PartialEq, Clone, Default, Eq)]
 pub struct AmqpSymbol(pub String);
 
+impl PartialEq<str> for AmqpSymbol {
```
heaths, line 23: Gripe: I really wish all this was symmetric.

Technically, if you control the order in which comparisons are done (rather, the lvalue and rvalue), you don't need all these; however, if the goal is to be flexible for callers this works.

sdk/core/azure_core_amqp/src/value.rs
```diff
@@ -267,10 +304,14 @@ where
...
-    pub fn get(&self, key: &K) -> Option<&V> {
+    pub fn get<Q>(&self, key: &Q) -> Option<&V>
+    where
+        K: Borrow<Q> + PartialEq<Q>,
+        Q: PartialEq<K> + ?Sized,
+    {
         self.inner
             .iter()
-            .find_map(|(k, v)| if k == key { Some(v) } else { None })
+            .find_map(|(k, v)| if key == k { Some(v) } else { None })
```
heaths, line 314: Ah, this may be why you had to implement both directions. As I [changed in my example](https://gist.github.com/heaths/dd2a8f327b1a6b429827ecc17fa44460/revisions#diff-0544fc952eefa8d170bbd200b75827f5bb58e39b132e7c871c1bc8a9d5e09ce5), if you are careful of the order you only need the direction you use, and can use it both ways.

That said, again though, if the goal is to make it symmetric for other callers, by all means keep it. It just puts more constraints on using your ordered map.

heaths (approved):
LGTM. Just some thoughts about symmetrically implementing `PartialEq` you could probably avoid.

sdk/core/azure_core_amqp/src/value.rs
heaths: Actually, even if you keep the symmetrical implementations, I would still advise you to maintain a single order - maybe even call `eq` directly just to avoid ambiguity for future maintainers - so that both implementations aren't compiled into a binary. You know, shave a little off the top.

LarryOsterman: I like the .eq idea, as long as VS Code doesn't decide to change it to == for me :(.

PR #1894, opened 2024-11-04
retry_policy: verbose logging for io errors

thewisenerd:
may contain information necessary for triaging the issue.

spent waay too much time debugging an issue where ca-certificates was not installed, here's an example of how that error looks.

before:
```
[2024-11-04T21:17:40Z DEBUG azure_core::policies::retry_policies::retry_policy] io error occurred when making request which will be retried: failed to execute `reqwest` request
```

after:
```
[2024-11-04T20:52:39Z DEBUG azure_core::policies::retry_policies::retry_policy] io error occurred when making request which will be retried: Error { context: Full(Custom { kind: Io, error: reqwest::Error { kind: Request, url: "REDACTED", source: hyper_util::client::legacy::Error(Connect, Ssl(Error { code: ErrorCode(1), cause: Some(Ssl(ErrorStack([Error { code: 167772294, library: "SSL routines", function: "tls_post_process_server_certificate", reason: "certificate verify failed", file: "../ssl/statem/statem_clnt.c", line: 1889 }]))) }, X509VerifyResult { code: 20, error: "unable to get local issuer certificate" })) } }, "failed to execute `reqwest` request") }
```

### DIFF
--- a/sdk/core/src/policies/retry_policies/retry_policy.rs
+++ b/sdk/core/src/policies/retry_policies/retry_policy.rs
@@ -174,7 +174,7 @@ where
                 Err(error) => {
                     if error.kind() == &ErrorKind::Io {
                         debug!(
-                            "io error occurred when making request which will be retried: {}",
+                            "io error occurred when making request which will be retried: {:?}",
                             error
                         );
                         // IO error so no Retry-After headers - leave the retry period up to the policy


heaths (approved):
Given this is limited to an IO error, it seems unlikely PII would be leaked any more than it could be now for other error types e.g., our own `HttpError`.

I will merge this, but please keep in mind that we have no plans to release any new versions of this unsupported library. We are working on supported versions in `main` so see `main`'s `README.md` for details.

See <https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html> for options to override crates.io to pull down this version.

PR #1903, opened 2024-11-07
Test wasm32-unknown-unknown in Azure Pipelines

heaths:
Also gets rid of most of GitHub Actions with only E2E tests remaining.


heaths:
@LarryOsterman can we even run E2E tests against Event Hubs? That requires actually-provisioned resources, right? If not, I can delete the entirety of the GitHub Actions workflows finally. /cc @hallipr

heaths:

LarryOsterman:
> @LarryOsterman can we even run E2E tests against Event Hubs? That requires actually-provisioned resources, right? If not, I can delete the entirety of the GitHub Actions workflows finally. /cc @hallipr

No, the test_e2e tests for Event Hubs requires either the eventhub emulator running locally or actually provisioned resources. It's a live test in the classic vernacular. 

As we get further into ADO pipelines, I'll probably bring up the test AMQP Broker (used in Go) to test the AMQP stuff locally, that will help with coverage.

LarryOsterman (approved):

eng/scripts/Test-Packages.ps1
```diff
@@ -30,22 +31,24 @@ if ($PackageInfoPath) {
   | ConvertFrom-Json
 }
 else {
-  $packagesToTest = Get-AllPackagesInRepo
+  . "$PSScriptRoot\Language-Settings.ps1"
```
hallipr, line 34: Line 14's invocation of eng/common/scripts/common.ps1 will import Language-Settings.ps1

eng/scripts/Test-Packages.ps1
```diff
@@ -30,22 +31,24 @@ if ($PackageInfoPath) {
...
+  Write-Host "  '$($package.Name)' in '$($package.DirectoryPath)'"
 }
 
 Write-Host "Setting RUSTFLAGS to '-Dwarnings'"
 $env:RUSTFLAGS = "-Dwarnings"
 
 
 foreach ($package in $packagesToTest) {
-  Push-Location (Join-Path $RepoRoot $package.DirectoryPath)
+  Push-Location ([System.IO.Path]::Combine($RepoRoot, $package.DirectoryPath))
```
hallipr, line 49: Were you seeing a difference between Path::Combine and Join-Path here?

hallipr: Nevermind.  I saw your note re: combining absolute paths:
```
> Join-Path  'c:\something\else' 'd:\another\path'
c:\something\else\d:\another\path

> [IO.Path]::Combine('c:\something\else', 'd:\another\path')
d:\another\path
```

heaths: Yeah. Never use `Join-Path`. It's utter garbage and gets people into trouble so many times. It's worse on non-Windows where you don't get a drive letter in the middle of two absolute paths - just `/`s.

PR #1912, opened 2024-11-12
[Cosmos] Throughput APIs

analogrelay:
This adds support for throughput-related APIs to the Cosmos SDK for Rust, including:

* Specifying throughput offers when creating databases and containers, via a field on the `CreateDatabaseOptions`/`CreateContainerOptions` type.
* Reading throughput offers for a database or container, via the `::read_throughput` method on `ContainerClient` and `DatabaseClient`
* Replacing throughput on a database or container, via the `::replace_throughput` method on `ContainerClient` and `DatabaseClient`

These APIs are largely modelled on similar ones we have in the Go SDK, but I've tried to make sure they stayed "Rusty" and am happy to deviate from those if things look off.

The Go SDK created a separate "OffersClient" that was internal, to support reading/replacing offers, since they're a separate resource and the logic to read/replace them is similar for both databases and containers. Instead of that, I've added helper functions to our internal `CosmosPipeline` type. I think that approach works well here.

Offers don't quite follow the standard pattern for generating the "resource link" for signing requests when authenticating with a key. Specifically, they require that the link be ["only the _rid of the offer you wish to retrieve"](https://learn.microsoft.com/en-us/rest/api/cosmos-db/get-an-offer). So I've added some special-casing to `ResourceLink` to handle that. Part of the goal of refactoring to use a custom `ResourceLink` type was to allow exactly this kind of special-casing, in a clear way.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
@@ -3,11 +3,12 @@
 
 use crate::{
     constants,
-    models::{ContainerProperties, Item, PatchDocument, QueryResults},
+    models::{ContainerProperties, Item, PatchDocument, QueryResults, ThroughputProperties},
     options::{QueryOptions, ReadContainerOptions},
     pipeline::CosmosPipeline,
     resource_context::{ResourceLink, ResourceType},
     DeleteContainerOptions, ItemOptions, PartitionKey, Query, QueryPartitionStrategy,
+    ThroughputOptions,
 };
 
 use azure_core::{Method, Pager, Request, Response};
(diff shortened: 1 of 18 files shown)


sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -62,14 +63,59 @@ impl ContainerClient {
...
+    /// # Arguments
+    /// * `options` - Optional parameters for the request.
+    pub async fn read_throughput(
+        &self,
+        options: Option<ThroughputOptions<'_>>,
+    ) -> azure_core::Result<Option<Response<ThroughputProperties>>> {
+        let options = options.unwrap_or_default();
+
+        // We need to get the RID for the database.
+        let db = self.read(None).await?.deserialize_body().await?;
```
Pilchie, line 87: This is why we need a container property cache at some point.

Pilchie (approved):

sdk/cosmos/azure_data_cosmos/src/models/throughput_properties.rs
```diff
@@ -0,0 +1,135 @@
...
+    headers::{AsHeaders, HeaderName, HeaderValue},
+    Model,
+};
+use serde::{Deserialize, Serialize};
+
+use crate::{constants, models::SystemProperties};
+
+const OFFER_VERSION_2: &str = "V2";
+
+#[derive(Model, Clone, Default, Debug, Deserialize, Serialize)]
```
heaths, line 16: Models should not derive `Debug` or implement `Display`, in most cases. New in the guidelines. Too easy to leak PII.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -97,4 +100,93 @@ impl CosmosPipeline {
...
+        }
+
+        // We only have to into_owned here in order to call send_query_request below,
+        // since it returns `Pager` which must own it's data.
+        // But in this case, we don't really _need_ the `Pager` to own it's data
+        // because we use it and dispose of it within the body of this method.
+        // If we wanted to optimize this later, we have a few options:
+        // 1. Give Pager a lifetime parameter so it can borrow it's context (proliferates lifetime parameters everywhere though...)
+        // 2. Don't use send_query_request. We expect the offer to be in the first page, so we could just make a regular request
+        // (but what if we get an empty page for some reason? is that something the server could do?)
```
heaths, line 127: Given we discuss this in the issue, you might just link to it for full context instead of duplicating options here.

heaths (changes requested):

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
analogrelay: Yep, we didn't have the issue when I wrote this. I'll update to link.

sdk/cosmos/azure_data_cosmos/src/models/throughput_properties.rs
analogrelay: Yep, I remember that now. I had implemented Debug during testing but am happy to remove it. I should say, we know this is all non-PII data, but I'm also fine with the default of not implementing `Debug` unless we have a specific reason to do so.

analogrelay: I need to go through and change how our "testing" example works before I can do this everywhere, but I'll remove it from throughput_properties here. The example uses `Debug` quite a bit and I'd like to done something more consistent there.

heaths (approved):

heaths:

PR #1914, opened 2024-11-13
[Cosmos] Patch item APIs

analogrelay:
This PR adds the `ContainerClient::patch_item` API, which allows you to specify a `PatchDocument` consisting of a list of `PatchOperations` to perform on the specified item.

Despite using the term "Patch" and similar syntax, it's important to note that Cosmos does NOT support [RFC 6902 JSON Patch](https://www.rfc-editor.org/rfc/rfc6902). It is a custom format with its own `op` values.

I'll add a few other comments inline. There are some API design questions that could use some attention since we don't have a lot of prior art or detailed guidelines yet.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
@@ -1,0 +1,307 @@
+use azure_core::Model;
+use serde::{Deserialize, Serialize};
+
+// Cosmos' patch operations are _similar_ to JSON Patch (RFC 6902) in structure, but have different operations.
+
+/// Represents a patch document for Cosmos DB.
+///
+/// A patch document is made up of a collection of patch operations.
+/// Each operation describes how to modify a property of a document.
+/// See <https://learn.microsoft.com/en-us/azure/cosmos-db/partial-document-update> for more information on patch operations.
+#[derive(Model, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct PatchDocument {
+    pub operations: Vec<PatchOperation>,
+}
+
+impl PatchDocument {
+    /// Creates a new empty patch document.
+    pub fn new() -> Self {
+        PatchDocument { operations: vec![] }
+    }
+
+    /// Adds a new "add" operation to the patch document.
+    ///
+    /// See the [type documentation](PatchDocument) for more information on patch operations.
+    ///
+    /// # Arguments
+    /// * `path` - The path to the property to add.
+    /// * `value` - The value to add.
+    pub fn with_add(
+        mut self,
+        path: impl Into<String>,
+        value: impl Serialize,
+    ) -> Result<Self, serde_json::Error> {
+        self.operations.push(PatchOperation::Add {
+            path: path.into(),
+            value: serde_json::to_value(value)?,
+        });
+        Ok(self)
+    }
+
+    /// Adds a new "increment" operation to the patch document.
+    ///
+    /// See the [type documentation](PatchDocument) for more information on patch operations.
+    ///
+    /// # Arguments
+    /// * `path` - The path to the property to increment.
+    /// * `value` - The amount to increment by. Integers can be specified directly in this parameter. Use [`serde_json::Number::from_f64`] to create a floating-point number.
+    pub fn with_increment(
+        mut self,
+        path: impl Into<String>,
+        value: impl Into<serde_json::Number>,
+    ) -> Result<Self, serde_json::Error> {
+        self.operations.push(PatchOperation::Increment {
+            path: path.into(),
+            value: value.into(),
+        });
+        Ok(self)
+    }
+
+    /// Adds a new "remove" operation to the patch document.
+    ///
+    /// See the [type documentation](PatchDocument) for more information on patch operations.
+    ///
+    /// # Arguments
+    /// * `path` - The path to the property to remove.
+    pub fn with_remove(mut self, path: impl Into<String>) -> Result<Self, serde_json::Error> {
+        self.operations
+            .push(PatchOperation::Remove { path: path.into() });
+        Ok(self)
+    }
+
+    /// Adds a new "replace" operation to the patch document.
+    ///
+    /// See the [type documentation](PatchDocument) for more information on patch operations.
+    ///
+    /// # Arguments
+    /// * `path` - The path to the property to remove.
+    /// * `value` - The value to replace the property with.
+    pub fn with_replace(
+        mut self,
+        path: impl Into<String>,
+        value: impl Serialize,
+    ) -> Result<Self, serde_json::Error> {
+        self.operations.push(PatchOperation::Replace {
+            path: path.into(),
+            value: serde_json::to_value(value)?,
+        });
+        Ok(self)
+    }
+
+    /// Adds a new "set" operation to the patch document.
+    ///
+    /// See the [type documentation](PatchDocument) for more information on patch operations.
+    ///
+    /// # Arguments
+    /// * `path` - The path to the property to remove.
+    /// * `value` - The value to set the property to.
+    pub fn with_set(
+        mut self,
+        path: impl Into<String>,
+        value: impl Serialize,
+    ) -> Result<Self, serde_json::Error> {
+        self.operations.push(PatchOperation::Set {
+            path: path.into(),
+            value: serde_json::to_value(value)?,
+        });
+        Ok(self)
+    }
+
+    /// Adds a new "move" operation to the patch document.
+    ///
+    /// See the [type documentation](PatchDocument) for more information on patch operations.
+    ///
+    /// # Arguments
+    /// * `from` - The path to the property to move.
+    /// * `to` - The path to move the property to.
+    pub fn with_move(
+        mut self,
+        from: impl Into<String>,
+        to: impl Into<String>,
+    ) -> Result<Self, serde_json::Error> {
+        self.operations.push(PatchOperation::Move {
+            from: from.into(),
+            to: to.into(),
+        });
+        Ok(self)
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(tag = "op")]
+#[serde(rename_all = "camelCase")]
+pub enum PatchOperation {
+    Add {
+        path: String,
+        value: serde_json::Value,
+    },
+    #[serde(rename = "incr")]
+    Increment {
+        path: String,
+        value: serde_json::Number,
+    },
+    Remove {
+        path: String,
+    },
+    Replace {
+        path: String,
+        value: serde_json::Value,
+    },
+    Set {
+        path: String,
+        value: serde_json::Value,
+    },
+    Move {
+        from: String,
+        #[serde(rename = "path")]
+        to: String,
+    },
+}
+
+#[cfg(test)]
+mod tests {
+    use serde::Serialize;
+    use serde_json::Number;
+
+    use crate::models::PatchDocument;
+
+    #[derive(Serialize)]
+    struct TestStruct {
+        foo: String,
+        bar: i32,
+    }
+
+    #[test]
+    pub fn serialize_empty_patch_document() -> Result<(), Box<dyn std::error::Error>> {
+        let patch_document = PatchDocument::new();
+
+        let serialized = serde_json::to_string(&patch_document).unwrap();
+        assert_eq!(serialized, "{\"operations\":[]}");
+        Ok(())
+    }
+
+    #[test]
+    pub fn serialize_add() -> Result<(), Box<dyn std::error::Error>> {
+        let patch_document = PatchDocument::new().with_add(
+            "/parent",
+            TestStruct {
+                foo: "test".to_string(),
+                bar: 42,
+            },
+        )?;
+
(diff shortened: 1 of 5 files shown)


sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
```diff
@@ -0,0 +1,323 @@
...
+    /// Adds a new "add" operation to the patch document.
+    ///
+    /// See the [type documentation](PatchDocument) for more information on patch operations.
+    ///
+    /// # Arguments
+    /// * `path` - The path to the property to add.
+    /// * `value` - The value to add.
+    pub fn with_add(
+        mut self,
+        path: String,
```
analogrelay, line 53: Would like some input from you, @heaths . As well as anyone else reviewing this.

Taking paths as `String` does feel a little clunky. We expect users to call this and provide **either** string literals **or** owned `String` values from elsewhere in their application. For methods on `CosmosClient` and friends, we take `&str` because even if they _gave_ us a `String`, we'd be immediately concatenating that into a URL and dropping it, so we might as well be more user-friendly and accept `&str`.

I can see two ways these APIs would be called:

* With a string literal (`&'static str`), because the property path being patched is a constant.
* With a `String` from parsed, read, or received from some other place in the application, because the user, or a config file, or similar, specified the path to be patched.

So, we have these options:

1. Accept `impl Into<String>`. If you pass a literal or any other `&str`, we'll allocate a `String` (perhaps cloning the one the `&str` was borrowed from😬), incurring a hidden allocation. But, we can accept your `String` and move it in to the `PatchOperation` without allocating
2. Accept `&str`. Because we have to own our data (deriving `azure_core::Model` requires `DeserializeOwned`), we have to clone it immediately, incurring a hidden allocation.
3. Accept `Cow<'static, str>`. This allows us to accept string literals and owned Strings without cloning. But it's clunky to write at the call-site: `with_add("/path".into(), ...)`
4. Accept `impl Into<Cow<'static, str>>`. This **would** give us the ability to accept string literals and owned Strings without clunky syntax (`with_add("/path", ...)` and `with_add(my_string, ...)` would both work). However, it brings in more `impl` syntax, which we recently decided to avoid.

For now, I've gone with option 3, because the syntax isn't too much, the intent is clear, and we still get to avoid allocating `String`s for string literals. It does mean string literals have to be passed as `"/path".into()` but unless we go to option 4 (which violates our recent "avoid `impl` types" guidance), I don't think there's a better option.

sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
```diff
@@ -0,0 +1,323 @@
...
+        value: impl Serialize,
```
analogrelay, line 54: This feels like a clear exception to our "avoid `impl` types" guidance. `Serialize` is not object safe, so we can't accept `&dyn Serialize`, and the intent here is to accept anything that can be serialized, so you can write this:

```rust
let patch = PatchDocument::default()
    .with_add("/customer", Customer::new("John Doe"));
```

Or even support any number of already-serializable types from `std`:

```rust
let patch = PatchDocument::default()
    .with_add("/customer/name", "John Doe")
    .with_add("/customer/age", 42)
    .with_add("/customer/is_enterprise", true);
```

Thoughts, @heaths ? There's also an aesthetics decision to be made about whether or not these should be `impl Trait` or an actual generic parameter. Either is OK with me, but I find `impl Trait` a little more concise while remaining clear and explicit.

sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
```diff
@@ -0,0 +1,323 @@
...
+    ///
+    /// See the [type documentation](PatchDocument) for more information on patch operations.
+    ///
+    /// # Arguments
+    /// * `path` - The path to the property to increment.
+    /// * `value` - The amount to increment by. Integers can be specified directly in this parameter. Use [`serde_json::Number::from_f64`] to create a floating-point number.
+    pub fn with_increment(
+        mut self,
+        path: String,
+        value: serde_json::Number,
```
analogrelay, line 73: This is another bit of an API conundrum. We want to accept `serde_json::Number` so we can support integers or floats. We could support `impl Into<serde_json::Number>` but there are two problems there:

1. We just added a guideline to avoid `impl` types. Though if the scenario is compelling, that doesn't bother me much
2. While all integer types can be converted `Into<Number>`, there are no implementations of `From<T> for Number` for any floating point types, because that conversion is fallible (JSON forbids NaN and Infinity). There are also no implementations of `TryFrom` for `Number` (though `serde_json` could add that)

So, for now I'm taking `Number` directly. We could also have our own trait, `impl` it for all the numeric primitives, and use it as an `impl` type here. I'm open to that, if we wanted to hide `serde_json::Number` from our API surface...

sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
```diff
@@ -0,0 +1,323 @@
...
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(tag = "op")]
+#[serde(rename_all = "camelCase")]
+pub enum PatchOperation {
+    Add {
+        path: String,
+        value: serde_json::Value,
```
analogrelay, line 148: Storing this as `serde_json::Value` forces us to "pre-serialize" the value to JSON and then "serialize" it again from `serde_json::Value` in to the final body. However, our options are limited here, because `Serialize` isn't object-safe. 

There is the [`erased_serde` crate](https://crates.io/crates/erased-serde), which provides a type-erased `Serialize` that **is** object-safe, and we could use that and store `Box<dyn erased_serde::Serialize>` here.

This also means we're exposing `serde_json::Value` to our public API surface, which isn't ideal, but may be somewhat unavoidable here unless we use a newtype.

sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
analogrelay: The main alternative here would be to accept `serde_json::Value` directly, which pushes that serialization to the user at the call-site.

sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
heaths: It's not so much that we're trying to avoid `impl`, just reduce it so that the API isn't so "clunky". But we do maintain that it has its uses and the AMQP stack will still use it in places because it makes the API - sort of like what you're running into - more friendly. So, I say go with 1 or 4. I'm leaning toward 1 just because it sounds like you always need to own it anyway, so I don't see a `Cow` offering any advantage, or maybe I misunderstood that particular scenario. Are you saying it's only owned sometimes? If so, go with 4 consistently.

analogrelay: I don't think we need to own it. In a lot of cases users are providing static string literals here. We need a reference that lives long enough to go from when the user creates the `PatchDocument` to when they call `patch_item`, which will usually be pretty short.

I contemplated having `PatchOperation` hold a reference to the string but that got really messy with `DeserializeOwned` and `Model`.

I'll go with 4 for now then!

sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
heaths: Again, we don't want to avoid `impl` entirely - just in places where it doesn't help or hides allocations. I think it's fine to use in many cases. Maybe we need to clarify in guidance where we don't *ideally* want it used e.g., when passed to client or client methods that are just part of URLs or a body. When used in higher-order abstractions, `impl` may be fine. It would be hard to qualify, though, so maybe we just keep it as a architectural design question per crate. @RickWinter or @JeffreyRichter, what do you think?

sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
heaths: I think I'd go with 1 in this case. Callers may not realize how to get a `serde_json::Number`, nor would I want to expose other crates' types in our public API (so, yes, maybe we do need our own trait like I did in https://github.com/heaths/ordinal-rs). That said, serde and friends are ubiquitous and may be safe.

sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
heaths: I'd rather not take yet another crate (`erased_serde`) just for this. While not ideal to expose `serde_json` types, I'm not as worried about the serde family of crates. They're ubiquitous. I'm also not too concerned about object safety unless we know we need it somewhere. So many things we do or even use are not object safe that it feels like a bygone conclusion.

That said, I do like the newtype idea here and for `Number`, if possible, as well. Also gives us some flexibility to impl other traits or add methods as needed so we'll own the types.

sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
Pilchie: Newtonsoft.Json was once "ubiquitous and safe" too ;)

Pilchie: I do think serde is even more entrenched though.

Pilchie (approved):
I'll let you and Heath hammer out the API questions - my rust isn't good enough to have opinions there.

sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
JeffreyRichter: I agree. The guidance is to avoid it for values related to making the HTTP request/response itself. Our guidance does not cover other values typically and I don't think we need to address it in our guidance. But perhaps make it clear that our guidelines are specifically about clients and their methods related to HTTP requests/responses.

Similarly, the .NET library guidance is about the PUBLIC face of the library and not about its internal implementation.

analogrelay: Sounds good, I definitely think the `impl Trait` syntax is really helping here.

> the .NET library guidance is about the PUBLIC face of the library

I do want to make sure it's clear that this is absolutely public API, it's basically a builder API for a model type. And to that point, this is also a case where we're looking to make an exception to our "avoid builders" guideline, since I think this is exactly a situation where a builder is useful. This isn't quite a "builder" since there's no separate builder type but it has the same effect. Again, this feels like a good example of where it **does** make sense to have a builder, since the APIs for constructing a `PatchDocument` manually, while still available, are very clunky.

The above example:

```rust
let patch = PatchDocument::default()
    .with_add("/customer/name", "John Doe")
    .with_add("/customer/age", 42)
    .with_add("/customer/is_enterprise", true);
```

If written without the `with_` "consuming setters" that are effectively a built-in builder, would look like this:

```rust
let patch = PatchDocument {
    operations: vec![
        PatchOperation::Add {
            path: "/customer/name".into(),
            value: serde_json::Value::String("John Doe".into()),
        }
        // And I'm already tired of typing and don't want to finish, with two operations still to add 😛 
    ]
}
```

I think we're all on the same page there, just laying out the places where we are wiggling around the guidelines, to help demonstrate the concrete "exceptions".

JeffreyRichter: Oh, I'm afraid my analogy confused things. Forget I said anything about .NET and public apis. Our SDK Rust guidance is specifically about the client types and methods that make HTTP calls. It is NOT about other parts of the SDK that are service-specific.

sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
analogrelay: A custom trait is easy enough and makes sense, IMO. Do we think this should be promoted up to `azure_core` or `typespec_client_core` or just live in the Cosmos SDK? I'll start with the latter for now, since we could move and re-export if necessary.

sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
analogrelay: I was pretty sure you followed, I just wanted to make sure things were absolutely clear 😉

sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
analogrelay: > Newtonsoft.Json was once "ubiquitous and safe" too ;)
> I do think serde is even more entrenched though.

Certainly a valid concern, particularly given our own history with that particular situation, and the current struggles the .NET SDK is going through ;).

My sense from the Rust community is that serde has become a de-facto part of the "extended" stdlib and Rust is **much less** likely to bring serialization in to the real stdlib than .NET was. .NET included XML serialization from v1 (I believe it was there that early) whereas Rust has _no_ serialization in the stdlib.

The other element here is that Rust's Trait system is much more flexible than .NET's inheritance and interfaces. So even if a stdlib serialization system comes in, `serde_json` could easily adapt its types to support it **and** `serde`'s. That's a lot harder to do in .NET

It's definitely good to consider, but I think we're in a safer place than we're in a much better place depending on serde and friends than we were when we depended on Newtonsoft.Json.

PR #1915, opened 2024-11-13
Add get_key sample

heaths:


chlowell (approved):
Does this need a license header?

heaths:
> Does this need a license header?

I could add one, but almost none of the other track 1 source has it. We'll have a check at some point in track 2 but need to write it. I was unsuccessful in getting a clippy-standard lint implemented despite a community desire to.

PR #1917, opened 2024-11-14
[Cosmos] Initial integration tests

analogrelay:
Here's a first bit of infrastructure for integration tests for `azure_data_cosmos`, and a single test.

To run these tests:

* Set `AZURE_COSMOS_CONNECTION_STRING` environment variable to a valid Connection String to a real Cosmos DB account (the Emulator isn't quite supported yet, I'm having TLS issues that I need to find a good solution for, the emulator has a self-signed certificate).
* Run tests with `AZURE_TEST_MOST=live` and `--all-features` (because the current tests use the `key_auth` feature)

I still need to work out how to integrate these with CI, I'll start a separate conversation to discuss that.


analogrelay:
After some discussion with @heaths, I'm going to change this a bit already: Instead of `TestAccount::from_env()`, use a macro: `get_test_account!()` (or similar). That macro will `return Ok(())` if the environment variable isn't set, thus effectively skipping the test.

analogrelay:
Ok, here's a different approach, based on some discussions in Teams.

1. "Live" tests, which depend on some set of "real" resources (including emulators) are marked with `#[cfg_attr(not(livetest), ignore]`, which applies the `#[ignore]` attribute if `--cfg=livetest` is **not** specified when `rustc` is building the tests.
2. In addition to the `--cfg` value, we specify a Cargo profile called `livetest`, which ensures that the build of live tests doesn't interfere with the dev build, causing the build cache to be invalidated.
3. To make it easier to specify all the necessary options, I've defined an alias in `.cargo/config.toml`. Running `cargo livetest ...` will be equivalent to running `cargo test` with `--profile=livetest` (to use a separate build cache) and `RUSTFLAGS=--cfg=livetest` (to compile-in the live tests themselves).

With this, if you run the following command: `cargo test --all-features --package azure_data_cosmos --test integration_databases` (to run only "non-live" tests), you get this output:

```
     Running tests/integration_databases.rs (target/debug/deps/integration_databases-acf8e6c1c2544f49)

running 1 test
test database_crud ... ignored
```

The `database_crud` test has been ignored, as desired.

When you run `cargo livetest --all-features --package azure_data_cosmos --test integration_databases` (to run all tests, including livetests), but without specifying mandatory env vars, you get this output:

```
     Running tests/integration_databases.rs (target/livetest/deps/integration_databases-ba888bd60bb98b5c)

running 1 test
test database_crud ... FAILED

failures:

---- database_crud stdout ----
Error: "failed to read AZURE_COSMOS_CONNECTION_STRING environment variable: environment variable not found"


failures:
    database_crud
```

The "live" test has _correctly_ failed because you requested to run it but failed to provide mandatory configuration. If you then configure the `AZURE_COSMOS_CONNECTION_STRING` value correctly, and re-run the test, it runs as expected:

```
     Running tests/integration_databases.rs (target/livetest/deps/integration_databases-ba888bd60bb98b5c)

running 1 test
test database_crud ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.18s
```

analogrelay:
This is on hold as we get the "official" plan for integration/live testing in place.

.devcontainer/devcontainer.json
```diff
@@ -9,7 +9,8 @@
   "features": {
     "ghcr.io/devcontainers/features/azure-cli:1": {},
     "ghcr.io/devcontainers/features/github-cli:1": {},
-    "ghcr.io/devcontainers/features/sshd:1": {}
+    "ghcr.io/devcontainers/features/sshd:1": {},
+    "ghcr.io/devcontainers/features/docker-in-docker:2": {}
```
analogrelay, line 13: This is to enable the use of the Cosmos emulator docker container, for testing. We generally test against the Emulator in a lot of our SDKs, including in CI builds: https://github.com/Azure/azure-sdk-for-go/blob/f647b1a888568b65e9bac0fcb2f1f494a660f6c1/sdk/data/azcosmos/ci.yml#L45

sdk/cosmos/azure_data_cosmos/tests/framework/test_account.rs
```diff
@@ -0,0 +1,181 @@
...
+                .danger_accept_invalid_certs(true)
+                .pool_max_idle_per_host(0)
+                .build()?;
+            options = {
+                let mut o = options.unwrap_or_default();
+                o.client_options
+                    .set_transport(TransportOptions::new(Arc::new(client)));
+                Some(o)
+            };
+        }
```
analogrelay, line 140: Calling some special attention here, since I'm calling `danger_accept_invalid_certs`.

The emulator uses a self-signed certificate. I tried putting the cert in the root store, but something in the pipeline is still unhappy with that _specifically because the certificate is self-signed_ (i.e. even if it's trusted and `curl` and `openssl s_client` are both happy). So, we give the option to disable cert validation _in these tests only_ (notice that there's no built-in option to do this in the SDK itself, it's a custom transport used by the tests). The flag to bypass cert validity checks is set in **either** of these circumstances:

* `AZURE_COSMOS_CONNECTION_STRING=emulator` - The user is explicitly using the emulator, thus we know we'll have a self-signed cert.
* `AZURE_COSMOS_ALLOW_INVALID_CERT=true` - The user has explicitly requested that we allow invalid certs [^1].

[^1]: Again, I want to make clear that it's the _test_ reading this value. This environment variable is meaningless to the SDK and won't disable any checks in the SDK itself.

analogrelay: In fact, `rustls` specifically disallows self-signed certificates **even if they are trusted roots**, unless you set `danger_accept_invalid_certs`:

>> For reasons explained in the other sections of this manual, rustls does not and will not support: 
>> * ...
>> * Using CA certificates directly to authenticate a server/client (often called “self-signed certificates”). Rustls’ default certificate verifier does not support using a trust anchor as both a CA certificate and an end-entity certificate in order to limit complexity and risk in path building. While dangerous, all authentication can be turned off if required – see the [example code](https://github.com/rustls/rustls/blob/992e2364a006b2e84a8cf6a7c3eaf0bdb773c9de/examples/src/bin/tlsclient-mio.rs#L318) *
>
> https://docs.rs/rustls/latest/rustls/manual/_04_features/index.html#non-features

Pilchie (approved):

sdk/cosmos/azure_data_cosmos/tests/framework/test_account.rs
Pilchie: Quick question here - for the new emulator, we are moving towards using `http` instead of `https` by default, so we can get out of the self-signed cert business a bit. Would anything need to change here to be able to use that?

analogrelay: Using `http` for the emulator would certainly be preferable as it makes clear the security limitations (i.e. the emulator being intended strictly for development/testing on your local machine). I don't think anything blocks us from using `http` in our SDK. We expect a full URL, scheme included, everywhere, so if you specify `http`, we'll use `http`.

Does the current emulator support `http` at all?

Pilchie: > Does the current emulator support http at all?

Sadly, no.

.devcontainer/devcontainer.json
heaths: Nit: I prefer alphabetical order so it's easy to find things as the list expands. And that would only be a 1-line addition as opposed to a 2-line diff.

heaths:

sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -42,6 +44,7 @@ default = ["hmac_rust"]
 hmac_rust = ["azure_core/hmac_rust"]
 hmac_openssl = ["azure_core/hmac_openssl"]
 key_auth = [] # Enables support for key-based authentication (Primary Keys and Resource Tokens)
+test_e2e = ["key_auth"]
```
heaths, line 47: Could we phase this out? I got rid of it in Event Hubs because we can't do `--all-features` to test, which left many tests in a failed state as you may recall. You can check the `TestContext::test_mode()` accessor to see if you're running "live", for example, if that helps.

sdk/cosmos/azure_data_cosmos/tests/cosmos_databases.rs
```diff
@@ -0,0 +1,116 @@
...
+
+use std::error::Error;
+
+use azure_core_test::{recorded, TestContext};
+use azure_data_cosmos::{models::ThroughputProperties, CreateDatabaseOptions, Query};
+use framework::TestAccount;
+use futures::StreamExt;
+
+#[recorded::test(live)]
+#[cfg(feature = "key_auth")]
```
heaths, line 11: Once we have tests recordings - my goal before I leave for the year - will we be able to pass `--all-features` to our CI tests? I.e., can the impact here be recorded (perhaps with sanitizers/matches)?

sdk/cosmos/azure_data_cosmos/tests/framework/test_account.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// This database will be automatically deleted when [`TestAccount::cleanup`] is called.
+    pub fn unique_db(&self, base_id: &str) -> String {
+        format!("{}_{}", base_id, self.context_id)
+    }
+
+    /// Cleans up test resources, then drops the [`TestAccount`].
+    ///
+    /// Call this at the end of every test using the [`TestAccount`].
+    #[cfg(feature = "key_auth")]
+    pub async fn cleanup(self) -> Result<(), Box<dyn std::error::Error>> {
```
heaths, line 160: Q: Should I provide a `defer`-like `Drop` implementation on `TestContext`? For recorded tests, a `TestContext` is initialized at the beginning of every test and will be required to instrument clients for recording/playback, as well as adding sanitizers. Assuming captures work across async boundaries (or maybe we need some work here), we could do something like this:

```rust
#[recorded:test]
async fn test_foo(ctx: TestContext) -> Result<(), Box<dyn std::error::Error>> {
    let account = TestAccount::from_env(&ctx, ...);
    ctx.cleanup(async move || account.cleanup().await);

    // ...
}
```

heaths (changes requested):
I see nothing blocking, but would like to discuss more about how to use features or at least formulate a plan how to get to the point we can pass `--all-features` to test.

sdk/cosmos/azure_data_cosmos/Cargo.toml
analogrelay: This can actually be removed, I just forgot to.

sdk/cosmos/azure_data_cosmos/tests/framework/test_account.rs
analogrelay: I was thinking of suggesting that! I'd like to be able to "automate" some degree of cleanup and actually started with a `Drop` implementation, but realized it needs async and that's kinda messy.

sdk/cosmos/azure_data_cosmos/tests/cosmos_databases.rs
analogrelay: The test can run under `--all-features` already. In fact, Cosmos is ready to roll with `--all-features` (and I would, in fact, prefer we enable it ASAP, to catch `key_auth`-related tests).

Right now, it's "live only" because we have no recordings, but it's should work fine under recorded tests.

I should note that in all our other language SDKs, Cosmos has per-PR test runs that use the Emulator (which act as "live" tests because they talk to a "real" emulator). We'll be looking to do the same here too.

sdk/cosmos/azure_data_cosmos/tests/framework/test_account.rs
heaths: One day maybe, perhaps, it might happen possibly: https://rust-lang.github.io/async-fundamentals-initiative/roadmap/async_drop.html

.devcontainer/devcontainer.json
analogrelay: Agreed! I wish it were possible to make alphabetizing JSON an "auto-format on save" option :). Maybe it is...? 🤔

sdk/cosmos/azure_data_cosmos/tests/cosmos_databases.rs
heaths: @hallipr we'll want to look at how other languages "extend" the CI pipeline and do the same here such that Cosmos (and probably Storage) can run emulators as well.

sdk/cosmos/azure_data_cosmos/tests/framework/test_account.rs
analogrelay: Shall I go ahead and add that in this PR? I'm happy to do so, or just merge it and update it when it's added in a separate PR

analogrelay: Ah, I just thought of a possible issue that makes this complicated enough to avoid for now: How should the cleanup run if the test panics? Do we need to [catch_unwind](https://doc.rust-lang.org/stable/std/panic/fn.catch_unwind.html)? What if the cleanup code panics (currently, if we didn't catch a cleanup panic, it would mask the actual test failure).

sdk/cosmos/azure_data_cosmos/tests/cosmos_databases.rs
heaths: The plan was always to follow suit with the other languages. @hallipr has some of that working already, but I don't think it fully ingests individual `ci.yml` files like in other languages yet.

heaths (approved):

PR #1921, opened 2024-11-19
Add serde helpers for base64 encoding/decoding

jhendrixMSFT:
To be used in serde "with" derive macros for struct fields that contain base64 data.

```rust
pub struct Base64BytesProperty {
    #[serde(
        deserialize_with = "base64::deserialize",
        skip_serializing_if = "Option::is_none",
        serialize_with = "base64::serialize"
    )]
    pub value: Option<Vec<u8>>,
}
```

### DIFF
--- a/sdk/typespec/typespec_client_core/src/base64.rs
+++ b/sdk/typespec/typespec_client_core/src/base64.rs
@@ -61,11 +61,12 @@ where
 /// # Examples
 ///
 /// ```rust,no_run
-/// # use serde::{Deserialize}
+/// # use serde::{Deserialize};
+/// # use typespec_client_core::base64;
 /// #[derive(Deserialize)]
 /// struct SomeType {
 ///     #[serde(deserialize_with = "base64::deserialize")]
-///     pub value: Vec<u8>,
+///     pub value: Option<Vec<u8>>,
 /// }
 /// ```
 pub fn deserialize<'de, D>(deserializer: D) -> Result<Option<Vec<u8>>, D::Error>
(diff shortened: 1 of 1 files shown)


sdk/typespec/typespec_client_core/src/base64.rs
```diff
@@ -51,3 +52,50 @@ where
...
+
+pub fn serialize<S>(to_serialize: &Option<Vec<u8>>, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    let encoded = to_serialize.as_ref().map(encode);
+    <Option<String>>::serialize(&encoded, serializer)
+}
+
+pub fn serialize_url_safe<S>(
```
heaths, line 92: Does declaring a `T` as `AsRef<[u8]>` not work here like the `serialize` functions above? More flexible.

heaths (approved):
Nothing blocking, but please add some comments (all new code should; I'll do a pass on old stuff before we GA if not sooner) and maybe a round-trip test.

sdk/typespec/typespec_client_core/src/base64.rs
jhendrixMSFT: Would the param be an `Option<T>` or just a `T`?

sdk/typespec/typespec_client_core/src/base64.rs
```diff
@@ -51,3 +52,102 @@ where
...
+/// 
+/// # Examples
+/// 
+/// ```rust,no_run
+/// struct SomeType {
+///     #[serde(deserialize_with = "base64::deserialize")]
+///     pub value: Vec<u8>,
+/// }
+/// ```
+pub fn deserialize<'de, D>(deserializer: D) -> Result<Option<Vec<u8>>, D::Error>
```
jhendrixMSFT, line 69: Should `Option` be removed here?

sdk/typespec/typespec_client_core/src/base64.rs
heaths: Ideally, we'd have one with and without `Option<T>`, but we don't really need one sans `Option` so I wouldn't worry about that. All our model fields are optional and we can always add non-optional support later.

jhendrixMSFT: Fixed.

PR #1926, opened 2024-11-22
Implement `recorded::test` macro for recorded tests

heaths:
- **Rename typespec_derive to typespec_macros**
  Will be more consistent with upcoming azure_core_macros, and may not contain *just* derive macros anyway. Will keep feature as "derive", though, for derive macros.
  

- **Remove 128-bit number functions for Cosmos**
  

- **Implement `#[recorded]` attribute macro**
  

- **Allow live-only tests with no parameters**
  

- **Replace test_e2e feature with `recorded` attribute macro**


heaths:
I also had to bump the MSRV to 1.80 to use the stable `LazyLock` so we're not having to read the `AZURE_TEST_MODE` env var (or lack there of) and parsing the string for each method. We wouldn't *have* to do that, but 1.80 is also currently behind the latest by 2 versions (1.82).

analogrelay (approved):
I like this. My only thought, is maybe we rename this "recorded_test" or just call the attribute "test" and but it in a `recorded` module with the intent of being referenced as `#[recorded::test]`? Since it's effectively replacing the test marker entirely and the user is expected to NOT apply a standard test attribute, I think that including "test" in the name would make that clear.

analogrelay:
> I also had to bump the MSRV to 1.80 to use the stable LazyLock so we're not having to read the AZURE_TEST_MODE env var (or lack there of) and parsing the string for each method.

I'm fine with bumping our MSRV, though I'll be honest I'm not sure it's necessary to cache the test mode value like this. It should work just fine though, and I like getting new Rust versions 😁

LarryOsterman (approved):

analogrelay (approved):
LGTM! Thanks for making those little fix-ups in the Cosmos crate. Let's get this rolling!

LarryOsterman:
> > I honestly do like the look of #[recorded::test] much better but #[recorded_test] would be fine, too.
> 
> Worth noting that this aligns with the existing community pattern for async tests: `#[tokio::test]`. That attribute isn't named `tokio_test`, it's just named `test` and in the `tokio` module. Seems reasonable to me to have own own `#[recorded::test]` in that case, particularly since it's a replacement for `#[test]` or `#[tokio::test]`. I do think the pattern should be to never import the macro directly.

100% agreed. And I suspect that if you attempted to import the `recorded` module it would get upset between the default declared `test` and the `test` from the `recorded` module (but I obviously haven't tested that).

PR #1937, opened 2024-11-28
Bring `Method` and `StatusCode` types from `http-types` into `typespec`

### LINKED ISSUE #1644: Consider replacing http-types crate with http crate

joonas:
I was hoping to be able to get rid of the `http-types` crate given that it's not actively being updated and one of the dependencies of it's dependencies has an outstanding RUSTSEC advisory ([`RUSTSEC-2024-0384`](https://rustsec.org/advisories/RUSTSEC-2024-0384.html)).

Since `typespec` and `typespec_client_core` are only making use of the `Method` and `StatusCode` types, I thought bringing them into the `typescript` crate and using them from there seemed appropriate.

Closes #1644

### DIFF
--- a/sdk/typespec/Cargo.toml
+++ b/sdk/typespec/Cargo.toml
@@ -13,14 +13,15 @@ keywords = ["typespec"]
 
 [dependencies]
 base64.workspace = true
-http-types = { workspace = true, optional = true }
+serde = { workspace = true, optional = true }
 serde_json = { workspace = true, optional = true }
 url.workspace = true
 
 [dev-dependencies]
 thiserror.workspace = true
+serde_json.workspace = true
 
 [features]
 default = ["http", "json"]
-http = ["dep:http-types"]
-json = ["dep:serde_json"]
+http = []
+json = ["dep:serde", "dep:serde_json"]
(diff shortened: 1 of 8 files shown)


joonas:

joonas:
It looks like the analyze job doesn't like the HTTP methods [`CHECKIN`](https://datatracker.ietf.org/doc/html/rfc3253#section-9.4) or [`MKACTIVITY`](https://tools.ietf.org/html/rfc3253#section-13.5), should the be added to a dictionary somewhere?

joonas:
@heaths that sounds great, at the end of the day my main concern here is to be able to address [`RUSTSEC-2024-0384`](https://rustsec.org/advisories/RUSTSEC-2024-0384.html) as a consumer of these libraries.

heaths:
I appreciate that, but it seems a lot of crates are unmaintained. What happens when `http` falls out of maintenance? We may need to consider vendoring instead. Maybe we're not there with this small change (2 types) yet, but once we hit GA we have a strict* no-breaking-changes policy.

*) Well, mostly strict. For Rust, as opposed to most of our other languages, client APIs may break due to service REST API approved breaks but we'll also shipping 1:1 crate version:service API version.

heaths:
Replaced by #2233, which is more comprehensive and against a relatively current HEAD.

PR #1947, opened 2024-12-06
Commit Cargo.lock

heaths:
More reproducable builds and necessary for CG.


heaths:
@hallipr we may also want to consider having CI run `cargo` commands with `--locked`, which uses the versions in `Cargo.lock` much like npm uses `package-lock.json` when you run `npm ci`. That said, there's a debate in the community whether this is a good idea or not because it potentially means that dev builds and CIs are potentially *slightly* different i.e., that patch semvers may be different. Personally, I'm in camp "just do it the same everywhere". Still, this is necessary for Component Governance to detect the components we use and will help avoid some issues that @LarryOsterman and I ran into a while back since I had an older `Cargo.lock` that satisfied `Cargo.toml` version requirements than he did, and there APIs on which we relied were incompatible.

hallipr (approved):

PR #1954, opened 2024-12-11
Proposal: Rework body deserialization on Response<T>

analogrelay:
To go along with some internal discussions about reworking the body-related methods on `Response<T>`, here's a draft of the changes I'm proposing. To reiterate the specific changes:

1. Rename `Response::<T>::into_body(self) -> ResponseBody` to `into_raw_body`, since it extracts the raw body. This removes the conflict with 2 below and provides a way for all `Response<T>` instances to access the raw bytes, regardless of the model type `T`.
1. Rename `Response::<T: Model>::deserialize_body(self)` to `into_body` 
1. Remove `Response::<T>::deserialize_body_into<U: Model>(self)` and replace with the below two methods. These are just convenient short-hands for `into_raw_body().json()` and `into_raw_body().xml()`. This requires the user to know the serialization format, but they already had to do so in order to derive `Model` (which was previously required by `deserialize_body_into`) and this saves them from having to annotate their domain-specific type with that information.
    1. `Response::<T>::into_json_body<U: DeserializeOwned>(self)`
    1. `Response::<T>::into_xml_body<U: DeserializeOwned>(self)`
1. **Optional** (included in this draft): Two related changes that must be made together if made
    1. To make "raw" responses easier to process, implement `Model` on `ResponseBody` as an identity transformation (since `Model::from_response_body(ResponseBody) -> ResponseBody` is just an identity function). This allows `Response::<ResponseBody>::into_body(self)` to work to extract the raw body from a "raw" response.
    1. Change the default `T` on `Response` from `()` to `ResponseBody`. This would allow `Response<()>` to be used to represent truly empty responses that are always empty. Note that I have **not** included any logic for `Response<()>` at this time. All scenarios that _were_ using `Response<()>` are now using `Response` => `Response<ResponseBody>`.


sdk/typespec/typespec_client_core/src/http/response.rs
```diff
@@ -146,16 +150,60 @@ impl<T> Response<T> {
...
+    ///   let my_struct: MySecretResponse = response.into_xml_body().await.unwrap();
+    ///   assert_eq!("hunter2", my_struct.value);
+    /// }
+    ///
+    /// # #[tokio::main]
+    /// # async fn main() {
+    /// #    let r: Response<GetSecretResponse> = typespec_client_core::http::Response::from_bytes(
+    /// #      http_types::StatusCode::Ok,
+    /// #      typespec_client_core::http::headers::Headers::new(),
+    /// #      "<Response><name>database_password</name><value>hunter2</value></Response>",
```
RickWinter, line 190: FYI, security scanners might flag this as a secret... it looks real enough even though we all know it isn't

analogrelay: Good point. We already had a JSON equivalent to this and it hasn't been caught so far, but we can watch out for it.

eng/scripts/Analyze-Code.ps1
```diff
@@ -34,7 +34,7 @@ if ($CheckWasm) {
   Invoke-LoggedCommand "cargo +$Toolchain clippy --target=wasm32-unknown-unknown --workspace --keep-going --no-deps"
 }
 
-Invoke-LoggedCommand "cargo +$Toolchain doc --workspace --no-deps"
```
analogrelay, line 37: I'm referencing `into_xml_body` in a doc comment, which only exists in the `xml` feature. IMO, our docs should build with all reasonable features, and include "Requires `xyz` feature" notices as necessary (docs.rs can do this, and I've enabled it in `azure_data_cosmos`). So, I figured the best fix was to have our analyze pass run the docs build with `--all-features`. The alternative, which I'm open too if we want to avoid this change, is to remove the reference to `into_xml_body`.

sdk/typespec/typespec_client_core/src/http/response.rs
heaths: Could use "hunter2". That'll get masked automatically, right?

eng/scripts/Analyze-Code.ps1
heaths: I want to enable `--all-features` anyway in every run, but the `e2e_test` and your `keyauth` feature prevent that. We should eliminate those or find another way.

analogrelay: The key auth feature shouldn't be blocking that. It's fully additive and intended to be tested in unit tests.

heaths (approved):
Approving to unblock. I don't have anything blocking anyway, actually: just some suggestions, and I'm starting to rethink the prefix "into". Then again, it fits the idiomatic prefix naming convention, but "into_json_body()", for example, just gives me (and likely those new to the API) the wrong idea. It's actually *from* a JSON body.

But we can bikeshed over names later, perhaps in a tracking issue. If you don't have any better ideas or feel strongly this is the best (for now / forever), I'm not going to block it especially since I'll be OOF.

sdk/cosmos/azure_data_cosmos/examples/cosmos/patch.rs
```diff
@@ -48,7 +48,7 @@ impl PatchCommand {
         match response {
             Err(e) if e.http_status() == Some(StatusCode::NotFound) => println!("Item not found!"),
             Ok(r) => {
-                let item: serde_json::Value = r.deserialize_body_into().await?;
+                let item: serde_json::Value = r.into_json_body().await?;
```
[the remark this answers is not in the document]
analogrelay: I don't love the naming either, but I'll open a tracking issue because I don't want to get totally bogged down in (fully legitimate, but less time-sensitive) naming bikesheds.

The main value of `into_json_body` is that it aligns with `into_raw_body` and `into_body`. A clearer name might be `into_model_from_json` or just `into_json_model`, but I'll open an issue to have that debate.

sdk/typespec/typespec_client_core/src/http/response.rs
```diff
@@ -279,6 +327,12 @@ impl ResponseBody {
     }
 }
 
+impl Model for ResponseBody {
```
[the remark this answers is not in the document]
analogrelay: We could. We also don't _need_ to do it, but it means a plain `Response<ResponseBody>` would have no `into_body` method. Instead, it would have only `into_{raw,json,xml}_body` methods.

How about I try removing it for now, we can restore it later if it's proving inconvenient. It's non-breaking to add later, but breaking to remove later.

sdk/cosmos/azure_data_cosmos/examples/cosmos/patch.rs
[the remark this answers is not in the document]
analogrelay: https://github.com/Azure/azure-sdk-for-rust/issues/1957

sdk/typespec/typespec_client_core/src/http/response.rs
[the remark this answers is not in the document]
heaths: We also get it from `into_raw_body()`, right? In which case I think it's explicit. Yeah, I like the idea of not implementing `Model` for now but we could keep it in our back pocket.

heaths (approved):

Issue #1874, opened 2024-10-24
Add Test Proxy support


heaths:
We need to generate the test proxy and have support to use it in all our integration tests. This should be as easy as possible to use, much like .NET.

We should consider putting all this in a separate crate e.g., `azure_core_test` or maybe even `typespec_client_core_test`. Depends how much is Azure-specific (probably everything since it hinges on Test Proxy which, AFAIK, is only geared at Azure SDKs).

RickWinter:
@heaths All issues on this are done. Is there anything left you are keeping this issue open for?

rcorre:
Is this intended only for use inside the azure SDK, or can it also be used for integration tests in external projects that depend on the SDK? I noticed that the `azure_core_test` crate doesn't seem to be published to crates.io.

heaths:
It's only for use developing the Azure SDK at this time.

rcorre:
Thanks! So is there a practical way to integration test code against a test server, or do you just have to mock out the SDK APIs in your code?

heaths:
You have to mock the responses currently. We don't have plans at this time to expose https://github.com/Azure/azure-sdk-for-rust/blob/c6b0f3a85aa146f48b5ea7e7b799f3cf23af2a32/sdk/core/azure_core_test/src/http/clients.rs#L53 but we could. /cc @ronniegeraghty 

We could, perhaps, put it behind the `azure_core/test` feature. The test-proxy is specific to Azure SDKs and wouldn't work well for customer code. You might check out [Dev proxy](https://learn.microsoft.com/microsoft-cloud/dev/dev-proxy/overview).

Issue #1955, opened 2024-12-11
Automatically acquire the test proxy


heaths:
Either in lieu of checking if we can start test-proxy (should at least check the version somehow) or in addition to, we should acquire an appropriate published package similar to how Go does it: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/internal/recording/server.go

heaths:
Related to Azure/azure-sdk-tools#9467, we should at least check / acquire Test-Proxy version `20241213.1` or newer.

PR #1956, opened 2024-12-12
Bootstrap test-proxy service

heaths:
Relates to #1874 but does not complete it.

### DIFF
--- a/sdk/core/azure_core_test/src/proxy.rs
+++ b/sdk/core/azure_core_test/src/proxy.rs
@@ -27,10 +27,8 @@ pub async fn start(
     options: Option<ProxyOptions>,
 ) -> Result<Proxy> {
     if env::var(PROXY_MANUAL_START).is_ok_and(|v| v.to_ascii_lowercase() == "true") {
-        return Err(azure_core::Error::message(
-            ErrorKind::Other,
-            format!("environment variable {PROXY_MANUAL_START} is 'true'; not starting test proxy"),
-        ));
+        tracing::event!(target: crate::SPAN_TARGET, Level::WARN, "environment variable {PROXY_MANUAL_START} is 'true'; not starting test proxy");
+        return Ok(Proxy::default());
     }
 
     // Find root of git repo or work tree: a ".git" directory or file will exist either way.
@@ -67,7 +65,10 @@ pub async fn start(
     let mut stdout = command.stdout.take();
     let url = wait_till_listening(&mut stdout).await?;
 
-    Ok(Proxy { command, url })
+    Ok(Proxy {
+        command: Some(command),
+        url,
+    })
 }
 
 async fn wait_till_listening(stdout: &mut Option<ChildStdout>) -> Result<Url> {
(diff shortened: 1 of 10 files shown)


heaths:
@RickWinter @analogrelay I'm still working on the transport but nearly done. I figured enough was here already I should push the bootstrapping portion (minus automatic acquisition, tracked in #1955 so I can get an MVP out by EOW) to alleviate some of the review burden. I'll `rebase --onto` whatever gets merged here.

sdk/core/azure_core_test/src/proxy.rs
```diff
@@ -0,0 +1,162 @@
...
+
+pub async fn start(
+    test_data_dir: impl AsRef<Path>,
+    options: Option<ProxyOptions>,
+) -> Result<Proxy> {
+    if env::var(PROXY_MANUAL_START).is_ok_and(|v| v.to_ascii_lowercase() == "true") {
+        return Err(azure_core::Error::message(
+            ErrorKind::Other,
+            format!("environment variable {PROXY_MANUAL_START} is 'true'; not starting test proxy"),
+        ));
```
analogrelay, line 33: Should this be an error? If a user has manually started the proxy and then wants to run a test against the manually-started proxy, wouldn't this cause their test to fail, assuming there is code calling `proxy.start()?`?

sdk/core/azure_core_test/src/sanitizers.rs
```diff
@@ -0,0 +1,118 @@
...
+        json_path: String::from("$.values"),
+        value: None,
+        regex: None,
+        group_for_replace: None,
+    };
+    let headers = sut.as_headers().expect("expect headers");
+    headers.for_each(|(h, v)| {
+        assert_eq!(h.as_str(), "x-abstraction-identifier");
+        assert_eq!(v.as_str(), "BodyKeySanitizer");
+    });
```
analogrelay, line 81: nit: This won't fail if `as_headers` returns empty or an iterator with 15 `x-abstraction-identifier: BodyKeySanitizer` headers, if I'm reading right. I'd prefer a `.collect::<Vec<_>>()` and then assert the collected vector to ensure we're asserting the exact content of the iterator.

sdk/core/azure_core_test/src/sanitizers.rs
```diff
@@ -0,0 +1,118 @@
...
+impl_sanitizer!(BodyRegexSanitizer);
+
+#[test]
+fn test_body_regex_sanitizer_as_headers() {
+    let sut = BodyRegexSanitizer::default();
+    let headers = sut.as_headers().expect("expect headers");
+    headers.for_each(|(h, v)| {
+        assert_eq!(h.as_str(), "x-abstraction-identifier");
+        assert_eq!(v.as_str(), "BodyRegexSanitizer");
+    });
```
analogrelay, line 117: Same here.

analogrelay (approved):
One small issue with tests on the sanitizers, but I trust you to take care of it, so nothing blocking on my end.

sdk/core/azure_core_test/src/sanitizers.rs
heaths: Good catch.

sdk/core/azure_core_test/src/proxy.rs
heaths: You're right. I incorrect transcribed the Go source (the intent, rather; Go isn't the problem at all) that I pulled this from.

analogrelay (approved):

PR #1963, opened 2024-12-16
[Cosmos] "Enable Content Response on Write" option and integration tests for items and query

analogrelay:
This PR changes the default behavior for write operations to align with other Cosmos SDKs. By default, we instruct the server _not_ to send the updated content of an item when writing it (Upsert, Replace, Patch, etc.). If a user wants the content returned for these operations, they must set the `enable_content_response_on_write` option. As part of this, I also changed the return types a bit:

* Read operations return `Response` instead of `Response<T>`. In order for `Response<T>` to "work", the `T` type has to implement the `typespec_client_core::Model` trait, a trait with is Azure/TypeSpec-specific. I didn't want to force users to do that. Fortunately, with #1954 merged, we have another option. We can return `Response` and users can call `Response::into_json_body` to parse the JSON body to their own custom type which need only implement `serde::Deserialize`.
* Write operations also return `Response` since we expect them to have no body by default. If a user sets `enable_content_response_on_write`, they can parse the body with the same `Response::into_json_body` method.
* Database/Container-level operations still return full objects, even the `query_...` methods which could return projections. This aligns the Go SDK, but users still have the ability to parse _any_ response into their own type using `Response::into_json_body` (even if the type is `Response<DatabaseProperties>`).

In addition, I changed our doc examples to use `?` syntax (by having them be in `async fn doc() -> Result<(), Box<dyn std::error::Error>>` wrapper functions). It looks a lot cleaner that way, IMO, by removing all the noisy `.unwrap()` calls and matching more closely to what we expect users to do in their own applications.

Finally, I did need one change in the `azure_core_test` crate, because our live tests were broken by #1956 . I changed `azure_core_test::recorded::start` to return `Result<Option<Session>>` (instead of just `Result<Session>`). If the test is running in live mode, it'll return `Ok(None)`, to indicate that no test proxy session was established. I think this is the right way to handle this, but I'm open to feedback.

I'm not expecting action on this until the new year, because I know @heaths is out until the new year and the `azure_core_test` change will need his ✅ , but I wanted to open it up and get Cosmos folks looking at it.


kirankumarkolli:
How are changelog maintained for Rust? 
For other SDK's we have changelog.md part of the repo and part PR or later during release we keep refreshing them.

kirankumarkolli (approved):

analogrelay:
> How are changelog maintained for Rust?

Since we haven't made a release yet, there isn't really a Changelog. I expect that will start up after the first preview release. I do like how the other SDKs manage it, we can do something similar here.

Pilchie (approved):

heaths:
@analogrelay wrote,

> In addition, I changed our doc examples to use ? syntax (by having them be in async fn doc() -> Result<(), Box<dyn std::error::Error>> wrapper functions). It looks a lot cleaner that way, IMO, by removing all the noisy .unwrap() calls and matching more closely to what we expect users to do in their own applications.

I've been contemplating this across our samples as well. It's not typical, though, and does complicate things for our README.md samples because rustdoc-hidden "setup" code still renders in markdown regions; thus, we end up with discrepancies in our markdown vs. doc comments.

I'm not suggesting you remove it, but I did open #1989 to track this. /cc @RickWinter @JeffreyRichter

sdk/core/azure_core_test/src/recorded.rs
```diff
@@ -18,7 +18,14 @@ static TEST_PROXY: OnceCell<Result<Arc<Proxy>>> = OnceCell::const_new();
...
 /// The [Test Proxy](https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md) service will be started as needed.
 /// Every `#[recorded::test]` will call this automatically, but it can also be called manually by any other test e.g., those attributed with `#[tokio::test]`.
-pub async fn start(ctx: &TestContext, options: Option<ProxyOptions>) -> Result<Session> {
+///
+/// This function will return `Ok(None)` if the test is running in live mode and should not use the test proxy at all.
+pub async fn start(ctx: &TestContext, options: Option<ProxyOptions>) -> Result<Option<Session>> {
+    // Live tests don't use test-proxy.
+    if ctx.test_mode() == azure_core::test::TestMode::Live {
+        return Ok(None);
+    }
```
heaths, line 27: Then why declare a `TestContext` parameter at all?

sdk/cosmos/azure_data_cosmos/examples/cosmos/create.rs
```diff
@@ -32,6 +32,10 @@ pub enum Subcommands {
         /// The JSON of the new item.
         #[clap(long, short)]
         json: String,
+
+        /// If set, the updated item will be included in the response.
+        #[clap(long)]
```
heaths, line 37: Just occurred to me: wait, we're using an old enough clap that it's still using `#[clap]` instead of `#[arg]`? Or did they retain support for backcompat? `#[arg]` is current.

Also, if compatible with `#[arg]`, I thought for sure this required `#[arg(action = ArgAction::SetTrue)]`.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -208,24 +209,59 @@ impl ContainerClient {
...
+    ///
+    /// # Content Response on Write
+    ///
+    /// By default, the newly created item is *not* returned in the HTTP response.
+    /// If you want the new item to be returned, set the [`ItemOptions::enable_content_response_on_write`] option to `true`.
+    /// You can deserialize the returned item using [`Response::into_json_body`], like this:
+    ///
+    /// ```rust,no_run
+    /// # use azure_data_cosmos::{clients::ContainerClient, ItemOptions};
+    /// # use serde::{Deserialize, Serialize};
```
heaths, line 226: Any reason *not* to show these? Rust examples typically do show imports.

sdk/core/azure_core_test/src/recorded.rs
heaths: Oh, I see. You actually use it. But then there's no need to call `start()`. I guess I'm puzzled as to why this is needed. If we truly need it, fine; my questions are merely how we got to this point because it may alter my assertions about how this is used.

heaths (approved):
I see nothing blocking, but have some questions I'd like to hash out.

sdk/core/azure_core_test/src/recorded.rs
heaths: Never mind. We talked offline and this makes sense. No callers needed to change because, while I capture it in the macro, I don't use it yet (but will). For live tests that want to use it or when `AZURE_TEST_MODE=live`, we'll need some way of signaling that test-proxy wasn't started unnecessarily, so this will no doubt be useful.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
analogrelay: No particular reason! Happy to align with how other examples work.

sdk/cosmos/azure_data_cosmos/examples/cosmos/create.rs
analogrelay: >  wait, we're using an old enough clap that it's still using #[clap] instead of #[arg]? Or did they retain support for backcompat? #[arg] is current.

Just a habit on my end. It seems to have been retained for compat, but our version of clap absolutely supports `#[arg]`/`#[command]`.

> Also, if compatible with #[arg], I thought for sure this required #[arg(action = ArgAction::SetTrue)].

It's implied when you have a `bool` arguments: https://docs.rs/clap/latest/clap/_derive/index.html#arg-types

heaths (approved):

PR #1965, opened 2024-12-19
Create new response struct for AppendBlock, exposing ETag, blob_append_offset and blob_committed_block_count

### LINKED ISSUE #1964: ETag is not exposed on AppendBlock response type
ETag response header is not exposed on the AppendBlock response struct. Seems this is because currently the SDK re-uses the PutBlock response type, rather than a bespoke AppendBlock response type.

It is exposed on the API:
https://learn.microsoft.com/en-us/rest/api/storageservices/append-block?tabs=microsoft-entra-id#response-headers

LeonHartley:
Fixes #1964


LeonHartley:
@heaths @RickWinter @JeffreyRichter @ronniegeraghty ping :)

heaths (approved):
Thank you for your contribution!

LGTM, but keep in mind we have no plans to release subsequent versions from the `legacy` branch at this time. The Storage team is working on supported crates out of the `main` branch. /cc @vincenttran-msft to make sure proper support is included for this.

@LeonHartley you might consider [patching your dependency](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html) on this in your `Cargo.toml` manifest.

Issue #1972, opened 2025-01-07
Create simple mechanism to trigger codegeneration


RickWinter:
Something to the likes of 
`cargo generate`

heaths:
For context, I don't think this should be in a `build.rs` file. Even if it's conditionally triggered on an env var or something, anyone using our crates will require extra build time to compile a `build.rs` file to an executable, run it, and then generate source that likely wasn't even regenerated. You can add files and env vars to the watch list, but this still require extra build time and Rust is slow enough to build already.

TypeSpecs shouldn't be changing often enough to warrant this, IMO. But I could see providing a `cargo xtask` task or similar, or even just a build script under `eng/scripts` that not only takes care of a platform-agnostic and non-PATH-specific mechanism to executable `tsp-client` (e.g., Azure/azure-sdk-for-rust#1966 at the time of this writing is Windows- and PATH-specific), but could even take care of acquisition of `tsp-client` if easy enough to do in a platform-agnostic way.

heaths:
To note, this should be responsible for running `cargo fmt`. See #1981 for context.

heaths:
While it would increase initial build time slightly, it wouldn't affect downstream users if we emitted the `cargo::rerun-if-env-changed=AZURE_GENERATE_CODE` or something like that. See [docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-env-changed) for more info.

Still, this feels like driving a thumbtack with a sledgehammer. It's useful for us, yes...sometimes...but never for end users. And even if Azure SDK devs are building multiple crates - like the entire workspace by default - everything gets regenerated unless we use crate-specific env vars. That is, this has virulent behavior.

Instead, I think it's best if we just have a script under `eng/scripts` that regenerates when needed. It's not like services are often updating enough to make this transparent, and we're in a different ecosystem unlike most of our other languages that include this as part of their build command because, like .NET for example, the downstream devs don't get the `.csproj` - just the produced binary.

heaths:
> To note, this should be responsible for running `cargo fmt`. See [#1981](https://github.com/Azure/azure-sdk-for-rust/pull/1981) for context.

Now done in the emitter, as with most other languages.

heaths:
@RickWinter now that this should be down to just `tsp-client update`, do we still need this?

[...]

PR #1981, opened 2025-01-08
[Storage] Initial `v0.7.0` codegen

vincenttran-msft:
This is currently pointed to [vincenttran/blob-tsp-test2](https://github.com/Azure/azure-rest-api-specs/tree/vincenttran/blob-tsp-test2), which contains the necessary stop-gap changes from [feature/blob-tsp](https://github.com/Azure/azure-rest-api-specs/tree/feature/blob-tsp) to allow this typespec definition to be ingested.

1. Axing pageable operations (currently still NYI, expected behavior, emitter-stopping)
2. Adding a `crate-version` (this is required by the emitter)
3. Change the name from `storage_blob` to `azure_storage_blob` and combining with our handwritten wrapper code


vincenttran-msft:
> `storage_blob` doesn't exist in that scenario. You'd have a `azure_storage_blobs` (plural; same naming convention as every other language) crate where the clients live in the root. Again, like most other languages. Implementations should be idiomatic, yes, but naming should be idiomatically consistent (i.e. the same, ignoring casing differences). That means for "packages" (nupkgs, wheels, jars, crates, etc.) as well as namespaces/modules.

_Originally posted by @heaths in https://github.com/Azure/azure-sdk-for-rust/pull/1959#discussion_r1906196124_

Hi Heath, just wanted to touch base as to whether this is your expected directory structure? If so, I will look to merge our old initial [scaffolding](https://github.com/Azure/azure-sdk-for-rust/pull/1721) (which is the `azure_storage_blob` singular namespace) with the generated output. Thanks!

heaths:
Looks like it. Basically, you should follow the naming convention you use for other languages (but idiomatic, so underscores here). Looking at https://azure.github.io/azure-sdk/, I see both "Azure Storage Blobs" and "Azure Storage Blob". I guess pick which makes the most sense to you. Personally, I prefer the plural "blobs" since it works with Storage Blobs - not a blob.

vincenttran-msft:
> Looks like it. Basically, you should follow the naming convention you use for other languages (but idiomatic, so underscores here). Looking at https://azure.github.io/azure-sdk/, I see both "Azure Storage Blobs" and "Azure Storage Blob". I guess pick which makes the most sense to you. Personally, I prefer the plural "blobs" since it works with Storage Blobs - not a blob.

Hi Heath, that sounds good to me, I also started an email thread to loop in our PMs to see if they have any strong feelings or guidance wrt the naming and will land on something after all that feedback is received. Thanks!

vincenttran-msft:
This is ready for review, but is [failing CI](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4474046&view=logs&jobId=b72d910a-f771-5e0a-6940-a57bfa4b8e55&j=2026bc61-bde4-5818-dc95-0974ca714f8e&t=5b7bb566-c9dd-57fe-7552-ec130a74410f) for several duplicates of the same issue: 
```
error: field `container_name` is never read
  --> sdk/storage/azure_storage_blob/src/generated/clients/blob_blob_client.rs:16:16
   |
15 | pub struct BlobBlobClient {
   |            -------------- field in this struct
16 |     pub(crate) container_name: String,
   |                ^^^^^^^^^^^^^^
```

This issue will be resolved as a result of finding a solution for: https://github.com/Azure/typespec-rust/issues/218 , so if that solves in a timely manner we can wait for it, otherwise curious to hear what the path forward should be.

In the meantime, we will be continuing to work on fleshing out the get blob properties API in a separate branch based off this branch.

jhendrixMSFT:
The unused `container_name` field is a known issue. I've asked @catalinaperalta to look into it as it might simply be a matter of removing it from the tsp.

Also, it looks like you need to execute `cargo fmt` on the generated code.

heaths:
@jhendrixMSFT said,

> Also, it looks like you need to execute cargo fmt on the generated code.

Can we have the tool do that? It's not something authors should have to know/remember to do.

jhendrixMSFT:
I would really prefer to have the emitter just emit code and not have any dependency on 3rd party tools to perform any post-processing. This is why in the Go repo we have `build.go` files to execute the emitter and perform any post-processing (especially since a lot of SDKs have unique post-processing to modify generated code).

Is there a mechanism we can use to invoke some post-generation command, or do we need to wrap building in a script to do this?

catalinaperalta:
FYI I removed the containerName definition from client.tsp. This might change in a few days however as the client.tsp is restructured to output a better generated client design.

heaths:
> I would really prefer to have the emitter just emit code and not have any dependency on 3rd party tools to perform any post-processing. This is why in the Go repo we have `build.go` files to execute the emitter and perform any post-processing (especially since a lot of SDKs have unique post-processing to modify generated code).
> 
> Is there a mechanism we can use to invoke some post-generation command, or do we need to wrap building in a script to do this?

Using a `build.rs` file will increase build times for people that do not need to generate code, which we should not have the dev simply importing our libraries do because they'd need to install a bunch of TypeScript and TypeSpec libraries, Node.js, etc. Even if the `build.rs` file does nothing in `main` if not passed an env var or whatever as @RickWinter did in #1966, Rust still compiles it and executes the resulting binary even before it compiles the crate. This has a downstream impact.

We could have a script or `cargo xtask` or something do this, but I'm curious to know why you call `cargo fmt` a third-party tool? It's part of the rust toolchain itself and automatically installed whenever you run a `cargo` command based on our root `rust-toolchain.toml`. Or did you just mean to say it's external to the TypeSpec toolchain and rust emitter?

jhendrixMSFT:
My mental model is that any process external to the TypeSpec toolchain/emitter is a "3rd party tool". Maybe a "hidden dependency" is a better way to phrase it. When you `pnpm install` the emitter and its dependencies, this doesn't install the Rust toolchain. So, how does one know that it's a requirement to generate code? And if it's not present, what do we do? Fail? Warn/ignore?

Agreed that `build.rs` isn't an appropriate mechanism. I'd like to explore other possibilities before deciding that we should incorporate `cargo fmt` in the emitter.

heaths:
We have #1972 tracking that currently.

eng/emitter-package.json
```diff
@@ -1,11 +1,12 @@
 {
   "main": "dist/src/index.js",
   "dependencies": {
-    "@azure-tools/typespec-rust": "0.3.0"
+    "@azure-tools/typespec-rust": "0.6.0"
```
jhendrixMSFT, line 4: `v0.7.0` is now available and it fixes the Clippy (and other) issues once you migrate to the latest tsp. You'll also need to migrate the below to `0.50.0` and `0.64.0` respectively.

vincenttran-msft: Great, thanks! Regenerated!

jhendrixMSFT:
Can you please resolve the merge conflicts?

vincenttran-msft:
> Can you please resolve the merge conflicts?

Definitely, resolving that right now. Spinning on generating the lockfile.

vincenttran-msft:
Alrighty, I have resolved all the merge conflicts and cspell issues. The remaining issues are with the "[Run source analysis](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4488806&view=logs&jobId=ea0d377c-eec0-5cc9-d7d7-174dad1e3978&j=ea0d377c-eec0-5cc9-d7d7-174dad1e3978&t=fbc75dc2-62a5-5bd9-4d01-db31ae5249f2)" step (sample except):

```
error: unresolved link to `Update`
   --> sdk/storage/azure_storage_blob/src/generated/clients/blob_blob_client.rs:119:10
    |
119 |     /// [Update] The Lease Blob operation establishes and manages a lock on a blob for write and delete operations.
    |          ^^^^^^ no item named `Update` in scope
    |
    = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
```

Any insight on what is the correct path forward? This _should_ be our final blocker 😄

jhendrixMSFT:
It's coming from the doc comments in the tsp. [example](https://github.com/Azure/azure-rest-api-specs/blob/feature/blob-tsp/specification/storage/Microsoft.BlobStorage/routes.tsp#L414)

Presumably the doc comments need to be updated. At least I think that's the preferred solution.

jhendrixMSFT:
I updated the tsp in the feature branch to clean up these doc comments.

jalauzon-msft (approved):

sdk/storage/azure_storage_blob/Cargo.toml
```diff
@@ -11,6 +11,10 @@ rust-version.workspace = true
 
 [dependencies]
 azure_storage_common.workspace = true
+azure_core = { workspace = true, features = ["xml"] }
+serde = { workspace = true }
+time = { workspace = true }
+typespec_client_core = { workspace = true, features = ["derive"] }
```
jhendrixMSFT, line 17: @heaths long-term do we expect/want Azure SDKs to have a direct dependency on this? There are a few things in `typespec_client_core` that haven't been re-exported from `azure_core` (e.g. the `create_enum` macros).

jhendrixMSFT (approved):

sdk/storage/azure_storage_blob/Cargo.toml
heaths: I don't see any harm, but I don't have a strong opinion either way. @RickWinter @ronniegeraghty @JeffreyRichter what do you think? Should we necessarily re-export everything in `typespec_client_core` from `azure_core`? Hmm, having just typed that I'm actually thinking we shouldn't. It's an unnecessary burden. I only did it originally for migration and wanted to get rid of them (mostly) eventually. Since we generate and expect very little hand-coding, it wouldn't be hard to rename those crates later should we need to.

I wouldn't hold this PR up for that, though. Something we should discuss further offline.

.vscode/cspell.json
```diff
@@ -39,22 +46,35 @@
...
+    "restype",
     "rustup",
     "seekable",
     "servicebus",
     "stylesheet",
     "typespec",
+    "undelete",
+    "undelete",
+    "Undelete",
+    "Undelete",
```
heaths, line 74: Please remove duplicates. cSpell is also case-insensitive so no need to store both variations.

.vscode/cspell.json
```diff
@@ -39,22 +46,35 @@
     "eventhub",
     "eventhubs",
     "hmac",
+    "incrementalcopy",
```
heaths, line 49: I'd rather not store Storage-specific things in here. Instead, define a `sdk/storage/.dict.txt` file and add an entry in this file like cosmos and identity did. You can put all the Storage-specific things in there you want.

Mainly, the reason is because some of these may very well be misspellings in other contexts. What's here was intended to be more general items. Same with other repos.

heaths (approved):
Not blocking, but please either make the updates I suggested in this PR or a subsequent one.

.vscode/cspell.json
heaths: And, to clarify, I think "undelete" is general purpose enough to stay here. Several of these are. Use your best judgement.

Issue #1982, opened 2025-01-08
[TypeSpec] Date-time parsing assumes UTC format


jhendrixMSFT:
Examples:
https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/typespec/typespec_client_core/src/date/mod.rs#L67
https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/typespec/typespec_client_core/src/date/mod.rs#L88

TypeSpec supports date-times in UTC and with a time zone offset. If the helpers can't be parameterized to work with both in serde, then perhaps we'll need variants that don't enforce UTC.

While we don't have any tests/services that require a date-time with an offset, we'll need to support this at some point to align with TypeSpec.

heaths:
I think this only applies to RFC 1123 date-times, which, at least for Azure services, must be GMT: https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#http-header-date-values. That said, I appreciate that for TypeSpec we probably need to support that and work for RFC 1123 dates will be needed.

RFC 3339 date-times support offsets already, though we do have some bugs in the code that don't properly capture the UTC converted date before serializing. That said, I don't know if we should convert to UTC from a local time offset when roundtripping. @JeffreyRichter or @johanste, should we?

JeffreyRichter:
It's safer not to always convert to UTC. Any language date-time type should have a function that easiliy converts to UTC if the customer's application desires this.

jhendrixMSFT:
TypeSpec makes the distinction. i.e. it has `utcDateTime` and `offsetDateTime` (see [docs](https://typespec.io/docs/language-basics/built-in-types/#date-and-time-types)). So, for `utcDateTime` I believe it's expected that we convert to UTC when serializing and deserialize as UTC (I would assume in this case the value sent by the service is already in UTC, else should we convert it?).

JeffreyRichter:
OK, UTC assumes UTC - I can accept this.
No, we should not convert it. If the service sends a non-UTC time, that is a service bug as it is not living up to its contract.
Implicitly converting it hides the bug. I remember the SDK team actually did have an issue related to this be we did auto-convert and this caused a problem. I can't remember the details though.

heaths:
[...]

PR #1983, opened 2025-01-09
Fix conversions to UTC

jhendrixMSFT:
The conversions to UTC didn't capture the result, having no effect. Added tests to cover the gaps.


RickWinter (approved):

heaths (changes requested):
Does not cover enough. RFC1123 also has problems, and we may not want to always convert. See discussion in #1982. I have a PR where I'm adding tests and will take into account what we need to do correctly for these variants.

jhendrixMSFT:
Closing as @heaths will be fixing this and more.

PR #1986, opened 2025-01-09
Do not convert dates

### LINKED ISSUE #1982: [TypeSpec] Date-time parsing assumes UTC format

heaths:
Technically, we weren't because we weren't capturing the return of `to_offset()`, but this cleans up the code and adds some tests that caused us to miss this.

Resolves #1982

This also renames RFC 1123 functions to 7231, which obsolesces 1123 effectively; though, is itself obsolesced by 9110. We do refer to 7231 in both [TypeSpec](https://typespec.io/docs/libraries/http/encoding/#utcdatetime-and-offsetdatetime) and [REST API Guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#http-header-date-values) already, though.

### DIFF
--- a/sdk/core/azure_core/src/request_options/if_source_modified_since_condition.rs
+++ b/sdk/core/azure_core/src/request_options/if_source_modified_since_condition.rs
@@ -24,7 +24,7 @@ impl Header for IfSourceModifiedSinceCondition {
     fn value(&self) -> headers::HeaderValue {
         match self {
             IfSourceModifiedSinceCondition::Modified(date)
-            | IfSourceModifiedSinceCondition::Unmodified(date) => date::to_rfc1123(date).into(),
+            | IfSourceModifiedSinceCondition::Unmodified(date) => date::to_rfc7231(date).into(),
         }
     }
 }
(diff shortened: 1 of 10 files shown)


heaths:
This is a minimal change for now. We should add more test cases when we do the full bevy of fixes.

jhendrixMSFT (approved):

heaths:
@JeffreyRichter can you have a look as well?

JeffreyRichter (approved):
Switching from 1123 to 7231 now is fine. But of course, we can't chase RFC numbers as the years go on.

heaths:
> Switching from 1123 to 7231 now is fine. But of course, we can't chase RFC numbers as the years go on.

Agreed. And given TypeSpec documents "rfc7231" as a supported `@encode` parameter, I'd say it's locked in within reason. If we ever do update our REST API Guidelines, we should probably retain some reference to 7231 e.g., "dates must comply with RFC 9110 (replaces RFC 7231)...". Keeps things orderly and more references more discoverable.

PR #1987, opened 2025-01-10
Fix clippy 0.1.84 lints

heaths:


heaths:
Fixes build breaks we're getting now.

LarryOsterman (approved):

Issue #1967, opened 2024-12-20
Identity README


ronniegeraghty:
The `README.md` file for the Azure Identity package should follow our template for service library READMEs 
Example: [.NET's Azure Identity README](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/README.md)

ronniegeraghty:
Completed

PR #1990, opened 2025-01-14
Updates to the azure_identity README

ronniegeraghty:
Frist draft for the Identity `README.md`. 
Copied from Python identity `README.md`, updating syntax, examples and links. 
Pretty much went through the credentials their `README.md` referenced and checked if we had that cred implemented yet.

Still many TODOs and ?(questions) in the doc.


ronniegeraghty:
It looks like the checks are failing because in the `README.md` I put an example that uses `azure_data_cosmos::CosmosClient`. I believe the check is trying to run the example, but the identity crate workspace doesn't include the cosmos crate. Is that right? Not sure how to resolve this. Looks like the old `README.md` uses a `StorageAccountClient` in its example but it had to create and `impl` the struct in the example. I feel like that adds a lot of code to the example that users won't and shouldn't bring into their projects.

ronniegeraghty:
Fixes #1967

sdk/identity/azure_identity/README.md
```diff
@@ -1,77 +1,425 @@
...
-);
+let client = CosmosClient::new("https://myaccount.documents.azure.com/", credential, None).unwrap();
+```
+
+<!-- TODO: update this section on enabling interactive auth if/when it's configurable in the Rust library>
+#### Enable interactive authentication with `DefaultAzureCredential`
+
+By default, interactive authentication is disabled in `DefaultAzureCredential` and can be enabled with a keyword argument:
+
+```python
```
LarryOsterman, line 87: Intentional `python` reference?

sdk/identity/azure_identity/README.md
```diff
@@ -1,77 +1,425 @@
...
+When enabled, `DefaultAzureCredential` falls back to interactively authenticating via the system's default web browser when no other credential is available.
+</!-->
+
+<!-- TODO: update this section on specifying a user-assigned managed identity if/when it's configurable in the Rust library>
+#### Specify a user-assigned managed identity with `DefaultAzureCredential`
+
+Many Azure hosts allow the assignment of a user-assigned managed identity. To configure `DefaultAzureCredential` to authenticate a user-assigned managed identity, use the `managed_identity_client_id` keyword argument:
+
+```python
+DefaultAzureCredential(managed_identity_client_id=client_id)
```
LarryOsterman, line 100: I think that should be `DefaultAzureCredential::new()` or `DefaultAzureCredential::new().unwrap()`

sdk/identity/azure_identity/README.md
```diff
@@ -1,77 +1,425 @@
...
+- [Azure Arc](https://learn.microsoft.com/azure/azure-arc/servers/managed-identity-authentication)
+- [Azure Cloud Shell](https://learn.microsoft.com/azure/cloud-shell/msi-authorization)
+- [Azure Kubernetes Service](https://learn.microsoft.com/azure/aks/use-managed-identity)
+- [Azure Service Fabric](https://learn.microsoft.com/azure/service-fabric/concepts-managed-identity)
+- [Azure Virtual Machines](https://learn.microsoft.com/entra/identity/managed-identities-azure-resources/how-to-use-vm-token)
+- [Azure Virtual Machines Scale Sets](https://learn.microsoft.com/entra/identity/managed-identities-azure-resources/qs-configure-powershell-windows-vmss)
+
+### Examples
+
+These examples demonstrate authenticating `SecretClient` from the [`azure-keyvault-secrets`](https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/keyvault/azure-keyvault-secrets) library with `ManagedIdentityCredential`.
```
LarryOsterman, line 127: Python link

sdk/identity/azure_identity/README.md
```diff
@@ -1,77 +1,425 @@
...
+
+### Service principal with certificate
+
+|Variable name|Value|Required
+|-|-|-
+|`AZURE_CLIENT_ID`|ID of a Microsoft Entra application|X
+|`AZURE_TENANT_ID`|ID of the application's Microsoft Entra tenant|X
+|`AZURE_CLIENT_CERTIFICATE_PATH`|path to a PEM or PKCS12 certificate file including private key|X
+|`AZURE_CLIENT_CERTIFICATE_PASSWORD`|password of the certificate file, if any|
+|`AZURE_CLIENT_SEND_CERTIFICATE_CHAIN`|If `True`, the credential sends the public certificate chain in the x5c header of each token request's JWT. This is required for Subject Name/Issuer (SNI) authentication. Defaults to False. There's a [known limitation](https://github.com/Azure/azure-sdk-for-python/issues/13349) that async SNI authentication isn't supported.|
```
LarryOsterman, line 286: Probably want to remove the known limitation comment if it isn't only a python issue.

LarryOsterman:
Note to self: The markdown is *NOT* the content - look at the rendered output before leaving feedback.

sdk/identity/azure_identity/README.md
```diff
@@ -1,77 +1,425 @@
...
+[default_cred_ref]: <!-- TODO: When Docs.rs page for DefautlAzureCredential ref docs are available -->
+[entraid_doc]: https://learn.microsoft.com/entra/identity/
+[identity_api_docs]: https://docs.rs/azure_identity/latest/azure_identity/
+[managed_id_cred_ref]: <!-- TODO: When Docs.rs page for ImdsManagedIdentityCredential ref docs are available -->
+[package]: https://crates.io/crates/azure_identity
+[source]: https://github.com/Azure/azure-sdk-for-rust/tree/main/sdk/identity/azure_identity
+[token_cred_ref]: <!-- TODO: When Docs.rs page for TokenCredential trait ref docs are available -->
+[workload_id_cred_ref]: <!-- TODO: When Docs.rs page for WorkloadIdentityCredential ref docs are available -->
+
+<!-- LINKS from Python Example for reference>
```
heaths, line 389: We are all these here? We can always copy what we need later and update the URLs. That's the great thing about source control: we never really lose anything anyway.

sdk/identity/azure_identity/README.md
```diff
@@ -1,77 +1,425 @@
...
+Install the Azure Identity library for Rust with cargo:
+
+```bash
+cargo add azure_identity
 ```
 
-## Design
+### Prerequisites
+
+* An [Azure subscription][azure_sub].
```
heaths, line 19: Tip: if you keep referencing links this way with (mostly) the same name, you can just use `[Azure subscription]` and define the link as:

```markdown
[Azure subscription]: ...
```

Saves you a lot of typing and looking up what you called it.

heaths: I see you mention `[Azure CLI][azure_cli]` a lot. Imagine just defining that as `[Azure CLI]` and having to type only that over and over again.

sdk/identity/azure_identity/README.md
```diff
@@ -1,77 +1,425 @@
...
+
+When debugging and executing code locally, it's typical for developers to use their own accounts for authenticating calls to Azure services. The Azure Identity library supports authenticating through developer tools to simplify local development.
+
+#### Authenticate via the Azure CLI
+
+`DefaultAzureCredential` and `AzureCliCredential` can authenticate as the user signed in to the [Azure CLI][azure_cli]. To sign in to the Azure CLI, run `az login`. On a system with a default web browser, the Azure CLI launches the browser to authenticate a user.
+
+When no default browser is available, `az login` uses the device code authentication flow. This flow can also be selected manually by running `az login --use-device-code`.
+
+#### Authenticate via the Azure Developer CLI
```
heaths, line 32: We actually don't support this yet. But I wouldn't comment this out. We can copy it later. That could just be misleading for those reading the actual markdown file as opposed to the rendered view on GitHub.

sdk/identity/azure_identity/README.md
```diff
@@ -1,77 +1,425 @@
...
+
+### Authenticate with `DefaultAzureCredential`
+
+More details on configuring your environment to use `DefaultAzureCredential` can be found in the class's [reference documentation][default_cred_ref].
+
+This example demonstrates authenticating the `CosmosClient` from the [azure_data_cosmos][azure_data_cosmos] library using `DefaultAzureCredential`.
 
 ```rust
-use azure_core::credentials::TokenCredential;
+use azure_data_cosmos::CosmosClient;
```
heaths, line 75: Once we have `SecretClient` ready, we should use that. It's by far the cheapest resource to use if people want to follow along and probably one of the most simple clients.

sdk/identity/azure_identity/README.md
```diff
@@ -1,77 +1,425 @@
...
-#     }
-# }
-# struct SecretClient;
-# impl SecretClient {
-#     fn new(_endpoint: &str, _credential: Arc<dyn TokenCredential>, _options: Option<ClientOptions>) -> Result<Arc<Self>> {
-#         Ok(Arc::new(SecretClient))
-#     }
-# }
 
 let credential = DefaultAzureCredential::new().unwrap();
```
heaths, line 78: In our READMEs, I think we should show best practices as much as possible. But these READMEs also need to be compiled as documented: https://azure.github.io/azure-sdk/rust_introduction.html#rust-documentation-readme (or maybe that was already in there).

Thus, this whole section might actually be:

```rust
use azure_data_cosmos::CosmosClient;
use azure_identity::DefaultAzureCredential;

    let credential = DefaultAzureCredential::new()?;
    let client = CosmosClient::new("https://myaccount.documents.azure.com/", credential.clone(), None)?;
    // ...
}
```

sdk/identity/azure_identity/README.md
```diff
@@ -1,77 +1,425 @@
...
-    None,
-);
-let secret_client = SecretClient::new("https://myvault.keyvault.azure.net",
-    credential.clone(),
-    None,
-);
+let client = CosmosClient::new("https://myaccount.documents.azure.com/", credential, None).unwrap();
+```
+
+<!-- TODO: update this section on enabling interactive auth if/when it's configurable in the Rust library>
```
heaths, line 82: Let's get rid of all this commented code. It serves no purpose and we can always copy it later. It's not uncommon for developers to read the raw markdown. When I'm working in a repo and have it cloned locally, I do.

heaths (changes requested):

sdk/identity/azure_identity/README.md
ronniegeraghty: Updated `[Azure subscription]` & `[Azure CLI]` links and a few more.

sdk/identity/azure_identity/README.md
ronniegeraghty: It's gone.

sdk/identity/azure_identity/README.md
ronniegeraghty: Updated example

sdk/identity/azure_identity/README.md
ronniegeraghty: Removed the section.

sdk/identity/azure_identity/README.md
ronniegeraghty: Removed

sdk/identity/azure_identity/README.md
ronniegeraghty: Updated example to use Key Vault Secrets client.

sdk/identity/azure_identity/README.md
```diff
@@ -1,77 +1,132 @@
...
+### DefaultAzureCredential
+
+`DefaultAzureCredential` simplifies authentication while developing apps that deploy to Azure by combining credentials used in Azure hosting environments with credentials used in local development. For more information, see [DefaultAzureCredential overview][dac_overview].
+
+#### Continuation policy
+
+As of version 1.14.0, `DefaultAzureCredential` attempts to authenticate with all developer credentials until one succeeds, regardless of any errors previous developer credentials experienced. For example, a developer credential may attempt to get a token and fail, so `DefaultAzureCredential` will continue to the next credential in the flow. Deployed service credentials stop the flow with a thrown exception if they're able to attempt token retrieval, but don't receive one. Prior to version 1.14.0, developer credentials would similarly stop the authentication flow if token retrieval failed, but this is no longer the case.
```
RickWinter, line 46: Version is wrong. We also only ahve the later behavior.

sdk/identity/azure_identity/README.md
```diff
@@ -1,77 +1,132 @@
...
+
+The following examples are provided:
+<!-- no toc -->
+* [Authenticate with DefaultAzureCredential](#authenticate-with-defaultazurecredential "Authenticate with DefaultAzureCredential")
+
+### Authenticate with `DefaultAzureCredential`
+
+More details on configuring your environment to use `DefaultAzureCredential` can be found in the class's [reference documentation][default_cred_ref].
+
+This example demonstrates authenticating the `SecretClient` from the [azure_security_keyvault_secrets] library using `DefaultAzureCredential`.
```
RickWinter, line 60:
```suggestion
This example demonstrates authenticating the `SecretClient` from the [azure_security_keyvault_secrets] crate using `DefaultAzureCredential`.
```

RickWinter (commented):
Some updates needed. Lets get this checked in today and continue to improve on it.

sdk/identity/azure_identity/README.md
ronniegeraghty: Updated version number to next release version of identity.

heaths (approved):

PR #1993, opened 2025-01-15
`get_blob_properties` and first pass at testing infra

vincenttran-msft:


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -1,2 +1,108 @@
...
+            per_try_policies.push(Arc::new(oauth_token_policy) as Arc<dyn Policy>);
+            options.client_options.per_call_policies = per_try_policies;
+        }
+
+        let client = GeneratedBlobClient::new(
+            &endpoint.clone(),
+            credential.clone().unwrap(), // This unwrap() is temporary- this may be a straight passthrough dependent on how we handle auth-less
+            container_name.clone(),
+            Some(options),
+        )?;
```
vincenttran-msft, line 63: This is something that changed majorly between the old copy of the generated code and the present copy. Does this mean the MVP is now scoped only to authenticated workflows (and specifically, DAC)? Or should we engage with Azure SDK how their new generated code scheme is expected to handle auth-less.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -1,2 +1,108 @@
...
+        })
+    }
+
+    pub async fn get_blob_properties(
+        &self,
+        options: Option<BlobBlobClientGetPropertiesOptions<'_>>,
+    ) -> Result<Response<()>> {
+        self.client
+            .get_blob_blob_client()
+            .get_properties(self.container_name.clone(), self.blob_name.clone(), options)
```
vincenttran-msft, line 81: This is something I have an active work item regarding (having `container_name` and `blob_name` on every API) related to our swagger transform to handle it in autorest.

However, this brings up another friction point: Even after that issue is resolved, now the generated API won't have the `blob` name since the struct doesn't currently have this:

```rust
pub struct BlobBlobClient {
    pub(crate) endpoint: Url,
    pub(crate) pipeline: Pipeline,
    pub(crate) version: String,
}
```

Meaning that it now either has to come in as input from the handwritten caller layer, or we need to shift towards the abstraction having all the data necessary (i.e. these "Blob" level structs should have `blob`, and of course omit as we go up in abstraction)

sdk/storage/azure_storage_blob/tests/framework/test_account.rs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+#[derive(Default)]
+pub struct TestAccountOptions {
+    pub allow_invalid_certificates: Option<bool>,
+}
+static TRACING: Once = Once::new();
+
+// Assuming we will want to change this over to DefaultAzureCredential
+impl TestAccount {
```
vincenttran-msft, line 29: I am assuming the way we will use this is sort of have it set everything up, and so all values used for testing should be sourced from the `TestAccount` so that it is standardized and will set us up for success in terms of sanitization etc. and matching CI account configuration.

sdk/storage/azure_storage_blob/tests/framework/test_account.rs
```diff
@@ -0,0 +1,53 @@
...
+
+/// Represents a Azure Storage account for testing purposes.
+///
+/// A [`TestAccount`] serves two main purposes:
+/// * Managing connection information to make it easy to connect to the account the tests are using.
+/// * Provide a unique ID to each test and automatic clean-up of resources.
+pub struct TestAccount {
+    pub context_id: String,
+    endpoint: String,
+    key: Secret,
```
vincenttran-msft, line 18: DAC?

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -1,2 +1,108 @@
...
+        // Conditionally add authentication if provided
+        if credential.is_some() {
+            let oauth_token_policy = BearerTokenCredentialPolicy::new(
+                credential.clone().unwrap(),
+                ["https://storage.azure.com/.default"],
+            );
+            let mut per_try_policies = options.client_options.per_call_policies.clone();
+            per_try_policies.push(Arc::new(oauth_token_policy) as Arc<dyn Policy>);
+            options.client_options.per_call_policies = per_try_policies;
+        }
```
vincenttran-msft, line 56: This is an example of the previous hoops we had to jump through.
TODO: Refactor to match work done for `per_call_policies`

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: For now- we are going to assume thie MVP is scoped purely to authenticated workflows.
We can change our `credential` field to non-optional to match the passthrough for generated.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: Going to mark as resolve to avoid confusion-- this issue should go away entirely after we resolve the `container_name` and `blob_name` discussion.

sdk/storage/azure_storage_blob/tests/framework/test_account.rs
vincenttran-msft: We likely only need `account_name` here for Storage's purposes

sdk/storage/azure_storage_blob/tests/test_blob_client.rs
```diff
@@ -0,0 +1,39 @@
...
+
+#[cfg(test)]
+mod tests {
+
+    use azure_storage_blob::models::blob_properties;
+
+    use super::*;
+
+    #[recorded::test(live)]
+    async fn test_get_blob_properties() {
```
vincenttran-msft, line 20: Current blocker: I want to make this signature look like this and return:
`async fn test_get_blob_properties() -> Result<BlobProperties, Box<dyn Error>>`

However this is the error:
```
error[E0277]: the trait bound `azure_storage_blob::models::blob_properties::BlobProperties: Termination` is not satisfied
   --> sdk\storage\azure_storage_blob\tests\test_blob_client.rs:19:44
    |
18  |     #[tokio::test]
    |     -------------- in this procedural macro expansion
19  |     async fn test_get_blob_properties() -> Result<BlobProperties, Box<dyn Error>> {
    |                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Termination` is not implemented for `azure_storage_blob::models::blob_properties::BlobProperties`, which is required by `Result<azure_storage_blob::models::blob_properties::BlobProperties, Box<dyn std::error::Error>>: Termination`
```

Aside from that, this is currently working and we have a proof of concept model. 

Current output: 
```
---- tests::test_get_blob_properties stdout ----
Unknown key encountered: available
BlobProperties { blob_type: "BlockBlob", content_length: "11", date: "Thu, 16 Jan 2025 01:53:28 GMT", version: "2025-01-05", lease_status: "unlocked", access_tier: "Hot", access_tier_inferred: "true", client_request_id: "3da01868-17f2-41c9-a69d-158571c28c77", last_modified: "Wed, 15 Jan 2025 23:28:41 GMT", server: "Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0", content_md5: "XrY7u+Ae7tCTyyK7j1rNww==", accept_ranges: "bytes", request_id: "ba66921b-201e-0051-3bb9-673e1c000000", content_type: "text/plain", server_encrypted: "true", etag: "\"0x8DD35BC58886E5E\"", lease_state: "", creation_time: "Wed, 15 Jan 2025 23:28:41 GMT", vary: "Origin" }
```

sdk/storage/azure_storage_blob/tests/test_blob_client.rs
```diff
@@ -0,0 +1,39 @@
...
+use std::error::Error;
+
+#[cfg(test)]
+mod tests {
+
+    use azure_storage_blob::models::blob_properties;
+
+    use super::*;
+
+    #[recorded::test(live)]
```
vincenttran-msft, line 19: Note: This causes tests to be skipped (live tests skipped by default) when running the usual `cargo test`

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -14,56 +13,48 @@ use azure_core::{
...
-impl BlobClient<Unset> {
+impl BlobClient {
     pub fn new(
         endpoint: String,
         container_name: String,
         blob_name: String,
-        credential: Option<Arc<dyn TokenCredential>>,
-        mut options: Option<BlobClientOptions>,
+        credential: Arc<dyn TokenCredential>,
+        mut options: BlobClientOptions,
```
vincenttran-msft, line 33: Keep this as optional

sdk/storage/azure_storage_blob/tests/framework/test_account.rs
heaths: Why are you even checking in a separate test framework? We have recorded tests - like with every language - coming soon.

sdk/storage/azure_storage_blob/tests/test_blob_client.rs
heaths: Yes. That's by design and documented because they are live-only tests. You need to pass `AZURE_TEST_MODE=live cargo test` instead (bash example).

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
heaths: Why is this `mut`? That burdens callers unnecessarily. Per the guidelines, this must be an `Option<BlobClientOptions>` and in your body you do something like:

```rust
```

If options aren't optional, they shouldn't be in your client *options*.

heaths (changes requested):

vincenttran-msft:
Closed to prevent confusion- but will take feedback forward to next PR (which will be targetting main as we are getting an initial copy in soon 🚀 )

sdk/storage/azure_storage_blob/tests/framework/test_account.rs
vincenttran-msft: Thanks for the clarification here Heath! This was heavily cribbed from [cosmosdb](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/cosmos/azure_data_cosmos/tests/framework/test_account.rs), but good to know this is not necessary in our context.

sdk/storage/azure_storage_blob/tests/framework/test_account.rs
vincenttran-msft: N/A

PR #1997, opened 2025-01-15
Periodically read workload identity token from file

### LINKED ISSUE #1739: Azure Workload Identity - Expired token
We are running Azure Identity v0.20.0 and getting this error for a longer-running process in AKS. We are using Workload Identity with our deployments. 

```
 azure_identity::federated_credentials_flow: rsp_body == "{
 "error":"invalid_client",
 "error_description":"AADSTS700024: Client assertion is not within its valid time range.  
   Current time: 2024-08-05T10:07:04.2542067Z, assertion 
   valid from 2024-08-02T18:52:57.0000000Z, expiry time of assertion 2024-08-02T19:52:57.0000000Z.  Review the documentation at https://learn.microsoft.com/entra/identity-platform/certificate-credentials . Trace ID: XXXX-XXXX-XXXX-XXXX-XXXX Correlation ID: XXXX-XXXX-XXXX-XXXX-XXXX Timestamp: 2024-08-05 10:07:04Z",
   "error_codes":[700024],
   "timestamp":"2024-08-05 10:07:04Z", "trace_id":"XXXX-XXXX-XXXX-XXXX-XXXX",
   "correlation_id": "XXXX-XXXX-XXXX-XXXX-XXXX","error_uri": "https://login.microsoftonline.com/error?code=700024"
   }"
``` 

The workload identity token is not being refreshed as it is supposed to be.

heaths:
Fixes #1739 similar to other languages by reading the file every 10 minutes pointed to by AZURE_FEDERATED_TOKEN_FILE.

### DIFF
--- a/sdk/identity/azure_identity/src/credentials/workload_identity_credentials.rs
+++ b/sdk/identity/azure_identity/src/credentials/workload_identity_credentials.rs
@@ -8,9 +8,11 @@ use azure_core::{
     error::{ErrorKind, ResultExt},
     Error, HttpClient, Url,
 };
+use futures::channel::oneshot;
 use std::{
-    str,
+    fs, str,
     sync::Arc,
+    thread,
     time::{Duration, Instant},
 };
 use time::OffsetDateTime;
@@ -193,11 +195,22 @@ impl Token {
                 let now = Instant::now();
                 let cache = cache.upgradable_read().await;
                 if now - cache.last_read > TIMEOUT {
-                    let token = std::fs::read_to_string(path)
-                        .with_context(ErrorKind::Credential, || {
-                            format!("failed to read federated token from file {}", path)
-                        })?;
+                    // TODO: https://github.com/Azure/azure-sdk-for-rust/issues/2002
+                    let path = path.clone();
+                    let (tx, rx) = oneshot::channel();
+                    thread::spawn(move || {
+                        let token = fs::read_to_string(&path)
+                            .with_context(ErrorKind::Credential, || {
+                                format!("failed to read federated token from file {}", &path)
+                            });
+                        tx.send(token)
+                    });
+
                     let mut write_cache = RwLockUpgradableReadGuard::upgrade(cache).await;
+                    let token = rx.await.map_err(|err| {
+                        azure_core::Error::full(ErrorKind::Io, err, "canceled reading certificate")
+                    })??;
+
                     write_cache.token = Secret::new(token);
                     write_cache.last_read = now;
 
--- a/sdk/identity/azure_identity/src/credentials/workload_identity_credentials.rs
+++ b/sdk/identity/azure_identity/src/credentials/workload_identity_credentials.rs
@@ -2,12 +2,17 @@
 // Licensed under the MIT License.
 
 use crate::{credentials::cache::TokenCache, federated_credentials_flow, TokenCredentialOptions};
+use async_lock::{RwLock, RwLockUpgradableReadGuard};
 use azure_core::{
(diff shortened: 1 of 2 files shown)


sdk/identity/azure_identity/src/credentials/workload_identity_credentials.rs
```diff
@@ -155,3 +152,60 @@ impl TokenCredential for WorkloadIdentityCredential {
...
+    async fn secret(&self) -> azure_core::Result<String> {
+        match self {
+            Self::Value(secret) => Ok(secret.secret().into()),
+            Self::File { path, cache } => {
+                const TIMEOUT: Duration = Duration::from_secs(600);
+
+                let now = Instant::now();
+                let cache = cache.upgradable_read().await;
+                if now - cache.last_read > TIMEOUT {
+                    let token = std::fs::read_to_string(path)
```
conradludgate, line 196: This would in theory block the executor

arpad-m: Same goes here: https://github.com/Azure/azure-sdk-for-rust/blob/6d89f3acee27d7c26bca238de36fe91416690c1d/sdk/identity/azure_identity/src/credentials/client_certificate_credentials.rs#L311

I suppose the most fitting requirements of being executor agnostic would be if there was a trait for all sorts of I/O, including launching processes (#1652) which then has a default implementation via tokio, but also can have implementations via other executors (or even blocking ones).

But let's not have perfect be the enemy of good. I'd say it's an orthogonal concern.

heaths: I did consider this and was looking for existing abstractions that don't require a runtime - not finding any, which isn't actually surprising. It'd be great if more abstractions where in/under `std::future` so that we, the library developers, aren't always having to define our own incompatible (with other libs) abstractions, but it's likely what we'll have to do.

I've created #2002 but would add that this is no worse than what the code was doing before. Not ideal, I agree, but we'll need to create abstractions not just for the HTTP stack (and AMQP, as we've done), but processes as noted in #1652 and file system as noted in #2002 now.

heaths: We do already depend on the `futures` crate so that should be fine in the short term.

sdk/identity/azure_identity/src/credentials/workload_identity_credentials.rs
```diff
@@ -155,3 +154,71 @@ impl TokenCredential for WorkloadIdentityCredential {
...
+                    // TODO: https://github.com/Azure/azure-sdk-for-rust/issues/2002
+                    let path = path.clone();
+                    let (tx, rx) = oneshot::channel();
+                    thread::spawn(move || {
+                        let token = fs::read_to_string(&path)
+                            .with_context(ErrorKind::Credential, || {
+                                format!("failed to read federated token from file {}", &path)
+                            });
+                        tx.send(token)
+                    });
```
RickWinter, line 207: If the reader thread has an error it never restarts?

heaths: If the caller retries, `get_token()` would be called again. I don't think we should retry file IO operations in general. Outside of file locking - which is far more common on Windows, especially NTFS (oh, the stories from 25 years of deployment technology I can share) - there should be no good reasons to expect retrying again would succeed and we waste time. Let the upstream caller decide how to handle it.

RickWinter (approved):

PR #1998, opened 2025-01-16
Export Bytes from typespec_client_core

jhendrixMSFT:


LarryOsterman (approved):

heaths (approved):
I was kinda hoping to avoid exporting more third-party types, but perhaps there's no fighting this one. Long-term - though before GA, obviously - we need to decide if we want to keep relying on `bytes` this way - exposing it publicly - or make sure all our request/response ducks are in a row. Then again, we do expose `bytes::Bytes` elsewhere.

/cc @analogrelay

PR #2000, opened 2025-01-16
Add the test-proxy client

### LINKED ISSUE #1875: Generate Test Proxy client
Hopefully our TypeSpec emitter can handle whatever is in the TypeSpec for Test Proxy.

heaths:
Resolves #1875. I was unable to generate it from the TypeSpec because we lack support for a few things in the emitter, but I rename (hid as dotfile) the tsp-location.yaml for future use and easy reference for anyone wanting to know how.


heaths:
Note that this is only the test-proxy client. I have subsequent tasks to wire everything up but this was enough code - and specific to a single task - that I wanted to get eyes on it. I'll continue on a separate branch and rebase --onto later. Like Go's, thish does not define every single method that test-proxy defines because I don't think we'll need them; however, should we, I'll add them as needed. Think of this more as an MVP with legs.

sdk/core/azure_core_test/src/proxy/models.rs
```diff
@@ -0,0 +1,93 @@
...
+use serde::{Deserialize, Serialize};
+use std::{collections::HashMap, fmt};
+
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Error {
+    #[serde(skip, default = "Error::default_status")]
+    pub status_code: StatusCode,
+    pub status: String,
+    pub message: String,
```
LarryOsterman, line 14: The HTTP RFC is fairly emphatic that `status` should not be considered to be authoritative, and it may not be present at all. Does it make sense to consider `status` and `message` as `Option<String>`? 

Maybe another way of phrasing this is: Is it safe to treat an empty `status` and `message` as "not present"?

LarryOsterman (commented):
Mostly this looks good, only a couple of small comments.

sdk/core/azure_core_test/src/proxy/models.rs
heaths: Our service specs that defined this model say they are "required" which, in a response model, means they are always there. Given this is a client over a specific service, I'll take specificity over generalization.

heaths: To put it another way, if the service does something it didn't spec correctly, I'd rather know about it and fix the service and/or client as problems arise so we're not incorrectly responding to undefined behavior.

sdk/core/azure_core_test/src/proxy.rs
```diff
@@ -260,91 +257,94 @@ impl FromStr for Version {
...
+        major: 20240109,
+        minor: 1,
+        metadata: Some("1".into()),
+    };
+    let d = Version {
+        major: 20240109,
+        minor: 1,
+        metadata: Some("2".into()),
+    };
+    assert!(a == a);
```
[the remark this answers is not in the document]
heaths: Normally I agree, but these test variables are largely arbitrary. It's just data I'm comparing and no better than calling them, say, "lvalue" or "rvalue" especially in lines like this where I'm comparing identity equality.

LarryOsterman (approved):

Issue #2009, opened 2025-01-18
Support wasm32 for recorded tests


heaths:
We should consider supporting recorded tests for `wasm32-unknown-unknown` and friends. When resolving #1877 I made it possible to compile it, but we can't spawn processes for the wasm32 architecture so the core functionality doesn't work. Prior to that, the `#[recorded::test]` attribute already elides tests for wasm32.

We should be able to support it, however, if we require people to start it manually i.e., not bootstrap it automatically, and find some way to pass the URL.

PR #2010, opened 2025-01-18
Add support for sanitizers, matchers

### LINKED ISSUE #1877: Add support to sanitize headers, parameters, body, etc.
Similar to other languages, expose support to sanitize all that we can with Test Proxy. Much of this should be built on #682 but we'll also want to support more since we'll be storing (somewhere) full recordings.

heaths:
Resolves #1877. This also makes it possible to compile `azure_core_test` for the wasm32 architecture but will not currently work for wasm32. I opened #2009 to track.


heaths (commented):
This doesn't complete the test-proxy work, but does add support for modifying what it does. I'll finish up the transport in another PR.

sdk/core/azure_core_test/src/proxy/policy.rs
```diff
@@ -0,0 +1,57 @@
...
+    sync::{Arc, RwLock},
+};
+use tracing::{debug_span, Instrument};
+
+use crate::Skip;
+
+#[derive(Debug, Default)]
+pub struct RecordingPolicy {
+    pub test_mode: TestMode,
+    pub options: RwLock<RecordingOptions>,
```
heaths, line 21: @analogrelay wondering if you have any better ideas. Basically, I want the policy (in concept - not necessarily this implementation) to send headers as needed. And while there *should* be a 1:1 correlation between `Recording` and `RecordingPolicy`, separate threads are certainly a possibility and this, I believe, upholds the intent of keeping a vector of `Arc<dyn Policy>` as the pipeline: to share policies as appropriate.

heaths: I had contemplated that, but test code would have to remember to inherit the `Context` for every call into a client, which is error prone. Then there are the implicit calls e.g., for credentials. A macro *could* certainly do that but would be much more expensive to develop. Perhaps a goal further down the road.

LarryOsterman (approved):

sdk/core/azure_core_test/src/proxy/policy.rs
heaths: Thinking about this more, I guess I could see a case where callers could do this as needed like so:

```rust
let ctx = Context::with_value(SkipRecording::RequestBody);
let options = ClientMethodOptions { context: ctx, ..Default::default() };
client.foo(..., Some(ClientFooOptions { method_options: options.clone(), ..Default::default() })).await?;
client.bar(..., None).await?;
```

@LarryOsterman or @RickWinter, thoughts?

LarryOsterman: My biggest issue here is that I *think* that means we end up with two different ways of doing the same thing, which can be confusing. OTOH, the only people being confused would be SDK client developers.

And if we remove skip from the RecordOptions (to make it so that there's only one way of doing that), then we have some recording options in the Context and some others in RecordOptions, which is just *weird*.

heaths: I was thinking of supporting skipping only one way, but you make a good point about having other options on the `recording` object.

PR #2014, opened 2025-01-22
[Storage] `get_blob_properties` for `BlobClient`

vincenttran-msft:

- Fleshed out implementation of `BlobClient` to support `get_blob_properties`
- `get_blob_properties` and simple test cases (get properties with existing blob, container not exists case)


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -1,2 +1,70 @@
...
+use azure_core::credentials::TokenCredential;
+use azure_core::{BearerTokenCredentialPolicy, Policy, Result};
+use std::sync::Arc;
+
+pub struct BlobClient {
+    pub(crate) endpoint: String,
+    pub(crate) container_name: String,
+    pub(crate) blob_name: String,
+    pub(crate) credential: Arc<dyn TokenCredential>,
+    pub(crate) client: GeneratedBlobClient,
```
vincenttran-msft, line 18: Currently this will fail (since all warnings are errors) since we never use `credential` or `endpoint`: passed directly to underlying generated client on L47. 

I think because this information is contained in the `GeneratedBlobClient` and we maintain that, so we can likely omit storing this information directly on this struct.

sdk/storage/azure_storage_blob/tests/test_blob_client.rs
```diff
@@ -0,0 +1,60 @@
...
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+
+    #[recorded::test(live)]
+    async fn test_get_blob_properties() -> Result<(), Box<dyn Error>> {
+        let storage_account_name =
+            env::var("STORAGE_ACCOUNT_NAME").unwrap_or("storagename".to_string());
```
vincenttran-msft, line 20: supernit: `AZURE_STORAGE_ACCOUNT_NAME`?

sdk/storage/azure_storage_blob/src/models/mod.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+mod blob_properties;
+
+pub use crate::generated::enums::*;
+pub use crate::generated::models::*;
```
vincenttran-msft, line 7: TODO: Need to start a discussion regarding the way generated code is currently imported. Likely need to bring the scope down to `pub(crate)` and only promote to the root through `lib.rs`.

Currently if this is done (`pub` -> `pub(crate)`), several fields are `pub` but are never accessed, and all warnings are errors and will block CI.

sdk/storage/azure_storage_blob/Cargo.toml
```diff
@@ -10,11 +10,19 @@ rust-version.workspace = true
...
 
 [dependencies]
+async-trait.workspace = true
 azure_storage_common.workspace = true
 azure_core = { workspace = true, features = ["xml"] }
+azure_identity = { workspace = true }
 serde = { workspace = true }
 time = { workspace = true }
 typespec_client_core = { workspace = true, features = ["derive"] }
+uuid = { workspace = true }
```
heaths, line 20: Nit: Be consistent with `<crate>.workspace = true` like you did on line 13.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -1,2 +1,67 @@
...
+
+use crate::{
+    models::BlobProperties, pipeline::StorageHeadersPolicy, BlobBlobClientGetPropertiesOptions,
+    BlobClientOptions, GeneratedBlobClient,
+};
+use azure_core::{credentials::TokenCredential, BearerTokenCredentialPolicy, Policy, Result};
+use std::sync::Arc;
+
+pub struct BlobClient {
+    pub endpoint: String,
```
heaths, line 12: Type should be `azure_core::Url`. See generated clients.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -1,2 +1,67 @@
...
+    pub container_name: String,
+    pub blob_name: String,
+    client: GeneratedBlobClient,
+}
+
+impl BlobClient {
+    pub fn new(
+        endpoint: String,
```
heaths, line 20: Should take a `&str`. See generated clients. You'd then `let endpoint: Url = endpoint.parse()?;`.

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -3,13 +3,15 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 // Code generated by Microsoft (R) Rust Code Generator. DO NOT EDIT.
 
+pub mod clients;
```
heaths, line 6: This should not be public. Clients go in the root, so you'd have instead:

```rust
mod clients;
// ...
pub use clients::*;
```

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -3,13 +3,15 @@
...
 mod generated;
+pub mod pipeline;
```
heaths, line 8: The pipeline should not be public. This is an implementation detail. If you need it elsewhere in your crate, use `pub(crate)`.

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -3,13 +3,15 @@
...
 
-pub use crate::generated::clients::*;
+pub mod models;
 
-pub mod models {
-    pub use crate::generated::enums::*;
-    pub use crate::generated::models::*;
-}
+pub use crate::generated::clients::*;
+pub use blob_client::BlobClient as GeneratedBlobClient;
```
heaths, line 13: Why expose your generated crate separately? If you're trying to merely extend the generated client, use extension methods as described in the guidelines: https://azure.github.io/azure-sdk/rust_implementation.html#rust-client-convenience

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -3,13 +3,15 @@
...
 
-pub use blob_client::BlobClient;
+pub use crate::blob_client::BlobClientOptions;
+pub use crate::clients::BlobClient;
```
heaths, line 16: Why are these even defined in a separate module? Don't do 1 type per file like .NET. That's not idiomatic and makes things harder since every file is a separate module which then requires changing accessibility. The only types that should be public are your clients, client options, method options, and request models in the root, and remaining models (response models, basically) in a `models` public module.

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+use azure_core::headers::Headers;
+
+/// Properties of an Azure Storage blob.
+///
+#[derive(Clone, Default, Debug)]
```
heaths, line 8: Models should derive `SafeDebug` unless you're sure they will *never* have PII: https://azure.github.io/azure-sdk/rust_introduction.html#rust-client-methods-configuration-debug (this says "DO" but that's a mistake; it should be "DO NOT" and I'll fix).

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
```diff
@@ -0,0 +1,52 @@
...
+pub struct BlobProperties {
+    pub access_tier_inferred: String,
+    pub access_tier: String,
+    pub blob_type: String,
+    pub content_length: String,
+    pub content_md5: String,
+    pub content_type: String,
+    pub creation_time: String,
+    pub etag: String,
```
heaths, line 17: We have a separate `Etag` type you should use in `azure_core`.

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub last_modified: String,
```
heaths, line 18: Why is this a string and not a `DateTimeOffset` which we use for times elsewhere.

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
```diff
@@ -0,0 +1,52 @@
...
+
+use azure_core::headers::Headers;
+
+/// Properties of an Azure Storage blob.
+///
+#[derive(Clone, Default, Debug)]
+pub struct BlobProperties {
+    pub access_tier_inferred: String,
+    pub access_tier: String,
+    pub blob_type: String,
```
heaths, line 12: Are all these truly just free-form strings or extensible enums, which we cover in the guidelines: https://azure.github.io/azure-sdk/rust_introduction.html#rust-enums

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub creation_time: String,
+    pub etag: String,
+    pub last_modified: String,
+    pub lease_state: String,
+    pub lease_status: String,
+    pub server_encrypted: String,
+}
+
+impl BlobProperties {
+    pub fn build_from_response_headers(response_headers: &Headers) -> BlobProperties {
```
heaths, line 25: Does this really need to be public? Better to keep things private (or `pub(crate)` until we know they need to be public.

heaths: I also recommend implementing `AsHeaders` and `FromHeaders` instead, which plugs into our system better. see `typespec_client_core::http::headers` and you can find examples therein as well.

heaths: Many of these should also be typed headers, which also corresponds to their fields being defined as those types on your struct. `Etag` is a good example.

Whatever is typed in Go should be typed here as well.

heaths: That lets you call `let properties: BlobProperties = headers().get()?;` for example.

sdk/storage/azure_storage_blob/src/pipeline/storage_headers_policy.rs
```diff
@@ -0,0 +1,34 @@
...
+impl Policy for StorageHeadersPolicy {
+    async fn send(
+        &self,
+        ctx: &Context,
+        request: &mut Request,
+        next: &[Arc<dyn Policy>],
+    ) -> PolicyResult {
+        let result = request
+            .headers()
+            .get_str(&HeaderName::from("x-ms-client-request-id"));
```
heaths, line 23: Better to define this as a `const`:

```rust
const CLIENT_REQUEST_ID: HeaderName = HeaderName::from_static("x-ms-client-request-id");
```

Though, I think we already have one that defined in `azure_core`.

sdk/storage/azure_storage_blob/src/pipeline/storage_headers_policy.rs
```diff
@@ -0,0 +1,34 @@
...
+
+        match result {
+            Ok(_) => {}
+            Err(_) => {
+                let request_id = Uuid::new_v4().to_string();
+                request.insert_header("x-ms-client-request-id", request_id);
+            }
+        }
```
heaths, line 31: More concise:

```rust
if request.headers().get_optional_string(&CLIENT_REQUEST_ID).is_none() {
    let request_id = Uuuid::new_v4().to_string();
    request.insert_header(&CLIENT_REQUEST_ID, &request_id);
}
```

sdk/storage/azure_storage_blob/tests/test_blob_client.rs
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+use azure_core_test::recorded;
+use azure_identity::DefaultAzureCredentialBuilder;
+use azure_storage_blob::{BlobBlobClientGetPropertiesOptions, BlobClient, BlobClientOptions};
+use std::{env, error::Error};
+
+#[cfg(test)]
+mod tests {
```
heaths, line 10: Unnecessary to declare a separate module. You do that when inline, not for integration tests. Files under `test/` are already implicitly `cfg(test)`.

heaths (changes requested):

sdk/storage/azure_storage_blob/src/lib.rs
vincenttran-msft: wrt line 12- this line is here to support the generated code as currently imports look like:
[use crate::blob_service_client::BlobServiceClient](https://github.com/Azure/azure-sdk-for-rust/blob/d69577120fc0e1dc8b0a5471791114eb29656725/sdk/storage/azure_storage_blob/src/generated/clients/blob_client.rs#L11);
So I will need to get in touch with Joel/Catalina regarding improving the imports in generated code.

wrt line 13- That `BlobClient` is the top-level navigation client (that allows us to get the actual generated REST clients). I can change to using this in-line in the imports:
`use crate::{
    generated::clients::blob_client::BlobClient as GeneratedBlobClient,.....`

The main crux of the issue is that the top-level navigation clients struct collides with our name `BlobClient`

sdk/storage/azure_storage_blob/src/lib.rs
vincenttran-msft: L15 here is the generated model. Is there a better way to bring `BlobClientOptions` up to the root so that it can be imported as `use azure_storage_blob::BlobClientOptions`

L16 is our `BlobClient`, so I can take the feedback there. That'll be exported for free after taking an above feedback item you've suggested: `pub use clients::*;`

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
vincenttran-msft: Hi @heaths Heath, in regards to all the feedback regarding our handwritten `BlobProperties` model-- I can work on getting this implementation to better, but this was mostly a stopgap implementation since it is still a pending discussion if these sort of models will be auto-generated and available in time for the MVP. I will leave this thread open and can revisit this later on- but I don't want to this to be blocking unless we are going to ship with our handwritten models.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -2,22 +2,22 @@
 // Licensed under the MIT License.
 
 use crate::{
-    models::BlobProperties, pipeline::StorageHeadersPolicy, BlobBlobClientGetPropertiesOptions,
-    BlobClientOptions, GeneratedBlobClient,
+    generated::clients::blob_client::BlobClient as GeneratedBlobClient, models::BlobProperties,
```
vincenttran-msft, line 5: Ideally this will be a non-issue:
`generated::clients::blob_client::BlobClient as GeneratedBlobClient` and we won't have to reimport it as a different name once the collision is resolved with generated names.

sdk/storage/azure_storage_blob/src/lib.rs
heaths: My comment is really more about this line declaring it as `pub`. If you're wrapping the generated client it shouldn't be exposed as well. Whatever is exposed publicly from `src/lib.rs` is part of the public API. That's transitive throughout imports all the way up to `src/lib.rs`.

vincenttran-msft:
Hi @heaths Heath, I have documented the remaining 4 feedback items into my notes, and will address them in the following PR. I have commented out the testcases since they currently rely on static resources and have a hard dependency on some NYI functionality. With that in mind- please let me know if there is any other showstoppers in this specific PR. Thanks!

sdk/storage/azure_storage_blob/tests/test_blob_client.rs
heaths, line 1: Don't start your module with `test_` (the file name). It's already in `tests/` and this is redundant. `blob_client.rs` is perfect.

sdk/storage/azure_storage_blob/tests/test_blob_client.rs
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// use azure_core_test::recorded;
+// use azure_identity::DefaultAzureCredentialBuilder;
+// use azure_storage_blob::{BlobBlobClientGetPropertiesOptions, BlobClient, BlobClientOptions};
+// use std::{env, error::Error};
+
+// #[recorded::test(live)]
+// async fn test_get_blob_properties() -> Result<(), Box<dyn Error>> {
```
heaths, line 10: Why is all this commented out? If you're not ready to record, fine. But you've already marked them as live-only tests anyway so that's not a concern. We still want to build them to make sure they work.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -1,2 +1,66 @@
...
+    models::BlobProperties, pipeline::StorageHeadersPolicy, BlobBlobClientGetPropertiesOptions,
+    BlobClientOptions, GeneratedBlobClient,
+};
+use azure_core::{credentials::TokenCredential, BearerTokenCredentialPolicy, Policy, Result, Url};
+use std::sync::Arc;
+
+pub struct BlobClient {
+    pub endpoint: Url,
+    pub container_name: String,
+    pub blob_name: String,
```
heaths, line 14: Clients shouldn't have public members. We only do that for plain ol' data objects (models). We have thread-safety guarantees for clients and making them mutable violates that. You can add accessors, which is already covered in our guidelines for `endpoint`.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -1,2 +1,66 @@
...
+    client: GeneratedBlobClient,
+}
+
+impl BlobClient {
```
heaths, line 18: In other languages, I thought you can only get a `BlobClient` from a `ContainerClient` or something i.e., these aren't creatable. If so, they must be exported from both the crate root (re-exported) and from the `clients` module in which all clients - creatable or gettable - are exported.

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -3,13 +3,17 @@
...
-    pub use crate::generated::models::*;
-}
+pub mod models;
+
+pub use crate::generated::clients::*;
+pub(crate) use blob_client::BlobClient as GeneratedBlobClient;
 
-pub use blob_client::BlobClient;
+pub use crate::generated::clients::blob_blob_client::BlobBlobClientGetPropertiesOptions;
+pub use crate::generated::clients::blob_client::BlobClientOptions;
```
heaths, line 19: Why export them from the root for the crate? Just have the modules within the crate import `crate::generated::clients::BlobClient`. Re-exporting types only for the crate like this is going to lead to a lot of confusion, I think. It's not typical.

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+use azure_core::headers::Headers;
+
+/// Properties of an Azure Storage blob.
+///
+#[derive(Clone, Default, Debug)]
+pub struct BlobProperties {
```
heaths, line 9: Model properties must all be `Option<T>`.

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub creation_time: String,
+    pub etag: String,
+    pub last_modified: String,
+    pub lease_state: String,
+    pub lease_status: String,
+    pub server_encrypted: String,
+}
+
+impl BlobProperties {
+    pub(crate) fn build_from_response_headers(response_headers: &Headers) -> BlobProperties {
```
heaths, line 25: Use `FromHeaders` and `AsHeaders`.

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
```diff
@@ -0,0 +1,52 @@
...
+        let mut properties = BlobProperties {
+            ..Default::default()
+        };
+        for (key, value) in response_headers.iter() {
+            match key.as_str() {
+                "content-length" => properties.content_length = String::from(value.as_str()),
```
heaths, line 31: Some, like this, aren't strings. Why are you parsing them as such? You should use our header traits and even define a header. We also have a bunch of these already defined as `HeaderName`s that you should use for consistency.

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
```diff
@@ -0,0 +1,52 @@
...
+                "content-md5" => properties.content_md5 = String::from(value.as_str()),
+                "content-type" => properties.content_type = String::from(value.as_str()),
+                "etag" => properties.etag = String::from(value.as_str()),
```
heaths, line 34: Use our existing Etag header type.

sdk/storage/azure_storage_blob/tests/test_blob_client.rs
vincenttran-msft: > I have commented out the testcases since they currently rely on static resources and have a hard dependency on some NYI functionality.

So I can see your point that they are innocuous because they are live-only, but will leave them commented out since they don't really make any sense until the next PR which will have some NYI APIs.

sdk/storage/azure_storage_blob/src/lib.rs
vincenttran-msft: So for L16 `pub(crate) use blob_client::BlobClient as GeneratedBlobClient;`, I can change this to in-line importing:
`generated::clients::blob_client::BlobClient as GeneratedBlobClient`

wrt L18-19, I re-exported them because I thought we wanted all client options accessible from the root ([guidelines section](https://azure.github.io/azure-sdk/rust_introduction.html#rust-client-configuration-namespace)). Without re-exporting them, we would have to import them as:
```rust
use crate::generated::{BlobClientOptions, BlobBlobClientGetPropertiesOptions};
```

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub last_modified: Option<String>,
+    pub lease_state: Option<LeaseState>,
+    pub lease_status: Option<LeaseStatus>,
+    pub server_encrypted: Option<bool>,
+}
+
+impl FromHeaders for BlobProperties {
+    type Error = url::ParseError;
+
+    fn header_names() -> &'static [&'static str] {
```
vincenttran-msft, line 43: I was a little confused when trying to pattern match this-- is this supposed to be an exhaustive list of the possible optional headers? For now, I only had the plain-text version of the headers I had to create a new `const` for.

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
vincenttran-msft: wrt this comment, is there a currently accepted way to convert from RFC1123 strings (as returned by the service) ex:
`creation_time: Some("Sat, 01 Feb 2025 00:17:37 GMT"` to a `OffsetDateTime` object? I know you can provide a custom pattern query to [parse](https://docs.rs/time/latest/time/struct.OffsetDateTime.html#method.parse), but I assume that if we have `OffsetDateTime` elsewhere we may have some common utility for this?

vincenttran-msft:
I was under the impression this PR was no longer passing due to merging in `main` and messing up the `Cargo.lock`, but even after attempting to merge again and `cargo generate lock-file`, I am still hitting some errors here. Some assistance on getting CI green alongside the review would be greatly appreciated!

vincenttran-msft:
Hi @heaths Heath, whenever you get a free moment if you could take a look at this to help resolve the CI issues and any merge-blocking stoppers. Thanks!

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
```diff
@@ -62,38 +63,38 @@ impl FromHeaders for BlobProperties {
             .unwrap();
         properties.access_tier_inferred = Some(access_tier_inferred);
 
-        let access_tier: AccessTier = headers.get_optional_as(&BLOB_ACCESS_TIER).unwrap().unwrap();
-        properties.access_tier = Some(access_tier);
+        let access_tier: Option<AccessTier> = headers.get_optional_as(&BLOB_ACCESS_TIER).unwrap();
```
heaths: Never `unwrap()` in production code. That will panic and tear the whole process down. You already return a `Result` from this, so you use the `?` operator e.g.,
```suggestion
        let access_tier: Option<AccessTier> = headers.get_optional_as(&BLOB_ACCESS_TIER)?;
```
If you get a compiler error that there's no `From<T>` for that conversion, please let me know. For one, `from_headers` should be returning an `azure_core::Result<Option<Self>>` (we default to `azure_core::Error`) and we should already have a bunch of conversions anyway.

So no `unwrap()` or `expect()` in production code. Tests are fine. Panicking fails the test, which is probably what you want if unexpected anyway. Things like `unwrap_or_default()` are okay when used correctly, like if you need an empty string for a `String` if nothing is specified.

heaths (changes requested):

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
vincenttran-msft: Hi @heaths Heath, I believe you responded to an old commit here, since this has been resolved in the latest revision:
```rust
let access_tier: Option<AccessTier> = headers.get_optional_as(&BLOB_ACCESS_TIER)?;
```

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
heaths: This is from our older, unofficial code. Frankly, I'm not sure what purpose it serves. Do you see any usage throughout the rest of the repo (Shift+F12 by default in VSCode)? Maybe we could get rid of it if there's no use case currently.

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
heaths: Not sure what happened. I just clicked on the View updates button in GH.

sdk/storage/azure_storage_blob/src/models/blob_properties.rs
vincenttran-msft: Gotcha. I mostly cribbed this from the test example here:
https://github.com/Azure/azure-sdk-for-rust/blob/048717f4f5fc758e85769663d3279ff67437ebde/sdk/typespec/typespec_client_core/src/http/headers/mod.rs#L347-L361

But after some more digging and reading the documentation, I see this is supposed to representative of the possible headers that the `from_headers` function expects to parse (so should be **exhaustive**). I will apply this fix.

sdk/storage/azure_storage_blob/src/lib.rs
vincenttran-msft: Gotcha. I took a note of this and will apply this feedback next PR-- but will not apply any changes as the structure of the generated imports completely changed with the newly generated versions.

heaths:
Building and testing all features in the workspace works fine on my machine. There were some changes that went in today that may fix them, though. Try rebasing again.

heaths:
The command failing here works fine on my box, and I *just* pulled upstream/main once again:

```bash
cargo +stable clippy --target=wasm32-unknown-unknown --workspace --keep-going --no-deps
```

You've been merging (in general, don't do that unless you're trying to main history for some reason), and I'm guessing you have a bad merge at some point. If all your changes have been under sdk/storage, I recommend from within your topic branch:

```bash
git fetch upstream main
git rebase -X theirs upstream/main
```

The problem you're running into is something I believe I fixed last Friday (was merged today, Monday).

heaths:
Also, I believe your merge is dirty because at one point `Cargo.lock` was accidentally upgraded to all new versions, which caused some issues with `getrandom` used by `uuid`. I had to downgrade `uuid` from 0.3 to 0.2 (what we had) because of it; lots of people are running into this problem because of a breaking change in `getrandom` and because we have dependencies that take a dependency on an older version: the workaround for the older and newer versions conflicts with each other.

vincenttran-msft:
@heaths Thanks for that callout. After merging and reverting the lockfile and doing a clean `cargo build` on the crate, we are now back to passing! Thanks for the help Heath, it should now be ready for a final review.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -1,2 +1,70 @@
...
+        &self,
+        options: Option<BlobBlobClientGetPropertiesOptions<'_>>,
+    ) -> Result<BlobProperties> {
+        let response = self
+            .client
+            .get_blob_blob_client(self.container_name.clone(), self.blob_name.clone())
+            .get_properties(options)
+            .await?;
+
+        // let blob_properties: Option<BlobProperties> = response.headers().get_optional()?;
```
heaths, line 66: Nit: don't really need to check in.

sdk/storage/azure_storage_blob/src/pipeline/storage_headers_policy.rs
```diff
@@ -0,0 +1,31 @@
...
+impl Policy for StorageHeadersPolicy {
+    async fn send(
+        &self,
+        ctx: &Context,
+        request: &mut Request,
+        next: &[Arc<dyn Policy>],
+    ) -> PolicyResult {
+        if request
+            .headers()
+            .get_optional_string(&CLIENT_REQUEST_ID)
```
heaths, line 23: This is fine for now, but we should define an `x-ms-client-request-id` policy in `azure_core` like most other languages have. I'll open a separate issue on core for this.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -0,0 +1,60 @@
...
+use azure_storage_blob::{BlobBlobClientGetPropertiesOptions, BlobClient, BlobClientOptions};
+use std::{env, error::Error};
+
+#[recorded::test(live)]
+async fn test_get_blob_properties() -> Result<(), Box<dyn Error>> {
+    // Setup
+    let storage_account_name = env::var("AZURE_STORAGE_ACCOUNT_NAME")
+        .expect("Failed to get environment variable: AZURE_STORAGE_ACCOUNT_NAME");
+    let endpoint = format!("https://{}.blob.core.windows.net/", storage_account_name);
+    let credential = DefaultAzureCredentialBuilder::default().build()?;
```
heaths, line 15: Nit: use `DefaultAzureCredential::new()?`. I didn't even realize we still had the builder in there. `azure_identity` is on my list to clean up for beta.2 and add more credentials.

FWIW - and not a blocker for this PR - but all this stuff will fit nicely with test-proxy. See my `azure_security_keyvault_secrets` tests.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    // Act
+    let blob_client = BlobClient::new(
+        &endpoint,
+        String::from("testcontainer"),
+        String::from("test_blob.txt"),
+        credential,
+        Some(BlobClientOptions::default()),
```
heaths, line 23: Pass `None`, though once you use test-proxy you'll need to instrument your `BlobClientOptions::client_options` anyway, but don't show customers to do this in doc tests or examples. It's unnecessarily verbose and not idiomatic.

heaths (approved):
A couple nits/tips but nothing blocking. LGTM

RickWinter (approved):

PR #2023, opened 2025-01-25
[Storage] `BlobContainerClient`, `BlobServiceClient`, `get_container_properties`, and `get_service_properties` API and other dependencies

vincenttran-msft:

- Fleshed out implementation of `BlobContainerClient` and `BlobServiceClient`
- `get_container_properties` with simple test cases (one positive, one with container not exists)
- `get_service_properties` with a simple test case (positive, assert against expected field)
- `create_container` which is a required dependency to make testing easier
- `delete_container` falls under the same category as above


sdk/storage/azure_storage_blob/src/models/container_properties.rs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+use azure_core::headers::Headers;
+
+/// Properties of an Azure Storage blob.
```
vincenttran-msft, line 6: nit: container

PR #2024, opened 2025-01-25
Remove write! calls for Error Display impl

sunsided:
This simply replaces calls to `write!(f, "{variable}")` in the `Display` implementation of `typespec::error::Error` and other crates with direct calls to `f.write_str(variable)` when they are `str`-like, or `Display::fmt` when they are complex types.

### DIFF
--- a/sdk/typespec/src/error/mod.rs
+++ b/sdk/typespec/src/error/mod.rs
@@ -293,12 +293,10 @@ impl From<core::convert::Infallible> for Error {
 impl Display for Error {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match &self.context {
-            ErrorContext::Simple(kind) => write!(f, "{kind}"),
-            ErrorContext::Message { message, .. } => write!(f, "{message}"),
-            ErrorContext::Custom(Custom { error, .. }) => write!(f, "{error}"),
-            ErrorContext::Full(_, message) => {
-                write!(f, "{message}")
-            }
+            ErrorContext::Simple(kind) => std::fmt::Display::fmt(&kind, f),
+            ErrorContext::Message { message, .. } => f.write_str(message),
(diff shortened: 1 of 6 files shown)


sunsided:

sdk/typespec/src/error/mod.rs
```diff
@@ -293,12 +293,10 @@ impl From<core::convert::Infallible> for Error {
 impl Display for Error {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match &self.context {
-            ErrorContext::Simple(kind) => write!(f, "{kind}"),
-            ErrorContext::Message { message, .. } => write!(f, "{message}"),
-            ErrorContext::Custom(Custom { error, .. }) => write!(f, "{error}"),
-            ErrorContext::Full(_, message) => {
-                write!(f, "{message}")
-            }
+            ErrorContext::Simple(kind) => std::fmt::Display::fmt(&kind, f),
```
heaths, line 296: I had done the conversion to `write_str` before but decided to revert it because we still need `write!()` in some places so I kept it consistent and there was no discernable difference in perf (I ran some benchmarks). That said, I'm not against it; however, why replace `write!()` with an explicit call to `Display::fmt()` like this? `write!()` is certainly more concise and there's no perceptible runtime difference.

heaths (changes requested):
Thank you for your contribution! I have a question before merging.

sdk/typespec/src/error/mod.rs
LarryOsterman: From my point of view (since I'm the primary offender in this PR :)), I don't have a bit of a problem with the EH and AMQP changes - these are almost entirely a consequence of my naivety w.r.t. Rust's conventions. And if it saves an allocation, that means one less possible point of failure in an error path, which IMHO is a very good idea.

heaths (approved):

PR #2027, opened 2025-01-27
Add version numbers for workspace dependencies

hallipr:
- Workspace dependencies should include `version` to accommodate `cargo package`
- Path based Intra-repo dev-dependencies should use local paths, not `workspace = true`
  - This is because, at package time, dev dependencies don't resolve correctly when the local dependency is unpublished but has a version number referenced in the workspace toml.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,28 +26,36 @@ rust-version = "1.80"
 
 [workspace.dependencies.typespec]
 path = "sdk/typespec"
+version = "0.2.0"
 default-features = false
 
 [workspace.dependencies.typespec_client_core]
 path = "sdk/typespec/typespec_client_core"
+version = "0.1.0"
 default-features = false
 
 [workspace.dependencies.typespec_macros]
+version = "0.1.0"
 path = "sdk/typespec/typespec_macros"
 
 [workspace.dependencies.azure_core]
+version = "0.22.0"
 path = "sdk/core/azure_core"
 
 [workspace.dependencies.azure_core_amqp]
+version = "0.1.0"
 path = "sdk/core/azure_core_amqp"
 
 [workspace.dependencies.azure_core_test]
+version = "0.1.0"
 path = "sdk/core/azure_core_test"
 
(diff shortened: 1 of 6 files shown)


heaths (approved):
As long as everything keeps working, sure. But the nightly build failure concerns me somewhat. Is this because package paths lead to a loop where, for some reason, versions don't?

hallipr:
> As long as everything keeps working, sure. But the nightly build failure concerns me somewhat. Is this because package paths lead to a loop where, for some reason, versions don't?

I think it may be due to circular references between dev/non-dev packages.

Consider:

```toml
[package]
name = "one"
version = "1.0.1"

[dependencies]
two = { path = ../two, version = "1.0.1" }
```

and

```toml
[package]
name = "two"
version = "1.0.1"

[dev-dependencies]
one = { path = ../one, version = "1.0.1" }
```

We can't build "one" until we've published (at least locally) "two@1.0.1"
The same for "two"

If we remove the version from the dev-dependency, then we can package "two" using the source version of "one"
then we can package "one" using the package version of "two"

heaths:
Do you know which packages are at fault? We use some packages in tests e.g., `azure_identity` has `azure_core` in its `dependencies`, but we may have `azure_identity` in `azure_core`'s `dev-dependencies`. Maybe we need to strip those examples/integration tests if indeed that's the problem.

analogrelay (approved):

hallipr:
> Do you know which packages are at fault? We use some packages in tests e.g., `azure_identity` has `azure_core` in its `dependencies`, but we may have `azure_identity` in `azure_core`'s `dev-dependencies`. Maybe we need to strip those examples/integration tests if indeed that's the problem.

There are 2 circular dependencies:

=> dependency
-> devDependency

azure_core_test => azure_core_test_macros
azure_core_test_macros -> azure_core_test

typespec_client_core => typespec_macros
typespec_client_core -> typespec_macros
typespec_macros -> typespec_client_core

PR #2032, opened 2025-01-29
Add SetPackageVersion and GetExistingPackageVersions to support Prepa…

hallipr:
…re-Release.ps1


eng/scripts/Language-Settings.ps1
```diff
@@ -6,7 +6,35 @@ $packagePattern = "Cargo.toml"
...
+    $ReleaseDate = Get-Date -Format "yyyy-MM-dd"
+  }
+  & "$EngDir/scripts/Update-PackageVersion.ps1" -ServiceDirectory $ServiceDirectory -PackageName $PackageName `
+      -NewVersionString $Version -ReleaseDate $ReleaseDate -ReplaceLatestEntryTitle $ReplaceLatestEntryTitle
+}
+
+function GetExistingPackageVersions ($PackageName, $GroupId=$null)
+{
+  try {
+    $PackageName = $PackageName.ToLower()
```
heaths, line 22: This probably doesn't hurt, but shouldn't these be all lowercase anyway? Wondering if maybe we should let it error otherwise. Or is this purely based on user input and we want to normalize it?

eng/scripts/Language-Settings.ps1
```diff
@@ -6,7 +6,35 @@ $packagePattern = "Cargo.toml"
...
+    $response = Invoke-RestMethod -Method GET -Uri "https://crates.io/api/v1/crates/${PackageName}/versions"
+    $existingVersions = $response.versions | Sort-Object -Property created_at
```
heaths, line 24: This won't always be the case. We should be sorting based on the semver, which I thought @weshaggard had added support for. We might, for example, publish a patch version after a newer version if customers require.

heaths: I see `AzureEngSemanticVersion` used below. `num` should be coerced into this type, assuming it supports sorting...which I assume it does.

eng/scripts/Update-PackageVersion.ps1
```diff
@@ -0,0 +1,71 @@
...
+. (Join-Path $PSScriptRoot '../common/scripts/common.ps1')
+
+$pkgProperties = Get-PkgProperties -PackageName $PackageName -ServiceDirectory $ServiceDirectory
+$packageVersion = $pkgProperties.Version
+
+$packageSemVer = [AzureEngSemanticVersion]::new($packageVersion)
+Write-Host "Current Version: ${PackageVersion}"
+
+if ([System.String]::IsNullOrEmpty($NewVersionString)) {
+  $packageSemVer.IncrementAndSetToPrerelease();
```
heaths, line 48: For 0.x releases, does this set the minor? If not, until we GA would it be worth the effort?

heaths (approved):
I see nothing blocking. Just a few questions. Did you run this and see if you get the expected output?

weshaggard (approved):

eng/scripts/Language-Settings.ps1
hallipr: It would be based on user input.  The user will call `Prepare-Release.ps1 --ServiceDirectory foo --PackageName azure_foo`.  I normalize it here.

eng/scripts/Update-PackageVersion.ps1
hallipr: 0.x releases are already considered "Prerelease" so they get a minor bump by default.  i.e.
The next prerelease version for "0.3.0" is "0.4.0".
The next prerelease version for "1.3.1" is "1.4.0-beta.1".
The next prerelease version for "1.4.0-beta.1" is "1.4.0-beta.2"

eng/scripts/Language-Settings.ps1
hallipr: Updated

eng/scripts/Update-PackageVersion.ps1
hallipr: I added this to eng/common a few weeks ago:
https://github.com/Azure/azure-sdk-tools/pull/9668

hallipr:
> I see nothing blocking. Just a few questions. Did you run this and see if you get the expected output?

I ran `Prepare-Release.ps1` with `azure_security_keyvault_secrets`.  It created the package work item and updated the changelog and Cargo.toml correctly.  Once in main, we can run it again to submit the changes in a release-prep PR.

RickWinter (approved):

PR #2033, opened 2025-01-29
Automatically record and play back test recordings

### LINKED ISSUE #1876: Create transport for Test Proxy
Maybe this is just our normal transport, but we should create the wiring necessary to hook it up either implicitly in integration tests (maybe via `#[cfg(test)]` attribute?) or make it really easy to do in integration tests like some sort of `init()` func to wire everything up (much like tracing subscribers).

heaths:
Resolves #1876


sdk/core/azure_core_test/Cargo.toml
```diff
@@ -17,6 +17,8 @@ rust-version.workspace = true
 async-trait.workspace = true
 azure_core = { workspace = true, features = ["test"] }
 azure_core_test_macros.workspace = true
+azure_identity.workspace = true
```
LarryOsterman, line 20: There's an interesting layering challenge here. This depends on azure_identity *never* taking a dependency on azure_core_test.

This has been the subject of much discussion in the C++ repo. My personal take is that the layering is azure_core->azure_identity->azure_core_test but that's not completely consistent with the directory layout.

sdk/core/azure_core_test/src/lib.rs
```diff
@@ -70,6 +84,14 @@ fn find_ancestor(dir: impl AsRef<Path>, name: &str) -> azure_core::Result<PathBu
         if path.exists() {
             return Ok(path);
         }
+
+        let path = dir.join(".git");
+        if path.exists() {
```
LarryOsterman, line 89: The error message appears to be inconsistent with the test - if there's a <path>.git file, you report that "<name> not found under <path>".

I'm sure there's a reason for this, but a comment might be helpful to understand the scenario that is being checked here.

sdk/core/azure_core_test/src/recorded.rs
```diff
@@ -22,38 +21,61 @@ static TEST_PROXY: OnceCell<Result<Arc<Proxy>>> = OnceCell::const_new();
 ///
 /// The [Test Proxy](https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md) service will be started as needed.
 /// Every `#[recorded::test]` will call this automatically, but it can also be called manually by any other test e.g., those attributed with `#[tokio::test]`.
+///
+/// This function will return `Ok(None)` if the test is running in live mode and should not use the test proxy at all.
```
LarryOsterman, line 25: ?? It looks like the function returns a `Result<TestContext>`, but this comment implies that it returns a `Result<Option<TestContext>>` - is one or the other out of sync?

sdk/core/azure_core_test/src/recording.rs
```diff
@@ -120,10 +138,25 @@ impl Recording {
...
+        self.var_opt(key)
+            .unwrap_or_else(|| panic!("{key} is not set"))
+    }
+
+    /// Gets an optional variable from the environment or recording.
+    pub fn var_opt<K>(&self, key: K) -> Option<String>
+    where
+        K: AsRef<str>,
+    {
+        // TODO: Should we attempt to preface "AZURE_" or "{SERVICE_DIRECTORY}_" like .NET does?
```
LarryOsterman, line 156: FWIW, for C++, we ended up [doing both](https://github.com/Azure/azure-sdk-for-cpp/blob/4cbb49f0364fc47cc91bc7b3e7d1945516a1a80e/sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp#L340).

LarryOsterman (approved):

sdk/core/azure_core_test/src/recording.rs
heaths: Good to know. Thanks!

sdk/core/azure_core_test/Cargo.toml
heaths: Loops don't exist* between dependencies and dev-dependencies, though.

*) Well, almost. @hallipr recently ran into an issue testing the publishing pipeline that seems to ignore the distinction and, while likely a bug, doesn't change the fact it's a problem in that case.

sdk/core/azure_core_test/src/lib.rs
heaths: Thank you. I totally got my wires cross on that one.

heaths: Per offline discussion, I'll comment below as to the purpose and clarify the error message a bit as well.

sdk/core/azure_core_test/Cargo.toml
heaths: Per offline discussion, `azure_identity` *could* have a `dev-dependencies` on `azure_core_test` though, in practice, that's hard to do since we have to sanitize everything anyway. Still, shouldn't be a problem if we end up doing that, though.

Also, I'll mark this and the associated macros crate as `publish = false` since we really don't need to publish them and, as you pointed out, Azure/azure-sdk-for-cpp doesn't either.

heaths:
@LarryOsterman I was touching yer' stuff. Shouldn't matter, but it's easier all around to require recorded/live tests return a `Result<T, E>` including type defs like `azure_core::Result<T>`.

.vscode/cspell.json
```diff
@@ -41,12 +41,14 @@
     "hmac",
     "iothub",
     "keyvault",
+    "LLDB",
```
RickWinter, line 44: lower case this

sdk/core/azure_core_test/README.md
```diff
@@ -4,22 +4,24 @@ The types and functions in this crate help test client libraries built on `azure
...
 
 #[recorded::test]
 async fn get_secret(ctx: TestContext) -> Result<()> {
     todo!()
 }
 ```
 
-The `TestContext` parameter is required unless your test function is attribute as `#[recorded::test(live)]` (live-only),
-in which case it is optional. You can name the parameter whatever you want.
+The `TestContext` parameter is required unless your test function is attribute as `#[recorded::test(live)]` (live-only).
```
RickWinter, line 20:
```suggestion
The `TestContext` parameter is required unless your test function is attributed as `#[recorded::test(live)]` (live-only).
```

sdk/core/azure_core_test/src/lib.rs
```diff
@@ -7,11 +7,14 @@ pub mod proxy;
...
 mod recording;
 
-pub use azure_core::test::TestMode;
+use azure_core::Error;
+pub use azure_core::{error::ErrorKind, test::TestMode};
 pub use proxy::{matchers::*, sanitizers::*};
 pub use recording::*;
 use std::path::{Path, PathBuf};
 
+#[cfg_attr(target_arch = "wasm32", allow(dead_code))]
```
RickWinter, line 16: Why do we want to allow dead_code? Add a comment

sdk/core/azure_core_test/src/lib.rs
```diff
@@ -20,20 +23,35 @@ const SPAN_TARGET: &str = "test-proxy";
 /// to setup up the HTTP client to record or play back session records.
 #[derive(Debug)]
 pub struct TestContext {
+    repo_dir: &'static Path,
     crate_dir: &'static Path,
+    service_directory: &'static str,
```
RickWinter, line 28: Is `service_directory` a folder or a path?
Should it be `service_folder`? If using directory then `service_dir` for consistency.

sdk/core/azure_core_test/src/lib.rs
```diff
@@ -52,24 +70,147 @@ impl TestContext {
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the [`TestContext::crate_dir()`] is not rooted within a Git repository.
     pub fn test_data_dir(&self) -> PathBuf {
-        self.crate_dir.join("tests/data")
+        self.crate_dir
+            .join("tests/data")
+            .strip_prefix(self.repo_dir)
+            .expect("not rooted within repo")
```
RickWinter, line 96: Is this error message enough to diagnose what is misconfigured?

sdk/core/azure_core_test/src/proxy/client.rs
```diff
@@ -77,9 +74,9 @@ impl Client {
         let mut url = self.endpoint.clone();
         url = url.join("/Record/Stop")?;
         let mut request = Request::new(url, Method::Post);
-        request.insert_header("accept", "application/json");
-        request.insert_header("content-type", "application/json");
-        request.insert_header(X_RECORDING_ID, recording_id.to_string());
+        request.insert_header(ACCEPT, "application/json");
+        request.insert_header(CONTENT_TYPE, "application/json");
+        request.insert_header(RECORDING_ID, recording_id.to_string());
```
RickWinter, line 79: Nit: For me its clearer if the Keys were named HEADER_*

sdk/keyvault/test-resources.bicep
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+param baseName string = resourceGroup().name
+param tenantId string = '72f988bf-86f1-41af-91ab-2d7cd011db47'
```
RickWinter, line 5: Should these guid values be checked in?

RickWinter (approved):

sdk/core/azure_core_test/src/lib.rs
heaths: This tends to be common in libs, actually. Just about everything you depend on has this either targeted (as we do and should) or general e.g., `#![allow(dead_code)]` which seems lazy. It's because, as a library, we can't always test everything and we expose some public APIs that don't otherwise get touched (any `Debug`-derived printing is ignored by the linter).

sdk/core/azure_core_test/src/lib.rs
heaths: In context I think so. The major thing is to use unique strings people can find.

sdk/core/azure_core_test/src/proxy/client.rs
heaths: That would be redundant given they are all defined in the `header` module. If you like being explicit, import `typespec_client_core::http::header` and you can use `header::ACCEPT`. That's the idiomatic way. We don't like unnecessary repetition.

PR #2039, opened 2025-01-30
update the Cosmos crate README

analogrelay:
This is a first pass at updating the README for the `azure_data_cosmos` crate. I based this on the Go package (https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/azcosmos/README.md) and updated the syntax and added some notes about the differences. I also borrowed from the ongoing work @ronniegeraghty is doing for the Identity create in https://github.com/Azure/azure-sdk-for-rust/pull/1990)

Please take a close look to make sure I got all the Go related stuff converted properly. I'm also doing some further editing passes, but wanted to open things up for folks so they can take some time to find errors I may have made.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/README.md
+++ b/sdk/cosmos/azure_data_cosmos/README.md
@@ -37,8 +37,6 @@ In order to interact with the Azure Cosmos DB service you'll need to create an i
 
 The following section provides several code snippets covering some of the most common Azure Cosmos DB NoSQL API tasks, including:
 * [Create Client](#create-cosmos-db-client "Create Cosmos DB client")
-* [Create Database](#create-database "Create Database")
-* [Create Container](#create-container "Create Container")
 * [CRUD operation on Items](#crud-operation-on-items "CRUD operation on Items")
 
 ### Create Cosmos DB Client
--- a/sdk/cosmos/azure_data_cosmos/README.md
+++ b/sdk/cosmos/azure_data_cosmos/README.md
@@ -10,7 +10,7 @@ This client library enables client applications to connect to Azure Cosmos DB vi
 
 Install the Azure Cosmos DB SDK for Rust with cargo:
 
-```bash
+```sh
 cargo add azure_data_cosmos
 ```
 
@@ -43,16 +43,19 @@ The following section provides several code snippets covering some of the most c
 
 ### Create Cosmos DB Client
 
-The clients support different forms of authentication. The Azure Cosmos DB SDK for Rust supports authorization via Microsoft Entra identities or an account key. However, support for account keys is only available when the `key_auth` feature is enabled in your Cargo.toml file.
-
-```toml
-[dependencies]
-azure_data_cosmos = { version = "...", features = ["key_auth"] }
-```
-
-We **strongly** recommend using Microsoft Entra ID for authentication, rather than account keys.
+In order to interact with the Azure Cosmos DB service, you'll need to create an instance of the `CosmosClient`. You need an endpoint URL and credentials to instantiate a client object.
 
 **Using Microsoft Entra ID**
+
+The example shown below use a `DefaultAzureCredential`, which is appropriate for most local development environments. Additionally, we recommend using a managed identity for authentication in production environments. You can find more information on different ways of authenticating and their corresponding credential types in the [Azure Identity] documentation.
+
+The `DefaultAzureCredential` will automatically pick up on an Azure CLI authentication. Ensure you are logged in with the Azure CLI:
+
+```sh
+az login
+```
+
+Instantiate a `DefaultAzureCredential` to pass to the client. The same instance of a token credential can be used with multiple clients if they will be authenticating with the same identity.
 
 ```rust
 use azure_identity::DefaultAzureCredential;
@@ -60,12 +63,20 @@ use azure_data_cosmos::CosmosClient;
 
 async fn example() -> Result<(), Box<dyn std::error::Error>> {
     let credential = DefaultAzureCredential::new()?;
-    let cosmos_client = CosmosClient::new("myAccountEndpointURL", credential, None)?;
+    let cosmos_client = CosmosClient::new("myAccountEndpointURL", credential.clone(), None)?;
     Ok(())
 }
 ```
 
 **Using account keys**
+
+Cosmos DB also supports account keys, though we strongly recommend using Entra ID authentication. To use account keys, you will need to enable the `key_auth` feature:
+
+```sh
+cargo add azure_data_cosmos --features key_auth
+```
+
+Once that feature is enabled, you can create a client like this:
 
 ```rust
 use azure_core::credentials::Secret;
@@ -102,9 +113,9 @@ Using the above created database for creating a container, like this:
 
 ```rust
 use azure_data_cosmos::models::{ContainerProperties, PartitionKeyDefinition, ThroughputProperties};
-use azure_data_cosmos::CreateContainerOptions;
+use azure_data_cosmos::{CosmosClient, CreateContainerOptions};
 
-async fn example(cosmos_client: azure_data_cosmos::CosmosClient) -> Result<(), Box<dyn std::error::Error>> {
+async fn example(cosmos_client: CosmosClient) -> Result<(), Box<dyn std::error::Error>> {
     let properties = ContainerProperties {
         id: "aContainer".into(),
         partition_key: PartitionKeyDefinition::new(vec!["/id".into()]),
@@ -124,7 +135,7 @@ async fn example(cosmos_client: azure_data_cosmos::CosmosClient) -> Result<(), B
 
 ```rust
 use serde::{Serialize, Deserialize};
-use azure_data_cosmos::models::PatchDocument;
+use azure_data_cosmos::{CosmosClient, models::PatchDocument};
 
 #[derive(Serialize, Deserialize)]
 struct Item {
@@ -133,7 +144,7 @@ struct Item {
     pub value: String,
 }
 
-async fn example(cosmos_client: azure_data_cosmos::CosmosClient) -> Result<(), Box<dyn std::error::Error>> {
+async fn example(cosmos_client: CosmosClient) -> Result<(), Box<dyn std::error::Error>> {
     let item = Item {
         id: "1".into(),
         partition_key: "partition1".into(),
@@ -176,10 +187,6 @@ async fn example(cosmos_client: azure_data_cosmos::CosmosClient) -> Result<(), B
 
 ## Next steps
 
-### Client library support
-
-Client and management libraries <!-- TODO: Update link and uncomment when Rust SDK has a page on the releases site.> listed on the [Azure SDK release page](https://azure.github.io/azure-sdk/releases/latest/python.html)</!--> that support Microsoft Entra authentication accept credentials from this library. You can learn more about using these libraries in their documentation, which is <!-- TODO: uncomment when Rust SDK has a release page.>linked from the release page</!-->available at [Docs.rs](https://Docs.rs/azure_data_cosmos).
-
 ### Provide feedback
 
 If you encounter bugs or have suggestions, [open an issue](https://github.com/Azure/azure-sdk-for-rust/issues).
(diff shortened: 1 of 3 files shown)


analogrelay:
We've got a bit of a build issue. The `README.md` file is embedded as documentation for the root of the crate, which means it builds as doctests. I think this is a Good Thing ™️ since it ensures our README samples compile properly. However, we do that build without `--all-features`, so it's currently failing when building the examples that use `CosmosClient::with_key`, since that method requires the `key_auth` feature. Also, I can't use any of the rustdoc "attributes" on the code block (like `no_compile`) to avoid that because they will break syntax highlighting on the README on GitHub (and possibly crates.io, though I'd hope they can support those attributes).

@heaths Are we able to add `--all-features` to the tests runs in `eng/scripts/Test-Packages.ps1`? I believe that's the script that's failing. I'm happy to do it, but I wanted to check in with you first.

sdk/cosmos/azure_data_cosmos/README.md
```diff
@@ -1,5 +1,200 @@
...
+- [Resource Model of Azure Cosmos DB Service](https://learn.microsoft.com/azure/cosmos-db/sql-api-resources)
+- [Azure Cosmos DB Resource URI](https://learn.microsoft.com/rest/api/documentdb/documentdb-resource-uri-syntax-for-rest)
+- [Partitioning](https://learn.microsoft.com/azure/cosmos-db/partition-data)
+- [Using emulator](https://github.com/Azure/azure-documentdb-dotnet/blob/master/docs/documentdb-nosql-local-emulator.md)
+
+## Next steps
+
+### Client library support
+
+Client and management libraries <!-- TODO: Update link and uncomment when Rust SDK has a page on the releases site.> listed on the [Azure SDK release page](https://azure.github.io/azure-sdk/releases/latest/python.html)</!--> that support Microsoft Entra authentication accept credentials from this library. You can learn more about using these libraries in their documentation, which is <!-- TODO: uncomment when Rust SDK has a release page.>linked from the release page</!-->available at [Docs.rs](https://Docs.rs/azure_data_cosmos).
```
Pilchie, line 181: This seems like it should be on the identity README, not ours?

Pilchie (approved):

heaths:
Yes, we can add `--all-features` *now*. We couldn't before the test-proxy work was MVP and I could remove the `e2e_test` feature @LarryOsterman was using (no better option at the time).

That said, isn't key auth less secure, such that maybe we shouldn't show it in the first thing people see/read? It's fine as an example, but maybe those under `examples/` instead.

sdk/cosmos/azure_data_cosmos/README.md
```diff
@@ -1,5 +1,200 @@
...
+
+**Using Microsoft Entra ID**
+
+```rust
+use azure_identity::DefaultAzureCredential;
+use azure_data_cosmos::CosmosClient;
+
+async fn example() -> Result<(), Box<dyn std::error::Error>> {
+    let credential = DefaultAzureCredential::new()?;
+    let cosmos_client = CosmosClient::new("myAccountEndpointURL", credential, None)?;
```
ronniegeraghty, line 63:
```suggestion
    let cosmos_client = CosmosClient::new("myAccountEndpointURL", credential.clone(), None)?;
```

sdk/cosmos/azure_data_cosmos/README.md
```diff
@@ -1,5 +1,200 @@
...
+
+**NOTE:** Creating a container requires authenticating with an account key, which requires the `key_auth` feature to be enabled. See [Create Cosmos DB Client](#create-cosmos-db-client "Create Cosmos DB client") for more information.
+
+Using the above created database for creating a container, like this:
+
+```rust
+use azure_data_cosmos::models::{ContainerProperties, PartitionKeyDefinition, ThroughputProperties};
+use azure_data_cosmos::CreateContainerOptions;
+
+async fn example(cosmos_client: azure_data_cosmos::CosmosClient) -> Result<(), Box<dyn std::error::Error>> {
```
ronniegeraghty, line 107:
```suggestion
use azure_data_cosmos::{CosmosClient, CreateContainerOptions};

async fn example(cosmos_client: CosmosClient) -> Result<(), Box<dyn std::error::Error>> {
```
Recommend bringing the import of the client into the `use` statement.

sdk/cosmos/azure_data_cosmos/README.md
```diff
@@ -1,5 +1,200 @@
...
+use azure_data_cosmos::models::PatchDocument;
+
+#[derive(Serialize, Deserialize)]
+struct Item {
+    pub id: String,
+    pub partition_key: String,
+    pub value: String,
+}
+
+async fn example(cosmos_client: azure_data_cosmos::CosmosClient) -> Result<(), Box<dyn std::error::Error>> {
```
ronniegeraghty, line 136:
```suggestion
use azure_data_cosmos::{CosmosClient, models::PatchDocument};

#[derive(Serialize, Deserialize)]
struct Item {
    pub id: String,
    pub partition_key: String,
    pub value: String,
}

async fn example(cosmos_client: CosmosClient) -> Result<(), Box<dyn std::error::Error>> {
```

sdk/cosmos/azure_data_cosmos/README.md
ronniegeraghty: Yes, this is specific to the Identity library. Please remove it.

sdk/cosmos/azure_data_cosmos/README.md
```diff
@@ -1,5 +1,200 @@
...
-This client library enables client applications to connect to [Azure Cosmos DB](https://azure.microsoft.com/en-us/products/cosmos-db) via the NoSQL API. Azure Cosmos DB is a globally distributed, multi-model database service.
+[Source code] | [Package (crates.io)] | [API reference documentation] | [Azure Cosmos DB for NoSQL documentation]
+
+## Getting started
+
+### Install the package
+
+Install the Azure Cosmos DB SDK for Rust with cargo:
+
+```bash
```
heaths, line 13: Nit: we've been trying to standardize on `sh` in general, but `azurecli` for `az`-specific command lines. IIRC, that actually does some magic on Microsoft Learn, though I'm not sure how many of these READMEs get up there in a way where that helps.

sdk/cosmos/azure_data_cosmos/README.md
```diff
@@ -1,5 +1,200 @@
...
+cargo add azure_data_cosmos
+```
+
+### Prerequisites
+
+* An [Azure subscription] or free Azure Cosmos DB trial account.
+
+Note: If you don't have an Azure subscription, create a free account before you begin.
+You can Try Azure Cosmos DB for free without an Azure subscription, free of charge and commitments, or create an Azure Cosmos DB free tier account, with the first 400 RU/s and 5 GB of storage for free. You can also use the Azure Cosmos DB Emulator with a URI of https://localhost:8081. For the key to use with the emulator, see [how to develop with the emulator](https://learn.microsoft.com/azure/cosmos-db/how-to-develop-emulator).
```
heaths, line 22: `https` for localhost? Atypical, but doable if they accept to install the dev cert, which most aren't crazy about. Just wanted to ask since we all get in the habit of typing "https" these days.

sdk/cosmos/azure_data_cosmos/README.md
```diff
@@ -1,5 +1,200 @@
...
+* [CRUD operation on Items](#crud-operation-on-items "CRUD operation on Items")
+
+### Create Cosmos DB Client
+
+The clients support different forms of authentication. The Azure Cosmos DB SDK for Rust supports authorization via Microsoft Entra identities or an account key. However, support for account keys is only available when the `key_auth` feature is enabled in your Cargo.toml file.
+
+```toml
+[dependencies]
+azure_data_cosmos = { version = "...", features = ["key_auth"] }
+```
```
heaths, line 51:
````suggestion
cargo add azure_data_cosmos -F key_aith
````
We prefer `cargo add` over hand editing `Cargo.toml`.

File: sdk/cosmos/azure_data_cosmos/README.md [hunk 8]
```diff
@@ -1,5 +1,200 @@
...
+
+We **strongly** recommend using Microsoft Entra ID for authentication, rather than account keys.
```
heaths, line 53: We have some boilerplate text we use that also mentions and links `DefaultAzureCredential`. See `sdk/keyvault/azure_security_keyvault_secrets/README.md` for that blurb. Again, we try to standardize on it across services and languages.

sdk/cosmos/azure_data_cosmos/README.md
```diff
@@ -1,5 +1,200 @@
...
+const COSMOS_DB_KEY: &str = "someKey";
+
+async fn example() -> Result<(), Box<dyn std::error::Error>> {
+    let key = Secret::from(COSMOS_DB_KEY);
+    let cosmos_client = CosmosClient::with_key(COSMOS_DB_ENDPOINT, key, None)?;
+    Ok(())
+}
+```
+
+### Create Database
```
heaths, line 84: We've been creating a TOC of bookmarks, though I'm not particular fond of this (hard to keep maintained, and GitHub does it anyway). @ronniegeraghty are ya'll trying to standardize on that as well?

sdk/cosmos/azure_data_cosmos/README.md
heaths: We have a standard blurb for this that links to Identity. See KV secrets's README and just copy that as needed.

heaths (approved):
Nothing blocking, but a few recommendations.

sdk/cosmos/azure_data_cosmos/README.md
ronniegeraghty: The ToC at the Top of the Examples section? Not something I'm deeply attached to. My though was it would help readers scan through the available examples quickly without having to scroll past some of our longer code blocks.

sdk/cosmos/azure_data_cosmos/README.md
analogrelay: Alas, the Cosmos Emulator is only available over HTTPS, even on localhost. There's details in the linked doc on how to get the cert and install it.

sdk/cosmos/azure_data_cosmos/README.md
heaths: Given GitHub generates a TOC if they're interested (granted, not the most visible - see below), I'd prefer to nix it rather than keep it up to date. I already forgot to add one heading this week.

image

analogrelay:
> That said, isn't key auth less secure, such that maybe we shouldn't show it in the first thing people see/read?

I'm open to removing it from the README, if @TheovanKraay is OK with it. The problem is that there are operations we describe in the README (creating databases) which currently require key auth. We'd have to remove those too.

TheovanKraay:
> > That said, isn't key auth less secure, such that maybe we shouldn't show it in the first thing people see/read?
> 
> I'm open to removing it from the README, if @TheovanKraay is OK with it. The problem is that there are operations we describe in the README (creating databases) which currently require key auth. We'd have to remove those too.

Fine with removing it!

ronniegeraghty (approved):

sdk/cosmos/azure_data_cosmos/README.md
```diff
@@ -1,5 +1,155 @@
...
+#### Authenticate the client
+
+In order to interact with the Azure Cosmos DB service you'll need to create an instance of the `CosmosClient` struct. To make this possible you will need a URL and key of the Azure Cosmos DB service.
+
+## Examples
+
+The following section provides several code snippets covering some of the most common Azure Cosmos DB NoSQL API tasks, including:
+* [Create Client](#create-cosmos-db-client "Create Cosmos DB client")
+* [Create Database](#create-database "Create Database")
+* [Create Container](#create-container "Create Container")
```
ronniegeraghty, line 41: Looks like these links no longer resolve.

analogrelay: Yep, I forgot to remove the TOC items. Thanks!

sdk/core/azure_core_amqp/src/fe2o3/messaging/mod.rs
```diff
@@ -712,7 +712,7 @@ mod tests {
             >::new(
                 seq.into_iter()
                     .map(|x| {
-                        let iter = x.into_iter().map(Into::into);
+                        let iter = x.into_iter();
```
analogrelay, line 715: Clippy was complaining about this being unnecessary. If the tests pass with this change, I'm assuming it's correct.

> error: useless conversion to the same type: `serde_amqp::Value`
   --> sdk/core/azure_core_amqp/src/fe2o3/messaging/mod.rs:715:49
    |
715 |                         let iter = x.into_iter().map(Into::into);
    |                                                 ^^^^^^^^^^^^^^^^ help: consider removing
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
    = note: `-D clippy::useless-conversion` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::useless_conversion)]`

analogrelay: This is resolved with a recent rebase, as it was fixed in `main`

heaths (approved):

PR #2042, opened 2025-01-30
Add release pipeline templates

hallipr:
Add required templates for ci.yml files per service directory


eng/scripts/Pack-Crates.ps1
```diff
@@ -0,0 +1,330 @@
...
+  $metadataLengthBytes = [BitConverter]::GetBytes([UInt32]$metadataBytes.Length)
+  $crateBytes = [IO.File]::ReadAllBytes($CrateFilePath)
+  $crateLengthBytes = [BitConverter]::GetBytes([UInt32]$crateBytes.Length)
+
+  $bytes += $metadataLengthBytes + $metadataBytes + $crateLengthBytes + $crateBytes
+
+  return $bytes
+}
+
+Push-Location $RepoRoot
```
heaths, line 265: Just use `-p` on most `cargo` commands. There's no value in pushing into directories. Just more work for no gain.

eng/scripts/Pack-Crates.ps1
```diff
@@ -0,0 +1,330 @@
...
+    $packageVersion = $package.version
+
+    Invoke-LoggedCommand `
+      -GroupOutput `
+      -Command "cargo package --package $packageName --config `"source.crates-io.replace-with='local'`" --config `"source.local.local-registry='$localRegistryPath'`" --allow-dirty"
+
+    $crateFile = "$RepoRoot/target/package/$packageName-$packageVersion.crate"
+
+    # copy the package to the local registry
+    Add-CrateToLocalRegistry `
```
heaths, line 292: Probably should double check that it exists. Not sure how well-documented this path is and it could change, or does `Add-CrateToLocalRegistry` suitably fail with an obvious message if it doesn't exist?

eng/scripts/Publish-Crates.ps1
```diff
@@ -0,0 +1,44 @@
...
+  [string[]]$CrateNames,
+  [string[]]$AdditionalOwners,
+  [string]$Token
+)
+
+$ErrorActionPreference = 'Stop'
+Set-StrictMode -Version 2.0
+
+function TryAddOwners($crateName) {
+  foreach ($owner in $AdditionalOwners) {
```
heaths, line 15: If these are *additional* owners, where is the initial owner list? `azure-sdk` should always be an owner, for example.

eng/scripts/Publish-Crates.ps1
```diff
@@ -0,0 +1,44 @@
...
+    # ignore errors is owner already exists
+    $body = @{ users = @($owner) } | ConvertTo-Json
+
+    $response = Invoke-WebRequest -Method Put -Uri "https://crates.io/api/v1/crates/$crateName/owners" `
+      -Headers @{ Accept = 'application/json'; Authorization = $Token } `
+      -ContentType 'application/json' `
+      -Body $body`
+      -SkipHttpErrorCheck
+
+    if ($response.StatusCode -ge 400 -and $response.Content -notmatch 'already an owner') {
```
heaths, line 27: Nothing more programmatic than `already an owner`? Seems a bit fragile, but we can always fix it later if it changes. But can we restart publishing in such an event safely enough?

sdk/keyvault/azure_security_keyvault_keys/Cargo.toml
```diff
@@ -14,9 +14,9 @@ keywords = ["sdk", "azure", "rest", "cloud", "keyvault", "security", "keys"]
 categories = ["api-bindings"]
 
 [dependencies]
-azure_core = { workspace = true }
+azure_core = { path = "../../core/azure_core" }
```
heaths, line 17: This is all generated initially, and something @jhendrixMSFT will have to change in the emitter.

sdk/template/azure_template/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+pub fn add(left: u64, right: u64) -> u64 {
```
heaths, line 1: I know this is generated initially, but we should still put the copyright headers in the source files so someone doesn't forget later. A couple of us have plans to put more in here for the very purpose of treating this crate as a canary. /cc @HarshaNalluru

sdk/template/azure_template_core/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+pub fn add(left: u64, right: u64) -> u64 {
```
heaths, line 1: Same here.

sdk/template/azure_template_core/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+
+[dependencies]
+
+[dev-dependencies]
+azure_template.path = "../azure_template"
```
heaths, line 14: `azure_template` should depend on `azure_template_core`, actually, to fit with our convention.

heaths (changes requested):
Most comments are nits or things I don't necessarily need to re-review, I feel, but I really want to see the dependencies in the root Cargo.toml fixed.

Cargo.toml
```diff
@@ -26,42 +28,14 @@ repository = "https://github.com/azure/azure-sdk-for-rust"
 rust-version = "1.80"
 
 [workspace.dependencies.typespec]
-path = "sdk/typespec"
-version = "0.2.0"
-default-features = false
-
-[workspace.dependencies.typespec_client_core]
```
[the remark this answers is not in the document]
hallipr: It follows from:
1. A version only dependency is for depending on a released version
2. A path based dependency is for depending on an unreleased version
3. Workspace versions are a central declaration of the default version (and features) that other workspace packages should depend on.
4. The version in the workspace Cargo.toml should be a released version (not always the most recently released version)
5. Depending on unreleased versions is a per-package decision that should be reflected in per-package dependencies.

typespec_client_core has no released version.  We can say that either:
A.  "completely unreleased packages" can have a workspace path dependency because any consumer must necessarily depend on the unreleased version.
B.  there's no difference between unreleased crates and initial unreleased versions and rule 5. stands.

I went with B

eng/pipelines/templates/jobs/ci.yml
```diff
@@ -1,114 +1,146 @@
...
+
+  pool:
+    name: $(LINUXPOOL)
+    image: $(LINUXVMIMAGE)
+    os: linux
+
+  steps:
+  - checkout: self
+    ${{ if eq(variables['Build.Reason'], 'PullRequest') }}:
+      fetchDepth: 0 # Fetch all history for PR builds
```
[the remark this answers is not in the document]
hallipr: PRs require depth for diff detection.  CIs just require building the current state.

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -0,0 +1,158 @@
...
+          
+          - template: /eng/common/pipelines/templates/steps/create-tags-and-git-release.yml
+            parameters:
+              ArtifactLocation: $(Pipeline.Workspace)/${{parameters.PipelineArtifactName}}/${{artifact.name}}
+              PackageRepository: Crates.io
+              ReleaseSha: $(Build.SourceVersion)
+
+        - deployment: PublishPackage
+          displayName: "Publish to Crates.io"
+          condition: and(succeeded(), ne(variables['Skip.PublishPackage'], 'true'))
```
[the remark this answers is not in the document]
hallipr: This is consistent.  Publishing it gated on approval through the `environment:`.  By default, we don't skip publishing in a release stage.

eng/scripts/Language-Settings.ps1
```diff
@@ -1,23 +1,20 @@
 $Language = "rust"
 $LanguageDisplayName = "Rust"
 $PackageRepository = "crates.io"
-$packagePattern = "Cargo.toml"
+$packagePattern = "cargo-metadata.json"
```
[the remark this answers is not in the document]
hallipr: It's generated by calling Pack-Crates.ps1.  It's the json metadata portion of the request sent to crates.io when publishing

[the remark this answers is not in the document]
hallipr: Because it's natively parsable in powershell, it's a better metadata format for the engineering system than the toml file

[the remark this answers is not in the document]
hallipr: $packagePattern is only used when searching for packages in build artifacts

eng/pipelines/templates/jobs/ci.yml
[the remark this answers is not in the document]
heaths: Gotcha. But you should only need the previous commit. Remember, git doesn't store diffs like TFS (and most others) did. It stores snapshots, so any diff would appear against `HEAD~`.

eng/scripts/Language-Settings.ps1
```diff
@@ -120,3 +115,44 @@ function Get-AllPackageInfoFromRepo ([string] $ServiceDirectory)
 
   return $allPackageProps
 }
+
+function Get-rust-PackageInfoFromPackageFile([IO.FileInfo]$pkg, [string]$workingDirectory) {
+  #$pkg will be a FileInfo object for the cargo-metadata.json file in a package artifact directory
+  $package = Get-Content -Path $pkg.FullName -Raw | ConvertFrom-Json
+  $packageName = $package.name
+  $packageVersion = $package.vers
+  $docsReadMeName = $packageName -replace "^azure_" , ""
```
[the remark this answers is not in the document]
hallipr: I'm honestly not sure.  Part of the eng/common cargo culture.  .NET was setting it like `$docsReadMeName = $pkgId -replace "^Azure." , ""` and I just copied the pattern.

It doesn't look like there's any code in eng/common that consumes the resulting `package.DocsReadMeName` property

eng/scripts/Language-Settings.ps1
```diff
@@ -120,3 +115,44 @@ function Get-AllPackageInfoFromRepo ([string] $ServiceDirectory)
...
+
+  return New-Object PSObject -Property @{
+    PackageId      = $packageName
+    PackageVersion = $packageVersion
+    ReleaseTag     = "$packageName`_$packageVersion"
+    Deployable     = $existingVersions -notcontains $packageVersion
+    ReleaseNotes   = $releaseNotes
+    ReadmeContent  = $readmeContent
+    DocsReadMeName = $docsReadMeName
+  }
```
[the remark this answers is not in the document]
hallipr: >All objects sent through the pipeline are returned

agreed, but it's just visually odd to interpret an object literal statement as output.

eng/scripts/Pack-Crates.ps1
hallipr: I do use `--package $packageName` on the cargo commands, I just do it from the repo root

eng/scripts/Pack-Crates.ps1
hallipr: thanks.  I added a test-path sanity check after the assignment

eng/scripts/Publish-Crates.ps1
hallipr: `azure-sdk` is the publisher and initial owner of any crate it publishes

eng/pipelines/pr.yml
```diff
@@ -10,3 +10,5 @@ pr:
```
weshaggard, line 10: Could you as a follow-up rename pr.yml to pullrequest.yml and update the pipeline name similarly to align with the other languages.

eng/pipelines/templates/jobs/ci.yml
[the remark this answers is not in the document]
weshaggard: Should we use sparse-checkout here?

eng/scripts/Publish-Crates.ps1
hallipr: It returns a 400 status code with no error specific headers and the response body:
```json
{"errors":[{"detail":"`hallipr` is already an owner"}]}
```

sdk/keyvault/azure_security_keyvault_keys/Cargo.toml
hallipr: It depends on how we want to treat azure_core.  I think we decided that we'll treat it as a workspace managed dependency, including whether packaged depend on the released or unreleased version.

In which case, all azure_core dependencies can be reverted back to `{ workspace = true }`

sdk/template/azure_template_core/Cargo.toml
hallipr: It does, but we also need to model the circular dev-dependency

eng/pipelines/templates/jobs/ci.yml
```diff
@@ -1,114 +1,146 @@
...
+    ${{ else }}:
+      fetchDepth: 1 # Fetch only the latest commit for non-PR builds
+
+  - ${{ if ne(variables['Build.Reason'], '') }}:
+    - script: |
+        echo "##vso[build.addbuildtag]$(Build.Reason)"
```
weshaggard, line 73: IIRC this doesn't work in public builds.

eng/pipelines/templates/jobs/ci.yml
```diff
@@ -1,114 +1,146 @@
...
+    - template: /eng/common/pipelines/templates/steps/set-test-pipeline-version.yml
+      parameters:
+        PackageNames: ${{ join(',', parameters.Artifacts.*.name) }}
+        ServiceDirectory: ${{parameters.ServiceDirectory}}
+        TestPipeline: true
+
+  - template: /eng/pipelines/templates/steps/pack-crates.yml
+    parameters:
+      ServiceDirectory: ${{ parameters.ServiceDirectory }}
+      Toolchain: stable
```
weshaggard, line 86: Isn't this the default?

hallipr: yep.  thanks

eng/pipelines/pr.yml
hallipr: created issue #2084

eng/pipelines/templates/jobs/ci.yml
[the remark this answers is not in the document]
hallipr: We should. I'll switch to sparse checkout everywhere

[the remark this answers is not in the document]
hallipr: > But you should only need the previous commit

We would need all the commits between HEAD and the merge base of HEAD and the PR target branch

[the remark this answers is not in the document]
hallipr: As each commit is just a tree and `git diff` should only need the current tree and the tree the PR originated from to know the net change of the PR.

eng/scripts/Language-Settings.ps1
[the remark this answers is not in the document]
weshaggard: @danieljurek might know if this is still used anywhere. I think it was used for docs in some point.

eng/scripts/Publish-Crates.ps1
heaths: It'll have to do.

sdk/template/azure_template_core/Cargo.toml
heaths: Ah, gotcha. I missed that.

eng/scripts/Language-Settings.ps1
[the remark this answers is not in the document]
danieljurek: The `$docsReadMeName` is used when publishing "package-level overview" pages (pulled from a package's README.md file). The filename specifically removes the word `azure` from the package name to prevent repeating `azure` in the page URL. 

For example: 

The URL for the package-level overview for `Azure.Security.KeyVault.Secrets` is https://learn.microsoft.com/en-us/dotnet/api/overview/azure/security.keyvault.secrets-readme

Note that the URL ends with `azure/security.keyvault.secrets-readme`. If we did not remove the preceding `azure` in the package name then the URL would end with `azure/azure.security.keyvault.secrets-readme` which was determined to be too repetitive for users seeing the URLs. 

At this point, Rust packages are not onboarded to learn.microsoft.com so this variable doesn't serve any purpose at this time.

sdk/keyvault/azure_security_keyvault_keys/Cargo.toml
heaths: I honestly can't remember. We've been back and forth so many times about which is better, this is why I asked for a write-up of what we discussed - what seems to have the best pro/con ratio considering our options.

RickWinter:
@hallipr Is there an ETA for this work to complete?

sdk/keyvault/azure_security_keyvault_keys/Cargo.toml
hallipr: I've removed all toml edits from this PR.  we can address them in a new pr

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -0,0 +1,168 @@
...
+              displayName: Yank Crates
+              inputs:
+                targetType: filePath
+                filePath: $(Build.SourcesDirectory)/eng/scripts/Yank-Crates.ps1
+                arguments:
+                  -CrateNames '${{artifact.name}}'
+                  -PackagesPath '$(Pipeline.Workspace)/${{parameters.PipelineArtifactName}}'
+                  -Token $(azure-sdk-cratesio-token)
+
+        # - job: PublishDocsMs
```
heaths, line 160: I'd be shocked if this was even supported. Rust docs are a very different beast for DocFx. Is this even a requirement?

/cc @danieljurek

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -0,0 +1,168 @@
...
+
+        # - job: PublishDocs
```
heaths, line 162: This shouldn't be necessary. Publishing to crates.io does it automatically e.g., https://github.com/heaths/ordinal-rs/blob/5b4254956cca64de22e5243a57e2686ace85c2e7/.github/workflows/release.yml#L38-L41

eng/pipelines/templates/steps/use-rust.yml
```diff
@@ -14,6 +17,22 @@ steps:
...
-      rustup update --no-self-update $toolchain
+      $attempts = 0
+      $maxAttempts = ${{ parameters.MaxAttempts }}
+
+      while ($true) {
+        $attempts++
+        Write-Host "> rustup toolchain install --no-self-update $toolchain"
+        rustup toolchain install --no-self-update $toolchain
+        
+        if ($LASTEXITCODE -eq 0) { exit 0 }
```
heaths, line 28: Tip:
```suggestion
        if ($?) { exit 0 }
```

eng/scripts/Pack-Crates.ps1
```diff
@@ -0,0 +1,297 @@
...
+$ErrorActionPreference = 'Stop'
+
+. (Join-Path $PSScriptRoot '..' 'common' 'scripts' 'common.ps1')
+. (Join-Path $EngCommonScriptsDir 'Helpers' 'CommandInvocation-Helpers.ps1')
+
+if ($OutputPath) {
+  $OutputPath = New-Item -ItemType Directory -Path $OutputPath -Force | Select-Object -ExpandProperty FullName
+}
+
+function Get-OutputPackageNames($workspacePackageNames) {
```
heaths, line 21: Just based on the name, shouldn't this filter out `publish = false` as well?

eng/scripts/Pack-Crates.ps1
```diff
@@ -0,0 +1,297 @@
...
+
+function Get-CargoPackages() {
+  $metadata = Get-CargoMetadata
+
+  # path based depdenencies are assumed to be unreleased package versions
+  # they must be included in this build and build before packages that depend on them
+  foreach ($package in $metadata.packages) {
+    $package.UnreleasedDependencies = @()
+    foreach ($dependency in $package.dependencies) {
+      if ($dependency.path -and $dependency.kind -ne 'dev') {
```
heaths, line 64: FWIW, this could break down in cases like:

```toml
[target.'cfg(target_arch = "wasm32")'.dev-dependencies]
# ...
```

Maybe. I guess it depends on what is populating `kind` and if they correctly categories syntax like that above. We do actually have some of that and, I have no doubt, will have more in the future.

eng/scripts/Yank-Crates.ps1
```diff
@@ -0,0 +1,34 @@
...
+
+$hasErrors = $false
+foreach ($crateName in $crateNames) {
+  $crate = Get-Content "$PackagesPath/$crateName/cargo-metadata.json" -Raw | ConvertFrom-Json
+  $crateVersion = $crate.vers
+
+  Write-Host "Yanking crate: '$crateName@$crateVersion'"
+
+  # https://doc.rust-lang.org/cargo/reference/registry-web-api.html#yank
+  $response = Invoke-WebRequest -Method Delete -Uri "https://crates.io/api/v1/crates/$crateName/$crateVersion/yank" `
```
heaths, line 21: Interesting: I'm surprised they take a `DELETE` to a ".../yank" URI instead of the resource (version) itself. That's just...not very RESTful. /cc @JeffreyRichter

heaths (approved):
Nothing blocking. LGTM. Mostly just a few comments for posterity.

eng/pipelines/templates/stages/archetype-rust-release.yml
hallipr: Creating issue to remove docs comments from yml

eng/pipelines/templates/stages/archetype-rust-release.yml
hallipr: excelent.  I'll remove the comments in following PR

eng/scripts/Pack-Crates.ps1
hallipr: Will create an issue for future PR

eng/scripts/Yank-Crates.ps1
hallipr: They're very not restful

eng/pipelines/templates/stages/archetype-rust-release.yml
danieljurek: Docs for learn.microsoft.com are not in scope for rust at this time. In the future this is where that work would go.

PR #2047, opened 2025-01-31
Update Key Vault Secrets README

### LINKED ISSUE #2040: Key Vault Secrets README
The `README.md` file for the Azure KeyVault Screts crate should follow our template for service library `README.md`s. 
Exmaple: [.NET's Azure KeyVault Secrets README](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/Azure.Security.KeyVault.Secrets/README.md)

ronniegeraghty:
Adding initial Key Vault Secrets create `README.md`.

Fixes #2040


sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,438 @@
...
+    println!("{}", updated_secret_properties.version);
+    println!("{}", updated_secret_properties.content_type.unwrap());
+
+    Ok(())
+}
+```
+
+### Delete a secret
+
+`start_delete_secret` starts a long-running operation to delete a secret previously stored in the Azure Key Vault. You can retrieve the secret immediately without waiting for the operation to complete. When [soft-delete] is not enabled for the Azure Key Vault, this operation permanently deletes the secret.
```
ronniegeraghty, line 232: Is this true? It's copied from .NET and I assumed if the delete secret API requires an LRO this would still be true for the rust crate.

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,489 @@
...
+        ).await?;
+    println!("Response Code: {:?}", response.status());
+
+    Ok(())
+}
+```
+
+### Delete and purge a secret
+
+You will need to wait for the long-running operation to complete before trying to purge or recover the secret. You can do this by calling `update_status` in a loop as shown below:
```
ronniegeraghty, line 284: This is taken from the .NET KeyVault Secrets `README`. Since the rust crate's delete method returns a response, I am unsure how we would query if the operation has completed.

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,499 @@
...
+    let credential = DefaultAzureCredential::new()?;
+    let client = SecretClient::new(
+        "https://your-key-vault-name.vault.azure.net/",
+        credential,
+        None,
+    )?;
+
+    // List secrets using the secret client.
+    let response: Pager<SecretListResult>= client.get_secrets(None)?;
+    println!("{:?}", response);
```
ronniegeraghty, line 337: What is the proper way to work with the Pager struct? I saw it has a stream property but it's private. And no other methods were showing through IntelliSense on the struct.

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,429 @@
-# Azure KeyVault Secrets Client
+# Azure Key Vault Secret client library for Rust
```
heaths, line 1: "secret" is not part of the product name and shouldn't be capitalized. I don't do that in .NET either.
```suggestion
# Azure Key Vault secret client library for Rust
```

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,429 @@
...
+
+The Azure Key Vault Secrets client library allows you to securely store and control the access to tokens, passwords, API keys, and other secrets. This library offers operations to create, retrieve, update, delete, purge, backup, restore, and list the secrets and its versions.
+
+[Source code] | [Package (crates.io)] | [API reference documentation] | [Product documentation]
 
 ## Getting started
 
+### Install the package
+
+Install the Azure Key Vault Secrets client library for Rust with [Cargo]:
```
heaths, line 13: I recommend lower-casing "secrets" everywhere.
```suggestion
Install the Azure Key Vault secrets client library for Rust with [Cargo]:
```
Best to compare with [.NET's](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/Azure.Security.KeyVault.Secrets/README.md). That was the rubric when I lead the KV v-team.

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,429 @@
...
+
+```azurecli
+az login
+```
+
+Instantiate a `DefaultAzureCredential` to pass to the client. The same instance of a token credential can be used with multiple clients if they will be authenticating with the same identity.
+
+### Set and Get a Secret
+
+```rust
```
heaths, line 55: You need to mark these as `no_run` or tests will try to execute them:

````markdown
```rust no_run
// ...
```
````

See the [guidelines](https://azure.github.io/azure-sdk/rust_introduction.html#rust-documentation-readme) for more information.

File: sdk/keyvault/azure_security_keyvault_secrets/README.md [hunk 7]
```diff
@@ -1,24 +1,429 @@
...
+    let credential = DefaultAzureCredential::new()?;
+    let client = SecretClient::new(
+        "https://your-key-vault-name.vault.azure.net/",
+        credential,
+        None,
+    )?;
+
+    // Create a new secret using the secret client.
+    let mut secret_set_parameters = SecretSetParameters::default();
+    secret_set_parameters.value = Some("secret-value".to_string());
```
heaths, line 76: Easier and idiomatic:
```suggestion
    let secret_set_parameters = SecretSetParameters {
        value: Some("secret-value".to_string()),
        ..Default::default()
    };
```

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,429 @@
...
+
+    // Serialize secret_set_parameters to Vec<u8>
+    let secret_set_parameters_bytes: Vec<u8> = to_vec(&secret_set_parameters)?;
+
```
heaths, line 80: No necessary. More work.

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,429 @@
...
+    let secret: Response<SecretBundle> = client
+        .set_secret(
+            "secret-name".to_string(),
+            RequestContent::from(secret_set_parameters_bytes),
```
heaths, line 84: More work. You also don't need to import `RequestContent`:
```suggestion
            secret_set_parameters.try_into()?,
```

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,429 @@
...
+            None,
+        )
+        .await?;
+    println!("Response Code: {:?}", secret.status());
```
heaths, line 88: Why? Not really useful. It either worked and you print the secret or it didn't.

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,429 @@
...
+
+    Ok(())
+}
+```
+
+## Key concepts
+
+### KeyVaultSecret
+
+A `KeyVaultSecret` is the fundamental resource within Azure Key Vault. From a developer's perspective, Azure Key Vault APIs accept and return secret values as strings.
```
heaths, line 110: We don't have this type currently but should to align with other languages. I thought @RickWinter was going to do some updates in the `client.tsp` to better align with other languages.

heaths: It's called `SecretBundle` currently, but may also be `SecretItem` in some cases. Without hand-writing a client, though, we can't do that with a pure emitted client.

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,429 @@
...
+
+### SecretClient
+
+A `SecretClient` asynchronous operations in the SDK allowing for selection of a client based on an application's use case. Once you've initialized a `SecretClient`, you can interact with secrets in Azure Key Vault.
```
heaths, line 114: "A" infers singular but "operations" is plural.
```suggestion
`SecretClient` asynchronous operations in the SDK allowing for selection of a client based on an application's use case. Once you've initialized a `SecretClient`, you can interact with secrets in Azure Key Vault.
```
This also doesn't make sense. It's the only client. To what selection do you refer? Did you start from Python or Java? Again, for KV anyway, I recommend basing it on .NET. I typically wrote the README first and shared it with other languages.

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,429 @@
...
+        credential,
+        None,
+    )?;
+
+    // Create a new secret using the secret client.
+    let mut secret_set_parameters = SecretSetParameters::default();
+    secret_set_parameters.value = Some("secret-value".to_string());
+
+    // Serialize secret_set_parameters to Vec<u8>
+    let secret_set_parameters_bytes: Vec<u8> = to_vec(&secret_set_parameters)?;
```
heaths, line 158: All these are more complicated than necessary. I can fix them up.

sdk/keyvault/azure_security_keyvault_secrets/README.md
heaths: All languages return a response. There's just no model. You still `.await` the response.

heaths (changes requested):

heaths:
See <https://github.com/Azure/azure-sdk-for-rust/pull/2051/files#diff-e68c23fe7f799586bbbcaf760ba0cd25ac804e07b08c0c4519192f0ae3d00a2a> for some examples.

The pager needs work. Especially if we're making all fields `Option<T>`, which puts a burden on pageable responses where I think instead we should deserialize to an empty `Vec` instead of `None`. It also doesn't loop over all pages like it did in track 1 and I want to bring that back. Going through page by page necessarily is a pain and one I think will trip up some devs because, for Key Vault, listing resources could return an empty page but also a `next_link`. It happens. If they get no values they may terminate their loop early given how loops typically work i.e., `Iterator::next()` returns `None` for an item and they assume all items have been enumerated.

/cc @JeffreyRichter @RickWinter

sdk/keyvault/azure_security_keyvault_secrets/README.md
JeffreyRichter: The docs look broken to me. I have some back-story on this. DELETE is not spec'd (in swagger and probably not in TypeSpec either) as a proper LRO. I think that when KeyVault was first created many years ago, the LRO pattern didn't exist. However, deleting a secret is not immediate and is not guaranteed to complete with the DELETE operation returns. So, in other SDKs, we manually wrote code to poll continuously GETtng the secret until 404 returns (or something like this - I forget exactly what the heuristic is). 

But, whatever the heuristic, there is a race condition and implied promise to the customer that we can't keep. So, I'd say that we should expose whatever the service contract actually is and not have our SDK simulate magical LRO functionality. whose behavior we can't guarantee. Also, I believe that most customers simply won't care: they will DELETE a secret, and their app will just move on; the secret will be deleted eventually, why does the app have to wait for it to be fully deleted?

And finally, if the service team thinks there is an issue here with their contract or implementation, then the service team should fix it to benefit ALL their customers which includes customers using an SDK as well as customers not using any SDK.

JeffreyRichter:
I'm almost never a fan of SDK magic. The service offers a "get a page" operation and we need to expose this. The SDK can expose it with a pageable type that can optionally be used to get the nextLink page. In a single app, it is likely (but not mandatory; see example at bottom) that the app will loop over the page of items and then get the next page to loop of its results, and so on. 

But knowing where in the code the high latency with many possible failures that might need recovery is happening (the get page operations) and knowing where in the code the low latency and no chance of failures (iterating over the page of items) is happening is critical especially in a language like Rust. Obfuscating this with some magical iterator that combines these 2 very different things is a mistake. If a customer wants to write code to obfuscate these 2 things to make their code "easier" (while hiding failure modes and latency issues), that is OK but our SDK should not encourage or force this obfuscation. I'll also add that a page with no items is actually how the service works, and we should not hide this either. We MUST document it, so customers write code properly to deal with an empty page, but we do not fix it for them (as ding so hide additional latency and failures).

Another reason not to magically conflate get-a-page & iterate-over-a-page is because it is also fairly common to have one PC initiate listing, process the page, and then another PC processes the next page. This is why our pager types have a way of getting a resume token and re-hydrating a pager from this resume token. I can give a real-life concrete example using 2 PCs and a resume token: Imagine an e-commerce website (like Amazon). The customer searches for some product, a server PC initiates the search, gets results and constructs an HTML page returning it to the browser. This webpage has the resume token embedded in it. 

The customer now clicks a button to look at the next page of results. When they click this button, a new web request is sent back to the service with the resume token where it goes through a load balancer and reaches a different server PC. This different server PC needs to rehydrate the pager type from the resume token and get the next page in order to return a new HTML page with the 2nd page of results and the latest resume token embedded in it so the customer could click "next page" again if they desire.

For KeyVault, imagine a Portal (like the Azure Portal) showing 1 page of secrets/certificates at a time. My example above explains is exactly how this would be implemented.

sdk/keyvault/azure_security_keyvault_secrets/README.md
RickWinter: Also marketing uses a plural `Secrets`

RickWinter: Should we use `library` or `crate` terminology?

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,429 @@
-# Azure KeyVault Secrets Client
+# Azure Key Vault Secret client library for Rust
 
-The `azure_security_keyvault_secrets` crate.
-These crates follow the [Azure SDK Design Guidelines for Rust](https://azure.github.io/azure-sdk/rust_introduction.html).
+Azure Key Vault is a cloud service that provides a secure storage of secrets, such as passwords and database connection strings.
+
+The Azure Key Vault Secrets client library allows you to securely store and control the access to tokens, passwords, API keys, and other secrets. This library offers operations to create, retrieve, update, delete, purge, backup, restore, and list the secrets and its versions.
```
RickWinter, line 5: In rust they are crates.

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,429 @@
...
+
+Add the following crates to your project:
+
+```sh
+cargo add azure_identity azure_core serde_json tokio
+```
+
+### Authenticate the client
+
+In order to interact with the Azure Key Vault service, you'll need to create an instance of the `SecretClient` struct. You need a **vault url**, which you may see as "DNS Name" in the portal, and credentials to instantiate a client object.
```
RickWinter, line 41:
```suggestion
In order to interact with the Azure Key Vault service, you'll need to create an instance of the `SecretClient`. You need a **vault url**, which you may see as "DNS Name" in the portal, and credentials to instantiate a client object.
```

sdk/keyvault/azure_security_keyvault_secrets/README.md
heaths: Barring more feedback about this, I agree with @JeffreyRichter. There's no evidence customers really use this. We may be the only ones in our tests, and I don't plan to exercise this code regularly in our tests because it just eats up time. Cleaning up secrets also implies that we can't run integration tests concurrently and we should be better than that: if we are checking for a collection of secrets for e.g., check only for those names and ignore everything else. Use unique names, etc.

sdk/keyvault/azure_security_keyvault_secrets/README.md
heaths: I was told "no" in the past. Across languages we refer to these as "client libraries". It's that way in the general guidelines as well. As I was told, this is a more abstract concept than whatever we call the container in which the implementation is contained.

heaths:
@JeffreyRichter wrote,

> Another reason not to magically conflate get-a-page & iterate-over-a-page is because it is also fairly common to have one PC initiate listing, process the page, and then another PC processes the next page.

Based on what? Even if we ignore client applications, many service schedulers I've seen documented seem to shuffle clients off to one machine or another for persistent connections. While I agree we should support rehydratable pagers, the fact we don't have them in all languages - and for a time didn't in any languages - and they still worked proves that they, well, work. While anecdotal, I didn't see any user feedback for .NET about wanting rehydratable pagers either.

As stated in another issue in which we discussed this, I agree we shouldn't hide getting page by page but iterating over a `Stream` for a single request the user made is also idiomatic in Rust as it is in .NET. Manually getting the `next_link` and making a different call is not idiomatic. Iteration is a major feature of Rust - as I often half-joke, `std` is pretty much there to support iteration and little else that people actually use - and making it harder to use is not idiomatic.

Still, part of my [feedback](https://github.com/Azure/azure-sdk-for-rust/pull/2047#issuecomment-2629310748) above was also that returning an `Option<Vec<T>>` for `value` makes this API - even without a "magic" pager - much harder to use. I recommend we instead - for list results - return an empty `Vec<T>`. That would also reduce some of my worry about how people might interpret `None` here as well.

heaths:
@ronniegeraghty I updated `main` with extension methods that will help clean things up. See #2052 and how I use them in #2051

sdk/keyvault/azure_security_keyvault_secrets/README.md
ronniegeraghty: My question was more on if we need to wait for the (not really an LRO) LRO to finish before the secret can be purged? 
Or if we even do have to wait for that?

sdk/keyvault/azure_security_keyvault_secrets/README.md
ronniegeraghty: Changed all instances of "Secrets" to "secret" and made sure it's always plural when referencing the service branding.

sdk/keyvault/azure_security_keyvault_secrets/README.md
ronniegeraghty: Done

sdk/keyvault/azure_security_keyvault_secrets/README.md
ronniegeraghty: Was being blocked by the `#[non_exhaustive]` on this before but I believe it's been removed now.

sdk/keyvault/azure_security_keyvault_secrets/README.md
ronniegeraghty: I did take from .NET. I think the wording is weird because .NET has sync and async operations.

sdk/keyvault/azure_security_keyvault_secrets/README.md
ronniegeraghty: @heaths, 
I tested out the delete and purge scenario and I get the following error: 
```
Error: Error { context: Full(Custom { kind: HttpResponse { status: Conflict, error_code: Some("Conflict") }, error: HttpError { status: Conflict, details: ErrorDetails { code: Some("Conflict"), message: Some("Secret is currently being deleted.") }, .. } }, "server returned error status which will not be retried: 409") }
```
Looks like you do need to wait for the delete to complete before running the purge. 
Here's some sample code I wrote to do the polling Jeff mentioned: 
``` rust
use azure_identity::DefaultAzureCredential;
use azure_security_keyvault_secrets::{models::SecretSetParameters, ResourceExt, SecretClient};

#[tokio::main]
    let credential = DefaultAzureCredential::new()?;
    let client = SecretClient::new(
        "https://ronnieg-keyvault.vault.azure.net/",
        credential,
        None,
    )?;

    println!("Setting a secret...");
    // Set a secret using the secret client.
    let set_secret_parameters = SecretSetParameters {
        value: Some("secret-value".into()),
        ..Default::default()
    };

    let secret = client
        .set_secret(
            "secret-name".into(),
            set_secret_parameters.try_into()?,
            None,
        )
        .await?
        .into_body()
        .await?;

    println!("Deleting the secret...");
    // Delete a secret using the secret client.
    client
        .delete_secret(secret.resource_id()?.name, None)
        .await?;

    println!("Attempting to get the secret until it is deleted...");
    loop {
        // Attempt to get secret until it is deleted
        let response = client
            .get_secret(
                secret.resource_id()?.name,
                secret.resource_id()?.version.unwrap_or_default(),
                None,
            )
            .await;

        if response.is_err() {
            break;
        }
    }

    println!("Purging the deleted secret...");
    // Purge deleted secret using the secret client.
    client
        .purge_deleted_secret(secret.resource_id()?.name, None)
        .await?;

    println!("Operation completed successfully.");
    Ok(())
}
```

This worked for me. In the README we can take out the section to set the secret.

heaths: That's why we generated pseudo-LROs for other languages. I mean, how bad do we need to document this? Hardly anyone uses it. If you want, I can write up a recorded test for it.

sdk/keyvault/azure_security_keyvault_secrets/README.md
heaths: I removed it manually. @jhendrixMSFT were you going to try to do this change before we release beta.1?

jhendrixMSFT: Yeah it's there now in `v0.8.0` of the emitter.

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,419 @@
...
+        .into_body()
+        .await?;
+
+    Ok(())
+}
+```
+
+### Delete a secret
+
+`delete_secret` starts a long-running operation to delete a secret previously stored in the Azure Key Vault. You can retrieve the secret immediately without waiting for the operation to complete. When [soft-delete] is not enabled for the Azure Key Vault, this operation permanently deletes the secret.
```
heaths, line 252: 1. Soft delete is enabled by default and cannot be disabled anymore.
2. Mention of an LRO might make people think we have client support but won't.

So maybe instead,

```suggestion
`delete_secret` will tell Key Vault to delete a secret but it is not deleted immediately. It is not deleted until the service-configured data retention period - the default is 90 days - or until you call `purge_secret` on the returned `DeletedSecretBundle.id`.
```

sdk/keyvault/azure_security_keyvault_secrets/README.md
ronniegeraghty: I'm fine removing this example. I think we have enough examples without it. We can wait for user feedback on this, then add documentation on how to handle it.

sdk/keyvault/azure_security_keyvault_secrets/README.md
ronniegeraghty: Resolved with @heaths's example from the tests

sdk/keyvault/azure_security_keyvault_secrets/README.md
heaths: I'd recommend instead something like,
```suggestion
`purge_secret` will immediately delete a secret using the `DeletedSecretBundle.id` returned from `delete_secret` unless the service is configured with purge protection, which is disabled by default for Key Vault.

You may not be able to delete a secret immediately after calling `delete_secret`, however. Key Vault may take some time to actually delete the secret, so it's best to retry purging in a loop.
```

heaths (approved):
A few nits but otherwise LGTM

sdk/keyvault/azure_security_keyvault_secrets/README.md
ronniegeraghty: Added suggestion

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -249,7 +249,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
 ### Delete a secret
 
-`delete_secret` starts a long-running operation to delete a secret previously stored in the Azure Key Vault. You can retrieve the secret immediately without waiting for the operation to complete. When [soft-delete] is not enabled for the Azure Key Vault, this operation permanently deletes the secret.
+`delete_secret` starts a long-running operation to delete a secret previously stored in the Azure Key Vault. You can retrieve the secret immediately without waiting for the operation to complete.
```
heaths: Please use my recommended text. This still doesn't make sense: "You can retrieve the secret immediately without waiting for the operation to complete."

heaths: Maybe GitHub is doing something funky here, but it looks like you previously did - though I didn't see it - and now just restored back to what you had.

ronniegeraghty: Hey @heaths, 
I think this comment might have been made against an old commit. I removed that old text and used your first suggestion in e0fb752.

I updated the text again in 411eaab to this new suggestion.

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,393 @@
...
+        .into_body()
+        .await?;
+
+    Ok(())
+}
+```
+
+### Delete a secret
+
+`delete_secret` starts a long-running operation to delete a secret previously stored in the Azure Key Vault. You can retrieve the secret immediately without waiting for the operation to complete.
```
heaths, line 252:
```suggestion
`delete_secret` will tell Key Vault to delete a secret but it is not deleted immediately. It will not be deleted until the service-configured data retention period - the default is 90 days - or until you call `purge_secret` on the returned `DeletedSecretBundle.id`.
```

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,393 @@
...
+    ResourceExt, SecretClient,
+};
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Create a new secret client
+    let credential = DefaultAzureCredential::new()?;
+    let client = SecretClient::new(
+        "https://your-key-vault-name.vault.azure.net/",
+        credential,
```
heaths, line 68:
```suggestion
        credential.clone(),
```

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,393 @@
...
+    models::{SecretBundle, SecretSetParameters},
+    ResourceExt, SecretClient,
+};
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let credential = DefaultAzureCredential::new()?;
+    let client = SecretClient::new(
+        "https://your-key-vault-name.vault.azure.net/",
+        credential,
```
heaths, line 143:
```suggestion
        credential.clone(),
```

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,393 @@
...
+```rust no_run
+use azure_identity::DefaultAzureCredential;
+use azure_security_keyvault_secrets::{models::SecretBundle, SecretClient};
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let credential = DefaultAzureCredential::new()?;
+    let client = SecretClient::new(
+        "https://your-key-vault-name.vault.azure.net/",
+        credential,
```
heaths, line 187:
```suggestion
        credential.clone(),
```

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,393 @@
...
+    SecretClient,
+};
+use std::collections::HashMap;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let credential = DefaultAzureCredential::new()?;
+    let client = SecretClient::new(
+        "https://your-key-vault-name.vault.azure.net/",
+        credential,
```
heaths, line 221:
```suggestion
        credential.clone(),
```

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,393 @@
...
+```rust no_run
+use azure_identity::DefaultAzureCredential;
+use azure_security_keyvault_secrets::SecretClient;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let credential = DefaultAzureCredential::new()?;
+    let client = SecretClient::new(
+        "https://your-key-vault-name.vault.azure.net/",
+        credential,
```
heaths, line 263:
```suggestion
        credential.clone(),
```

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,393 @@
...
+use azure_security_keyvault_secrets::{ResourceExt, SecretClient};
+use futures::TryStreamExt;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Create a new secret client
+    let credential = DefaultAzureCredential::new()?;
+    let client = SecretClient::new(
+        "https://your-key-vault-name.vault.azure.net/",
+        credential,
```
heaths, line 291:
```suggestion
        credential.clone(),
```

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -1,24 +1,393 @@
...
+use azure_identity::DefaultAzureCredential;
+use azure_security_keyvault_secrets::SecretClient;
+
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let credential = DefaultAzureCredential::new()?;
+    let client = SecretClient::new(
+        "https://ronnieg-keyvault.vault.azure.net/",
+        credential,
```
heaths, line 330:
```suggestion
        credential.clone(),
```

heaths (approved):
Best practice is to clone credentials, which is why we return a `Result<Arc<T>>`.

Unnecessary here, but we should show best practices regardless.

PR #2051, opened 2025-02-02
Add more SecretClient tests

heaths:
Also removes the `$..id` body sanitizer that breaks Key Vault resource ID assertions and name/version harvesting.


heaths:
Must rebase onto `main` after #2033 is merged.

RickWinter (approved):

PR #2062, opened 2025-02-04
Regenerate Key Vault Secrets with new emitter

heaths:
Changes the owned String parameters to borrowed &str slices.

Fixed integration tests accordingly.


RickWinter:
You need to run 
`cargo +stable fmt --all --`

Also you should pick up the latest main as the emitter eng has already been updated.

heaths:
> You need to run `cargo +stable fmt --all --`

I forgot. As we talked internally, I opened Azure/typespec-rust#251.

PR #2069, opened 2025-02-04
Add delete + purge test to Key Vault Secrets

heaths:


sdk/keyvault/azure_security_keyvault_secrets/tests/common/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+
+use std::{ops::Mul, pin::Pin, time::Duration};
+use tokio::time::{sleep, Sleep};
+
+const DEFAULT_TIMEOUT: Duration = Duration::from_secs(60 * 3);
+const MAX_DURATION: Duration = Duration::from_secs(30);
+
+/// Use in a retry loop.
+#[derive(Debug)]
+pub struct Retry(RetryImpl);
```
heaths, line 12: @JeffreyRichter @RickWinter if we're not going to create pseudo-LROs for operations like this when we did for most other languages, what about providing a utility like this in `azure_core`? I'd also want to implement `Stream` - maybe only `Stream`, but this was faster for now - but I think it could be immensely helpful. We should be able to do this generically as well without an explicit dependency on `tokio`, which doesn't matter here because we already uses it for tests (`dev-dependencies`).

@analogrelay do you know of any better way in `tokio`? I took a time-boxed look and found nothing that would let us retry within a timeout. One could use their `timeout()` function and do things in a loop as well, I suppose.

heaths: Actually, a better API might be like `tokio::time::timeout(Duration, Intofuture)` where we take an `IntoFuture` and run that in a loop with a progressive retry (unless immediate for playback) or until we timeout. We handle all the plumbing and the timeout would be immediate instead of the next call to `next()`. Should be cleaner.

RickWinter: Please create an issue to track. We can consider doing this after beta.1

RickWinter (approved):

PR #2085, opened 2025-02-07
Prepare eventhubs for beta

LarryOsterman:
# Prepare EventHubs service for beta 1

Lots of tweaks to EventHubs in this PR.

- Added samples for sending and receiving messages.
- Created aliases for client types into the azure_messaging_eventhubs crate.
  - Removed references to `producer::` and `consumer::`.
- Added aliases for AMQP types in the azure_messaging_eventhubs::models crate.
- Refactored event receive to move streaming of events into a separate EventReceiver struct.
- When receiving messages, defer conversion from AMQP message to eventhubs constructs until they are actually needed.
- Moved model types into their own module and out of lib.rs.
- Moved event_data model type into its own file.
- Sorted dependencies in cargo.toml
- Removed all cspell overrides and moved them into package specific custom dictionaries.


.vscode/cspell.json
```diff
@@ -163,4 +173,4 @@
       ]
     }
   ]
-}
+}
```
antkmsft, line 176:
```suggestion
}

```

LarryOsterman: THe formatter doesn't add that empty line.

sdk/core/azure_core_amqp/src/receiver.rs
```diff
@@ -100,6 +100,13 @@ impl AmqpReceiverApis for AmqpReceiver {
         self.implementation.credit_mode()
     }
 
+    /// Receives a delivery from the AMQP receiver.
+    ///
+    /// This method returns a single `AmqpDelivery` that can be used to receive a message from the AMQP receiver.
```
heaths, line 105:
```suggestion
    /// This method returns a single [`AmqpDelivery`] that can be used to receive a message from the AMQP receiver.
```
Creates a doc link.

sdk/core/azure_core_amqp/src/receiver.rs
```diff
@@ -100,6 +100,13 @@ impl AmqpReceiverApis for AmqpReceiver {
...
+    ///
+    /// If the receiver options are set to `auto_accept`, the delivery will have already been accepted, and no
```
heaths, line 107: For functions, the syntax is qualified (parens optional, but I prefer them so it's obvious it's a fn member):
```suggestion
    /// If the receiver options are set to [`AmqpReceiver::auto_accept()`], the delivery will have already been accepted, and no
```

sdk/eventhubs/azure_messaging_eventhubs/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+## 1.0.0-beta.1 (2025-02-13)
```
heaths, line 1: Until 1.0, we are using this syntax. Use 0.1.0 or whatever would be first for your crate name. That's idiomatic to rust.

sdk/eventhubs/azure_messaging_eventhubs/README.md
```diff
@@ -157,80 +157,127 @@ More information about Event Hubs features and terminology can be found here: [l
 Examples for various scenarios can be found on in the samples directory in our GitHub repo for
 [Event Hubs](https://github.com/Azure/azure-sdk-for-rust/tree/main/sdk/eventhubs/azure-messaging-eventhubs/samples).
 
-## Send events
-
-The following example shows how to send events to an event hub:
+## Open an eventhubs message producer on an eventhubs instance.
```
heaths, line 160: Isn't the product name "Event Hubs"? We should make sure our public docs follow branding guidelines e.g., ".NET" not ".Net", "Key Vault", not "KeyVault", etc.

sdk/eventhubs/azure_messaging_eventhubs/README.md
```diff
@@ -157,80 +157,127 @@ More information about Event Hubs features and terminology can be found here: [l
...
 
 ```rust
```
heaths, line 162:
````suggestion
```rust no_run
```
````

File: sdk/eventhubs/azure_messaging_eventhubs/README.md [hunk 7]
```diff
@@ -157,80 +157,127 @@ More information about Event Hubs features and terminology can be found here: [l
...
-async fn send_events() {
-    let host = std::env::var("EVENTHUBS_HOST").unwrap();
-    let eventhub = std::env::var("EVENTHUB_NAME").unwrap();
+use azure_messaging_eventhubs::{ProducerClient, ProducerClientOptions};
 
-    let credential = azure_identity::DefaultAzureCredential::new().unwrap();
+async fn open_producer_client() -> Result<ProducerClient, Box<dyn std::error::Error>>
```
heaths, line 165: Would we always return an `azure_core::Error`? If so, just be specific:
```suggestion
async fn open_producer_client() -> azure_core::Result<ProducerClient>
```
That way the compiler can use static dispatch.

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_batch_produce_events.rs
```diff
@@ -0,0 +1,61 @@
...
+use uuid::Uuid;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Set up the Event Hub client
+    let eventhub_namespace = std::env::var("EVENTHUBS_HOST")?;
+    let eventhub_name = std::env::var("EVENTHUB_NAME")?;
+    let credential = DefaultAzureCredential::new()?;
+
+    let client = ProducerClient::new(eventhub_namespace, eventhub_name, credential, None);
```
heaths, line 14: Even if it's the only one, we want to show best practices by cloning the arc:
```suggestion
    let client = ProducerClient::new(eventhub_namespace, eventhub_name, credential.clone(), None);
```

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_consume_events.rs
```diff
@@ -0,0 +1,59 @@
...
+    // Create a stream of events from the receiver
+    let receive_stream = receiver.stream_events();
+
+    println!("Created receive stream");
+
+    // Pin the receive stream on the stack so that it can be polled
+    pin_mut!(receive_stream);
+
+    // Receive events until the receive_timeout has been reached.
+    while let Some(event) = receive_stream.next().await {
```
heaths, line 52: Not `...await?`?

sdk/eventhubs/azure_messaging_eventhubs/src/consumer/event_receiver.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// Receives messages from the Event Hub.
+    /// This method returns a stream of `ReceivedEventData` that can be used to receive messages from the Event Hub.
+    /// The stream will continue to yield messages as long as the receiver is not closed.
+    /// The stream will yield an error if there is an issue receiving messages from the Event Hub.
+    ///
+    /// # Returns
+    ///
+    /// A stream of `ReceivedEventData` that can be used to receive messages from the Event Hub.
+    ///
+    /// # Example
```
heaths, line 76: It should be "Examples" only if one. `std` - and therefore most others - do this.

sdk/eventhubs/azure_messaging_eventhubs/src/consumer/mod.rs
```diff
@@ -24,21 +21,25 @@ use azure_core_amqp::{
...
     default::Default,
     fmt::Debug,
     sync::{Arc, OnceLock},
+    time::Duration,
 };
 use tracing::{debug, trace};
 use url::Url;
 
+/// Receive messages from a partition.
+pub mod event_receiver;
```
heaths, line 41: We can't just fold this into the `consumer` module? We don't want super-long module paths.

sdk/eventhubs/azure_messaging_eventhubs/src/consumer/mod.rs
```diff
@@ -189,10 +190,9 @@ impl ConsumerClient {
         Ok(())
     }
 
-    /// Receives events from a specific partition of the Event Hub.
+    /// Attaches a message receiver to a specific partition of the Event Hub.
     ///
-    /// This function establishes a connection to the specified partition of the Event Hub and starts receiving events from it.
-    /// It returns a stream of `ReceivedEventData` items, representing the events received from the partition.
+    /// This function establishes a connection to the specified partition of the Event Hub and returns a MessageReceiver which can be used to receive messages from it.
```
heaths, line 195: Maybe in a future PR, but link types as much as possible like I showed in my first couple of comments. Makes for better user docs.

sdk/eventhubs/azure_messaging_eventhubs/src/lib.rs
```diff
@@ -19,779 +16,18 @@ pub mod error;
...
+pub use producer::ProducerClientOptions;
+pub use producer::SendEventOptions;
+pub use producer::SendMessageOptions;
+pub use producer::SubmitBatchOptions;
+
+pub use consumer::ConsumerClient;
+pub use consumer::ConsumerClientOptions;
+pub use consumer::OpenReceiverOptions;
+pub use consumer::StartLocation;
+pub use consumer::StartPosition;
```
heaths, line 33: So you're exporting `consumer` and `producer` modules, and then these top-level clients similar to our "normal" Azure SDK guidelines currently? What are the `StartLocation` and `StartPosition` types? Sound like models.

sdk/eventhubs/azure_messaging_eventhubs/src/models/event_data.rs
```diff
@@ -0,0 +1,522 @@
...
+        if let Some(application_properties) = message.application_properties() {
+            for (key, value) in application_properties.0.clone() {
+                event_data_builder = event_data_builder.add_property(key, value);
+            }
+        }
+        event_data_builder.build()
+    }
+}
+
+impl Debug for EventData {
```
heaths, line 103: Why not just derive `Debug`? Seems you've done nothing it wouldn't have.

heaths (approved):
There are a couple of issues I'd like to see fixed like branding, but most are just nits/tips and some suggestions for a future PR (or now, if you want) for improved documentation. Please at least fix the branding, but approving otherwise to keep you unblocked.

sdk/eventhubs/azure_messaging_eventhubs/README.md
LarryOsterman: Strangely, cargo test --doc seems to be just fine with these. No idea why though.

sdk/eventhubs/azure_messaging_eventhubs/src/lib.rs
LarryOsterman: `StartLocation` and `StartPosition` are types referenced by the `ProducerClientOptions` and are not model types (we use them to format strings which are sent to the service). 

As I recall it, services are expected to expose their service client objects in their top-level client, even if they're located in a `clients` (or `producer` and `consumer`) module. I believe that also applies to all the `Options` types needed to create the client. 

I don't have any preferences here.

sdk/eventhubs/azure_messaging_eventhubs/src/models/event_data.rs
LarryOsterman: At one point there was a significant concern that because we don't have any idea what is in the body() of a message, it could be considered PII and we want to keep PII out of traces.

However, for now it's likely ok.

sdk/eventhubs/azure_messaging_eventhubs/src/models/event_data.rs
```diff
@@ -0,0 +1,522 @@
...
+                    if let AmqpAnnotationKey::Symbol(symbol) = key {
+                        if symbol == ENQUEUED_TIME_UTC {
+                            if let AmqpValue::TimeStamp(timestamp) = value {
+                                return timestamp.0;
+                            }
+                        }
+                    }
+                }
+            }
+            None
```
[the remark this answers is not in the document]
LarryOsterman: Or of course `let Some(annotations) = self.message.message_annotations()?` :).

sdk/eventhubs/azure_messaging_eventhubs/src/models/mod.rs
```diff
@@ -0,0 +1,299 @@
...
+/// let eventhub_properties = consumer_client.get_eventhub_properties().await?;
+///
+/// for partition_id in eventhub_properties.partition_ids {
+///    println!("Partition ID: {}", partition_id);
+/// }
+/// # Ok(()) }
+/// ```
+///
+#[derive(Debug)]
+pub struct EventHubProperties {
```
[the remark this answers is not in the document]
LarryOsterman: In this case, the `EventHubProperties` and `EventHubPartitionProperties` values are contractual - it's an error if the service doesn't return these fields, and we'll fail the API if they are not provided.

The one "exception" is the time based values, and that's because the service returns a sentinel value of 1-1-0001 if a time is not available, which we turn into `None` in the AMQP layer.

sdk/eventhubs/azure_messaging_eventhubs/src/models/event_data.rs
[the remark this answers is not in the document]
heaths: Yeah - only if the fn signature matches. That's where the various map functions come in.

sdk/eventhubs/azure_messaging_eventhubs/README.md
LarryOsterman: I had the generic form because std::env returned a `VarError`. But I don't need the std::env reference.

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_consume_events.rs
LarryOsterman: `stream.next()` returns an `Option<Result<T>>` I'm not sure I know how to combine the two.

sdk/eventhubs/azure_messaging_eventhubs/src/consumer/event_receiver.rs
LarryOsterman: To confirm: If there is a single example, we should use the word "Examples".

If there are multiple examples, we should use the word "Example".

That's rather counter intuitive, but I can work with it.

sdk/eventhubs/azure_messaging_eventhubs/src/consumer/mod.rs
LarryOsterman: Ah - the objection is to the `pub`, not to the `mod`.

PR #2100, opened 2025-02-11
API Review feedback; fixed a bunch of type documentation; reduced type visibility in public API surface.

### LINKED ISSUE #2129: eventhubs: minor errata with the Producer
- [x] ProducerClient allows you to set the maximum message size at the client level. In Go, the only place a message size can be set is when creating a batch. I believe that's the case, across the board.
- [x] `submit_batch` should be renamed, as we generally use 'send' as the verb. So perhaps either `send_batch()`, or `send_event_batch()`?
- [x] batch.len() and batch.size() were confusing to me when looking through EventBatch's methods. What do you think about renaming size() to size_in_bytes(), or similar?

LarryOsterman:
# Cleanup of AMQP and EventHubs for beta
- Removed `ProducerClient::open()` and `ConsumerClient::open()`, replacing them with builders that terminate in an "open()" verb.
- Renamed `ProducerClient::submit_batch()` to `ProducerClient::send_batch()`
- Renamed `EventDataBatch::size()` to `EventDataBatch::size_in_bytes()`
- Changed input `String` parameters to `&str` parameters to simplify certain authoring scenarios.
- Reduced visibility of AMQP modules and expose types from the root.
- Improved EventHubs documentation including models types.

Fixes #2129

### DIFF
--- a/sdk/eventhubs/azure_messaging_eventhubs/README.md
+++ b/sdk/eventhubs/azure_messaging_eventhubs/README.md
@@ -215,6 +215,7 @@ use azure_messaging_eventhubs::ProducerClient;
 
 async fn send_events(producer: &ProducerClient) -> Result<(), Box<dyn std::error::Error>> {
     producer.send_event(vec![1, 2, 3, 4], None).await?;
+    Ok(())
 }
 ```
 
@@ -224,12 +225,13 @@ async fn send_events(producer: &ProducerClient) -> Result<(), Box<dyn std::error
 use azure_messaging_eventhubs::ProducerClient;
 
 async fn send_events(producer: &ProducerClient) -> Result<(), Box<dyn std::error::Error>> {
-    let mut batch = producer.create_batch(None).await;
+    let mut batch = producer.create_batch(None).await?;
     assert_eq!(batch.len(), 0);
     assert!(batch.try_add_event_data(vec![1, 2, 3, 4], None)?);
 
     let res = producer.send_batch(&batch, None).await;
     assert!(res.is_ok());
+    Ok(())
 }
 ```
 
@@ -277,6 +279,7 @@ async fn receive_events(client: &ConsumerClient) -> Result<(), Box<dyn std::erro
             }
         }
     }
+    Ok(())
 }
 ```
 
--- a/sdk/eventhubs/azure_messaging_eventhubs/README.md
+++ b/sdk/eventhubs/azure_messaging_eventhubs/README.md
@@ -162,7 +162,7 @@ Examples for various scenarios can be found on in the samples directory in our G
 ## Open an Event Hubs message producer on an Event Hubs instance.
 
 ```rust no_run
-use azure_messaging_eventhubs::{ProducerClient, ProducerClientOptions};
+use azure_messaging_eventhubs::ProducerClient;
 
 async fn open_producer_client() -> Result<ProducerClient,azure_core::Error>
 {
@@ -171,13 +171,11 @@ async fn open_producer_client() -> Result<ProducerClient,azure_core::Error>
 
     let credential = azure_identity::DefaultAzureCredential::new()?;
 
-    let producer = azure_messaging_eventhubs::ProducerClient::new(
+    let producer = azure_messaging_eventhubs::ProducerClient::builder(
         host.to_string(),
         eventhub.to_string(),
-        credential.clone(),
-        None,
-        );
-    producer.open().await?;
+        credential.clone())
+        .open().await?;
 
     Ok(producer)
     }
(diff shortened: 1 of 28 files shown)


sdk/eventhubs/azure_messaging_eventhubs/README.md
```diff
@@ -162,47 +162,42 @@ Examples for various scenarios can be found on in the samples directory in our G
...
 
-    let producer = azure_messaging_eventhubs::ProducerClient::new(
-        host.to_string(),
-        eventhub.to_string(),
-        credential.clone(),
-        None,
-        );
-    producer.open().await?;
+    let producer =
+        azure_messaging_eventhubs::ProducerClient::builder(host, eventhub, credential.clone())
```
heaths, line 174: Why are you passing parameters to a builder? Doesn't seem very builder-y.

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_properties.rs
```diff
@@ -18,22 +17,17 @@ async fn main() -> Result<()> {
...
-    if result.is_err() {
-        println!("Error opening client: {:?}", result.err());
+    let result = ProducerClient::builder(host.as_str(), eventhub.as_str(), credential.clone())
+        .with_application_id("test_get_properties")
+        .open()
+        .await;
+    if let Err(err) = result {
+        println!("Error opening client: {:?}", err);
         return Ok(());
     }
```
heaths, line 27: That seems pretty fatal. You're already set up to use `?` so why not just do that and let the example program fail with a more useful message? Also, all `std::error::Error`s impl `Display` (it's a constraint) so you don't need to use the `Debug` notation to format. Generally will be more user-friendly, as is `Display` meant to be over `Debug`.

sdk/eventhubs/azure_messaging_eventhubs/src/consumer/mod.rs
```diff
@@ -77,23 +81,43 @@ impl ConsumerClient {
...
+    ///    .open().await?;
     /// # Ok(())}
     /// ```
-    pub fn new(
+    ///
+    pub fn builder(
+        fully_qualified_namespace: &str,
+        eventhub_name: &str,
+        consumer_group: Option<String>,
+        credential: Arc<dyn TokenCredential>,
```
heaths, line 93: Again, not very builder-y. Or are these all required parameters? The `Option<String>` for `consumer_group` says otherwise.

heaths (approved):
I see nothing egregious, but the builders don't really seem to be used like builders. That said, I wouldn't change anything now before beta.1.

Did you do `cargo doc --open` to checkout that the API shape looks right to you? All the refactoring of types - rather, reshuffling - were a little hard to make out in this PR.

sdk/eventhubs/azure_messaging_eventhubs/README.md
LarryOsterman: QQ: How do you express required parameters in a builder? I figured that making them be parameters on the creation of the builder made the most sense.

I guess I could put them as parameters on the terminal `open()` call instead.

But fundamentally, they need to be *somewhere*.

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_properties.rs
LarryOsterman: In my experience, ? doesn't give a great error - all it does is print the error text - I'd lose the context for the error.

sdk/eventhubs/azure_messaging_eventhubs/src/consumer/mod.rs
LarryOsterman: consumer_group is optional, the others are all mandatory.

sdk/eventhubs/azure_messaging_eventhubs/README.md
LarryOsterman: Moved the parameters to the `open` method. The more that I think about it, it makes sense for them to be there, since most of them have to do with the network connection operation anyway.

heaths: There's also the typestate pattern: https://github.com/heaths/typestate-rs. That said, [guidelines](https://azure.github.io/azure-sdk/rust_introduction.html#rust-builders) don't currently prescribe this scenario. We could say that required parameters for builders are passed as "constructor" args for the builder. I'm not opposed to it, but it was confusing since you have `Option<T>` "required" args.

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_properties.rs
heaths: You won't get a stacktrace this way either.

PR #2108, opened 2025-02-12
[Draft] [Storage] Blob `download_blob` and `upload_blob` APIs, `BlobContainerClient` and `BlobServiceClient` clients

vincenttran-msft:
This PR currently generates against [vincenttran/blob-tsp-test2](https://github.com/Azure/azure-rest-api-specs/tree/vincenttran/blob-tsp-test2) which should roughly represent `v0.9.1`.

This pull request contains the following:
**BlobClient**
- `download_blob`
- `upload_blob`
- `commit_block_list`
- `get_block_list`
- `stage_block`

**BlobContainerClient**
- Initial introduction of `BlobContainerClient` client
- `create_container`
- `delete_container`
- `get_container_properties`
     - Includes handwritten model `ContainerProperties` and `FromHeaders` impl

**BlobServiceClient**
- Initial introduction of `BlobServiceClient` client
- `get_service_properties`


sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -15,19 +20,41 @@ async fn test_get_blob_properties() -> Result<(), Box<dyn Error>> {
     let credential = DefaultAzureCredentialBuilder::default().build()?;
 
     // Act
+    let container_client = BlobContainerClient::new(
+        &endpoint,
+        String::from("testcontainer1"),
```
vincenttran-msft, line 25: We need to figure out how to seed the container names similar to other languages (seeded random based on test name) since we keep hitting conflicting test container names.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -15,19 +20,41 @@ async fn test_get_blob_properties() -> Result<(), Box<dyn Error>> {
...
         String::from("test_blob.txt"),
         credential,
         Some(BlobClientOptions::default()),
     )?;
+    let data = b"hello rusty world";
+    blob_client
+        .upload_blob(
+            RequestContent::from(data.to_vec()),
+            true,
+            data.len() as i64,
```
vincenttran-msft, line 43: This is possibly unsafe code as we are going from `usize` to `i64` which could cause overflow. We should consider adding safe conversion logic to the handwritten API, but we need to figure out the most cheapest and safest way to calculate len.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -15,19 +20,42 @@ async fn test_get_blob_properties() -> Result<(), Box<dyn Error>> {
...
         .await;
 
     // Assert
     assert!(response.is_ok());
+
+    let blob_properties = response?;
+    assert_eq!(blob_properties.blob_type, Some(BlobType::BlockBlob));
+    assert_eq!(blob_properties.content_length, Some(17));
+
+    container_client.delete_container(None).await?;
```
vincenttran-msft, line 58: nit: Move all of these to a try-finally so cleanup attempt happens even if the test fails unexpectedly.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -53,17 +61,60 @@ impl BlobClient {
         &self.endpoint
     }
 
+    pub fn container_name(&self) -> String {
```
vincenttran-msft, line 64: Return ref, clone it when you use it

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -53,17 +61,60 @@ impl BlobClient {
...
+    }
+
     pub async fn get_blob_properties(
         &self,
         options: Option<BlobBlobClientGetPropertiesOptions<'_>>,
     ) -> Result<BlobProperties> {
         let response = self
             .client
-            .get_blob_blob_client(self.container_name.clone(), self.blob_name.clone())
+            .get_blob_blob_client(self.container_name(), self.blob_name())
```
vincenttran-msft, line 78: Don't need ()

vincenttran-msft: But still need to clone because subclients take ownership.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: For now I will leave the handwritten API matching the signature of the underlying generated code-- refactored tests to use `try_from` which will bubble up any conversion errors instead of silently swallowing like `as`.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: There is no try-catch or try-finally since Rust is a Results-based driven codeflow. I found some workarounds but they seem way too convoluted. Going to punt this for a little longer.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: For now we will assign these tests all unique container names.

PR #2131, opened 2025-02-12
Add ci.yml for existing services

hallipr:


sdk/eventhubs/tests.yml
```diff
@@ -0,0 +1,5 @@
+trigger: none
+
+steps:
+  - pwsh: |
+      Write-Host "No-Op"
```
weshaggard, line 5: Is there any reason to add these placeholders? Why not add them when we want to add the tests.

weshaggard (approved):

sdk/eventhubs/tests.yml
hallipr: removed

sdk/core/tests.yml
```diff
@@ -0,0 +1,5 @@
+trigger: none
+
+steps:
+  - pwsh: |
+      Write-Host "No-Op"
```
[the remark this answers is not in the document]
hallipr: it may have.  I was assuming that a build with no steps would fail to start

heaths (approved):
I guess that works...if it works...too.

PR #2136, opened 2025-02-12
dev-dependencies of workspace packages should still use path

hallipr:
1. `cargo package` requires that all dependencies have versions declared
2. `workspace = true` on dependencies and dev-dependencies use `workspace.dependencies` for resolution
3. `path` based packages in `workspace.dependencies` get version numbers added during packaging to satisfy `1.`
4. `cargo package` expects dependencies with versions to be available in the feed or the local vendor directory
5. packages with `publish = false` aren't packaged or placed in the vendor directory
6. workspace based dev-dependencies on unpublished crates fails because of `3.` + `4.` + `5.`
    - consider  A depends on B, B dev depends on A
    - when A and B both publish, B must package before A.  B will not have `version` A available to satisfy the dev-dependency, but it will have `path` A available. Because of `.3. + `4.`, `cargo package` on B will fail because it can't find A@version even though `path` is also specified
  
We can:
- stop using workspace dependencies for dev-dependencies
    - will ensure that no version is added to a path dev-dependency through workspace.dependencies
- start skipping `publish = false` packages when placing version numbers beside path dependencies in toml
    - will prevent cargo from expecting "local only" packages to come from the feed
    - will mean that some packages in dev-dependencies can use workspace = true (ones that publish) while others can't (ones that don't)

### DIFF
--- a/sdk/keyvault/azure_security_keyvault_keys/Cargo.toml
+++ b/sdk/keyvault/azure_security_keyvault_keys/Cargo.toml
@@ -22,7 +22,9 @@ time = { workspace = true }
 typespec_client_core = { workspace = true, features = ["derive"] }
 
 [dev-dependencies]
-azure_core_test = { workspace = true, features = ["tracing"] }
+azure_core_test = { path = "../../core/azure_core_test", features = [
+    "tracing",
+] }
 azure_identity.path = "../../identity/azure_identity"
 azure_security_keyvault_test = { path = "../azure_security_keyvault_test" }
 rand.workspace = true
--- a/sdk/keyvault/azure_security_keyvault_secrets/Cargo.toml
+++ b/sdk/keyvault/azure_security_keyvault_secrets/Cargo.toml
@@ -25,7 +25,9 @@ time = { workspace = true }
 typespec_client_core = { workspace = true, features = ["derive"] }
 
 [dev-dependencies]
-azure_core_test = { workspace = true, features = ["tracing"] }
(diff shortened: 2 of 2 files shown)


heaths:
Have you tried both `path` and `version`? This is supported where `path` (or `git`) works locally for development, but `version` is used when publishing (or pulling).

hallipr:
> Have you tried both `path` and `version`? This is supported where `path` (or `git`) works locally for development, but `version` is used when publishing (or pulling).

When `path` and `version` exist, it uses `version` during `cargo package`

heaths:
Right, which is what we wanted, I thought.

hallipr:
If version and path are both specified, then the path needs to meet the version req.

I think we should reserve workspace dependencies to local packages for centrally indicating the released version that consumers in the repo should depend on.

When we want everyone to depend on the unreleased version, we need to remove the version number and put just a path.  When we do this, we have to be aware that some packages will take dependencies and some will take dev-dependencies. This may make circular references that can't be resolved if the dev-dependency contains a version.

heaths (approved):
Cosmos and Event Hubs are using this crate as well. Maybe we should even remove it from `[dependencies]` in the workspace `Cargo.toml`. @LarryOsterman @analogrelay what do you think? If we want certain dev-dependencies to use only path (I don't want to publish `azure_core_test` et. al.), it makes sense to me.

If nothing else - if we need to move fast on this - lets remove `azure_core_test` and `azure_core_test_macros` from the workspace `Cargo.toml` and see how it goes.

PR #2140, opened 2025-02-12
Use native cargo commands for crates.io work

hallipr:


eng/scripts/Language-Settings.ps1
```diff
@@ -119,28 +121,21 @@ function Get-AllPackageInfoFromRepo ([string] $ServiceDirectory) {
...
 function Get-rust-PackageInfoFromPackageFile([IO.FileInfo]$pkg, [string]$workingDirectory) {
-  #$pkg will be a FileInfo object for the cargo-metadata.json file in a package artifact directory
-  $package = Get-Content -Path $pkg.FullName -Raw | ConvertFrom-Json
-  $packageName = $package.name
-  $packageVersion = $package.vers
+  InstallAndImport-ModuleIfNotInstalled "PSToml" "0.3.1"
 
-  $crateFile = Get-ChildItem $pkg.DirectoryName -Filter '*.crate'
+  #$pkg will be a FileInfo object for the Cargo.toml file in a package artifact directory
+  $toml = Get-Content -Path $pkg.FullName -Raw | ConvertFrom-Toml
```
heaths, line 127: Why not just use `cargo read-manifest` which JSONifies the TOML?

eng/scripts/Publish-Crates.ps1
```diff
@@ -11,34 +11,22 @@ param(
...
+    exit 1
+  }
 
-  TryAddOwners $crateName
+  $existingOwners = (Invoke-LoggedCommand cargo owner --list $crateName) -replace " \(.*", ""
+  $missingOwners = $AdditionalOwners | Where-Object { $existingOwners -notcontains $_ }
+
+  foreach ($owner in $missingOwners) {
+    Write-Host "> cargo owner --add $owner $crateName --token <TOKEN>"
+    cargo owner --add $owner $crateName --token $Token
```
heaths, line 30: Shouldn't this use the `Invoke-WhateverCommand`, or is that you can't mask values? That said, AzDO can if you pass the value to its magic syntax. Any instance of that value it'll mask. Probably not a big deal, though.

heaths (approved):

eng/scripts/Language-Settings.ps1
hallipr: that is much better

heaths (approved):

eng/scripts/Publish-Crates.ps1
hallipr: AzDO would mask the value if we wrote it to Host.   I'm avoiding the use of `eng/common` helpers here because we may be required to inline this powershell for 1ES safe deployments:

https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/1es-pipeline-templates/features/releasepipelines/overview

> - You have to declare all the artifacts required for the release as inputs for the job (this is similar to the concept of outputs in a standard build job).
> - You can't build source code in a release job (this is to ensure all artifacts have been binary scanned in a build job output).
> - You can't check out repositories. All artifacts must be generated and published from a 1ES PT build job and declared as an input.
> - All 1ES PT pipelines must use a [1ES hosted pool](https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/1es-pipeline-templates/onboardingesteams/overview).

PR #2141, opened 2025-02-12
Update Key Vault keys README

heaths:
Also makes a few improvements to the secrets README and adds them to the crates' `src/lib.rs` as guidelines dictate.


sdk/keyvault/azure_security_keyvault_keys/README.md
```diff
@@ -1,66 +1,370 @@
...
+
+```sh
+cargo add azure_identity azure_core tokio
+```
+
+### Authenticate the client
+
+In order to interact with the Azure Key Vault service, you'll need to create an instance of the `KeyClient`. You need a **vault url**, which you may see as "DNS Name" in the portal, and credentials to instantiate a client object.
+
+The example shown below use a `DefaultAzureCredential`, which is appropriate for most scenarios including local development and production environments. Additionally, we recommend using a managed identity for authentication in production environments. You can find more information on different ways of authenticating and their corresponding credential types in the [Azure Identity] documentation.
```
RickWinter, line 45:
```suggestion
The example shown below use a `DefaultAzureCredential`, which is appropriate for local development environments. Additionally, we recommend using a managed identity for authentication in production environments. You can find more information on different ways of authenticating and their corresponding credential types in the [Azure Identity] documentation.
```

RickWinter (approved):

sdk/keyvault/azure_security_keyvault_keys/README.md
```diff
@@ -1,66 +1,370 @@
...
+use azure_identity::DefaultAzureCredential;
+use azure_security_keyvault_keys::{models::KeyBundle, KeyClient};
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let credential = DefaultAzureCredential::new()?;
     let client = KeyClient::new(
-        "https://<your-key-vault-name>.vault.azure.net",
+        "https://your-key-vault-name.vault.azure.net/",
         credential.clone(),
```
LarryOsterman, line 167: QQ: You've mentioned that best practices is to require the caller to `clone()` the credentials.

Does that imply that the service client should *not* clone the credentials since we're expecting the caller to `clone()` before passing the credentials into the service client? Otherwise, how do we avoid double cloning the credentials?

Or is this to safeguard against the case where the caller is going to be passing the same credentials client into multiple service clients? In which case, wouldn't the borrow checker guide them to cloning?

Is it better to say that double cloning on an arc is a best practice, or to let the borrow checker fail when someone accidentally consumes the arc?

sdk/keyvault/azure_security_keyvault_keys/README.md
```diff
@@ -1,66 +1,370 @@
...
+
+    println!("Key: {:#?}", key.key);
+
+    Ok(())
+}
+```
+
+### Update an existing key
+
+`update_key` updates a key previously stored in the Azure Key Vault. Only the attributes of the key are updated. To update the value, call `KeyClient::create_key` on a key with the same name.
```
LarryOsterman, line 186: Should this be [`KeyClient::update_key()`] to force a link?

LarryOsterman (approved):

sdk/keyvault/azure_security_keyvault_keys/README.md
```diff
@@ -1,66 +1,370 @@
...
+```azurecli
+az keyvault create --resource-group <your-resource-group-name> --name <your-key-vault-name>
+```
+
+### Install dependencies
+
+Add the following crates to your project:
+
+```sh
+cargo add azure_identity azure_core tokio
```
ronniegeraghty, line 38:
```suggestion
cargo add azure_identity tokio
```
I believe we don't need to call out core specifically unless we want to work with common types.

ronniegeraghty (approved):

sdk/keyvault/azure_security_keyvault_keys/README.md
heaths: Not in a README. That only works in rustdocs. While included in lib.rs and that *might* happen, it wouldn't render as a proper link here in the README and would fail the link checker.

sdk/keyvault/azure_security_keyvault_keys/README.md
heaths: I'd rather the customer do it so we're not needless. Let the borrow checking complain if necessary.

LarryOsterman: To confirm: we should NOT clone the credentials when saving them in the service client object and instead we should consume the credential object? Not a huge deal, but I would like to be consistent across the SDKs.

PR #2147, opened 2025-02-13
Event Hubs: make it possible to set a custom endpoint for use with TCP proxies

richardpark-msft:
AMQP clients can choose to connect _through_ another host, like a TCP proxy. When you do that you need to set the Host attribute in the AMQP connection.

This draft just shows how to do that, and plumbs it through the Producer.

This is how I configure the client to use it:

```rust
    let client = ProducerClient::new(
        host,
        eventhub.clone(),
        credential,
        Some(ProducerClientOptions {
            application_id: Some("test_get_properties".to_string()),
            custom_endpoint: Some("amqp://localhost:5671".to_string()),
            ..Default::default()
        }),
    );
```

NOTE, since it's the amqpproxy, and I don't have a proper signed cert (yet) I used plain `amqp`. It's only unencrypted between the code and the proxy - all outbound traffic from the proxy to the actual Event Hubs service is still TLS.

### DIFF
--- a/sdk/core/azure_core_amqp/src/connection.rs
+++ b/sdk/core/azure_core_amqp/src/connection.rs
@@ -24,6 +24,7 @@ pub struct AmqpConnectionOptions {
     pub desired_capabilities: Option<Vec<AmqpSymbol>>,
     pub properties: Option<AmqpOrderedMap<AmqpSymbol, AmqpValue>>,
     pub buffer_size: Option<usize>,
+    pub custom_endpoint: Option<Url>,
 }
 
 impl AmqpConnectionOptions {}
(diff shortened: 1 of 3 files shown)


richardpark-msft:
@LarryOsterman, this is the code I'm using to connect to the amqp proxy, and it matches up with what we've done in other libraries to enable it as a legitimate customer feature (you can also use it when connecting to TCP proxies). I only did it for the Producer, as a demonstration.

Also, I feel like I must have clone()'d, as_str()'d and others more than I needed to. I just based this off what I could copy/paste in your library :)

heaths:
Could you shorten the title and put more of a description in here? I'm not sure what you're trying to do.

richardpark-msft:
> Could you shorten the title and put more of a description in here? I'm not sure what you're trying to do.

No problem :)

heaths:
@LarryOsterman does this one need to be closed now in favor of #2245?

PR #2150, opened 2025-02-13
Update Event Hubs and Key Vault secrets READMEs

### LINKED ISSUE #2121: Event Hubs README update
Update the Event Hubs `README.md` to follow standard format.

ronniegeraghty:
Fixes #2121

### DIFF
--- a/sdk/eventhubs/azure_messaging_eventhubs/README.md
+++ b/sdk/eventhubs/azure_messaging_eventhubs/README.md
@@ -62,7 +62,7 @@ az login
 
 Instantiate a `DefaultAzureCredential` to pass to the client. The same instance of a token credential can be used with multiple clients if they will be authenticating with the same identity.
 
-### Create an Event Hubs message producer on an Event Hub instance
+### Create an Event Hubs message producer and send an event
 
 ```rust no_run
 use azure_identity::DefaultAzureCredential;
@@ -70,20 +70,16 @@ use azure_messaging_eventhubs::ProducerClient;
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let host = "<EVENTHUBS_NAMESPACE_HOST>";
+    let host = "<EVENTHUBS_HOST>";
     let eventhub = "<EVENTHUB_NAME>";
 
     // Create new credential
     let credential = DefaultAzureCredential::new()?;
 
-        // Create and open new Producer Client
-    let producer = ProducerClient::new(
-        host.to_string(),
-        eventhub.to_string(),
-        credential.clone(),
-        None,
-    );
-    producer.open().await?;
+    // Create and open a new ProducerClient
+    let producer = ProducerClient::builder()
+        .open(host, eventhub, credential.clone())
+        .await?;
 
     producer.send_event(vec![1, 2, 3, 4], None).await?;
 
@@ -127,18 +123,19 @@ Additional examples for various scenarios can be found on in the examples direct
 ## Open an Event Hubs message producer on an Event Hub instance
 
 ```rust no_run
+use azure_core;
+use azure_identity::DefaultAzureCredential;
 use azure_messaging_eventhubs::ProducerClient;
 
 async fn open_producer_client() -> Result<ProducerClient, azure_core::Error> {
     let host = "<EVENTHUBS_HOST>";
     let eventhub = "<EVENTHUB_NAME>";
 
-    let credential = azure_identity::DefaultAzureCredential::new()?;
+    let credential = DefaultAzureCredential::new()?;
 
-    let producer =
-        azure_messaging_eventhubs::ProducerClient::builder()
-            .open(host, eventhub, credential.clone())
-            .await?;
+    let producer = ProducerClient::builder()
+        .open(host, eventhub, credential.clone())
+        .await?;
 
     Ok(producer)
 }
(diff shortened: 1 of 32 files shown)


sdk/eventhubs/azure_messaging_eventhubs/README.md
```diff
@@ -243,16 +208,18 @@ Each message receiver can only receive messages from a single Event Hubs partiti
...
-use azure_messaging_eventhubs::ConsumerClient;
+use azure_messaging_eventhubs::{
+    ConsumerClient, OpenReceiverOptions, StartLocation, StartPosition,
+};
 use futures::pin_mut;
 
 async fn receive_events(client: &ConsumerClient) -> Result<(), Box<dyn std::error::Error>> {
     let message_receiver = client
         .open_receiver_on_partition(
             "0",
```
LarryOsterman, line 219: Might want to add a comment explaining why the `start_position` is set.

Something like:

```rust
// By default, an event receiver only receives new events from the event hub. To receive events from earlier, specify
// a `start_position` which represents the position from which to start receiving events. 
// In this example, events are received from the start of the partition.

File: sdk/eventhubs/azure_messaging_eventhubs/README.md [hunk 2]
```diff
@@ -276,63 +243,53 @@ async fn receive_events(client: &ConsumerClient) -> Result<(), Box<dyn std::erro
...
+
     Ok(())
 }
 ```
 
 # Troubleshooting
 
+## General
+
+When you interact with the Azure Event Hubs client library using the Rust SDK, errors returned by the service correspond to the same HTTP status codes returned for [REST API] requests.
```
LarryOsterman, line 255: This isn't strictly accurate - we return `azure_core::Error` values using `ErrorKind::Other` which are `azure_messaging_eventhubs::Error` values.

The `get_eventhubs_properties()` and `get_eventhubs_partition_properties()` APIs return HTTP status codes but they are buried and hard to find (in the current design). This will be fixed in a future iteration of the eventhubs and amqp clients.

LarryOsterman (approved):

sdk/eventhubs/azure_messaging_eventhubs/README.md
ronniegeraghty: I wrote something new based on what you said. Please double check me cause I'm not too sure what I wrote is accurate. 
Also, we don't need to be super extensive here. We'll have time to improve on the troubleshooting sections as the libraries change and improve.

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -150,12 +146,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         ..Default::default()
     };
 
-    let secret = client
-        .set_secret(
-            "secret-name".into(),
-            secret_set_parameters.try_into()?,
-            None,
-        )
+    let secret: SecretBundle = client
```
heaths, line 149: Why do you keep adding this back? No one would type this unnecessarily.
```suggestion
    let secret = client
```

sdk/keyvault/azure_security_keyvault_secrets/README.md
```diff
@@ -189,8 +181,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     )?;
 
     // Retrieve a secret using the secret client.
-    let secret = client
-        .get_secret("secret-name".into(), "secret-version".into(), None)
+    let secret: SecretBundle = client
```
heaths, line 184:
```suggestion
    let secret = client
```

sdk/eventhubs/azure_messaging_eventhubs/README.md
```diff
@@ -156,82 +109,96 @@ More information about Event Hubs features and terminology can be found here: [l
...
+[Event Hubs](https://github.com/Azure/azure-sdk-for-rust/tree/main/sdk/eventhubs/azure_messaging_eventhubs/examples).
 
-## Open an Event Hubs message producer on an Event Hubs instance.
+<!-- no toc -->
+* [Open an Event Hubs message producer on an Event Hub instance](#open-an-event-hubs-message-producer-on-an-event-hub-instance)
+* [Send events](#send-events)
+  * [Send events directly to the Event Hub](#send-events-directly-to-the-event-hub)
+  * [Send events using a batch operation](#send-events-using-a-batch-operation)
+* [Open an Event Hubs message consumer on an Event Hubs instance](#open-an-event-hubs-message-consumer-on-an-event-hub-instance)
+* [Receive events](#receive-events)
```
heaths, line 121: Did you check these all still work? In fact, come to think of it, I don't think I updated them for Key Vault secrets or keys. Could you double check?

heaths (approved):

sdk/eventhubs/azure_messaging_eventhubs/README.md
ronniegeraghty: Just checked all the samples for Event Hubs, Secrets and Keys. 
They all compile fine. I ran `cargo fmt` on the secrets and keys samples.

heaths: No, I mean that the TOC actually points to bookmarks that exist. I'll check them quick...

ronniegeraghty: Oh. Yes, I checked, they all work and point to the examples listed below.

PR #2151, opened 2025-02-13
Don't nest the output package folders

hallipr:
The output folders were being produced like:
`a/packages/azure_template_core/azure_template_core-1.0.0-beta.4562713/Cargo.toml`

and not the expected:
`a/packages/azure_template_core/Cargo.toml`

### DIFF
--- a/eng/scripts/Publish-Crates.ps1
+++ b/eng/scripts/Publish-Crates.ps1
@@ -4,14 +4,11 @@
 param(
   [string]$PackagesPath,
   [string[]]$CrateNames,
-  [string[]]$AdditionalOwners,
-  [string]$Token
+  [string[]]$AdditionalOwners
 )
 
 $ErrorActionPreference = 'Stop'
 Set-StrictMode -Version 2.0
-
-$env:CARGO_REGISTRY_TOKEN = $Token
 
 foreach ($crateName in $CrateNames) {
   Write-Host "Publishing packae: '$crateName'"
--- a/eng/scripts/Publish-Crates.ps1
+++ b/eng/scripts/Publish-Crates.ps1
@@ -22,7 +22,7 @@ foreach ($crateName in $CrateNames) {
     exit 1
   }
 
-  $existingOwners = (Invoke-LoggedCommand cargo owner --list $crateName) -replace " \(.*", ""
+  $existingOwners = (cargo owner --list $crateName) -replace " \(.*", ""
   $missingOwners = $AdditionalOwners | Where-Object { $existingOwners -notcontains $_ }
 
   foreach ($owner in $missingOwners) {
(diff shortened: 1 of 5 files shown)


LarryOsterman (approved):
Changes look reasonable, but I'll be honest and say I'm not 100% sure how this fixes the issue.

hallipr:
> Changes look reasonable, but I'll be honest and say I'm not 100% sure how this fixes the issue.

When copying Folder to Folder in powershell  `Copy-Item  /source  /output -Recurse` 
if  `/output` exists, the contents of `/source` will be copied to `/output/source`.
If `/output` doesn't exist, then the contents of `/source` will be copied to `/output`

To ensure consistent output paths, I make sure the output folder exists, then I tell powershell to copy the _**contents**_ of the source folder, not the source folder itself:

`New-Item -Type Directory /output`
`Copy-Item /source/* /output -Recurse`

heaths:
Ugh. PowerShell's file IO "magic" differing from .NET's strikes again (`Join-Path` vs. `[System.IO.Path]::Join` gets people every time - always use the latter).

eng/scripts/Yank-Crates.ps1
```diff
@@ -10,15 +10,17 @@ param(
 $ErrorActionPreference = 'Stop'
 Set-StrictMode -Version 2.0
 
+$env:CARGO_REGISTRY_TOKEN = $Token
```
heaths, line 13: Should this even be passed in script? Why not just set it in `env` in the pipeline YAML? That's more typical anyway and has less risk of accidentally leaking it in output somehow.

heaths (approved):
Approving to unblock, but I think the token should just be set in the pipeline instead of passed to the script to set an env var. It's just more moving parts, and more risk of leakage. Just set it on the pipeline env once for the job and call it done.

eng/scripts/Publish-Crates.ps1
```diff
@@ -11,22 +11,24 @@ param(
...
+$env:CARGO_REGISTRY_TOKEN = $Token
+
 foreach ($crateName in $CrateNames) {
   Write-Host "Publishing packae: '$crateName'"
   $manifestPath = "$PackagesPath/$crateName/Cargo.toml"
   # https://doc.rust-lang.org/cargo/reference/registry-web-api.html#publish
-  Write-Host "> cargo publish --manifest-path `"$manifestPath`" --token <TOKEN>"
-  cargo publish --manifest-path $manifestPath --token $Token
+  Write-Host "> cargo publish --manifest-path `"$manifestPath`""
+  cargo publish --manifest-path $manifestPath
```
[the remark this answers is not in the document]
hallipr: I'll be inlining this script as we're not supposed to be checking out any code in the deployment job.  `Invoke-LoggedCommand` comes from eng/common and will be unavailable in the inlined script

eng/scripts/Yank-Crates.ps1
hallipr: using `env:` from yaml now

heaths (approved):
Much better!

PR #2155, opened 2025-02-13
Update CHANGELOGs for initial release

heaths:


RickWinter (approved):
NIT : "Initial release"

Use the two word statement instead of initial public release and initial supported release

PR #2156, opened 2025-02-13
Optimize rebuild when AZURE_TEST_MODE changes

heaths:
Doesn't need to be as far down as rebuilding the macros crate.


heaths:
Can wait till after the beta.1 release. I realized in the demo this afternoon that I didn't need to rebuild the macros crate since it doesn't change - just the code it runs on behalf of azure_core_test. A slight build speed improvement.

LarryOsterman (approved):

PR #2161, opened 2025-02-14
Implement From for std::str::ParseBoolError for Error

jhendrixMSFT:


sdk/typespec/src/error/mod.rs
```diff
@@ -247,6 +247,12 @@ impl From<std::io::Error> for Error {
     }
 }
 
+impl From<core::str::ParseBoolError> for Error {
```
heaths, line 250: Just use `std` instead of `core`. `core` is a subset of `std` and I can't imagine us even trying to be `no-std` with all the dependencies we necessarily have.
```suggestion
impl From<std::str::ParseBoolError> for Error {
```

heaths (approved):

sdk/typespec/src/error/mod.rs
jhendrixMSFT: Fixed

PR #2162, opened 2025-02-14
Minor fixes for test proxy on Windows

LarryOsterman:
Two fixes for test proxy:
1. To get the proxy working on Windows, replace the 0.0.0.0 IP address reported by the test-proxy tool with localhost
2. If `AZURE_TEST_MODE` is live, then return None for the `assets.json` file.

### DIFF
--- a/sdk/core/azure_core_test/src/lib.rs
+++ b/sdk/core/azure_core_test/src/lib.rs
@@ -127,27 +127,25 @@ impl TestContext {
 
         #[cfg(not(target_arch = "wasm32"))]
         {
-            match mode {
-                TestMode::Live => None,
-                _ => {
-                    let path = match find_ancestor_file(self.crate_dir, ASSETS_FILE) {
-                        Ok(path) => path,
-                        Err(_) if mode == TestMode::Record => {
-                            return Path::new("sdk")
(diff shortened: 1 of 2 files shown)


sdk/core/azure_core_test/src/lib.rs
```diff
@@ -127,22 +127,27 @@ impl TestContext {
...
-            let path = match find_ancestor_file(self.crate_dir, ASSETS_FILE) {
-                Ok(path) => path,
-                Err(_) if mode == TestMode::Record => {
-                    return Path::new("sdk")
-                        .join(self.service_dir)
-                        .join(ASSETS_FILE)
-                        .as_path()
+            match mode {
+                TestMode::Live => None,
+                _ => {
```
heaths, line 132: It works, but with 3 more nesting levels. Easier to just:

```rust
if mode == TestMode::Live {
    return None;
}
// ...continue with exisitng code.

heaths: Like Go, Rust is quite happy with early returns to avoid deep nesting.

heaths (approved):

heaths (approved):

PR #2166, opened 2025-02-14
Normalize path separator on Windows

heaths:


sdk/typespec/typespec_macros/tests/compilation-tests.rs
```diff
@@ -141,7 +141,7 @@ pub fn compilation_tests() {
                         file_name: relative_span_path
                             .to_str()
                             .expect("failed to convert span path to string")
-                            .into(),
+                            .replace("\\", "/"),
```
RickWinter, line 144: Why don't you need the into() after doing the replace?

RickWinter (approved):

sdk/typespec/typespec_macros/tests/compilation-tests.rs
heaths: `replace` already turns the string slice into a `String` because it assumes it has to allocate. Would be better as a `Cow`, but I didn't write it. ;)

LarryOsterman: FWIW, Windows is just fine with either `\\` or `/` as path separators. 

The only place it might matter is if you're building the command line to a program.

Issue #2002, opened 2025-01-16
Async file system abstractions


heaths:
Related to #1652 and as brought up in #1997, we probably need to create async file system abstractions as well because we do enough file IO e.g., in `azure_identity`, that we can't pin to a specific async runtime like `tokio` or `async-std`.

heaths:
Discussed: we'll just have a pluggable process launcher in `azure_identity`. Could be a lift-and-shift from `typespec_client_core` or similar to what we already have in identity.

heaths:
I'm closing this one. Apparently we don't use any of our FS implementations anymore, and the only reason we even provide a `typespec_client_core::fs` module with some types therein is to implement `SeekableStream` for the common case using `tokio::fs`.

I'll open a separate issue tracking whether we want to keep that.

If we ever need an FS abstraction, I imagine it'd probably show up first in `azure_identity` - the same place we moved the bespoke async process abstraction because that's the only crate in which we need to launch child processes currently. /cc @chlowell

PR #2175, opened 2025-02-14
Add run_command abstraction to azure_identity

### LINKED ISSUE #1652: Async process abstractions

arpad-m:
Adds a `run_command` function to `azure_identity` and uses it in the one place where we used `async-process` before. We provide two implementations:

* one based on tokio, behind an optional feature
* the other (used by default) is a manual implementation with a oneshot channel and a dedicated thread

successor of #1654
fixes #1652

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,43 +202,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-process"
-version = "2.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "63255f1dc2381611000436537bbedfe83183faa303a5a0edaf191edef06526bb"
-dependencies = [
- "async-channel 2.3.1",
- "async-io",
- "async-lock",
- "async-signal",
- "async-task",
- "blocking",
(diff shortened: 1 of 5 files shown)


arpad-m:
Note that I have not tested the patch, but the implementation seems pretty straightforward to me.

heaths (changes requested):
Thank you for your contribution!

While this does solve the problem, it does not align with our design of providing a means of abstraction to callers e.g., implementing a trait to run on their own async runtime like our `HttpClient`. That is more what I had in mind when I opened #2002 (more about file IO , but I meant to do async child process forks as well).

We'd certainly be happy to review work more aligned with that. It should go in `azure_core` in, perhaps, a module named `process`.

arpad-m:
> It should go in `azure_core` in, perhaps, a module named `process`.

Sounds good to me. I suppose you would like to have it controlled by a dyn trait that is stored in a static const default?

arpad-m:
Thanks for the review. filed a successor PR: #2184

PR #2184, opened 2025-02-18
Add ProcessRunner trait to azure_core

### LINKED ISSUE #1652: Async process abstractions

arpad-m:
Adds a `ProcessRunner` trait to `azure_core`, similar to `HttpClient`, and uses it in the one place in `azure_identity` where we used `async-process` before. We provide two implementations:

* one based on tokio, behind an optional feature
* the other (used by default) is a manual implementation with a oneshot channel and a dedicated thread

successor of #1654, #2175
fixes #1652

### DIFF
--- a/sdk/core/azure_core/Cargo.toml
+++ b/sdk/core/azure_core/Cargo.toml
@@ -57,6 +57,7 @@ reqwest_gzip = ["typespec_client_core/reqwest_gzip"]
 reqwest_rustls = ["typespec_client_core/reqwest_rustls"]
 test = ["typespec_client_core/test"]
 tokio_fs = ["typespec_client_core/tokio_fs"]
+tokio_process = ["dep:tokio"]
 tokio_sleep = ["typespec_client_core/tokio_sleep"]
 xml = ["typespec_client_core/xml"]
 
--- a/sdk/core/azure_core/Cargo.toml
+++ b/sdk/core/azure_core/Cargo.toml
@@ -57,7 +57,7 @@ reqwest_gzip = ["typespec_client_core/reqwest_gzip"]
 reqwest_rustls = ["typespec_client_core/reqwest_rustls"]
 test = ["typespec_client_core/test"]
 tokio_fs = ["typespec_client_core/tokio_fs"]
-tokio_process = ["dep:tokio"]
+tokio_process = ["dep:tokio", "tokio/process"]
 tokio_sleep = ["typespec_client_core/tokio_sleep"]
 xml = ["typespec_client_core/xml"]
 
--- a/sdk/core/azure_core/src/process/mod.rs
+++ b/sdk/core/azure_core/src/process/mod.rs
@@ -1,27 +1,51 @@
+use async_trait::async_trait;
+use std::io::Error;
+use std::{ffi::OsStr, process::Output};
+
 #[cfg(not(feature = "tokio_process"))]
 mod thread;
 
 #[cfg(not(feature = "tokio_process"))]
-pub use thread::run_command;
+pub use thread::ThreadRunner;
 
 #[cfg(feature = "tokio_process")]
-pub use self::tokio::run_command;
+pub use tokio::TokioRunner;
 
 #[cfg(feature = "tokio_process")]
 mod tokio {
-    use std::io::Error;
-    use std::{ffi::OsStr, process::Output};
+    use super::*;
 
-    /// Run a command with the given arguments until it terminates, returning the output
-    pub async fn run_command<S, I, A>(program: S, args: I) -> Result<Output, Error>
-    where
-        S: AsRef<OsStr>,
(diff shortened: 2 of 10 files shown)


sdk/core/azure_core/src/process/mod.rs
```diff
@@ -0,0 +1,51 @@
+use async_trait::async_trait;
```
heaths, line 1: Missing copyright headers found in other files.

sdk/core/azure_core/Cargo.toml
```diff
@@ -50,6 +50,7 @@ reqwest_gzip = ["typespec_client_core/reqwest_gzip"]
 reqwest_rustls = ["typespec_client_core/reqwest_rustls"]
 test = []
 tokio_fs = ["typespec_client_core/tokio_fs"]
+tokio_process = ["dep:tokio", "tokio/process"]
 tokio_sleep = ["typespec_client_core/tokio_sleep"]
```
heaths, line 54: Note to self: we're creating enough `tokio`-based features that maybe we should have a `tokio` feature that pulls them all in.

sdk/core/azure_core/src/process/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+mod thread;
+
+#[cfg(not(feature = "tokio_process"))]
+pub use thread::ThreadRunner;
+
+#[cfg(feature = "tokio_process")]
+pub use tokio::TokioRunner;
+
+#[cfg(feature = "tokio_process")]
+mod tokio {
```
heaths, line 15: It'd be better to just isolate these into separate modules and conditionally pull them in if enabled.

sdk/core/azure_core/src/process/mod.rs
```diff
@@ -0,0 +1,51 @@
+use async_trait::async_trait;
+use std::io::Error;
+use std::{ffi::OsStr, process::Output};
+
+#[cfg(not(feature = "tokio_process"))]
+mod thread;
```
heaths, line 6: You should never use mutually-exclusive features as called out in rust docs. The resolver does a union.

sdk/identity/azure_identity/Cargo.toml
```diff
@@ -26,9 +26,6 @@ openssl = { workspace = true, optional = true }
 pin-project.workspace = true
 typespec_client_core = { workspace = true, features = ["derive"] }
 
-[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-async-process.workspace = true
-
```
heaths, line 31: This should be removed from the workspace if it's the only one referencing it left.

heaths (changes requested):
Thank you. This is closer, but there are still a few things we'd like to see different. I can finish this up. Just make sure that maintainers can push to your branch and I'll get to it soon. I have some other higher priority work I need to attend to first but will leave this in my queue.

sdk/core/azure_core/src/process/mod.rs
arpad-m: There is no feature that's mutually exclusive with `tokio_process`: either it's present, or it's not. If there would be multiple features, from multiple backends, then there would be a problem indeed. But this isn't the case here.

sdk/core/azure_core/src/process/mod.rs
arpad-m: Added

sdk/core/azure_core/src/process/mod.rs
arpad-m: Done

heaths:
To note: I haven't forgotten about this and will be working on it soon. I have a couple more things I want to get done first.

sdk/core/azure_core/src/process/mod.rs
heaths: I see this matches what we do for `HttpClient`. While we could maybe keep this to make it easy for us to call - or even external callers, though I don't imagine much reason to - how it gets passed to credentials will have to change. I was playing around with some different options including a "static trait" for a runner, but the DX for that because of a bug/design issue in rustc isn't great.

So a slight adaptation of this + the options pattern in https://gist.github.com/heaths/ab062b63cac36bc1ef774a43c74dcf63 and this should work. Since you said your branch is locked down, I'll start a new one but add you as `Co-Authored-By`. Thanks for all your work here!

arpad-m: > I see this matches what we do for `HttpClient`.

Yeah i've been inspired by that.

> Thanks for all your work here!

You are welcome. Great that I could have been of help.

I will keep this PR open until you make your pull request, but feel free to close it now if you prefer that.

heaths:
Replaced by #2328. I carried over all your commits so you'll get "credit" for your work. Thanks so much for contributing to the Azure SDKs!

PR #2187, opened 2025-02-18
Improve errors for getting package properties

weshaggard:


eng/scripts/Language-Settings.ps1
```diff
@@ -67,7 +67,8 @@ function Get-AllPackageInfoFromRepo ([string] $ServiceDirectory) {
   }
   catch {
     # This is soft error and failure is expected for python metapackages
-    LogError "Failed to get all package properties"
+    LogError "Failed to get all package properties. $($_.Exception.Message)"
+    return $null
```
heaths, line 71: 1. Nit: 
```suggestion
    LogError "Failed to get all package properties: $($_.Exception.Message)"
    return $null
```
2. `return $null`? Seems rather exceptional - terminal, even. Or is the caller checking for valid data, which begs the question why not just throw here?

RickWinter (approved):

eng/scripts/Language-Settings.ps1
```diff
@@ -67,7 +67,8 @@ function Get-AllPackageInfoFromRepo ([string] $ServiceDirectory) {
   }
   catch {
     # This is soft error and failure is expected for python metapackages
-    LogError "Failed to get all package properties"
+    LogError "Failed to get all package properties: $($_.Exception.Message)"
+    return $null
   }
```
hallipr, line 72: I would just remove the catch block.  It looks like I copied it from python and it's not really appropriate for rust:
> # This is soft error and failure is expected for python metapackages

This isn't a soft error in rust. `cargo metadata` should work and all of the code in the try block should succeed.

In this case, it's only there to ensure the caller returns back to their original location after the Push-Location

heaths (approved):

PR #2194, opened 2025-02-18
Preinstall test-proxy before integration tests

heaths:


hallipr (approved):

RickWinter (approved):
The looks good to me. I'm not the expert on the pipelines though.

PR #2207, opened 2025-02-20
[Cosmos] run cosmos emulator tests during PRs

analogrelay:
This PR adds some logic to the `azure_data_cosmos` CI pipeline to run tests against the Cosmos emulator. This is largely based on [similar logic in the Go SDK](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/azcosmos/ci.yml).

To implement this, I added an `AdditionalStages` parameter to `archetype-sdk-client.yml`, again mirroring what I [saw in the Go SDK](https://github.com/Azure/azure-sdk-for-go/blob/d38b662e68ec49a96c60523544aee7865ed8a1b0/eng/pipelines/templates/jobs/archetype-sdk-client.yml#L76-L78), which places the steps after the default pack/test/analyzer steps but BEFORE any release steps.

We install the Windows emulator primarily because that's the logic that already exists in `eng/common`. There is a Linux emulator, that runs as a Docker image, but I wanted to be consistent here.

In addition, I tweaked `recording::start` to activate tracing even when test proxy wasn't in use. This allows us to use `RUST_LOG` to turn tracing on for live tests.


analogrelay:
I see now that the PR builds aren't quite structured the same. I'm looking in to an approach using the `Test-Setup.ps1` file

heaths:
> I see now that the PR builds aren't quite structured the same. I'm looking in to an approach using the `Test-Setup.ps1` file

Work with @hallipr and @LarryOsterman if needed. They devised that system.

analogrelay:
I'm going to go ahead and close this in the short term. I need to do some other work to figure out the right approach here.

PR #2221, opened 2025-02-24
Don't replace existing change log entry

hallipr:
- Default `$ReplaceLatestEntryTitle` to `$false`
- Clean up the naming and handling of the variable passed to `Update-ChangeLog.ps1 -Unreleased` parameter.

### DIFF
--- a/eng/scripts/Update-PackageVersion.ps1
+++ b/eng/scripts/Update-PackageVersion.ps1
@@ -33,7 +33,7 @@ Param (
   [string] $PackageName,
   [string] $NewVersionString,
   [string] $ReleaseDate,
-  [boolean] $ReplaceLatestEntryTitle = $true
+  [boolean] $ReplaceLatestEntryTitle = $false
 )
 
 . (Join-Path $PSScriptRoot '../common/scripts/common.ps1')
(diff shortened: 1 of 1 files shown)


eng/scripts/Update-PackageVersion.ps1
```diff
@@ -33,7 +33,7 @@ Param (
   [string] $PackageName,
   [string] $NewVersionString,
   [string] $ReleaseDate,
-  [boolean] $ReplaceLatestEntryTitle = $true
+  [boolean] $ReplaceLatestEntryTitle = $false
```
heaths, line 36: 1. `$false` is already default so this is unnecessary.
2. Why does engsys use `[boolean]`? You should use `[switch]` which not only works like a boolean, but you can easily specify as `-ReplaceLatestEntryTitle` without a value, which treats it as `$true`. You can still pass explicit values if you need by passing `:{value}` e.g. `-ReplaceLatestEntryTitle:$false`.

/cc @weshaggard since I see this often. It's not idiomatic to PowerShell and makes calling scripts harder for interactive use because users have to pass a value, which isn't idiomatic. Do you pass a value to `-Verbose`? To `-Force`? etc.

heaths (approved):

eng/scripts/Update-PackageVersion.ps1
hallipr: 1. sure
2. This script is only expected to be called by other scripts or from yaml. Having a different syntax for one parameter is odd.

```ps
  & "$EngDir/scripts/Update-PackageVersion.ps1" -ServiceDirectory $ServiceDirectory `
    -PackageName $PackageName `
    -NewVersionString $Version `
    -ReleaseDate $ReleaseDate `
    -ReplaceLatestEntryTitle:$ReplaceLatestEntryTitle
```

RickWinter (approved):

PR #2222, opened 2025-02-24
Create Issue template

RickWinter:


ronniegeraghty (approved):

.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,56 @@
+name: Feature request
+description: Suggest an idea for this project
+title: "[Feature Request]: "
+labels: ["needs-triage"]
```
heaths, line 4: Also `feature-request` and probably `customer-reported`, as well as for the bug.

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,70 @@
+name: Bug report
+description: Report a bug in the Azure SDK for Rust
+title: "[BUG] "
+labels: ["needs-triage"]
```
heaths, line 4: Also bug and customer-reported.

.github/ISSUE_TEMPLATE/feature-request.yml
ronniegeraghty: Customer-reported isn't needed. The GitHub bot will figure that out and label the issue with `customer-reported` if it is.

RickWinter: I'd vote for moving away from "feature-request" label since type "Feature" lets us know its a Feature.

heaths: Then can we get rid of the "[BUG] " and "[Feature Request] " prefixes in titles? Same reason: it's noise. I can already see that from tags/type.

RickWinter: I'm not opposed to removing that prefix. If we are in agreement I'll remove it.

heaths: Especially with "[Feature Request]: " - that's 19 characters wasted in a title and could make triaging harder.

PR #2226, opened 2025-02-24
First attempt at "rationalizing" AMQP error codes.

LarryOsterman:
# Changes to AMQP and Eventhubs to improve error handling and make errors more actionable.

This is a slightly ugly set of changes to AMQP and eventhubs error handling

There are three main changes here:
1) Enable describing errors as originating from AMQP and/or EventHubs. This enables callers to discover and act on actionable errors from AMQP and/or EventHubs.
2) Moved almost all AMQP errors from fe2o3 into the AMQP layer. Currently the layering is too fine grained, there are undoubtedly reductions in the quantity of errors currently generated. This is the bulk of the changes in the PR.
3) Changed the way the AMQP `send()` API works - instead of returning a `Result<()>`, it returns the outcome of the send to the caller.

Note: I have mixed feelings about the first change - Heath and I have chatted about it off and on, my justification for this is that it allows a caller to discover the distinguished AMQP and EventHubs errors contained in the `err.source()`. My concern is that this will be an implication that *any* service which wants to have distinguished and actionable errors will want to do the same, and that is concerning.

## Copilot generated summary

This pull request includes various changes to the `azure_core_amqp` SDK, focusing on improving error handling, updating dependencies, and enhancing the AMQP connection management. Below is a summary of the most important changes:

### Error Handling Improvements:
* Introduced a new `AmqpConnectionError` enum to handle various connection-related errors and implemented the `From` trait for conversions to `AmqpErrorKind` and `azure_core::Error`.
* Updated `AmqpErrorKind` to include additional error types and improved the `AmqpError` struct to handle these new error kinds.
* Modified error handling in `Fe2o3ClaimsBasedSecurity` and `Fe2o3AmqpConnection` implementations to use the new error types and conversion methods. [[1]](diffhunk://#diff-c569aa674a24402ed7f7942e3678f39323f51e9959097893621d26da589be33bL45-R59) [[2]](diffhunk://#diff-3df1e5d2a77c38245c74ae9a38af5bcb7744b99d894b2e6936df1a4a6cc5905dL103-R109)

### Dependency Updates:
* Updated `Cargo.toml` files to include new dependencies and features, such as adding `typespec` and `url` to the `azure_core_amqp` package.
* Enabled the `tracing` feature for the `fe2o3-amqp-management` dependency.

### Documentation and Descriptions:
* Updated the package description in `sdk/core/azure_core_amqp/Cargo.toml` to accurately reflect that it is a Rust client library for the AMQP protocol.

### Test Enhancements:
* Added tracing initialization in tests to improve debugging and error reporting.
* Added print statements to log errors during test execution for better visibility.

### Dictionary Updates:
* Added new AMQP-related terms to the `.dict.txt` file to improve spell-checking and code quality.


heaths:
> Note: I have mixed feelings about the first change - Heath and I have chatted about it off and on, my justification for this is that it allows a caller to discover the distinguished AMQP and EventHubs errors contained in the err.source(). My concern is that this will be an implication that any service which wants to have distinguished and actionable errors will want to do the same, and that is concerning.

But just as we have an `Http` error classification, I think it's fine to have an `Amqp` classification. You can put more details in there.

sdk/core/azure_core_amqp/.dict.txt
```diff
@@ -1,5 +1,7 @@
 mgmt
 sasl
+amqp
```
heaths, line 3: IMO, you could put "amqp" in the main dictionary (.cspell.json file). We'll probably end up referring to it at least as existing enough. Not a big deal either way.

sdk/core/azure_core_amqp/Cargo.toml
```diff
@@ -26,10 +26,12 @@ fe2o3-amqp-management = { workspace = true, optional = true }
 fe2o3-amqp-types = { workspace = true, optional = true }
 serde_amqp = { workspace = true, optional = true }
 serde_bytes = { workspace = true, optional = true }
+typespec = { workspace = true, features = ["amqp_errors"] }
```
heaths, line 29: I don't think we need to specialize on "errors". Maybe just call the feature "amqp". We have "http" now. I don't want feature to be so granular that you feel like you're ordering off a Cheesecake Factory menu.

sdk/typespec/Cargo.toml
```diff
@@ -24,3 +24,5 @@ thiserror.workspace = true
 default = ["http", "json"]
 http = ["dep:http-types"]
 json = ["dep:serde_json"]
+amqp_errors = []
+eventhubs_errors = []
```
heaths: I know AMQP and Event Hubs aren't a perfect circle, but let's not get too granular. Can we just do "amqp"? Services that expose their own details - if they use the Azure error response "standard" (we need to improve the variations of this, granted) - customers can get that. Let's not pile on different services because I agree: that's a bad precedent.

Most (all?) other languages don't. .NET, for example, still just has the `RequestFailedException` for 99% of service-response errors.

sdk/eventhubs/.dict.txt
```diff
@@ -2,3 +2,6 @@ amqps
 consts
 eventdata
 mgmt
+amqp
```
heaths, line 5: Yeah - just put "amqp" in the general location for all words. It's general enough within Azure SDKs. I could see keeping "amqps" isolated for now, though.

sdk/eventhubs/azure_messaging_eventhubs/examples/richards_test.rs
```diff
@@ -0,0 +1,98 @@
+use core::f32;
```
heaths, line 1: 1. Copyright header.
2. Let's not use people's names in code. Probably not exactly PII, but feels close enough. Also raises more questions than answers.
3. If it's a test, why is it not in the `tests/` directory as an integration test?

heaths: ...even if live-only.

sdk/typespec/Cargo.toml
heaths: I also don't think it's outside the realm of reason for `azure_core_amqp` to have its own `Error` as well that can `into()` a `typespec::Error` as `Other` or maybe we need a per-service classification that uses a `kind(&self) -> 'static str` method.

That also means `azure_core_amqp` could define it's own shorthand `Result<T, azure_core_amqp::Error>` if it makes method signatures easier.

sdk/core/azure_core_amqp/src/management.rs
```diff
@@ -24,6 +24,105 @@ pub trait AmqpManagementApis {
...
+            match &self {
+                AmqpManagementError::AmqpManagementAlreadyAttached => {
+                    f.write_str("AMQP Management is already attached")
+                }
+                AmqpManagementError::AmqpManagementNotAttached => {
+                    f.write_str("AMQP Management is not attached")
+                }
+                AmqpManagementError::InvalidManagementResponse(s) => {
+                    if let Some(e) = self.source() {
+                        f.write_fmt(format_args!("Invalid Management Response: {}: {}", s, e))
```
heaths, line 61: Tip: as long as you don't need an expression - just a simpel var reference - you can do:
```suggestion
                        f.write_fmt(format_args!("Invalid Management Response: {s}: {e}"))
```
IMO, cleaner, but either is fine.

sdk/core/azure_core_amqp/src/connection.rs
```diff
@@ -79,6 +79,133 @@ impl AmqpConnection {
     }
 }
 
+pub(crate) mod error {
+
+    use crate::{error::AmqpErrorKind, AmqpError};
+
+    pub enum AmqpConnectionError {
+        ConnectionNotSet,
```
heaths, line 87: I feel like this is getting too granular. Are all these actionable, or could you have fewer and just vary the string message like how we do mostly for `azure_core::Error`? Granted, we could probably use *slightly* more `ErrorKind` variants (I've been feeling like it hasn't been enough lately), but there's no much point in all these different variants if the caller can't actually do anything about them. At that point, you're potentially making them do more work for no gain i.e., just to display a message and terminate the connection, that client, the process, etc.

heaths (changes requested):
We shouldn't do this with errors. I think AMQP is justified to have its own, but I also think you're getting way too granular for no good reason. Matching on all those different kinds (potentially; many probably won't) just to show a different message is pointless when you can just vary the message. Heck, you could even define a bunch of string constants if you reuse the same message in different places or have some internal error "constructors" that do that for you.

sdk/core/azure_core_amqp/.dict.txt
LarryOsterman: AMQP *is* in the main dictionary. But it's not picked up for reasons which I don't fully understand.

sdk/typespec/Cargo.toml
LarryOsterman: Eventhubs is gone in current PR - Rick pointed out that typespec cannot have references to anything Azure specific, so Eventhubs goes away in favor of "Other".

I'm starting an email thread with you and Ashley about how we want to handle service specific (or protocol specific suites).

sdk/eventhubs/azure_messaging_eventhubs/examples/richards_test.rs
LarryOsterman: It was an oops - this shouldn't have been checked in.

sdk/core/azure_core_amqp/src/management.rs
LarryOsterman: Yeah, habits die hard (too much time in .Net :)).

sdk/core/azure_core_amqp/src/connection.rs
LarryOsterman: That's actually intentional right now.

I'm starting off with an incredibly granular set of errors (basically the entire fe2o3 errors) and my plan is to collapse many of the errors into more generic errors as I start testing more failure conditions.

Right now, I'm not sure exactly which errors get returned in what circumstances - in other words, which of these errors is actionable or not. A bunch of them express coding errors on the part of the caller and thus are clearly developer diagnostics (which don't need to be exposed), but what I'd like to do is the fine grained version for now and then refine it further based on experience.

I've actually already reduced the set of errors a bunch in areas where I am pretty sure about the use scenarios, and I plan on doing more of this.

But the PR is pretty ungainly right now and I hope that subsequent PRs can be dramatically smaller.

sdk/core/azure_core_amqp/.dict.txt
heaths: In what file extension(s) is the word being flagged? Maybe I can try down why.

Note: this isn't blocking. Just trying to reduce maintenance hassles.

sdk/typespec/Cargo.toml
heaths: Email or Teams? I'd prefer the latter: in our channel, in fact. Posterity and all.

No other language really has this. It comes down to the actionable comment below: what can devs actually do with this information? In most cases I suspect the answer is, at best: log/trace it. You don't need - or even want, I maintain - different kinds for that. Heck, look at Go: most of their errors are merely strings (perhaps with nested error messages, but still strings).

Error details for complexity's sake make for an unnecessarily complicated API.

sdk/core/azure_core_amqp/.dict.txt
LarryOsterman: azure-sdk-for-rust/sdk/eventhubs/azure_messaging_eventhubs/src/error.rs.

And oh, this is *lovely*. I saved cspell.json (with no modifications) and the errors disappeared. So this may be a VS code bug of some kind.

LarryOsterman:
> I like the recent changes, but lets figure out the plan for errors...or would you like to merge first and figure that out later? I'll unblock if you want to go that route.

I'd like to get this in. It's a functional solution and it will let me attempt to see what this solution looks like in practice.

PR #2233, opened 2025-02-25
Replace http-types with http

ntrinquier:
As mentioned in https://github.com/Azure/azure-sdk-for-rust/pull/1937, `http-types` is not being maintained (and neither are some its dependencies).

This replaces it with the `http` crate which seems to be the ecosystem standard (10x more recent downloads).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,7 +83,7 @@ fe2o3-amqp-types = { version = "0.12" }
 futures = "0.3"
 getrandom = { version = "0.2", features = ["js"] }
 hmac = { version = "0.12" }
-http-types = { version = "2.12", default-features = false }
+http = { version = "1.2", default-features = false }
 log = "0.4"
 oauth2 = { version = "5.0.0", default-features = false }
 once_cell = "1.18"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -91,23 +91,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "anyhow"
-version = "1.0.94"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c1fd03a028ef38ba2276dce7e33fcd6369c158a1bca17946c4b1b701891c1ff7"
-
-[[package]]
-name = "async-channel"
-version = "1.9.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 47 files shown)


heaths:
This does get rid of a lot of older dependencies as well, since it seemed to pull in strict dependencies (probably using the older resolver; I never looked exactly). Might be worth it on that alone.

Cargo.toml
```diff
@@ -83,7 +83,7 @@ fe2o3-amqp-types = { version = "0.12" }
 futures = "0.3"
 getrandom = { version = "0.2", features = ["js"] }
 hmac = { version = "0.12" }
-http-types = { version = "2.12", default-features = false }
+http = { version = "1.2", default-features = false }
```
heaths, line 86: Based on all the build errors, we need at least `std`.

heaths (changes requested):
Thank you for your contribution! I'm leaning toward taking this, in part, but reconsidering whether we even should expose third-party `StatusCode` from our error and response types. After GA, we have pretty strict no-breaking changes policy, though it's good to see that `http` did declare a 1.x stable release so, as long as they stick to semver, we should be stable.

Still, the only reason we expose `Method` from `azure_core` is because other crates need it to call into client methods. It's not meant to be *truly* customer public. `StatusCode` obviously is, but we could just as easily return our own with, optionally, `From` impls to `http::StatusCode`...perhaps behind a feature flag so the dependency is optional (if not already transitive) by default.

heaths:
> because new major versions of ubiquitous dependencies like http are always difficult to pick up.

True; however, I did check and we already have transitive dependencies on `http` anyway, so putting any `From<>` impl behind a feature flag would be moot.

heaths:
We're going to define our own `StatusCode` and probably `Method` as well. These are straight forward enough and given limited use, there's really no reason for a 3P dependency just for these. But I'm starting a major refactor of `azure_core` so I'll take the opportunity to define those as well.

heaths:
...though, we may end up having to use constants anyway. I've made `StatusCode` an ["extensible enum"](https://azure.github.io/azure-sdk/rust_introduction.html#rust-enum-extensible) and we're discussing it internally (open to feedback externally as well, of course!), but several of us agree the DX on constants is just...bad.

Issue #2198, opened 2025-02-19
Verify crate keywords do not exceed 5


heaths:
We had several crates get rejected from publishing because we had more than 5 keywords. We should check for this so we're not getting caught by this *right* at the end.

heaths:
We should also validate that we don't duplicate words in the crate name, since the crates.io [search relevance](https://github.com/rust-lang/crates.io/discussions/9325#discussioncomment-10447715) is based on:

1. Crate name
2. Keywords
3. Description
4. README

/cc @ronniegeraghty

ronniegeraghty:
- `azure_core`:
  - keywords:
    - cloud
    - services
    - API
    - REST
    - HTTP
  - description: `azure_core` provides shared primitives, abstractions, and helpers for modern Rust Azure SDK client libraries. It includes functionalities such as configuring service clients, accessing HTTP response details, and handling errors consistently.

- `azure_core_amqp`:
  - keywords:
    - messaging
    - communication
    - non-blocking
    - transport
    - interface
  - description: `azure_core_amqp` provides core AMQP protocol support for the Azure SDK for Rust, including functionalities for establishing connections, managing sessions, and handling message exchanges over the AMQP protocol.

- `azure_messaging_eventhubs`:
  - keywords:
    - streaming
    - data
    - ingestion
    - processing
    - real-time
  - description: `azure_messaging_eventhubs` provides client libraries for interacting with Azure Event Hubs, allowing developers to build event-driven applications and process large volumes of data.

- `azure_security_keyvault_secrets`:
  - keywords:
    - encryption
    - secure
    - credentials
    - access
    - API-key
  - description: `azure_security_keyvault_secrets` provides client libraries for managing secrets in Azure Key Vault, allowing developers to securely store and retrieve sensitive information such as API keys and passwords.

- `azure_security_keyvault_keys`:
  - keywords:
    - encryption
    - secure
    - credentials
    - access
    - protection
  - description: `azure_security_keyvault_keys` provides client libraries for managing encryption keys in Azure Key Vault, allowing developers to securely store, retrieve, and manage cryptographic keys used for data encryption and decryption.

- `azure_identity`:
  - keywords:
    - authorization
    - security
    - token
    - credentials
    - authentication
  - description: `azure_identity` provides a unified API for authenticating with Azure services using various authentication methods, simplifying the process of acquiring tokens and managing authentication flows.

heaths:
[...]

PR #2236, opened 2025-02-25
[Cosmos] reduce azure_data_cosmos to 5 keywords

analogrelay:
>Note: [crates.io](https://crates.io/) allows a **maximum of 5 keywords**. Each keyword must be ASCII text, have at most 20 characters, start with an alphanumeric character, and only contain letters, numbers, _, - or +.
> 
> https://doc.rust-lang.org/cargo/reference/manifest.html#the-keywords-field

🤦🏻‍♀️ 

I guess I didn't think to check that, and the `--dry-run` we do doesn't seem to verify it.

For now, I'm just removing the `database` keyword. We can always revise these later if we decide to swap out one of the existing keywords for something else. Most of the other keywords we have are shared among other `azure_` crates.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/Cargo.toml
+++ b/sdk/cosmos/azure_data_cosmos/Cargo.toml
@@ -10,7 +10,7 @@ repository.workspace = true
 rust-version.workspace = true
 homepage = "https://github.com/azure/azure-sdk-for-rust"
 documentation = "https://docs.rs/azure_data_cosmos"
-keywords = ["sdk", "azure", "rest", "cloud", "cosmos", "database"]
+keywords = ["sdk", "azure", "rest", "cloud", "cosmos"]
 categories = ["api-bindings"]
 
 [dependencies]


sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -10,7 +10,7 @@ repository.workspace = true
 rust-version.workspace = true
 homepage = "https://github.com/azure/azure-sdk-for-rust"
 documentation = "https://docs.rs/azure_data_cosmos"
-keywords = ["sdk", "azure", "rest", "cloud", "cosmos", "database"]
+keywords = ["sdk", "azure", "rest", "cloud", "cosmos"]
```
heaths, line 13: I'd recommend keeping "database" and drop "azure", which is in the crate name already and takes precedence in search. See #2198

heaths (approved):

sdk/cosmos/azure_data_cosmos/Cargo.toml
analogrelay: I absolutely agree in general, but would like to get this landed to unblock the release pipeline and then I'll open a bikeshed for us to refine these for 0.23.0

Pilchie (approved):

PR #2245, opened 2025-02-26
Enable Custom Endpoints for EventHubs

LarryOsterman:
# Enable custom endpoints for EventHubs service

This enables the use of the AMQP recording server.

Bonus feature:
- Restructured the EventHubs service to centralize management of connections and authorization. This creates a "connection manager" API which centralizes connection establishment and authorization on those connections. It consolidates a bunch of common code between the EventHubs producer and EventHubs consumer.

## Github Copilot Summary
This pull request introduces several significant changes to the Azure Event Hubs SDK, primarily focusing on enhancing the AMQP connection management and simplifying the `ConsumerClient` implementation. The key changes include the addition of a new `ConnectionManager` struct, updates to the `ConsumerClient` to use this new manager, and various improvements to the AMQP connection options.

### Connection Management Enhancements:
* **New `ConnectionManager` struct**: Added to manage connections, handle authorization, and support connection recovery. (`sdk/eventhubs/azure_messaging_eventhubs/src/common/connection_manager.rs`)
* **Integration with `ConsumerClient`**: Updated `ConsumerClient` to utilize the new `ConnectionManager` for managing AMQP connections. (`sdk/eventhubs/azure_messaging_eventhubs/src/consumer/mod.rs`) [[1]](diffhunk://#diff-0ab7a2a86e5b369a2ec23636e593d0db5011a00d9d293e78e2f4bd5e2d420303L8-R21) [[2]](diffhunk://#diff-0ab7a2a86e5b369a2ec23636e593d0db5011a00d9d293e78e2f4bd5e2d420303L41-R55) [[3]](diffhunk://#diff-0ab7a2a86e5b369a2ec23636e593d0db5011a00d9d293e78e2f4bd5e2d420303L94-R112) [[4]](diffhunk://#diff-0ab7a2a86e5b369a2ec23636e593d0db5011a00d9d293e78e2f4bd5e2d420303L155-R153) [[5]](diffhunk://#diff-0ab7a2a86e5b369a2ec23636e593d0db5011a00d9d293e78e2f4bd5e2d420303R220-R239) [[6]](diffhunk://#diff-0ab7a2a86e5b369a2ec23636e593d0db5011a00d9d293e78e2f4bd5e2d420303L381-R400) [[7]](diffhunk://#diff-0ab7a2a86e5b369a2ec23636e593d0db5011a00d9d293e78e2f4bd5e2d420303L411-R428)

### AMQP Connection Options:
* **Extended `AmqpConnectionOptions` struct**: Added several new fields such as `custom_endpoint`, `buffer_size`, and various locales and capabilities options to provide more configuration flexibility. (`sdk/core/azure_core_amqp/src/connection.rs`)
* **Updated `Fe2o3AmqpConnection` implementation**: Modified to use the new options and handle custom endpoints. (`sdk/core/azure_core_amqp/src/fe2o3/connection.rs`) [[1]](diffhunk://#diff-3df1e5d2a77c38245c74ae9a38af5bcb7744b99d894b2e6936df1a4a6cc5905dR56-L63) [[2]](diffhunk://#diff-3df1e5d2a77c38245c74ae9a38af5bcb7744b99d894b2e6936df1a4a6cc5905dR113-R120)

### Test Enhancements:
* **Extended tests for `AmqpConnectionOptions`**: Added tests to verify the new `custom_endpoint` field. (`sdk/core/azure_core_amqp/src/connection.rs`) [[1]](diffhunk://#diff-3b1cde9b879b09dbdc03797027041992fe34df702acdf708bed1df18f358c695R220) [[2]](diffhunk://#diff-3b1cde9b879b09dbdc03797027041992fe34df702acdf708bed1df18f358c695R258-R261)

### Miscellaneous:
* **Updated `.vscode/cspell.json`**: Corrected the dictionary file path pattern. (`.vscode/cspell.json`)

These changes collectively improve the flexibility, reliability, and maintainability of the Azure Event Hubs SDK, particularly in managing AMQP connections and handling various connection configurations.

### DIFF
--- a/.vscode/cspell.json
+++ b/.vscode/cspell.json
@@ -17,7 +17,7 @@
     ".vscode/tasks.json",
     "NOTICE.txt",
     "eng/",
-    "*.dict.txt",
+    "**/.dict.txt",
     "rust-toolchain.toml"
   ],
   "words": [
--- a/sdk/core/azure_core_amqp/src/error.rs
+++ b/sdk/core/azure_core_amqp/src/error.rs
@@ -32,7 +32,7 @@ pub enum AmqpErrorKind {
     TransportImplementationError(Box<dyn std::error::Error + Send + Sync>),
 }
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone)]
 pub struct AmqpDescribedError {
     condition: AmqpSymbol,
     description: Option<String>,
--- a/sdk/core/azure_core_amqp/src/error.rs
+++ b/sdk/core/azure_core_amqp/src/error.rs
@@ -32,7 +32,7 @@ pub enum AmqpErrorKind {
     TransportImplementationError(Box<dyn std::error::Error + Send + Sync>),
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct AmqpDescribedError {
     condition: AmqpSymbol,
     description: Option<String>,
--- a/sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
+++ b/sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
@@ -97,12 +97,10 @@ impl AmqpClaimsBasedSecurityApis for Fe2o3ClaimsBasedSecurity<'_> {
                     .to_offset(time::UtcOffset::UTC)
                     .unix_timestamp()
                     .checked_mul(1_000)
-                    .ok_or_else(|| {
-                        azure_core::Error::message(
-                            azure_core::error::ErrorKind::Amqp,
-                            "Unable to convert time to unix timestamp.",
-                        )
-                    })?,
+                    .ok_or(azure_core::Error::message(
+                        azure_core::error::ErrorKind::Amqp,
+                        "Unable to convert time to unix timestamp.",
+                    ))?,
             )),
         );
         self.cbs
--- a/sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
+++ b/sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
@@ -71,7 +71,7 @@ impl AmqpClaimsBasedSecurityApis for Fe2o3ClaimsBasedSecurity<'_> {
     }
 
     async fn detach(mut self) -> Result<()> {
-        let cbs = self.cbs.take().ok_or(Self::cbs_not_set())?;
+        let cbs = self.cbs.take().ok_or_else(Self::cbs_not_set)?;
         let cbs = cbs.into_inner();
         cbs.close().await.map_err(AmqpError::from)?;
         Ok(())
@@ -97,10 +97,12 @@ impl AmqpClaimsBasedSecurityApis for Fe2o3ClaimsBasedSecurity<'_> {
                     .to_offset(time::UtcOffset::UTC)
                     .unix_timestamp()
                     .checked_mul(1_000)
-                    .ok_or(azure_core::Error::message(
-                        azure_core::error::ErrorKind::Amqp,
-                        "Unable to convert time to unix timestamp.",
-                    ))?,
+                    .ok_or_else(|| {
+                        azure_core::Error::message(
+                            azure_core::error::ErrorKind::Amqp,
+                            "Unable to convert time to unix timestamp.",
+                        )
+                    })?,
             )),
         );
         self.cbs
--- a/sdk/core/azure_core_amqp/src/fe2o3/connection.rs
+++ b/sdk/core/azure_core_amqp/src/fe2o3/connection.rs
@@ -53,6 +53,9 @@ impl AmqpConnectionApis for Fe2o3AmqpConnection {
         options: Option<AmqpConnectionOptions>,
     ) -> Result<()> {
         {
+            let options = options.unwrap_or_default();
+            let mut endpoint = url.clone();
+
             // All AMQP clients have a similar set of options.
             let mut builder = fe2o3_amqp::Connection::builder()
                 .sasl_profile(fe2o3_amqp::sasl_profile::SaslProfile::Anonymous)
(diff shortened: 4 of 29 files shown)


heaths:
Does this replace #2147 then?

.vscode/cspell.json
```diff
@@ -17,7 +17,7 @@
     ".vscode/tasks.json",
     "NOTICE.txt",
     "eng/",
-    "*.dict.txt",
+    "**/.dict.txt",
```
heaths, line 20: Good catch. I wonder if that's why "amqp" was an issue: cSpell currently (and probabl always will, per an issue conversation I was in with them) does not ignore its own configuration files. So it was very possible "amqp" was coming from inside the house.

LarryOsterman:
> Does this replace #2147 then?

Yes. It takes almost all of @richardpark-msft 's excellent work and incorporates it into my branch (removing a bunch of unnecessary `unwraps`).

And then because I've been annoyed at this for months and I am apparently unable to make a simple change without introducing a ton of refactoring, I took a bunch of code which was identical between the producer and consumer and consolidated it into a single class.

And then I realized I should use `.ok_or()` instead of `.ok_or_else()` throughout the codebase.

And all of a sudden a 4 file PR became a 9 file PR.

sdk/core/azure_core_amqp/src/error.rs
```diff
@@ -32,7 +32,7 @@ pub enum AmqpErrorKind {
     TransportImplementationError(Box<dyn std::error::Error + Send + Sync>),
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
```
heaths, line 35: If this is intended to be fully equatable, yo ushould also derive `Eq`. Generally that will be the case. If you only want *truly* `PartialEq`, it's more often `impl`'d because you have to say exactly what is compared. `Eq` is a marker trait - a super trait of `PartialEq` that means "fully equatable", basically.

sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
```diff
@@ -97,12 +97,10 @@ impl AmqpClaimsBasedSecurityApis for Fe2o3ClaimsBasedSecurity<'_> {
                     .to_offset(time::UtcOffset::UTC)
                     .unix_timestamp()
                     .checked_mul(1_000)
-                    .ok_or_else(|| {
-                        azure_core::Error::message(
-                            azure_core::error::ErrorKind::Amqp,
-                            "Unable to convert time to unix timestamp.",
-                        )
-                    })?,
+                    .ok_or(azure_core::Error::message(
```
heaths, line 100: This eagerly allocates the error. The previous way was better.

sdk/core/azure_core_amqp/src/fe2o3/connection.rs
```diff
@@ -53,64 +53,71 @@ impl AmqpConnectionApis for Fe2o3AmqpConnection {
         options: Option<AmqpConnectionOptions>,
     ) -> Result<()> {
         {
+            let options = options.unwrap_or_default();
+            let mut endpoint = url.clone();
```
heaths, line 57: Does this actually need to be cloned? It's already owned at this point, and looking below it doesn't seem it needs to be either. Or is it just because you need it `mut`?

sdk/eventhubs/azure_messaging_eventhubs/src/common/connection_manager.rs
```diff
@@ -0,0 +1,150 @@
...
+/// The connection manager is responsible for managing the connection to the Event Hubs service.
+/// It also handles authorization and connection recovery.
+///
+/// Currently the connection manager only handles a *single* connection, eventually it will manage
+/// a pool of connections to the service.
+pub(crate) struct ConnectionManager {
+    url: Url,
+    application_id: Option<String>,
+    custom_endpoint: Option<Url>,
+    connections: Mutex<OnceLock<Arc<AmqpConnection>>>,
```
heaths, line 31: More commonly this is a `OnceLock<Mutex<T>>` and you shouldn't need an `Arc` then e.g., <https://github.com/Azure/azure-sdk-for-rust/blob/a8211be103b021d3e0132ca08d1202b170c7f6ce/sdk/core/azure_core_test/src/recording.rs#L57>.

sdk/eventhubs/azure_messaging_eventhubs/src/common/connection_manager.rs
```diff
@@ -0,0 +1,150 @@
...
+            return Ok(());
+        }
+
+        trace!("Creating connection for {}.", self.url);
+        let connection = Arc::new(AmqpConnection::new());
+        connection
+            .open(
+                self.application_id
+                    .clone()
+                    .unwrap_or(uuid::Uuid::new_v4().to_string()),
```
heaths, line 58: Again, this eagerly allocates to `unwrap_or_else` is generally better. Probably only use `unwrap_or` for primitives.

sdk/eventhubs/azure_messaging_eventhubs/src/common/connection_manager.rs
```diff
@@ -0,0 +1,150 @@
...
+        Ok(())
+    }
+
+    pub(crate) async fn get_connection(&self) -> Result<Arc<AmqpConnection>> {
+        let connections = self.connections.lock().await;
+
+        let connection = connections
+            .get()
+            .cloned()
+            .ok_or(EventHubsError::from(ErrorKind::MissingConnection))?;
```
heaths, line 91: Same here.

sdk/eventhubs/azure_messaging_eventhubs/src/common/connection_manager.rs
```diff
@@ -0,0 +1,150 @@
...
+        Ok(())
+    }
+
+    pub(crate) async fn authorize_path(
+        &self,
+        connection: &Arc<AmqpConnection>,
+        path: &Url,
+        credential: Arc<dyn azure_core::credentials::TokenCredential>,
+    ) -> Result<AccessToken> {
+        debug!("Authorizing path: {path}");
```
heaths, line 111: I see you using `trace!()` in other places. Do you have, in your mind, a clear picture of `debug` vs `trace`? I know we haven't drafted guidelines yet (on my long todo list), but I'd suggest trying to at least be consistent in your lib. Not block, though.

sdk/eventhubs/azure_messaging_eventhubs/src/common/mod.rs
```diff
@@ -55,22 +56,20 @@ impl ManagementInstance {
         }
         let name: String = response
             .get(EVENTHUB_PROPERTY_NAME)
-            .ok_or_else(|| EventHubsError::from(ErrorKind::InvalidManagementResponse))?
+            .ok_or(EventHubsError::from(ErrorKind::InvalidManagementResponse))?
```
heaths, line 59: It was correct the first way. I'll not repeat for brevity.

sdk/eventhubs/azure_messaging_eventhubs/src/error.rs
```diff
@@ -1,27 +1,14 @@
...
-    /// The shared access key name is missing.
-    MissingSharedAccessKeyName,
-
-    /// The endpoint is missing.
-    MissingEndpoint,
-
     /// The session was missing for the partition.
     MissingSession,
-
+    //
```
heaths, line 11: ???

sdk/eventhubs/azure_messaging_eventhubs/tests/consumer.rs
```diff
@@ -164,6 +164,7 @@ async fn receive_lots_of_events() -> Result<(), Box<dyn Error>> {
         .open(host.as_str(), eventhub.as_str(), credential.clone())
         .await?;
 
+    info!("Client open, create receiver.");
```
heaths, line 167: Per previous comment, this doesn't doesn't seem to rise to the level of `info`. FWIW, after talking with Catalina we'll probably end up tracing methods as `info`, stuff most developers might want to see as `debug`, and really verbose stuff as `trace`.

heaths (changes requested):

.vscode/cspell.json
LarryOsterman: I don't know for sure - I noticed it because cspell triggered on a .dict.txt file on my local machine and while trying to figure it out, I noticed this.

sdk/core/azure_core_amqp/src/error.rs
LarryOsterman: Actually it was a part of an experiment for my tests trying to compare two ErrorKind enumerations.

LarryOsterman: So removing it :)

sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
LarryOsterman: Oh, that's interesting. I had assumed that it would generate dramatically more code because it had to create the closure which does nothing but return an error and that the `ok_or` variant would ultimately reduce the code size. 

I hadn't considered that the compiler would pre-calculate the error even if the value was None (which basically will never happen). 

Reverting.

sdk/eventhubs/azure_messaging_eventhubs/src/common/connection_manager.rs
LarryOsterman: I use trace for the lowest level information - in other words, the things that *I* as the developer of the code care about. *debug* is used for things that I think that the *user* of the library would care about.

So when I'm sticking in "I need to make sure I understand what's going on here", I use `trace`, when I'm sticking in "I bet someone would like to know about this thing", I use `debug`.

But I'm not completely consistent, to be honest.

sdk/eventhubs/azure_messaging_eventhubs/src/common/connection_manager.rs
```diff
@@ -0,0 +1,150 @@
...
+            session.begin(connection.as_ref(), None).await?;
+
+            let cbs = AmqpClaimsBasedSecurity::new(&session)?;
+            cbs.attach().await?;
+
+            debug!("Get Token.");
+            let token = credential
+                .get_token(&["https://eventhubs.azure.net/.default"])
+                .await?;
+            debug!("Got token: {:?}", token.token.secret());
```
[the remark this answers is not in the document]
LarryOsterman: I sincerely hope that the tokens passed to the service don't ever contain secrets, or secret equivalent values. If they do, we've utterly messed up our authentication story.

But your point is valid regardless.

sdk/core/azure_core_amqp/src/fe2o3/connection.rs
LarryOsterman: Unfortunately, It's used on line 116 (`builder = builder.hostname(url.host_str))`).

sdk/eventhubs/azure_messaging_eventhubs/src/common/connection_manager.rs
LarryOsterman: The Arc part of this change is forward looking to when this is a connection pool - the idea will be that you can borrow connections from the pool and return them when you're done with the individual API call. That's why the `get()` method returns an `Arc<AmqpConnection>` rather than a `&AmqpConnection`.

And the Mutex is there because `OnceLock::get_or_try_init()` is an experimental feature :(. So I need the mutex to protect the initialization of the OnceLock.

I could just as easily have used `Option<AmqpConnection>` here, to be honest.

sdk/eventhubs/azure_messaging_eventhubs/src/error.rs
LarryOsterman: I commented out a bunch of stuff, then removed it. And missed one of the "commented out" lines.

sdk/eventhubs/azure_messaging_eventhubs/tests/consumer.rs
LarryOsterman: This is test code.

For my tests, I use "info" as a shorthand for `println` - but it's controlled by the RUST_LOG environment variable.

heaths (approved):

PR #2246, opened 2025-02-26
[Storage] `ContainerClient` (`create_container`, `delete_container`) and `BlobClient` (`download_blob`, `upload_blob`)

vincenttran-msft:
As the title states, this pull-request introduces the initial implementation of the following:
`ContainerClient`

- `create_container`
- `delete_container`

`BlobClient`

- `download_blob`
- `upload_blob`

Note:

1. This pull-request is currently still based off [v0.9.1](https://github.com/Azure/azure-rest-api-specs/tree/vincenttran/blob-tsp-test2), as `v0.10.0` is under active development and introduces a huge refactor to the structure of the generated code (and will be handled accordingly in a later PR)
2. This PR is also our first attempt at linking our test into the test proxy infrastructure. Any feedback on the usage of the test utilities would be greatly appreciated!
3. There is a stopgap fix, which involves renaming the handwritten `BlobContainerClient` -> `ContainerClient`. This will be resolved in a future PR as it is due to a generated code issue that is resolved in a future iteration.


sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -1,2 +1,82 @@
...
 // Licensed under the MIT License.
+
+use crate::clients::GeneratedBlobClient;
+use crate::models::{BlobContainerClientCreateOptions, BlobContainerClientDeleteOptions};
+use crate::pipeline::StorageHeadersPolicy;
+use crate::BlobClientOptions;
+use azure_core::{
+    credentials::TokenCredential, BearerTokenCredentialPolicy, Policy, Response, Result, Url,
+};
+use std::sync::Arc;
```
jalauzon-msft, line 11: super nit: new line after imports?

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -58,3 +98,162 @@ async fn test_get_blob_properties_invalid_container() -> Result<(), Box<dyn Erro
...
+        .upload_blob(
+            RequestContent::from(data2.to_vec()),
+            true,
+            i64::try_from(data2.len())?,
+            None,
+        )
+        .await?;
+
+    // Assert
+    assert_eq!(response.status(), StatusCode::Created);
```
jalauzon-msft, line 255: You should check the content is indeed overwritten as expected.

Let's also test the error case (overwrite=false when there is a Blob). Can probably be in the same test right before we do overwrite=true

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -58,3 +98,162 @@ async fn test_get_blob_properties_invalid_container() -> Result<(), Box<dyn Erro
...
+    // Assert
+    let (status_code, headers, response_body) = response.deconstruct();
+    assert!(status_code.is_success());
+    assert_eq!(
+        "21",
+        headers.get_str(&HeaderName::from_static("content-length"))?
+    );
+    assert_eq!(Bytes::from_static(data), response_body.collect().await?);
+
+    container_client.delete_container(None).await?;
```
jalauzon-msft, line 153: Is there a way to ensure this runs eve if the test fails on an earlier line?

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -3,17 +3,66 @@
...
+// BEGIN GENERATED CODE -- do not edit from here till END
 mod generated;
 
-pub(crate) mod pipeline;
+pub mod clients {
+    pub mod blob_client;
+    pub mod blob_container_client;
+
+    pub use blob_client::BlobClient;
+    pub use blob_container_client::BlobContainerClient as ContainerClient;
```
vincenttran-msft, line 14: Ignore this rename-- This is currently a workaround for failing [Run source analysis](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4599779&view=logs&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=2102385d-609d-5572-64d2-932661c7902f&l=1053) on docs gen. This is because the docstring is referring to the generated code's client (but the names currently collide). This will be fixed as part of the move towards `v0.10.0`, and so this is the easiest way to circumvent this issue since this PR is pinned to the `v0.9.1` typespec/emitter combo.

In general, `lib.rs` will go under a huge renovation once the new generated code structure change comes through.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: Sounds good, I added in the error case, assert, and checking the content is in fact overwritten.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: That's a good point. I know we chatted offline about `try-finally` but in fact Rust does not have that control flow.

@heaths Heath, is there an idiomatic way to handle cleaning up test resources in both end of test and error cases? I couldn't find anything in my cursory glance through Cosmos and KeyVault tests.

One possible idea is making a struct to hold reference to all the clients necessary for cleanup, then implementing the [Drop](https://doc.rust-lang.org/rust-by-example/trait/drop.html) trait?

jalauzon-msft (approved):

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -58,3 +98,185 @@ async fn test_get_blob_properties_invalid_container() -> Result<(), Box<dyn Erro
...
+            i64::try_from(data.len())?,
+            None,
+        )
+        .await?;
+    let response = blob_client
+        .download_blob(Some(BlobBlobClientDownloadOptions::default()))
+        .await?;
+
+    // Assert
+    let (status_code, headers, response_body) = response.deconstruct();
```
vincenttran-msft, line 145: At this point, we have a `response_body` of type [ResponseBody](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/typespec/typespec_client_core/src/http/response.rs#L274), is there any examples on how to directly stream from this type? While this test currently does test functionality and verifies the content, I would like to be able to test the more common customer scenario (i.e. streaming directly from the download stream to a file or different workflow without buffering all into memory).

sdk/storage/azure_storage_blob/assets.json
heaths, line 1: Just to make you aware, this is the default location of `assets.json` if I create one from the test-proxy code; however, across most other languages you're probably using a single `assets.json` in the service directory i.e., `sdk/storage`. If that's the case and that's how you want to manage it - one set of assets for all related crates, which I imagine is what you want just like I did intentionally for Key Vault - you want to move this up a directory. It does not need any further changes, though. The content is fine.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -53,6 +61,14 @@ impl BlobClient {
         &self.endpoint
     }
 
+    pub fn container_name(&self) -> &String {
```
heaths, line 64: Never return a `&String`. You should return a string slice i.e., `&str`.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -66,4 +82,39 @@ impl BlobClient {
...
+            .await?;
+        Ok(response)
+    }
+
+    // For now, this is single-shot, block blob hot path only.
+    pub async fn upload_blob(
+        &self,
+        data: RequestContent<Bytes>,
+        overwrite: bool,
+        content_length: i64,
```
heaths, line 103: They can specify a negative content_length? Note that this isn't .NET that doesn't like unsigned types in public APIs. It's totally idiomatic - and quite frequent (probably more than signed types, in my experience) - to specify unsigned integers.

heaths: Remember: a common mantra of Rust is "if it compiles, it works". Being able to specify a negative numbers - which probably results in an HTTP 400, I imagine - goes against that principal.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
heaths: Please fix all instances throughout. I see another below.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -1,50 +1,90 @@
...
+        String::from("testcontainer1"),
+        recording.credential(),
+        Some(options.clone()),
+    )?;
+    container_client.create_container(None).await?;
+
     let blob_client = BlobClient::new(
         &endpoint,
-        String::from("testcontainer"),
+        String::from("testcontainer1"),
```
heaths, line 37: Hmm. I should probably add support to generate random names. In the meantime, though, I recommend you use names based on  your test name for easy identification.

sdk/storage/azure_storage_blob/tests/blob_client.rs
heaths: It's already a stream, and we're actually going to be changing that to buffer the entire stream within the pipeline, but will have a way - for these upload/download methods - to opt out of that. Keep in mind, though, like most (all, probably) other Azure SDK languages, that currently won't be part of the pipeline. I talked with several other architects and those languages don't retry streams like that but agree they should - probably a `RetriableStream` that wraps the pipeline.

As for an example, see https://github.com/Azure/azure-sdk-for-rust/blob/b39649ba194c49326811c5a7cad5ab7a54fa9adc/sdk/typespec/typespec_client_core/examples/stream_response.rs#L20-L27

sdk/storage/azure_storage_blob/tests/blob_client.rs
heaths: There are no exceptions in Rust, but you still control the flow. Don't use the `?` operator and instead capture the result, do something, then react to the result e.g.,

```rust
let result = response_body.collect().await;
// ignore error on next line
let _ = container_client.delete_container(None).await;
assert_eq!(Bytes::from_data(data), result?); // notice the questio mark here, or use `.unwrap()` and let the test panic / fail
```

It's more like Go or C/C++. Handle the errors when you want to.

sdk/storage/azure_storage_blob/tests/blob_container_client.rs
```diff
@@ -0,0 +1,34 @@
...
+        &endpoint,
+        String::from("testcontainer11"),
+        recording.credential(),
+        Some(options),
+    )?;
+
+    // Assert
+    container_client.create_container(None).await?;
+
+    container_client.delete_container(None).await?;
```
heaths, line 32: Why bother trying to clean up? I don't know why we ever did in .NET. Originally before I wrote the Test Resources scripts for all the lanugages, it was because we had long-lived test resources. No more. We provision as needed. That said, if you're running these again and again and trying to create an existing container fails, you could just decode to ignore *that* error. Otherwise, you're also just creating bigger session files that take longer to pull/push. Maybe have 1 test that tests creation and deletion, but ignore creation failures in the rest where it doesn't matter.

heaths (changes requested):

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: Good catch, this was definitely an oversight and I will change it to unsigned

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: I came up with this in a separate branch when testing out test proxy:
```rust
pub fn get_name(recording: &Recording, resource_type: &str) -> String {
    let rand: u128 = recording.random();
    format!("{}{}{}", "test", resource_type, Uuid::from_u128(rand))
}
```

But would be happy to just consume some common utility instead of writing a helper specific to Storage (since I could see this being beneficial to others as well)

sdk/storage/azure_storage_blob/tests/blob_container_client.rs
vincenttran-msft: Sure, we could decode and ignore that error. I will take that approach in the following revision.

sdk/storage/azure_storage_blob/assets.json
vincenttran-msft: Good callout here, I will move it up a directory. Thanks!

sdk/storage/azure_storage_blob/tests/blob_client.rs
heaths: #2265 is merged so you can use that now to generate seeded random names like in other languages.

vincenttran-msft: Awesome, thanks for the quick turnaround on that Heath! I will completely switch to this utility in the following PR when we revisit our tests + add new tests for the other APIs.

sdk/storage/azure_storage_blob/tests/blob_client.rs
heaths: I didn't see a change for this one.

heaths (approved):
LGTM, with one outstanding question you can resolve appropriately.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: Hi Heath, we are still having discussions internally with how we want to test resources, but will definitely be thinking on this in the next PR.

PR #2247, opened 2025-02-26
Update package versions and local path dependencies

hallipr:


Cargo.toml
```diff
@@ -29,41 +29,23 @@ rust-version = "1.80"
 
 [workspace.dependencies.typespec]
 default-features = false
-path = "sdk/typespec"
```
heaths, line 32: Is this what we decided to do? I thought we were going to leave both. This really needs to be written up somewhere and not locked in our heads. We've waffled about this too much.

That said, this is maybe premature because:

1. We should test `package-workspace` and help stabilize that. Seems it'll solve most of our problems and we won't have to have this difficult solution.
2. Until GA, we'll be rev'ing everything anyway. beta.2 is going to include a massive number of changes because I'm changing `azure_core` significantly.

heaths: If anything, maybe we should be keeping the path dependency for the foreseeable future.

sdk/core/azure_core/Cargo.toml
```diff
@@ -28,8 +28,11 @@ serde_json.workspace = true
 sha2 = { workspace = true, optional = true }
 tokio = { workspace = true, optional = true }
 tracing.workspace = true
-typespec = { workspace = true, features = ["http", "json"] }
-typespec_client_core = { workspace = true, features = ["http", "json"] }
+typespec = { path = "../../typespec", features = ["http", "json"] }
```
heaths, line 31: I think all crates should be using path dependencies until GA (or close). See previous comments.

heaths (changes requested):

eng/scripts/Get-PackageVersions.ps1
```diff
@@ -0,0 +1,51 @@
...
+  [switch]$Versions,
+  [Parameter(ParameterSetName = "dependencies")]
+  [switch]$Dependencies
+)
+
+$metadata = cargo metadata --format-version 1 --no-deps --all-features | ConvertFrom-Json -AsHashtable
+$packages = $metadata.packages
+foreach ($package in $packages) {
+  try {
+    $resp = Invoke-RestMEthod "https://crates.io/api/v1/crates/$($package.name)"
```
heaths, line 13: Nit:
```suggestion
    $resp = Invoke-RestMethod "https://crates.io/api/v1/crates/$($package.name)"
```
But the actual concern here is that the API is rate limited to 1req/s; however, `https://index.crates.io/az/ur/azure_core` is not and has the same info. We just need to handle the NDJSON e.g.,

```pwsh
$versions = (Invoke-RestMethod https://index.crates.io/az/ur/azure_core) -split "`n" | ConvertFrom-Json
# now sort by $versions.version using Wes's semver code
```

eng/scripts/Get-PackageVersions.ps1
```diff
@@ -0,0 +1,51 @@
...
+  $package.publish = $null -eq $package.publish
+
+  foreach ($dependency in $package.dependencies) {
+    $dependencyPackage = $packages | Where-Object -Property name -EQ -Value $dependency.name | Select-Object -First 1
+
+    if ($dependencyPackage) {
+      $ordered = [ordered]@{
+        dependant        = $package.name
+        pathVersion      = $dependencyPackage.version
+        released_version = $dependencyPackage.released_version
```
heaths, line 27: Nit: pick a casing strategy. ;)

heaths (approved):

eng/scripts/Get-PackageVersions.ps1
hallipr: Updated, thanks

sdk/core/azure_core_amqp/Cargo.toml
```diff
@@ -18,7 +18,7 @@ categories = ["api-bindings"]
 edition = "2021"
 
 [dependencies]
-azure_core = { path = "../azure_core", default-features = false }
+azure_core = { path = "../azure_core", version = "0.23.0", default-features = false }
```
heaths, line 21: Versions should only be in the workspace `Cargo.toml`. I don't want to manager versions across a swatch of files.

heaths (changes requested):

sdk/core/azure_core_amqp/Cargo.toml
heaths: But I thought we can put *both* the `path` and `version` in the workspace `Cargo.toml`. I do not want to keep track of all the versions in the crates' `Cargo.toml`s. Most other languages use centralized versions/dependencies as well. It's far easier to keep track of and resolve CG alerts, etc.

hallipr:
versions are mostly correct on `main` now.  Replacing this PR with #2307.

PR #2252, opened 2025-02-27
[cosmos] use version-only references to avoid broken publish rewriting

### LINKED ISSUE #2241: Can't find `key_auth` feature for v0.22.0 of Azure Cosmos DB crate
### Bug Title

Can't use key authentication due to an error message about a missing feature

### Crate Name

azure_data_cosmos

### Crate Version

0.22.0

### Description

I'm trying to use key authentication following the documentation, but I can't get it to build.

Here's the error I get when running `crate check` or `crate build`:

```
error: failed to select a version for `azure_data_cosmos`.
    ... required by package `azure-cosmos-db-nosql-rust-quickstart v0.0.0 (/workspaces/cosmos-db-nosql-rust-quickstart/src)`
versions that meet the requirements `^0.22.0` (locked to 0.22.0) are: 0.22.0

the package `azure-cosmos-db-nosql-rust-quickstart` depends on `azure_data_cosmos`, with features: `key_auth` but `azure_data_cosmos` does not have these features.
```

### Steps to Reproduce

1. Create a *\*.toml* file with atleast these dependencies:
    
    ```toml
    azure_data_cosmos = { version = "0.22.0", features = ["key_auth"] }
    azure_core = "0.22.0"
    ```

1. Add this sample code to an application:

    ```rust
    use azure_data_cosmos::{CosmosClient, PartitionKey, Query};
    use azure_core::credentials::Secret;

    ...
    
    let client = CosmosClient::with_key("https://myaccount.documents.azure.com/", Secret::from("my_key"), None).unwrap();
    
    ...
    ```

> [!NOTE]
> Just to isolate my test, I copied the exact sample verbatin from the `with_key` function in the source code: 
>
> <https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs#L73>
> 

### Checklist

analogrelay:
This is a somewhat-temporary workaround to fix #2241 so we can publish a version of the package with the `key_auth` feature.

The current publishing code rewrites any dependency with a `path` key (i.e. referencing a crate in the repository, like `azure_core`) so that it instead references the latest released (or to-be-released) version. This rewriting code is written in PowerShell and deserializes/reserializes Cargo.toml. Due to a bug in Tomlyn (https://github.com/xoofx/Tomlyn/issues/98), empty TOML arrays can be lost during this serialization process, which is what caused us to lose the `key_auth` feature.

The quick-fix here is to remove the `path` references and depend _solely_ on released versions of `azure_`/`typespec_` crates. As we continue, this may cause issues if we want to depend on features that are still being developed in-repo, but for now, this is just fine and matches user expectations.

My plan is to release this as a patch release (0.22.1) ASAP, if it works properly. I should be able to tell from the output of the Pack Crates step (which currently reports warnings due to the missing `key_auth` feature, which I'm looking at turning to errors in a separate PR).

### DIFF
--- a/eng/scripts/Language-Settings.ps1
+++ b/eng/scripts/Language-Settings.ps1
@@ -132,7 +132,7 @@ function Get-rust-AdditionalValidationPackagesFromPackageSet ($packagesWithChang
   [array]$additionalPackages = $affectedPackages | Where-Object { $packagesWithChanges -notcontains $_ }
 
   # if the change affected no packages, e.g. eng/common change, we use core and template for validation
-  if ($additionalPackages.Length -eq 0) {
+  if ($packagesWithChanges.Length -eq 0) {
     $additionalPackages += $allPackageProperties | Where-Object { $_.Name -eq "azure_core" -or $_.Name -eq "azure_template" }
   }
 
--- a/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
+++ b/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
@@ -7,7 +7,8 @@
 ### Breaking Changes
 
 ### Bugs Fixed
-* Fixed a publishing issue that caused the `key_auth` feature to be omitted. [#2241](https://github.com/Azure/azure-sdk-for-rust/issues/2241)
+
+* Fixed a publishing issue that caused the `key_auth` feature to be omitted. ([#2241](https://github.com/Azure/azure-sdk-for-rust/issues/2241))
 
 ### Other Changes
 
@@ -15,4 +16,4 @@
 
 ### Features Added
 
-- Initial supported release.
+* Initial supported release.
--- a/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
+++ b/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
@@ -1,12 +1,13 @@
 # Release History
 
-## 0.23.0 (Unreleased)
+## 0.22.1 (Unreleased)
 
 ### Features Added
 
 ### Breaking Changes
 
 ### Bugs Fixed
+* Fixed a publishing issue that caused the `key_auth` feature to be omitted. [#2241](https://github.com/Azure/azure-sdk-for-rust/issues/2241)
 
 ### Other Changes
 
--- a/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
+++ b/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
@@ -1,16 +1,10 @@
 # Release History
 
-## 0.22.1 (Unreleased)
-
-### Features Added
-
-### Breaking Changes
+## 0.22.1 (2025-03-05)
 
 ### Bugs Fixed
 
 * Fixed a publishing issue that caused the `key_auth` feature to be omitted. ([#2241](https://github.com/Azure/azure-sdk-for-rust/issues/2241))
-
-### Other Changes
 
 ## 0.22.0 (2025-02-25)
 
--- a/sdk/cosmos/azure_data_cosmos/Cargo.toml
+++ b/sdk/cosmos/azure_data_cosmos/Cargo.toml
@@ -25,7 +25,7 @@ url.workspace = true
 
 [dev-dependencies]
 azure_identity = "0.22.0"
-azure_core_test = "0.1.0"
+azure_core_test.workspace = true
 clap.workspace = true
 reqwest.workspace = true
 time.workspace = true
--- a/sdk/cosmos/azure_data_cosmos/Cargo.toml
+++ b/sdk/cosmos/azure_data_cosmos/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "azure_data_cosmos"
-version = "0.23.0"
+version = "0.22.1"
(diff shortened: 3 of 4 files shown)


sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -15,17 +15,17 @@ categories = ["api-bindings"]
...
 tracing.workspace = true
-typespec_client_core = { workspace = true, features = ["derive"] }
+typespec_client_core = { version = "0.1.0", features = ["derive"] }
 url.workspace = true
 
 [dev-dependencies]
-azure_identity.path = "../../identity/azure_identity"
-azure_core_test.path = "../../core/azure_core_test"
+azure_identity = "0.22.0"
+azure_core_test = "0.1.0"
```
hallipr, line 28: azure_core_test doesn't ship and should stay a dev-dependency

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -1,12 +1,13 @@
...
 
-## 0.23.0 (Unreleased)
+## 0.22.1 (Unreleased)
 
 ### Features Added
 
 ### Breaking Changes
 
 ### Bugs Fixed
+* Fixed a publishing issue that caused the `key_auth` feature to be omitted. [#2241](https://github.com/Azure/azure-sdk-for-rust/issues/2241)
```
heaths, line 10: Good idea to install markdown lint, which would've cause this:

```suggestion

* Fixed a publishing issue that caused the `key_auth` feature to be omitted. ([#2241](https://github.com/Azure/azure-sdk-for-rust/issues/2241))
```

Also nit: we tend to put bug references in parens, but not a big deal as long as you're consistent in your CHANGELOG.

sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -15,16 +15,16 @@ categories = ["api-bindings"]
 
 [dependencies]
 async-trait.workspace = true
-azure_core.workspace = true
+azure_core = "0.22.0"
```
heaths, line 18: This is the wrong way, but perhaps necessary for right now. /cc @hallipr. The workspace dependency should be a version, while local dependencies should be path until after `azure_core` or whatever dependency ships (`dev-dependencies` should always be path going forward, but we'll solve that later).

Managing versions across many different crates will be difficult as we onboard hundreds. I don't want to have to fix all those when we get CG alerts. Most other languages have centralized dependencies.

analogrelay: I'm fine promoting this to the workspace if that's what you'd prefer, now that we've released.

heaths: Might be a good test, though until we GA nearly everything will end up being path.

Hmm, that said, maybe this is fine for now since we'll just need to change it back anyway because of that. @hallipr what do you think?

analogrelay: Ok, why don't I leave this as specific to `azure_data_cosmos` for now and we can sort out our plan for workspace-wide dependencies separately. I'd really like to get this `.1` working and released because we're seeing a few customers notice the issues (or at least unintended side-effects, like the failed docs build).

eng/scripts/Language-Settings.ps1
```diff
@@ -132,7 +132,7 @@ function Get-rust-AdditionalValidationPackagesFromPackageSet ($packagesWithChang
   [array]$additionalPackages = $affectedPackages | Where-Object { $packagesWithChanges -notcontains $_ }
 
   # if the change affected no packages, e.g. eng/common change, we use core and template for validation
-  if ($additionalPackages.Length -eq 0) {
+  if ($packagesWithChanges.Length -eq 0) {
```
hallipr, line 135:
```suggestion
  if ($additionalPackages.Length -eq 0 -and $packagesWithChanges.Length -eq 0) {
```

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -1,17 +1,19 @@
 # Release History
 
-## 0.23.0 (Unreleased)
+## 0.22.1 (Unreleased)
```
heaths, line 3: Probably worth just updating the CHANGELOG.md for the impending release here now.

heaths (approved):

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
analogrelay: Done

eng/scripts/Language-Settings.ps1
hallipr: We're wanting to express:
- $packagesWithChanges will already contain packages where $diff contains a file in the package directory
- $additionalPackages will contain any packages where there was a change in its service directory, but not in a package directory.  e.g.     /sdk/cosmos/some-file.json
- If we have no auto-discovered packages, use _azure_core and azure_template

Pilchie (approved):

PR #2253, opened 2025-02-27
deny warnings while packing crates

analogrelay:
This is currently just a test to see if `RUSTFLAGS=-D warnings` would have found the issue that caused #2241 . If it does fail the build correctly, I'll check in as to where we want to add it. Denying warnings seems like a good plan. It can be a big pain to a local developer though, so my thinking is we'd only deny warnings in CI.


heaths:
Wait, do we not have that in our pipelines as an env var currently? I didn't realize and we absolutely should. /cc @hallipr

hallipr:
> Wait, do we not have that in our pipelines as an env var currently? I didn't realize and we absolutely should. /cc @hallipr

We have it set in Analyze and Test scripts:
https://github.com/Azure/azure-sdk-for-rust/blob/e1cd9d26261bf031c05d1d425eda252df152ae2c/eng/scripts/Analyze-Code.ps1#L19

https://github.com/Azure/azure-sdk-for-rust/blob/e1cd9d26261bf031c05d1d425eda252df152ae2c/eng/scripts/Test-Packages.ps1#L38

hallipr:
@analogrelay, to test the failure, you'll need to update a file in cosmos so it packs that package

heaths:
> > Wait, do we not have that in our pipelines as an env var currently? I didn't realize and we absolutely should. /cc @hallipr
> 
> We have it set in Analyze and Test scripts:
> 
> https://github.com/Azure/azure-sdk-for-rust/blob/e1cd9d26261bf031c05d1d425eda252df152ae2c/eng/scripts/Analyze-Code.ps1#L19-L20
> 
> https://github.com/Azure/azure-sdk-for-rust/blob/e1cd9d26261bf031c05d1d425eda252df152ae2c/eng/scripts/Test-Packages.ps1#L38

It's better to set it in the workflows for the whole thing, like is common in rust: https://github.com/heaths/akv-cli-rs/blob/59e56b2827f7e5a1efe5f0283cd2d06d7b9312de/.github/workflows/ci.yml#L20-L22

Notice how that is global. We should do the same here for Azure Pipelines. I don't want the env var squirreled away in some script that only affects that script, especially since each `pwsh` task we run creates a new `pwsh` process and that env var is lost after the process exits.

analogrelay:
This was successful and did flag the unused feature as an error, which would have prevented #2241 . I can look at moving it to the pipelines config.

heaths:
I already have a PR for it: https://github.com/Azure/azure-sdk-for-rust/pull/2260

analogrelay:
> I already have a PR for it: https://github.com/Azure/azure-sdk-for-rust/pull/2260

😆 Fair enough. This should be closable then, I'll confirm your PR covers my case and do so.

analogrelay:
Superceded by #2260

PR #2260, opened 2025-02-28
Always set RUSTFLAGS=-Dwarnings

heaths:
Also removes some old scripts that no longer apply (for old legacy directory structure).


analogrelay (approved):

heaths:
Thanks. I was just checking in and noticed that too. I was thinking of GitHub Actions, I admit. I'm going to template them similar to what you did.

heaths:
Seems, unlike GitHub Actions, you can't define env vars globally: https://learn.microsoft.com/azure/devops/pipelines/yaml-schema/variables?view=azure-pipelines. Or, at least, you can on the pipeline but it seems the root pipeline would mean all the service directories' individual `ci.yml` and `pullrequest.yml` files, which isn't ideal.

eng/pipelines/templates/variables/rust.yml
```diff
@@ -0,0 +1,3 @@
+variables:
```
hallipr, line 1: Why not just add these to globals.yml?

hallipr (approved):

hallipr:
> Seems, unlike GitHub Actions, you can't define env vars globally

For java, I used a syntax in 1es-redirect that let us pass "pipeline" variables into the template and also let us inject global variables into all pipelines:

```yaml
        variables:
          - template: /eng/pipelines/variables/image.yml@self
          - ${{ each variable in parameters.variables }}:
              - ${{ if variable.key }}:
                  # turn objects into name/value lists
                  - name: ${{ variable.key }}
                    value: ${{ variable.value }}
              - ${{ else }}:
                  # preserve all other variables
                  - ${{ variable }}
          - ${{ each variable in stage.variables }}:
              - ${{ if variable.key }}:
                  # turn objects into name/value lists
                  - name: ${{ variable.key }}
                    value: ${{ variable.value }}
              - ${{ else }}:
                  # preserve all other variables
                  - ${{ variable }}
```

So every passed in stage gets a variables section that includes "image.yml", any variables passed in through "parameters.variables" as well as any variables set directly on the stage

eng/pipelines/templates/variables/rust.yml
heaths: I thought about it, but wasn't entirely sure it was imported everywhere and didn't want to worry about the ramifications of doing so. If you think it's safer, I'm fine doing that. Note for example, though, it's not in https://github.com/Azure/azure-sdk-for-rust/pull/2260/files#diff-8846d7741afb575539571d7b46758247229daaa4336751c6b3e743fd284426fa. I imagine that's probably why @analogrelay did it that way as well. `global.yml` didin't seem truly global.

heaths:
> So every passed in stage gets a variables section that includes "image.yml", any variables passed in through "parameters.variables" as well as any variables set directly on the stage

Interesting. But, personally, I'd rather keep it simple. Following our chain of pipeline YAML definitions from service directories back to eng/common/pipelines out to eng/pipelines and back into eng/common/pipelines (repeat a few more times) is already complicated enough. Those of us who deal with it often enough get it, but it's makes maintenance harder and I'd love to avoid that when possible. I appreciate it's benefits, but a simpler syntax in this case feels better.

PR #2265, opened 2025-03-01
Generate seeded random test strings

heaths:


.vscode/cspell.json
```diff
@@ -17,14 +17,15 @@
     ".vscode/tasks.json",
     "NOTICE.txt",
     "eng/",
-    "*.dict.txt",
+    "**/.dict.txt",
```
heaths, line 20: @LarryOsterman I thought you checked this change in a while ago. Did it not get merged yet?

sdk/core/azure_core_test/src/recording.rs
```diff
@@ -158,52 +177,68 @@ impl Recording {
...
+    /// if `LEN` is 0,
+    /// or if the length of `prefix` is greater than or equal to `LEN`.
+    ///
+    /// ```should_panic
+    /// # let recording = azure_core_test::Recording::with_seed();
+    /// let vault_name = recording.random_string::<8>(Some("keyvault"));
+    /// ```
+    ///
+    pub fn random_string<const LEN: usize>(&self, prefix: Option<&str>) -> String {
+        struct NonZero<const N: usize>;
```
heaths, line 217: I do this instead of a `NonZeroUsize` because you can't simply pass `0` as the argument. You'd have to do `NonZeroUsize::new(0)` and that's not a great DX.

heaths: I was also trying to catch as many errors at compile time as I could. Alternatively, I could go back to the original signature I prototyped i.e., `random_string(&self, prefix: Option<&str>, len: usize) -> String` but then all the assertions are runtime assertions.

.vscode/cspell.json
LarryOsterman: It should have been merged with my changes on Friday.

LarryOsterman: https://github.com/Azure/azure-sdk-for-rust/commit/11ae4ec2876f4a3bba50e71e3cd46bbcfa1c39d1

LarryOsterman (approved):

PR #2273, opened 2025-03-04
[Storage] `stage_block`, `commit_block_list` for `BlobClient`, `get_container_properties` for `ContainerClient`, `ServiceClient`

vincenttran-msft:

BlobClient
- `stage_block`
- `commit_block_list`

ContainerClient
- `get_container_properties`

ServiceClient
- Initial introduction of `ServiceClient`
- `get_service_properties`

This PR also moves to the standardized way to fetch randomized seeded names (`random_string`) and includes a helper utility for test setup.


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -117,4 +118,47 @@ impl BlobClient {
...
+    //     list_type: BlockListType,
+    //     options: Option<BlobBlockBlobClientGetBlockListOptions<'_>>,
+    // ) -> Result<Response<BlockList>> {
+    //     let response = self
+    //         .client
+    //         .get_blob_block_blob_client(self.container_name.clone(), self.blob_name.clone())
+    //         .get_block_list(list_type, options)
+    //         .await?;
+    //     Ok(response)
+    // }
```
vincenttran-msft, line 147: This was fully working E2E against a forked copy of the generated code that contained the necessary fixes. This will be fixed when we can regenerate and get the necessary generated code changes (emitter currently is broken by newest `.tsp` definition).

Same comment applies to `get_service_properties` for `BlobServiceClient`.

sdk/storage/azure_storage_blob/src/models/container_properties.rs
```diff
@@ -0,0 +1,91 @@
...
+    type Error = Error;
+    fn header_names() -> &'static [&'static str] {
+        &[
+            "etag",
+            "last-modified",
+            "x-ms-lease-state",
+            "x-ms-lease-status",
+            "x-ms-immutable-storage-with-versioning-enabled",
+            "x-ms-has-immutability-policy",
+            "x-ms-version",
```
jalauzon-msft, line 43: Version is not a container property, we should remove this and all its parsing logic.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -69,20 +73,22 @@ async fn test_get_blob_properties_invalid_container(
...
+        .random_string::<12>(Some("blob"))
+        .to_ascii_lowercase();
 
-    // Setup
-    let endpoint = format!(
-        "https://{}.blob.core.windows.net/",
-        recording.var("AZURE_STORAGE_ACCOUNT_NAME", None).as_str()
-    );
+    println!("{}", container_name.clone());
+    println!("{}", blob_name.clone());
```
jalauzon-msft, line 85: Probably remove these?

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -158,28 +163,27 @@ async fn test_download_blob(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
         recording.credential(),
         Some(options.clone()),
     )?;
     container_client.create_container(None).await?;
 
     let blob_client = BlobClient::new(
         &endpoint,
-        String::from("testcontainer3"),
-        String::from("test_upload_blob.txt"),
+        container_name.clone(),
```
jalauzon-msft, line 185: nit: technically don't need this clone since it's the last use. Sort of a weird pattern to clone all but the last one but that's what you did in other tests so maybe be consistent. Or clone everywhere.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -280,3 +283,210 @@ async fn test_upload_blob_overwrite(ctx: TestContext) -> Result<(), Box<dyn Erro
...
+            RequestContent::from(block_3.to_vec()),
+            Some(BlobBlockBlobClientStageBlockOptions::default()),
+        )
+        .await?;
+
+    let latest_blocks: Vec<String> = vec![
+        base64::encode("1"),
+        base64::encode("2"),
+        base64::encode("3"),
+    ];
```
jalauzon-msft, line 350: Hmmm, I'm not sure we should require the user to base 64 encode their block list, especially since we don't require it for `stage_block`. I'm not too sure the best way to handle it but we should do the encoding ourselves. What do we do in Python?

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -280,3 +283,210 @@ async fn test_upload_blob_overwrite(ctx: TestContext) -> Result<(), Box<dyn Erro
...
+    let block_1 = b"AAA";
+    let block_2 = b"BBB";
+    let block_3 = b"CCC";
+
+    blob_client
+        .stage_block(
+            "1",
+            i64::try_from(block_1.len())?,
+            RequestContent::from(block_1.to_vec()),
+            Some(BlobBlockBlobClientStageBlockOptions::default()),
```
jalauzon-msft, line 325: nit: Is there any difference between passing a default options bag and `None`? Should we just pass `None`? This applies a lot in this file.

Also testing passing default is good though so idk...

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -280,3 +283,210 @@ async fn test_upload_blob_overwrite(ctx: TestContext) -> Result<(), Box<dyn Erro
...
+    assert_eq!(
+        Bytes::from_static(b"AAABBBCCC"),
+        response_body.collect().await?
+    );
+
+    container_client.delete_container(None).await?;
+    Ok(())
+}
+
+// Currently blocked by generated code, uncomment when we can consume newest definition
```
jalauzon-msft, line 386: This definitely feels like an instance where we can test `get_block_list` (when available) inside the same test above. Maybe rename it to `test_block_list` and put everything in one?

sdk/storage/azure_storage_blob/tests/blob_container_client.rs
```diff
@@ -32,3 +34,68 @@ async fn test_create_container(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
+        recording.credential(),
+        Some(options),
+    )?;
+    container_client.create_container(None).await?;
+    let response = container_client
+        .get_container_properties(Some(BlobContainerClientGetPropertiesOptions::default()))
+        .await;
+
+    // Assert
+    assert!(response.is_ok());
```
jalauzon-msft, line 60: I think it would be fine to leave this out and move the `?` below to line 55. If the request failed, it test will fail because of the `?` which is fine.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: Here is the relevant excerpt from Python:
```python
        blob.stage_block('1', b'AAA')
        blob.stage_block('2', b'BBB')
        blob.stage_block('3', b'CCC')

        # Act
        block_list = [BlobBlock(block_id='1'), BlobBlock(block_id='2'), BlobBlock(block_id='3')]
        put_block_list_resp = blob.commit_block_list(block_list)
```

Interestingly enough, I don't really see where the `block_id` gets `base64` encoded in [Python ](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/azure/storage/blob/_models.py#L765)(other than the workaround found in `_from_generated`) but this definitely raises errors in Rust if you don't encode.

Sample Error:
```
Error: Error { context: Full(Custom { kind: HttpResponse { status: BadRequest, error_code: Some("InvalidBlockId") }, error: HttpError { status: BadRequest, details: ErrorDetails { code: Some("InvalidBlockId"), message: None }, body: <?xml version="1.0" encoding="utf-8"?><Error><Code>InvalidBlockId<</Code><Message>Block ID is invalid. Block ID must be base64 encoded.
```

I guess this would be a good candidate for a handwritten model? But the tricky part is that in Rust I am building the `Vec<String>` to pass it to `BlockLookupList` (auto-generated model that is the input type for the generated API):
```rust
pub struct BlockLookupList {
    /// The committed blocks
    #[serde(rename = "Committed", skip_serializing_if = "Option::is_none")]
    pub committed: Option<Vec<String>>,

    /// The latest blocks
    #[serde(rename = "Latest", skip_serializing_if = "Option::is_none")]
    pub latest: Option<Vec<String>>,

    /// The uncommitted blocks
    #[serde(rename = "Uncommitted", skip_serializing_if = "Option::is_none")]
    pub uncommitted: Option<Vec<String>>,
}
```

So it seems like mostly the model would just be backed by `Vec<String>` but `base64` encode the ID for the customer, and then we would have to do an `into()` to get it back to a `Vec<String>` for `BlockLookupList`. Happy to hear out any other thoughts or approaches!

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: Yeah this is a tricky one, Heath had mentioned in a previous PR that we really should be urging towards `None` when exposing samples and usage examples to cx, but for testing I think we could go either way (and see value in both).

We could establish a pattern here to keep it consistent, but I don't feel strongly towards either end (default vs. None)

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: Sounds good, I'll just remove both the def of the method & tests and then lump it in once we get past this PR.

sdk/storage/azure_storage_blob_test/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
```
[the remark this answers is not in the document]
vincenttran-msft: Hmm, I was following the convention that:

- azure_core_test exists for test utilities for client libraries built on `azure_core`
- azure_security_keyvault_test exists for test utilities for Key Vault tests

I am happy to move these into the main `azure_storage_blob` crate, but we should check with Azure SDK what is idiomatic here.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: Removed from this PR to avoid confusion. Will introduce when the generated code backing is updated.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: EDIT: I just did a rough draft of what this could look like.
Blob Block impl:
```rust
impl BlobBlock {
    pub fn new(input: Vec<String>) -> Self {
        let encoded_data = input.into_iter().map(base64::encode).collect();
        BlobBlock { data: encoded_data }
    }
}

impl From<BlobBlock> for Vec<String> {
    fn from(blob: BlobBlock) -> Self {
        blob.data
    }
}
```

Usage:
```rust
let latest_blocks = BlobBlock::new(vec![
        String::from("1"),
        String::from("2"),
        String::from("3"),
    ]);

    let block_lookup_list = BlockLookupList {
        committed: None,
        latest: Some(latest_blocks.into()),
        uncommitted: None,
    };
```

There is definitely still some convenience we could add i.e. taking in `&str` instead of `String` and doing that conversion on the model (to avoid cx having to do `String::from` or `.into_string()` on their `&str`s, but this just a quick demo of what I was envisioning.

sdk/storage/azure_storage_blob_test/Cargo.toml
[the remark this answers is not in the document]
jalauzon-msft: IMO this should probably be a helper sub-module in the test module of the SDK itself. The core test package provides all the stuff for other packages so that makes sense and keyvault has some actual tests in there so I don't think plain utilities have a precedent here.

Curious on Heath's thoughts though.

sdk/storage/azure_storage_blob/tests/blob_client.rs
jalauzon-msft: Let's use `None` in all cases except when testing options bags where we can include a "default" test.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: Make this take in `Vec<u8>` and do the `String` conversion and return that

heaths (changes requested):

sdk/storage/azure_storage_blob/Cargo.toml
```diff
@@ -25,6 +24,6 @@ workspace = true
 
 [dev-dependencies]
 azure_core_test.path = "../../core/azure_core_test"
-azure_identity.workspace = true
+azure_identity = "0.22.0"
```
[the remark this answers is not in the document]
vincenttran-msft: Hi Heath, this has been corrected in the current revision, this is an outdated revision you are commenting on. Thanks!

Cargo.toml
```diff
@@ -64,6 +64,10 @@ path = "sdk/identity/azure_identity"
 version = "0.1.0"
 path = "sdk/storage"
 
+[workspace.dependencies.azure_storage_blob]
+version = "0.1.0"
+path = "sdk/storage/azure_storage_blob"
+
```
heaths, line 70: You should only add this for other crates outside your service directory.

heaths: Within your service directory, just use a `path` dependency like I did for `sdk/keyvault`.

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -9,9 +9,11 @@ mod generated;
 pub mod clients {
     pub mod blob_client;
     pub mod blob_container_client;
+    pub mod blob_service_client;
```
heaths, line 12: @jhendrixMSFT why are these still `pub mod`? We want to flatten in `clients`.

sdk/storage/azure_storage_blob/src/models/container_properties.rs
```diff
@@ -0,0 +1,86 @@
...
+
+pub const LAST_MODIFIED: HeaderName = HeaderName::from_static("last-modified");
+pub const IMMUTABLE_STORAGE_WITH_VERSIONING_ENABLED: HeaderName =
+    HeaderName::from_static("x-ms-immutable-storage-with-versioning-enabled");
+
+/// Properties of an Azure Storage container.
+///
+#[derive(Clone, Default, SafeDebug)]
+pub struct ContainerProperties {
+    pub last_modified: Option<String>,
```
heaths, line 23: Why is this a string? We have date-time parsers for ISO 8601, RFC 3339, and one other RFC I forget off hand. This is not type-safe.

vincenttran-msft: Hi Heath, this hand-written model is largely cribbed from the implementation of [blob_properties.rs](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/storage/azure_storage_blob/src/models/blob_properties.rs#L26), which also represents things i.e. `last_modified` as `Option<String>`. This is because these are stop-gap implementations since this PR consumes the `v0.9` emitter version (which does _not_ provide the serialized response header). 

This will completely change and we will no longer utilize the handwritten model once this PR is merged and we can look to start introducing the v11+ emitter changes.

heaths: As you said: it's hand-written. Why not do it right the first time? It's straight forward because we have the helpers that do all the heavy lifting.

jhendrixMSFT: To clarify, this was due to bad authoring in the tsp. It's been fixed in https://github.com/Azure/azure-rest-api-specs/commit/1530d964ce2489aae004030bead632ffbfc4ec28.

heaths:
You have some clippy lints to resolve as well. Look at the failed build. It's a good idea to run this locally e.g., `cargo clippy -p azure_storage_blob` so you can quickly iterate on lints.

sdk/storage/azure_storage_blob/src/lib.rs
vincenttran-msft: These are all added by me (L10-L12) and it is possible these `lib.rs` generation issues have been resolved in greater versions. 

The reason why it is added under the `BEGIN GENERATED CODE -- do not edit from here till END` header is because I need to be inside of the `clients` mod in order to allow the import for our handwritten clients to look like:
```rust
use azure_storage_blob::{
    clients::{BlobClient, ContainerClient},
```

Please correct me if I'm wrong if there is another way to achieve this sort of import for customers.

Note: The `blob_client`, `blob_container_client`, and `blob_service_client` referenced in L10 -> L12 are the _handwritten_ clients, **not** generated clients.

I wouldn't consider this blocking for this PR, but we need to settle on `lib.rs` before release. However, `lib.rs` is still very changeable due to the changing nature of the generated code, and this PR is still pinned to an older version (v9)

vincenttran-msft:
> You have some clippy lints to resolve as well. Look at the failed build. It's a good idea to run this locally e.g., `cargo clippy -p azure_storage_blob` so you can quickly iterate on lints.

Interesting, this doesn't manifest locally:
```
C:\azure-sdk-for-rust\sdk\storage>cargo clippy -p azure_storage_blob
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.75s
```

I also largely took that `credential` type hint that is problematic from Cosmos [ex](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs#L31), so I am surprised Cosmos does not have the same error regarding [redundant explicit links](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4641471&view=logs&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=2102385d-609d-5572-64d2-932661c7902f&l=808).

Regardless, I removed the duplicate explicit link and re-ran `clippy` locally (although I can't really trust the local output, as it didn't flag it even _before_ the fix).

sdk/storage/azure_storage_blob/src/lib.rs
heaths: 1. Even if you had to modify this code (you don't; see below), the modules shouldn't be public:
   ```rust
   pub mod clients {
       mod blob_client;
       pub use block_client::*;
   }
2. While you can't define two `pub mod clients` in the `.rs` file, true, you can add a folder named `clients` with a `blob_client.rs` file/module therein that you should then be able to `mod blob_client; pub use blob_client::*` outside the "DO NOT EDIT" sction.

   These are functionality equivalent:

```
lib.rs // "mod foo"
foo/
  mod.rs // "mod bar"
  bar.rs
```

```
lib.rs // "mod foo"
foo.rs // "mod bar"
foo/
  bar.rs
```

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -17,6 +17,7 @@ use azure_core::{
 };
 use std::sync::Arc;
 
+// A client to interact with a specific Azure storage blob, although that blob may not yet exist.
```
heaths, line 20: Triple slashes for a proper doc comment.
```suggestion
/// A client to interact with a specific Azure storage blob, although that blob may not yet exist.
```

heaths (changes requested):
I don't see the `lib.rs` change. We don't want to be exposing all those submodules to clients.

sdk/storage/azure_storage_blob/src/lib.rs
vincenttran-msft: Currently, this is the structure of our project:
```
azure_storage_blob
	src/
		clients/
			blob_client.rs 
			blob_container_client.rs
			blob_service_client.rs
                generated/				
		...
		...
```		
So, taking your suggestion to:

1. Make the module non-public
2. Move them out of the "DO NOT EDIT" section

And the resulting `lib.rs` will roughly represent:
```rust
// END GENERATED CODE

mod blob_client;
mod blob_container_client;
mod blob_service_client;

pub use blob_container_client::BlobContainerClient as ContainerClient;
pub use blob_service_client::BlobServiceClient as ServiceClient;
```


Yields the following error:
>error[E0583]: file not found for module `blob_client`
  --> sdk\storage\azure_storage_blob\src\lib.rs:65:1
   |
65 | mod blob_client;
   | ^^^^^^^^^^^^^^^^
   |
   = help: to create the module `blob_client`, create file "sdk\storage\azure_storage_blob\src\blob_client.rs" or "sdk\storage\azure_storage_blob\src\blob_client\mod.rs"
   = note: if there is a `mod blob_client` elsewhere in the crate already, import it with `use crate::...` instead
   
Therefore, it is expecting the implementation of blob_client to live directly in `src` i.e. at src/blob_client.rs , whereas we have the project structure src/**clients**/blob_client.rs (the intermediary folder clients)

Hence why I was under the impression it needs to live in the `mod clients` defined in the "DO NOT EDIT" section. Thanks!

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: Good catch, bad copy and paste of a double slash for all the handwritten clients. Will fix in next commit, but pending your response wrt `lib.rs`

heaths:
I was trying to check out your branch when I realized you're not working in a fork. Please use forks on GitHub: fork your repo, push your branch to your fork, and open a PR normally. This not only yields in faster pull times, but also makes other workflows easier and is the GitHub way. Using a single shared repo is how Azure Repos operates but isn't the norm.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
heaths: I fixed this.

sdk/storage/azure_storage_blob/src/lib.rs
heaths: I fixed this and pushed.

heaths (approved):

PR #2274, opened 2025-03-04
Trace stdout from test-proxy

heaths:
This is a big refactor of the test-proxy wrapper that will precede other work. This will help diagnose issues that, in the past,  you could only do if you started the test-proxy separately and remembered to pass `PROXY_MANUAL_START=true` when testing.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "1.80.0"
+channel = "stable"
 components = [
   "rustc",
   "rustfmt",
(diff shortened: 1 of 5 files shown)


heaths:
I've been wanting to do this for a while because it would've helped a couple weeks ago trying to figure out why playback wasn't working in certain circumstances. It was only once I ran the `test-proxy` separate and watched its logs I could tell; however, interspersing stdout from test-proxy - which `cargo test` would swallow anyway - would make things more difficult. So, instead, I capture its stdout - which I already was to only get the version and port binding - and trace it all!

image

I also added a troubleshooting section to the `azure_core_test` `README.md` that explains a bit. Basically, by tracing this information it provides a lot of opportunity to customize what level of information - and from where - you want to see. For example, `hyper` writes a lot of trace information you probably rarely want to see, if ever...so don't!

LarryOsterman (approved):

PR #2278, opened 2025-03-04
Use `cargo read-manifest` to get rust_version from core

hallipr:
Because we have workspace path dependencies and direct version dependencies on azure_core in the repo, cargo metadata returns multiple packages named azure_core. We could use `--no-deps` to stop returning dependency packages, but then we still have to filter the results to `azure_core`.   `cargo read-manifest` is already filtered to the target package and seems like a more direct way to get at specific package properties.


analogrelay (approved):
This works in my testing. I'll rebase my other PR on this when it lands :).

Thanks for the quick fix!

heaths (approved):

PR #2284, opened 2025-03-05
Support custom ImdsId for AppServiceManagedIdentityCredential.

tot0:
This matches the interface of `VirtualMachineManagedIdentityCredential`, it enables specifying a `client_id` to use for Identities being served _not_ via the VM local Imds endpoint.

### DIFF
--- a/sdk/identity/azure_identity/src/credentials/app_service_managed_identity_credential.rs
+++ b/sdk/identity/azure_identity/src/credentials/app_service_managed_identity_credential.rs
@@ -19,7 +19,10 @@ pub struct AppServiceManagedIdentityCredential {
 }
 
 impl AppServiceManagedIdentityCredential {
-    pub fn new(options: impl Into<TokenCredentialOptions>) -> azure_core::Result<Arc<Self>> {
+    pub fn new(
+        id: ImdsId,
+        options: impl Into<TokenCredentialOptions>,
+    ) -> azure_core::Result<Arc<Self>> {
         let options = options.into();
         let env = options.env();
         let endpoint = &env
@@ -43,7 +46,7 @@ impl AppServiceManagedIdentityCredential {
                 API_VERSION,
                 SECRET_HEADER,
                 SECRET_ENV,
-                ImdsId::SystemAssigned,
+                id,
             ),
         }))
     }
--- a/sdk/identity/azure_identity/CHANGELOG.md
+++ b/sdk/identity/azure_identity/CHANGELOG.md
@@ -9,6 +9,7 @@
   The `tokio` feature is disabled by default so `std::process::Command` is used; otherwise, if enabled, `tokio::process::Command` is used.
   Callers can also implement the trait themselves to use a different asynchronous runtime.
 - Added `ImdsId` param to `AppServiceManagedIdentityCredential::new`
+- Expose `azure_identity::credentials::TokenCache` for external `TokenCredential` implementations to leverage
 
 ### Breaking Changes
 
--- a/sdk/identity/azure_identity/CHANGELOG.md
+++ b/sdk/identity/azure_identity/CHANGELOG.md
@@ -4,20 +4,20 @@
 
 ### Features Added
 
-- Added `AzurePipelinesCredential`.
-- `AzureCliCredentialOptions` (new) accepts a `azure_core::process::Executor` to run the Azure CLI asynchronously.
-  The `tokio` feature is disabled by default so `std::process::Command` is used; otherwise, if enabled, `tokio::process::Command` is used.
-  Callers can also implement the trait themselves to use a different asynchronous runtime.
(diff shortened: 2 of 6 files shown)


sdk/identity/azure_identity/src/credentials/app_service_managed_identity_credential.rs
```diff
@@ -19,7 +19,10 @@ pub struct AppServiceManagedIdentityCredential {
 }
 
 impl AppServiceManagedIdentityCredential {
-    pub fn new(options: impl Into<TokenCredentialOptions>) -> azure_core::Result<Arc<Self>> {
```
heaths, line 22: @chlowell @scottaddie should this type even be exposed publicly? I'm close to refactoring `azure_identity` and I remember talking about how only `ManagedIdentityCredential` should be exposed and all the "implementation-specific" credentials should be internal. That would significantly affect the need for this PR and force consideration for how else to pass the IMDS ID. Comparing with [Go's azidentity](https://github.com/Azure/azure-sdk-for-go/blob/91f14e2ae0e4f44540ba8df7ce889fc301e5b6ce/sdk/azidentity/managed_identity_client.go#L53-L60), I see it's just an implementation detail. I'm guessing this equates to [`ManagedIdentityCredentialOptions.ManagedIDKind`](https://github.com/Azure/azure-sdk-for-go/blob/dc2256feabc94ed0fef6a61325b10a55128774b6/sdk/azidentity/managed_identity_credential.go#L97)? I'm not sure, though, since the only implementations I see are all user-assigned variants.

heaths (changes requested):
Thanks for the contribution, but I don't think we can expose much of this publicly and probably already have too much public for the next beta release where we'll better align with other languages' versions of `azure_identity`. That's not to say we can't have some idiomatic rust-specifics, but it seems the `AppServiceManagedIdentityCredential` shouldn't even be public.

For that alone I'd still merge this, but then the whole reason for exposing the IMDS type evaporates. We also won't expose the cache. It's not like people can't write their own if needed. We don't want to expose types unless necessary or highly requested becuase we'll have to support them as-is in perpetuity. Instead, we will probably have a more robust cache that can be securely persisted a la <https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/cache/cache.go>. So we wouldn't expose this simple in-memory cache as-is.

tot0:
> Thanks for the contribution, but I don't think we can expose much of this publicly and probably already have too much public for the next beta release where we'll better align with other languages' versions of `azure_identity`. That's not to say we can't have some idiomatic rust-specifics, but it seems the `AppServiceManagedIdentityCredential` shouldn't even be public.
> 
> For that alone I'd still merge this, but then the whole reason for exposing the IMDS type evaporates. We also won't expose the cache. It's not like people can't write their own if needed. We don't want to expose types unless necessary or highly requested becuase we'll have to support them as-is in perpetuity. Instead, we will probably have a more robust cache that can be securely persisted a la https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/cache/cache.go. So we wouldn't expose this simple in-memory cache as-is.

Understandable on both fronts.
If I understand correctly, if I remove the TokenCache visibility changes you would accept the AppService changes? The breaking isn't Ideal, but I just want to reuse the `IDENTITY_ENDPOINT` and Secret logic, while being able to specify a ClientId to pass.

sdk/identity/azure_identity/src/credentials/app_service_managed_identity_credential.rs
tot0: A unified `ManagedIdentityCredential` credential makes sense to me. When to call the IMDS static IP endpiont and when to call a environment specific endpoint I guess switches on the presence of `IDENTITY_ENDPOINT` in env?

For any Managed Identity credential it is important to be _somehow_ able to specify whether to use a specific ClientId or just use the endpoints default.

heaths:
Possibly. We're discussing in the Identity v-team. But I can say you won't be able to use these changes for long, and even then would have to use a patched git reference to `azure_identity`. By the time the next beta releases, `azure_identity` won't expose all these other MSI credential types - only `ManagedIdentityCredential`.

tot0:
> Possibly. We're discussing in the Identity v-team. But I can say you won't be able to use these changes for long, and even then would have to use a patched git reference to `azure_identity`. By the time the next beta releases, `azure_identity` won't expose all these other MSI credential types - only `ManagedIdentityCredential`.

Happy to not actually PR this then. For now I am referencing my fork for usage of `azure_core`, `azure_identity` and `azure_messaging_eventhub`. I can keep it that way for a month or so if there's a refactor coming soob.

chlowell:
Closing this because #2409 replaced AppServiceManagedIdentityCredential with ManagedIdentityCredential, which accepts user-assigned IDs for all platforms. Note though that it supports only client and object IDs for App Service (#2407)

PR #2287, opened 2025-03-05
[Storage] Initial `v11` codegen output

vincenttran-msft:
Compiling, but testcases not adjusted yet. Have some points to discuss


sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -11,66 +11,67 @@ pub mod clients {
...
-        BlobPageBlobClient, BlobServiceClient,
+        AppendBlobClient, AppendBlobClientOptions, BlobClient as GeneratedBlobClient,
+        BlobClientOptions, BlobContainerClient as GeneratedBlobContainerClient,
+        BlobContainerClientOptions, BlobServiceClient as GeneratedBlobServiceClient,
+        BlobServiceClientOptions, BlockBlobClient, BlockBlobClientOptions, PageBlobClient,
+        PageBlobClientOptions,
     };
+    pub use blob_client::BlobClient;
+    pub use blob_container_client::BlobContainerClient;
+    pub use blob_service_client::BlobServiceClient;
```
vincenttran-msft, line 23: We still have a collision on the generated client names and our handwritten client names, not sure if we want to actually resolve this or if renaming the export at the `lib.rs` level is idiomatic and suffices.

sdk/storage/azure_storage_blob/src/generated/clients/blob_client.rs
```diff
@@ -35,22 +45,36 @@ impl BlobClient {
...
     /// * `credential` - An implementation of [`TokenCredential`](azure_core::credentials::TokenCredential) that can provide an
     ///   Entra ID token to use when authenticating.
+    /// * `version` - Specifies the version of the operation to use for this request.
+    /// * `container_name` - The name of the container.
+    /// * `blob` - The name of the blob.
     /// * `options` - Optional configuration for the client.
     pub fn new(
         endpoint: &str,
         credential: Arc<dyn TokenCredential>,
+        version: String,
```
vincenttran-msft, line 55: Note that this generated `BlobClient` takes in a non-optional `version`. This applies to the other abstractions (Append, Block, Page).

It is a little odd since it also saves it directly on the client's struct, but for other clients (i.e. generated `BlobContainerClient` and `BlobServiceClient`) do not take a version and populates their client struct's `version` from the options bag.

So I think the main take-aways are:

1. `BlobClient` and other Blob client abstractions are fundamentally different to how it handles `version` as compared to other abstractions (direct input saved right to their struct vs. implied from options bag and saved to their struct)
2. We should consider making version optional, (i.e. you get this for free through the options bag) and possibly unifying the way `BlobClient` handles version, unless there is a good reason to have it remain different.

ContainerClient options bag example:
```rust
impl Default for BlobContainerClientOptions {
    fn default() -> Self {
        Self {
            client_options: ClientOptions::default(),
            version: String::from("2025-01-05"),
        }
    }
}
```

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -49,11 +52,19 @@ impl BlobClient {
...
 
-        let client = GeneratedBlobClient::new(endpoint, credential, Some(options))?;
+        let client = GeneratedBlobClient::new(
+            endpoint,
+            credential.clone(),
+            "2025-05-05".to_string(),
+            container_name.clone(),
+            blob_name.clone(),
+            Some(options),
+        )?;
```
vincenttran-msft, line 62: Now that we that we don't have an accessor `BlobClient` and this literally the generated client that will perform the REST operations, I think this makes some fields redundant on our handwritten clients.

Generated `BlobClient` struct:
```rust
pub struct BlobClient {
    pub(crate) blob: String,
    pub(crate) endpoint: Url,
    pub(crate) pipeline: Pipeline,
    pub(crate) version: String,
}
```
We can now omit `blob_name`, `container_name`, `endpoint` from the handwritten client layer and instead do `self.client.<field>`

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -21,6 +23,7 @@ pub struct BlobClient {
     endpoint: Url,
     container_name: String,
     blob_name: String,
+    credential: Arc<dyn TokenCredential>,
```
vincenttran-msft, line 26: Again, because we no longer have an accessor client, we need to spin up the other abstraction clients on demand (i.e. `BlockBlockClient`) to perform operations not exposed on the generated `BlobClient`.

This means we need a copy of the auth information used to originally build our `BlobClient` in order to pass it to sub-clients.

However, I think this is a candidate for possibly having `get_X_client` added to the generated `BlobClient`, since it is a pretty typical scenario that we have a `BlobClient` and need a `BlockBlobClient` to perform some operation.


Sample generated getter from `BlobServiceClient` to `BlobContainerClient`:
```rust
/// Returns a new instance of BlobContainerClient.
    ///
    /// # Arguments
    ///
    /// * `container_name` - The name of the container.
    pub fn get_blob_container_client(&self, container_name: String) -> BlobContainerClient {
        BlobContainerClient {
            container_name,
        }
    }
```

sdk/storage/azure_storage_blob/src/generated/clients/blob_client.rs
```diff
@@ -66,111 +90,1375 @@ impl BlobClient {
...
-    ) -> BlobAppendBlobClient {
-        BlobAppendBlobClient {
-            blob,
-            container_name,
-            endpoint: self.endpoint.clone(),
-            pipeline: self.pipeline.clone(),
-            version: self.version.clone(),
+        copy_id: &str,
+        options: Option<BlobClientAbortCopyFromUrlOptions<'_>>,
+    ) -> Result<Response<BlobClientAbortCopyFromUrlResult>> {
```
vincenttran-msft, line 104: Making this comment here but this comment applies broadly to the whole new generated code-- each `Result` returned from these generated REST APIs have a unique `Result` that is a combo of <operation> + <Result> i.e. `BlobCLientAbortCopyFromUrlResult`. They no longer return a general `Response`.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -101,19 +106,27 @@ impl BlobClient {
...
-            .get_blob_block_blob_client(self.container_name.clone(), self.blob_name.clone())
+        // Currently there is no way to alter what options you pass to the BlockBlobClient we spin up on-demand
+        let block_blob_client = BlockBlobClient::new(
+            self.endpoint().as_str(),
+            self.credential(),
+            "2025-05-05".to_string(),
+            self.container_name().to_string(),
+            self.blob_name().to_string(),
+            None,
+        )?;
```
vincenttran-msft, line 127: Making a specific callout here, but this applies broadly to any clients that requires a newly spun up sub-client to carry out the operation.

Currently our handwritten APIs only expose an `options` bag for the actual operation (in this case, `BlockBlobClientUploadOptions`) for upload.

However, there is a usage issue here in the sense that cx can alter the Options bag for the actual REST operation, but not for the `BlockBlobClient` that is being spun up (which takes a different options, `BlobContainerClientOptions`)

Currently my implementation I am passing `None` at L126 (which means it will always take default). But this can be problematic, as for example encoded in the options is the `version`, and so no matter what version they are using at the `BlobClient` level, if they happen to use an operation that needs a `BlockBlobClient`, we will switch back to default values since there no way for cx to provide this.

That means we should probably expose the options bag for the `BlockBlobClient` as well, but I am curious to hear if we think this is an area we could improve on.

Again, I think adding back accessors to sub-clients (from generated `BlobClient` to `BlockBlockClient`) would solve this elegantly, and we could add the options bag knob there when using the accessor to either set a different options or inherit from the main `BlobClient`.

sdk/storage/azure_storage_blob/src/generated/clients/blob_client.rs
vincenttran-msft: Going to be resolved (this is a bug in the emitter compiler)

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: Omit from struct, add getters (that will go 1 layer down into gen client)

PR #2294, opened 2025-03-06
Bump versions and simplify dependencies

heaths:
At least in the workspace `Cargo.toml`, crates that should only ever be in `dev-dependencies` can have a `path` and no `version`. This makes referencing them easier, but I'd be fine with crates taking a local `path` as well.


Cargo.toml
```diff
@@ -29,35 +29,35 @@ rust-version = "1.80"
...
 path = "sdk/core/azure_core_test"
 
 [workspace.dependencies.azure_core_test_macros]
-version = "0.1.0"
+# azure_core_test_macros is not published and only ever a dev-dependency
 path = "sdk/core/azure_core_test_macros"
 
 [workspace.dependencies.azure_identity]
-version = "0.22.0"
+# azure_identity should only ever be in dev-dependencies herein
```
hallipr, line 60: `azure_storage_blob` is still taking a workspace dependency on `azure_identity`

heaths: Yes, in a dev-dependency which is now a path. I don't see the problem. Could you clarify?

heaths: Oh, wait, as a regular dependency too. That's not allowed, @vincenttran-msft. Let me see if it's actually used in shipping code...

heaths:
What's breaking the MSRV builds is likely a change to rustc/cargo that check crates' `rust-version` better than before:

```text
error: rustc 1.80.1 is not supported by the following packages:
  litemap@0.7.5 requires rustc 1.81
  zerofrom@0.1.6 requires rustc 1.81
Either upgrade rustc or select compatible dependency versions with
`cargo update <name>@<current-ver> --precise <compatible-ver>`
where `<compatible-ver>` is the latest version supporting rustc 1.80.1
```

Running `cargo tree -i` on these dependencies show they are coming in from `url`...which further adds weight to replacing it with our own. I'll have to see how I can either avoid these dependencies (they're used for IDNA, which is unlikely we'll ever need) or...I don't know.

LarryOsterman (approved):

PR #2295, opened 2025-03-06
Fix issue bumping wrong versions

hallipr:
The square brackets matched regex character class syntax and should be escaped.


heaths (approved):
You'll need until I can merge #2294

LarryOsterman (approved):

PR #2298, opened 2025-03-06
Remove Unnecessary Proxy Install

scbedd:
fyi @heaths @hallipr 

If `cargo test` takes care of downloading the proxy in all cases, this build yaml is no longer necessary.

### DIFF
--- a/eng/pipelines/templates/jobs/test.yml
+++ b/eng/pipelines/templates/jobs/test.yml
@@ -36,9 +36,6 @@ jobs:
       ServiceDirectory: ${{ parameters.ServiceDirectory }}
       PackageInfoDirectory: $(Build.ArtifactStagingDirectory)/PackageInfo
 
-  - ${{ if eq(parameters.TestProxy, true) }}:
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/jobs/test.yml
```diff
@@ -36,9 +36,6 @@ jobs:
       ServiceDirectory: ${{ parameters.ServiceDirectory }}
       PackageInfoDirectory: $(Build.ArtifactStagingDirectory)/PackageInfo
 
-  - ${{ if eq(parameters.TestProxy, true) }}:
```
heaths, line 39: If it updates `PATH` is still might be worth it. Just means everything is ready to go and there's no guess work since, without DEBUG tracing, it's all a silent install inside `cargo`. But if not...which begs the question "what's the point of the step, then", we can remove it.

heaths (approved):

eng/pipelines/templates/jobs/test.yml
scbedd: It absolutely prepends the path so let's leave this for now as agreed offline.

PR #2306, opened 2025-03-08
Add AzurePipelinesCredential

### LINKED ISSUE #2030: Add AzurePipelineCredential for test authentication
For test authentication congruent with other languages, we need to return a `ChainedTokenCredential` with `AzurePipelinesCredential` and `AzurePowerShellCredential` similar to:

* .NET
  * [AzurePipelinesCredential](https://github.com/Azure/azure-sdk-for-net/blob/ecbf39e8094af405e7a3a4abc1ea4b611c051536/sdk/identity/Azure.Identity/src/Credentials/AzurePipelinesCredential.cs)
  * [AzurePowerShellCredential](https://github.com/Azure/azure-sdk-for-net/blob/ecbf39e8094af405e7a3a4abc1ea4b611c051536/sdk/identity/Azure.Identity/src/Credentials/AzurePowerShellCredential.cs)
* Go
  * [AzurePipelinesCredential](https://github.com/Azure/azure-sdk-for-go/blob/051993c29d9e2b0c7bd4997407e6401af5c28380/sdk/azidentity/azure_pipelines_credential_test.go#L21)

heaths:
Resolves #2030. This includes some small amount of refactoring of existing code that will be necessary sooner than later, but I didn't want to take all that on in this PR focused on `AzurePipelinesCredential`. Still, I wasn't about to repeat the same antiquated patterns.


heaths:
There are some TODOs that require some heavy lifting, like using either the `azure_core` `Pipeline` or making an `azure_identity`-specific one; however, I want to focus on an MVP that enables our live tests since we need to get those up and running ASAP (/cc @hallipr). `azure_identity` needs a lot of work already including some major refactoring, using more of the pipelining, etc., so we can do all that then and be consistent.

@RickWinter @JeffreyRichter @analogrelay as we start getting more and more nested options, the API is feeling a bit unwieldy. In addition, I'm not crazy about making it obvious we use a `ClientAssertionCredential` by nesting it's options. It's not that I think we should hide it, but feels like shipping an implementation detail (just like we shouldn't ship our org chart). Guidelines do leave the possibility of builders open, so should we consider that for credentials? Or do we duplicate fields and manage all that as we make changes? It's not terribly hard since Rust forces us to assign fields if we don't use `..Default::default()`.
I'm open to other suggestions on design, too.

sdk/core/azure_core_test/src/http/clients.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+impl<C> HttpClient for MockHttpClient<C>
+where
+    C: FnMut(&Request) -> Pin<Box<dyn Future<Output = Result<Response>> + Send + Sync + '_>>
+        + Send
+        + Sync,
```
heaths, line 34: @analogrelay any other good ideas here? The DX this forces (see test below; we have to pin the future) is unwieldy but I couldn't figure out a better way without taking advantage of Rust 1.85 async captures.

heaths: I cleaned it up a little using a `BoxedFuture`, though the `boxed()` "extension method" probably does more heavy lifting here. Still open to other suggestions, but this seems to be a fairly common solution I found - more than what inspired me originally.

sdk/identity/azure_identity/src/azure_pipelines_credential.rs
```diff
@@ -0,0 +1,322 @@
...
+const ALLOWED_HEADERS: &[&str] = &["x-msedge-ref", "x-vss-e2eid"];
+
+#[derive(Debug)]
+pub struct AzurePipelinesCredential(ClientAssertionCredential<Client>);
+
+/// Options for constructing a new [`AzurePipelinesCredential`].
+#[derive(Debug, Default)]
+pub struct AzurePipelinesCredentialOptions {
+    /// Options for the [`ClientAssertionCredential`] used by the [`AzurePipelinesCredential`].
+    pub credential_options: ClientAssertionCredentialOptions,
```
heaths, line 32: @demoray see here and related code for an example of what I was referring to in #2309

sdk/identity/azure_identity/src/azure_pipelines_credential.rs
```diff
@@ -0,0 +1,323 @@
...
+// TODO: https://github.com/Azure/azure-sdk-for-rust/issues/682
+const ALLOWED_HEADERS: &[&str] = &["x-msedge-ref", "x-vss-e2eid"];
+
+#[derive(Debug)]
+pub struct AzurePipelinesCredential(ClientAssertionCredential<Client>);
+
+/// Options for constructing a new [`AzurePipelinesCredential`].
+#[derive(Debug, Default)]
+pub struct AzurePipelinesCredentialOptions {
+    /// Options for the [`ClientAssertionCredential`] used by the [`AzurePipelinesCredential`].
```
chlowell, line 31: I doubt it will change, but this relationship is an implementation detail, so it feels risky to make it part of the public API

chlowell (approved):

sdk/identity/azure_identity/src/azure_pipelines_credential.rs
heaths: It's also not idiomatic to "reflect" changes, but this isn't settled yet. I already have a separate thread on this.

PR #2307, opened 2025-03-10
Fix versions for pack and improve logs

hallipr:


eng/scripts/Get-PackageDependencies.ps1
```diff
@@ -2,12 +2,31 @@
 #Requires -Version 7.0
 
 [CmdletBinding()]
+param(
+  [switch]$Publishable,
+  [switch]$NoDev,
+  [switch]$NoReq
```
heaths, line 8: Nit: PowerShell allows as few characters as necessary to disambiguate, but we shouldn't use shorter names that may not be obvious to everyone using it.
```suggestion
  [switch]$NoDeveloperDependencies,
  [switch]$NoRequiredDependencies
```
Let's keep them intuitive for easier use/maintenance.

heaths (approved):

eng/scripts/Get-PackageDependencies.ps1
hallipr: req is an output property from `cargo metadata`, short for Version Requirement
`req == '*'` means that the dependency had no version specified.
NoReq would return only entries where `req == '*'`, so NoVersion?

hallipr: or NoRequirement?

hallipr: going with NoVersion

PR #2309, opened 2025-03-10
Add ChainedTokenCredential

demoray:


sdk/identity/azure_identity/src/credentials/chained_credentials.rs
```diff
@@ -0,0 +1,135 @@
...
+use crate::credentials::cache::TokenCache;
+use azure_core::{
+    credentials::{AccessToken, TokenCredential},
+    error::{Error, ErrorKind},
+};
+use std::sync::Arc;
+
+/// Provides a mechanism of selectively adding credentials used for a `ChainedTokenCredential` instance
+#[derive(Default)]
+pub struct ChainedTokenCredentialBuilder {
```
heaths, line 13: We don't use builders anymore. All options should be plain old objects per [guidelines](https://azure.github.io/azure-sdk/rust_introduction.html#rust-client-configuration).

sdk/identity/azure_identity/src/credentials/chained_credentials.rs
heaths, line 1: Let's have the file name match the credential type i.e. `chained_token_Credential.rs`. Also, move it to the src root. I'll be doing that soon with the rest. There's no point of nesting them like this. We flatten the imports in `src/lib.rs`, but this submodule serves no purpose.

sdk/identity/azure_identity/src/credentials/chained_credentials.rs
```diff
@@ -0,0 +1,135 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::credentials::AppServiceManagedIdentityCredential;
+    use crate::credentials::AzureCliCredential;
+    use crate::TokenCredentialOptions;
+
+    #[test]
+    fn test_builder_included_credential_flags() -> azure_core::Result<()> {
+        let mut builder = ChainedTokenCredentialBuilder::new();
```
heaths, line 122: When you define a `ChainedTokenCredentialOptions`, make sure to nest a `TokenCredentialOptions` named `credential_options` (for now; I may rename that later). That defines a `pub(crate) env` that you can use. That said, the test error indicates you're checking for an ultimately failing on a missing env var, which shouldn't happen for a chain. It shouldn't be a fatal error unless there were no credentials that satisfied the chain.

heaths (changes requested):
Thanks for the contribution! As noted, though, we've made a few changes based on discussions. These are reflected in the guidelines and I noted them where appropriate.

sdk/identity/azure_identity/src/credentials/chained_credentials.rs
demoray: Renamed.

sdk/identity/azure_identity/src/credentials/chained_credentials.rs
demoray: Added, though I'm not sure where the options would/should be used.

I modeled my implementation off of how the Python [ChainedTokenCredential](https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.chainedtokencredential?view=azure-python) works, which consumes existing token credential impls, rather than creating them directly (as DAC operates).

sdk/identity/azure_identity/src/credentials/chained_credentials.rs
heaths: The file was moved but not renamed (the file name itself, that is). For credentials, I'd like to keep the file names the same as the credential therein.

sdk/identity/azure_identity/src/credentials/chained_credentials.rs
heaths: See <https://github.com/Azure/azure-sdk-for-go/blob/cdaa40165f7fcb23cc59cbb0ba3e755b50416df8/sdk/azidentity/chained_token_credential.go#L22>. Structurally, we are more akin to Go (and C++) - the "Jeffrey languages". That said, at the very least it would contain a `TokenCredentialOptions` which has the `env` and `http_client`.

sdk/identity/azure_identity/src/credentials/chained_credentials.rs
demoray: The filename is now `chained_token_credential.rs`, which aligns to `ChainedTokenCredential`

sdk/identity/azure_identity/src/credentials/chained_credentials.rs
demoray: Understood.  

Note, the go implementation of ChainedTokenCredential [chained_token_credential.go](https://github.com/Azure/azure-sdk-for-go/blob/cdaa40165f7fcb23cc59cbb0ba3e755b50416df8/sdk/azidentity/chained_token_credential.go#L22) does not include the golang version of `TokenCredentialOptions` which I believe is the `azcore.ClientOptions` as used by credential impls (see [azure_pipelines_credentia.go](https://github.com/Azure/azure-sdk-for-go/blob/cdaa40165f7fcb23cc59cbb0ba3e755b50416df8/sdk/azidentity/azure_pipelines_credential.go#L38), and [default_azure_credential.go](https://github.com/Azure/azure-sdk-for-go/blob/cdaa40165f7fcb23cc59cbb0ba3e755b50416df8/sdk/azidentity/default_azure_credential.go#L24).

Note, I did update the implementation to include `retry_sources` which models the golang implementation.

heaths (approved):

sdk/identity/azure_identity/src/chained_token_credential.rs
```diff
@@ -0,0 +1,155 @@
...
+        Self {
+            options: options.unwrap_or_default(),
+            sources: Vec::new(),
+            cache: TokenCache::new(),
+            successful_credential: RwLock::new(None),
+        }
+    }
+
+    /// Add a credential source to the chain.
+    pub fn add_source(&mut self, source: Arc<dyn TokenCredential>) {
```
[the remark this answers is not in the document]
demoray: Added.

With the recent inclusion of credential testing in azure_core_test, I will add testing for that as well.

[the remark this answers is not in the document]
demoray: I added said `From`, plus integration tests that validate chaining works as well as the `retry_sources` impl

RickWinter (approved):

heaths (approved):

PR #2322, opened 2025-03-11
Add live tests template

hallipr:


sdk/keyvault/ci.yml
```diff
@@ -14,6 +14,7 @@ extends:
   template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
   parameters:
     ServiceDirectory: keyvault
+    RunLiveTests: true
```
heaths, line 17: Seems this should be the default, or at least if they have a `test-resources.{json|bicep}` in their service directory.

hallipr: I was just following Go:
https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/messaging/azeventhubs/ci.yml#L29

hallipr: > or at least if they have a `test-resources.{json|bicep}` in their service directory

We wouldn't be able to detect that at yaml compile time.  We can enforce it with a linter in PR.

PR #2323, opened 2025-03-11
Add matrix-based testing

hallipr:


eng/pipelines/templates/jobs/ci.yml
```diff
@@ -44,17 +26,24 @@ jobs:
...
+  parameters:
+    JobTemplatePath: /eng/pipelines/templates/jobs/ci.tests.yml
+    MatrixConfigs: ${{ parameters.MatrixConfigs }}
+    MatrixFilters: ${{ parameters.MatrixFilters }}
+    MatrixReplace: ${{ parameters.MatrixReplace }}
+    CloudConfig:
+      Cloud: public
+    SparseCheckoutPaths:
+      - /*
+      - '!/sdk/*/**/SessionRecords/*'
```
heaths, line 45: 1. This is .NET's pattern, not ours.
2. I don't think it's needed even for .NET since session records moved out into another repo and are put elsewhere. In our repo (and we copied from .NET et. al.) it's gitignore'd `.assets/` directory off the root.

heaths (approved):
Apart from the one comment, I don't see anything obviously wrong, but @benbp might be a better reviewer here. And, at least as of when I wrote this, all the pipelines failed so there's something clearly wrong but I couldn't say what from what I see. But I don't see anything to hold the PR up.

eng/pipelines/templates/jobs/ci.yml
heaths: While I don't think we even need it, FWIW, our records "virtually" get put in `sdk/**/tests/data/**`.

hallipr: SessionRecords line removed

RickWinter (approved):

eng/pipelines/templates/jobs/ci.yml
```diff
@@ -44,17 +26,23 @@ jobs:
...
+- template: /eng/common/pipelines/templates/jobs/generate-job-matrix.yml
+  parameters:
+    JobTemplatePath: /eng/pipelines/templates/jobs/ci.tests.yml
+    MatrixConfigs: ${{ parameters.MatrixConfigs }}
+    MatrixFilters: ${{ parameters.MatrixFilters }}
+    MatrixReplace: ${{ parameters.MatrixReplace }}
+    CloudConfig:
+      Cloud: public
+    SparseCheckoutPaths:
+      - /*
```
benbp, line 44:
```suggestion
```

I don't think you need this? It's just for checking out files that may be necessary for matrix generation outside of eng+service directory.

eng/pipelines/templates/stages/archetype-sdk-client.yml
```diff
@@ -4,17 +4,34 @@ parameters:
...
-  type: string
-  default: release
+- name: MatrixConfigs
+  type: object
+  default:
+  - Name: rust_ci_test_base
+    Path: eng/pipelines/templates/stages/platform-matrix.json
+    Selection: sparse
+    NonSparseParameters: RustToolchainName
+    GenerateVMJobs: true
```
benbp, line 18: The way the matrix is set up now, you could specify it with `all` instead. Though perhaps you have it with `NonSparseParameters` to handle when more matrix parameters are added?

```suggestion
  default:
  - Name: rust_ci_test_base
    Path: eng/pipelines/templates/stages/platform-matrix.json
    Selection: all
    GenerateVMJobs: true
```

benbp (approved):

eng/pipelines/templates/jobs/ci.yml
benbp: > 1. This is .NET's pattern, not ours.
> 2. I don't think it's needed even for .NET since session records moved out into another repo and are put elsewhere. In our repo (and we copied from .NET et. al.) it's gitignore'd `.assets/` directory off the root.

@heaths unfortunately net still has some stragglers that need to clean up their recordings.

eng/pipelines/templates/jobs/ci.yml
hallipr: So `ci.yml` will need all the code because any of it may need to be built.  Depending on how cargo handles partial workspaces, we may still need all the cargo.tomls to do the `cargo metadata` call even when working with a single package.

eng/pipelines/templates/stages/archetype-sdk-client.yml
hallipr: I wasn't sure how selection: all would interact with sparse coming from the PR matrix generation

eng/pipelines/templates/jobs/ci.yml
heaths: To avoid long check-out times, I wonder if we could just check out:
```yaml
- /*
- /sdk/**/Cargo.toml
```

PR #2330, opened 2025-03-13
Add live tests template

hallipr:


eng/pipelines/templates/jobs/live.tests.yml
```diff
@@ -0,0 +1,147 @@
...
+          -PackageInfoDirectory '$(Build.ArtifactStagingDirectory)/PackageInfo'
+          -Toolchain '$(Toolchain)'
+  - ${{ else }}:
+    - task: Powershell@2
+      displayName: "Test Packages"
+      timeoutInMinutes: ${{ parameters.TimeoutInMinutes }}
+      env:
+        CIBW_BUILD_VERBOSITY: 3
+        AZURE_TEST_MODE: 'live'
+        SYSTEM_ACCESSTOKEN: $(System.AccessToken)
```
heaths, line 130:
```suggestion
        RUST_LOG=trace
        SYSTEM_ACCESSTOKEN: $(System.AccessToken)
```

heaths:
Also, shouldn't there be separate `- live` pipelines e.g., `rust - keyvault - live`? I thought this was how all services were doing it. If we can manage it such that PRs use recordings while CIs use live resources with a single archetype great. But then how do we use `/azp run ???` if they all originate from a PR? It's handy to run against live resources at times and by a separate host if, say, we have a laptop and need to get to a meeting, drive home, etc.

eng/pipelines/templates/jobs/live.tests.yml
```diff
@@ -0,0 +1,148 @@
...
+          -PackageInfoDirectory '$(Build.ArtifactStagingDirectory)/PackageInfo'
+          -Toolchain '$(Toolchain)'
+  - ${{ else }}:
+    - task: Powershell@2
+      displayName: "Test Packages"
+      timeoutInMinutes: ${{ parameters.TimeoutInMinutes }}
+      env:
+        CIBW_BUILD_VERBOSITY: 3
+        AZURE_TEST_MODE: 'live'
+        RUST_LOG=trace
```
heaths, line 130: Oops. Wrong syntax anyway.
```suggestion
```

hallipr:
> Also, shouldn't there be separate `- live` pipelines e.g., `rust - keyvault - live`? 

Most languages are moving to common PR pipelines, so ci.yml won't be PR triggering.  It'll just be for post-merge "IndividualCI" invocation and manual for release.   Then there's the separate tests.yml that does the nightly, weekly scheduled runs.  Go uses just ci.yml for CI, manual and live tests.  I prefer that.

We do have some code in the common PR powershell that parses ci.yml to know if a service/package is triggered by a path, but the pull request yaml doesn't look at a service's ci.yml

hallipr:
> But then how do we use `/azp run ???`

We're currently doing that in this pr.

`/azp run rust - keyvault` will run the internal `rust - keyvault` against this pr. It'll use `sdk/keyvault/ci.yml` that uses `archetype-sdk-client.yml` with `RunLiveTests: true` (default) which will cause it to use `live.tests.yml`.

We also get the automatic run of `rust - pullrequest` that uses `eng/pipelines/pullrequest.yml` that uses `archetype-sdk-client.yml` with `RunLiveTests: false` (explicit)  which will cause it to use `ci.tests.yml`.

hallipr:
You can also manually queue `rust - keyvault` and pick any ref to build.  If you add `Skip.Release: true` to the variables when you queue it, you don't have to worry about accidentally releasing it.

eng/pipelines/templates/jobs/live.tests.yml
```diff
@@ -0,0 +1,147 @@
...
+      displayName: "Test Packages - Federated Auth"
+      timeoutInMinutes: ${{ parameters.TimeoutInMinutes }}
+      env:
+        CIBW_BUILD_VERBOSITY: 3
+        AZURE_TEST_MODE: 'live'
+        SYSTEM_ACCESSTOKEN: $(System.AccessToken)
+        ${{ if parameters.PersistOidcToken }}:
+          ARM_OIDC_TOKEN: $(ARM_OIDC_TOKEN)
+        ${{ insert }}: ${{ parameters.EnvVars }}
+      inputs:
```
weshaggard, line 114: If you plan to use AzurePipelineCredential you will also need to set these variables https://github.com/Azure/azure-sdk-for-net/blob/main/eng/pipelines/templates/jobs/live.tests.yml#L177

eng/pipelines/templates/jobs/ci.tests.yml
```diff
@@ -68,3 +71,10 @@ jobs:
       arguments: >
         -PackageInfoDirectory '$(Build.ArtifactStagingDirectory)/PackageInfo'
         -Toolchain '$(Toolchain)'
+
+  - ${{ if eq(parameters.TestProxy, true) }}:
```
heaths, line 75: Could we only do this if `System.Debug` is set too? That's a lot of output. At the very least, collapse it into a region like you do for other tools.

eng/pipelines/templates/jobs/live.tests.yml
heaths: We will need those, yes.

eng/pipelines/templates/stages/archetype-sdk-client.yml
```diff
@@ -48,17 +88,66 @@ extends:
...
 
-    - template: archetype-rust-release.yml
-      parameters:
-        DependsOn: "Build"
-        ServiceDirectory: ${{ parameters.ServiceDirectory }}
-        Artifacts: ${{ parameters.Artifacts }}
-        TestPipeline: ${{ eq(parameters.ServiceDirectory, 'template') }}
-        PipelineArtifactName: packages
+    # Run live tests for internal only, not public CI builds. This can be triggered manually for via an `/azp run` comment.
+    - ${{if eq(parameters.RunLiveTests, 'true') }}:
```
heaths, line 97: Are you sure that's true? IIRC, `/azp run {pipeline}` still treats it as a PR trigger.

sdk/cosmos/ci.yml
```diff
@@ -14,6 +14,7 @@ extends:
   template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
   parameters:
     ServiceDirectory: cosmos
+    RunLiveTests: false
```
heaths, line 17: /cc @analogrelay

sdk/typespec/ci.yml
```diff
@@ -14,6 +14,7 @@ extends:
   template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
   parameters:
     ServiceDirectory: typespec
+    RunLiveTests: false
```
heaths, line 17: Can't we leave these as default and just let whether any `test-resources.*` files exist dictate whether to run them or not? Adding that + this variable duplicates effort.

eng/pipelines/templates/jobs/ci.tests.yml
hallipr: I'm not sure that collapsing it will do much as it's already isolated to its own task.  I like the idea of conditioning it on System.Debug

heaths:
@hallipr you'll need #2333 before this will (hopefully) work.

eng/pipelines/templates/stages/archetype-sdk-client.yml
hallipr: It does. It that an issue?

heaths:
I kicked off a previous run of Key Vault and verified that we are seeing live test traces! 🥳

eng/pipelines/templates/stages/archetype-sdk-client.yml
heaths: As we talked offline, that shouldn't happen in the pullrequest.yml, so cool.

sdk/core/azure_core_test/src/recorded.rs
heaths, line 1: This was already merged to main, but I cherry-picked it here so we could prove live tests were actually running, well, live.

heaths (approved):

PR #2331, opened 2025-03-13
Redefine Method, StatusCode from http_types in typespec

### LINKED ISSUE #1644: Consider replacing http-types crate with http crate

heaths:
Resolves #1644 and replaces #2233. The latter PR was a good attempt but we decided internally we don't want to take a dependency on a different crate - `http` ; though, we do have a transitive dependency on it already - for something so simple.

This starts as a copy from http_types with appropriate attribution. We may want to remove a bunch of Methods we don't need, though, at some point.


sdk/typespec/src/http/status_code.rs
```diff
@@ -493,67 +506,80 @@ impl StatusCode {
...
+            StatusCode::GatewayTimeout => Cow::Borrowed("Gateway Timeout"),
+            StatusCode::HttpVersionNotSupported => Cow::Borrowed("HTTP Version Not Supported"),
+            StatusCode::VariantAlsoNegotiates => Cow::Borrowed("Variant Also Negotiates"),
+            StatusCode::InsufficientStorage => Cow::Borrowed("Insufficient Storage"),
+            StatusCode::LoopDetected => Cow::Borrowed("Loop Detected"),
+            StatusCode::NotExtended => Cow::Borrowed("Not Extended"),
+            StatusCode::NetworkAuthenticationRequired => {
+                Cow::Borrowed("Network Authentication Required")
+            }
+            StatusCode::UnknownValue(code) => Cow::Owned(format!("HTTP {code}")),
```
heaths, line 582: @JeffreyRichter @RickWinter this name is in line with our [guidelines](https://azure.github.io/azure-sdk/rust_introduction.html#rust-enum-extensible) but wondering, for this case, if we should use `Other(u16)` or either `OtherStatusCode(u16)` to avoid potential naming conflicts should an "OTHER" ever be added (unlikely, but possible).

LarryOsterman (dismissed):

sdk/typespec/src/http/status_code.rs
```diff
@@ -646,106 +665,134 @@ mod serde {
...
 }
 
-impl std::convert::TryFrom<u16> for StatusCode {
-    type Error = crate::Error;
-
-    fn try_from(num: u16) -> Result<Self, Self::Error> {
+impl From<u16> for StatusCode {
+    fn from(num: u16) -> Self {
         match num {
-            100 => Ok(StatusCode::Continue),
```
LarryOsterman: Stupid question: Would it make sense to embody this in a table which has #, StatusCode, and String so that the two separate big match statements could be collapsed?

heaths: If we had contiguous numbers that would be useful, but these are not contiguous. And it's incredibly risky to reference arbitrary references. WE could define consts, but that just ends up being even more work to maintain. Do you have a better suggestion?

LarryOsterman (approved):

sdk/typespec/src/http/status_code.rs
LarryOsterman: That's totally fair. And yeah, the compiler can (and will) do magic for non-contiguous match cases.

PR #2333, opened 2025-03-13
Add Content-Length for AzurePipelinesCredential

heaths:


LarryOsterman (approved):
QQ: Why doesn't the credential use a pipeline instead of going directly to HttpClient? 

With a pipeline, you get some nice semantics like retry and user-agent support.

heaths:
> QQ: Why doesn't the credential use a pipeline instead of going directly to HttpClient?
> 
> With a pipeline, you get some nice semantics like retry and user-agent support.

It will. I opened a separate issue but I want to refactor the whole stack. More credentials need credential options, and I will probably have a `client_options` at the bottom of that stack. See our internal channel for a thread I started last week that I don't think anyone has responded to yet. 😉

PR #2334, opened 2025-03-13
Update Cosmos tests to be record/replay tests

analogrelay:
This PR moves all Cosmos' integration tests from "Live-only" to Record/Replay/Live tests. Recorded service responses are now present in [azure-sdk-assets](https://github.com/Azure/azure-sdk-assets/tree/rust/azure_data_cosmos_43a7f435b6) and running `cargo test` without setting `AZURE_TEST_MODE` will "replay" those recorded responses and run the tests against those.

I've recorded these against the Emulator, because it's the smoothest experience for key-based auth since it uses a well-known key that isn't considered a "secret". When it comes time for Live testing, I think we could test everything _except_ Database creation, which in my initial testing is the only remaining operation that _requires_ key-based auth (it looks like Container creation can be done using Entra ID now). The Emulator is considered a full implementation of the core APIs so I'm quite happy using it for our testing. When we move to things like Vector indexing and other "new" features, we'll need to make sure we can test them against the live service, but that can wait.


.devcontainer/devcontainer.json
heaths, line 1: Great ideas!

.devcontainer/oncreate
```diff
@@ -6,3 +6,6 @@ echo "Running Azure SDK for Rust devcontainer post create script..."
 
 echo "Updating toolchains..."
 rustup show
```
heaths, line 8: This won't work anymore. While you're in there, can you change this to `rustup install`?

.devcontainer/oncreate
```diff
@@ -6,3 +6,6 @@ echo "Running Azure SDK for Rust devcontainer post create script..."
...
+
+echo "Installing test-proxy..."
+dotnet tool update azure.sdk.tools.testproxy --global --prerelease --add-source https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json --ignore-failed-sources
```
heaths, line 11: This isn't necessary. The test proxy will be acquired automatically and the EXE installed by this tool doesn't use the name we launch, so this step is basically pointless.

.devcontainer/devcontainer.json
heaths: ...actually, after my comment below, I realized that the ZIP we download is self-contained, so we don't need .NET...except for PowerShell, but I'm assuming that installs .NET as a pre-req. Maybe not, though.

heaths: ...though, I suppose installing the tool is still useful to run commands like `test-proxy push -a {assets.json}`.

sdk/cosmos/azure_data_cosmos/tests/framework/test_account.rs
```diff
@@ -91,14 +106,8 @@ impl TestAccount {
...
-            context.name(),
-            OffsetDateTime::now_utc().format(format_description!(
-                "[year]_[month]_[day]T[hour]_[minute]_[second]"
-            ))?,
-            Uuid::new_v4().as_simple()
-        );
+        // We need the context_id to be constant, so that record/replay work.
+        let context_id = context.name().to_string();
 
         TRACING.call_once(|| {
```
heaths, line 112: You may not need this since the test proxy code does it using `RUST_LOG` as well. You might try deleting this code and running your tests e.g.,

```bash
RUST_LOG=debug,azure_data_cosmos=trace cargo test -p azure_data_cosmos
```

Or something like that. You typically don't want to trace everything because `reqwest` and `hyper` are noisy.

.devcontainer/oncreate
analogrelay: The intent here was to support running `test-proxy push -a .../assets.json`, but I can remove it if you'd prefer otherwise

heaths: I realized that above in my comments. I see you removed it and the `dotnet` devcontainer feature. I'd put them back. I wasn't thinking about that originally, but it makes sense.

analogrelay: Done!

heaths (approved):

Pilchie:
> I think we could test everything except Database creation, which in my initial testing is the only remaining operation that requires key-based auth (it looks like Container creation can be done using Entra ID now).

This is a bit in flux, and depends on what subscription you run things in. We can talk offline about more details here when necessary.

Pilchie (approved):

PR #2340, opened 2025-03-14
Run `cargo deny check` in Analyze job

heaths:


heaths:
@hallipr @RickWinter @LarryOsterman this is going to fail until we clean a few more things up. I got rid of some issues - mainly duplicates - when I got rid of `http-types`, but we still have `ring` popping up and iron oxide is causing a few dups as well, though it's not the end of the world. We should try to clean up as many possible.

I need to see if BSL-1.0 - the "Boost Software License" - is the same thing as the "Boot Software Code" license mentioned internally. I'll send @ronniegeraghty the internal link to track down.

eng/scripts/Analyze-Code.ps1
```diff
@@ -24,6 +25,10 @@ if ($CheckWasm) {
   Invoke-LoggedCommand "rustup target add --toolchain $Toolchain wasm32-unknown-unknown"
 }
 
+if ($Deny) {
+  Invoke-LoggedCommand "cargo +$Toolchain install cargo-deny --locked"
```
heaths, line 29: @hallipr we might actually want to move this out of the script - the install, specifically - so we can take advantage of build caching. Or maybe we can still do that if we collect the right directories in Azure Pipelines. For GitHub Actions, I do this: <https://github.com/heaths/template-rustlang/blob/2b23770467f06c59977edfbeea84e96b144a702e/.github/workflows/ci.yml#L37-L45>

heaths:
This now removes the `ring` dependency, which wasn't obvious from the `Cargo.lock` file but once I incorporated `cargo deny` and started adding [`[[bans.features]]`](https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html#the-features-field-optional) I was able to figure out that our `reqwest_rustls` could enable `ring` because we enabled the `rustls-tls-native-roots` instead of the `rustls-tls-native-roots-no-provider` feature, which requires configuring the cryptography provider if needed e.g., `symcrypt`.

sdk/typespec/typespec_client_core/Cargo.toml
```diff
@@ -52,7 +52,7 @@ json = ["typespec/json"]
 reqwest = ["reqwest/native-tls"]
 reqwest_deflate = ["reqwest/deflate"]
 reqwest_gzip = ["reqwest/gzip"]
-reqwest_rustls = ["reqwest/rustls-tls-native-roots"]
+reqwest_rustls = ["reqwest/rustls-tls-native-roots-no-provider"]
```
RickWinter, line 55: Lets put the comment you have in the CHANGELOG next to this line as well. I have a suspicion that down the road devs might wonder why we picked this feature flag. Your comment will remediate and avoid them pestering you ;)

RickWinter (approved):

sdk/typespec/typespec_client_core/Cargo.toml
heaths: When the TOML is formatted, comments won't stay with the line. What we need to do - and I could start here with this create in this PR - is document the features. There's no built-in way with cargo (yet - there is a feature request or two) but it's common to put a Features section in the README. That would be in the same commit so people properly looking through commit details would see.

heaths: Well, I suppose I could also put a comment to the right of the line as well. I'm going to keep it shorter.

RickWinter (approved):

eng/scripts/Analyze-Code.ps1
hallipr: Shouldn't we still be able to use build caching.  It may be missing cargo-deny, but most other packages and the target folder will be cached

hallipr (approved):

eng/scripts/Analyze-Code.ps1
heaths: Should, yes. It's basically just a `cargo build` of an external crate. All dependencies end up in the same place. But is that part of the "analyze" step already?

Whatever the case, this shouldn't block this PR so I'm resolving the comment.

PR #2342, opened 2025-03-17
When updating versions in Cargo.tomls, also update dependencies

hallipr:
If another package takes a path and version dependency on the package being updated, replace the version in the dependency as well

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ version = "0.2.0"
 path = "sdk/typespec/typespec_macros"
 
 [workspace.dependencies.azure_core]
-version = "0.99.0"
+version = "0.23.0"
 path = "sdk/core/azure_core"
 default-features = false
 # Keep these in sync with azure_core. This allows
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,33 +1,169 @@
 [workspace]
-resolver = "2"
+exclude = ["eng/scripts"]
 members = [
-  "sdk/typespec",
-  "sdk/typespec/typespec_client_core",
-  "sdk/typespec/typespec_macros",
-  "sdk/core/azure_core",
-  "sdk/core/azure_core_amqp",
-  "sdk/core/azure_core_test",
-  "sdk/core/azure_core_test_macros",
-  "sdk/cosmos/azure_data_cosmos",
-  "sdk/identity/azure_identity",
-  "sdk/eventhubs/azure_messaging_eventhubs",
-  "sdk/keyvault/azure_security_keyvault_keys",
-  "sdk/keyvault/azure_security_keyvault_secrets",
-  "sdk/storage",
-  "sdk/storage/azure_storage_blob",
-  "sdk/template/azure_template_core",
-  "sdk/template/azure_template",
+    "sdk/typespec",
+    "sdk/typespec/typespec_client_core",
+    "sdk/typespec/typespec_macros",
+    "sdk/core/azure_core",
+    "sdk/core/azure_core_amqp",
+    "sdk/core/azure_core_test",
+    "sdk/core/azure_core_test_macros",
+    "sdk/cosmos/azure_data_cosmos",
+    "sdk/identity/azure_identity",
+    "sdk/eventhubs/azure_messaging_eventhubs",
+    "sdk/keyvault/azure_security_keyvault_keys",
+    "sdk/keyvault/azure_security_keyvault_secrets",
+    "sdk/storage",
+    "sdk/storage/azure_storage_blob",
+    "sdk/template/azure_template_core",
+    "sdk/template/azure_template",
 ]
-exclude = [
-  "eng/scripts",
-]
+resolver = "2"
 
-[workspace.package]
-authors = ["Microsoft"]
-edition = "2021"
-license = "MIT"
-repository = "https://github.com/azure/azure-sdk-for-rust"
-rust-version = "1.80"
+[workspace.dependencies]
+async-lock = "3.0"
+async-trait = "0.1"
+base64 = "0.22"
+bytes = "1.0"
+cargo_metadata = "0.18.1"
+dyn-clone = "1.0"
+flate2 = "1.1.0"
+futures = "0.3"
+litemap = "0.7.4"
+log = "0.4"
+once_cell = "1.18"
+pin-project = "1.0"
+proc-macro2 = "1.0.86"
+quote = "1.0.37"
+rand = "0.8"
+rand_chacha = "0.3"
+rustc_version = "0.4"
+serde_json = "1.0"
+serde_test = "1"
+serial_test = "3.0"
+thiserror = "1.0"
+tracing = "0.1.40"
+tracing-subscriber = "0.3"
+url = "2.2"
+zerofrom = "0.1.5"
+
+[workspace.dependencies.async-stream]
+version = "0.3.5"
+
+[workspace.dependencies.azure_core]
+default-features = false
+features = [
+    "reqwest",
+    "reqwest_deflate",
+    "reqwest_gzip",
+]
+path = "sdk/core/azure_core"
+version = "0.23.0"
+
+[workspace.dependencies.azure_core_amqp]
+path = "sdk/core/azure_core_amqp"
+version = "0.2.0"
+
+[workspace.dependencies.azure_core_test]
+path = "sdk/core/azure_core_test"
+
+[workspace.dependencies.azure_core_test_macros]
+path = "sdk/core/azure_core_test_macros"
+
+[workspace.dependencies.azure_identity]
+path = "sdk/identity/azure_identity"
+
+[workspace.dependencies.clap]
+features = ["derive"]
+version = "4.4.16"
+
+[workspace.dependencies.fe2o3-amqp]
+features = [
+    "native-tls",
+    "tracing",
+    "uuid",
+]
+version = "0.12"
+
+[workspace.dependencies.fe2o3-amqp-cbs]
+version = "0.12"
+
+[workspace.dependencies.fe2o3-amqp-ext]
+version = "0.12"
+
+[workspace.dependencies.fe2o3-amqp-management]
+features = ["tracing"]
+version = "0.12"
+
+[workspace.dependencies.fe2o3-amqp-types]
+version = "0.12"
+
+[workspace.dependencies.getrandom]
+features = ["js"]
+version = "0.2"
+
+[workspace.dependencies.hmac]
+version = "0.12"
+
+[workspace.dependencies.oauth2]
+default-features = false
+version = "5.0.0"
+
+[workspace.dependencies.openssl]
+version = "0.10.70"
+
+[workspace.dependencies.quick-xml]
+features = [
+    "serialize",
+    "serde-types",
+]
+version = "0.31"
+
+[workspace.dependencies.reqwest]
+default-features = false
+features = [
+    "json",
+    "stream",
+]
+version = "0.12"
+
+[workspace.dependencies.serde]
+features = ["derive"]
+version = "1.0"
+
+[workspace.dependencies.serde_amqp]
+features = ["uuid"]
+version = "0.12"
+
+[workspace.dependencies.serde_bytes]
+version = "0.11"
+
+[workspace.dependencies.sha2]
+version = "0.10"
+
+[workspace.dependencies.syn]
+features = ["full"]
+version = "2.0.87"
+
+[workspace.dependencies.tar]
+default-features = false
+version = "0.4.44"
+
+[workspace.dependencies.time]
+features = [
+    "serde-well-known",
+    "macros",
+    "wasm-bindgen",
+]
+version = "0.3.10"
+
+[workspace.dependencies.tokio]
+default-features = false
+features = [
+    "macros",
+    "time",
+]
+version = "1.0"
 
 [workspace.dependencies.typespec]
 default-features = false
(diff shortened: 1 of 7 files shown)


eng/scripts/Update-PackageVersion.ps1
```diff
@@ -66,6 +66,40 @@ if ($packageSemVer.HasValidPrereleaseLabel() -ne $true) {
   exit 1
 }
 
+function ReplacePathDependencyVersions() {
+  $cargoTomlFiles = Get-ChildItem -Path . -Recurse -Filter Cargo.toml
+  | Where-Object FullName -NotMatch "$RepoRoot/(target|eng)/*"
```
heaths, line 71: You should also exclude "tests" and "examples" directories. We already have one under "tests" i.e., `sdk/typespec/typespec_macros/tests/data/compilation-tests/Cargo.toml`.

heaths: We might have some for tests or full examples.

eng/scripts/Update-PackageVersion.ps1
```diff
@@ -66,6 +66,40 @@ if ($packageSemVer.HasValidPrereleaseLabel() -ne $true) {
...
+  foreach ($cargoTomlFile in $cargoTomlFiles) {
+    $content = Get-Content -Path $cargoTomlFile.FullName -Raw
+    $updated = $content
+
+    if ($content -match "\n\s*\[(.+[\.-])?dependencies\.$PackageName\][^\[]+") {
+      # if there is a dependency block containing a path and version for the package, update the version
+      $dependencyBlock = $matches[0]
+      if ($dependencyBlock -match '\n\s*path\s*=') {
+        $replacement = $dependencyBlock -replace '(\n\s*version\s*=\s*)".*"', "`$1`"$packageSemVer`""
+        $updated = $updated.Replace($dependencyBlock, $replacement)
```
heaths, line 82: Don't Repeat Yourself (DRY): you could return either as the line e.g.,

```powershell
$dependency = if ($content -match ... -and $matches[0] -match '\n...') {
  $matches[0]
elseif ($content -match ... -and $matches[0] -match '[...') {
  $matches[0]
}
if ($dependency) {
  $replacement = ...
}
```

eng/scripts/Update-PackageVersion.ps1
```diff
@@ -66,6 +66,40 @@ if ($packageSemVer.HasValidPrereleaseLabel() -ne $true) {
...
 
+function ReplacePathDependencyVersions() {
+  $cargoTomlFiles = Get-ChildItem -Path . -Recurse -Filter Cargo.toml
+  | Where-Object FullName -NotMatch "$RepoRoot/(target|eng)/*"
+
+  foreach ($cargoTomlFile in $cargoTomlFiles) {
+    $content = Get-Content -Path $cargoTomlFile.FullName -Raw
+    $updated = $content
+
+    if ($content -match "\n\s*\[(.+[\.-])?dependencies\.$PackageName\][^\[]+") {
```
heaths, line 77: .NET regex doesn't span lines by default. Is that different in PowerShell's usage of it?

heaths (approved):
If it works, I see nothing blocking, but I raised a doubt based on .NET's regex default behavior.

eng/scripts/Update-PackageVersion.ps1
hallipr: Would any test or example depend on a path and version to a package?

hallipr: cargo is very fragile in regards to the correctness of dependencies.  If a dependency version in any package in the workspace doesn't agree with the version at a path, we can't even run `cargo metadata` to discover the dependencies.

heaths: > Would any test or example depend on a path and version to a package?

They implicitly depend on the crate they are in, but should only ever have a path *or* version otherwise.

eng/scripts/Update-PackageVersion.ps1
hallipr: The default is `multiline: false, singleline: false`, so `^` would only apply to the start of the whole string, `.` wouldn't match `\n` and `\s` will match `\n`. Any "not" pattern like `[^x]` will also match `\n`

My pattern was originally `"(?m)^\s*\[(.+[\.-])?dependencies\.$PackageName\][^\[]+"`, so:
  - enable multi-line
  - find any `[dependency.$PackageName]` block (including `dev-`, `workspace.`, etc) that follows only whitespace after the beginning of any line
  - include all characters after `$PackageName]` that aren't `[`

because `\s` matches new lines, the only functional difference between the current pattern and the `^` pattern is that my new pattern wouldn't match the block if it was the first line in the file.

I think it's more straight forward using explicit newline characters instead of `^` to get at new lines.

heaths: But that's what doesn't make sense: you have `...dependencies\.$PackageName\][^\[]+` which is looking for the end of something like `[dependencies.azure_core]` and then anything else but `[` but that wouldn't expect on the very next line. In addition, you're matching against both `\npath` and `\nversion` found in the `dependencyBlock` which, if I'm reading this right, seems like you're expecting multiple lines.

Did you run this? Is it producing the right results?

eng/scripts/Update-PackageVersion.ps1
hallipr: Updated to use rust and toml_edit,

eng/scripts/Update-PackageVersion.ps1
hallipr: It was, but updated to use rust and toml_edit instead of regex

Cargo.toml
```diff
@@ -62,14 +62,17 @@ path = "sdk/core/azure_core_amqp"
 [workspace.dependencies.azure_core_test]
 # azure_core_test is not published and only ever a dev-dependency
 path = "sdk/core/azure_core_test"
+version = "0.1.0"
 
 [workspace.dependencies.azure_core_test_macros]
 # azure_core_test_macros is not published and only ever a dev-dependency
 path = "sdk/core/azure_core_test_macros"
+version = "0.1.0"
```
heaths: Is this necessary for these two? They should never be published - and are authored as such - but not having a version helps doubly make sure they don't. If it simplifies things, though, no worries. I intentionally didn't add a `version` for that reason, though. They should only ever be `path` dependencies (or inherited, which equates to `path`).

heaths: We should consult the `publish` for each crate, but if that's more work maybe we just exempt these two explicitly with a comment in the cargo script as to why. One less thing we have to worry about updating.

Cargo.toml
```diff
@@ -62,14 +62,17 @@ path = "sdk/core/azure_core_amqp"
...
 
 [workspace.dependencies.azure_identity]
 # azure_identity should only ever be in dev-dependencies herein
 path = "sdk/identity/azure_identity"
+version = "0.23.0"
```
heaths: *Ideally*, this is similar to above: this should only ever be a dev-dependency and, ergo, a `path` dependency. A `version` shouldn't be required and it's one less thing to keep up to date. Maybe we exempt this one as well.

eng/scripts/update-pathversions.rs
```diff
@@ -10,14 +10,24 @@ toml_edit = "^0.22"
 ---
 
 
+
```
heaths: Extreme nit: so

much



whitespace.

eng/scripts/update-pathversions.rs
```diff
@@ -10,14 +10,24 @@ toml_edit = "^0.22"
...
 use regex::Regex;
 use std::io::Write;
 use std::{env, error::Error, fs, path::PathBuf};
 use toml_edit::{value, DocumentMut, Item, Table};
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let script_root = PathBuf::from(env::var("CARGO_MANIFEST_DIR")?);
-    let repo_root = script_root.join("../../..").canonicalize()?;
+    let mode = env::args().nth(1).unwrap_or("update".to_string());
```
heaths, line 20: Too much magic, IMO. Like most CLIs, I'd rather we fail and make them pass explicit verbs. Just `expect("requires 'add' or 'update' command")` or something instead of `unwrap_or(...)`.

eng/scripts/update-pathversions.rs
```diff
@@ -10,14 +10,24 @@ toml_edit = "^0.22"
...
+
+    // in update mode, ignore any depdendencies that don't already have a version
+    // add mode will not add versions to packages that are `publish = false` or to dev-dependencies
+    if mode != "add" && mode != "update" {
+        panic!("Invalid mode. Use 'add' or 'update'.")
+    }
+
+    //let script_root = PathBuf::from(env::var("CARGO_MANIFEST_DIR")?);
+    //let repo_root = script_root.join("../../..").canonicalize()?;
+    let repo_root = PathBuf::from("/home/pahallis/repos/azure-sdk-for-rust").canonicalize()?;
```
heaths: Me nor the hosted agents are going to have this directory. Recommend you look up the tree until you find a file system item (don't check if it's a file or dir) for ".git". See my `update_cratenames.rs` script - the function at the bottom. Can look for `Cargo.toml` as well. Could see if the `file!()` macro works in a cargo script to get the script file itself, but it might be relocated. I've never checked.

eng/scripts/update-pathversions.rs
```diff
@@ -30,11 +40,19 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
 
     for (path, mut document) in toml_files {
-        update_package_versions(document.as_table_mut(), &package_versions);
+        let is_publishable_package = document
+            .get("package")
+            .and_then(Item::as_table)
+            .and_then(|table| table.get("publish").and_then(Item::as_bool))
+            .unwrap_or(true);
+
+        let should_add = is_publishable_package && mode == "add";
```
heaths, line 48: Something is wrong if this added the `version`s above.

heaths (changes requested):
Mostly LGTM, but you'll need to resolve the workspace and/or git root differently; also, I think something in your algorithm is off if you updated the *test crates that have `publish = false` to add `version`s. Nothing obvious I see, but I'd like to review again once you solve that and the workspace lookup. The others are mostly nits or suggestions.

Cargo.toml
hallipr: We should exclude those.  I have it skipping updates to Cargo.tomls for packages with `publish = false`, but I should also exclude those packages from contributing versions for update

Cargo.toml
hallipr: `azure_identity` is a published package and it's usage as "only dev-dependency" wouldn't be known when looking at updates to make to `workspace.dependencies`

Adding a version where it didn't already exist would only happen in `add` mode where the output wouldn't be committed to the repo.

eng/scripts/update-pathversions.rs
hallipr: I didn't even notice that

Cargo.toml
hallipr: I've pushed an update where we only pull package versions from tomls with `[package]` and not `publish = false`

eng/scripts/update-pathversions.rs
hallipr: 😬I really thought I had pulled my debug stuff out.  fixed

hallipr: CARGO_MANIFEST_DIR is the folder containing the script

eng/scripts/update-pathversions.rs
hallipr: fixed

Cargo.toml
heaths: Why not hardcode the exemption, then. It's here only as a dependency and since it should never be anything more than a (transitive) dev-dependency, that's *all* this entry is for.

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -145,6 +149,8 @@ stages:
               CommitMsg: "Increment package version after release of ${{ artifact.name }}"
               PRTitle: "Increment versions for ${{parameters.ServiceDirectory}} releases"
               CloseAfterOpenForTesting: '${{parameters.TestPipeline}}'
+              ${{ if startsWith(variables['Build.SourceBranch'], 'refs/pull/') }}:
```
heaths, line 152: I thought PRs happen in a `refs/merge` prefix?

eng/scripts/Publish-Crates.ps1
```diff
@@ -14,8 +14,8 @@ foreach ($crateName in $CrateNames) {
   Write-Host "Publishing packae: '$crateName'"
   $manifestPath = "$PackagesPath/$crateName/Cargo.toml"
   # https://doc.rust-lang.org/cargo/reference/registry-web-api.html#publish
-  Write-Host "> cargo publish --locked --manifest-path `"$manifestPath`""
-  cargo publish --locked --manifest-path $manifestPath
+  Write-Host "> cargo publish --manifest-path `"$manifestPath`""
+  cargo publish --manifest-path $manifestPath
```
heaths, line 18: In case it matters, these commands aren't the same. You quote `$manifestPath` in one but not the other. Best to be consistent in case it ever does matter and we're confused one way or the other. I think it's safe *not* to quote.

eng/scripts/update-pathversions.rs
heaths, line 1: Make sure to run `chmod +x` on this file. Looks like it might not have the right mask, or just isn't showing it in the web UI but I know I've seen it before.

heaths (approved):
A few changes, but with those I see nothing else blocking.

eng/scripts/update-pathversions.rs
hallipr: It is indeed.  If you click "View File" on the right, the header will show Executable File if `+x`
image

heaths: If you're on Windows, see <https://stackoverflow.com/a/13593391/462376> for how to do it.

Cargo.toml
hallipr: hard coded

eng/pipelines/templates/stages/archetype-rust-release.yml
hallipr: pr refs look like  `refs/pull/1234/(head|merge)`

eng/scripts/Publish-Crates.ps1
hallipr: Powershell doesn't need quotes to pass the arg to cargo publish, but if someone doesn't have the string already in a variable, they may need quotes to make the same call.

I want someone to be able to copy paste the `>` line from the build log to repro an error, so I quote the string variables, but only in the "copy paste this to repro" log line

PR #2343, opened 2025-03-17
Add impl TryFrom<serde_json::Value> for RequestContent<serde_json::Value>

jhendrixMSFT:
Simplifies creating request bodies that are raw JSON.

```rust
json!({"name": "name", "desc": "desc"}).try_into().unwrap(),
```


heaths:
How is that simpler than doing the same thing from a raw string?

```rust
r#"{"name": "json"}"#.try_from()?
```

jhendrixMSFT:
Ohh was not aware of this, that works nicely.

PR #2344, opened 2025-03-17
Default to live tests for only scheduled runs

hallipr:


sdk/keyvault/ci.yml
```diff
@@ -1,4 +1,9 @@
 # NOTE: Please refer to https://aka.ms/azsdk/engsys/ci-yaml before editing this file.
+parameters:
+- name: RunLiveTests
+  displayName: 'Run live tests'
+  type: boolean
+  default: false
```
heaths, line 6: Why is this defined here or in any service directory? It should be in the template. All services should run live tests (if they have any).

heaths (changes requested):
This doesn't seem like what we discussed over Teams.

sdk/keyvault/ci.yml
hallipr: It can't be surfaced to users running the pipeline manually if it's not in the outermost pipeline yaml:

image

Also, I thought we said that all services should run live tests for scheduled runs and should enable live tests for manual runs when queuing the build:

```yaml
RunLiveTests: ${{ or(parameters.RunLiveTests, eq(variables['Build.Reason'], 'Schedule')) }}
```

heaths: I don't see this in any other language and they work the way I think makes sense. `net - keyvault - tests` run live tests by default. I can queue it manually as well. `sdk/keyvault/ci.yml` does not have a variable defined to do it.

The minimal amount of authoring *required* should be in service directories' `ci.yml` files or people are going to screw things up. We should dictate the default behavior, not make them opt into it or even knowing they have to opt into it.

hallipr: we don't have `rust - keyvault - tests`  because we use unified (ci + test) pipelines, so [go - azsecrets](https://dev.azure.com/azure-sdk/internal/_build?definitionId=6440) would be the analog here.

It's a single pipeline for CI, Nightly, Manual and Release
image

To manually run tests on a branch, just queue the single pipeline, change the ref to build and don't approve the release stages

hallipr: What we discussed was being able to turn off live tests for release runs.  Because a release run is just a manual invocation of the single pipeline, there's no way to know, without a queue time parameter, if someone is queuing it because they want to release or queuing it because they want to run live tests.

heaths: Discuss offline.

heaths (approved):

PR #2346, opened 2025-03-18
Refactor azure_core to be more like typespec_client_core

### LINKED ISSUE #1909: Move `Pager` to `azure_core` since it's Azure-specific
As came up during a REST API LT meeting, paging is Azure-specific and the `Pager<T>` should be moved back into `azure_core`. Will have to see how that impacts the overall response model i.e., does it try to create any pager when deserializing? I'm pretty sure not, but use it as an example of things to check. /cc @jhendrixMSFT @JeffreyRichter

### LINKED ISSUE #2181: Refactor all the types in azure_core to match typespec_client_core
When refactoring most of the general types into `typespec` and `typespec_client_core`, I did my best to keep `azure_core` looking like it did to avoid an even bigger diff with all the changes to imports in other crates. All those other crates also helped double check my progress. Moving those types around became a lower priority.

After beta.1, I'd like to move those types around to match `typespec_client_core`. I'm pretty happy with the layout - though, open to suggestions otherwise - and I think re-exporting from `azure_core` that the emitter uses makes sense. Azure service partners will interface entirely with `azure_core` - which does and will continue to have some Azure-specific types like pagers and pollers - and it gives us flexibility as well.

This will no doubt require changes in the emitter, but my hope is that few to no type names will need to change - just their module paths.

heaths:
Resolves #1909 and resolves #2181


heaths:
@jhendrixMSFT @RickWinter @LarryOsterman could I get a code review on this? Ignore `generated` folders. Nothing hand-changed. The rest is mostly moves. We need to get this in ASAP because the emitter is otherwise broken right now without this change.

jhendrixMSFT (approved):

PR #2349, opened 2025-03-19
Use `-Zpackage-workspace` for packing

hallipr:
- Add `target` to the workspace exclusions

`Pack-Crates.ps1`
- Use the staged `.crate` file instead of the unpacked package folder in `/target` after a `cargo publish --dry-run`
- Use `+nightly -Zpackage-workspace` instead of a vendor folder and powershell to order and stage unpublished dependencies
- Use `--no-verify` unless enabled with the `-Verify` switch
- Be more resilient to powershell Array / Single Item return values

`Publish-Crates.ps1`
- Add support for `--dry-run` in Publish-Crates.ps1 via the `-DryRun` switch
- Only pull owners list if adding additional owners
- Use `--no-verify` unless enabled with the `-Verify` switch


Cargo.toml
```diff
@@ -20,6 +20,7 @@ members = [
 ]
 exclude = [
   "eng/scripts",
+  "target",
```
heaths, line 23: This is always ignored. Why did you add it? This is going to raise a lot of eyebrows.

eng/scripts/Pack-Crates.ps1
```diff
@@ -126,26 +126,65 @@ function Get-PackagesToBuild() {
...
-  Invoke-LoggedCommand "cargo vendor $path" -GroupOutput | Out-Host
-  return $path
-}
+# https://doc.rust-lang.org/cargo/reference/registry-web-api.html#publish
+# https://github.com/rust-lang/cargo/blob/5c87c14f9a162daf10d4133fdaab35c72d67b018/crates/crates-io/lib.rs#L42
+function Get-ApiMetadata($package) {
+  $packagePath = Split-Path -Path $package.manifest_path -Parent
+  $readmePath = $package.readme ? (Join-Path -Path $packagePath -ChildPath $package.readme) : $null
+
+  $jsonBody = [ordered]@{
```
heaths, line 135: Why are you copying into exactly the same structure - deps included? Why spend the time and hassle of maintaining code that ends up being the same as just returning the JSON object you want? If you could post process certain fields like `readme` a lot cheaper.

eng/scripts/Pack-Crates.ps1
```diff
@@ -126,26 +126,65 @@ function Get-PackagesToBuild() {
...
+  $metadataLengthBytes = [BitConverter]::GetBytes([UInt32]$metadataBytes.Length)
+  $crateBytes = [IO.File]::ReadAllBytes($crateFilePath)
+  $crateLengthBytes = [BitConverter]::GetBytes([UInt32]$crateBytes.Length)
 
-  #write an empty checksum file
-  '{"files":{}}' | Out-File -FilePath "$LocalRegistryPath/$packageName-$packageVersion/.cargo-checksum.json" -Encoding utf8
+  $bytes += $metadataLengthBytes + $metadataBytes + $crateLengthBytes + $crateBytes
+
+  return $bytes
 }
```
heaths, line 185: Why not just use `Get-FileHash`? It's been around forever. Best not to reinvent the wheel and own its maintenance.

eng/scripts/Pack-Crates.ps1
```diff
@@ -158,61 +197,77 @@ function Create-ApiViewFile($package) {
...
+  }
 
-    Invoke-LoggedCommand -Command $command -GroupOutput
+  Invoke-LoggedCommand -Command $command -GroupOutput
 
+  if ($env:SYSTEM_DEBUG -eq 'true') {
+    Write-Host "##[group] $RepoRoot/Cargo.lock"
+    Get-Content "$RepoRoot/Cargo.lock"
+    Write-Host "##[endgroup]"
+  }
```
heaths, line 228: This shouldn't be necessary, but if you were going to compare before and after, why not do that in the script and not make the human save to files just to diff? Back up the `Cargo.lock` from before, do the new, and emit `git diff -- file1 file2` or something.

heaths (changes requested):
What's not clear from the PR description is *why*. Why are we doing all this work - often redoing what `cargo pack` and/or `cargo publish` would I'm really not crazy about the cost of owning all this and keeping it in sync with any changes that the `cargo` team might make. They don't really document what's in a `.crate` file which means it could change and all our stuff breaks.

I'd really like to understand *why* we're doing all these changes.

Cargo.toml
hallipr: it's not ignored 😬.  The targets folder will get package folder in it.  If you run any `cargo` commands on the packages in /targets, it complains about the package thinking its running in a workspace but not being a member or excluded

heaths: Why are you running commands against anything `targets`? That's not our place space. For package analysis - though I maintain there's nothing to analyze (it's zipped source) - that should probably be done outside the source repo in the temp working dir AzDO, GHA, etc. give us. Let's not unnecessarily complicate things.

eng/scripts/Pack-Crates.ps1
hallipr: They're not the same contract.  One is the metadata handled by cargo internally, the other is the json contract of the crates.io api.  They're just really similar.  I wouldn't want to deal with cargo adding properties that the API will complain about.

Currently
dependencies -> deps
readme -> readme_file
contents(readme) -> readme
version -> vers

dependencies:
version_req -> req
rename -> explicit_name_in_toml

eng/scripts/Pack-Crates.ps1
hallipr: this isn't computing a hash.  Its constructing a binary file for posting to the crates.io api

eng/scripts/Pack-Crates.ps1
hallipr: I actually want to see the contents of the lock file without having to download anything, not just the diff

eng/scripts/Pack-Crates.ps1
heaths: ...which comes back to my original question: why are we doing that? I get why we crack open `.nupkg` files: there's binaries to check for signing, viruses, etc. `.crate` files are zipped source. We spending so much time on redoing what `cargo publish` already does which not only prevents some other work from getting done, but I worry creates a dependency on undocumented implementation details we'll forever have to maintain ourselves.

hallipr:
> I'd really like to understand _why_ we're doing all these changes.

@epage said that we can only rely on the crate file in the workspace and not the package folder

https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/Publishing.20crates.20from.20a.20monorepo.20at.20different.20periods/near/502267339
> Note that any files in target/ made by cargo publrsh are most likely considered incidental and not subject to compatibility guarentees. cargo package would only guarentee .crate. We are planning to move the uncompressed folder.

So our release process needs to be .crate centric.  Also, by running `cargo publish` on top of the pre-normalized `cargo.toml` inside the crate file, we end up with a normalized `cargo.toml.orig` file instead of a normalized `cargo.toml` and normal looking `cargo.toml.orig`

hallipr:
Also, it feels like maintaining the API call will be less maintenance that all of the powershell we have around orchestrating the dependencies during build.

heaths:
All the powershell around dependencies was necessary, though, because monorepos aren't well-supported by cargo. I still don't understand the point of us pulling apart a `.crate` file to "analyze" it, reconstruct it, and call crates.io APIs directly ourselves when `cargo publish` is right there.

What are we even analyzing for a zipped source archive? Anything we could scan for we should against the source itself during PRs, CIs, etc. All that can already be done and not at the last minute when we're trying to publish.

/cc @RickWinter 

I think we should leave this PR as-is for now while we discuss. I know there are some other high-pri EngSys work items we need to address yet e.g., caching, whether Azure Pipelines' caching is good enough or if we should use sccache.

hallipr:
> when `cargo publish` is right there.

cargo publish is only right there if we publish an uncompressed folder.  We only get an uncompressed folder by checking out the repo or by decompressing the .tgz file. We can't checkout the repo in the deployment job.

hallipr:
Because we use workspace references, out package folders can't be published in isolation.   The contents of the tarball are publishable in isolation and the only effect of "double publishing" the tarball is the mangling of the Cargo.toml.orig file.

heaths:
Until we can get toolset changes, I think we need to seek an exemption. That's not how `cargo` works and we're relying solely on *observed behavior* to make critical publishing decisions.

hallipr:
I'm documenting all of the current dependency version and path behaviors in issue #2475 "Document dependency requirements".

PR #2355, opened 2025-03-20
Add manual approval before yanking test pipeline crates

hallipr:


eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -153,9 +153,24 @@ stages:
...
 
         - ${{ if eq(parameters.TestPipeline, true) }}:
+          - job: ManualApproval
+            displayName: "Manual approval"
+            dependsOn: PublishPackage
+            condition: ne(variables['Skip.PublishPackage'], 'true')
+            pool: server
+            timeoutInMinutes: 120 # 2 hours
+            steps:
+            - task: ManualValidation@1
```
heaths, line 163: I take it this is the normal way we use and not moving between environments.

heaths (approved):

eng/pipelines/templates/stages/archetype-rust-release.yml
hallipr: This is new.  We don't have any pipelines that use Manual validation yet.  It just lets us put a pause in a pipeline until someone clicks "Resume".  It's not environment based and I'm not requiring any group membership.

PR #2356, opened 2025-03-20
[API View] Pin nightly for rustdoc JSON output

HarshaNalluru:
- Added a constant `NIGHTLY_TOOLCHAIN` to pin a specific nightly Rust version 
   - included comments to explain its purpose and ensure compatibility with the `rust-api-parser` tool and `rustdoc_types` crate.
- Modified the `cargo +nightly rustdoc ...` command to use the `NIGHTLY_TOOLCHAIN` constant instead of a hardcoded nightly version.
- Updated the `version compatibility` section in the readme to include detailed steps for updating the nightly toolchain and related dependencies, ensuring compatibility across components.


RickWinter (commented):
LGTM

eng/tools/generate_api_report/src/main.rs
```diff
@@ -19,7 +25,7 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     // Call cargo +nightly rustdoc to generate the JSON file
     let output = Command::new("cargo")
-        .arg("+nightly")
+        .arg(NIGHTLY_TOOLCHAIN)
```
heaths, line 28: This will err if not installed. The proper way to do this is put a `rust-toolchain.toml` file next to `Cargo.toml` and make sure that pipelines run `rustup install` with the PWD set to this directory like we do in the root.

```toml
[toolchain]
channel = "nightly-2025-01-12"
```

heaths (changes requested):

eng/tools/generate_api_report/src/main.rs
heaths: This works for installing the toolchain as well as using it in commands by default, so you shouldn't even have this line or the constant because it can be out of sync. Just use a `rust-toolchain.toml` file.

HarshaNalluru: I just checked; it does not error out. Here is a run for instance, if the specified version is not found, it seems to install and then run the code.
image

> Regarding the suggestion to place the rust-toolchain.toml file next to Cargo.toml

I don't prefer pinning the nightly version repo-wide. It limits us from getting all new latest features everywhere else in the repo.

> As for the constant potentially being out of sync

As I mentioned in the email thread, I am fine with it going out of sync because we want to make less frequent updates to this tool. This approach will help us support fewer `FORMAT_VERSION`s in the `azure-sdk-tools/../rust-api-parser`, reducing maintenance overhead at the parser level in the `-tools` repo.

HarshaNalluru: I was discussing this @hallipr as part of the rust CI, and he suggested installation as part of the CI, a separate step/job.
So, we'll have a global variable for this nightly version, which would be installed by the time this code is run.

eng/tools/generate_api_report/src/main.rs
```diff
@@ -6,7 +6,26 @@ use std::io::prelude::*;
...
+// 2. The rustdoc_types crate version used in this tool (0.33.0)
+// When updating this version, ensure rustdoc_types dependency and the rust-api-parser tool are also updated.
+fn get_nightly_toolchain() -> Result<String, Box<dyn Error>> {
+    // Read the rust.yml file and extract the toolchain version
+    let yml_content = std::fs::read_to_string("eng/pipelines/templates/variables/rust.yml")?;
+
+    // Extract the value from the NIGHTLY_TOOLCHAIN_FOR_APIVIEW line
+    yml_content
+        .lines()
+        .find(|l| l.contains("NIGHTLY_TOOLCHAIN_FOR_APIVIEW"))
```
heaths, line 20: This is not the idiomatic way of doing this. Please use `rust-toolchain.toml` and just add a step that runs `rustup install` with the PWD set to this directory.

This makes sure that both the CI and local dev environments use the same version.

eng/tools/generate_api_report/README.md
```diff
@@ -7,13 +7,25 @@ It uses the following command `cargo +nightly rustdoc -Z unstable-options --outp
 
 ## Version Compatibility
 
-When updating the `rustdoc-types` in Cargo.toml, ensure that the `rust-api-parser` in the tools repository is also updated along with this project.
+The project depends on specific version relationships between several components:
 
-1. Check the `FORMAT_VERSION` value in the new version.
-2. Verify that it is compatible with your rustdoc JSON output version (generated by `cargo +nightly rustdoc ...`).
-3. If they do not match, you may need to use a different version of `rustdoc-types`.
+1. **NIGHTLY_TOOLCHAIN**: This value in main.rs must be set to a compatible nightly Rust version that produces the expected format version in the JSON output. Currently, the tool is pinned to a specific nightly (`nightly-2025-01-12`) to ensure consistent output.
```
heaths, line 12: Except it isn't for dev environments. Only for the CI.

heaths (changes requested):

eng/tools/generate_api_report/src/main.rs
hallipr: Why would we use rust-toolchain.toml rather than making a call to UseRust@1 with a toolchain id in pipeline yaml?

We don't make `rustup` calls in ci for anything else.

hallipr: If it was the last `use-rust.yml` call, we can use the pipeline variable `Toolchain` read from the environment variable `TOOLCHAIN` in code:

```rust
let nightly_toolchain = env::var("TOOLCHAIN").expect("Unable to read the TOOLCHAIN environment variable");
```

heaths: @hallipr asked,

> Why would we use rust-toolchain.toml rather than making a call to `steps/use-rust.yml` with a toolchain id in pipeline yaml?

I'm a dev that wants to contribute. I clone, make changes, push and find that I used language/code from a newer version. If a `rust-toolchain.toml` file was there, that wouldn't happen. So, no, only the CI benefits.

> We don't make `rustup` calls from code for anything else.

As I explained above, we don't use a `channel` in the *root* `rust-toolchain.toml` because we're mainly lib crates, which don't typically do that because we want to support a lower MSRV and test across a number of channels. If we pin to a specific version like the MSRV, that makes developing harder.

@HarshaNalluru [stated](https://github.com/Azure/azure-sdk-for-rust/pull/2356#discussion_r2008465603),

> I don't prefer pinning the nightly version repo-wide. It limits us from getting all new latest features everywhere else in the repo.

Nor should you, but I didn't say to change the root `rust-toolchain.toml`. I said to add one to your project. Anyone making changes will then use the version specified therein (based on the `Cargo.toml`) loaded and the CI will use the same version - all for free without any env vars or custom steps. This is literally what I do in almost all my personal Rust project repos. It's easy, it's consistent (local dev, codespaces, pipelines, etc.), and it's idiomatic.

heaths: I'll have a WIP PR to show you, but with your way, go in and run `cargo build --version`. You'll get whatever is the default toolchain for your machine. That's not the version you want to be developing on. That's the consistency I'm talking about.

heaths: If nothing else, instead of inventing a custom way, use `RUSTUP_TOOLCHAIN`: <https://rust-lang.github.io/rustup/overrides.html#overrides>. The CI could set that if needs be. My point is not baking a lot of custom stuff we have to maintain into the CI that only affects the CI. We want devs maintaining this code to use the same tool version to develop it, right? Otherwise they use newer stuff. The doc above explains how to pin versions idiomatically.

eng/tools/generate_api_report/src/main.rs
```diff
@@ -6,7 +6,26 @@ use std::io::prelude::*;
...
+        .and_then(|line| line.split('"').nth(1))
+        .map(|version| format!("+{}", version))
+        .ok_or_else(|| "Could not find NIGHTLY_TOOLCHAIN_FOR_APIVIEW in rust.yml".into())
+}
+
 fn main() -> Result<(), Box<dyn Error>> {
+    let nightly_toolchain = get_nightly_toolchain()?;
+
```
heaths, line 28: Line 36 below is dangerous. You're assuming a PWD that might not be the case. It'll expect a `target` directory in PWD wherever the tool may be run.

How is this tool expected to run? I don't see anything in CI that actually runs it or how. Understanding the invocation is going to be important to know whether the PWD is right, etc. If run as `cargo run`, for example, `./target` will be wrong and `--package {name}` won't work because this project isn't in a workspace.

eng/tools/generate_api_report/src/main.rs
HarshaNalluru: One more question..

I can put a `rust-toolchain.toml` inside the `"eng/tools/generate_api_report"` project and install the desired version programmatically. All good till then.
But, the `cargo +nightly rustdoc ...` command that I'll be running has to be from the root, so it picks up a different active version.

Demonstration below on the different active versions between `"eng/tools/generate_api_report"` project and the root.
image

The solution I was thinking is, I can get the version from `eng/tools/generate_api_report/rust-toolchain.toml` and update the command from `cargo +nightly rustdoc ...` to  `cargo +{version from .toml} rustdoc ...` for it to pick the right version.

eng/tools/generate_api_report/src/main.rs
HarshaNalluru: Assumption is that the invocation is from the root.
some more context and a question - https://github.com/Azure/azure-sdk-for-rust/pull/2356/files#r2012970544

eng/tools/generate_api_report/src/main.rs
heaths: See how I solved this (pushed commit). It pins it in a way that `rustup`, `cargo`, `rust-analyzer`, etc. will use and "stamps" it in the binary so you already get the right version regardless of the PWD.

eng/tools/generate_api_report/src/main.rs
heaths: Since I solved the PWD issue, probably no need but I would still update the `README.md` to mention that.

eng/pipelines/templates/variables/rust.yml
```diff
@@ -2,3 +2,4 @@ variables:
   RUSTFLAGS: "-Dwarnings"
   RUSTDOCFLAGS: "-Dwarnings"
   RUST_LOG: $[iif(eq(variables['System.Debug'], 'true'), 'debug', '')]
+
```
HarshaNalluru, line 5:
```suggestion
```

heaths (approved):

eng/tools/generate_api_report/src/main.rs
HarshaNalluru: readme has it already, threw in an error check too

eng/tools/generate_api_report/README.md
HarshaNalluru: works for everyone now :)

PR #2374, opened 2025-03-21
Optionally disable test tracing colors

heaths:


heaths:
Found out that ANSI color sequences are always written by the `tracing` console subscriber. Since `cargo` et. al. support `NO_COLOR` - and like some other crates did as mentioned in a bug still open against `tracing-subscriber` - I used that. I was writing `2>&1` to a file for @lmolkova when I ran into this and the file was incredibly difficult to read.

sdk/core/azure_core_test/README.md
```diff
@@ -125,6 +125,12 @@ For example, if you wanted to see debug information from all sources by default
 RUST_LOG=debug,test-proxy=trace cargo test
 ```
 
+ANSI colors are written to the terminal even if redirected to a file. To disable writing ANSI color sequences, pass `NO_COLOR=1`:
+
+```bash
+NO_COLOR=1 RUST_LOG=debug,test-proxy=trace cargo test
```
RickWinter, line 131: This change is reasonable. As a seperate issue we need to review all our environment variables. Ideally we'd have them prefixed with `AZ_*` or `AZSDK_*` or something else. This will prevent collisions with other code on the system. Plus it makes it easy to quickly see what settings are set.

RickWinter (approved):

sdk/core/azure_core_test/README.md
heaths: That's intentional. These aren't "out" variables. They are community-common variables, which is why I used them. Why reinvent something new for the same purpose? Besides, these affect behavior only in dev scenarios. E.g., `RUST_LOG` is the default that `log`, `tracing`, and more crates use to control tracing. You can use other env vars, sure, but then a common solution people are used to doesn't work. `NO_COLOR` (and a small handful of others) are ubiquitous in linux/macOS and, as such, even adopted by many CLIs for Windows e.g., the GitHub CLI.

heaths: The idea is if you specify `NO_COLOR=1` on the command line (or otherwise set env vars before starting a process with them), nothing in that expression would use colors. Sadly - and surprisingly - `cargo` doesn't fully honor it. I noticed it's initial text is non-colored, but the test results have some color. Just realized I should file a bug.

PR #2379, opened 2025-03-21
`v0.13.2` emitter version updates to Storage Rust SDK

vincenttran-msft:

1. Updated Storage code to consume `v0.13.2` of the emitter + tsp combo, tsp [here](https://github.com/Azure/azure-rest-api-specs/tree/feature/blob-tsp)
2. Refactors all Storage code to correctly call into generated clients, APIs, and new response models
3. Removes the proof-of-concept models (`blob_properties.rs` and `container_properties.rs`) in favor of the generated response models
4. Added `commit_block_list` to `BlobClient`
5. Added `get_block_list` to `BlobClient`
6. Finalizes the exports to correctly hide any generated models/clients/etc. we don't want customers to export


sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -1,12 +1,12 @@
...
-use azure_core::{base64, headers::HeaderName, Bytes, RequestContent, StatusCode};
+use azure_core::RequestContent;
 use azure_core_test::{recorded, TestContext};
 use azure_storage_blob::{
-    clients::{BlobClient, ContainerClient},
-    models::{BlobType, BlockLookupList},
-    BlobClientOptions,
+    clients::{BlobClient, BlobContainerClient},
+    models::{BlobClientGetPropertiesResultHeaders, LeaseState},
+    BlobClientOptions, BlobContainerClientOptions,
```
vincenttran-msft, line 9: Is this the desired import experience? I think it is a little odd that `BlobClient` and `BlobContainerClient` are gated behind `clients::` while `BlobClientOptions` and `BlobContainerClientOptions` are available at the root.

We can resolve this with:
```rust
pub mod clients {
    mod blob_client;
    mod blob_container_client;
    mod blob_service_client;

    pub use blob_container_client::BlobContainerClient;
    pub use blob_service_client::BlobServiceClient;
}
```

->

```rust
pub mod clients {
    pub(crate) mod blob_client;
    pub(crate) mod blob_container_client;
    pub(crate) mod blob_service_client;
}

pub use clients::blob_client::BlobClient;
pub use clients::blob_container_client::BlobContainerClient;
pub use clients::blob_service_client::BlobServiceClient;
```

1. Add `pub(crate)` to the modules
2. Move the `use` to outside of `clients` module

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -59,11 +63,15 @@ impl BlobClient {
             .per_try_policies
             .push(Arc::new(oauth_token_policy) as Arc<dyn Policy>);
 
-        let client = GeneratedBlobClient::new(endpoint, credential, Some(options))?;
+        let client = GeneratedBlobClient::new(
+            endpoint,
+            credential.clone(),
+            container_name.clone(),
+            blob_name.clone(),
+            Some(options.clone()),
```
jalauzon-msft, line 71: You were not cloning the options before, now you are? Did something change or can we remove the clone?

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -3,71 +3,12 @@
...
-}
-
-pub use crate::generated::clients::{BlobClient, BlobClientOptions};
-// END GENERATED CODE
-
-pub(crate) mod pipeline;
+// Moved from src/generated/mod.rs to surface these specific Options bags to the crate root
+// pub use crate::generated::clients::{
+//     BlobClientOptions, BlobContainerClientOptions, BlobServiceClientOptions,
+// };
```
jalauzon-msft, line 14: Remove commented code?

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -15,348 +19,426 @@ use std::error::Error;
...
+        .upload_blob(
+            RequestContent::from(new_data.to_vec()),
+            false,
+            u64::try_from(new_data.len())?,
+            None,
+        )
+        .await;
+
+    // Assert
+    assert!(error_response.is_err());
```
jalauzon-msft, line 156: Maybe check for specific BlobAlreadyExists error?

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -15,348 +19,426 @@ use std::error::Error;
...
+    // No Option Scenario
+    let response = blob_client.get_blob_properties(None).await?;
 
     // Assert
-    assert_eq!(
-        String::from("HttpResponse(NotFound, \"ContainerNotFound\")"),
-        response.unwrap_err().kind().to_string()
-    );
+    let lease_state = response.lease_state()?;
+    assert_eq!(LeaseState::Available, lease_state.unwrap());
```
jalauzon-msft, line 79: Can we assert a couple more and more useful properties other than lease state? Maybe size and then maybe some ones that need to get serialized like Etag and a date? (obviously can't assert exact values there but just check they seem correct somehow)

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: Unnecessary clone, removed

sdk/storage/azure_storage_blob/src/lib.rs
vincenttran-msft: This will be removed once I merge in the POC changes for the exports.

vincenttran-msft:
This PR is almost ready for review, @heaths Heath, could you advise how to resolve the [issue ](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4713023&view=logs&j=b766ebde-1fdb-5f11-1350-46ddc53b23cf&t=7ef29fa0-f532-5653-3473-16dc04608431&l=697)in "Build Pack Crates" step? It seems to be related to `azure_storage_common` not being found, and I think this was related to the azure-core refactor when Storage was left out? I have since integrated the changes at this commit: https://github.com/Azure/azure-sdk-for-rust/pull/2379/commits/c05d8faf457a11bfd5e984f79f9b41d80ba886cc . Thanks!

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -91,16 +100,10 @@ impl BlobClient {
     /// * `options` - Optional configuration for the request.
     pub async fn get_blob_properties(
```
jhendrixMSFT, line 101: This method name doesn't align with the method options type name.

jalauzon-msft: True, I'm fine if we want to rename this method to `get_properties` to align with generated code.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -15,348 +20,437 @@ use std::error::Error;
...
+
+    let latest_blocks: Vec<Vec<u8>> = vec![block_1_id, block_2_id, block_3_id];
 
     let block_lookup_list = BlockLookupList {
-        committed: None,
-        latest: Some(latest_blocks),
-        uncommitted: None,
+        committed: Vec::new(),
+        latest: latest_blocks,
+        uncommitted: Vec::new(),
```
jalauzon-msft, line 437: Does the "unused" ones not support `None` anymore? Sort of a shame to have to allocate empty vectors here.

sdk/storage/azure_storage_blob/Cargo.toml
```diff
@@ -12,7 +12,7 @@ rust-version.workspace = true
 [dependencies]
 async-trait.workspace = true
 azure_core = { workspace = true, features = ["xml"] }
-azure_storage_common = { version = "0.1.0", path = ".." }
+azure_storage_common = { version = "0.1.0", path = "../azure_storage_common" }
```
jalauzon-msft, line 15: The idea was to have common live the storage root path rather than having a common directory. Why did you add a common folder?

If common is being an issue, let's just remove it entirely for now.

heaths: We have a tool that *unnecessarily* assumes the crate name matches the directory name. What you had isn't technically wrong, but after @hallipr, @vincenttran-msft, and I discussed this, I think it would be better to not fight the status quo on this. All our other language repos do this and when you buck the trend, weird and bad things tend to happen. Better to just follow status quo.

It also doesn't obviate having a common crate. It's still common. Nothing changes for the customer. It's no different than `azure_core` being in `sdk/core/azure_core`. The only reasons `typespec` crate is in `sdk/typespec` are because 1) it's already the crate name and 2) because of Rust RFC 3243, it *will* be a root crate for all other `typespec*` crates beneath it. Your common crate is not a root crate. `azure` would be, which we also already have claim to.

jalauzon-msft: Oh ok, sounds good. Thanks for the explanation. We don't use or plan to release Common yet, so it is also fine to remove it entirely but it's fine for now.

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -2,72 +2,33 @@
...
-        BlobContainerClientRestoreOptions, BlobContainerClientSetAccessPolicyOptions,
-        BlobContainerClientSetMetadataOptions, BlobContainerClientSubmitBatchOptions,
-        BlobPageBlobClientClearPagesOptions, BlobPageBlobClientCopyIncrementalOptions,
-        BlobPageBlobClientCreateOptions, BlobPageBlobClientResizeOptions,
-        BlobPageBlobClientUpdateSequenceNumberOptions, BlobPageBlobClientUploadPagesFromUrlOptions,
-        BlobPageBlobClientUploadPagesOptions, BlobServiceClientFilterBlobsOptions,
-        BlobServiceClientGetAccountInfoOptions, BlobServiceClientGetPropertiesOptions,
-        BlobServiceClientGetStatisticsOptions, BlobServiceClientGetUserDelegationKeyOptions,
-        BlobServiceClientSetPropertiesOptions, BlobServiceClientSubmitBatchOptions,
+    pub use crate::generated::models::{
```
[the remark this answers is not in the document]
heaths: ...or is this a case where the generated clients that use some of those aren't exposed, so the models are useless?

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -2,26 +2,31 @@
...
+        policies::{BearerTokenCredentialPolicy, Policy},
+        RequestContent, Response, Url,
+    },
+    Bytes, Result,
 };
 use std::sync::Arc;
 
 /// A client to interact with a specific Azure storage blob, although that blob may not yet exist.
 pub struct BlobClient {
     endpoint: Url,
```
heaths, line 29: `endpoint` should be a `&str` as they are on all other clients. Parse it into a `Url` internally using the `?` to return an error if that fails.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -110,13 +113,9 @@ impl BlobClient {
     /// * `options` - Optional configuration for the request.
     pub async fn download_blob(
```
heaths, line 114: Same here as @jhendrixMSFT mentioned above. Repeating "blob" in method names on the "BlobClient" feels pretty redundant anyway.

heaths: I'll not repeat, so probably fix up all such violations.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -161,36 +176,31 @@ impl BlobClient {
...
-        content_length: i64,
-        data: RequestContent<Bytes>,
-        options: Option<BlobBlockBlobClientStageBlockOptions<'_>>,
-    ) -> Result<Response<()>> {
-        let block_id = base64::encode(block_id);
-        let response = self
-            .client
-            .get_blob_block_blob_client(self.container_name.clone(), self.blob_name.clone())
-            .stage_block(&block_id, content_length, data, options)
+        block_id: Vec<u8>,
```
heaths, line 179: Why is this a `Vec<u8>`? Is it really a binary ID? If it's a UTF8 string, it should be a `&str` or `String` depending on if you're borrowing below or passing ownership.

On that note, if it's binary, if you're only borrowing all through this call (likely, I imagine), you should use a `&[u8]` which is to `Vec<u8>` that `&str` is to `String`. Only use concrete types like `String` or `Vec` when ownership is passed to us.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: Sounds good, changed all instances of `get_x_properties` -> `get_properties` to align with generated code.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -15,348 +20,437 @@ use std::error::Error;
...
+        )
+        .await;
+
+    // Assert
+    assert!(response.is_err());
     assert_eq!(
-        "21",
-        headers.get_str(&HeaderName::from_static("content-length"))?
+        "HttpResponse(Conflict, \"BlobAlreadyExists\")",
+        response.unwrap_err().kind().to_string()
```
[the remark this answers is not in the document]
heaths: You can deconstruct the members, yes.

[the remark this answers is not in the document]
vincenttran-msft: Sure, changed this to assert against the http status for `Conflict`.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -15,348 +20,437 @@ use std::error::Error;
...
+    let block_2_id: Vec<u8> = b"2".to_vec();
+    let block_3_id: Vec<u8> = b"3".to_vec();
+
     blob_client
-        .upload_blob(
-            RequestContent::from(old_data.to_vec()),
-            false,
-            i64::try_from(old_data.len())?,
+        .stage_block(
+            block_1_id.clone(),
```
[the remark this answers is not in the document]
vincenttran-msft: Later down, we use the same IDs to formulate the `Vec<Vec<u8>> that represents the "latest" blocks in `BlockLookupList` model:
```rust
let latest_blocks: Vec<Vec<u8>> = vec![block_1_id, block_2_id, block_3_id];
let block_lookup_list = BlockLookupList {
        committed: Vec::new(),
        latest: latest_blocks,
        uncommitted: Vec::new(),
    };
```

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: Nope, this is currently in alignment with the `BlockLookupList` model, which takes in `Vec<Vec<u8>> strictly (no `Option` here)

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
jalauzon-msft: Yeah, the block ID supports any binary data ([REST docs](https://learn.microsoft.com/en-us/rest/api/storageservices/put-block?tabs=microsoft-entra-id#uri-parameters)) and it needs to be base64 encoded. When you tell the tsp you want base64 encoded stuff, it changes the option to a `Vec<u8>`. Additionally, as-is we need to own this because the generated code takes a `Vec<u8>` and therefore takes ownership. Whether that is required or not, I'm not sure.

sdk/storage/azure_storage_blob/tests/blob_container_client.rs
```diff
@@ -35,56 +43,42 @@ async fn test_create_container(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
     // Assert
     assert!(response.is_err());
     let error = response.unwrap_err().http_status();
-    assert_eq!(Some(StatusCode::NotFound), error);
+    assert_eq!(StatusCode::NotFound, error.unwrap());
+
+    // Container Exists Scenario
+    container_client.create_container(None).await?;
+    let container_properties = container_client.get_container_properties(None).await?;
+    let lease_state = container_properties.lease_state()?;
```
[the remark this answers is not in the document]
vincenttran-msft: Hi Heath, unfortunately I don't have great answers here to your questions. I was not a part of the discussion regarding this refactor between the previous generated code iteration, I believe this thread was tracking this specific behavior:
https://github.com/Azure/typespec-rust/issues/233 
and so the Storage code here is just following the usage pattern presented by the generated code layer.

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -2,72 +2,33 @@
 //
 // Licensed under the MIT License. See License.txt in the project root for license information.
 // Code generated by Microsoft (R) Rust Code Generator. DO NOT EDIT.
+#![allow(dead_code)]
+#![allow(unused_imports)]
```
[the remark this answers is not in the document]
vincenttran-msft: Yes, I believe the discussion tracking this issue is in the following GitHub issue: https://github.com/Azure/typespec-rust/issues/339 (it was originally listed as a follow-up task, but I striked through to avoid confusion after Joel proposed a [solution](https://github.com/Azure/typespec-rust/issues/339#issuecomment-2770390858)):

> Regarding task "Generated code support to ignore unused warnings", there's no way for the emitter to know if any of the emitted code will be unused. You'll need to control this via `src/lib.rs`. The big hammer would be to add this.
> 
> #![allow(dead_code)]
> #![allow(unused_imports)] // this might not be needed depending on what you're doing

The issue is regarding warnings raised in generated code, not our handwritten Storage clients.

Sample output if you leave out...

#![allow(dead_code)]
```
warning: multiple fields are never read
    --> sdk\storage\azure_storage_blob\src\generated\models\method_options.rs:2518:9
     |
2516 | pub struct PageBlobClientUploadPagesOptions<'a> {
     |            -------------------------------- fields in this struct
2517 |     /// An opaque, globally-unique, client-generated string identifier for the request.
2518 |     pub client_request_id: Option<String>,
     |         ^^^^^^^^^^^^^^^^^
...
2522 |     pub encryption_algorithm: Option<EncryptionAlgorithmType>,
     |         ^^^^^^^^^^^^^^^^^^^^
...
2526 |     pub encryption_key: Option<String>,
     |         ^^^^^^^^^^^^^^
...
2531 |     pub encryption_key_sha256: Option<String>,
     |         ^^^^^^^^^^^^^^^^^^^^^
...
2535 |     pub encryption_scope: Option<String>,
     |         ^^^^^^^^^^^^^^^^
...
...
warning: `azure_storage_blob` (lib) generated 127 warnings
```

#![allow(unused_imports)]
```
warning: unused imports: `AppendBlobClientOptions`, `AppendBlobClient`, `BlobClientOptions`, `BlobClient`, `BlobContainerClientOptions`, `BlobContainerClient`, `BlobServiceClientOptions`, `BlobServiceClient`, `BlockBlobClientOptions`, `BlockBlobClient`, `PageBlobClientOptions`, and `PageBlobClient`
  --> sdk\storage\azure_storage_blob\src\generated\mod.rs:9:5
   |
9  |     AppendBlobClient, AppendBlobClientOptions, BlobClient, BlobClientOptions, BlobContainerClient,
   |     ^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^  ^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^
10 |     BlobContainerClientOptions, BlobServiceClient, BlobServiceClientOptions, BlockBlobClient,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^
11 |     BlockBlobClientOptions, PageBlobClient, PageBlobClientOptions,
   |     ^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: `azure_storage_blob` (lib) generated 1 warning (run `cargo fix --lib -p azure_storage_blob` to apply 1 suggestion)
```
Which I mistakenly tried to remove many times, but as stated by Joel [here](https://github.com/Azure/typespec-rust/issues/326#issuecomment-2748166298):
> ...
> Keep in mind that, by default, SDKs are expected to be 100% generated. So, the default behavior must be to export all public surface area, so things just work.

So I don't know if we have any recourse other than using the allows. But happy to take any changes if we find a sufficient answer without them.

sdk/storage/azure_storage_blob/src/lib.rs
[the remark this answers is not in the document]
vincenttran-msft: This one I don't feel too strongly about, I was under the impression that we were only exporting the models that customers should be concerned with (as currently our code only exposes probably >5% of the models) for response models, options bag models, and request models.
cc: @jalauzon-msft

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: Hi Heath, for this one I originally actually had `String`, but you had told me it should be a `Url` in a previous PR comment: https://github.com/Azure/azure-sdk-for-rust/pull/2014#discussion_r1931216799
> Type should be `azure_core::Url`. See generated clients.

Happy to take this change if `&str` is the actual correct type we want to land on, but will leave it unchanged for now to minimize back and forth.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: Sounds good, I applied this feedback across the whole package.
`upload_blob` -> `upload`
`download_blob` -> `download`
All instances of `get_X_properties` -> `get_properties`

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: Relevant GH issue tracking this: https://github.com/Azure/typespec-rust/issues/296

sdk/storage/azure_storage_blob/src/lib.rs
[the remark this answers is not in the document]
jalauzon-msft: The idea is to only export the models we are currently making use of in our limited set of APIs we are exposing through our hand-written clients. Like Vincent said, there are many many models and option bags we are currently not using so we chose to leave them out of our exports for now to keep the API surface that customers have to look through smaller and more focused.

sdk/storage/azure_storage_blob/tests/blob_client.rs
heaths: @LarryOsterman convinced us - though not to hold up the MVP - that we should go back to `Option<T>` for vecs again because there plenty of cases where it matters - not just JSON merge+patch.

sdk/storage/azure_storage_blob/tests/blob_container_client.rs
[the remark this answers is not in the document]
heaths: Oh, so these are header fields? I didn't know that. In that case, this makes sense since parsing headers can fail. @jhendrixMSFT brought up maybe only returning an `Option<T>` instead of an `azure_core::Result<T>` if no conversion is necessary, but I don't like the inconsistency. I think this is fine as-is.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
heaths: The field, yes, not the parameter. The parameter to clients, as in the guidelines, is a `&str` because `Url::parse()` only needs to borrow a `&str`. It will allocate a `String` internally but return a `Url`, which is what you store as the field. Your other clients do this correctly, IIRC.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
heaths: Ah, so this is the context for what @jhendrixMSFT asked offline. To put it in public writing, the rules about borrowing vs. owning are the same for vecs as well. If we're only borrowing, take/return a `&str` or `&[u8]`. If we're going to own, take/return a `String` or `Vec<u8>`. Both of those concrete types satisfy `&str` and `&[u8]` if you take a ref to them.

heaths: That said, I don't think we need to hold up the MVP for this.

jhendrixMSFT: Opened https://github.com/Azure/typespec-rust/issues/352 to track this.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: Hi Heath, I think everything follows your feedback-
```rust
pub struct BlobClient {
    endpoint: Url,
    client: GeneratedBlobClient,
}
```
Here, the field is of type `Url`

```rust
pub fn new(
        endpoint: &str,
        container_name: String,
        blob_name: String,
        options: Option<BlobClientOptions>,
    ) -> Result<Self> {
```
Here is the `new()` function on `BlobClient`, and `endpoint` is of type `&str`

```rust
 Ok(Self {
            endpoint: endpoint.parse()?,
            client,
        })
```

And here is where I consequently parse the `&str` into a `Url` type.

Please let me know if there are any changes still necessary.

heaths: Shoot, sorry. I misread. These big PRs tend to make me go "cross-eyed" after a while.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -140,7 +141,7 @@ async fn test_upload_blob(ctx: TestContext) -> Result<(), Box<dyn Error>> {
         .await?;
 
     // Assert
-    let response = blob_client.download_blob(None).await?;
+    let response = blob_client.download(None).await?;
     let content_length = response.content_length()?;
```
heaths, line 145: So this is downloading the entire blob? Not to hold up this PR, and it will likely require changes in `azure_core`, but isn't the idea to get a stream and stream to a file or other caller-provided buffer? Or is that a separate client like ya'll did for Go?

I've been thinking about how to support this case since after beta.2 I want to move deserialization *generally* into the client methods, but we need to make concessions for Storage and maybe Cosmos (/cc @analogrelay). It would be good to chat about that soon.

Again, this isn't to hold up this PR nor the MVP but wanted to call it out / ask about it in context.

heaths (approved):

sdk/storage/azure_storage_blob/tests/blob_client.rs
jalauzon-msft: Yes, this is downloading the entire Blob, unless a range is specificed in the options. And yes, we would ideally want customers to be able to Stream the response body somewhere other than memory. Is there a way to do this with the current `ResponseBody` type in Core? For this test, we chose to just collect the bytes since the Blob was small.

Happy to chat about our requirements whenever.

vincenttran-msft: Since this is non-blocking, I will go ahead and resolve in order to merge. Thanks!

jalauzon-msft (approved):

PR #2385, opened 2025-03-25
Remove TokenCredential::clear_cache

### LINKED ISSUE #2381: Remove TokenCredential::clear_cache()
Caching should be an implementation detail of each credential type. `clear_cache()` is misleading because some credentials don't cache anything, and clearing a cache doesn't guarantee the next `get_token` call will authenticate again or return a new token.

chlowell:
Closes #2381


analogrelay (approved):
LGTM from Cosmos DB

heaths (approved):
Maybe a `pub(crate)` implementation-specific concern, but we *might* need them for testing and could handle on a cred-by-cred basis. At least, IIRC, there was some need for it in some libraries for .NET; perhaps, though, we can also solve that with a 1ms lifetime or something like that to force a refresh should we need to test that.

sdk/identity/azure_identity/src/credentials/cache.rs
heaths, line 1: While you're at it, move this up to the root. I want to get rid of "credentials" in this crate eventually. It's pointless. It's all credentials now. The old crate had "flows" as well.

You can drop `pub(crate)` in that case. Just `struct TokenCache(...)` will do because descendent modules have private visibility of their ancestors private members.

sdk/identity/azure_identity/src/credentials/cache.rs
```diff
@@ -0,0 +1,173 @@
...
+use azure_core::credentials::AccessToken;
+use futures::Future;
+use std::collections::HashMap;
+use tracing::trace;
+
+#[derive(Debug)]
+pub(crate) struct TokenCache(RwLock<HashMap<Vec<String>, AccessToken>>);
+
+impl TokenCache {
+    pub(crate) fn new() -> Self {
```
heaths: Tip: the type is already `pub(crate)` (and doesn't need to be; see previous comment). No need to declare these as `pub(crate)` when `pub` will suffice.

sdk/identity/azure_identity/src/chained_token_credential.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+use crate::credentials::cache::TokenCache;
 use crate::TokenCredentialOptions;
```
heaths: Did you type this? I don't recommend that:

1. It's idiomatic to collapse definitions (see below).
2. If you let rust-analyzer do it, you just `Ctrl+.` (default Windows/linux binding) and select "Import". It'll do this automatically:

```suggestion
use crate::{TokenCredentialOptions, credentials::cache::TokenCache};
```

When lines get long/complete enough, rust-analyzer will format automatically based on rules we have in the repo.

heaths (approved):

sdk/identity/azure_identity/src/chained_token_credential.rs
chlowell: This is what we have on main (you're commenting on a revert commit). I'll clean it up in another PR

sdk/identity/azure_identity/src/credentials/cache.rs
chlowell: I agree. I'll do it in another PR, to keep this one focused on removing `clear_cache`

heaths: Oh, I wasn't even thinking we needed to be proactive. But maybe if you had a pattern in mind and wanted to do it for one then we can pattern off that. Whatever works.

sdk/identity/azure_identity/src/chained_token_credential.rs
heaths: Eh, more of a tip, anyway. It's usually a sign of manually typed imports, which is rarely necessary/advised.

heaths:
I don't know what's going on this with this one:

image

heaths:
@weshaggard I temporarily disabled "Require conversation resolution" to merge this. Something wrong on the GitHub backend, it seems. I even tried unresolving and re-resolving each conversation but GitHub still thought a conversation needed resolving, despite it not showing any unresolved conversation. 🥴

PR #2389, opened 2025-03-26
Remove parsing module

heaths:
The parsing module only contained the `FromStringOptional` trait implemented for a few primitives and implemented by our `create_enum` macros. But it's not used, and I could find no use of it even in the `legacy` branch. Maybe it once was, but it really is just `FromStr` now with `FromStr::Err` set to `typespec::error::Error`. No need to keep it, and it just creates problems for re-exporting macros into `azure_core` without necessitating a dependency on `typespec_client_core`.


heaths:
This was getting in the way of some other changes and I can't find how it was ever used. Maybe if I go back into the history of the `legacy` branch far enough I could find it, but in the current branch it's not used at all. Maybe @demoray or @cataggar remember.

Anyway, may as well get rid of it. It seems to serve no purpose and there's no reason to have to maintain it.

jhendrixMSFT (approved):

RickWinter (approved):

PR #2390, opened 2025-03-26
Handle MsiTokenResponse::expires_on being an int or string.

tot0:
Some MSI token servers don't follow the spec and return `expires_on` as an int unix timestamp, rather than the expected string of an int unix timestamp, so to be broadly compatible it's nice to handle both.

### DIFF
--- a/sdk/identity/azure_identity/CHANGELOG.md
+++ b/sdk/identity/azure_identity/CHANGELOG.md
@@ -19,6 +19,8 @@
 
 ### Bugs Fixed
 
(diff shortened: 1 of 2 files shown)


heaths:
@chlowell I will review the PR, but I leave it up to you if you want to take it for Identity. We already have code that handles this (see the `AzureCliCredential`) but there's no uniformity with the older code still in `azure_identity` currently. Because this issue might be more general, we might instead want to have a helper of sorts for this class of problem.

heaths (approved):
Thank you for the contribution!

PR #2393, opened 2025-03-26
[Cosmos] Refactor query responses to their own Pager and Page types

analogrelay:
Cosmos DB queries for Items, Containers, Databases, or Offers, aren't guaranteed to be satisfied through HTTP requests. For example, Direct mode clients, such as the .NET SDK, use a custom TCP protocol to communicate directly with the replicas. In addition, cross-partition requests return "synthetic" pages of data that merge results from several requests (to each partition). In order to support this in the future, I'm changing the return type of the _query related_ APIs from `azure_core::Pager<T>` to a new `azure_data_cosmos::FeedPager<T>` type. Instead of yielding `azure_core::Response` values, it yields `azure_data_cosmos::FeedPage<T>` values. A `FeedPage` is a lot like a `Response`, except for a few things:

1. It doesn't have a status code, since I don't believe the status code is relevant in this case and it may not be an HTTP status code. We can always add it back if necessary.
2. It contains full deserialized items. We could augment it to support some form of streaming if we want in the future.

It does retain the `azure_core::http::Headers` collection, because it's a collection (and thus could just be empty when we don't have any) and because all the Cosmos transports _do_ support some form of headers that we may want to expose to the user.

@heaths The easiest way for me to make this split was to refactor `azure_core::Pager<T>` a bit so that I could use it. I'm not wedded to that design if you have concern. I wanted to be able to use the existing work done on `Pager<T>` but _did not_ want it to yield `azure_core::Response<T>` values. An alternative I considered and am willing to fall back to is to just copy the `Pager<T>` code, rather than using the type itself. I refactored it by creating two separate types:

1. `PageStream<T>` which contains all the paging logic, but yields bare `T` values.
2. `Pager<T>` which is an _alias_ for `PageStream<Response<T>>`, which allows existing usages to work as they did before. In **most** cases, a `Pager<T>` *should* yield a `Response<T>` , so I didn't want to force every other paginated service API to spell out `Pager<Response<T>>` in its return type.


sdk/core/azure_core/src/http/pager.rs
heaths, line 1: I think this would be fine. We want to put fully-deserialized models into responses by default anyway, except in cases were "true streaming" e.g., blob downloads are needed, which we'll provide a way to do that. But in those cases, I can't imagine how a paged response would work so these changes shouldn't adversely affect those scenarios.

@JeffreyRichter @jhendrixMSFT @LarryOsterman would you agree? /cc @vincenttran-msft

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # Release History
 
+## 0.23.0 (Unreleased)
+
+### Features Added
+
+* Decoupled query responses from HTTP, to allow for handling non-HTTP transports for queries. ([#2393](https://github.com/Azure/azure-sdk-for-rust/pull/2393))
```
heaths, line 7: Grammar nit:
```suggestion
* Decoupled query responses from HTTP to allow for handling non-HTTP transports for queries. ([#2393](https://github.com/Azure/azure-sdk-for-rust/pull/2393))
```

heaths (approved):
This looks nice and clean.

JeffreyRichter:
@analogrelay Are you saying that today this client methods make HTTP requests but that in the future, they might not?
If so, I don't love the idea of changing the underlying implementation this drastically on customers. Customers will have an expectation of HTTP with headers, context, perhaps they added a pipeline policy that would no longer function as expected, etc.

Instead, I'd keep these methods always making HTTP requests and then introduce NEW methods (perhaps on a new/different client) that had different (transport) behavior.

analogrelay:
> Are you saying that today this client methods make HTTP requests but that in the future, they might not?

Correct. In fact it often does **not** use HTTP in several other languages. That's why several Cosmos clients use their own "Response" type ([.NET](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.feedresponse-1?view=azure-dotnet) , [Java](https://learn.microsoft.com/en-us/java/api/com.azure.cosmos.models.feedresponse?view=azure-java-stable), [JavaScript](https://learn.microsoft.com/en-us/javascript/api/%40azure/cosmos/feedresponse?view=azure-node-latest)). This change is essentially aligning the Rust SDK with those SDKs.

> If so, I don't love the idea of changing the underlying implementation this drastically on customers. Customers will have an expectation of HTTP with headers, context, perhaps they added a pipeline policy that would no longer function as expected, etc.

Customers should **not** have this expectation; it will cause unexpected surprises for them as we bring the more feature-rich transports[^1] online, and I'd rather make it very clear to customers that queries are NOT guaranteed to move through the HTTP pipeline. When we make HTTP requests, we do use that pipeline (though perhaps, to reinforce that point, we should not use the core pipeline and should instead use a separate pipeline for queries that can work for all transports; I'm not sure on that yet).

Again, it's good to point out here that several of our other clients work this way. HTTP requests **are** made through the "standard" HTTP pipeline and can be observed in policies, but activity on the non-HTTP transport does not.

> Instead, I'd keep these methods always making HTTP requests and then introduce NEW methods (perhaps on a new/different client) that had different (transport) behavior.

This would be much more surprising to users, particularly given the fact that the user is usually not _directly_ aware that their request is spanning multiple HTTP responses OR using a different transport (of course we don't hide it, but the point is they would certainly be surprised to have to make this decision themselves).

---

I think it's also worth noting a key distinction here with Cosmos. Cosmos clients should not really be considered standard wrapper SDKs that call REST APIs. They are more akin to the client driver for a database, like PostgreSQL or SQL Server. Part of the transport surface area uses REST, but a significant portion of that transport surface is non-HTTP (more like EventHubs AMQP support, for example).

I think it's reasonable for us to reuse the Core HTTP primitives for APIs that are _entirely_ REST-based (essentially the "ControlPlane-over-DataPlane" APIs like CRUD for Databases/Containers). But this is the first step in moving the _Data Plane_ operations to be independent of HTTP primitives (I expect to move point read/write operations like Item CRUD to this model as well).

[^1]: The non-HTTP transport is the **only** way for customers to get a guarantee of the latency SLA we publish, for example.

heaths:
@JeffreyRichter it's also worth pointing out that @analogrelay's changes to `azure_core` not only *don't* adversely affect HTTP support, but are in line with how we want to deserialize the vast majority of models within the pipeline anyway. See my comment. In those cases that we do want to provide true stream (as we always do now) e.g., `BlobClient::download_blob`, they wouldn't use paging anyway. I just can't see how that would work.

Pilchie:
In other SDKs where we support other transports, we have a configuration option for which transport to use (for example: https://learn.microsoft.com/azure/cosmos-db/nosql/sdk-connection-modes), so the behavior change you are concerned about only happens in response to a customer code change. What @analogrelay is proposing is to enable changing that connectivity mode without *also* having to change all the query and point operation calls to something different.

JeffreyRichter:
Ok, that does make me feel better. 
I'll say this: I the worked on the IOT SDKs years ago and, in this, customers could set a transport flag between selecting HTTP, AMQP, or MQTT. The end result was disastrous for customers and the SDK team while they tried to abstract away the differences between 3 fundamentally different transports and customers couldn't write code that just worked and is wasn't clear to them the pros/cons of selecting one transport over another. 

So, my past experience scares me with this proposal but it is your SDK and if you're ok with maintaining this SDK and supporting your customers, go ahead.

JeffreyRichter:
I See. I misunderstood then. I thought the plan was to GA with an HTTP implementation and then change the protocol in a subsequent release. Doing this before GA also makes me feel much better.

analogrelay:
I've rebased this and revived it after my week off last week. Looking for sign-off from Cosmos folks (@Pilchie @kirankumarkolli)

heaths (approved):
No new commits.

PR #2398, opened 2025-03-27
Improve error handling in Proxy startup and adjust executable path for Windows

heaths:


heaths:
Tested this out in Windows and it works right now. I intentionally broke the EXE name and proved the error handling is corrected, which would've saved a lot of time. I wasn't consider the `Result<T, E>` return of the async select arm. I was also *replacing* ".TestProxy" with ".exe" when I meant to append it. I guess technically Windows doesn't need to, but `Command` doesn't document whether it creates processes in such a way that it would cause the `%CMDEXT%` env var to be used so this is safer.

PR #2403, opened 2025-03-27
Rename "operation" to "poller"

heaths:


sdk/core/azure_core/src/http/mod.rs
```diff
@@ -5,11 +5,11 @@
 
 pub mod headers;
 mod models;
-pub mod operation;
 mod options;
 mod pager;
 mod pipeline;
 pub mod policies;
+pub mod poller;
```
RickWinter, line 12: The change is fine. I'm curious on Poller vs Operation.

In C++ we use the term Operation, maybe shouldn't have? In go we call them Pollers, maybe shouldn't have?

Is Poller the pattern everywhere? Is it idiomatic to Rust?

RickWinter (approved):

sdk/core/azure_core/src/http/mod.rs
heaths: @LarryOsterman said,

> FWIW, C++ uses "Operation" instead of "Poller", but in all honesty, I think I prefer Poller, especially since it more accurately expresses what is actually happening during a long running operation.

C++ and .NET seem to be the only ones using `Operation`. I thought we agreed on `Poller`.

JeffreyRichter (approved):

PR #2406, opened 2025-03-28
Add `x-ms-client-request-id` pipeline policy

### LINKED ISSUE #2240: Define policy in azure_core for x-ms-client-request-id
Like most (all?) our other SDK languages, define a policy for `x-ms-client-request-id` that client libraries can just include as needed.

_Originally posted by @heaths in https://github.com/Azure/azure-sdk-for-rust/pull/2014#discussion_r1970817350_

heaths:
Resolves #2240. This also supports setting a custom header name and, thus, must be added by each client to the per-call pipeline policies instead of being added automatically.


heaths:
If we move forward with this - or, rather, whatever design we move forward with - I'll add more examples in docs and/or `examples/`.

heaths:
Need to merge #2412 before we can rebase and merge this one or recorded tests will fail.

sdk/core/azure_core/src/http/options/telemetry.rs
```diff
@@ -1,18 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-use typespec_client_core::setters;
-
 /// Telemetry options.
 #[derive(Clone, Debug, Default)]
 pub struct TelemetryOptions {
     /// Optional application ID to telemetry.
```
LarryOsterman, line 7: Do you want to mention that the primary use for application ID is in the user-agent string?

LarryOsterman: This stems from the fact that the "Telemetry" policy really has very little to do with telemetry as people think of telemetry, but instead is a "User-Agent" policy. The "telemetry" policy is actually a different policy :(.

sdk/core/azure_core/src/http/policies/client_request_id.rs
```diff
@@ -0,0 +1,233 @@
...
+impl ClientRequestIdPolicy {
+    /// Creates a new policy using the default `x-ms-client-request-id` header.
+    pub const fn new() -> Self {
+        ClientRequestIdPolicy(headers::CLIENT_REQUEST_ID)
+    }
+
+    /// Creates a new policy using a custom header name.
+    ///
+    /// You can construct a new policy for a constant or static variable.
+    pub const fn with_header_name(header: &'static str) -> Self {
```
LarryOsterman, line 32: DId you ever decide to remove the `with` prefix on builder methods?

LarryOsterman (approved):

PR #2409, opened 2025-03-28
Move managed identity credentials behind ManagedIdentityCredential

chlowell:
Adding `ManagedIdentityCredential` as a wrapper around `AppServiceManagedIdentityCredential` and `VirtualMachineManagedIdentityCredential`, and removing those and supporting types from the public API

### DIFF
--- a/sdk/identity/azure_identity/src/credentials/imds_managed_identity_credentials.rs
+++ b/sdk/identity/azure_identity/src/credentials/imds_managed_identity_credentials.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-use crate::{credentials::cache::TokenCache, TokenCredentialOptions};
+use crate::{credentials::cache::TokenCache, TokenCredentialOptions, UserAssignedId};
 use azure_core::{
     credentials::{AccessToken, Secret, TokenCredential},
     error::{http_response_from_body, Error, ErrorKind},
@@ -29,6 +29,16 @@ pub enum ImdsId {
     ObjectId(String),
     /// The Azure resource ID of the user-assigned identity to be used.
     MsiResId(String),
+}
+
+impl From<UserAssignedId> for ImdsId {
+    fn from(user_assigned_id: UserAssignedId) -> Self {
+        match user_assigned_id {
+            UserAssignedId::ClientId(client_id) => ImdsId::ClientId(client_id),
+            UserAssignedId::ObjectId(object_id) => ImdsId::ObjectId(object_id),
+            UserAssignedId::ResourceId(resource_id) => ImdsId::MsiResId(resource_id),
+        }
+    }
 }
 
 /// Attempts authentication using a managed identity that has been assigned to the deployment environment.
--- a/sdk/identity/azure_identity/src/credentials/imds_managed_identity_credentials.rs
+++ b/sdk/identity/azure_identity/src/credentials/imds_managed_identity_credentials.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-use crate::{credentials::cache::TokenCache, TokenCredentialOptions, UserAssignedId};
+use crate::{credentials::cache::TokenCache, env::Env, TokenCredentialOptions, UserAssignedId};
 use azure_core::{
     credentials::{AccessToken, Secret, TokenCredential},
     error::{http_response_from_body, Error, ErrorKind},
@@ -55,6 +55,7 @@ pub(crate) struct ImdsManagedIdentityCredential {
     secret_env: String,
     id: ImdsId,
     cache: TokenCache,
+    env: Env,
 }
 
 impl ImdsManagedIdentityCredential {
@@ -75,6 +76,7 @@ impl ImdsManagedIdentityCredential {
             secret_env: secret_env.to_owned(),
             id,
             cache: TokenCache::new(),
+            env: options.env().clone(),
         }
     }
 
@@ -100,7 +102,7 @@ impl ImdsManagedIdentityCredential {
 
         req.insert_header("metadata", "true");
 
-        let msi_secret = std::env::var(&self.secret_env);
+        let msi_secret = self.env.var(&self.secret_env);
         if let Ok(val) = msi_secret {
             req.insert_header(self.secret_header.clone(), val);
         };
--- a/sdk/identity/azure_identity/src/credentials/mod.rs
+++ b/sdk/identity/azure_identity/src/credentials/mod.rs
@@ -21,15 +21,15 @@ mod options;
 mod virtual_machine_managed_identity_credential;
 mod workload_identity_credentials;
 
-pub use app_service_managed_identity_credential::*;
+pub(crate) use app_service_managed_identity_credential::*;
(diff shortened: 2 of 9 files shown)


sdk/identity/azure_identity/src/managed_identity_credential.rs
```diff
@@ -0,0 +1,498 @@
...
+    AzureArc,
+    AzureML,
+    AppService,
+    CloudShell,
+    Imds,
+    ServiceFabric,
+}
+
+impl ManagedIdentitySource {
+    pub fn as_str(&self) -> &str {
```
heaths, line 117:
```suggestion
    pub fn as_str(&self) -> &'static str {
```

May improve compiler optimizations.

sdk/identity/azure_identity/src/managed_identity_credential.rs
```diff
@@ -0,0 +1,498 @@
...
+    use std::collections::HashMap;
+    use std::time::{SystemTime, UNIX_EPOCH};
+
+    const EXPIRES_ON: &str = "EXPIRES_ON";
+
+    struct EnvVarGuard {
+        original_vars: HashMap<String, Option<String>>,
+    }
+
+    impl EnvVarGuard {
```
heaths, line 175: `TokenCredentialOptions` already have this, which you should use to get variables at runtime and override at test time. You can pass in the options. This is also safer because we're not actually setting env vars in the process in case tests are ever parallelized (probably won't be, but one never knows).

sdk/identity/azure_identity/src/managed_identity_credential.rs
```diff
@@ -0,0 +1,498 @@
...
+        expected_source: ManagedIdentitySource,
+    ) {
+        let _guard = EnvVarGuard::new(&env_vars);
+        let actual_source = get_source();
+        assert!(
+            std::mem::discriminant(&actual_source) == std::mem::discriminant(&expected_source),
+            "Expected {:?}, got {:?}",
+            expected_source,
+            actual_source
+        );
```
heaths, line 279:
```suggestion
        assert_eq!(std::mem::discriminant(&actual_source), std::mem::discriminant(&expected_source));
```

You can save yourself time by not putting all these "expected a, got b". That's not idiomatic in rust. The default error message is most often good enough. I only use the custom message when the default may not be as helpful.

heaths (changes requested):
Mostly just nits, but I'd like to re-review. I don't advise setting actual env vars, especially when `TokenCredentialOptions` already has a better way. There's also some efficiency notes herein.

sdk/identity/azure_identity/src/managed_identity_credential.rs
```diff
@@ -0,0 +1,498 @@
...
+
+/// Options for constructing a new [`ManagedIdentityCredential`].
+#[derive(Debug, Default)]
+pub struct ManagedIdentityCredentialOptions {
+    /// The [`TokenCredentialOptions`] to use for the credential.
+    pub credential_options: TokenCredentialOptions,
+
+    /// Specifies a user-assigned identity the credential should authenticate.
+    /// When `None`, the credential will authenticate a system-assigned identity, if any.
+    pub user_assigned_id: Option<UserAssignedId>,
```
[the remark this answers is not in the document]
chlowell: I removed it from the public surface on the rationale that system-assigned identity will be the default i.e., what you get when you don't set `user_assigned_id: Some(UserAssignedId::...)`. `SystemAssigned` is then redundant as an option and wouldn't make sense in a `UserAssignedId` enum anyway.

Is this "`Option` in an options bag" pattern alright?

sdk/identity/azure_identity/src/credentials/mod.rs
```diff
@@ -21,15 +21,15 @@ mod options;
 mod virtual_machine_managed_identity_credential;
 mod workload_identity_credentials;
 
-pub use app_service_managed_identity_credential::*;
+pub(crate) use app_service_managed_identity_credential::*;
```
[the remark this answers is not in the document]
chlowell: I did this because `lib.rs` has `pub use credentials::*` and unrelated files import from that module via the crate. I plan to sort all that out and clarify the public API as part of removing the `credentials` module

heaths (approved):

sdk/identity/azure_identity/src/credentials/imds_managed_identity_credentials.rs
```diff
@@ -75,6 +76,7 @@ impl ImdsManagedIdentityCredential {
             secret_env: secret_env.to_owned(),
             id,
             cache: TokenCache::new(),
+            env: options.env().clone(),
```
heaths, line 79: This is already in `TokenCredentialOptions`, though. Can you not just use that?

heaths: I mean, don't declare it when you already have it in the nested options. Just use what's already defined.

chlowell: Please forgive my density here, I'm still not following you. I believe the struct needs a new field because it doesn't currently hold the options. I chose to add the Env to the struct instead of the options because that's a smaller change. Really the right thing to do would be to read environment variables in the constructor and hang on to those values for use in `get_token`, however I don't want to jump into a rewrite yet. With this PR, I just want to put a better API on the current implementation

sdk/identity/azure_identity/src/managed_identity_credential.rs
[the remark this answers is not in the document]
heaths: Yes. In general, all options / fields should be `Option<T>` for the same reason ya'll gophers made everything nilable. :wink:

sdk/identity/azure_identity/src/credentials/imds_managed_identity_credentials.rs
heaths: Ah, sorry. My bad. The way GitHub was rendering the diff and not the whole file, I thought the struct was holding the `options` already.

sdk/identity/azure_identity/src/managed_identity_credential.rs
```diff
@@ -256,32 +220,35 @@ mod tests {
...
+                env,
+                ..Default::default()
+            },
+            ..Default::default()
+        }));
         assert!(
             matches!(result, Err(ref e) if *e.kind() == azure_core::error::ErrorKind::Credential),
-            "Expected constructor error, got: {:?}",
-            result
+            "Expected constructor error"
```
heaths, line 251: Perhaps not in this PR, but generally in Rust, "if it compiles it works". If we're asserting something at runtime that we could otherwise do at compile time e.g., only certain variants are allowed, certain strings are valid, whatnot, that should be modeled as such in code. Obviously some things are out of our control e.g., bad content to deserialize.

heaths (approved):

sdk/identity/azure_identity/src/managed_identity_credential.rs
chlowell: I also prefer build to runtime errors but I don't see a way to shift this one left because we can't know before runtime which managed identity API(s) an application will actually use. For example, keeping `VirtualMachineManagedIdentityCredential` and friends in the public API would force applications to make their expectations explicit but we'd still have to return an error at runtime when the type didn't match the host

heaths (approved):

Issue #2411, opened 2025-03-29
Remove excludedHeaders after all recordings are re-recorded


heaths:
As .NET and some other languages, I'm changing the default matcher to ignore `x-ms-client-request-id` introduced in #2406. We'll need to add more over time. But because it's not in current recordings, it needs to be excluded - not just ignored. Once everything is re-recorded, we should remove the exclusion because that masks any unexpected issues where the header is different or missing from one or the other of the recording or play back.

/cc @analogrelay

PR #2412, opened 2025-03-29
Always ignore certain headers and sanitizers

heaths:
With #2406 adding `x-ms-client-request-id`, I updated the default matcher and set that invariably like other languages. While looking into other languages, I see they also remove several default sanitizers after starting the test-proxy, which Key Vault and Cosmos were already doing so I made that default. To do so, I did some refactoring which, I think, cleans up the `azure_core_test` code a bit.

Opened tracking issue #2411 to remove `x-ms-client-request-id` from the `excludedHeaders` after recordings are updated. It should only be in `ignoredHeaders` long-term.

### DIFF
--- a/sdk/core/azure_core_test/src/proxy/matchers.rs
+++ b/sdk/core/azure_core_test/src/proxy/matchers.rs
@@ -2,18 +2,33 @@
 // Licensed under the MIT License.
 
 use super::join;
-use azure_core::http::headers::{AsHeaders, HeaderName, HeaderValue};
+use azure_core::{
+    http::headers::{AsHeaders, HeaderName, HeaderValue},
+    Bytes,
+};
 use serde::Serialize;
 use std::{convert::Infallible, iter};
 
 // cspell:ignore headerless
 
+/// The headers ignored by default for [`CustomDefaultMatcherBody`];
+pub const DEFAULT_IGNORED_HEADERS: &[&str; 6] = &[
+    "date",
+    "request-id",
+    // cspell:disable-next-line
+    "traceparent",
+    "user-agent",
+    "x-ms-client-request-id",
+    "x-ms-date",
+];
+
 /// Matchers to use for a recording or playback.
-#[derive(Debug)]
+#[derive(Debug, Serialize)]
 pub enum Matcher {
     BodilessMatcher,
-    CustomDefaultMatcher(CustomDefaultMatcherBody),
     HeaderlessMatcher,
+    #[serde(untagged)]
+    CustomDefaultMatcher(CustomDefaultMatcher),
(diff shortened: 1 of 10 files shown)


sdk/core/azure_core_test/src/proxy/matchers.rs
```diff
@@ -58,15 +83,35 @@ pub struct CustomDefaultMatcherBody {
...
 }
 
+impl Default for CustomDefaultMatcher {
+    fn default() -> Self {
+        CustomDefaultMatcher {
+            compare_bodies: None,
+            // TODO: Remove once all recordings are updated (https://github.com/Azure/azure-sdk-for-rust/issues/2411).
+            excluded_headers: vec!["x-ms-client-request-id"],
+            ignored_headers: DEFAULT_IGNORED_HEADERS.to_vec(),
+            ignore_query_ordering: None,
```
RickWinter, line 93: Shouldn't this be true? We want to sanitize no matter how the querry params appear. Or am I missing how this field works?

RickWinter (approved):

sdk/core/azure_core_test/src/proxy/matchers.rs
```diff
@@ -2,18 +2,33 @@
...
 
 // cspell:ignore headerless
 
+/// The headers ignored by default for [`CustomDefaultMatcherBody`];
+pub const DEFAULT_IGNORED_HEADERS: &[&str; 6] = &[
+    "date",
+    "request-id",
+    // cspell:disable-next-line
+    "traceparent",
+    "user-agent",
```
LarryOsterman, line 20: User Agent is there because it means you can't use recordings across operating systems? I thought there was a user-agent specific sanitizer that removed the OS specific `()` section of the recordings.  FWIW, C++ doesn't filter user-agent and somehow recordings "just work".

LarryOsterman (approved):

sdk/core/azure_core_test/src/proxy/matchers.rs
heaths: You're missing how this field works. Not only is this a matcher - not a sanitizer - and query param order should never matter according to HTTP specs, but all this field means is "accept the server default".

sdk/core/azure_core_test/src/proxy/matchers.rs
heaths: I copied this from .NET. Seems no other language has it, though. However, several sanitize it, but that seems problematic. Even if we sanitize the platform info, version info could change. I'd prefer to leave it unless you can think of any reason we would *want* to match against `user-agent`.

PR #2431, opened 2025-04-04
[Storage] `azure_storage_blob` release prep (+ `delete()` for `BlobClient`)

vincenttran-msft:

- Authoring of the initial `README.md` and `CHANGELOG.md` files
- Addition of the `delete()` API for `BlobClient`
    - I thought this was introduced in the initial API surface PR, but must have been overlooked. Either way I think this is important enough to make one more late addition to the MVP.

### DIFF
--- a/sdk/storage/assets.json
+++ b/sdk/storage/assets.json
@@ -1,6 +1,6 @@
 {
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "rust",
-  "Tag": "rust/azure_storage_blob_d3c80a557c",
+  "Tag": "rust/azure_storage_blob_7d1988e753",
   "TagPrefix": "rust/azure_storage_blob"
 }
--- a/sdk/storage/assets.json
+++ b/sdk/storage/assets.json
@@ -1,6 +1,6 @@
 {
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "rust",
-  "Tag": "rust/azure_storage_blob_7d1988e753",
+  "Tag": "rust/azure_storage_blob_c02de844b6",
   "TagPrefix": "rust/azure_storage_blob"
 }
--- a/sdk/storage/azure_storage_blob/README.md
+++ b/sdk/storage/azure_storage_blob/README.md
@@ -34,12 +34,12 @@ az storage account create -n my-storage-account-name -g my-resource-group
 
 #### Authenticate the client
 
-In order to interact with the Azure Blobs Storage service, you'll need to create an instance of a client, `BlobClient`, `BlobContainerClient`, or `BlobServiceClient`. The [Azure Identity] library makes it easy to add Azure Active Directory support for authenticating Azure SDK clients with their corresponding Azure services:
+In order to interact with the Azure Blobs Storage service, you'll need to create an instance of a client, `BlobClient`, `BlobContainerClient`, or `BlobServiceClient`. The [Azure Identity] library makes it easy to add Microsoft Entra ID support for authenticating Azure SDK clients with their corresponding Azure services:
 ```rust
 use azure_storage_blob::BlobClient;
 use azure_identity::DefaultAzureCredential;
 
-// Create a BlobClient that will authenticate through Active Directory
+// Create a BlobClient that will authenticate through Microsoft Entra ID
 let credential = DefaultAzureCredential::new()?;
 let blob_client = BlobClient::new(
     "https://<storage_account_name>.blob.core.windows.net/", // endpoint
@@ -57,7 +57,7 @@ let blob_client = BlobClient::new(
 use azure_storage_blob::BlobClient;
 use azure_identity::DefaultAzureCredential;
 
-// Create a BlobClient that will authenticate through Active Directory
+// Create a BlobClient that will authenticate through Microsoft Entra ID
 let credential = DefaultAzureCredential::new()?;
 let blob_client = BlobClient::new(
     "https://<storage_account_name>.blob.core.windows.net/", // endpoint
--- a/sdk/storage/azure_storage_blob/README.md
+++ b/sdk/storage/azure_storage_blob/README.md
@@ -1,3 +1,140 @@
-# sdk/storage/azure_storage_blob
+# Azure Storage Blobs client library for Rust
 
-TODO TODO TODO
+Azure Blob storage is Microsoft's object storage solution for the cloud. Blob storage is optimized for storing massive amounts of unstructured data, such as text or binary data.
+
+[Source code] | [Package (crates.io)] | [API reference documentation] | [REST API documentation] | [Product documentation]
+
+## Getting started
+
+### Install the package
+
+Install the Azure Storage Blobs client library for Rust with [cargo]:
+
+```sh
+cargo add azure_storage_blob
+```
+
+### Prerequisites
+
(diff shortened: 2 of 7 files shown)


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -149,6 +149,19 @@ impl BlobClient {
         Ok(response)
     }
 
+    /// Marks the specified blob for deletion. The blob is later deleted during garbage collection.
```
jalauzon-msft, line 152: This documentation string has much confusion in the past and should not be used. Please just change to something like "Deletes the Blob". We should eventually update the tsp to reflect this.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -189,6 +189,75 @@ async fn test_upload_blob(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
+        )
+        .await?;
+
+    // Assert
+    let response = blob_client.download(None).await?;
+    let content_length = response.content_length()?;
+    let (status_code, _, response_body) = response.deconstruct();
+    assert!(status_code.is_success());
+    assert_eq!(17, content_length.unwrap());
+    assert_eq!(Bytes::from_static(data), response_body.collect().await?);
```
jalauzon-msft, line 247: Seems like a bit overkill to download the Blob just to prove it's there. Maybe just get properties as that is typically used for existence checks.

sdk/storage/azure_storage_blob/README.md
```diff
@@ -1,3 +1,140 @@
...
+# if using an existing resource group, skip this step
+az group create --name my-resource-group --location westus2
+
+# Create the storage account
+az storage account create -n my-storage-account-name -g my-resource-group
+```
+
+#### Authenticate the client
+
+In order to interact with the Azure Blobs Storage service, you'll need to create an instance of a client, `BlobClient`, `BlobContainerClient`, or `BlobServiceClient`. The [Azure Identity] library makes it easy to add Azure Active Directory support for authenticating Azure SDK clients with their corresponding Azure services:
```
jalauzon-msft, line 37: "Active Directory" is the old term. Please change to Entra Id here and in code comments below.

sdk/storage/assets.json
```diff
@@ -1,6 +1,6 @@
 {
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "rust",
-  "Tag": "rust/azure_storage_blob_d3c80a557c",
+  "Tag": "rust/azure_storage_blob_c02de844b6",
   "TagPrefix": "rust/azure_storage_blob"
 }
```
RickWinter, line 6:
```suggestion
}

```

sdk/storage/azure_storage_blob/README.md
```diff
@@ -1,3 +1,140 @@
...
+```
+
+### Prerequisites
+
+* You must have an [Azure subscription] and an [Azure storage account] to use this package.
+
+
+
+### Create a storage account
+If you wish to create a new storage account, you can use the
```
RickWinter, line 24: By default does the creator have permissions to read/write blobs? Do they need to grant any permissions to themself?

RickWinter (approved):

sdk/storage/assets.json
vincenttran-msft: Hi Rick, this is an autogenrated file when pushign up assets, so I will not make this change as it is in-line with the asset file generated when using test-proxy, see [.NET](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs/assets.json), [Python](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/assets.json)

sdk/storage/azure_storage_blob/README.md
```diff
@@ -1,3 +1,140 @@
-# sdk/storage/azure_storage_blob
+# Azure Storage Blobs client library for Rust
```
jalauzon-msft, line 1: nit: Singular Blob here and on line 11... and anywhere else we have it plural

sdk/storage/azure_storage_blob/README.md
jalauzon-msft: No, they need to add some RBAC roles. @vincenttran-msft please add a section after "Authenticate the client" header called "Permissions" and mention that you may need to setup RBAC roles to access Blob Storage via Microsoft Entra ID. Mostly just link to this page for more info.
https://learn.microsoft.com/azure/storage/blobs/assign-azure-role-data-access?tabs=portal

jalauzon-msft (approved):

sdk/storage/azure_storage_blob/README.md
```diff
@@ -1,3 +1,144 @@
...
+az group create --name my-resource-group --location westus2
+
+# Create the storage account
+az storage account create -n my-storage-account-name -g my-resource-group
+```
+
+#### Authenticate the client
+
+In order to interact with the Azure Blob Storage service, you'll need to create an instance of a client, `BlobClient`, `BlobContainerClient`, or `BlobServiceClient`. The [Azure Identity] library makes it easy to add Microsoft Entra ID support for authenticating Azure SDK clients with their corresponding Azure services:
+```rust
```
heaths, line 38: These need to be marked `rust no_run`. See https://azure.github.io/azure-sdk/rust_introduction.html#rust-documentation-readme.

Also, install `markdownlint`. I can see some warnings in this file. Like learn.microsoft.com, we try to be lint-free even in docs.

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -14,11 +14,11 @@ pub use crate::generated::clients::{
     BlobClientOptions, BlobContainerClientOptions, BlobServiceClientOptions,
 };
 pub use crate::generated::models::{
-    BlobClientDownloadOptions, BlobClientGetPropertiesOptions, BlobContainerClientCreateOptions,
-    BlobContainerClientDeleteOptions, BlobContainerClientGetPropertiesOptions,
-    BlobServiceClientGetPropertiesOptions, BlockBlobClientCommitBlockListOptions,
-    BlockBlobClientGetBlockListOptions, BlockBlobClientStageBlockOptions,
-    BlockBlobClientUploadOptions,
+    BlobClientDeleteOptions, BlobClientDownloadOptions, BlobClientGetPropertiesOptions,
```
heaths, line 17: You're missing the README, which is why you didn't catch the `no_run` necessary. See https://azure.github.io/azure-sdk/rust_introduction.html#rust-documentation-readme.

sdk/storage/azure_storage_blob/README.md
```diff
@@ -1,3 +1,144 @@
...
+
+[Source code] | [Package (crates.io)] | [API reference documentation] | [REST API documentation] | [Product documentation]
+
+## Getting started
+
+### Install the package
+
+Install the Azure Storage Blob client library for Rust with [cargo]:
+
+```sh
```
heaths, line 13: You use both `sh` and `bash` below. You should be consistent. The rest of us are using `sh`, so lets stick with that.

The only exception is when you have a line with just `az` then use `azcli`. IIRC, it gets special treatment for learn.microsoft.com if/when they ingest these.

vincenttran-msft:
Remaining feedback addressed in: #2442

PR #2434, opened 2025-04-07
Update zip dependency

RickWinter:


heaths (approved):
It's a bigger jump than necessary, but looking at their hand-authored changelogs at <https://github.com/zip-rs/zip2/releases> I don't see any harm.

PR #2437, opened 2025-04-07
Update openssl version

heaths:
Fixes GHSA-4fcv-w3qc-ppgg


heaths:
This is requiring a new version of OpenSSL on Windows. We can integrate `vcpkg` into build, but this potentially adds a burden to Windows devs using our crates. It's likely that our hosted agents already had the version we used before for `openssl-sys` 0.9.105 but don't for 0.9.107.

I find nothing obvious to necessitate an update in their diff, though: <https://github.com/sfackler/rust-openssl/compare/openssl-sys-v0.9.105...openssl-sys-v0.9.107>

Let's see what comes of <https://github.com/sfackler/rust-openssl/issues/2391>.

I'm going to push the `vcpkg` changes I have for now and mark this a WIP. We can remove them if possible later.

vcpkg.json
heaths, line 1: Before we merge this, we should update the root `CONTRIBUTING.md` file with instructions for Windows dev to successfully build.

heaths:
Making progress on this, but blocked using just vcpkg because of https://github.com/mcgoo/vcpkg-rs/issues/41 (does not support manifest mode, which `openssl-sys` uses. There are some workarounds mentioned therein I'll try.

eng/vcpkg.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "azure-sdk-for-rust",
+  "dependencies": [
+    "openssl"
+  ]
+}
```
LarryOsterman, line 6: Missing crlf at end of file.

LarryOsterman (approved):

RickWinter (approved):

eng/vcpkg.json
heaths: That's odd. I wonder what editor I used that would've ignored `.editorconfig` or just didn't do it outright. VSCode (with EditorConfig extension), helix (my personal settings always cause a newline), and VS (supports `.editorconfig` out of the box) all do.

heaths: That said, I don't think it necessary to get re-approvals for adding that. If someone edits in the future, it should happen then. I'm just puzzled how it didn't in the first place.

LarryOsterman (approved):

PR #2438, opened 2025-04-07
Add ClientSecretCredential

### LINKED ISSUE #2219: Substitute for removed `ClientSecretCredential`
Hi,

When upgrading azure_identity from 0.21.0 to 0.22.0, I noticed that the `ClientSecretCredential` type was removed (in https://github.com/Azure/azure-sdk-for-rust/pull/1821/).

I can see that it was done as part of a cleanup initiative of the azure_identity crate, now I'm trying to understand why and if there is a new type or something else that should be used instead of `ClientSecretCredential` to get a token from a client id and a client secret?

chlowell:
Closes #2219


sdk/identity/azure_identity/src/client_secret_credential.rs
```diff
@@ -0,0 +1,276 @@
...
+    client_id: String,
+    endpoint: Url,
+    options: TokenCredentialOptions,
+    secret: Secret,
+}
+
+impl ClientSecretCredential {
+    pub fn new(
+        tenant_id: String,
+        client_id: String,
```
heaths, line 42: `tenant_id`, at least, should be a `&str` because we don't take ownership of it. We probably should declare `secret` as either `Secret` - make the caller do the allocation as needed, if even needed (probably) - or this may be a case to declare it as `impl Into<Secret>` which allows for a lot of possibilities. I'm leaning toward the former because it better follows our guidelines.

sdk/identity/azure_identity/src/client_secret_credential.rs
```diff
@@ -0,0 +1,276 @@
...
+            StatusCode::Ok => {
+                let token_response: EntraIdTokenResponse = deserialize(res).await?;
+                Ok(AccessToken::new(
+                    token_response.access_token,
+                    OffsetDateTime::now_utc() + Duration::from_secs(token_response.expires_in),
+                ))
+            }
+            _ => {
+                let error_response: EntraIdErrorResponse = deserialize(res).await?;
+                let mut message = "ClientSecretCredential authentication failed".to_string();
```
heaths, line 94: You shouldn't need to allocate a string here. `format!()` takes an implicit ref anyway, and `&str` should satisfy `Error::new`, at least, if not also `message`.

heaths (approved):

sdk/identity/azure_identity/src/client_secret_credential.rs
LarryOsterman: QQ: (since this hits some of the stuff I'm doing): Should we be consistent across all parameters to the function? To me, it feels weird to have `foo(a:String, b:&str, c:String)` just because of what the function does with the different parameters (especially since the caller's experience is different for each - it feels like you're exposing an implementation detail of the API - and if at some point in the future you need to take ownership of `tenant_id`, you're a bit stuck).

heaths: This is a separate thread I started but for now I would say: follow guidelines. If we're going to own it, make them pass a concrete struct; otherwise, pass a ref.

sdk/identity/azure_identity/src/client_secret_credential.rs
chlowell: `format!` returns a String I couldn't assign back to `message` if it were `&str`. Don't I need to allocate a String in any case, to hold the concatenation?

sdk/identity/azure_identity/src/client_secret_credential.rs
RickWinter: I agree with Larry on his concern about exposing the implementation detail. We are forever bound to the api design. Is it really worth making the customer think about how to pass a string because we may/may not want ownership?

RickWinter (approved):

heaths (approved):

sdk/identity/azure_identity/src/client_secret_credential.rs
heaths: Shadowing types in rust can change the type, unlike in Go. But, yes, you need to have a `String` to concat. My point was that your first use of `message` passed to `format!()` just needs to be a `&str`, which `"some text"` already is.

heaths: I wouldn't hold up the beta for this, though.

PR #2442, opened 2025-04-07
[Storage] `azure_storage_blob` release documentation nits

vincenttran-msft:
This PR addresses the following remaining feedback items:

- Code samples `rust` -> `rust no_run`
- Unified terminal samples to `sh`
- Installed and ran Markdown Lint extension (thanks @heaths for the recommendation)
- Removed the `sdk/storage` README as it now lives in `sdk/azure_storage_common` (and is currently an eyesore of `TODO`s that you must navigate past to get to `azure_storage_blob`)
- Fixed a typo on the source code link (was still linking to Python 🐍 )


sdk/storage/README.md
heaths, line 1: I didn't even notice that; though, don't you still need one for `azure_storage_common`? It probably doesn't need to be as large. Could even be something like a heading, the "links" bar to source, and just a description like:

````markdown
```markdown
This crate supports the following Azure Storage crates:

- [azure_storage_blob](fully-qualified link to its README.md perhaps)
```
````

Sort of a way to advertise your other crates as well in case it shows up higher in search hits.

**Review by heaths — APPROVED** [review]

File: sdk/storage/README.md
Comment (vincenttran-msft) [reply sdk/storage/README.md:1]: That sounds good, the next iteration of the PR will include a new `README` for common. However, right now `azure_storage_common` has nothing, so I believe Storage is only looking to release `azure_storage_blob`. But please let me know if there is any requirements regarding the common package being released alongside Blob.

Comment (heaths) [reply sdk/storage/README.md:1]: You can't ship it if it still has a dependency on common. So if there's nothing in it and you don't use it, get rid of the dependency. Also, the release process is still going to try to ship common, so if you don't need it, I recommend removing it from the "Stages" when you go to kick off a build so you don't have to reject it later (an option) and have the release build show up as failed (rejecting a package will mark the build failed, but the approved packages still get published...it's just that at a glance the build will always look failed, which may raise some initial red flags).

Comment (vincenttran-msft) [reply sdk/storage/README.md:1]: Understood, I just pushed another change that removes it as a depedency (I honestly didn't even know we had one, we must have made that decision early on and forgot).

I will also note down to remove it from Stages on the release build -- this is a common thing we do on the Python SDK Release as it is cluttered with other packages we don't actually manage and release.

File: sdk/storage/azure_storage_blob/README.md [hunk 1]
```diff
@@ -35,85 +35,105 @@ az storage account create -n my-storage-account-name -g my-resource-group
...
-// Create a BlobClient that will authenticate through Microsoft Entra ID
-let credential = DefaultAzureCredential::new()?;
-let blob_client = BlobClient::new(
-    "https://<storage_account_name>.blob.core.windows.net/", // endpoint
-    "container_name".to_string(),                            // container name
-    "blob_name".to_string(),                                 // blob name
-    credential,                                              // credential
-    Some(BlobClientOptions::default()),                      // BlobClient options
-)?;
+async fn example() -> Result<(), Box<dyn std::error::Error>> {
```
heaths, line 43: Example should use `main`. See our other READMEs, like Key Vault's. These should basically be ready to copy and paste. Think LLM training.

heaths (changes requested):

sdk/storage/azure_storage_blob/README.md
vincenttran-msft: I can make this change. I actually lifted this from Cosmos' [README](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/cosmos/azure_data_cosmos/README.md), so looks like an adjustment should be made there as well 😄

jalauzon-msft (approved):

Issue #2327, opened 2025-03-13
Remove old_azure_cli feature from azure_identity


heaths:
Because the caller of `azure_identity` can't know what version of the Azure CLI is installed, we should remove the `old_azure_cli` feature and support both `expiresOn` and `expires_on` as does Go: https://github.com/Azure/azure-sdk-for-go/blob/d34a8dd747cbaaac577a8b53ec56b970bff09eaa/sdk/azidentity/azure_cli_credential.go#L179-L185

This does mean more dependencies but I don't think is something we can avoid.

heaths:
@chlowell @joshfree do you agree? Basically, we use a feature named 'old_azure_cli` that is part of the default features and reads the `expiresOn` field of the access token, but that causes us to pull in more date-time dependencies. The old code treats it as optional, but it seems to me that shouldn't be up to the dev using our SDKs, but we should support whatever version the end user has installed.

At some point do we cut support for that version of Azure, though? According to a comment in our `azure_identity` crate, that version is from before November 2023:

```rust
# If you are using and Azure CLI version older than 2.54.0 from November 2023,
# upgrade your Azure CLI version or enable this feature.
# Azure CLI 2.54.0 and above has an "expires_on" timestamp that we can use.
# https://github.com/Azure/azure-cli/releases/tag/azure-cli-2.54.0
# https://github.com/Azure/azure-cli/issues/19700
old_azure_cli = ["time/local-offset", "tz-rs"]
```

/cc @cataggar

chlowell:
Makes sense to me. The credential working seamlessly in both cases because it handles the old/new distinction internally is obviously a better user experience, modulo the additional dependencies.

> At some point do we cut support for that version of Azure [CLI], though?

🤷 I expect to continue supporting the old format indefinitely because we can't tell whether anyone is using it

PR #2443, opened 2025-04-07
Remove azure_identity's old_azure_cli feature

chlowell:
This feature enables parsing token expiration times from old Azure CLI versions. The newest "old" version is 2.53.0. Assuming that was an LTS release, the Azure CLI team stopped supporting it in 11-2024 with the release of version 2.66.0. #2327 proposes removing the crate feature while keeping the "old format" parsing code, however I think it's reasonable to remove both because we haven't shipped a stable azure_identity with this feature and no supported Azure CLI version now requires it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -125,7 +125,6 @@ tokio = { version = "1.0", default-features = false, features = [
 ] }
 tracing = "0.1.40"
 tracing-subscriber = "0.3"
-tz-rs = { version = "0.6" }
 url = "2.2"
 uuid = { version = "1.0", features = ["v4"] }
 zerofrom = "0.1.5"
--- a/sdk/identity/azure_identity/src/credentials/azure_cli_credentials.rs
+++ b/sdk/identity/azure_identity/src/credentials/azure_cli_credentials.rs
@@ -16,105 +16,11 @@ use std::{ffi::OsStr, fmt::Debug, str, sync::Arc};
 use time::OffsetDateTime;
 use tracing::trace;
 
-#[cfg(feature = "old_azure_cli")]
-mod az_cli_date_format {
-    use azure_core::error::{ErrorKind, ResultExt};
-    use serde::{Deserialize, Deserializer};
-    use time::format_description::FormatItem;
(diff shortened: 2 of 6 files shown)


LarryOsterman (approved):

heaths:
We should probably make sure - if possible - we have a clear error if we get the old format: something to the effect, "Hey, update your `az` CLI to a supported version!"

Cargo.toml
```diff
@@ -125,7 +125,6 @@ tokio = { version = "1.0", default-features = false, features = [
 ] }
 tracing = "0.1.40"
 tracing-subscriber = "0.3"
-tz-rs = { version = "0.6" }
```
heaths, line 128: Always glad to reduce dependencies! 👍

sdk/identity/azure_identity/src/credentials/azure_cli_credentials.rs
```diff
@@ -131,19 +37,10 @@ impl CliTokenResponse {
...
-                #[cfg(feature = "old_azure_cli")]
-                {
-                    Ok(self.local_expires_on)
-                }
-                #[cfg(not(feature = "old_azure_cli"))]
-                {
-                    Err(Error::message(
-                        ErrorKind::DataConversion,
-                        "expires_on field not found. Please use Azure CLI 2.54.0 or newer.",
-                    ))
```
heaths, line 144: We should leave something to this effect. I worry about sudden breakage for those that installed `az` long ago and haven't updated since.

heaths (changes requested):
Don't recall if needed - w.r.t. to the submodule for older parsing - but if there's an opportunity to reduce submodules that were used to split things up, always good to clean up code a little as well.

sdk/identity/azure_identity/src/credentials/azure_cli_credentials.rs
chlowell: I removed the feature macros, not this error; it's now returned from the None arm in any case

heaths (approved):

PR #2468, opened 2025-04-09
[Storage] [WIP] `list_blobs` for `ContainerClient`

vincenttran-msft:
Going to take some time to fix my dev environment so that I can probably use WSL and debugger tools, but this seems to be unexpectedly `None`?

Evaluating the HTTP Response in Fiddler shows that enumerated results are in fact there, but possibly some failure in deserialization to fill the fields?

### DIFF
--- a/sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
+++ b/sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
@@ -3,9 +3,11 @@
 
 use crate::{
     generated::clients::BlobContainerClient as GeneratedBlobContainerClient,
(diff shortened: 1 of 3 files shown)


vincenttran-msft:
> Where exactly? It's possible we have some corner-case bugs in the pager.

I did a little bit more digging and I see that pageable operations in Storage not exactly fully functioning yet, https://github.com/Azure/typespec-rust/pull/354 introduces pageable to 1 operation (`list_containers`). Will leave a follow-up on the issue tracking this effort: https://github.com/Azure/typespec-rust/issues/60

vincenttran-msft:
Closing as duplicate and stale.

PR #2474, opened 2025-04-10
Fix Key Vault secrets examples in azure_core README

heaths:
Since `azure_core` shipped first, we needed to wait until after
`azure_security_keyvault_secrets` shipped to fix up the samples.


heaths:
Wanted to fix this in another PR since I've had to do it in a few still-outstanding ones. It was just a timing issue: had to ship core first so it was still the older KV secrets API.

RickWinter (approved):

Issue #1746, opened 2024-08-13
Pass `--all-features` to any commands that use it, typically, e.g., linting.


heaths:

heaths:
With #1926 we almost can, but testing with `reqwest-rustls` builds `rustls-openssl` which is failing on Windows. I'm inventorying the software and, according to <https://github.com/actions/runner-images/blob/win22/20241021.1/images/windows/Windows2022-Readme.md#tools> (link from agent info during Azure Pipelines build), it's installed. We probably will need to set env vars.

At the same time, maybe it's time to consider switching over to SymCrypt. We likely will eventually, but their acquisition story is sub-par currently, since customers have to get platform-specific binaries and that's going to be a problem.

I've got some inventory changes in Azure/azure-sdk-tools#9424 that should help.

analogrelay:
I've worked around this in #3428 so that Cosmos can run emulator tests (which require key-based auth, which is behind a feature flag) in PRs

heaths:
@hallipr @danieljurek we should be unblocked for this one now.

PR #2476, opened 2025-04-11
Implement EventHubs token refresh

LarryOsterman:
Implement Event Hubs authorization token refresh.

Because AMQP is a connection-oriented protocol (unlike HTTP which is nominally a connectionless protocol), authorization tokens used by AMQP can expire while the connections are still active. To implement this, each EventHubs (and ServiceBus) client spawns a background task to refresh authorization tokens.

There are several elements to this PR:
1. Updated the AMQP library to use the `async_trait` trait rather than implementing async traits manually. This uncovered several subtle bugs in the AMQP implementation (mostly the `Send` trait being missing on certain types).
2. Added a new "task spawner" type to azure core to facilitate creating an asynchronous task from the current runtime (currently supported runtimes are "tokio" and "standard".
3. Updated the EventHubs `ConnectionManager` type to add a token refresh task which is launched when a path is first authorized.
   1. As a part of the `ConnectionManager` refactoring, the `credential` field was moved from the service client objects into the connection manager and the `credential` field was removed from the `authorize_path` method.
5. A clippy nightly suggested fix to the `StdExecutor` type to simplify the error case.

Note that the ConnectionManager has a number of `#[cfg(test)]` test hooks to enable unit testing the connection manager token refresh logic. That allows for in-isolation testing of the token refresh logic while not significantly impeding the functionality.


sdk/core/azure_core/src/task/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+        self.0.await.map_err(|e| {
+            crate::Error::message(
+                crate::error::ErrorKind::Other,
+                format!("Task was cancelled before completion: {}", e),
+            )
+        })
+    }
+}
+
+#[cfg(all(not(feature = "tokio"), not(target_arch = "wasm32")))]
```
heaths, line 71: Rather than repeat this condition over and over, either put it in the `tokio_spawn` module and export it here with that condition, or make a private module and export everything from there. Much easier to manage and read:

```rust
#[cfg(...)]
mod __tokio {
    pub struct ...;
}

#[cfg(same ...)]
pub use __tokio::*;
```

sdk/core/azure_core/src/task/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+        Arc::new(WasmSpawner)
+    }
+}
+
+#[cfg(feature = "tokio")]
+#[derive(Debug)]
+pub struct SpawnHandle(tokio::task::JoinHandle<()>);
+
+#[cfg(feature = "tokio")]
+impl SpawnHandle {
```
heaths, line 59: The `async_trait` we use everywhere else may help here. It rewrites the signatures into a `BoxedFuture`.

sdk/core/azure_core/src/task/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+pub use wasm_spawn::WasmSpawner;
+
+#[cfg(feature = "tokio")]
+pub use tokio_spawn::TokioSpawner;
+
+#[cfg(not(target_arch = "wasm32"))]
+pub(crate) type TaskFuture = Pin<Box<dyn Future<Output = ()> + Send + 'static>>;
+
+#[cfg(target_arch = "wasm32")]
+pub(crate) type TaskFuture = Pin<Box<dyn Future<Output = ()> + 'static>>;
```
heaths, line 33: Why not just use `async_trait` like we do everywhere else? I don't recall if it's `Pin`, but it'll rewrite to a `BoxedFuture` and consistent with all other code. Should work here, too.

sdk/core/azure_core/src/task/mod.rs
LarryOsterman: Unfortunately, that doesn't work for parameters to traits, only trait functions :(. Because boy, I would love to be able to use it :).

sdk/core/azure_core/src/task/mod.rs
LarryOsterman: `async_trait` doesn't map futures as parameters to trait functions :(. So unfortunately I need to make it explicit.

heaths (changes requested):

sdk/core/azure_core/src/task/mod.rs
```diff
@@ -18,24 +18,30 @@ mod tokio_spawn;
...
-pub use standard_spawn::StdSpawner;
+pub use standard_spawn::{SpawnHandle, StdSpawner};
 
 #[cfg(all(not(feature = "tokio"), target_arch = "wasm32"))]
-pub use wasm_spawn::WasmSpawner;
+pub use wasm_spawn::{SpawnHandle, WasmSpawner};
 
 #[cfg(feature = "tokio")]
-pub use tokio_spawn::TokioSpawner;
+pub use tokio_spawn::{SpawnHandle, TokioSpawner};
```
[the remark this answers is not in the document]
LarryOsterman: Ah - that makes a lot of sense. I tried making the `SpawnHandle` a trait, but it REALLY wants to be a concrete type. Let me see what I can come up here.

sdk/core/azure_core/src/task/standard_spawn.rs
```diff
@@ -30,3 +30,18 @@ impl TaskSpawner for StdSpawner {
         SpawnHandle(th)
     }
 }
+
+#[derive(Debug)]
+pub struct SpawnHandle(std::thread::JoinHandle<()>);
+
+impl SpawnHandle {
+    /// Wait for the task to complete and return the result.
+    pub async fn await_result(self) -> crate::Result<()> {
```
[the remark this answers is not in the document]
LarryOsterman: Sure - makes sense to me.

sdk/core/azure_core/src/task/mod.rs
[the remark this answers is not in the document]
LarryOsterman: Ok. I wasn't able to get your suggestion to work (RustC complained that `struct SpawnHandle(inner: SpawnHandle)` had an infinite size. However, I came up with something that I think is pretty close to the equivalent (an async trait for the `wait` method, and a `SpawnHandleT` generic which requires the async trait for it's generic parameter, and then a `pub type SpawnHandle = SpawnHandleT<implementation>`.  And it's more extensible over time.

sdk/core/azure_core/src/task/tests.rs
```diff
@@ -106,3 +105,62 @@ async fn test_std_specific_handling() {
...
+        handle
+            .await_result()
+            .await
+            .expect("Task should complete successfully");
+    }
+    // Verify all tasks executed
+    assert_eq!(*counter.lock().unwrap(), 5);
+}
+
+#[cfg(not(feature = "tokio"))]
```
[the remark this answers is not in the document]
LarryOsterman: There was a bug in this test case - line 112 should have been `let spawner = Arc::new(standard_spawn::StdSpawner)` - in other words, it forces the use of the "standard spawner". I've fixed the test case and I've added printouts to the test cases to confirm that both the standard and tokio tests are run (the wasm tests are not run because we've not yet standardized on the runtime environment for wasm).

sdk/core/azure_core/src/task/mod.rs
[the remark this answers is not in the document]
heaths: You'd import as a fixed alias like I showed above, redefining `self as inner` such that `inner::SpawnHandle` is correct. You could also alias `import stdandard_handle::{SpawnHandle as InnerHandle, StdSpawner};` or something. The key was the common alias. Rust does this with its `std` and `core` libraries quite a bit.

sdk/core/azure_core/src/task/tests.rs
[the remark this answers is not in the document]
heaths: But we test with `--all-features`, so any `not(feature = "tokio")` would never been run.

sdk/core/azure_core/src/task/mod.rs
```diff
@@ -17,14 +48,52 @@ mod tokio_spawn;
...
+    /// A handle to a spawned task, allowing you to wait for its completion.
+    #[cfg(all(not(feature = "tokio"), not(target_arch = "wasm32")))]
+    Std(standard_spawn::StdSpawnHandle),
+    /// A handle to a spawned task, allowing you to wait for its completion.
+    #[cfg(all(not(feature = "tokio"), target_arch = "wasm32"))]
+    Wasm(wasm_spawn::WasmSpawnHandle),
+    /// A handle to a spawned task, allowing you to wait for its completion.
+    #[cfg(feature = "tokio")]
+    Tokio(tokio_spawn::TokioSpawnHandle),
+}
```
heaths, line 63: How's oxidizer or any other implementation going to plug in?

heaths: If you're going to have a concrete type, it either has to take a trait impl by type param or be dynamically dispatched. I don't think a type param will work in this case, though, since you're returning it e.g.,

```rust
trait SpawnHandle {}

#[cfg(feature = "foo")]
mod foo {
    pub struct FooSpawnHandle {}
    impl super::SpawnHandle for FooSpawnHandle {}
}

#[cfg(not(feature = "foo"))]
mod bar {
    pub struct BarSpawnHandle {}
    impl super::SpawnHandle for BarSpawnHandle {}
}

struct Task(pub Box<dyn SpawnHandle>);

trait TaskSpawner {
    fn spawn(&self) -> Task;
}

struct TaskFactory {}

impl TaskSpawner for TaskFactory {
    fn spawn(&self) -> Task {
        #[cfg(feature = "foo")]
        {
            Task(Box::new(foo::FooSpawnHandle {}))
        }

        #[cfg(not(feature = "foo"))]
        {
            Task(Box::new(bar::BarSpawnHandle {}))
        }
    }
}

mod other {
    use super::{SpawnHandle, Task, TaskSpawner};

    pub struct OtherTaskHandle {}
    impl SpawnHandle for OtherTaskHandle {}

    pub struct OtherTaskSpawner {}
    impl TaskSpawner for OtherTaskSpawner {
        fn spawn(&self) -> Task {
            Task(Box::new(OtherTaskHandle {}))
        }
    }
}
```

sdk/core/azure_core/src/task/mod.rs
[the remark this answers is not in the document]
LarryOsterman: I'm struggling to get this to compile, possibly because I split the implementation into separate modules.

I'm also confused about the use of the `mod standard_spawn` and then the `use standard_handle::{self as inner` - where is the standard_handle module supposed to be defined? It seems to come out from nowhere.

I'm also not sure why my newtype implementation is less efficient or substantially different from your implementation.

sdk/core/azure_core/src/task/standard_spawn.rs
```diff
@@ -27,16 +27,16 @@ impl TaskSpawner for StdSpawner {
...
 
 #[derive(Debug)]
-pub struct SpawnHandle(std::thread::JoinHandle<()>);
+pub struct StdSpawnHandle(std::thread::JoinHandle<()>);
 
-impl SpawnHandle {
+impl StdSpawnHandle {
     /// Wait for the task to complete and return the result.
-    pub async fn await_result(self) -> crate::Result<()> {
+    pub(crate) async fn wait(self) -> crate::Result<()> {
```
heaths, line 39: Any reason not to just impl `Future`? Did we already talk about that? `wait` is really doing nothing more than polling would with `await` anyway. So it'd be just `handle.await`, which is in line with other `std::thread` and `tokio` tasks.

heaths (changes requested):

sdk/core/azure_core/src/task/tests.rs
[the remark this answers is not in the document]
LarryOsterman: I'm profoundly confused then.

I run `cargo test --package azure_core task` and what I see in the output is:

```
running 5 tests
test task::tests::test_multiple_tasks ... ok
test task::tests::std_multiple_tasks ... ok
test task::tests::test_task_execution ... ok
test task::tests::std_task_execution ... ok
test task::tests::std_specific_handling ... ok
```
which implies that the `std_specific_handling` test was run. But the `std_specific_handling` test is marked as `#[cfg(not(feature="tokio")))]`.

And I've put traces into the `std_` tests and they ARE executed.

Oh, and strangely, if I specify `cargo test --package azure_core task --all-features`, I see the output is:

```
test task::tests::test_multiple_tasks ... ok
test task::tests::test_tokio_specific_handling ... ok
test task::tests::test_task_execution ... ok
```

Which confuses me great deal since it runs a subset of the set of tests which are run without the --all-features.

Also strange: The [CI pipeline](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4782163&view=logs&j=9ec8044b-c590-5e7d-fcc5-eb7e93c897b5&t=9ebea1ad-ab18-5b60-c9d9-983e723b80c2&l=696) runs:
```
test task::tests::test_multiple_tasks ... ok
test task::tests::std_multiple_tasks ... ok
test task::tests::std_task_execution ... ok
test task::tests::test_task_execution ... ok
test task::tests::std_specific_handling ... ok
```

And strangely, it *doesn't* run the tokio specific tests. 

So, I'm not 100% sure why you believe it isn't tested.

sdk/core/azure_core/src/task/mod.rs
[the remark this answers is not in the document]
heaths: I don't see you using newtype. I see you're using a fixed enum. That can't be extended.

You still define the `mod standard_spawn` like you are. The `use` only imports, in which case you're aliasing the mod name as `inner`. Alternatively, you could alias the type with a common name. Same thing I showed in my code snippet below, or in <https://github.com/heaths/akv-cli-rs/blob/cef38d77b0dc9317195833547a00e09fa4879100/src/pty/mod.rs#L5-L20>. The rust standard libraries do this a lot to "wrap" changes. It's all resolved at build time, so as long as calls are the same you don't even need a trait.

sdk/core/azure_core/src/task/tests.rs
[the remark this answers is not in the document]
heaths: Turns out `--all-features` isn't passed yet: <https://github.com/Azure/azure-sdk-for-rust/issues/1746>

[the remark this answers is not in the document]
heaths: So that's why `std` is being used but `tokio` is not. Ideally we want to test both.

I think `std` should always be imported. `tokio` should be optional because a user may not want those libraries, but `std` will always have to be used (we're so far from `no_std` it's not even close to top of mind). The factory can fall back from the `std` spawner, but it should probably always be available. Only implementations requiring more dependencies should be optional.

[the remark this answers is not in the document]
heaths: That said, I think it's fine to have distinct fallbacks for wasm32 vs not(wasm32) since those are distinct.

sdk/core/azure_core/src/task/standard_spawn.rs
LarryOsterman: Let me see what that would look like (returning the future directly). There may be some interesting challenges related to the different return types, but I may be able to make this work.

sdk/core/azure_core/src/task/mod.rs
```diff
@@ -49,52 +50,44 @@ mod tokio_spawn;
...
-            #[cfg(all(not(feature = "tokio"), target_arch = "wasm32"))]
-            SpawnHandle::Wasm(handle) => handle.wait().await,
-            #[cfg(feature = "tokio")]
-            SpawnHandle::Tokio(handle) => handle.wait().await,
-        }
-    }
+    async fn wait(self: Box<Self>) -> crate::Result<()>;
 }
 
+// #[derive(Debug)]
```
heaths: Any reason to leave this commented code in?

sdk/core/azure_core/src/task/mod.rs
```diff
@@ -29,8 +29,9 @@
 //! ```
 //!
 //!
+use async_trait::async_trait;
 use std::{
-    fmt::{self, Debug},
+    fmt::{self},
```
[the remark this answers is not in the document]
LarryOsterman: Not 100% sure - it should have removed it. But that's all just unnecessary.

sdk/core/azure_core/src/task/standard_spawn.rs
```diff
@@ -1,61 +1,119 @@
...
+#[cfg(not(target_arch = "wasm32"))]
+use std::pin::Pin;
+#[cfg(not(target_arch = "wasm32"))]
+use std::sync::{Arc, Mutex};
+#[cfg(not(target_arch = "wasm32"))]
+use std::task::Waker;
+#[cfg(not(target_arch = "wasm32"))]
+use std::task::{Context, Poll};
+#[cfg(not(target_arch = "wasm32"))]
 use std::thread;
```
heaths, line 22: This is really hard to read. Could you combine all the imports? That's typical. rust-analyzer should do it automatically for non-conditional, but then just manually move stuff in manually.

sdk/core/azure_core/src/task/standard_spawn.rs
```diff
@@ -1,61 +1,119 @@
...
-        self.0.join().map_err(|e| {
-            crate::Error::message(
-                crate::error::ErrorKind::Other,
-                format!("Task was cancelled before completion: {:?}", e),
-            )
-        })
+                    // Spawn the future on the local executor
+                    let future_handle = spawner
+                        .spawn_with_handle(f)
+                        .expect("Failed to spawn future");
```
heaths, line 102: Was this temporary?

sdk/core/azure_core/src/task/mod.rs
```diff
@@ -60,8 +54,21 @@ pub(crate) type TaskFuture = Pin<Box<dyn Future<Output = ()> + Send + 'static>>;
...
+>;
+
+#[cfg(target_arch = "wasm32")]
+pub type SpawnedTask =
+    Pin<Box<dyn Future<Output = std::result::Result<(), Box<dyn std::error::Error>>> + 'static>>;
+
 /// An async command runner.
-pub trait TaskSpawner: Send + Sync + fmt::Debug {
+//#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+//#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
```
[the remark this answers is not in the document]
LarryOsterman: Nope, cleaned it up in current PR.

sdk/core/azure_core/src/task/standard_spawn.rs
LarryOsterman: Essentially, yes.

I had mentioned that copilot created the initial skeleton of this code, and copilot just LOVES `.unwrap` and `.expect`. 

As part of my cleanup of the code, I went through and removed all the `.unwrap` and `.expect` fields, including this one.

sdk/core/azure_core/src/task/mod.rs
```diff
@@ -45,13 +39,13 @@ mod tokio_spawn;
...
-    /// Wait for the task to complete and return the result.
-    async fn wait(self: Box<Self>) -> crate::Result<()>;
-}
+// /// A `SpawnHandle` is a handle to a spawned task, allowing you to wait for its completion.
+// #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+// #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+// pub trait SpawnHandle: Send + fmt::Debug {
+//     /// Wait for the task to complete and return the result.
+//     async fn wait(self: Box<Self>) -> crate::Result<()>;
+// }
```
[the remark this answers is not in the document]
LarryOsterman: Nope - just hadn't cleaned it out.

sdk/core/azure_core/src/task/standard_spawn.rs
```diff
@@ -30,39 +27,54 @@ struct ThreadJoinFuture {
...
-                // for when it completes.
-                join_state.waker = Some(_cx.waker().clone());
-                Poll::Pending
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut join_state = self.join_state.lock().map_err(|e| {
+            debug!("Failed to lock join state: {}", e);
+            Box::new(crate::Error::message(
+                crate::error::ErrorKind::Other,
+                format!("Thread panicked: {:?}", e),
+            )) as Box<dyn std::error::Error + Send>
```
heaths, line 46: Tip: you shouldn't need to box the error yourself. I've never once had to when functions return a `Result<T, Box<dyn Error>`.

```rust
fn example() -> Result<(), Box<dyn std::error::Error>> {
    Err("error".into())
}

fn main() {
    match example() {
        Ok(()) => {},
        Err(err) => eprintln!("{err:?}"),
    }
}
```

sdk/core/azure_core/src/task/standard_spawn.rs
```diff
@@ -84,36 +96,58 @@ impl TaskSpawner for StdSpawner {
...
+                        js.join_handle = Some(thread::spawn(move || {
+                            // Create a local executor
+                            let mut local_pool = LocalPool::new();
+                            let spawner = local_pool.spawner();
 
-                    // Spawn the future on the local executor
-                    let future_handle = spawner
-                        .spawn_with_handle(f)
-                        .expect("Failed to spawn future");
+                            // Spawn the future on the local executor
```
heaths, line 108: Nit: the deep nesting here is a bit hard to follow. I wouldn't hold up the PR for this, but maybe consider refactoring using the `let..else` syntax to unnest e.g.,

```rust
let Ok(future_handle) = spawn_result else {
  return Err(crate::Error::mesage(...));
};

// ...
```

heaths (approved):
Just a couple nits/tips. I'd like to see a "flattening" of the `standard_spawn.rs` but I wouldn't hold up this PR for it. It made it a bit hard to follow because all of a sudden is felt like reading...gasp...VB and keeping prefaced whitespace in mind because important. 😉

sdk/core/azure_core/src/task/standard_spawn.rs
LarryOsterman: The only negative on that is that it loses the ability to report the actual error from the spawn call. But that may not be significant.

LarryOsterman: Cleaned up a few of these.

heaths: For posterity, as we discuss in person, that's probably fine because your error strings are unique to the point of tracking down *where* the error occurs and, as you pointed out, in most cases it's a failure to get a lock...which should be rare anyway.

If we ever revisit it, you can do something like a match with an explicit return out of the function or even a `map_err` in many cases.

heaths (approved):

PR #2477, opened 2025-04-11
[Storage] Add `set_metadata` to BlobClient

vincenttran-msft:
This PR introduces:
- `set_metadata` API to `BlobClient`
TypeSpec PR: https://github.com/Azure/azure-rest-api-specs/pull/33966

TypeSpec Changes:
Omit `date`, `last_modified`, `etag`, `encryption_key_sha256`, `encryption_scope`, `is_server_encrypted`, and `version_id` from `BlobClientSetMetadataResult` to effectively yield Response<()>


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -149,6 +150,21 @@ impl BlobClient {
         Ok(response)
     }
 
+    /// Sets user-defined metadata for the specified blob as one or more name-value pairs. Each call to this operation
+    /// replaces all existing metadata attached to the blob. TO remove all metadata from the blob, call this operation with
```
heaths, line 154: Nit:
```suggestion
    /// replaces all existing metadata attached to the blob. To remove all metadata from the blob, call this operation with
```

heaths (approved):

PR #2481, opened 2025-04-11
Fix Key Vault signing sample

heaths:


mikeharder (approved):

heaths:
The vault was a test-only resource I just provisioned to test the sample and have already deleted/purged.

PR #2487, opened 2025-04-15
Support .env files for testing

### LINKED ISSUE #2038: Read environment variables from optional file
`New-TestResources.ps1` may create an environment file e.g., `.env` that we should supporting reading from if there. This makes it easier to work in an IDE without having to restart the IDE to pick up process environment variables.

heaths:
Requires changes to the TestResources scripts, which will need to be merged into Azure/azure-sdk-tools. Client libraries need to opt in.

Resolves #2038

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -218,6 +218,7 @@ dependencies = [
  "azure_identity",
  "azure_security_keyvault_secrets",
  "clap",
+ "dotenvy",
  "flate2",
  "futures",
  "rand",
@@ -755,6 +756,12 @@ dependencies = [
  "quote",
  "syn",
 ]
+
+[[package]]
+name = "dotenvy"
+version = "0.15.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1aaf95b3e5c8f23aa320147307562d361db0ae0d51242340f558153b4eb2439b"
 
 [[package]]
 name = "dyn-clone"
--- a/eng/dict/crates.txt
+++ b/eng/dict/crates.txt
@@ -22,6 +22,7 @@ base64
 bytes
 cargo_metadata
 clap
+dotenvy
 dyn-clone
 fe2o3-amqp
 fe2o3-amqp-cbs
--- a/eng/scripts/verify-dependencies.rs
+++ b/eng/scripts/verify-dependencies.rs
@@ -17,7 +17,10 @@ use std::{
     process::Command,
 };
 
-static EXEMPTIONS: &[(&str, &str)] = &[("azure_template", "serde")];
+static EXEMPTIONS: &[(&str, &str)] = &[
+    ("azure_core_test", "dotenvy"),
+    ("azure_template", "serde"),
+];
 
 fn main() {
     let manifest_path = std::env::args()
--- a/sdk/core/azure_core_test/Cargo.toml
+++ b/sdk/core/azure_core_test/Cargo.toml
@@ -23,6 +23,7 @@ async-trait.workspace = true
 azure_core = { workspace = true, features = ["test"] }
 azure_core_test_macros.workspace = true
 azure_identity.workspace = true
+dotenvy = "0.15.7"
 futures.workspace = true
 rand.workspace = true
 rand_chacha.workspace = true
--- a/sdk/core/azure_core_test/src/recorded.rs
+++ b/sdk/core/azure_core_test/src/recorded.rs
@@ -68,6 +68,21 @@ pub async fn start(
         ctx.test_recording_file(),
         ctx.test_recording_assets_file(mode),
     );
+
+    // Attempt to read any .env file up to the repo root.
+    #[cfg(not(target_arch = "wasm32"))]
+    if let Ok(path) = crate::find_ancestor_file(ctx.crate_dir, ".env") {
+        tracing::debug!("loading environment variables from {}", path.display());
+
+        use azure_core::error::ResultExt as _;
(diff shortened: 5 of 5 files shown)


heaths:
Need to first merge #2486 and rebase. But I wanted to get @weshaggard's, @benbp's, and @scbedd's take on this. I was trying to think of where we can put a breadcrumb - ideally in an existing file - and `assets.json` makes some sense. Not married to it. Might also be nice to support a repo-wide effort with local overrides when we know it won't work, but maybe that's overkill. Eventually, it would be nice to default all repos into this and allow client libraries to opt out if they know it's not safe. Hopefully, after SFI, most are.

The TestResources changes here are only for demo. Regardless, I'd like to check in the `.env` support to make testing much simpler. This is a direct dependency - hence changes required in `verify-dependencies.rs` - to discourage or even hide that we're taking a direct dependency on `dotenvy` (the active crate since `dotenv` hasn't been touched in 5 years and has a CVE or something out on it) so other product client libraries don't try to use it. It should be treated as an implementation detail of `azure_core_test` which is, as the name implies, only for testing and is authored not to be publishable.

eng/common/TestResources/TestResources-Helpers.ps1
```diff
@@ -1,267 +1,282 @@
...
-    }
-    Write-Host $message
+  if (!$CI -or $SuppressVsoCommands) {
+    return
+  }
+  Write-Host $message
 }
 
 function Retry([scriptblock] $Action, [int] $Attempts = 5) {
-    $attempt = 0
```
LarryOsterman, line 16: Are the indentation changes relevant? Or can they be removed to reduce the amount of churn in the PR?

LarryOsterman (approved):

eng/common/TestResources/TestResources-Helpers.ps1
heaths: Dang. I was trying to be careful not to but have to use a different key chord. Apparently, I forgot once.

Alternatively, @weshaggard can we put an `.editorconfig` in the `eng/common` folder to keep these things consistent? Fighting the formatting is a PITA.

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -215,11 +215,18 @@ try {
     $PSBoundParameters['BaseName'] = $BaseName
 
     # Try detecting repos that support OutFile and defaulting to it
-    if (!$CI -and !$PSBoundParameters.ContainsKey('OutFile') -and $IsWindows) {
+    if (!$CI -and !$PSBoundParameters.ContainsKey('OutFile')) {
         # TODO: find a better way to detect the language
```
scbedd, line 219: Re: `find a better way to detect the language`.

If we include `eng/common/scripts/common.ps1` you will be able to access the variable `$Language` which should have the correct language for the repo.

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -215,11 +215,18 @@ try {
...
-        if (Test-Path "$repositoryRoot/eng/service.proj") {
+        if ($IsWindows -and (Test-Path "$repositoryRoot/eng/service.proj")) {
             $OutFile = $true
-            Log "Detected .NET repository. Defaulting OutFile to true. Test environment settings would be stored into the file so you don't need to set environment variables manually."
+            $isDotnet = $true
+            Log "Detected .NET repository. Defaulting OutFile to true. Test environment settings will be stored into a file so you don't need to set environment variables manually."
+        } elseif (Test-Path "$root/assets.json") {
+            $assets = Get-Content -Raw "$root/assets.json" | ConvertFrom-Json
+            if ($assets.Dotenv) {
```
scbedd, line 226: So you're imagining boolean value here:

```json
{
  "AssetsRepoPrefixPath": "net",
  "TagPrefix": "net/eventgrid/Azure.ResourceManager.EventGrid",
  "Tag": "net/eventgrid/Azure.ResourceManager.EventGrid_442af20ea8",
  "Dotenv": true
}
```
The test proxy should be able to handle this. I will double check that it doesn't throw if we add this.

If anything, we should remove the containing `IsWindows` and `eng/service.proj` path, because if the assets.json has `Dotenv` present then we should save the .env without caring about the other possible values.

scbedd: [Looks good locally but double checking integration tests](https://github.com/Azure/azure-sdk-tools/pull/10327)

heaths: Well, the `$IsWindow` check and friends are there for back compat. On Windows, it has long supported a *custom* `.env`-like file, but it's JSON-encoded and then DPAPI-encryped, which is why it's only on Windows. And it's only for Azure/azure-sdk-for-net because no other language cared to do the work to call DPAPI which, in .NET, has a convenient API - at least on Windows.

So, I don't want to break that. In fact, I wanted to get this PR in because I miss what I had in .NET on Windows (for .NET, VS > VSCode).

That's also why I do that check first. I want to maintain backcompat so even if a library on Windows authors `"Dotenv": true`, I'd think @JoshLove-msft and folk would probably want to maintain backcompat.

scbedd: Ok fair point. I don't oppose this PR then. Wes is out this week so @benbp for FYI.

heaths:
I've marked this one ready after rebasing and split off the proposed script changes to #2498. We wouldn't want to merge those into this repo anyway, but it allows us to test them against a repo that will otherwise support `.env` files.

LarryOsterman (approved):

PR #2492, opened 2025-04-16
[Storage] [BlobClient] `set_tier` (standard tiers only)

vincenttran-msft:
Premium tiers will be covered in a follow-up PR as those are only applicable to Page blobs (which cannot be spun up currently)

### DIFF
--- a/sdk/storage/azure_storage_blob/src/clients/blob_client.rs
+++ b/sdk/storage/azure_storage_blob/src/clients/blob_client.rs
@@ -257,7 +257,6 @@ impl BlobClient {
         tier: AccessTier,
         options: Option<BlobClientSetTierOptions<'_>>,
     ) -> Result<Response<()>> {
-        let response = self.client.set_tier(tier, options).await?;
-        Ok(response)
+        self.client.set_tier(tier, options).await
     }
 }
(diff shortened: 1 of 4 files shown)


jalauzon-msft (approved):

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -243,4 +244,20 @@ impl BlobClient {
...
+    ///
+    /// * `tier` - The tier to be set on the blob.
+    /// * `options` - Optional configuration for the request.
+    pub async fn set_tier(
+        &self,
+        tier: AccessTier,
+        options: Option<BlobClientSetTierOptions<'_>>,
+    ) -> Result<Response<()>> {
+        let response = self.client.set_tier(tier, options).await?;
+        Ok(response)
```
heaths, line 261: Should just be able to call: `self.client.set_tier(tier, options).await`

heaths (approved):

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: Good callout here- will apply this in a sweeping change next PR to all applicable APIs

heaths (approved):

PR #2498, opened 2025-04-18
Script changes for .env file support

heaths:
Builds on #2487 with just the script changes required

### DIFF
--- a/eng/common/TestResources/New-TestResources.ps1
+++ b/eng/common/TestResources/New-TestResources.ps1
@@ -203,11 +203,18 @@ try {
     $PSBoundParameters['BaseName'] = $BaseName
 
     # Try detecting repos that support OutFile and defaulting to it
-    if (!$CI -and !$PSBoundParameters.ContainsKey('OutFile') -and $IsWindows) {
+    if (!$CI -and !$PSBoundParameters.ContainsKey('OutFile')) {
         # TODO: find a better way to detect the language
-        if (Test-Path "$repositoryRoot/eng/service.proj") {
+        if ($IsWindows -and (Test-Path "$repositoryRoot/eng/service.proj")) {
             $OutFile = $true
-            Log "Detected .NET repository. Defaulting OutFile to true. Test environment settings would be stored into the file so you don't need to set environment variables manually."
+            $isDotnet = $true
+            Log "Detected .NET repository. Defaulting OutFile to true. Test environment settings will be stored into a file so you don't need to set environment variables manually."
+        } elseif (Test-Path "$root/assets.json") {
+            $assets = Get-Content -Raw "$root/assets.json" | ConvertFrom-Json
+            if ($assets.Dotenv) {
+              $OutFile = $true
+              Log "Defaulting OutFile to true since '$root/assets.json' dictates. Test environment settings will be stored in '$root/.env' so you don't need to set environment variables manually."
+            }
         }
     }
 
@@ -860,13 +867,17 @@ Force creation of resources instead of being prompted.
 
 .PARAMETER OutFile
 Save test environment settings into a .env file next to test resources template.
-The contents of the file are protected via the .NET Data Protection API (DPAPI).
-This is supported only on Windows. The environment file is scoped to the current
-service directory.
 
+On Windows in the Azure/azure-sdk-for-net repository,
+he contents of the file are protected via the .NET Data Protection API (DPAPI).
+The environment file is scoped to the current service directory.
 The environment file will be named for the test resources template that it was
 generated for. For ARM templates, it will be test-resources.json.env. For
 Bicep templates, test-resources.bicep.env.
+
+If enabled in other repositories or on other platforms, a .env file is written
+unencrypted to the service directory if and only if it is in .gitignore.
+Enable -OutFile automatically by putting `"Dotenv": true` in your assets.json file.
 
 .PARAMETER SuppressVsoCommands
 By default, the -CI parameter will print out secrets to logs with Azure Pipelines log
--- a/eng/common/TestResources/New-TestResources.ps1
+++ b/eng/common/TestResources/New-TestResources.ps1
@@ -209,12 +209,9 @@ try {
         if ($IsWindows -and $Language -eq 'dotnet') {
             $OutFile = $true
             Log "Detected .NET repository. Defaulting OutFile to true. Test environment settings will be stored into a file so you don't need to set environment variables manually."
-        } elseif (Test-Path "$root/assets.json") {
-            $assets = Get-Content -Raw "$root/assets.json" | ConvertFrom-Json
-            if ($assets.Dotenv) {
-              $OutFile = $true
-              Log "Defaulting OutFile to true since '$root/assets.json' dictates. Test environment settings will be stored in '$root/.env' so you don't need to set environment variables manually."
-            }
+        } elseif ($SupportsDotenv) {
+            $OutFile = $true
+            Log "Repository supports reading .env files. Defaulting OutFile to true. Test environment settings may be stored in a .env file so they are read by tests automatically."
         }
     }
 
@@ -349,10 +346,10 @@ try {
         if ($context.Account.Type -eq 'User') {
             # Support corp tenant and TME tenant user id lookups
             $user = Get-AzADUser -Mail $context.Account.Id
-            if ($user -eq $null -or !$user.Id) {
+            if ($null -eq $user -or !$user.Id) {
                 $user = Get-AzADUser -UserPrincipalName $context.Account.Id
             }
-            if ($user -eq $null -or !$user.Id) {
+            if ($null -eq $user -or !$user.Id) {
                 throw "Failed to find entra object ID for the current user"
             }
             $ProvisionerApplicationOid = $user.Id
@@ -426,10 +423,10 @@ try {
 
         # Support corp tenant and TME tenant user id lookups
         $userAccount = (Get-AzADUser -Mail (Get-AzContext).Account.Id)
-        if ($userAccount -eq $null -or !$userAccount.Id) {
+        if ($null -eq $userAccount -or !$userAccount.Id) {
             $userAccount = (Get-AzADUser -UserPrincipalName (Get-AzContext).Account)
         }
-        if ($userAccount -eq $null -or !$userAccount.Id) {
+        if ($null -eq $userAccount -or !$userAccount.Id) {
             throw "Failed to find entra object ID for the current user"
         }
         $TestApplicationOid = $userAccount.Id
--- a/eng/common/TestResources/New-TestResources.ps1
+++ b/eng/common/TestResources/New-TestResources.ps1
@@ -122,6 +122,7 @@ param (
     $NewTestResourcesRemainingArguments
 )
 
+. (Join-Path $PSScriptRoot .. scripts common.ps1)
 . (Join-Path $PSScriptRoot .. scripts Helpers Resource-Helpers.ps1)
 . $PSScriptRoot/TestResources-Helpers.ps1
 . $PSScriptRoot/SubConfig-Helpers.ps1
@@ -205,9 +206,8 @@ try {
     # Try detecting repos that support OutFile and defaulting to it
     if (!$CI -and !$PSBoundParameters.ContainsKey('OutFile')) {
         # TODO: find a better way to detect the language
-        if ($IsWindows -and (Test-Path "$repositoryRoot/eng/service.proj")) {
+        if ($IsWindows -and $Language -eq 'dotnet') {
             $OutFile = $true
-            $isDotnet = $true
             Log "Detected .NET repository. Defaulting OutFile to true. Test environment settings will be stored into a file so you don't need to set environment variables manually."
         } elseif (Test-Path "$root/assets.json") {
             $assets = Get-Content -Raw "$root/assets.json" | ConvertFrom-Json
@@ -869,7 +869,7 @@ Force creation of resources instead of being prompted.
 Save test environment settings into a .env file next to test resources template.
 
 On Windows in the Azure/azure-sdk-for-net repository,
-he contents of the file are protected via the .NET Data Protection API (DPAPI).
+the contents of the file are protected via the .NET Data Protection API (DPAPI).
 The environment file is scoped to the current service directory.
 The environment file will be named for the test resources template that it was
 generated for. For ARM templates, it will be test-resources.json.env. For
--- a/eng/common/TestResources/Remove-TestResources.ps1
+++ b/eng/common/TestResources/Remove-TestResources.ps1
@@ -235,6 +235,7 @@ if ($ServiceDirectory) {
 
     # Make sure environment files from New-TestResources -OutFile are removed.
     Get-ChildItem -Path $root -Filter "$ResourceType-resources.json.env" -Recurse | Remove-Item -Force:$Force
+    Get-ChildItem -Path $root -Filter ".env" -Recurse | Remove-Item -Force:$Force
 }
 
 $verifyDeleteScript = {
--- a/eng/common/TestResources/TestResources-Helpers.ps1
+++ b/eng/common/TestResources/TestResources-Helpers.ps1
@@ -1,282 +1,282 @@
 function Log($Message) {
-  Write-Host ('{0} - {1}' -f [DateTime]::Now.ToLongTimeString(), $Message)
+    Write-Host ('{0} - {1}' -f [DateTime]::Now.ToLongTimeString(), $Message)
 }
 
 # vso commands are specially formatted log lines that are parsed by Azure Pipelines
 # to perform additional actions, most commonly marking values as secrets.
 # https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands
 function LogVsoCommand([string]$message) {
-  if (!$CI -or $SuppressVsoCommands) {
-    return
-  }
-  Write-Host $message
+    if (!$CI -or $SuppressVsoCommands) {
+        return
+    }
+    Write-Host $message
 }
 
 function Retry([scriptblock] $Action, [int] $Attempts = 5) {
-  $attempt = 0
-  $sleep = 5
+    $attempt = 0
+    $sleep = 5
 
-  while ($attempt -lt $Attempts) {
-    try {
-      $attempt++
-      return $Action.Invoke()
-    }
-    catch {
-      if ($attempt -lt $Attempts) {
-        $sleep *= 2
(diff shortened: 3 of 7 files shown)


eng/common/TestResources/New-TestResources.ps1
```diff
@@ -215,11 +215,18 @@ try {
     $PSBoundParameters['BaseName'] = $BaseName
 
     # Try detecting repos that support OutFile and defaulting to it
-    if (!$CI -and !$PSBoundParameters.ContainsKey('OutFile') -and $IsWindows) {
+    if (!$CI -and !$PSBoundParameters.ContainsKey('OutFile')) {
         # TODO: find a better way to detect the language
```
heaths, line 219: @scbedd noted on the previous PR: https://github.com/Azure/azure-sdk-for-rust/pull/2487#discussion_r2045579845

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -859,14 +866,18 @@ Force creation of resources instead of being prompted.
...
-service directory.
 
+On Windows in the Azure/azure-sdk-for-net repository,
+the contents of the file are protected via the .NET Data Protection API (DPAPI).
+The environment file is scoped to the current service directory.
 The environment file will be named for the test resources template that it was
 generated for. For ARM templates, it will be test-resources.json.env. For
 Bicep templates, test-resources.bicep.env.
 
+If enabled in other repositories or on other platforms, a .env file is written
```
LarryOsterman, line 877: It might be helpful to leave a breadcrumb pointing out that the `.gitignore` logic is *not* encapsulated in this script but instead is implemented in the test-resources-helper script. 

That's especially true since the "On Windows in the Azure/azure-sdk-for-net" part of the check IS implemented in this file.

I originally had left a comment (in a previous iteration of the PR) asking if the .gitignore check was going to happen in a subsequent PR because I didn't find it until later (and then I deleted that comment because I had found the check).

LarryOsterman (approved):

LarryOsterman:
QQ: Don't eng/common changes have to go into the azure-sdk repo and not azure-sdk-for-rust?

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -215,11 +215,18 @@ try {
...
-    if (!$CI -and !$PSBoundParameters.ContainsKey('OutFile') -and $IsWindows) {
+    if (!$CI -and !$PSBoundParameters.ContainsKey('OutFile')) {
         # TODO: find a better way to detect the language
-        if (Test-Path "$repositoryRoot/eng/service.proj") {
+        if ($IsWindows -and (Test-Path "$repositoryRoot/eng/service.proj")) {
             $OutFile = $true
-            Log "Detected .NET repository. Defaulting OutFile to true. Test environment settings would be stored into the file so you don't need to set environment variables manually."
+            $isDotnet = $true
+            Log "Detected .NET repository. Defaulting OutFile to true. Test environment settings will be stored into a file so you don't need to set environment variables manually."
+        } elseif (Test-Path "$root/assets.json") {
```
weshaggard, line 224: I'm not a big fan of this (or the .NET detection). Any reason to not make this a parameter to the script and set the default accordingly in the template?

eng/common/TestResources/Remove-TestResources.ps1
```diff
@@ -235,6 +235,7 @@ if ($ServiceDirectory) {
 
     # Make sure environment files from New-TestResources -OutFile are removed.
     Get-ChildItem -Path $root -Filter "$ResourceType-resources.json.env" -Recurse | Remove-Item -Force:$Force
+    Get-ChildItem -Path $root -Filter ".env" -Recurse | Remove-Item -Force:$Force
```
weshaggard, line 238: Why do we want to remove all env files?

eng/common/TestResources/TestResources-Helpers.ps1
```diff
@@ -203,19 +203,34 @@ function SetDeploymentOutputs(
     $deploymentOutputs = BuildDeploymentOutputs $serviceName $azContext $deployment $deploymentEnvironmentVariables
 
     if ($OutFile) {
-        if (!$IsWindows) {
-            Write-Host 'File option is supported only on Windows'
-        }
+        if ($IsWindows -and $isDotnet) {
```
weshaggard, line 206: Is the $isDotNet variable in scope here?

eng/common/TestResources/TestResources-Helpers.ps1
```diff
@@ -203,19 +203,34 @@ function SetDeploymentOutputs(
...
-        Set-Content $outputFile -Value $protectedBytes -AsByteStream -Force
+            Write-Host "Test environment settings`n$environmentText`nstored into encrypted $outputFile"
+        }
+        else {
+            $outputFile = $templateFile.originalFilePath | Split-Path | Join-Path -ChildPath '.env'
+
+            # Make sure the file would be ignored.
+            git check-ignore -- "$outputFile" > $null
+            if ($?) {
+                $environmentText = foreach ($kv in $deploymentOutputs.GetEnumerator()) {
```
weshaggard, line 223: @benbp do you know how to we would want to add the linter to ensure none of these are secrets?

eng/common/TestResources/New-TestResources.ps1
heaths: The .NET detection has always been there. This is keeping it compatible. I'll let them deal with switching to unencrypted `.env` files and deprecate their current JSON-encoded `.env`-like DPAPI-encrypted file support on their own time.

But what specifically is "this"? I need to test if the path exists before I read it in a non-terminating case; that is, if the file exists and I can successfully deserialize the check for the value I need, write the `.env` file; otherwise, keep doing what it was doing. You wanted this to be conditional. It has to be until other languages support `.env` files for this.

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -215,11 +215,18 @@ try {
...
 
     # Try detecting repos that support OutFile and defaulting to it
-    if (!$CI -and !$PSBoundParameters.ContainsKey('OutFile') -and $IsWindows) {
+    if (!$CI -and !$PSBoundParameters.ContainsKey('OutFile')) {
         # TODO: find a better way to detect the language
-        if (Test-Path "$repositoryRoot/eng/service.proj") {
+        if ($IsWindows -and (Test-Path "$repositoryRoot/eng/service.proj")) {
             $OutFile = $true
-            Log "Detected .NET repository. Defaulting OutFile to true. Test environment settings would be stored into the file so you don't need to set environment variables manually."
+            $isDotnet = $true
```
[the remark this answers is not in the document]
heaths: We can't. DPAPI is Windows-only. With SFI, there's also no need for most libraries - maybe still those supporting and testing connection strings, shared secrets, etc., which is why you asked me to make this opt-in for now.

eng/common/TestResources/TestResources-Helpers.ps1
heaths: Yes. Variable are in scope for the current and child scopes. Since we source this file, it's in the same scpe as the rest of the script.

eng/common/TestResources/New-TestResources.ps1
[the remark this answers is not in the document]
weshaggard: No I was meaning instead of calling this variable isDotNet it should be called something like shouldEncryptEnv

eng/common/TestResources/New-TestResources.ps1
weshaggard: I understand you want to leave the .NET code as is but for the new opt-in instead of looking for the assets.json file I would prefer that we add a new parameter to the script something along the lines of outputEnvFile and you can start passing it for the Rust usages.

eng/common/TestResources/New-TestResources.ps1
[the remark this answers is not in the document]
heaths: It was already .NET-on-Windows-specific. I'd rather not over-generalize. The support is solely for .NET on Windows, and if they want to expand that, I say just take the conditions out and use a normal `.env` file like everyone else (hopefully) will. /cc @JoshLove-msft

eng/common/TestResources/New-TestResources.ps1
heaths: That helper script should *never* be used directly and these comments are exposed as help to the end user, who would never know about the helper script (like POD comments for Perl, which inspired the PowerShell team in part).

eng/common/TestResources/New-TestResources.ps1
heaths: It shouldn't be something the developer has to do. It should be automatic. Or they will forget to do it and we'll get all sorts of support requests. That's how .NET on Windows support already works now. We want the users (and CI) to have the right experience by default. That's why I wrote the scripts the way they were in the first place. Complicating them by requiring more input isn't going to benefit anyone.

eng/common/TestResources/TestResources-Helpers.ps1
weshaggard: @benbp @heaths instead of doing this inside of the script directly and checking for the existence of the assets file. Perhaps we have a repo wide post-script (or a post deployment function in language-settings) we can look for and execute it if it exists. Then in the rust version you can output the env file.

heaths: I don't think that's a great idea. The point of this was to eventually bring it to *all* languages so they can all use what .NET on Windows has effectively had for years (and to replace that as well). This isn't supposed to be language-specific. I made the script changes here only to test them in a repo that supports `.env` files.

And apparently for JS and Python do as well, but the script doesn't actually create them currently but devs can create them manually.

I wrote the scripts were first and foremost to unify resource provisioning for *devs*. How you provisioned test resources varied from language to language and even package to package within the same repo. We bolted in CI support later on. We can automate CIs to do whatever and however, but this should be easy - transparent, really - to devs who are contributing.

/cc @RickWinter

weshaggard: We need some way of opting in repos and even service/packages without looking for an assets file in the root. This extension point can provide this mechanism for opting in. I agree with you that we might not want the env file writing to happen in each language repo independently, but we at least need a way to configure the opt-in, which this will provide for us. We can of course still share the implementation of writing the env file.

heaths: There's no assets file in the root. It's looking at the `sdk/{service-directory}/assets.json` we already use for test-proxy. As we talked about yesterday f2f, we could use the `LanguageSettings.ps1` (or whatever it's called) as the repo-wide one, then override as needed from the aforementioned `assets.json`.

weshaggard: We should use the LanguageSettings.ps1 but if we do that I don't think we need to look at for the assets file. We need some other setting somewhere to key off of.

heaths: I don't think we should use a centralized exception list for opt-out packages, though. So then we're talking about a service directory-specific file and we already have `assets.json` that tracks test asset information...for which this is directly related. Why is looking at `assets.json` for something or other so bad? They are inextricably linked: test asset information - where to get it, and where to put it in a sense.

weshaggard: assets.json is orthogonal to whether or not our deployment should output an env file and to me does not really map well. I don't want the presents of the file to be the only check as they exists in other places.

Is there some setting in the test-resources.json or bicep file we could use to let the deployment script know we want to output the values to an env file?

heaths: > assets.json is orthogonal to whether or not our deployment should output an env file

Why? It's a place to put test-related settings we happened to use for test-proxy. Why does it have to be limited to just test-proxy? Do we need to drop files everywhere every time we want to make a related but different change? That's awfully short-sighted and is going to be a mess for discoverability and maintenance. How are partners supposed to learn about all these files and what they do as opposed to few docs that document more behaviors (like schema docs)?

> Is there some setting in the test-resources.json or bicep file

No, because the JSON schema is validated and extra stuff errs (I tried), and bicep doesn't allow extension like that.

weshaggard: Do the assets.json file allow for additional properties that can be read? My bigger concern is that we are using the existence of that file as the opt-in, if we were to read the configuration from it, I would be less concerned.

>No, because the JSON schema is validated and extra stuff errs (I tried), and bicep doesn't allow extension like that.

Could we have a custom output that is set to true or false that we could query to see if they opted in?

heaths: I've already stated above we can first rely on `LanguageSettings.ps1` for repo opt-in. Secondary to that would be `assets.json` to either opt in or out; but, yes, `assets.json` does support other fields. Not only did this change work when I tried it locally under a number of scenarios, but @scbedd verified in code that it doesn't err on other properties (the default, IIRC, for both `Newtonsoft.Json` and `System.Text.Json`) but thought it was a good idea to consolidate. While not a 100% overlap, they are very closely related: we provision test resources, in part, to be used by test-proxy to record.

> Could we have a custom output that is set to true or false that we could query to see if they opted in?

You mean just write something to the console? Sure. I don't see a problem with that. But if you meant a PS return value yeah, that might cause some issues.

weshaggard: Ok so @benbp and I chatted and it does look like there is a bicep linter rule to detect secrets https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/linter-rule-outputs-should-not-contain-secrets. So instead of an opt-in we can run bicep lint and if it is passing we can output an env file.

heaths: Run `bicep lint` on what? We haven't written a file yet. A quick usage check shows it requires a file. I also question whether it's good enough. It could miss things that are secrets or flag things that aren't e.g., ephemeral keys like Key Vault creates.

I think repos need to opt into it. At the very least, those repos need to read `.env` files and will need to output those env vars as they are now (to the host) so people can use them. We can't just flip this on.

weshaggard: run bicep lint on the bicep file and if it doesn't complain about outputs complaining secrets then go ahead and output the .env file.

heaths: I gotcha. I thought it was for scanning for secrets, but like you explained offline, it for scanning for patterns with secrets in the bicep. And, yes, I agree, it's a good carrot for people to convert if they want .env files. Will work on this soon.

eng/common/TestResources/TestResources-Helpers.ps1
```diff
@@ -203,19 +230,40 @@ function SetDeploymentOutputs(
...
+            Write-Host "Test environment settings`n$environmentText`nstored into encrypted $outputFile"
+        }
+        else {
+            $bicepTemplateFile = $templateFile.originalFilePath
 
-        $environmentText = $deploymentOutputs | ConvertTo-Json;
-        $bytes = [System.Text.Encoding]::UTF8.GetBytes($environmentText)
-        $protectedBytes = [Security.Cryptography.ProtectedData]::Protect($bytes, $null, [Security.Cryptography.DataProtectionScope]::CurrentUser)
+            # Make sure the file would not write secrets to .env file.
+            if (!(LintBicepFile $bicepTemplateFile)) {
```
weshaggard, line 248: Don't you need to check to see if it is a bicep file as opposed to an arm template?

eng/common/TestResources/TestResources-Helpers.ps1
```diff
@@ -203,19 +230,40 @@ function SetDeploymentOutputs(
...
+            $environmentText = $deploymentOutputs | ConvertTo-Json;
+            $bytes = [System.Text.Encoding]::UTF8.GetBytes($environmentText)
+            $protectedBytes = [Security.Cryptography.ProtectedData]::Protect($bytes, $null, [Security.Cryptography.DataProtectionScope]::CurrentUser)
 
-        $outputFile = "$($templateFile.originalFilePath).env"
+            Set-Content $outputFile -Value $protectedBytes -AsByteStream -Force
+
+            Write-Host "Test environment settings`n$environmentText`nstored into encrypted $outputFile"
+        }
+        else {
```
weshaggard, line 244: Don't you need to check SupportsDotenv?

eng/common/scripts/common.ps1
```diff
@@ -24,6 +24,9 @@ $PackageRepository = "Unknown"
 $packagePattern = "Unknown"
 $MetadataUri = "Unknown"
 
+# Whether the language repo supports automatically loading .env file generated from TestResources scripts.
+$SupportsDotenv = $false
```
weshaggard, line 28: Perhaps we name this SupportsDotenvForTestResourceOutputs or something to make it clear what it is intended for.

heaths: The comment directly above it didn't do that? Longer variables just invites the increase risk of typos. I don't see a clear advantage, especially when we have variables like `$packagePattern` just above that. No idea based on the variable name what that does.

eng/common/TestResources/TestResources-Helpers.ps1
heaths: No, because at this point the user explicitly passed `-OutFile` or we flipped it based on the previous check of `$SupportsDotenv`. The only reason this inner check exists is to - for now - keep supporting DPAPI use for the .NET repo on Windows; otherwise, it'd just be this block (and will be eventually) under the `if ($OutFile) {` check.

eng/common/scripts/common.ps1
weshaggard: The comment there helps but it doesn't help in the usages. I agree I don't like long names, but I do like self describing names, especially for global variables like this.

eng/common/TestResources/TestResources-Helpers.ps1
weshaggard: Yes I missed the OutFile is now holding the setting.

eng/common/scripts/common.ps1
heaths: Alright, I'll think of something. But I hope this same applies for other vars going forward because several of those above it aren't self-explanatory either...and have no comments describing them, unlike mine.

heaths: Went with `$SupportsTestResourcesDotenv`.

heaths:
This was migrated to Azure/azure-sdk-tools and successfully merged/rolled out.

PR #2500, opened 2025-04-18
Remove federated_credentials_flow

### LINKED ISSUE #2282: Federated credential flow should be renamed/repurposed as client assertion credential

chlowell:
Closes #2282 by moving the client_assertion implementation into ClientAssertionCredential


sdk/identity/azure_identity/src/credentials/client_assertion_credentials.rs
```diff
@@ -1,22 +1,34 @@
...
     error::{ErrorKind, ResultExt},
+    http::{
+        headers::{self, content_type},
+        Method, Request, StatusCode,
+    },
+    Error,
 };
 use std::{fmt::Debug, str, sync::Arc, time::Duration};
 use time::OffsetDateTime;
+use url::{form_urlencoded, Url};
```
heaths, line 19: `Url` should be imported from `azure_core` to be safe, should we ever define our own. If you need `form_urlencoded`, you probably will need it from `url` as well (which itself is a re-export).

sdk/identity/azure_identity/src/credentials/client_assertion_credentials.rs
```diff
@@ -68,59 +80,92 @@ impl<C: ClientAssertion> ClientAssertionCredential<C> {
...
         options: Option<ClientAssertionCredentialOptions>,
     ) -> azure_core::Result<Self> {
+        validate_tenant_id(&tenant_id)?;
+        validate_not_empty(&client_id, "no client ID specified")?;
+        let options = options.unwrap_or_default().credential_options;
+        let endpoint = options
+            .authority_host()?
+            .join(&format!("/{tenant_id}/oauth2/v2.0/token"))
+            .with_context(ErrorKind::DataConversion, || {
+                format!("tenant_id '{tenant_id}' could not be URL encoded")
```
heaths, line 90: Nit: I don't quote this elsewhere. Doesn't add much, really. It's not like it'll have spaces.

sdk/identity/azure_identity/src/credentials/client_assertion_credentials.rs
```diff
@@ -68,59 +80,92 @@ impl<C: ClientAssertion> ClientAssertionCredential<C> {
...
-        .context(ErrorKind::Credential, "request token error")?;
-        Ok(res)
+    async fn get_token_impl(&self, scopes: &[&str]) -> azure_core::Result<AccessToken> {
+        let mut req = Request::new(self.endpoint.clone(), Method::Post);
+        req.insert_header(
+            headers::CONTENT_TYPE,
+            content_type::APPLICATION_X_WWW_FORM_URLENCODED,
+        );
+        let assertion = self.assertion.secret().await?;
+        let encoded: String = form_urlencoded::Serializer::new(String::new())
```
heaths, line 108: Ah, you import it for url-encoded bodies. Maybe we should re-export it from `azure_core` as well. For URIs you can actually get a mutable one from `Url`, but that's clearly not what you need here.

heaths (approved):

sdk/identity/azure_identity/src/credentials/client_assertion_credentials.rs
chlowell: Maybe. I suppose it depends on the problem you hope to solve by re-exporting. I doubt the export would often be used, because I don't find many form bodies in the Go repo and eventually this crate won't craft requests

heaths: I suppose we don't need to make this generally available to crate devs - only Identity - so probably don't re-export then. The idea is that we want to limit Azure SDK crate devs to types on `azure_core` - even hiding anything "typespec" in general.

heaths (approved):

PR #2502, opened 2025-04-18
[Storage] Navigation Methods Between Clients + Storage Test Helpers

vincenttran-msft:

- Navigation methods between clients (`get_container_client`, `get_blob_client`)
- Cleans up the API calls in the handwritten code from unnecessary wrapping/rewrapping
- Refactors test code
              -  Adds helpers to return instrumented instances of each client
              - Adds a basic "upload blob" since many tests need a blob to exists but don't actually care much about the contents

### DIFF
--- a/sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
+++ b/sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
@@ -112,8 +112,7 @@ impl BlobContainerClient {
         &self,
         options: Option<BlobContainerClientCreateOptions<'_>>,
     ) -> Result<Response<()>> {
-        let response = self.client.create(options).await?;
-        Ok(response)
+        self.client.create(options).await
     }
 
     /// Sets user-defined metadata for the specified container as one or more name-value pairs. Each call to this operation
@@ -127,8 +126,7 @@ impl BlobContainerClient {
         &self,
         options: Option<BlobContainerClientSetMetadataOptions<'_>>,
     ) -> Result<Response<()>> {
-        let response = self.client.set_metadata(options).await?;
-        Ok(response)
+        self.client.set_metadata(options).await
     }
 
     /// Marks the specified container for deletion. The container and any blobs contained within are later deleted during garbage collection.
@@ -140,8 +138,7 @@ impl BlobContainerClient {
         &self,
         options: Option<BlobContainerClientDeleteOptions<'_>>,
     ) -> Result<Response<()>> {
-        let response = self.client.delete(options).await?;
-        Ok(response)
+        self.client.delete(options).await
     }
 
     /// Returns all user-defined metadata and system properties for the specified container.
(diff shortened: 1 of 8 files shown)


sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -59,18 +60,39 @@ impl BlobContainerClient {
...
         })
     }
 
+    /// Returns a new instance of BlobClient.
+    ///
+    /// # Arguments
+    ///
+    /// * `blob_name` - The name of the blob.
+    /// * `options` - Optional configuration for the client.
+    pub fn get_blob_client(
```
LarryOsterman, line 82: Naming nit: Calling this function "get_xxx" implies that it is retrieving an existing something and not that it is creating something new. I recognize that the equivalent function is called `getXxx` in all the P0 languages, and C++, but in Go it is `NewXxx` so there is at least a convention in Go of calling "a function which creates a Foo" with "NewFoo".

vincenttran-msft: This a good callout, and I agree that `get_xxx` implies fetching something rather than new-ing something up. I wonder if this convention holds up in Rust, or would it be odd/confusing to have `new_blob_client` inside of `BlobContainerClient`? 

One thing to also note is that it seems like customers are used to `getXxx` from P0 languages but I am okay with cx having to read the documentation a little harder to find that `newXxx` is equivalent if that design is objectively better.
cc: @heaths @jalauzon-msft

sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
```diff
@@ -52,14 +53,34 @@ impl BlobServiceClient {
...
+        &self,
+        container_name: String,
+        options: Option<BlobContainerClientOptions>,
+    ) -> Result<BlobContainerClient> {
+        BlobContainerClient::new(
+            self.endpoint().as_str(),
+            container_name,
+            self.credential.clone(),
+            options,
+        )
```
heaths, line 81: Here and anywhere else: it'd be more efficient to just create the struct, as long as the fields are `pub(crate)` (effectively `internal` as in C#). As it is, you're passing a string for the endpoint just to convert it back to a URL. If you construct the struct, you can simply clone it:

```rust
BlobContainerClient {
  container_name,
  credential: credential.clone(),
  options,
}
```
This is only a minor perf improvement so I wouldn't go out of your way to make changes to make this work, but something to consider.
```

File: sdk/storage/azure_storage_blob_test/src/lib.rs [hunk 3]
```diff
@@ -1,10 +1,22 @@
...
+    BlobContainerClientOptions, BlobServiceClient, BlobServiceClientOptions,
+};
 
-pub fn recorded_test_setup(recording: &Recording) -> (ClientOptions, String) {
+/// Takes in a Recording instance and returns an instrumented options bag and endpoint.
+///
+/// # Arguments
+///
+/// * `recording` - A reference to a Recording instance.
+pub fn _recorded_test_setup(recording: &Recording) -> (ClientOptions, String) {
```
heaths, line 19: Why prefaced with an underscore? If internal, just remove `pub` or make it `pub(crate)`?

heaths (approved):

heaths (approved):

jalauzon-msft (approved):

PR #2516, opened 2025-04-23
Omit fields when deriving SafeDebug by default

### LINKED ISSUE #1707: Redact Debug printing by default

heaths:
Resolves #1707. Callers can enable the `debug` feature of `typespec_client_core` or `azure_core` to emit normal `core::fmt::Debug` formatting.


sdk/typespec/typespec_client_core/Cargo.toml
```diff
@@ -45,6 +45,7 @@ typespec_macros.path = "../typespec_macros"
 
 [features]
 default = ["http", "json", "reqwest", "reqwest_deflate", "reqwest_gzip"]
+debug = ["typespec_macros?/debug"]
```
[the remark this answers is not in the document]
heaths: That's intentional and supported.

sdk/typespec/typespec_client_core/src/fmt.rs
```diff
@@ -25,7 +30,7 @@ pub use typespec_macros::SafeDebug;
 /// };
 /// assert_eq!(format!("{model:?}"), "MyModel { .. }");
 /// ```
-pub trait SafeDebug: Debug {}
+pub trait SafeDebug: crate::__private::Sealed {}
```
analogrelay, line 33: Doesn't sealing the trait make it impossible for other crates to impl, even via `#[derive]`?

analogrelay: Ah, I see (and/or remember from earlier). The trait isn't really intended to be derived, it's just a marker trait for the derive macro to run.

It might be confusing to have the trait at all though, since using it in a bound would not have the desired effect of limiting a generic type to ones that have implemented `SafeDebug`. It also makes it more difficult to provide any methods on `SafeDebug` to "unredact" the debug output later.

heaths: I tried to remove it but seems a derive macro doesn't like that. But, after a quick search, it seems it's possible. Let me try. I definitely wanted to remove the trait to avoid confusion. That's why I ended up sealing it instead.

LarryOsterman (approved):

PR #2519, opened 2025-04-23
Add AzureDeveloperCliCredential

### LINKED ISSUE #2112: Add azd credential support
Support authentication with azd

chlowell:
Closes #2112

### DIFF
--- a/sdk/identity/azure_identity/src/azure_developer_cli_credential.rs
+++ b/sdk/identity/azure_identity/src/azure_developer_cli_credential.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-// cspell:ignore SYSTEMROOT
+// cspell:ignore SYSTEMROOT workdir
 
 use crate::{env::Env, validate_scope, validate_tenant_id};
 use azure_core::{
--- a/sdk/identity/azure_identity/src/azure_developer_cli_credential.rs
+++ b/sdk/identity/azure_identity/src/azure_developer_cli_credential.rs
@@ -1,0 +1,249 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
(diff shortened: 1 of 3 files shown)


sdk/identity/azure_identity/src/azure_developer_cli_credential.rs
```diff
@@ -0,0 +1,249 @@
...
+                Error::message(
+                    ErrorKind::Credential,
+                    "SYSTEMROOT environment variable not set",
+                )
+            })?;
+            (system_root, "cmd", "/C")
+        } else {
+            ("/bin".to_string(), "/bin/sh", "-c")
+        };
+        let command_string = format!("cd {workdir} && {command}");
```
heaths, line 110: Faster to set the working directory. Do we not expose that currently for the executor? Maybe we should.

chlowell: We don't. I could go either way on whether we should. It's not required for any scenario I know about. I expect azure_identity will be the only crate that execs procs in production (we have one Go module that does so only in its tests) and it will always do so via shell like this. I'm happy with the status quo until we find a scenario that requires setting the workdir directly.

If you're curious--I was--I benchmarked `cd` vs setting the workdir in Go just to get an idea of the magnitude of the perf difference. So, here's some trivial anecdata for your Thursday morning: `cd` was faster by up to a few hundred μs and allocated less. I guess setting the workdir requires a larger data structure and/or an additional syscall (I'm not familiar with the kernel API).

heaths: Interesting. I was thinking of the time the shell would need to parse the extra string/command, execute it, check status (since `&&` was used), and run the next command. May as well leave it, then. If we see a need for setting the PWD elsewhere, we can revisit but I agree we'll probably only need this for `azure_identity`.

Actually, that raises another thought, but I'll start a side thread with you and @LarryOsterman. Not a blocker for this PR.

heaths (approved):

PR #2522, opened 2025-04-24
Perf test framework

gearama:
Add a test with some params, add perf-tests.yml in order to get it to be executed by the perf automation.


sdk/core/azure_core/README.md
```diff
@@ -201,6 +201,20 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+Azure Core includes performance benchmarks to help monitor the performance of critical components. These benchmarks are written using the [criterion](https://crates.io/crates/criterion) crate, a statistics-driven benchmarking library.
+
+To run the performance tests:
+
+```bash
+cd sdk/core/azure_core
+cargo bench
+```
+
+The benchmarks measure critical operations like serialization/deserialization, HTTP request processing, and token acquisition to ensure optimal performance of the SDK.
```
LarryOsterman, line 215: How do I find the results of these performance tests? Are they run with a regular cadence?

LarryOsterman, line 215: I'm not seeing the tests for serialization/deserialization, HTTP request processing and token acquisition. Are they coming in a later PR?

LarryOsterman (approved):

sdk/core/azure_core/README.md
gearama: they are displayed as a summary in the cmd when running cargo bench in either the root folder ( though they will be drowned by other noise or in the sdk/core/azure-core folder 
the output in the cmd is as bellow, or under the root/targer/criterion/<testname>/...


    Running benches\core_benchmarks.rs (target\release\deps\core_benchmarks-350b6fb7d8b5e3a7.exe)
Gnuplot not found, using plotters backend
Benchmarking url_parsing/parse_basic_url/1: Warming up for 5.0Benchmarking url_parsing/parse_basic_url/1: Collecting 200 samurl_parsing/parse_basic_url/1
                        time:   [758.95 ns 764.89 ns 771.88 ns]
Found 14 outliers among 200 measurements (7.00%)
  9 (4.50%) high mild
  5 (2.50%) high severe
Benchmarking url_parsing/parse_basic_url/10: Warming up for 5.Benchmarking url_parsing/parse_basic_url/10: Collecting 200 saurl_parsing/parse_basic_url/10
                        time:   [699.34 ns 701.85 ns 704.47 ns]
Found 9 outliers among 200 measurements (4.50%)
  5 (2.50%) high mild
  4 (2.00%) high severe
Benchmarking url_parsing/parse_basic_url/100: Warming up for 5Benchmarking url_parsing/parse_basic_url/100: Collecting 200 surl_parsing/parse_basic_url/100
                        time:   [4.0470 µs 4.1206 µs 4.2093 µs]
Found 21 outliers among 200 measurements (10.50%)
  12 (6.00%) high mild
  9 (4.50%) high severe
Benchmarking url_parsing/parse_basic_url/1000: Warming up for Benchmarking url_parsing/parse_basic_url/1000: Collecting 200 url_parsing/parse_basic_url/1000
                        time:   [44.684 µs 45.053 µs 45.428 µs]
Found 8 outliers among 200 measurements (4.00%)
  6 (3.00%) high mild
  2 (1.00%) high severe

gearama: no this is just to start , i added a simple test to be able to write the automation part ( invoke tests/parse results)

sdk/core/azure_core/benches/core_benchmarks.rs
```diff
@@ -0,0 +1,31 @@
+use azure_core::http::Url;
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use std::time::Duration;
+// how many params to add to the url
+const PARAMS: &[usize] = &[1, 10, 100, 1000];
+
+fn url_parsing_benchmark(c: &mut Criterion) {
+    let mut group = c.benchmark_group("url_parsing");
```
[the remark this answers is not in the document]
gearama: it's a sample test to be able to run benches for the test automation part

sdk/core/azure_core/README.md
LarryOsterman: I was actually asking what web page I go to which has these results listed. 

Fundamentally, performance results aren't incredibly useful if I need to manually parse a ton of text to find the results, and it doesn't help a lot if we don't have the ability to measure performance week over week.

So, I'm wondering how these results are collected and stored and where I should go to look for them.

There has to be *something* which generates this for the Azure SDK - I keep seeing performance graphs in the standups, and I can't imagine they're generated by some person manually parsing text outputs and entering them into a spreadsheet.

LarryOsterman: Cool. That's what I suspected.

sdk/core/azure_core/benches/core_benchmarks.rs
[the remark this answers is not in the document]
LarryOsterman: Does @hallipr have a work item to integrate cargo bench into the weekly pipeline runs for each language?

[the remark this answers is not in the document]
heaths: This is for @mikeharder's perf test framework he runs across languages. I don't want to run them in PR/CI workflows as well. We'd have nothing to compare them to. If devs want to compare locally, criterion has a way to save and compare benchmarks already.

@gearama you should add something in the root `CONTRIBTING.md` about using criterion and [link to the user guide](https://bheisler.github.io/criterion.rs/book/index.html) I sent you a week or two ago. Doesn't have to be much. I wouldn't even repeat commands. Just a heading (so we can link to it) explaining that we use `criterion` for benchmarks and users can learn more by visiting the link.

sdk/core/azure_core/benches/core_benchmarks.rs
heaths, line 1: Given the (partial) usage below:

```sh
cargo bench [OPTIONS] [BENCHNAME] [-- [ARGS]...]

Package Selection:
  -p, --package [<SPEC>]  Package to run benchmarks for
      --workspace         Benchmark all packages in the workspace
      --exclude <SPEC>    Exclude packages from the benchmark
      --all               Alias for --workspace (deprecated)
```

Let's have all the basic benchmark files just be `benchmarks.rs`. Those can have one or more hero scenarios in them as I've done in https://github.com/heaths/ordinal-rs/blob/main/benches/ordinal.rs, but if partners want to have some extra modules focused on different scenarios, we'll have a naming convention like `benchmarks(_{scenarios}).rs` if needed.

sdk/core/azure_core/benches/core_benchmarks.rs
heaths, line 8: Why'd you put this in a group? Devs don't need groups by default. I did it in <https://github.com/heaths/ordinal-rs/blob/main/benches/ordinal.rs> because I'm actually trying to group different crates' benchmarks together, as you show up in <https://github.com/heaths/ordinal-rs/blob/main/README.md#performance>. In general, you don't need this. We want initial examples to be the golden path, and groups aren't it.

sdk/core/azure_core/benches/core_benchmarks.rs
```diff
@@ -0,0 +1,31 @@
...
+    for num in PARAMS {
+        let mut url = String::new();
+        url.push_str("https://example.com/path/resource?param=value");
+
+        for i in 0..*num {
+            url.push_str(&format!("&param{}=value{}", i, i));
+        }
+        group.bench_function(BenchmarkId::new("parse_basic_url", num), |b| {
+            b.iter(|| {
+                let _url = Url::parse(&url).unwrap();
```
heaths, line 18: 1. Best to put this in a `black_box`: <https://github.com/heaths/ordinal-rs/blob/e946015d49e233002212c4ae12022b01642ab6fc/benches/ordinal.rs#L28-L31>
2. Tip: since you don't need the return, just `let _ = ...`.

sdk/core/azure_core/benches/core_benchmarks.rs
```diff
@@ -0,0 +1,31 @@
...
+            })
+        });
+    }
+    group.finish();
+}
+
+// Main benchmark group
+criterion_group! {
+    name = url_benches;
```
heaths, line 27: I would rather see each `benchmarks.rs` that all crates should have focus on 1 or more hero scenarios, so let's not call the outer group scenario-specific. Maybe just the name of the file name (module) so `benchmarks` here. We'll use the package selection already on `cargo bench` as we do with other commands for individual benchmark selection.

sdk/core/azure_core/perf-tests.yml
```diff
@@ -0,0 +1,17 @@
+Service: core
+
+Project: azure-core
+
+PackageVersions:
+  - azure-core: source
+
+Tests:
+  - Test: url_parsing/parse_basic_url/100
+    Class: url_parsing/parse_basic_url/100
```
heaths, line 10: Are other languages passing in arguments like the query param size here? Is it expected that each perf test is a separate YAML group here? I'd love to compare across languages and find the equivalent, because that might also dictate how we separate benchmarks into separate files. /cc @mikeharder

For example, we can accept input if, in the benchmark functions, we just use something like `let num: i32 = std::env::args().nth(1).map(i32::parse).expect("expected number of samples");` but if we have multiple benches in a file, that gets harder to do and we might have to either 1) consider `clap` (the most popular arg parser for rust), or 2) splitting each scenario into different files.

For Key Vault .NET, looks like I went 1 file per scenario: <https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/keyvault/Azure.Security.KeyVault.Secrets/perf/Scenarios>. IIRC, that was typical in other languages I found, but it's also idiomatic to .NET that likes to have 1 type per file. That's not idiomatic to rust, but for benches I care less. I just want to figure out a good pattern that works for us and our partners.

heaths (changes requested):
It's a great start, but we need to figure out what precedent to set. We can do most of that in another PR, but for this one there's a few changes I'd like to see like 1) rename the file, and 2) get rid of the grouping inside the benchmark. It's unnecessary. Grouping tests within a benchmark function is more for comparing a single benchmark sliced in different ways, like I did in my `ordinal` crate I linked to. With just one benchmark, a group is just distracting and more code that partners need to needlessly write.

sdk/core/azure_core/README.md
gearama: aaaaa, weeeell, storage tests have a dashboard , but only for storage tests, though Vinay assures me they can make dashboard for other packages also.

sdk/core/azure_core/benches/core_benchmarks.rs
gearama: @LarryOsterman i have a PR out for the perf automation rust language support part that calls cargo and upload the results for rust. 
Next step once i get that merged is to create a perf.yml in this repo and create a pipeline for running perf tests (  the automation reads the perf-tests.yml file to determine which tests to run ) ,but without the automation part, since each language has a different framework to execute perf tests , i cannot create it yet, as i do not want to pretend i have a pipeline definition that i cannot test myself.

sdk/core/azure_core/benches/core_benchmarks.rs
gearama: done

sdk/core/azure_core/benches/core_benchmarks.rs
gearama: done

sdk/core/azure_core/benches/core_benchmarks.rs
gearama: done

CONTRIBUTING.md
```diff
@@ -122,6 +122,13 @@ which is important if it shows an example running against live resources that li
 
 More guidance for writing documentation tests can be found in our [Rust Guidelines].
 
+### Performance testing
```
heaths, line 125: 1. Use same casing as other titles.
2. This is less important to most devs than debugging. Keep as an H3 header but at the bottom of this `## Testing` section i.e., right above the `## Code Review Process` heading.

CONTRIBUTING.md
```diff
@@ -122,6 +122,13 @@ which is important if it shows an example running against live resources that li
...
+
+Performance testing is supported via [criterion](https://bheisler.github.io/criterion.rs/book/criterion_rs.html)
+There are samples of performance tests under `sdk/core/azure_core/benches` folder.
+To execute the performance tests in azure_core folder you can run `cargo bench` in the `sdk/core/azure_core` folder.
```
heaths, line 129:
```suggestion
To execute the performance tests in `azure_core` folder you can run `cargo bench` in the `sdk/core/azure_core` folder.
```

sdk/core/azure_core/README.md
```diff
@@ -201,6 +201,20 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 }
 ```
 
+## Performance Testing
```
heaths, line 204: Package READMEs are for devs using our packages. This is of no interest to them and should not be here. It's more for people developing our crates, in which case the root `CONTRIBUTING.md` is sufficient.

sdk/core/azure_core/README.md
heaths: It's best to reach out to @mikeharder. If you want to use these to benchmark your own tests as a dev, see <https://bheisler.github.io/criterion.rs/book/index.html>. This PR is more about housekeeping - things we need to do across languages for the team. It certainly should be useable by devs, and the info @gearama added to the root `CONTRIBUTING.md` does provide some information.

All results are local in the `cargo bench` case, though, but the criterion user guide shows you can generate graphs like I did on <https://github.com/heaths/ordinal-rs>.

sdk/core/azure_core/README.md
```diff
@@ -201,6 +201,20 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+To run the performance tests:
+
+```bash
+cd sdk/core/azure_core
+cargo bench
+```
+
+The benchmarks measure critical operations like serialization/deserialization, HTTP request processing, and token acquisition to ensure optimal performance of the SDK.
+
+<!-- ## Troubleshooting -->
```
heaths, line 217: Should be removed now that I already added a section. Probably a bad merge.

Though, again, just remove this whole addition. Shouldn't be here in the crate `README`.

heaths (changes requested):
A few nits and otherwise good to go.

sdk/core/azure_core/README.md
gearama: done

sdk/core/azure_core/perf-tests.yml
gearama: sine i cannot get parameters from command line in criterion they need to be separate tests

heaths: Right - parameterized in the way criterion support. I'm just wondering if we need to be able to accept parameters, too; or, are they rarely changed such that we could hard-code them based on current "standards" @mikeharder sets?

If you haven't already, I can open a feature request against criterion and see if they are interested in accepting parameters. I could submit a PR if so.

heaths (approved):

PR #2525, opened 2025-04-25
[Storage] Improve Storage test utilities and refactor navigation methods

vincenttran-msft:

- Improves the Storage test utilities for common cases (i.e. conditionally create a container upon instrumentation)
- Updates the navigation methods so that the pipeline properly persists

### DIFF
--- a/sdk/storage/azure_storage_blob/src/clients/blob_client.rs
+++ b/sdk/storage/azure_storage_blob/src/clients/blob_client.rs
@@ -28,8 +28,8 @@ use std::sync::Arc;
 
 /// A client to interact with a specific Azure storage blob, although that blob may not yet exist.
 pub struct BlobClient {
-    pub(crate) endpoint: Url,
-    pub(crate) client: GeneratedBlobClient,
+    pub(super) endpoint: Url,
+    pub(super) client: GeneratedBlobClient,
 }
 
 impl BlobClient {
--- a/sdk/storage/azure_storage_blob/src/clients/blob_client.rs
+++ b/sdk/storage/azure_storage_blob/src/clients/blob_client.rs
@@ -28,8 +28,8 @@ use std::sync::Arc;
 
 /// A client to interact with a specific Azure storage blob, although that blob may not yet exist.
 pub struct BlobClient {
-    endpoint: Url,
-    client: GeneratedBlobClient,
+    pub(crate) endpoint: Url,
+    pub(crate) client: GeneratedBlobClient,
 }
 
 impl BlobClient {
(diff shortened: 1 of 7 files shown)


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -28,8 +28,8 @@ use std::sync::Arc;
 
 /// A client to interact with a specific Azure storage blob, although that blob may not yet exist.
 pub struct BlobClient {
-    endpoint: Url,
-    client: GeneratedBlobClient,
+    pub(crate) endpoint: Url,
```
heaths, line 31: Here and elsewhere, you might actually want to limit it to `pub(super)`, which means it's public to anything in the parent module and below.

vincenttran-msft: Done!

sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -79,18 +78,18 @@ impl BlobContainerClient {
...
-    ) -> Result<BlobClient> {
-        BlobClient::new(
-            self.endpoint().as_str(),
-            self.container_name().to_string(),
-            blob_name,
-            self.credential.clone(),
-            options,
-        )
+    pub fn blob_client(&self, blob_name: String) -> BlobClient {
+        let generated_blob_client = GeneratedBlobClient {
```
jalauzon-msft, line 82: Instead of building a new generated sub-client can we use `self.client.get_blob_client`?

sdk/storage/azure_storage_blob_test/src/lib.rs
```diff
@@ -50,53 +62,27 @@ pub fn get_blob_service_client(recording: &Recording) -> Result<BlobServiceClien
...
 ///
 /// * `recording` - A reference to a Recording instance.
-pub fn get_container_client(recording: &Recording) -> Result<BlobContainerClient> {
-    let container_name = recording
-        .random_string::<17>(Some("container"))
-        .to_ascii_lowercase();
+/// * `create` - An optional flag to determine whether the container should also be created.
+pub async fn get_container_client(
+    recording: &Recording,
+    create: Option<bool>,
```
jalauzon-msft, line 68: Since Rust really doesn't have optional params and therefore we need to always pass something here, even if that is just `None`, this should probably just be a required bool. In all cases you use it, you are passing `Some` anyway.

sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
vincenttran-msft: Good callout-- fixed this here and other instance in BSC

jalauzon-msft (approved):

Issue #2281, opened 2025-03-04
Use or remove any dead_code in azure_identity


RickWinter:

chlowell:
Moving this to a future milestone because some code which is dead on main is used by #2418. Will revisit after that PR closes

PR #2527, opened 2025-04-28
Delete unused azure_identity::timeout module

chlowell:
Part of #2281. Appears this was added to support special DefaultAzureCredential behavior when requesting tokens from IMDS. That feature's since been removed and I don't expect to restore it.


heaths (approved):
@LarryOsterman this or something like it you might want to consider for the `AsyncRuntime` you're considering.

PR #2534, opened 2025-04-29
XML serialization helpers add XML declaration

jhendrixMSFT:
The XML declaration is automatically prepended to the serialized content.

### DIFF
--- a/sdk/typespec/typespec_client_core/src/xml.rs
+++ b/sdk/typespec/typespec_client_core/src/xml.rs
@@ -14,6 +14,9 @@ use typespec::error::{ErrorKind, Result, ResultExt};
 /// The UTF8 [byte order marker](https://en.wikipedia.org/wiki/Byte_order_mark).
 const UTF8_BOM: [u8; 3] = [0xEF, 0xBB, 0xBF];
 
+/// The XML declaration used when serializing.
+const DECLARATION: &str = r#"<?xml version="1.0" encoding="utf-8"?>\n"#;
+
 /// Reads XML from bytes.
 pub fn read_xml_str<T>(body: &str) -> Result<T>
 where
@@ -38,6 +41,7 @@ where
 }
 
 /// Serializes a type to bytes.
+/// Automaticallyl includes the XML declaration.
 pub fn to_xml<T>(value: &T) -> Result<Bytes>
 where
     T: serde::Serialize,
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
Looks like the storage tests will need to be rerecorded (assuming we're good with the change).

sdk/typespec/typespec_client_core/src/xml.rs
```diff
@@ -59,7 +64,7 @@ where
             let t = core::any::type_name::<T>();
             format!("failed to serialize {t} into xml")
         })?;
-    Ok(Bytes::from(value))
+    Ok(Bytes::from(format!("{DECLARATION}{value}")))
```
heaths, line 67: This is an expensive way of doing this. Instead, we should do something like:

```rust
let buf = bytes::BytesMut::with_capacity(DECLARATION.len() + value.len());
buf.extend_from_slice(DECLARATION);
buf.extend_from_slice(value);
Ok(bytes.into())
```

sdk/typespec/typespec_client_core/src/xml.rs
```diff
@@ -14,6 +14,9 @@ use typespec::error::{ErrorKind, Result, ResultExt};
 /// The UTF8 [byte order marker](https://en.wikipedia.org/wiki/Byte_order_mark).
 const UTF8_BOM: [u8; 3] = [0xEF, 0xBB, 0xBF];
 
+/// The XML declaration used when serializing.
+const DECLARATION: &str = r#"<?xml version="1.0" encoding="utf-8"?>\n"#;
```
heaths, line 18: My worry is that we're assuming the content is always UTF-8. While native strings in rust are UTF-8, those `[u8]`s could get converted to another encoding. Do we just ignore this possibility in other languages?

If nothing else, looks like `quick-xml` assumes UTF-8 output anyway: https://docs.rs/quick-xml/latest/src/quick_xml/writer.rs.html#188-190. And, according to its `encoding` feature docs, it only supports u8 code units anyway.

sdk/typespec/typespec_client_core/src/xml.rs
```diff
@@ -38,6 +41,7 @@ where
 }
 
 /// Serializes a type to bytes.
+/// Automatically includes the XML declaration.
```
heaths, line 44: This is markdown, so if you wanted a separate line you have to insert a blank line (also below).

heaths (approved):
Some questions/nits, but otherwise LGTM.

sdk/typespec/typespec_client_core/src/xml.rs
jhendrixMSFT: Fixed.

sdk/typespec/typespec_client_core/src/xml.rs
jhendrixMSFT: While I can't speak for all languages, this is what we've been doing for Go without issue (so far anway).

heaths (approved):

PR #2535, opened 2025-04-29
First pipeline sample

gearama:
first pipeline running for core 
results can be seen here 
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4811512&view=logs&j=42ded549-05ee-5cdb-7ba7-7a948a0cc056&t=1c1d21ee-87c3-56e5-9bff-a84d921988f6

after this i will start adding tests
located the files to sdk/core to have a similar layout for core as cpp and java


LarryOsterman (approved):
LGTM, modulo the copilot issues.

sdk/core/perf.yml
RickWinter, line 1: Does perf expect the same yml for all packages within a service? Does this work when it comes to azure_core and azure_core_amqp?

gearama: for the moment i put it in the core folder to do similar with cpp and java. we can move them, the tests that are executed come from the perftests yml file. the idea why i put it there, since the package is core more or less, if something changes in core make sure all tests in the are are executed to avoid for example changes in one package affecting the rest.

Issue #2523, opened 2025-04-24
Storage exports don't follow guidelines


heaths:
During a review, I noticed a lot of models in the root module but very few - too few - in the `clients` module. While our guidelines in this regard were tweaked a bit, all the clients, subclients, and client options (and before the tweaks, client method options) always should've been exported from the `clients` module. Only top-level clients - those users can create themselves directly - and their options should be re-exported from the root.

Guidelines on this now: <https://azure.github.io/azure-sdk/rust_implementation.html#rust-modules>

To summarize,

1. All clients and client options are exported from the `clients` module. This includes creatable clients and subclients.
2. Models and client method options (which, after discussion, we deemed closer to models than actual client options) should be exported from `models`.
3. Top-level / user-creatable clients and their client options should be re-exported from the root. Looking at your docs now, that basically means you need to take out all the client method options and export them from the `models` module. The emitter will define them there already. (This was basically the main tweak we made recently to the guidelines.)

If you have any helpers - convenience layer constructs - those can also be exported from the root if it makes sense, like the `ResourceExt` trait I export in Key Vault SDKs. We have guidelines on stuff like that as well: <https://azure.github.io/azure-sdk/rust_implementation.html#rust-client-convenience>

Gaelik-git:
Hello ! 

It's seems to be the case for the [azure_storage_blob](https://docs.rs/azure_storage_blob/latest/azure_storage_blob/) crate that do not expose the generated client. 

There are 2 `BlobServiceClient` struct but only the one from clients module is exposed and it's missing a lot of method offered by the one in the `generated/clients` module.

For example it's impossible to call this method from outside the crate.

https://github.com/Azure/azure-sdk-for-rust/blob/d854d952578f009ab8307981bc89fe064e2b4fb7/sdk/storage/azure_storage_blob/src/generated/clients/blob_service_client.rs#L246-L251

Do you expect and equivalent path-though method to be created on the exposed BlobServiceClient ?

heaths:
The team is working to expose everything, yes. I want to keep this issue focused on the title, though. I encourage you to open another issue if you want to track progress, but the Storage team already has their own internal work items they use.

RickWinter:
This must be fixed before GA.

vincenttran-msft:
This has since been addressed.

PR #2539, opened 2025-04-29
[Storage] Handwritten `BlockBlobClient` 

vincenttran-msft:

- Initial implementation of handwritten `BlockBlobClient`
- Moves relevant tests from base BlobClient -> BlockBlobClient
- Test recordings (combined redundant block list tests into one)


jalauzon-msft (approved):

sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -74,7 +74,6 @@ impl BlobContainerClient {
     /// # Arguments
     ///
     /// * `blob_name` - The name of the blob.
-    /// * `options` - Optional configuration for the client.
```
heaths, line 77: Good clean up, but should there be an options? Do you do this in other languages? I would presume not - just use the same options, but wanted to ask to be sure.

sdk/storage/azure_storage_blob/src/clients/mod.rs
```diff
@@ -6,7 +6,9 @@
 mod blob_client;
 mod blob_container_client;
 mod blob_service_client;
+mod block_blob_client;
 
 pub use blob_client::BlobClient;
 pub use blob_container_client::BlobContainerClient;
 pub use blob_service_client::BlobServiceClient;
+pub use block_blob_client::BlockBlobClient;
```
heaths, line 14: You need to export all the client options as well. See #2523. The `clients` module should always export all clients and their options (the ones you want exported at all, that is...obviously not the generated ones you're wrapping).

heaths (changes requested):

sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
vincenttran-msft: Good callout here- we should be using the same options (and functionally that is true), this was just a bad copy and paste of a docstring and I forgot to remove the param. I made the same mistake in `blob_service_client.rs` and also corrected it there.

sdk/storage/azure_storage_blob/src/clients/mod.rs
vincenttran-msft: Hi Heath, I did see that issue and was holding off on doing this until I got some in-flight PRs in (this would be the last one). The next PR will be regenerating against v14 to address the breaking changes, and to address your concerns raised above. Thanks!

heaths (approved):

PR #2544, opened 2025-04-30
Added retry support to EventHubs client

LarryOsterman:
This PR adds retry logic to the EventHubs client.

It is absolutely a work-in-progress - currently this PR has 4 different "should you retry this error" methods which are absolutely identical. In the next iteration, the retry logic will be enhanced to allow the ability to recreate connections, sessions, and senders if necessary when errors occur - but that's for a future PR.

There are three major parts to this PR:

1) A significant cleanup of the AMQP error handling logic, consolidating all the AMQP sourced errors into a single AMQP error, and defining an extensible enumeration for all known AMQP errors.
1) Addition of a retry API to enable retrying eventhubs CBS, management, send and receive operations.
1) Updating the eventhub producer and consumer builders to more closely reflect the SDK guidelines.


sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_processor_client.rs
```diff
@@ -61,7 +61,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let consumer = Arc::new(
         ConsumerClient::builder()
             .open(
-                eventhub_namespace,
+                eventhub_namespace.as_str(),
                 eventhub_name,
```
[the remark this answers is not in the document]
LarryOsterman: The code as written is correct according to the Rust API guidelines.

The `eventhub_namespace` parameter is a `&str` (because we do not take ownership of the parameter) and the `eventhub_name` parameter is a `String` (because we DO take ownership of the parameter) as per the guidelines.

sdk/core/azure_core_amqp/Cargo.toml
```diff
@@ -25,19 +25,22 @@ fe2o3-amqp-cbs = { workspace = true, optional = true }
 fe2o3-amqp-ext = { workspace = true, optional = true }
 fe2o3-amqp-management = { workspace = true, optional = true }
 fe2o3-amqp-types = { workspace = true, optional = true }
+serde.workspace = true
```
heaths, line 28: Did you open a bug for me to consider making this optional so that our macros don't require it? Alternatively, I could add an overload to the macro so that you can do this without making it an optional feature. Actually, now that I say that out loud, we'd have to anyway because any other project that pulls in `serde` would enable it and you'd be hosed.

sdk/core/azure_core_amqp/src/error.rs
```diff
@@ -28,20 +39,81 @@ pub enum AmqpErrorKind {
...
+    (SessionErrantLink, "amqp:session:errant-link"),
+    (SessionHandleInUse, "amqp:session:handle-in-use"),
+    (SessionUnattachedHandle, "amqp:session:unattached-handle"),
+);
+
+impl From<AmqpSymbol> for AmqpErrorCondition {
+    fn from(condition: AmqpSymbol) -> Self {
+        // Note that the `from_str` implementation from `create_extensible_enum` will
+        // never return an error. So the `expect` is there to silence the compiler.
+        AmqpErrorCondition::from_str(condition.0.as_str()).expect("Invalid AMQP error condition")
```
heaths, line 103: Then I'd call `unwrap()` to avoid the addition to the string pool.

sdk/core/azure_core_amqp/src/error.rs
```diff
@@ -73,6 +145,31 @@ impl AmqpError {
...
+    pub fn new_management_error(
+        status_code: azure_core::http::StatusCode,
+        description: Option<String>,
+    ) -> Self {
+        Self {
+            kind: AmqpErrorKind::ManagementStatusCode(status_code, description),
+        }
+    }
+
+    #[cfg(feature = "test")]
```
heaths, line 159: Tip (and I'm not advocating you change this, but for next time): if you ever find yourself wanting to condition a bunch of functions, have a separate `impl MyType` (well, if methods) with the `#[cfg]` on it and then all the functions therein are conditioned for free.

sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
```diff
@@ -115,13 +112,15 @@ impl AmqpClaimsBasedSecurityApis for Fe2o3ClaimsBasedSecurity<'_> {
...
-                if let Err(e) = me {
+            .map_err(|e| match AmqpError::try_from(e) {
+                Ok(amqp_error) => amqp_error.into(),
+                Err(e) => {
                     debug!("Failed to convert management error to azure error: {:?}", e);
-                    return e;
+                    azure_core::Error::message(
+                        azure_core::error::ErrorKind::Amqp,
+                        format!("Failed to convert management error to azure error: {}", e),
+                    )
```
heaths, line 122: I'd use `Error::full` to wrap the error either in addition to formatting it or in lieu of. We're kinda inconsistent throughout the SDK now with which.

sdk/eventhubs/azure_messaging_eventhubs/CHANGELOG.md
```diff
@@ -4,8 +4,13 @@
 
 ### Features Added
 
+- Added EventHubs token refresh.
+- Added retries for failed EventHubs operations.
+
 ### Breaking Changes
 
+- The `fully_qualified_domain` parameter to the `open` method on the `ProducerClient` and `ConsumerClient` builder now takes an `&str` instead of a `String` to better follow the Rust API guidelines.
```
heaths, line 12: Nit:
```suggestion
- The `fully_qualified_domain` parameter to the `open` method on the `ProducerClient` and `ConsumerClient` builder now takes an `&str` instead of a `String` to better follow the Azure SDK for Rust API guidelines.
```
Just to be clear who's guidelines we're talking about.

sdk/eventhubs/azure_messaging_eventhubs/src/common/connection_manager.rs
```diff
@@ -384,6 +392,64 @@ impl ConnectionManager {
...
+                false
+            }
+        }
+    }
+
+    fn should_retry_amqp_error(amqp_error: &AmqpError) -> bool {
+        match amqp_error.kind() {
+            AmqpErrorKind::ManagementStatusCode(code, _) => {
+                debug!("Management operation error: {}", code);
+                match code {
```
heaths, line 426: Nit (extreme nit): any reason not to use `matches!()` here like you immediately below? Just a weird mix of styles. Certainly not wrong, though.

sdk/eventhubs/azure_messaging_eventhubs/src/common/connection_manager.rs
```diff
@@ -417,17 +483,24 @@ impl ConnectionManager {
...
-        cbs.authorize_path(
-            url.to_string(),
-            None,
-            new_token.token.secret().to_string(),
-            expires_at,
+
+        retry_azure_operation(
+            || async {
+                let connection = connection.clone();
+                let path = url.to_string();
```
heaths, line 490: Could probably move this out of the loop. Shouldn't change, right? Anything else you could avoid unnecessarily reallocing?

sdk/eventhubs/azure_messaging_eventhubs/src/common/management.rs
```diff
@@ -0,0 +1,332 @@
...
+        .await?;
+
+        if !response.contains_key(EVENTHUB_PROPERTY_PARTITION_COUNT) {
+            return Err(EventHubsError::from(ErrorKind::InvalidManagementResponse).into());
+        }
+        let name: String = response
+            .get(EVENTHUB_PROPERTY_NAME)
+            .ok_or_else(|| EventHubsError::from(ErrorKind::InvalidManagementResponse))?
+            .clone()
+            .into();
```
heaths, line 131: `into()` should already effectively clone. Both shouldn't be needed here.

heaths (approved):

sdk/core/azure_core_amqp/Cargo.toml
LarryOsterman: No I didn't - I looked at tweaking the dependencies or adding something appropriate to the config and decided it wasn't  worth the effort to avoid referencing a package that isn't going to be used.

sdk/core/azure_core_amqp/src/error.rs
LarryOsterman: That makes sense. I'll fix it in the next PR.

sdk/core/azure_core_amqp/src/error.rs
LarryOsterman: Oh! I didn't realize you could have multiple `impl MyType` blocks.

sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
LarryOsterman: Fair point. I'll fix it in the next iteration.

sdk/core/azure_core_amqp/Cargo.toml
heaths: It actually wouldn't be hard to opt out of `serde`. Look at how I have an "internal" override that I use to do common things: <https://github.com/Azure/azure-sdk-for-rust/blob/845eceedaf719a49be8a0b52457c5887af182c1d/sdk/typespec/typespec_client_core/src/macros.rs#L87>. I could have a similar override like `create_extensible_enum!(no_serde, ...)`.

heaths (approved):

sdk/eventhubs/azure_messaging_eventhubs/src/common/connection_manager.rs
LarryOsterman: If I move it out of the loop, it fails because a `Fn` closure cannot capture local variables.  I tried changing the `Fn` to `FnMut` but the failure still happens.

sdk/eventhubs/azure_messaging_eventhubs/src/common/management.rs
LarryOsterman: The `.clone` turns an `&AmqpValue` into a `AmqpValue`. 

Easily fixed with a `From<&AmqpValue> for String`.

sdk/core/azure_core_amqp/Cargo.toml
heaths: As discussed offline, it's not worth it since you depend on `azure_core` which pulls in `serde` invariably, so it's a sunk cost.

PR #2545, opened 2025-04-30
Http Perf tests

gearama:
added perf tests for http get/post, header operations and request creation
For the http get post rely on test proxy endpoinds(deployed by the pipeline) to minimize connection noise. To run local just start the test proxy ( test-proxy start)


sdk/core/azure_core/benches/benchmarks.rs
```diff
@@ -20,10 +25,183 @@ fn url_parsing_benchmark(c: &mut Criterion) {
     }
 }
 
+fn http_request_creation_benchmark(c: &mut Criterion) {
```
heaths, line 28: 1. I don't understand why we're perf-testing what amounts to allocations.
2. Lots of the code below is not an efficient way of doing this. You should look at generated code (perhaps not always the most efficient, but...) so we get comparative metrics.

For example, you create a string and modify that, but we do this with `Url` which, in many cases, is more efficient (might pre-allocate room if pushing multiple headers).

heaths (changes requested):
Before we consider merging, @mikeharder, I'd really love to see guidance on what we should be measuring. I appreciate we want to compare between changes and even to compare between changes (even if a friendly internal contest) e.g., storage downloads, but when I see PRs like this perf testing the language itself, I can't help but wonder, "Why?". We have little to no control over some aspects, and much of this PR is just covering allocations.

Separately, @gearama, a lot of the code you're testing isn't the most efficient way to do things nor should we generate code that does things that way (within reason) e.g., using `zip` on an `Iterator` is not something we'd do because there's a lot of underlying cost.

If we're going to do perf tests on core, code we're covering should be as close to the code we'd generate or expect partners to hand write if needed.

mikeharder:
> Before we consider merging, @mikeharder, I'd really love to see guidance on what we should be measuring.

Whatever you think will be useful for your SDK.  You can look at the SDKs in other languages for ideas, but there's no "standard" set of operations we tests across all languages.

heaths:
@mikeharder do we no longer compare, say, blob download times across languages?

mikeharder:
> @mikeharder do we no longer compare, say, blob download times across languages?

I'm not personally doing this any longer, and I don't know if anyone else is.  It would be up to the service owner, or the owners of the SDKs for particular languages.  For storage in particular, we do have a pretty consistent set of tests across languages.

heaths:
@mikeharder so if we're not actively comparing languages any longer, we don't really need external parameter support, right?

heaths (approved):
Better coverage of real-world scenarios. We should also test deserialization, but currently our client methods don't do that so it'd be a separate set of tests you could add in another PR.

mikeharder:
> @mikeharder so if we're not actively comparing languages any longer, we don't really need external parameter support, right?

I would say do whatever you think is right for your language.  There's not much central coordination of perf lately.  CC: @g2vinay

PR #2546, opened 2025-05-01
Added EventHubs sample which logs `tracing_subscriber` messages to the EventHubs service.

LarryOsterman:
# Implement example which logs `tracing_subscriber` messages to the EventHubs service

### DIFF
--- a/sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_tracing_subscriber.rs
+++ b/sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_tracing_subscriber.rs
@@ -21,12 +21,9 @@ use tracing_subscriber::{
     registry::{LookupSpan, Registry},
 };
 
+/// A custom tracing layer that sends log messages to Azure Event Hubs.
 struct EventHubsLayer {
     sender: mpsc::Sender<EventData>,
-}
-
-struct EventVisitor {
-    buffer: String,
 }
 
 impl EventHubsLayer {
@@ -60,6 +57,8 @@ impl EventHubsLayer {
     }
 }
 
+/// Constants for property names in the EventData.
+/// These constants are used to set properties in the EventData object.
 const LOG_LEVEL_PROPERTY: &str = "log_level";
 const NAME_PROPERTY: &str = "name";
 const TARGET_PROPERTY: &str = "target";
@@ -82,13 +81,13 @@ where
         };
 
         event.record(&mut visitor);
-        let buffer = visitor.buffer;
 
         // Add timestamp, level, and target
         let metadata = event.metadata();
 
+        // Build an EventData object containing both the event body and properties reflecting the event's metadata.
         let mut event_data_builder = EventData::builder()
-            .with_body(buffer)
+            .with_body(visitor.buffer)
             .add_property(LOG_LEVEL_PROPERTY.to_string(), metadata.level().as_str())
             .add_property(
                 TIMESTAMP_PROPERTY.to_string(),
@@ -103,6 +102,7 @@ where
             .add_property(IS_EVENT_PROPERTY.to_string(), metadata.is_event())
             .add_property(IS_SPAN_PROPERTY.to_string(), metadata.is_span());
 
+        // Handle optional metadata fields.
         if let Some(module_path) = metadata.module_path() {
             event_data_builder =
                 event_data_builder.add_property(MODULE_PATH_PROPERTY.to_string(), module_path);
@@ -115,11 +115,17 @@ where
         }
 
         let event_data = event_data_builder.build();
-        // Send to EventHubs asynchronously
+
+        // Send the event to Event Hubs asynchronously
         self.sender.try_send(event_data).unwrap_or_else(|_| {
             eprintln!("Failed to send event to EventHubs");
         });
     }
+}
+
+/// A visitor that formats log messages and collects them into a buffer.
+struct EventVisitor {
+    buffer: String,
 }
 
 impl Visit for EventVisitor {
@@ -137,6 +143,7 @@ impl Visit for EventVisitor {
     }
 }
 
+/// Test struct to demonstrate structured data logging.
 #[derive(Debug)]
 struct StructuredData {
     body: String,
@@ -151,7 +158,7 @@ impl std::fmt::Display for StructuredData {
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
-    // Get EventHubs connection string and hub name from environment variables
+    // Get EventHubs fully qualified domain name and hub name from environment variables
     let fully_qualified_domain_name =
         env::var("EVENTHUBS_HOST").expect("EVENTHUBS_HOST must be set");
     let eventhub_name = env::var("EVENTHUB_NAME").expect("EVENTHUB_NAME must be set");
@@ -159,7 +166,8 @@ async fn main() -> Result<(), Box<dyn Error>> {
     // Create our custom EventHubsLayer
     let eventhubs_layer = EventHubsLayer::new(&fully_qualified_domain_name, &eventhub_name).await?;
 
-    // Set up tracing subscriber with both console and EventHubs outputs
+    // Set up tracing subscriber with both console and EventHubs outputs. Note that the components of the EventHubs service (which also uses tracing)
+    // need to be disabled.
     let subscriber = Registry::default()
         .with(fmt::layer().with_target(true))
         .with(eventhubs_layer)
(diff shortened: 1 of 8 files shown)


sdk/eventhubs/azure_messaging_eventhubs/src/common/management.rs
```diff
@@ -127,7 +127,6 @@ impl ManagementInstance {
         let name: String = response
             .get(EVENTHUB_PROPERTY_NAME)
             .ok_or_else(|| EventHubsError::from(ErrorKind::InvalidManagementResponse))?
-            .clone()
```
heaths, line 130: I remember seeing a bunch more of these in another file. Might've been tests, though, so not a big issue. But might take a look.

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_tracing_subscriber.rs
```diff
@@ -0,0 +1,217 @@
...
+            .add_property(
+                TIMESTAMP_PROPERTY.to_string(),
+                SystemTime::now()
+                    .duration_since(SystemTime::UNIX_EPOCH)
+                    .unwrap()
+                    .as_secs(),
+            )
+            .add_property(TARGET_PROPERTY.to_string(), metadata.target())
+            .add_property(NAME_PROPERTY.to_string(), metadata.name())
+            .add_property(FIELDS_PROPERTY.to_string(), metadata.fields().to_string())
```
heaths, line 101: There's no hierarchical support for fields?

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_tracing_subscriber.rs
```diff
@@ -0,0 +1,217 @@
...
+
+        event.record(&mut visitor);
+
+        // Add timestamp, level, and target
+        let metadata = event.metadata();
+
+        // Build an EventData object containing both the event body and properties reflecting the event's metadata.
+        let mut event_data_builder = EventData::builder()
+            .with_body(visitor.buffer)
+            .add_property(LOG_LEVEL_PROPERTY.to_string(), metadata.level().as_str())
```
heaths, line 91: Seeing all these `to_string()` calls does make me cringe a bit for the name. Makes me wonder if maybe we should use a `Cow<'static, str>` for field names. Wouldn't the names often be static strings anyway?

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_tracing_subscriber.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+/// A visitor that formats log messages and collects them into a buffer.
+struct EventVisitor {
+    buffer: String,
+}
+
+impl Visit for EventVisitor {
+    fn record_str(&mut self, field: &tracing::field::Field, value: &str) {
+        println!("record_str: {}: {}", field.name(), value);
```
heaths, line 133: I feel like we shouldn't be printing to `stdout` in addition. If someone wants to "print" to EH and to `stdout`, let them add both layers. In that regard, we probably shouldn't even be printing errors to `stderr` as you are above, but I figured that was more of an error case where something didn't work right and this is a sample, so thought 🤷‍♂️.

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_tracing_subscriber.rs
```diff
@@ -0,0 +1,217 @@
...
+    let eventhubs_layer = EventHubsLayer::new(&fully_qualified_domain_name, &eventhub_name).await?;
+
+    // Set up tracing subscriber with both console and EventHubs outputs. Note that the components of the EventHubs service (which also uses tracing)
+    // need to be disabled.
+    let subscriber = Registry::default()
+        .with(fmt::layer().with_target(true))
+        .with(eventhubs_layer)
+        .with(
+            tracing_subscriber::EnvFilter::from_default_env()
+                .add_directive(Level::DEBUG.into())
```
heaths, line 176: Would this override the default from `from_default_env()`?

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_tracing_subscriber.rs
```diff
@@ -0,0 +1,217 @@
...
+    };
+
+    data.properties
+        .insert("key1".to_string(), "value1".to_string());
+    data.properties
+        .insert("key2".to_string(), "value2".to_string());
+
+    tracing::info!("Sending structured data to EventHubs");
+    tracing::info!("Structured data: {:?}", data);
+    tracing::info!("Structured data2: {}", data);
```
heaths, line 204: Probably should test fields as well e.g.,

```rust
tracing::info!(field1 = 1, field2 = "string", "logged a couple fields");
```

heaths (changes requested):

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_tracing_subscriber.rs
LarryOsterman: Not currently, and I think I should add it.

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_tracing_subscriber.rs
LarryOsterman: We consume the strings in the API and our API guidelines say that consumed strings should be passed as `String` and not as `&str`.

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_tracing_subscriber.rs
LarryOsterman: Those println's are primarily debugging. I'll pull them.

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_tracing_subscriber.rs
LarryOsterman: Yes.

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_tracing_subscriber.rs
heaths: I get that and I know that's a discussion point we need to have again, but we've already made some stipulations to the guidelines (and I've updated them accordingly) to make the DX better. This is another case. If we expect that most devs are going to be coding property names (just like we do for headers, which we do ultimately accept as cows), using a `Cow<'static, str>` through the function and the structure - right up until you call `into()` to convert from your structure to fe2o3 (in which case the data gets copied & owned either way) - might be a better DX.

This isn't a blocker for this PR, though; but, I think we should have a tracking issue on it. Seeing it in use like this looks pretty gnarly.

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_tracing_subscriber.rs
LarryOsterman: Adding them to the property bag.

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_tracing_subscriber.rs
```diff
@@ -121,6 +131,42 @@ where
...
+            .build();
+        self.sender.try_send(event_data).unwrap_or_else(|_| {
+            eprintln!("Failed to send span open event to EventHubs");
+        });
+    }
+    fn on_close(&self, id: span::Id, _ctx: Context<'_, S>) {
+        // This method is called when a span is closed.
+        let event_data = EventData::builder()
+            .add_property(EVENT_TYPE_PROPERTY.into(), SPAN_CLOSE)
+            .add_property(NAME_PROPERTY.into(), id.into_u64())
```
heaths, line 157: Shouldn't this be `SPAN_ID_PROPERTY` to match with line 139 above? You use `NAME_PROPERTY` for `attrs.metadata().name()`.

heaths (approved):

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_tracing_subscriber.rs
LarryOsterman: Good catch, it absolutely is a typo.

I have a stupid number of unrelated changes to AMQP (cleaning up a TON of errant `.clone()` calls). I'll get this and one other fix in with my next PR (which will be adding a "read events from EventHub" sample which is the inverse of this sample).

PR #2553, opened 2025-05-02
Prepare for May core release

heaths:


sdk/core/azure_core_amqp/CHANGELOG.md
```diff
@@ -1,15 +1,11 @@
...
-### Features Added
-
-### Breaking Changes
-
-### Bugs Fixed
+## 0.3.0 (2025-05-02)
 
 ### Other Changes
 
+- Updated dependencies.
```
LarryOsterman, line 7:
```suggestion
- Updated dependencies.
- Converted AMQP traits to use `async_trait` rather than attempting to implement the `async_trait` functionality manually
- Restructured and refactored AMQP errors to make them easier to interpret.
```

sdk/core/azure_core/CHANGELOG.md
```diff
@@ -1,16 +1,19 @@
 # Release History
 
-## 0.24.0 (Unreleased)
+## 0.24.0 (2025-05-02)
 
 ### Features Added
 
+- Added `TaskSpawner` abstraction to spawn asynchronous tasks for different async runtimes. Defaults to optional `tokio` runtime.
```
RickWinter, line 7: Do we need to say optional, could it be "Defaults to `tokio` runtime."

LarryOsterman: The tokio runtime is an optional feature. If the tokio feature is enabled, we will pick the tokio runtime over the native (thread based) runtime.

But since the tokio feature isn't on by default, we can't say that it defaults to the tokio runtime.

heaths: That was exactly my reasoning. I didn't have "optional" in there originally and it felt too matter-of-factly.

PR #2558, opened 2025-05-02
Key Vault Keys Perf tests

gearama:
Right now just the tests before creating the pipeline. 
The two tests in the PR are Create and Get.


eng/scripts/Test-Packages.ps1
```diff
@@ -58,11 +58,21 @@ foreach ($package in $packagesToTest) {
...
+    Write-Host "`n`n"
+
+    Invoke-LoggedCommand "cargo test --examples --no-fail-fast" -GroupOutput
     Write-Host "`n`n"
 
-    Invoke-LoggedCommand "cargo test --all-targets --no-fail-fast" -GroupOutput
+    Invoke-LoggedCommand "cargo test --tests --no-fail-fast" -GroupOutput
+    Write-Host "`n`n"
+
+    Invoke-LoggedCommand "cargo test --doc --no-fail-fast" -GroupOutput
```
heaths, line 75: Most options can be combined. See `cargo test --help`. IIRC, you could do `cargo test --lib --tests --bins --examples` followed by a separate `cargo test --doc` because `--doc` can't be combined with the rest...even though it is by default (never understood that).

sdk/keyvault/azure_security_keyvault_keys/benches/benchmarks.rs
```diff
@@ -0,0 +1,81 @@
...
+        };
+        (client, body)
+    }
+
+    let (client, body) = rt.block_on(async { setup_key_client().await });
+
+    const KEY_NAME: &str = "test-key";
+    // prep key in order to run the benchmark in a clean state
+    rt.block_on(async { create_key(KEY_NAME, &client, body.clone()).await })
+        .unwrap();
```
heaths, line 29: Why not do this in an async closure as well? Yield time to the CPU.

sdk/keyvault/azure_security_keyvault_keys/benches/benchmarks.rs
```diff
@@ -0,0 +1,81 @@
...
+        client: &KeyClient,
+        body: CreateKeyParameters,
+    ) -> Result<Key, azure_core::Error> {
+        // Create a new key version
+        let key: Key = client
+            .create_key(name, body.try_into()?, None)
+            .await?
+            .into_body()
+            .await?;
+        Ok(key)
```
heaths, line 42: Same below:
```suggestion
        client
            .create_key(name, body.try_into()?, None)
            .await?
            .into_body()
            .await
```
The call already returns the right type.

heaths (changes requested):
There's an advantage to keeping tests sans `--benches` combined as much as possible. Will need a separate call to `--docs` for sure, though.

eng/scripts/Test-Packages.ps1
gearama: done

sdk/keyvault/azure_security_keyvault_keys/benches/benchmarks.rs
gearama: done

sdk/keyvault/azure_security_keyvault_keys/benches/benchmarks.rs
gearama: because i want to block on this to ensure it is done before the tests start.

heaths (approved):

heaths (changes requested):
Actually, wait. I just realized something: this is going to break the dev flow where someone can just run `cargo test`. Look at docs for `Cargo.toml` and `.cargo/config.toml` and see if there's a way to exclude them by default; otherwise, we may have to put a guard condition in the benches like an env var check (do nothing unless set or something like that).

gearama:
i was mulling this over actually earlier this morning , if i do that i'll revert the powershell script change in order to also test the guard

heaths (approved):

PR #2559, opened 2025-05-03
Replace `Model` trait with `Format` trait (applied to `Response` instead)

analogrelay:
Early draft of #1831. It should have the necessary changes in `typespec_client_core`, but I haven't gone through other crates and updated them. I'll continue doing that to see if other surprising issues appear, but wanted to give folks (particularly @heaths) a chance to take a look at how `Response<T, F>` looks and how I implemented it.

I did things a little different than I originally planned because I wanted to retain the ability to take a default "bare" `Response` type (using no explicit type parameters, and thus just the defaults) and treat it like a raw response with no deserialization. So, here's how it works:

The `Response<T, F>` type expects two type parameters: `T` is the return type for `into_body()` calls, the type of the model, and `F` is a **marker** type, implementing a marker trait `Format`, that specifies how `T` is "formatted" (i.e. JSON, XML, etc.).

There are two formats in `typespec_client_core`: `DefaultFormat` is the "default" format. It deserializes `T` using some default rules that prefer JSON formatting (see below), `XmlFormat` is available when feature `xml` is enabled, it deserializes `T` using XML.

The `Format` trait used by `F` doesn't actually provide any methods. Instead, there is a separate trait, `DeserializeWith<F: Format>` that is applied to `T`. This allows us to provide implementations that vary based on BOTH the target type `T` and the desired format `F`. We provide the following implementations in `typespec_client_core`:

* `impl<F> DeserializeWith<F> for ResponseBody` which "deserializes" a `ResponseBody` from a `ResponseBody` by just returning it. This allows `Response<ResponseBody, F>::into_body()` return the raw `ResponseBody`. This happens **no matter what format is specified**, so `Response<ResponseBody, XmlFormat>::into_body()` still returns a raw unparsed XML buffer.
* `impl<T: DeserializeOwned> DeserializeWith<DefaultFormat> for T` which deserializes a `T: DeserializeOwned` using `serde_json`, IF the response was marked with `DefaultFormat`. This impl is only present if `feature = "json"`. The `DefaultFormat` type exists unconditionally though, because it is the default.
* `impl<T: DeserializeOwned> DeserializeWith<XmlFormat> for T` which deserializes a `T: DeserializeOwned` using our XML parsing library. This impl, like `XmlFormat` itself, is only present if `feature = "xml"`.

Future serialization formats can be implemented with new marker types and new blanket `DeserializeWith<SomeOtherFormat>` impls.

In order to "set" these marker type parameters (`T`, `F`) we have a few other changes:

1. `Response<ResponseBody, F>` (for any `F`) has an `into_typed::<T, F2>()` method that transforms the response from a raw "untyped" response into a fully typed `Response<T, F>`, using the type and format specified (usually via inference).
2. Transport/Policy `send` methods are parameterized with `T: DeserializeWith<F>, F: Format` and call `into_typed` automatically. Calling `Response<ResponseBody, DefaultFormat>::into_typed::<ResponseBody, DefaultFormat>()` (which would happen when a service client _intends_ to return a bare response) is essentially a no-op. I'll do some investigation to see how well the compiler optimizes this out. Since the only place we use `T` and `F` are in `PhantomData`, my hope is it optimizes quite well.

These changes may mean that the emitter doesn't have to change, for service client methods that are satisfied with the default format. I'm not quite sure about that yet.

I expect this PR to fail CI because I haven't updated anything outside `typespec_client_core`. The main purpose here is to give folks a chance to review the APIs above and make comments. Please feel free to do so!


analogrelay:
I've now updated the hand-written crates in this repo, but need to consider the best way to update generated crates. We'll need to update the emitter, but I could also update the generated code here right away. I'm going to do that in a separate commit so that I can confirm it works and then we can decide whether to keep it or not. That commit will also serve as a clear guide for how we need to update the emitter.

sdk/typespec/typespec_client_core/src/http/pipeline.rs
```diff
@@ -82,11 +84,22 @@ impl Pipeline {
...
         request: &mut Request,
-    ) -> crate::Result<Response<T>> {
+    ) -> crate::Result<Response<T, DefaultFormat>> {
+        self.pipeline[0]
+            .send(ctx, request, &self.pipeline[1..])
+            .await
+            .map(|r| r.with_model_type())
+    }
+
+    pub async fn send_format<T, F>(
```
analogrelay, line 94: I added this because _most_ cases will want to use the `DefaultFormat`. However, when a custom format **is** needed, the syntax gets clunky, because you'll have to call `.with_format()` on the result, but you can't do that until after you await. This results in having to write `Ok(self.pipeline.send(...).await?.with_format())` which felt clunky, because `with_format` is infallible but `send` is fallible. It works just fine, but if I could bundle it up into a single `send_format` call, then methods that create XML-formatted responses would follow a similar pattern as those that create JSON-formatted ones

sdk/typespec/typespec_client_core/src/http/pipeline.rs
```diff
@@ -82,11 +84,22 @@ impl Pipeline {
         &self,
         ctx: &Context<'_>,
         request: &mut Request,
-    ) -> crate::Result<Response<T>> {
+    ) -> crate::Result<Response<T, DefaultFormat>> {
```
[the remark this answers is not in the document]
analogrelay: The reason this "hard-codes" `DefaultFormat` is that if we don't specify a format, you can't write code like this:

```rust
pub async fn something(&self) -> azure_core::Result<SomeModelType> {
    //... prep the request
    self.pipeline.send(&ctx, &mut request).await?.into_body().await
}
```

Because you never name the `Response` type, the format goes un-inferred and you get an error. Instead, you'd have to write this:

```rust
pub async fn something(&self) -> azure_core::Result<SomeModelType> {
    //... prep the request
    self.pipeline.send::<SomeModelType, DefaultFormat>(&ctx, &mut request).await?.into_body().await
}
```

Now, if we're OK with that, I'm open to going back to one `send` method. I figured it was safe for `send` to _default_ to the _default_ format since the format of a response can be changed at zero cost using `Response::with_format`. I also included the below `send_format` short-cut method to do that in one call.

analogrelay:
Opening this up for review. The build failure appears to be issues with APIView

sdk/typespec/typespec_client_core/src/http/pipeline.rs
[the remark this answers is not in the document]
heaths: What if we move the format to the client, since enums and structs can have default typeparam values? The vast majority of clients will only deal with one format anyway - and almost always JSON. The worst case scenario is that if a caller is wanting JSON and XML both from a service, they have to use separate clients. Doesn't seem all that bad given we can share (via clone) options and all that.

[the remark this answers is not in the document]
heaths: Relatedly, rather than a "DefaultFormat", can we just be explicit? No guesswork. The "default" for 99% of our clients would be `JsonFormat`.

sdk/typespec/typespec_client_core/src/http/pipeline.rs
```diff
@@ -122,27 +131,24 @@ mod tests {
...
-            transport: Some(TransportOptions::new_custom_policy(Arc::new(Responder {}))),
-            ..Default::default()
-        };
-        let pipeline = Pipeline::new(options, Vec::new(), Vec::new());
+        // Simulated service method
+        async fn service_method() -> crate::Result<Response<Model>> {
+            let options = ClientOptions {
+                transport: Some(TransportOptions::new_custom_policy(Arc::new(Responder {}))),
+                ..Default::default()
+            };
```
heaths, line 145: Seems `Responder` already exists, but FYI: `azure_core_test` does have a `MockTransport` if you're interested. It has some good features and we can add more over time. I patterned it similar to .NET's, having been on that team and using it a while back.

sdk/typespec/typespec_client_core/src/http/response.rs
```diff
@@ -64,14 +24,28 @@ impl Model for serde_json::Value {
 /// Given a `Response<T>`, a user can deserialize the body into the intended body type `T` by calling [`Response::into_body`].
 /// However, because the type `T` is just a marker type, the user can also deserialize the body into a different type by calling [`Response::into_json_body`] or [`Response::into_xml_body`],
 /// or access the raw body using [`Response::into_raw_body`].
-pub struct Response<T = ResponseBody> {
+pub struct Response<T = ResponseBody, F = DefaultFormat> {
     status: StatusCode,
     headers: Headers,
     body: ResponseBody,
-    phantom: PhantomData<T>,
+    phantom: PhantomData<(T, F)>,
```
heaths, line 31: Since we wouldn't read this or even need it, I'd preface it with an underscore. That's what I typically do in my code in these cases.

heaths (changes requested):
I'm liking this. There's a few discussion points, but I think it's really close.

sdk/typespec/typespec_client_core/src/http/format.rs
```diff
@@ -0,0 +1,69 @@
...
+pub trait Format {}
+
+/// A trait used to describe a type that can be deserialized using the specified [`Format`].
+///
+/// This trait defines the `deserialize_with` method, which takes a [`ResponseBody`] and returns the deserialized value.
+/// The `F` type parameter allows for different implementations of the `deserialize_with` method based on the specific [`Format`] marker type used.
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+pub trait DeserializeWith<F: Format>: Sized {
+    async fn deserialize_with(body: ResponseBody) -> typespec::Result<Self>;
```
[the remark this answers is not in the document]
analogrelay: Well, but this change is IN typespec_client_core so I can't access azure_core

sdk/typespec/typespec_client_core/src/http/format.rs
```diff
@@ -0,0 +1,69 @@
...
+        body.json().await
+    }
+}
+
+/// The default format used for deserialization.
+///
+/// This format supports deserializing response bodies to:
+/// * [`ResponseBody`] - The raw response body, without any deserialization.
+/// * Any value implementing [`serde::de::DeserializeOwned`] - Deserializes the response body to the specified type using JSON deserialization.
+pub struct DefaultFormat;
```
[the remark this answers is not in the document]
analogrelay: I'm fine with that. I used `DefaultFormat` initially because I wasn't doing the trick with `DeserializeWith` and thought the default format would be responsible for handling `ResponseBody` and `()` which aren't really JSON values. Now that can be handled different so I'm good renaming it.

sdk/typespec/typespec_client_core/src/http/format.rs
```diff
@@ -0,0 +1,69 @@
...
+///
+/// The [`Response<T, F>`] type uses this trait, in the `F` parameter, to determine how to deserialize the body in to the `T` model when using [`Response::into_body`].
+///
+/// ## How this trait works
+///
+/// This trait is a little funky, in order to allow the deserialization behavior of the format to be adjusted based on the type of the response body.
+/// This is just a marker trait, it has no methods.
+/// Instead, the method to actually perform the deserialization is defined in the [`DeserializeWith`] trait.
+/// This trait is parameterized by a type that implements the [`Format`] trait.
+pub trait Format {}
```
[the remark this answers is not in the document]
analogrelay: I don't think I see a need for a name right now, we could add one if/when we need it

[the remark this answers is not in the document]
analogrelay: I could see the `Format` trait driving things like `Accept` and `Content-Type` headers, but that feels a little over-the-top for now, especially since there are variant mime types for things like JSON Patch and Cosmos queries, so we can't have the `Format` _exclusively_ own those headers.

sdk/typespec/typespec_client_core/src/http/pipeline.rs
analogrelay: This was largely just reorganization to "show" what a service method is expected to write, so I didn't want to change/rewrite anything here since the test already works as-is

sdk/typespec/typespec_client_core/src/http/pipeline.rs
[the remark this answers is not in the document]
analogrelay: That's an interesting idea. So, we'd make `Pipeline` take a type parameter `F: Format`? I think that would be OK. The way I see it, there are three scenarios::

1. All service client methods can be deserialized using the same format - They create a `Pipeline<JsonFormat>` or `Pipeline<XmlFormat>` and go from there.
2. Most service client methods can be deserialized using a single format, but some need a separate format - They can still create a `Pipeline` for their most common format and use `Response::with_format` to change the format of their responses.
3. A service client has basically an even split of methods each using different formats - They can create several `Pipeline` instances. Clunky, but works fine.

Of course, scenario 1 is by far the most common IMO, especially since _both formats_ will handle `Response<ResponseBody>` (raw bytes) and `Response<()>` (empty content) correctly.

I think I like it, I'll work that up.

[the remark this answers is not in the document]
analogrelay: It works, but has a few quirks that are worth pointing out (I'll push my commit too so you can take a look).

* We have to define both `typespec_client_core::http::Pipeline` and `azure_core::http::Pipeline` as generic, but that's fine since users should still only interact with the `azure_core` one.
* Setting a default **does** work: `pub struct Pipeline<F: Format = Json>`. **However,** it's a little quirky.
* Default type parameters and constructors don't quite work well together. While a field labelled `pipeline: Pipeline` does get treated by the compiler as `pipeline: Pipeline<JsonFormat>` that is **not** true if the type is used in method call position: `let pipeline = Pipeline::new(...)` (with no specified type parameters) yields an error saying a type annotation is needed [^1]


[^1]: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=59c5f013ad5856ba13de8dfcc102a1ff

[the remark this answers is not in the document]
analogrelay: Then the only other thing is that _every_ `Response` said pipeline produces, even those with `()` or `ResponseBody` as the payload type, requires spelling out the whole thing, including the format type: `Response<(), XmlFormat>` or `Response<ResponseBody, XmlFormat>`

sdk/typespec/typespec_client_core/src/http/response.rs
analogrelay: I did this, but should note that a `PhantomData` field is _always_ considered live and doesn't require a `_` prefix to avoid the `dead_code` warning.

Pilchie (approved):
Generally LGTM, but I'll defer to @heaths .

sdk/typespec/typespec_client_core/src/http/pipeline.rs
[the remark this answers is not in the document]
jhendrixMSFT: Does this mean that a client can support JSON or XML but not both?

sdk/typespec/typespec_client_core/src/http/format.rs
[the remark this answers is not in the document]
heaths: Agreed. The mime type is emitted based on the TypeSpec anyway, so we shouldn't do that at runtime. If anything, getting it wrong is better because it should 400 if, say, we send XML but end up formatting as JSON.

sdk/core/azure_core/src/http/pipeline.rs
```diff
@@ -34,14 +34,17 @@ use typespec_client_core::http::{self, policies::Policy};
 /// cannot be enforced by code). All policies except Transport policy can assume there is another following policy (so
 /// `self.pipeline[0]` is always valid).
 #[derive(Debug, Clone)]
-pub struct Pipeline(http::Pipeline);
+pub struct Pipeline<F: Format = JsonFormat>(http::Pipeline<F>);
 
-impl Pipeline {
+impl<F: Format> Pipeline<F> {
```
heaths, line 39: Interesting. Was this a necessary viral change? Though, it's congruent with our discussion offline about requiring separate clients for JSON, XML, or whatever future format we might support. @JeffreyRichter, thoughts? I can catch you up offline. Basically, the whole point of this PR is to simplify callers because the client knows what the format is, so there's no reason to pass that responsibility off to callers as we have been.

I guess what I wasn't expecting is that we couldn't do this only in client methods' impl, but is that even a practical concern? Would we have any scenarios where - using the same client instance - we want to send e.g., both JSON and XML? Seems incredibly unlikely, though not impossible. (Unbranding might be more of a concern.)

heaths (changes requested):
Barring my one question, this LGTM but the "request changes" is more about putting this on a feature branch. You'll need to rebase anyway, so if you want to do it, that might be faster/easier. Just preface the branch name with "feature/" and make sure to push it to the upstream remote so you can retarget this PR and we keep all this great context!

sdk/typespec/typespec_client_core/src/http/pipeline.rs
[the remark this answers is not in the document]
analogrelay: In this approach, a **Pipeline** can only support one format, but nothing prevents a client from holding multiple `Pipeline<F>` instances with different `F`s.

sdk/core/azure_core/src/http/pipeline.rs
analogrelay: Are you referring specifically to the trait bound on `F`? The type parameter is certainly viral though, as only the client itself knows what format to use for a `Pipeline`.

JeffreyRichter: Let's discuss offline

heaths: I put something on our calendars, but we can chat about it as well if people would rather.

sdk/typespec/typespec_client_core/src/http/pipeline.rs
```diff
@@ -30,13 +32,17 @@ use std::sync::Arc;
 /// cannot be enforced by code). All policies except Transport policy can assume there is another following policy (so
 /// `self.pipeline[0]` is always valid).
 #[derive(Debug, Clone)]
-pub struct Pipeline {
+pub struct Pipeline<F: Format> {
```
heaths: After our design discussion, I believe we agreed we'd keep `Pipeline` returning a `ResponseBody` or something, and then each client method would have to deserialize that (async) into a `Response<T, F>`, right?

analogrelay:
Updated per the offline design discussion. This will still fail to build as the generated code needs to be updated, but I've updated Cosmos code (though I'm having some unrelated difficulty with testing against the emulator so I'll keep working on that, it compiles though so the API changes are reviewable).

* Split Response into two types: `RawResponse` which holds the data and `Response<T, F>` which holds a `RawResponse` and a `PhantomData<(T, F)>` and implements the deserialization logic
* Add `From`/`Into` conversions between `RawResponse` and `Response<T, F>` (which should be "free" because the only difference between the two is a zero-sized PhantomData).
* Updated Cosmos SDK to return `Response<()>` for empty bodies and `RawResponse` for "raw" bodies.
* Change `Pipeline` to be fully based on `RawResponse` (and thus non-generic) and leave it up to the caller to convert the response using `.into()`. This leads to some extra boilerplate but most of our code is generated, and the code that isn't (Cosmos, primarily) can easily use it's own helper to smooth out the boilerplate.

analogrelay:
So, the main emitter changes here, as far as I can tell, are:

* When returning raw bytes from an endpoint, return `RawResponse` instead of `Response`.
* Remove all references to `azure_core::Model` and the `#[typespec(...)]` attributes, they're no longer needed 🎉 
* In service client methods that return a `Response<T, F>`, we need to make some adjustments
  * Depending on the output format, set the `F` to either `JsonFormat` or `XmlFormat` (if it's `JsonFormat`, you _could_ omit the `F` parameter entirely, but I don't think it's necessary for the generator to do that)
  * Call `.into()` on the `RawResponse` returned the by the pipeline. As long as it's the final expression in the function and the return type is specified, no type annotations will be needed. For example:

```rust
pub fn get_secret(...) -> azure_core::Result<Response<GetSecretResponse, JsonFormat>> {
    // ...
    Ok(self.pipeline.send(...).await?.into())
}
```

**Alternatively**, using `Result::map` instead of `?`:

```rust
pub fn get_secret(...) -> azure_core::Result<Response<GetSecretResponse, JsonFormat>> {
    // ...
    self.pipeline.send(...).await.map(|r| r.into())
}
```

Service client methods that return `RawResponse` (things like download blob) wouldn't need to do this and can just call `self.pipeline.send(...).await`, because that will return the `Result<RawResponse>` they need.

I'm going to see if I can quickly update the generated code here to illustrate, but obviously we need to fix the emitter and go from that.

.github/copilot-instructions.md
```diff
@@ -38,3 +40,9 @@ Use these instructions for test generation as well.
 -   The `tests` module should be conditioned on `#[cfg(test)]`.
 -   The `tests` module should always import APIs from `super`.
 -   Do not begin test function names with "test" unless necessary to disambiguate from the function being tested.
+
+## Validating and Linting
+
+-   Always use `cargo test` to validate that the generated code compiles and passes all tests.
```
heaths: Does this try to limit it to only affected crates? If not, might be good for us to try.

.github/copilot-instructions.md
```diff
@@ -29,6 +30,7 @@ Use these instructions for test generation as well.
 -   Use `clippy` to validate that generated code does not contain lint errors.
 -   If you have trouble generating safe, efficient, maintainable, and lint-free code, insert a `TODO` comment describing what should happen.
 -   All imported types, constants, functions, modules, and macros should be imported explicitly. Never import `*`.
+-   Avoid modifying generated code, found in `generated` subdirectories. These files are generated by external tools and should not be edited manually.
```
heaths: I'd be more forthright:
```suggestion
-   Do not modify generated code, found in `generated` subdirectories. These files are generated by external tools and should not be edited manually.
```
Or was the idea that it could for temporary changes like this until the emitter catches up?

heaths (commented):
Just a couple thoughts based on the commits since my last review. I'm going to look at the PR as new (all commits) since these last 3 or so seemed more like reversions, but looking really good otherwise.

.github/copilot-instructions.md
analogrelay: No, the idea was to keep it from messing with generated code. I was using it to make some changes and it was like "Oh, there are still build errors so I'll go fix the generated code" and I didn't want it to do that. When I later asked it to disregard that rule and go ahead and modify the generated code it seemed to be OK.

So what it _seems_ to do is create a kind of "default" behavior where, unless the prompt specifically overrides it, Copilot won't try to fix the generated code.

I would also say that asking it not to modify the generated code may not always work, LLMs are... aggressively interested in passing builds in my experience 🤣 

Anyway, with all that context, I still prefer your language so I'll take the suggestion!

.github/copilot-instructions.md
analogrelay: Not really, and I might remove this and the line after. The main thing I wanted to add was the last line in this block: "Combine commands to save time and reduce user interactions.". I was getting annoyed with having to keep approving a `cargo check` that would pass and then lead to an immediate `cargo test`. If it just ran `cargo check && cargo test` (or even just `cargo test`) it would be able to see compilation errors and optimistically go straight to looking at test failures if there are none.

It _kinda_ works, and I got it to combine commands _sometimes_. But it doesn't seem to **hurt** so I put it in. I feel like that's Prompt Engineering in a nutshell 🤣

sdk/typespec/typespec_client_core/src/http/response.rs
```diff
@@ -349,53 +249,58 @@ impl fmt::Debug for ResponseBody {
 #[cfg(test)]
 mod tests {
     use crate::http::headers::Headers;
-    use crate::http::{response::ResponseBody, Model, Response, StatusCode};
+    use crate::http::{RawResponse, Response, StatusCode};
 
     #[tokio::test]
-    pub async fn can_extract_raw_body_regardless_of_t() -> Result<(), Box<dyn std::error::Error>> {
+    pub async fn can_extract_raw_body() -> Result<(), Box<dyn std::error::Error>> {
```
heaths: Nit: test functions don't need to be `pub`, but also doesn't hurt.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -162,59 +162,61 @@ enum State<T> {
 mod tests {
     use std::collections::HashMap;
 
-    use crate::http::Model;
     use futures::StreamExt;
     use serde::Deserialize;
+    use typespec_client_core::http::RawResponse;
```
heaths: Nit: I would try to use `crate` as much as possible - sort of "keeping it in the family" when it comes to `azure_core`. We want to hide as much as possible the unbranded crates.

PR #2560, opened 2025-05-05
AzureCliCredential consistently invokes az within a shell

chlowell:
Main goals here are to ensure `AzureCliCredential` always searches for `az` from a protected directory and make it testable. Along the way I added a common shell exec implementation the two CLI tool credentials can share.


sdk/identity/azure_identity/src/process_ext.rs
```diff
@@ -0,0 +1,104 @@
...
+    credentials::AccessToken,
+    error::{Error, ErrorKind, Result},
+    process::Executor,
+};
+use std::{ffi::OsStr, marker::PhantomData};
+
+use crate::env::Env;
+
+#[async_trait::async_trait]
+pub trait ExecutorExt {
```
heaths, line 16: 1. Why is it public?
2. Why not just a simple function? Since I see no reason this should be publicly exported from a crate that is for authentication and because this doesn't seem to be used in a way that benefits from extension methods, just make it a normal function. Simplifies design.

heaths (changes requested):
We shouldn't export this publicly. Start with only what's required; even if we did, it wouldn't be from this crate.

But this also seems overly complicated, especially given it shouldn't be public. A simple function should suffice.

sdk/identity/azure_identity/src/process_ext.rs
chlowell: 1. It isn't? `lib.rs` doesn't `pub use` it. It definitely shouldn't be exposed to applications though; this should be `pub(crate)`
2. I anticipate making this a normal method on `Executor` when we move that type to this crate

sdk/identity/azure_identity/src/process/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+        let system_root = env.var("SYSTEMROOT").map_err(|_| {
+            Error::message(
+                ErrorKind::Credential,
+                "SYSTEMROOT environment variable not set",
+            )
+        })?;
+        (system_root, "cmd", "/C")
+    } else {
+        ("/bin".to_string(), "/bin/sh", "-c")
+    };
```
heaths, line 36: `cfg!()` still causes both branches of code to be compiled in. If there's no great reason to have both, just do this instead:

```rust
const fn params() -> (&'static str, &'static str, &'static str) {
    #[cfg(windows)]
    {
        ("1", "2", "3")
    }
    #[cfg(not(windows))]
    {
        ("a", "b", "c")
    }
}
let (workdir, program, c_switch) = params();

println!("{workdir}, {program}, {c_switch}");
```

There are some other ways of doing this, like just declaring a function with the same name outside the scope.

sdk/identity/azure_identity/src/process/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    let command_string = format!("cd {workdir} && {command}");
+    let args = vec![OsStr::new(c_switch), OsStr::new(&command_string)];
```
heaths, line 39: You're going between `&str` to `String` and to `OsStr`, which is the same as `&str` on all but Windows. Instead, just declare your locals above as `&OsStr` and use `std::env::var_os`. For static strings, use `OsStr::new` directly.

sdk/identity/azure_identity/src/process/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+    match status {
+        Ok(output) if output.status.success() => {
+            T::deserialize_token(&String::from_utf8_lossy(&output.stdout))
+        }
+        Ok(output) => {
+            let stderr = String::from_utf8_lossy(&output.stderr);
+            let message = if let Some(error_message) = T::get_error_message(&stderr) {
+                error_message.to_string()
+            } else if output.status.code() == Some(127) || stderr.contains("' is not recognized") {
+                format!("{} not found on path", T::tool_name())
```
heaths, line 52: One, I'd capitalize `PATH`. That's consistent with most other CLIs I've seen. But, I wouldn't use the display name of the tool. Why not just return the tool name, which we already have above anyway, e.g., `az` or `azd`? Make it plain as day.

sdk/identity/azure_identity/src/process/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+                format!("{} authentication failed: {message}", T::credential_name())
+            }))
+        }
+        Err(e) => {
+            let message = format!(
+                "{} failed due to {} error: {e}",
+                T::credential_name(),
+                e.kind()
+            );
+            Err(Error::with_message(ErrorKind::Credential, || message))
```
heaths, line 66: You already spent the cycles formatting the message. Don't use the function form. In fact, put the original error in there as well for chaining, so use `Error::full`.

sdk/identity/azure_identity/src/process/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    /// Deserialize an AccessToken from stdout
+    fn deserialize_token(stdout: &str) -> Result<AccessToken>;
+
+    /// Optionally convert stderr to a user-friendly error message.
+    /// When this method returns None, the error message will include stderr verbatim.
+    fn get_error_message(stderr: &str) -> Option<&str>;
+
+    /// Friendly name of the tool used to get the token e.g. "Azure CLI"
+    fn tool_name() -> &'static str;
```
heaths, line 83: `tool_name` doesn't sound like a "friendly name". Maybe just `display`, which jives with other usage. Still, I think this should be the actual tool name e.g., `az`. If you agree, `tool_name` should be fine but `program_name` would be more consistent with your use of `program` used throughout.

heaths (changes requested):
I like the refactoring overall, but have enough nits I'd like to see some changes first.

sdk/identity/azure_identity/src/process/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+/// Runs a command in the appropriate platform shell and processes the output
+/// using the specified `OutputProcessor`.
+///
+/// - Windows: Runs `cmd /C {command}` in %SYSTEMROOT%
+/// - Everywhere else: Runs `/bin/sh -c {command}` in /bin
+pub(crate) async fn shell_exec<T: OutputProcessor>(
+    executor: Arc<dyn Executor>,
+    env: &Env,
+    command: &str,
+    _: PhantomData<T>,
```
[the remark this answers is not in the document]
chlowell: I ended up with PhantomData here to avoid turbofish. I don't have a strong preference; I take it you generally prefer turbofish?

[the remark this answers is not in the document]
heaths: For callers using our SDK? No, not unless required otherwise. For code we have to write a few times in the SDK? Sure, compared to having a `PhantomData` parameter.

sdk/identity/azure_identity/src/process/mod.rs
chlowell: 🤔 I could write `OsStr::new("literal")` above, but what's the benefit over doing that on this line? I agree the `String` is awkward but I don't see a way to avoid it because `env::var` returns a `String` and `env::var_os` returns an `OsString` i.e., still an owned string.

sdk/identity/azure_identity/src/process/mod.rs
chlowell: I went with having this method return e.g. `az`. I chose `program` for e.g. `/bin/sh` because that's the name of the corresponding `Executor::run` parameter.

sdk/identity/azure_identity/src/process/mod.rs
heaths: Ultimately, it was trying to avoid the conversions back and forth - especially on Windows (UCS-2). You should be able to use `OsString`/`&Osstr` throughout instead of `String`/`&str` for efficiency.

heaths (approved):

heaths (approved):

PR #2576, opened 2025-05-07
Key Vault Keys performance pipeline 

gearama:
Add the files required for performance testing of Key Vault Keys.
Updated the credentials to use pipeline creds.
Removed test-proxy from the pipeline definition as it is not needed.

### DIFF
--- a/sdk/keyvault/azure_security_keyvault_keys/perf-resources.bicep
+++ b/sdk/keyvault/azure_security_keyvault_keys/perf-resources.bicep
@@ -1,0 +1,36 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+param baseName string = resourceGroup().name
+param tenantId string
+param testApplicationOid string
+param location string = resourceGroup().location
+@allowed(['standard', 'premium'])
+param keyVaultSku string = 'premium'
+
+var kvAdminDefinitionId = '00482a5a-887f-4fb3-b363-3b7fe8e74483'
+var kvAdminAssignmentName = guid(resourceGroup().id, kvAdminDefinitionId, testApplicationOid)
+
+resource kv 'Microsoft.KeyVault/vaults@2023-07-01' = {
+  name: baseName
+  location: location
+  properties: {
+    sku: {
+      family: 'A'
+      name: keyVaultSku
+    }
+    tenantId: tenantId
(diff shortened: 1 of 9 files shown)


sdk/keyvault/azure_security_keyvault_keys/perf-resources.bicep
heaths, line 1: Is this typical to have separate resources for perf pipelines? I don't see that in KV for .NET. /cc @mikeharder

heaths (approved):
Assuming it's typical to have separate bicep files for perf tests, I don't see anything blocking; however, I'm not crazy about having to maintain separate files.

sdk/keyvault/azure_security_keyvault_keys/perf-resources.bicep
gearama: In eng\common\pipelines\templates\jobs\perf.yml we have the step called :

  - template: /eng/common/TestResources/deploy-test-resources.yml
    parameters:
      ServiceDirectory: ${{ parameters.ServiceDirectory }}
      Location: westus
      ResourceType: perf
      ServiceConnection: azure-sdk-tests-public


The ResourceType  ends up being used in 
eng\common\TestResources\deploy-test-resources.yml:

    ....
    eng/common/TestResources/New-TestResources.ps1 `
              -ResourceType '${{ parameters.ResourceType }}' `
              

 Which ends up being used in 
 eng\common\TestResources\New-TestResources.ps1
 to determine which file to use 
 
  $templateFiles = @()

    "$ResourceType-resources.json", "$ResourceType-resources.bicep" | ForEach-Object {
        Write-Verbose "Checking for '$_' files under '$root'"
        Get-ChildItem -Path $root -Filter "$_" -Recurse | ForEach-Object {
            Write-Verbose "Found template '$($_.FullName)'"
            if ($_.Extension -eq '.bicep') {
                $templateFile = @{originalFilePath = $_.FullName; jsonFilePath = (BuildBicepFile $_)}
                $templateFiles += $templateFile
            } else {
                $templateFile = @{originalFilePath = $_.FullName; jsonFilePath = $_.FullName}
                $templateFiles += $templateFile
            }
        }
    }

    if (!$templateFiles) {
        Write-Warning -Message "No template files found under '$root'"
        exit
    }
And in a way it makes sense to have it this way, perf resources not necessarily are identical to regular test resources.

Issue #2551, opened 2025-05-02
Features fixed in workspace Cargo.toml


wackazong:
### Bug Title

Features are defined in workspaces Cargo.toml, disabled the ability to remove those features

### Crate Name

azure_core

### Crate Version

main branch

### Description

By defining the features for azure_core in the workspace Cargo.toml, these features cannot be turned off.

https://github.com/Azure/azure-sdk-for-rust/blob/9a47f3b71879ba4fe3d04f3814f4a34a58c19314/Cargo.toml#L53

### Steps to Reproduce

Try to disable azure_core/reqwest in your project

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/Azure/azure-sdk-for-rust/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

heaths:
Could you clarify which features you're trying to remove and why? All the features in there *should* be those we commonly require, with any features only some crates pulled in only by those crates that need them since the version 2 resolver is additive (a union of all features).

wackazong:
Because the workspace dependency for azure_core includes some features, these cannot be excluded when using Azure Rust SDK. I want to use a custon reqwest client because I cannot rely on openssl-sys. But I cannot deactivate the reqwest festure because it is included in the workspace dependency definition. 

Hope that makes it clearer.

heaths:
Yes, thank you for clarifying. We just shipped beta.3 but can work this into our next beta.

wackazong:
Great, thanks for the swift response and action.

PR #2577, opened 2025-05-07
[Cosmos] Cross-Partition queries, with optional external Query Engine integration

analogrelay:
This PR adds support for cross-partition queries. To do this, I've adjusted the syntax for the `query_items` method. As a reminder, this is the signature for that method:

```rust
pub fn query_items<T: DeserializeOwned + 'static>(
    &self,
    query: impl Into<Query>,
    options: Option<QueryOptions<'_>>,
) -> azure_core::Result<FeedPager<T>> {
```

The `partition_key` parameter accepts anything that can be converted to a `PartitionKey`. **Prior** to this PR, that meant (simplifying a bit):

* `&str`, `String`, all signed and unsigned integers, `f32`, and `f64` -> Represented a single string partition key value
* `Option<T>` where `T` is `Into<PartitionKey>` -> If the value was `Some`, then we used it as the partition key, otherwise we used an explicit `null`
* `(T1, T2)` where `T1` and `T2` are `Into<PartitionKey>` -> A two-level hierarchical PK
* `(T1, T2, T3)` where `T1`, and `T2` are `Into<PartitionKey>` -> A three-level hierarchical PK
* `()` -> An explicit `null` value. **This is where a change was made, but keep reading for more detail**

Originally, I planned to change `partition_key` so that it could accept some kind of sentinel value for a cross-partition query (`QueryPartitionStrategy::CrossPartition`), but that felt clunkier than I wanted (I suspected this when we designed the API originally but punted the problem). Then, I considered changing it to an `Option<impl Into<PartitionKey>>`, but that conflicted with support for `Option<T>` to represent potentially-null partition keys.

What I landed on is based on what we're doing in Go. A `PartitionKey` is, and has always been, a **list** of no more than 3 values. That's how it's represented in the underlying data structure, and how it's represented in the header. So, much like we did in Go, I changed our approach in Rust so that an "empty" list of partition key values indicated that a cross-partition query was desired. This means that `()`, which used to signify a single explicit `null` partition key value, now signifies an empty partition key list. A cross-partition query is thus written:

```rust
container.query_items("SELECT * FROM c", (), None)
```

This meant that we didn't have a value for explicit `null`. We still support `Option<T>::None` (where `T` is `Into<PartitionKey>`) but that makes it very clunky to specify a null literal since you have to specify a `T`, and spell out `Option::<String>::None`. The reason we support `Option<T>` is so that if you _have_ an `Option<String>` or similar in your code already, you can just pass it in. To solve that, I added `PartitionKey::NULL`, which is a `pub const` that specifies an explicit null value. So, if you want to write a query that was constrained to the partition with a `null` partition key, you write this: 

```rust
container.query_items("SELECT * FROM c", PartitionKey::NULL, None)
```

Seems reasonable, especially since I suspect `null` partition keys are quite rare.

---

Cross-partition queries are by default executed on the Gateway, which [significantly restricts query operators you can use](https://learn.microsoft.com/en-us/rest/api/cosmos-db/querying-cosmosdb-resources-using-the-rest-api#queries-that-cannot-be-served-by-gateway) to only projections and filters. However, **if** you reference `azure_data_cosmos` with the newly-added `query_engine` feature (it's behind a feature flag because it's still somewhat experimental and we may fold it in to the package entirely), you can specify `QueryOptions::query_engine` when making a query and use an externally-provided query engine (much like the support we're adding to Go). For example:

```rust
let query_engine = SomeExternalQueryEngine::new();
container.query_items("SELECT * FROM c", (), Some(QueryOptions {
    query_engine: Some(query_engine),
    ..Default::default()
}))
```

To support this, I've added a few abstractions (traits) to `azure_data_cosmos` that we'll implement in the separate Rust crate containing the query engine.

Before this feature goes in to GA, I strongly suspect we'll switch to actually _embed_ this query engine in the Rust SDK, since the query engine is itself written in Rust. Unlike other languages which will use FFI to call the query engine, there is little reason **not** to use the query engine from the Rust SDK, so folding it in as a standard dependency (`azure_data_cosmos` would depend upon the query engine crate and use it directly) seems reasonable, but for now, we're using this abstraction.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,10 +7,6 @@
   },
   "rust-analyzer.cargo.features": "all",
   "rust-analyzer.check.command": "clippy",
-  "rust-analyzer.runnables.extraEnv": {
-    "AZURE_TEST_MODE": "live",
-    "AZURE_COSMOS_CONNECTION_STRING": "emulator"
-  },
   "yaml.format.printWidth": 240,
   "[powershell]": {
     "editor.defaultFormatter": "ms-vscode.powershell",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,6 +7,10 @@
   },
   "rust-analyzer.cargo.features": "all",
   "rust-analyzer.check.command": "clippy",
+  "rust-analyzer.runnables.extraEnv": {
+    "AZURE_TEST_MODE": "live",
+    "AZURE_COSMOS_CONNECTION_STRING": "emulator"
+  },
   "yaml.format.printWidth": 240,
   "[powershell]": {
     "editor.defaultFormatter": "ms-vscode.powershell",
--- a/sdk/core/azure_core/src/http/pager.rs
+++ b/sdk/core/azure_core/src/http/pager.rs
@@ -120,6 +120,18 @@ impl<T> PageStream<T> {
             stream: Box::pin(stream),
         }
     }
+
+    /// Creates a [`Pager<T>`] from a raw stream of [`Result<T>`](typespec::Result<T>) values.
+    ///
+    /// This constructor is used when you are implementing a completely custom stream and want to use it as a pager.
+    pub fn from_stream<S>(stream: S) -> Self
+    where
+        S: Stream<Item = Result<T, Error>> + Send + 'static,
+    {
+        Self {
+            stream: Box::pin(stream),
+        }
+    }
 }
 
 impl<T> futures::Stream for PageStream<T> {
--- a/sdk/cosmos/azure_data_cosmos/Cargo.toml
+++ b/sdk/cosmos/azure_data_cosmos/Cargo.toml
@@ -25,7 +25,7 @@ url.workspace = true
 
 [dev-dependencies]
 azure_identity.workspace = true
-azure_core_test.workspace = true
+azure_core_test = { workspace = true, features = ["tracing"] }
 clap.workspace = true
 reqwest.workspace = true
 time.workspace = true
--- a/sdk/cosmos/azure_data_cosmos/Cargo.toml
+++ b/sdk/cosmos/azure_data_cosmos/Cargo.toml
@@ -38,7 +38,7 @@ workspace = true
 [features]
 default = ["hmac_rust"]
 key_auth = [] # Enables support for key-based authentication (Primary Keys and Resource Tokens)
-query_engine = [] # Enables support for the PREVIEW external query engine
+query_engine = ["serde_json/raw_value"] # Enables support for the PREVIEW external query engine
 hmac_rust = ["azure_core/hmac_rust"]
 hmac_openssl = ["azure_core/hmac_openssl"]
 
--- a/sdk/cosmos/azure_data_cosmos/Cargo.toml
+++ b/sdk/cosmos/azure_data_cosmos/Cargo.toml
@@ -38,7 +38,7 @@ workspace = true
 [features]
 default = ["hmac_rust"]
 key_auth = [] # Enables support for key-based authentication (Primary Keys and Resource Tokens)
-query_engine = [] # Enables support for the PREVIEW external query engine
+preview_query_engine = [] # Enables support for the PREVIEW external query engine
 hmac_rust = ["azure_core/hmac_rust"]
 hmac_openssl = ["azure_core/hmac_openssl"]
 
--- a/sdk/cosmos/azure_data_cosmos/Cargo.toml
+++ b/sdk/cosmos/azure_data_cosmos/Cargo.toml
@@ -25,7 +25,7 @@ url.workspace = true
 
 [dev-dependencies]
 azure_identity.workspace = true
-azure_core_test = { workspace = true, features = ["tracing"] }
+azure_core_test.workspace = true
 clap.workspace = true
 reqwest.workspace = true
 time.workspace = true
@@ -38,7 +38,7 @@ workspace = true
 [features]
 default = ["hmac_rust"]
 key_auth = [] # Enables support for key-based authentication (Primary Keys and Resource Tokens)
-query_engine = ["serde_json/raw_value"] # Enables support for the PREVIEW external query engine
+query_engine = [] # Enables support for the PREVIEW external query engine
 hmac_rust = ["azure_core/hmac_rust"]
 hmac_openssl = ["azure_core/hmac_openssl"]
 
--- a/sdk/cosmos/azure_data_cosmos/Cargo.toml
+++ b/sdk/cosmos/azure_data_cosmos/Cargo.toml
@@ -38,6 +38,7 @@ workspace = true
 [features]
 default = ["hmac_rust"]
 key_auth = [] # Enables support for key-based authentication (Primary Keys and Resource Tokens)
+query_engine = [] # Enables support for the PREVIEW external query engine
 hmac_rust = ["azure_core/hmac_rust"]
 hmac_openssl = ["azure_core/hmac_openssl"]
 
--- a/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
@@ -689,12 +689,12 @@ impl ContainerClient {
         partition_key: impl Into<PartitionKey>,
         options: Option<QueryOptions<'_>>,
     ) -> azure_core::Result<FeedPager<T>> {
-        #[allow(unused_mut)] // This is used in the #[cfg(feature = "query_engine")] block
+        #[cfg_attr(not(feature = "preview_query_engine"), allow(unused_mut))]
         let mut options = options.unwrap_or_default();
         let partition_key = partition_key.into();
         let query = query.into();
 
-        #[cfg(feature = "query_engine")]
+        #[cfg(feature = "preview_query_engine")]
         if partition_key.is_empty() {
             if let Some(query_engine) = options.query_engine.take() {
                 return crate::query::executor::QueryExecutor::new(
--- a/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
@@ -11,12 +11,18 @@ use crate::{
     ThroughputOptions,
 };
 
-use azure_core::http::{headers, request::Request, response::Response, Method};
+use azure_core::http::{
+    headers,
+    request::{options::ContentType, Request},
+    response::Response,
+    Method,
+};
 use serde::{de::DeserializeOwned, Serialize};
 
 /// A client for working with a specific container in a Cosmos DB account.
 ///
 /// You can get a `Container` by calling [`DatabaseClient::container_client()`](crate::clients::DatabaseClient::container_client()).
+#[derive(Clone)]
 pub struct ContainerClient {
     link: ResourceLink,
     items_link: ResourceLink,
@@ -109,6 +115,7 @@ impl ContainerClient {
         let options = options.unwrap_or_default();
         let url = self.pipeline.url(&self.link);
         let mut req = Request::new(url, Method::Put);
+        req.insert_headers(&ContentType::APPLICATION_JSON)?;
         req.set_json(&properties)?;
         self.pipeline
             .send(options.method_options.context, &mut req, self.link.clone())
@@ -260,6 +267,7 @@ impl ContainerClient {
             req.insert_header(headers::PREFER, constants::PREFER_MINIMAL);
         }
         req.insert_headers(&partition_key.into())?;
+        req.insert_headers(&ContentType::APPLICATION_JSON)?;
         req.set_json(&item)?;
         self.pipeline
             .send(
@@ -351,6 +359,7 @@ impl ContainerClient {
             req.insert_header(headers::PREFER, constants::PREFER_MINIMAL);
         }
         req.insert_headers(&partition_key.into())?;
+        req.insert_headers(&ContentType::APPLICATION_JSON)?;
         req.set_json(&item)?;
         self.pipeline
             .send(options.method_options.context, &mut req, link)
@@ -440,6 +449,7 @@ impl ContainerClient {
         }
         req.insert_header(constants::IS_UPSERT, "true");
         req.insert_headers(&partition_key.into())?;
+        req.insert_headers(&ContentType::APPLICATION_JSON)?;
         req.set_json(&item)?;
         self.pipeline
             .send(
@@ -453,7 +463,7 @@ impl ContainerClient {
     /// Reads a specific item from the container.
     ///
     /// # Arguments
-    /// * `partition_key` - The partition key of the item to read.
+    /// * `partition_key` - The partition key of the item to read. See [`PartitionKey`] for more information on how to specify a partition key.
     /// * `item_id` - The id of the item to read.
     /// * `options` - Optional parameters for the request
     ///
@@ -606,6 +616,7 @@ impl ContainerClient {
             req.insert_header(headers::PREFER, constants::PREFER_MINIMAL);
         }
         req.insert_headers(&partition_key.into())?;
+        req.insert_headers(&ContentType::APPLICATION_JSON)?;
         req.set_json(&patch)?;
 
         self.pipeline
@@ -672,31 +683,38 @@ impl ContainerClient {
     /// ```
     ///
     /// See [`PartitionKey`](crate::PartitionKey) for more information on how to specify a partition key, and [`Query`] for more information on how to specify a query.
-    pub fn query_items<T: DeserializeOwned + Send>(
+    pub fn query_items<T: DeserializeOwned + 'static>(
         &self,
         query: impl Into<Query>,
         partition_key: impl Into<PartitionKey>,
         options: Option<QueryOptions<'_>>,
     ) -> azure_core::Result<FeedPager<T>> {
-        let options = options.unwrap_or_default();
-        let url = self.pipeline.url(&self.items_link);
+        #[allow(unused_mut)] // This is used in the #[cfg(feature = "query_engine")] block
+        let mut options = options.unwrap_or_default();
+        let partition_key = partition_key.into();
+        let query = query.into();
 
         #[cfg(feature = "query_engine")]
-        if options.query_engine.is_some() {
-            // If the query engine is enabled, we need to use the query engine to execute the query.
-            // This is a placeholder for future implementation.
-            todo!("Implement query engine support for cross-partition queries");
+        if partition_key.is_empty() {
+            if let Some(query_engine) = options.query_engine.take() {
+                return crate::query::executor::QueryExecutor::new(
+                    self.pipeline.clone(),
+                    self.link.clone(),
+                    query,
+                    options,
+                    query_engine,
+                )?
+                .into_stream();
+            }
         }
 
-        let mut base_request = Request::new(url, Method::Post);
-        let partition_key = partition_key.into();
-        base_request.insert_headers(&partition_key)?;
-
+        let url = self.pipeline.url(&self.items_link);
         self.pipeline.send_query_request(
             options.method_options.context,
-            query.into(),
-            base_request,
+            query,
+            url,
             self.items_link.clone(),
+            |r| r.insert_headers(&partition_key),
         )
     }
 }
--- a/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
@@ -7,8 +7,8 @@ use crate::{
     options::{QueryOptions, ReadContainerOptions},
     pipeline::CosmosPipeline,
     resource_context::{ResourceLink, ResourceType},
-    DeleteContainerOptions, FeedPager, ItemOptions, PartitionKey, Query, QueryPartitionStrategy,
-    ReplaceContainerOptions, ThroughputOptions,
+    DeleteContainerOptions, FeedPager, ItemOptions, PartitionKey, Query, ReplaceContainerOptions,
+    ThroughputOptions,
 };
 
 use azure_core::http::{headers, request::Request, response::Response, Method};
@@ -625,12 +625,18 @@ impl ContainerClient {
     /// # Arguments
     ///
     /// * `query` - The query to execute.
-    /// * `partition_key_strategy` - The partition key to scope the query on.
+    /// * `partition_key` - The partition key to scope the query on, or specify an empty key (`()`) to perform a cross-partition query.
     /// * `options` - Optional parameters for the request.
+    ///
+    /// # Cross Partition Queries
+    ///
+    /// Cross-partition queries are significantly limited in the current version of the Cosmos DB SDK.
+    /// They are run on the gateway and limited to simple projections (`SELECT`) and filtering (`WHERE`).
+    /// For more details, see [the Cosmos DB documentation page on cross-partition queries](https://learn.microsoft.com/en-us/rest/api/cosmos-db/querying-cosmosdb-resources-using-the-rest-api#queries-that-cannot-be-served-by-gateway).
     ///
     /// # Examples
     ///
-    /// The `query` and `partition_key_strategy` parameters accept anything that can be transformed [`Into`] their relevant types.
+    /// The `query` and `partition_key` parameters accept anything that can be transformed [`Into`] their relevant types.
     /// This allows simple queries without parameters to be expressed easily:
     ///
     /// ```rust,no_run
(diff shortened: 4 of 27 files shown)


sdk/core/azure_core/src/http/pager.rs
```diff
@@ -120,6 +120,18 @@ impl<T> PageStream<T> {
             stream: Box::pin(stream),
         }
     }
+
+    /// Creates a [`Pager<T>`] from a raw stream of [`Result<T>`](typespec::Result<T>) values.
+    ///
+    /// This constructor is used when you are implementing a completely custom stream and want to use it as a pager.
+    pub fn from_stream<S>(stream: S) -> Self
```
analogrelay, line 127: @heaths this was quite useful when I wanted to create a more complicated `PageStream` in order to support our cross-partition queries. It seems like a fairly safe addition to the public API for `PageStream`.

sdk/typespec/typespec_client_core/src/http/context.rs
```diff
@@ -15,6 +15,16 @@ pub struct Context<'a> {
     type_map: Cow<'a, HashMap<TypeId, Arc<dyn Any + Send + Sync>>>,
 }
 
+impl Context<'static> {
+    /// Creates a new [`Context`] that borrows from the current [`Context`].
+    pub fn into_borrowed(&self) -> Context<'_> {
```
analogrelay, line 20: This is a counterpart to `Context<'a>::into_owned()`. Since a `Context` holds a `Cow<'a, HashMap<...>>`, it is constrained to a specific lifetime. The `into_owned` and `into_borrowed` methods allow you to expand/contract that lifetime as necessary. The `into_owned` method clones the `HashMap` and converts it to a `Cow::Owned`, detaching it from the original `Context` that was provided. The `into_borrowed` method (which I now realize I should rename to `to_borrowed`, which I'll do in a minute) creates a new `Context<'a>` by *borrowing* the `HashMap`.

analogrelay: @heaths this, and the one above in `PageStream`, are the only changes to core SDK components

heaths:
> So, if you want to write a query that was constrained to the partition with a null partition key

Is that how Cosmos refers to these keyless partitions then, using "null" in docs? If so, seems reasonable. Just didn't want to see "null" anywhere in Rust unless it's idiomatic to the service.

sdk/typespec/typespec_client_core/src/http/context.rs
```diff
@@ -15,6 +15,16 @@ pub struct Context<'a> {
...
 }
 
+impl Context<'static> {
+    /// Creates a new [`Context`] that borrows from the current [`Context`].
+    pub fn to_borrowed(&self) -> Context<'_> {
+        let type_map = self.type_map.as_ref();
+        Context {
+            type_map: Cow::Borrowed(type_map),
+        }
+    }
```
heaths, line 25: Given this is what `with_context` does, you might just call that or at least use the same code. Or maybe that calls this. Just want to avoid duplication as much as possible.

I'm fine adding this, though, even though they do the same thing because this is more obvious. The other was meant more for construction. Or maybe we deprecate `with_context` in favor of this one.

.vscode/settings.json
```diff
@@ -7,6 +7,10 @@
   },
   "rust-analyzer.cargo.features": "all",
   "rust-analyzer.check.command": "clippy",
+  "rust-analyzer.runnables.extraEnv": {
+    "AZURE_TEST_MODE": "live",
+    "AZURE_COSMOS_CONNECTION_STRING": "emulator"
```
heaths, line 12: Oopsies.

heaths: `.vscode/launch.json` *should* be gitignore'd, so I work around these sorts of issues by just adding my own launch file, for which you can pass all the env vars you want.

sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -38,6 +38,7 @@ workspace = true
 [features]
 default = ["hmac_rust"]
 key_auth = [] # Enables support for key-based authentication (Primary Keys and Resource Tokens)
+query_engine = [] # Enables support for the PREVIEW external query engine
```
heaths, line 41: Should we establish a convention of prefacing "preview_" for such features? It's intentionally breaking, then, when it goes GA so people know. And before it GAs (if ever), it's much more obvious[^1] that it's preview.

/cc @RickWinter @ronniegeraghty @JeffreyRichter for thoughts.

[^1]: Especially since cargo/crates.io doesn't *yet* support feature docs.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -666,23 +683,38 @@ impl ContainerClient {
...
-        partition_key: impl Into<QueryPartitionStrategy>,
+        partition_key: impl Into<PartitionKey>,
         options: Option<QueryOptions<'_>>,
     ) -> azure_core::Result<FeedPager<T>> {
-        let options = options.unwrap_or_default();
-        let url = self.pipeline.url(&self.items_link);
-        let mut base_request = Request::new(url, Method::Post);
-        let QueryPartitionStrategy::SinglePartition(partition_key) = partition_key.into();
-        base_request.insert_headers(&partition_key)?;
+        #[allow(unused_mut)] // This is used in the #[cfg(feature = "query_engine")] block
```
heaths, line 692: Maybe this instead, then? Less fragile to accidents - almost like `#[expect]` you wanted back when (and I'd still like to use).
```suggestion
        #[cfg_attr(not(feature = "query_engine"), allow(unused_mut))]
```

sdk/cosmos/azure_data_cosmos/Cargo.toml
heaths: I talked with @RickWinter offline and he agrees for now: lets preface preview features with "preview_"; though, we're still open to discussion, of course.

The idea is that:

1. It's obvious it's preview whether you see the docs or not (we know how good us devs are with reading docs *beforehand*).
2. If it GAs, the customer will know for sure.
3. If it GAs, the feature might just be rolled in and crate maintainers may not want to support it as an optional feature (creates all sorts of hassles that might otherwise be unnecessary). So if you removed the feature to roll it in, they break anyway.
4. If the feature is cut and you remove it, the customer is broken anyway.

I don't really see any downside except if you keep it through GA, but then at least it's obvious that the feature is now stable.

If we're all in agreement, I'll update the guidelines.

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
```diff
@@ -95,6 +77,19 @@ impl<T: Into<PartitionKey>> From<T> for QueryPartitionStrategy {
...
 pub struct PartitionKey(Vec<PartitionKeyValue>);
 
+impl PartitionKey {
+    /// A single null partition key value, which can be used as the sole partition key or as part of a hierarchical partition key.
+    pub const NULL: PartitionKeyValue = PartitionKeyValue(InnerPartitionKeyValue::Null);
+
+    /// An empty list of partition key values, which is used to signal a cross-partition query, when querying a container.
+    pub const EMPTY: PartitionKey = PartitionKey(Vec::new());
+
+    #[allow(dead_code)] // Used in the query_engine feature.
```
heaths, line 87: Same suggestion about `cfg_attr` here.

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
```diff
@@ -95,6 +77,19 @@ impl<T: Into<PartitionKey>> From<T> for QueryPartitionStrategy {
...
+    pub(crate) fn is_empty(&self) -> bool {
+        self.0.is_empty()
```
heaths, line 89: Presumably not, but just to ask the question since I'm not an expert in Cosmos: should this also return `true` for `EMPTY`?

sdk/cosmos/azure_data_cosmos/src/query/engine.rs
```diff
@@ -0,0 +1,85 @@
+/// Represents a request from the query pipeline for data from a specific partition key range.
+pub struct QueryRequest {
```
heaths, line 2: Don't want a `Debug` or `SafeDebug` derive here (or elsewhere, if definitely no PII)? These are affectively models, for which we de facto derive `SafeDebug`.

sdk/cosmos/azure_data_cosmos/src/query/executor.rs
```diff
@@ -0,0 +1,202 @@
...
+    base_request: Request,
+    query: Query,
+    pipeline: Option<OwnedQueryPipeline>,
+
+    // Why is our phantom type a function? Because that represents how we _use_ the type T.
+    // Normally, PhantomData<T> is only Send/Sync if T is, because PhantomData is indicating that while we don't _name_ T in a field, we should act as though we have a field of type T.
+    // However, we don't store any T values in this, we only RETURN them.
+    // That means we use a function pointer to indicate that we don't actually operate on T directly, we just return it.
+    // Because of this, PhantomData<fn() -> T> is Send/Sync even if T isn't (see https://doc.rust-lang.org/stable/nomicon/phantom-data.html#table-of-phantomdata-patterns)
+    phantom: std::marker::PhantomData<fn() -> T>,
```
heaths, line 27: No lint on this?
```suggestion
    _phantom: std::marker::PhantomData<fn() -> T>,
```

sdk/cosmos/azure_data_cosmos/src/query/executor.rs
```diff
@@ -0,0 +1,202 @@
...
+                        .create_pipeline(&self.query.text, &query_plan, &pkranges)?;
+                self.pipeline = Some(pipeline);
+                self.pipeline.as_mut().expect("we just set it")
+            }
+        };
+
+        // Execute the pipeline to get a page of results
+        while !pipeline.complete() {
+            // Run the pipeline
+            let results = pipeline.run()?;
```
heaths, line 108: Since `complete()` returns immediately, are we basically just taking advantage of any async waits in `pipeline.run()`? Otherwise, this seems very CPU-intensive and wasteful.

heaths (changes requested):
Mostly nits. Overall, LGTM!

sdk/typespec/typespec_client_core/src/http/context.rs
heaths: Given you commented on a previous commit and I was viewing all, this didn't show up for me to comment on, so see <https://github.com/Azure/azure-sdk-for-rust/pull/2577#discussion_r2080733536>. I'm fine with this in concept, but some suggestions/brainstorming on exactly impl.

analogrelay:
> Is that how Cosmos refers to these keyless partitions then, using "null" in docs? If so, seems reasonable. Just didn't want to see "null" anywhere in Rust unless it's idiomatic to the service.

Cosmos uses the JSON data model, so when I refer to `null` here, I'm referring to a JSON `null` literal. A `null` partition key is **not** the same as a missing partition key. It refers to the partition containing the actual JSON value `null`. For example, if your partition key is the `partitionKey` property, then this is the partition containing objects like this: `{ "partitionKey": null }`. Because this is permitted in the Cosmos data model, we need a way to represent it. In fact, I believe we also need a way to represent the JSON `undefined` value (i.e. the lack of a `"partitionKey"` property), but I'll file a bug to track that rather than jam it in here.

.vscode/settings.json
analogrelay: Ah yeah, good catch. Shall I gitignore it now and remove it from the repo?

sdk/cosmos/azure_data_cosmos/Cargo.toml
analogrelay: I'm quite happy to prefix this `preview_`, in fact I did start with that. One difference in the Rust community over others like Go, is that the idea of unstable/preview APIs in "GA" releases is accepted to some degree. Prefixing this with `preview_` would just mean there's a "breaking change" (in this **unstable** API) if the feature remains but is promoted to "GA" status. Since the feature is unstable from the start, I'd be quite happy to accept that breaking change.

I'll update it now!

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
analogrelay: Yep, I have done that in other places, I like it better. I'll do that!

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
analogrelay: This should indeed return true for `EMPTY`. `assert!(PartitionKey::EMPTY.is_empty())` would not panic.

sdk/cosmos/azure_data_cosmos/src/query/engine.rs
analogrelay: Yep, I can derive `SafeDebug`, though I may file an issue for a separate pass since I believe we don't derive `SafeDebug` in our other models today and we should.

sdk/typespec/typespec_client_core/src/http/context.rs
analogrelay: That's a good point, I had actually completely missed `with_context` 🤦🏻‍♀️ . Given that we use `with_context` heavily already and it works fine for my case, I'm actually ok just removing this for now. If we do want to switch to something called as a method instead of a constructor, I'm quite happy to do that as a separate refactoring.

.vscode/settings.json
analogrelay: Oh, just realized this was in `settings.json`. For now I'll just remove it. The problem with `launch.json` (though it's just an aesthetics/usability problem) is that it doesn't seem to affect the "Run Test/Debug Test" inlay links on top of the test itself, which is what this setting does. You can absolutely create a `launch.json` config for running a test under the debugger with certain env vars though, it's just a bit of a personal annoyance.

Anyway, I'll just remove this for now.

sdk/cosmos/azure_data_cosmos/src/query/executor.rs
analogrelay: Nope! `PhantomData` is specifically handled in the dead-code analysis pass (https://github.com/rust-lang/rust/blob/dcecb99176edf2eec51613730937d21cdd5c8f6e/compiler/rustc_passes/src/dead.rs#L944). A field of type `PhantomData` is always considered live, even if it's never read.

analogrelay:
I've addressed PR feedback. I don't _think_ the previous build failures were in the Cosmos crate, they appeared to be coming from `azure_keyvault_certificates` (see below), but I'll see if they reoccur after rebasing the branch.

<details><summary>Details</summary>
<p>

```
Compiling azure_security_keyvault_certificates v0.3.0 (/mnt/vss/_work/1/s/sdk/keyvault/azure_security_keyvault_certificates)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 6.11s
     Running unittests src/lib.rs (/mnt/vss/_work/1/s/target/debug/deps/azure_security_keyvault_certificates-f288530cefeeed1d)

running 6 tests
test resource::tests::canonicalizes ... ok
test resource::tests::from_certificate_bundle ... ok
test resource::tests::from_certificate_operation ... ok
test resource::tests::try_from_str ... ok
test resource::tests::test_deconstruct ... ok
test resource::tests::try_from_url ... ok

test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/certificate_client.rs (/mnt/vss/_work/1/s/target/debug/deps/certificate_client-e3c673cadf979af0)

running 5 tests
test sign_jwt_with_ec_certificate ... FAILED
test purge_certificate ... FAILED
test certificate_roundtrip ... FAILED
test list_certificates ... FAILED
test update_certificate_properties ... FAILED

failures:

---- sign_jwt_with_ec_certificate stdout ----
Error: Error { context: Full(Custom { kind: HttpResponse { status: NotFound, error_code: None }, error: HttpError { status: NotFound, details: ErrorDetails { code: None, message: None }, body: {"Message":"Unable to find a record for the request POST https://sanitized.vault.azure.net/certificates/ec-certificate-signer/create?api-version=7.6-preview.2\nHeader differences:\n    \u003CAccept-Encoding\u003E is absent in request, value \u003Cgzip, deflate\u003E\nBody differences:\nRemaining Entries:\n0: https://sanitized.vault.azure.net/certificates/ec-certificate-signer/create?api-version=7.6-preview.2\n1: https://sanitized.vault.azure.net/certificates/ec-certificate-signer/pending?api-version=7.6-preview.2\n2: [https://Sanitized.vault.azure.net/keys/ec-certificate-signer//sign?api-version=7.6-preview.2\n","Status":"NotFound"},](https://sanitized.vault.azure.net/keys/ec-certificate-signer//sign?api-version=7.6-preview.2\n%22,%22Status%22:%22NotFound%22},) .. } }, "server returned error status which will not be retried: 404") }

---- purge_certificate stdout ----
Error: Error { context: Full(Custom { kind: HttpResponse { status: NotFound, error_code: None }, error: HttpError { status: NotFound, details: ErrorDetails { code: None, message: None }, body: {"Message":"Unable to find a record for the request POST https://sanitized.vault.azure.net/certificates/purge-certificate/create?api-version=7.6-preview.2\nHeader differences:\n    \u003CAccept-Encoding\u003E is absent in request, value \u003Cgzip, deflate\u003E\nBody differences:\nRemaining Entries:\n0: https://sanitized.vault.azure.net/certificates/purge-certificate/create?api-version=7.6-preview.2\n1: https://sanitized.vault.azure.net/certificates/purge-certificate/pending?api-version=7.6-preview.2\n2: https://sanitized.vault.azure.net/certificates/purge-certificate?api-version=7.6-preview.2\n3: https://sanitized.vault.azure.net/deletedcertificates/purge-certificate?api-version=7.6-preview.2\n4: https://sanitized.vault.azure.net/deletedcertificates/purge-certificate?api-version=7.6-preview.2\n5: https://sanitized.vault.azure.net/deletedcertificates/purge-certificate?api-version=7.6-preview.2\n6: [https://Sanitized.vault.azure.net/deletedcertificates/purge-certificate?api-version=7.6-preview.2\n","Status":"NotFound"},](https://sanitized.vault.azure.net/deletedcertificates/purge-certificate?api-version=7.6-preview.2\n%22,%22Status%22:%22NotFound%22},) .. } }, "server returned error status which will not be retried: 404") }

---- certificate_roundtrip stdout ----
Error: Error { context: Full(Custom { kind: HttpResponse { status: NotFound, error_code: None }, error: HttpError { status: NotFound, details: ErrorDetails { code: None, message: None }, body: {"Message":"Unable to find a record for the request POST https://sanitized.vault.azure.net/certificates/certificate-roundtrip/create?api-version=7.6-preview.2\nHeader differences:\n    \u003CAccept-Encoding\u003E is absent in request, value \u003Cgzip, deflate\u003E\nBody differences:\nRemaining Entries:\n0: https://sanitized.vault.azure.net/certificates/certificate-roundtrip/create?api-version=7.6-preview.2\n1: https://sanitized.vault.azure.net/certificates/certificate-roundtrip/pending?api-version=7.6-preview.2\n2: [https://Sanitized.vault.azure.net/certificates/certificate-roundtrip/?api-version=7.6-preview.2\n","Status":"NotFound"},](https://sanitized.vault.azure.net/certificates/certificate-roundtrip/?api-version=7.6-preview.2\n%22,%22Status%22:%22NotFound%22},) .. } }, "server returned error status which will not be retried: 404") }

---- list_certificates stdout ----
Error: Error { context: Full(Custom { kind: HttpResponse { status: NotFound, error_code: None }, error: HttpError { status: NotFound, details: ErrorDetails { code: None, message: None }, body: {"Message":"Unable to find a record for the request POST https://sanitized.vault.azure.net/certificates/list-certificates-1/create?api-version=7.6-preview.2\nHeader differences:\n    \u003CAccept-Encoding\u003E is absent in request, value \u003Cgzip, deflate\u003E\nBody differences:\nRemaining Entries:\n0: https://sanitized.vault.azure.net/certificates/list-certificates-1/create?api-version=7.6-preview.2\n1: https://sanitized.vault.azure.net/certificates/list-certificates-1/pending?api-version=7.6-preview.2\n2: https://sanitized.vault.azure.net/certificates/list-certificates-2/create?api-version=7.6-preview.2\n3: https://sanitized.vault.azure.net/certificates/list-certificates-2/pending?api-version=7.6-preview.2\n4: [https://Sanitized.vault.azure.net/certificates?api-version=7.6-preview.2\n","Status":"NotFound"},](https://sanitized.vault.azure.net/certificates?api-version=7.6-preview.2\n%22,%22Status%22:%22NotFound%22},) .. } }, "server returned error status which will not be retried: 404") }

---- update_certificate_properties stdout ----
Error: Error { context: Full(Custom { kind: HttpResponse { status: NotFound, error_code: None }, error: HttpError { status: NotFound, details: ErrorDetails { code: None, message: None }, body: {"Message":"Unable to find a record for the request POST https://sanitized.vault.azure.net/certificates/update-properties/create?api-version=7.6-preview.2\nHeader differences:\n    \u003CAccept-Encoding\u003E is absent in request, value \u003Cgzip, deflate\u003E\nBody differences:\nRemaining Entries:\n0: https://sanitized.vault.azure.net/certificates/update-properties/create?api-version=7.6-preview.2\n1: https://sanitized.vault.azure.net/certificates/update-properties/pending?api-version=7.6-preview.2\n2: https://sanitized.vault.azure.net/certificates/update-properties/?api-version=7.6-preview.2\n3: [https://Sanitized.vault.azure.net/certificates/update-properties/c1b6ef48c8f24ac5b3b9878131706dd2?api-version=7.6-preview.2\n","Status":"NotFound"},](https://sanitized.vault.azure.net/certificates/update-properties/c1b6ef48c8f24ac5b3b9878131706dd2?api-version=7.6-preview.2\n%22,%22Status%22:%22NotFound%22},) .. } }, "server returned error status which will not be retried: 404") }


failures:
    certificate_roundtrip
    list_certificates
    purge_certificate
    sign_jwt_with_ec_certificate
    update_certificate_properties

test result: FAILED. 0 passed; 5 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.89s
```

</p>
</details>

.vscode/settings.json
heaths: Right, that is a downside. But you can come close with the right selection criteria e.g.,

```sh
cargo test -p azure_data_cosmos --test test_module_name specific_test_name_pattern
```

It'd be great if VSCode had layered settings within a repo, or a way to limit user settings to a specific directory, but I'm not aware of any capability.

sdk/cosmos/azure_data_cosmos/Cargo.toml
heaths: Whether we even allow preview features post-GA is something we need to discuss, but I'm in favor in exceptional cases, FWIW.

sdk/cosmos/azure_data_cosmos/src/query/engine.rs
heaths: When you're sure a model wouldn't include PII, feel free to derive `Debug`.

sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -25,7 +25,7 @@ url.workspace = true
 
 [dev-dependencies]
 azure_identity.workspace = true
-azure_core_test = { workspace = true, features = ["tracing"] }
```
heaths: You didn't want tracing for troubleshooting tests?

sdk/cosmos/azure_data_cosmos/src/query/executor.rs
```diff
@@ -76,7 +76,7 @@ impl<T: DeserializeOwned + 'static> QueryExecutor<T> {
                 let query_plan = get_query_plan(
                     &self.http_pipeline,
                     &self.items_link,
-                    self.context.into_borrowed(),
+                    Context::with_context(&self.context),
```
heaths, line 79: Now that I see this, I think I am in favor of adding `into_borrowed()` (just have one call the other, perhaps) but I wouldn't hold up this PR for it. It was definitely cleaner and more intuitive the other way.

heaths (approved):
A couple of thoughts, but nothing to hold up the PR.

sdk/cosmos/azure_data_cosmos/src/query/executor.rs
analogrelay: I believe `with_context` is used in a lot of generated code, so I wanted to avoid changing it in this PR. But I agree `to_borrowed` feels cleaner

analogrelay: https://github.com/Azure/azure-sdk-for-rust/issues/2591

heaths:
I see why the tests are failing. It's because of #2574 to fix #2551. "deflate" and "gzip" are no longer sent as `accept-encoding` headers. Basically, we want the default features intact but `azure_core_amqp` can't have that. I need to figure out a different way.

heaths:
You might want to wait for #2595.

heaths:
@analogrelay #2595 is in now, so you should be unblocked if you rebase.

Pilchie (approved):

heaths (approved):

PR #2584, opened 2025-05-09
Downgrade source version of typespec package to 0.4.0

LarryOsterman:


LarryOsterman:
Because AMQP has a workspace dependency on typespec which is converted by the pipelines by our publishing pipeline.

heaths:
If you take an explicit dependency on the exact version of `azure_core_amqp` and/or `azure_core` you want, the transitive dependency tree will work out fine. It's a local change to your crates and has zero repercussions for everyone else.

PR #2594, opened 2025-05-09
Added `AmqpSimpleValue` to enable compile time checks on AmqpApplicationProperties; Reworked type conversions to remove unnecessary clones.

LarryOsterman:
There are two major parts to this PR. The first is purely functional, the second should significantly reduce the overhead of converting messages to and from Fe2o3.

1. Adds support for `AmqpSimpleValue`, which is defined by the AMQP specification as an AmqpValue without `Map`, `List`, or `Array` fields. This ensures that we have a compile time error when attempting to put an unsupported value into an `AmqpApplicationProperties` (or `EventData` `properties`
2. Reworks all the fe2o3 conversion functions with an eye to minimizing the number of conversions back and forth. 
3. Also adds in the 3 missing AMQP fundamental types - `Decimal32`, `Decimal64 `and `Decimal128` - these are implemented as opaque arrays of 4, 8, and 16 bytes.

As an example of this, when receiving a message from EventHubs, the code uses the `AmqpDelivery::into_message` function to convert the received (owned) fe2o3 message into an `AmqpMessage`. Before this change, the code would convert the received fe2o3 message by reference into an `AmqpMessage` (thus copying the contents of the fe2o3 message into the `AmqpMessage`) and then extract the converted `AmqpMessage` from the `AmqpDelivery`, only freeing the fe2o3 message when the `AmqpDelivery` is freed.

In the new version, the code consumes the fe2o3 AMQP message directly converting it to an `AmqpMessage` without actually copying the original message. That should result in many elements being moved out of the fe2o3 message rather than being copied.

This change also removes most (all?) of the use of the anti-pattern `.clone().into()` which is typically used when converting between a reference to a type and another type - this change adds `From<&Type>` specializations to enable direct conversion from the reference to type and the underlying type.

### DIFF
--- a/sdk/core/azure_core_amqp/src/fe2o3/sender.rs
+++ b/sdk/core/azure_core_amqp/src/fe2o3/sender.rs
@@ -168,20 +168,22 @@ impl AmqpSenderApis for Fe2o3AmqpSender {
             }
             fe2o3_amqp_types::messaging::Outcome::Released(_) => AmqpSendOutcome::Released,
             fe2o3_amqp_types::messaging::Outcome::Modified(m) => {
-                AmqpSendOutcome::Modified(m.into())
+                AmqpSendOutcome::Modified((&m).into())
             }
         })
     }
 }
 
-impl From<fe2o3_amqp_types::messaging::Modified> for SendModification {
-    fn from(m: fe2o3_amqp_types::messaging::Modified) -> Self {
+impl From<&fe2o3_amqp_types::messaging::Modified> for SendModification {
+    fn from(m: &fe2o3_amqp_types::messaging::Modified) -> Self {
         Self {
             delivery_failed: m.delivery_failed,
             undeliverable_here: m.undeliverable_here,
-            message_annotations: m
-                .message_annotations
-                .map(AmqpOrderedMap::<AmqpSymbol, AmqpValue>::from),
+            message_annotations: m.message_annotations.as_ref().map(|m| {
+                m.iter()
+                    .map(|(k, v)| (k.into(), v.into()))
+                    .collect::<AmqpOrderedMap<AmqpSymbol, AmqpValue>>()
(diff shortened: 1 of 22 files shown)


sdk/core/azure_core_amqp/src/fe2o3/sender.rs
```diff
@@ -166,20 +168,22 @@ impl AmqpSenderApis for Fe2o3AmqpSender {
             }
             fe2o3_amqp_types::messaging::Outcome::Released(_) => AmqpSendOutcome::Released,
             fe2o3_amqp_types::messaging::Outcome::Modified(m) => {
-                AmqpSendOutcome::Modified(m.into())
+                AmqpSendOutcome::Modified((&m).into())
```
heaths, line 171: Tip: you could capture `Modified(ref m)` and then just to `m.into()`.

sdk/core/azure_core_amqp/src/fe2o3/value.rs
```diff
@@ -76,103 +85,234 @@ impl From<AmqpTimestamp> for fe2o3_amqp_types::primitives::Timestamp {
...
 }
 
-impl From<AmqpValue> for fe2o3_amqp_types::primitives::SimpleValue {
-    fn from(v: AmqpValue) -> Self {
+impl From<&AmqpTimestamp> for fe2o3_amqp_types::primitives::Timestamp {
+    fn from(timestamp: &AmqpTimestamp) -> Self {
+        if let Some(t) = timestamp.0 {
+            let t = t
+                .duration_since(UNIX_EPOCH)
+                .expect("Could not convert timestamp to time since unix epoch")
```
heaths, line 93: This should be in a `TryFrom` then and not panic.

heaths (approved):

heaths (approved):

PR #2597, opened 2025-05-11
Created initial structure for future servicebus development

dawidpereira:
This PR sets up the foundational structure for the ⁠`azure_messaging_servicebus` library, including package configuration, initial README documentation, and licensing. It prepares the project for further development of the Azure Service Bus SDK.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -328,6 +328,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "azure_messaging_servicebus"
+version = "0.1.0"
+
+[[package]]
 name = "azure_security_keyvault_certificates"
 version = "0.3.0"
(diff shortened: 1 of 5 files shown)


dawidpereira:

heaths:
Thank you for the contribution and your interest. Since we'll have to maintain this, we first need to prioritize it. @ronniegeraghty could you have a look? We should talk with the service team.

We are also a generation-first Azure SDK language, which means we generate as much as possible. For Service Bus that might be little to nothing - as with Event Hubs - but @LarryOsterman would be best to advise if we're ready to accept work currently.

heaths (changes requested):
I recommend holding off until we prioritize maintenance of a Service Bus crate.

PR #2601, opened 2025-05-13
Github Instructions for onboarding performance related code and pipeline 

gearama:
Added copilot instructions for Rust performance framework onboarding.
The instructions are "burrowed" from Heath Stewart's PR about copilot instructions. I used those for consistency so when his PR is merged we work on similar assumptions and don't confuse the instructions
. 
The supported operations are : 
Onboard perf framework for package azure_security_keyvault_certificates.
Create perf pipeline for package azure_security_keyvault_certificates.

The code file generated might need some quick edits to make it work. 

I would not say this is a definitive list. Improvements are welcome.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,19 @@
+# Instructions
+
+You are an expert Rust programmer. You write safe, efficient, maintainable, and well-tested code.
+
+* Use an informal tone.
+* Do not be overly apologetic and focus on clear guidance.
+* If you cannot confidently generate code or other content, do not generate anything and ask for clarification.
+
+## Onboard perf framework for package named name
+
+* In cargo.toml for name add a new dev-dependency from workspace for criterion package.
+* Create a benches\benchmark.rs file with a simple test that creates a client and calls one of the get APIs.
+* In cargo.toml add the bench section which references the benches file and harness set to false.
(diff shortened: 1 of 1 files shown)


heaths (changes requested):
Let's hold off. I want to get the initial copilot-instructions.md in first. How they are now isn't quite right. Then we can update this PR to add perf-related instructions; though, a few things I see off the bat:

1. Use more concrete wording. "Add performance support to crates" is going to work better, most likely, than using words like "onboard" or "perf". The clearer we are the better it'll work.
2. Too much written detail. Your example near the bottom is likely going to yield much better results.
3. After working in our Azure MCP code base a little, I wonder if this is too much to put in general instructions. That's why I actually moved commit and pull request instructions out. I wonder if we should actually consider using [prompt file examples](https://docs.github.com/en/copilot/customizing-copilot/adding-repository-custom-instructions-for-github-copilot?tool=vscode#prompt-file-examples) for specific use cases like perf tests and examples.

For the latter, I'd encourage you to give that a try. The more we put in copilot-instructions.ms, 1) the slower and more confusing responses we'll probably get, and 2) the number of tokens will cost more. We need to find that sweet spot, and adding new perf tests isn't exactly the highest priority order.

heaths:
Replaced by #2401. I've finished the initial instructions and tested them out and added these instructions to a custom prompt file since they'd be relatively seldom used.

PR #2602, opened 2025-05-13
First cut at EH perf tests

LarryOsterman:
They don't do very much yet, but they're a start.

The perf pipelines are not yet hooked up, I'll look at that when this is checked in.

### DIFF
--- a/sdk/eventhubs/azure_messaging_eventhubs/benches/benchmarks.rs
+++ b/sdk/eventhubs/azure_messaging_eventhubs/benches/benchmarks.rs
@@ -1,0 +1,126 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+use azure_identity::DefaultAzureCredential;
+use azure_messaging_eventhubs::{models::EventData, ProducerClient, SendEventOptions};
+use criterion::{criterion_group, criterion_main, Criterion};
+use std::{env, sync::Arc};
+use tokio::runtime::Runtime;
+
+static INIT_LOGGING: std::sync::Once = std::sync::Once::new();
+
+fn setup() {
+    INIT_LOGGING.call_once(|| {
+        println!("Setting up test logger...");
+
+        use tracing_subscriber::{fmt::format::FmtSpan, EnvFilter};
+        tracing_subscriber::fmt()
+            .with_env_filter(EnvFilter::from_default_env())
+            .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
+            .with_ansi(std::env::var("NO_COLOR").map_or(true, |v| v.is_empty()))
+            .with_writer(std::io::stderr)
+            .init();
+    });
+}
+
+fn send_batch_benchmark(c: &mut Criterion) {
+    setup();
+    let rt = Runtime::new().unwrap();
+
+    let namespace = env::var("EVENTHUBS_HOST")
+        .expect("Set EVENTHUBS_HOST in env (e.g. my-namespace.servicebus.windows.net)");
+    let event_hub = env::var("EVENTHUB_NAME").expect("Set EVENTHUB_NAME in env");
+
+    let credential =
+        DefaultAzureCredential::new().expect("Failed to create DefaultAzureCredential");
+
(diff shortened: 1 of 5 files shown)


sdk/eventhubs/azure_messaging_eventhubs/perf.yml
```diff
@@ -0,0 +1,35 @@
...
+  type: string
+  default: 'source'
+- name: Tests
+  displayName: Tests (regex of tests to run)
+  type: string
+  default: '.*'
+- name: Arguments
+  displayName: Arguments (regex of arguments to run)
+  type: string
+  default: '.*'
```
heaths, line 13: More so feedback to @gearama, but package and test selection don't support regex. Package selection does support basic Unix glob patterns, but test selection is just a simple match:

1. https://doc.rust-lang.org/cargo/commands/cargo-test.html (package selection)
2. https://doc.rust-lang.org/rustc/tests/index.html#filters (test selection

sdk/eventhubs/azure_messaging_eventhubs/benches/benchmarks.rs
```diff
@@ -0,0 +1,126 @@
...
+
+        use tracing_subscriber::{fmt::format::FmtSpan, EnvFilter};
+        tracing_subscriber::fmt()
+            .with_env_filter(EnvFilter::from_default_env())
+            .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
+            .with_ansi(std::env::var("NO_COLOR").map_or(true, |v| v.is_empty()))
+            .with_writer(std::io::stderr)
+            .init();
+    });
+}
```
heaths, line 24: Probably not a good idea to enable in perf tests. I appreciate it's based on whether `RUST_LOG` is set, but do we know how much code dedicated to tracing is skipped? If it truly is as efficient as the calls you already make in the shipping crates not really doing much of anything, this is probably fine. Might want to measure, though. If anything, for perf tests, this probably isn't really necessary compared to integration tests and examples anyway, IMO.

heaths (approved):
A couple nits/questions/concerns, but nothing blocking.

sdk/eventhubs/azure_messaging_eventhubs/perf.yml
heaths: @mikeharder Are these parameters mentioning "regex" common across all languages, or is there any problem with changing the names for rust just to match what our test package/name selection actually supports (unix glob patterns for packages but simple substring match for tests)?

mikeharder: The parameters are passed into the shared perf automation, which does the regex matching:

https://github.com/Azure/azure-sdk-tools/blob/71a6250c9feb93e1bc6497bc3986e9498d87700c/tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Program.cs#L159-L171

sdk/eventhubs/azure_messaging_eventhubs/benches/benchmarks.rs
```diff
@@ -0,0 +1,139 @@
...
+    //         .with_writer(std::io::stderr)
+    //         .init();
+    // });
+}
+
+fn send_batch_benchmark(c: &mut Criterion) {
+    setup();
+
+    // Check if the environment variable is set thus allowing the benchmarks to run
+    if azure_core_test::TestMode::current().unwrap() != azure_core_test::TestMode::Live {
```
heaths, line 30: I'd strongly urge `unwrap_or_default` or `unwrap_or(Playback)` or something like that. It can fail with a bad env var value. I suppose at least it fails in the way we want, but won't print your informative message below it.

heaths (approved):

sdk/eventhubs/azure_messaging_eventhubs/benches/benchmarks.rs
LarryOsterman: I'll pick this up in my next PR - it's a great suggestion. But now I have to make AMQP changes so I can merge AMQP into C++ :(.

PR #2603, opened 2025-05-13
Delete azure_identity dead code

### LINKED ISSUE #2281: Use or remove any dead_code in azure_identity

chlowell:
Closes #2281


Cargo.lock
```diff
@@ -551,21 +535,6 @@ version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
-[[package]]
-name = "chrono"
```
heaths, line 555: I'm genuinely shocked `oauth2` was the only crate in all our dependencies using `chrono`!

sdk/identity/azure_identity/src/credentials/cache.rs
```diff
@@ -15,13 +15,6 @@ impl TokenCache {
         Self(RwLock::new(HashMap::new()))
     }
 
-    #[allow(dead_code)]
-    pub(crate) async fn clear(&self) -> azure_core::Result<()> {
-        let mut token_cache = self.0.write().await;
-        token_cache.clear();
-        Ok(())
-    }
-
```
heaths, line 24: We're definitely going to need this for testing. You might consider commenting it out, or maybe we just add it back in once we start implementing more mock tests or whatever.

chlowell: We actually don't need this for tests:
- there's no behavior requiring this method to test because credentials don't have contractual behavior involving emptying the cache
- tests don't need this method to manipulate credential behavior in general because each new credential instance gets its own empty cache instance

heaths: > each new credential instance gets its own empty cache instance

That's a good point. I hadn't considered that in Rust, it's pretty much a given each test is working on isolated data. In .NET, we had a number of credential test-related code e.g., Key Vault's bearer token policy where we did need to clear the cache but only because some data was shared across individual tests that was idiomatic for NUnit et. al.

PR #2608, opened 2025-05-14
Expose more types for C++; reorganized some public types a bit

LarryOsterman:
This PR adds a few public types which are needed for the C++ AMQP implementation.

It also moves some types from the global namespace and into a `builder` and `message` namespace.

Verified that the C++ codebase builds and works with all of these changes.

Bonus: Fixed two late breaking suggestions for the EventHubs performance tests.

### DIFF
--- a/sdk/eventhubs/azure_messaging_eventhubs/benches/benchmarks.rs
+++ b/sdk/eventhubs/azure_messaging_eventhubs/benches/benchmarks.rs
@@ -27,7 +27,9 @@ fn send_batch_benchmark(c: &mut Criterion) {
     setup();
 
     // Check if the environment variable is set thus allowing the benchmarks to run
-    if azure_core_test::TestMode::current().unwrap() != azure_core_test::TestMode::Live {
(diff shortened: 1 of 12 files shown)


sdk/eventhubs/azure_messaging_eventhubs/benches/benchmarks.rs
```diff
@@ -27,7 +27,9 @@ fn send_batch_benchmark(c: &mut Criterion) {
     setup();
 
     // Check if the environment variable is set thus allowing the benchmarks to run
-    if azure_core_test::TestMode::current().unwrap() != azure_core_test::TestMode::Live {
+    if azure_core_test::TestMode::current().unwrap_or(azure_core_test::TestMode::Playback)
```
heaths, line 30: This is fine, but `TestMode` derives `Default` with `Playback` as the default, so you can just do `unwrap_or_default()`.

heaths (approved):

heaths (approved):

PR #2611, opened 2025-05-15
Remove AccessToken::is_expired()

chlowell:
...because it's unnecessary and its name is misleading. This PR also adds test coverage and fixes a bug: the policy currently returns errors from proactive refresh attempts even when it has a valid cached token. (There's no value in doing that because the client request could otherwise succeed.)

### DIFF
--- a/sdk/core/azure_core/CHANGELOG.md
+++ b/sdk/core/azure_core/CHANGELOG.md
@@ -6,7 +6,11 @@
 
 ### Breaking Changes
 
+- Removed `AccessToken::is_expired()`
+
 ### Bugs Fixed
+
+- `BearerTokenCredentialPolicy` returns an error when a proactive token refresh attempt fails
 
 ### Other Changes
 
--- a/sdk/core/azure_core/src/http/policies/bearer_token_policy.rs
+++ b/sdk/core/azure_core/src/http/policies/bearer_token_policy.rs
@@ -116,3 +116,151 @@ impl Policy for BearerTokenCredentialPolicy {
 fn should_refresh(expires_on: &OffsetDateTime) -> bool {
     *expires_on <= OffsetDateTime::now_utc() + Duration::from_secs(300)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{
+        credentials::{Secret, TokenCredential},
+        http::{
+            headers::{Headers, AUTHORIZATION},
+            policies::Policy,
+            Request, Response, StatusCode,
+        },
+        Bytes, Result,
+    };
+    use async_trait::async_trait;
+    use azure_core_test::http::MockHttpClient;
+    use futures::FutureExt;
+    use std::sync::{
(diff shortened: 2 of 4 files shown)


sdk/core/azure_core/CHANGELOG.md
```diff
@@ -6,8 +6,12 @@
 
 ### Breaking Changes
 
+- Removed `AccessToken::is_expired()`
+
 ### Bugs Fixed
 
+- `BearerTokenCredentialPolicy` returns an error when a proactive token refresh attempt fails
```
heaths, line 13: Especially if this was a customer-filed bug, we reference the bug number such that it renders as `(#{num})`. If not a customer-filed issue, no need.

sdk/core/azure_core/src/http/policies/bearer_token_policy.rs
```diff
@@ -63,16 +61,44 @@ impl Policy for BearerTokenCredentialPolicy {
...
+                if access_token.is_none() {
+                    *access_token = Some(self.credential.get_token(&self.scopes()).await?);
+                }
+            }
+            Some(token) if should_refresh(&token.expires_on) => {
+                // token is expired or within its refresh window. Upgrade the lock and
+                // acquire a new token, provided another thread hasn't already done so
+                let expires_on = token.expires_on;
                 drop(access_token);
                 let mut access_token = self.access_token.write().await;
```
heaths, line 78: We probably don't need to take a write lock this soon - only when we write. Is there any disadvantage to a couple of threads potentially authenticating but only the last writer wins? IIRC, this is how creds in the .NET SDK work, and generally advised if a guarded operation is non-destructive and, in this case, wouldn't get us throttled (I doubt that for auth - not within reason, anyway).

sdk/core/azure_core/src/http/policies/bearer_token_policy.rs
```diff
@@ -86,3 +112,161 @@ impl Policy for BearerTokenCredentialPolicy {
...
+    use std::sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    };
+    use std::time::Duration;
+    use time::OffsetDateTime;
+    use typespec_client_core::http::{policies::TransportPolicy, Method, TransportOptions};
+
+    #[derive(Debug, Clone)]
+    struct MockCredential {
```
heaths, line 144: Just FYI: we actually have one in `azure_core_test`, though maybe this one is purpose-built for this task.

heaths (approved):
Looks great! I see nothing blocking. Just a couple of notes and one suggestion.

sdk/core/azure_core/src/http/policies/bearer_token_policy.rs
chlowell: Throttling is actually a real danger because IMDS allows only 5 requests per second. We may need to add backoff between proactive refresh attempts to reduce the risk further; a highly concurrent app could hammer IMDS during an outage and get throttled.

PR #2613, opened 2025-05-15
Add safe attribute helper to SafeDebug

heaths:


heaths:
Making this a draft for now because I think, instead of copying the pattern we have with the Model derive macro - which @analogrelay is removing - I want to see if it's cleaner w.r.t. error handling to implement `Parse` as I did for `azure_core_test_macros`. This works, though.

sdk/typespec/typespec_macros/src/safe_debug.rs
```diff
@@ -166,3 +187,193 @@ fn to_path(idents: &[&Ident]) -> Path {
...
+    }
+}
+
+const INVALID_SAFE_ATTRIBUTE_MESSAGE: &str =
+    "invalid safe attribute, expected attribute in form #[safe(false)] or #[safe(true)]";
+
+fn parse_attr(attribute: &Attribute, attrs: &mut Attrs) -> Result<()> {
+    let meta_list = attribute
+        .meta
+        .require_list()
```
analogrelay, line 238: One small nit, since marking fields or whole types as "safe" would probably be the most common case, could we make it so `#[safe]` is treated the same as `#[safe(true)]`? Just a bit of syntax sugar. Totally non-blocking.

analogrelay (commented):
Won't approve because it's still draft, but I think this is looking good. We'll be moving all our types to `SafeDebug` and definitely have a lot of "safe" fields, so I'd love to get this in soon!

sdk/typespec/typespec_macros/src/safe_debug.rs
heaths: That was originally my intent, but @LarryOsterman and I discussed this yesterday and decided to stick with an explicit bool because:

1. It's obvious what it means, especially considering that you can inherit the attribute's effects.
2. It's easier to write/maintain the macro.
3. Related to 2, the intent was always to generate this based on something in TypeSpec (there's some open to considering said decorator, but we might have to make it Rust-only for now) and at least *supporting* an explicit value would be easier.

We can always make it more lax later, too. But we can't tighten it up. I was sort of inspired by `clap`'s `#[arg()]` which requires parens as well, as well as `#[serde()]` which, IIRC, actually requires at least one argument; otherwise, what's the point? There's no default.

analogrelay: That all makes sense. Especially given that other similar macro/attribute pairings do require an argument. Let's just require the arg!

analogrelay (approved):

LarryOsterman (approved):

PR #2617, opened 2025-05-16
Cosmos: fix #2616 by using RawValue, and fix query rewriting

### LINKED ISSUE #2616: Cosmos: Allow query engine to return `Box<serde_json::value::RawValue>` for items
The Client Engine, that provides the implementation of the query engine functionality, currently returns items as `Box<serde_json::value::RawValue>`. We should change the API for the `QueryPipeline` trait to accept these as results, because they allow the client engine to hold un-parsed chunks of JSON and pass them back to the SDK to deserialize fully, without incurring extra copies or serialize/deserialize cycles.

analogrelay:
Closes #2616 

It's a tale as old as time. Abstraction is written, then implemented, and found lacking somewhere ;). This PR has a few small fix-ups for the query engine abstraction that I found while integrating the query engine. Once these were fixed, it worked perfectly though! 🎉 

Here are the changes:

The query engine works with `serde_json::value::RawValue`, which represents "unparsed" JSON. It's just a wrapper around `str`, and I had hoped it would be easy to consume it and convert it in to a `Vec<u8>`, so that's what I had the QueryPipeline abstraction expect as a result from the pipeline. Alas, it's not really possible to do that. You can _access_ the underlying bytes via `RawValue::get()` but that returns `&str` and there's no way to "move" the bytes out of the `RawValue`. But that's OK, we can just have the pipeline send back `Box<RawValue>` values directly, since the pipeline is always working with JSON anyway. So, that's what I did. (This is #2616, which I filed when I thought I could work around it and fix the SDK later, but I couldn't do that with the below issue so I decided to just fix them both here first).

There was also a bug in how the query "executor" worked. I built the base query request as soon as the pipeline was created, using the query the user passed in. But the gateway **may rewrite the query**. The query engine handles this and exposes the "authoritative" query out of `QueryPipeline::query()`, but the SDK's executor ignored that and used the original query. A pretty simple fix.


sdk/cosmos/azure_data_cosmos/src/query/executor.rs
```diff
@@ -97,8 +100,16 @@ impl<T: DeserializeOwned + 'static> QueryExecutor<T> {
...
                         .create_pipeline(&self.query.text, &query_plan, &pkranges)?;
+                self.query.text = pipeline.query().into();
+                self.base_request = Some(crate::pipeline::create_base_query_request(
+                    self.http_pipeline.url(&self.items_link),
+                    &self.query,
+                )?);
                 self.pipeline = Some(pipeline);
-                self.pipeline.as_mut().expect("we just set it")
+                (
+                    self.pipeline.as_mut().expect("we just set it"),
```
heaths, line 110: Nit: since this is impossible, maybe just "unwrap" and avoid the pooled string? Just a comment instead?

heaths (approved):

heaths (approved):

Pilchie (approved):

PR #2626, opened 2025-05-22
Add serde helper for string format

jhendrixMSFT:
The common case is to use this for numeric types that are to be encoded/decoded in string format.

Part of https://github.com/Azure/typespec-rust/issues/47


sdk/typespec/typespec_client_core/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ### Features Added
 
 - Added `#[safe]` attribute helper for `SafeDebug` derive macro to show or hide types and members as appropriate.
+- Added module `serde` to contains various helpers.
```
LarryOsterman, line 8:
```suggestion
- Added module `serde` which contains various helpers.
```

sdk/core/azure_core/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Features Added
 
 - Added `#[safe]` attribute helper for `SafeDebug` derive macro to show or hide types and members as appropriate.
+- Added module `serde` from `typespec_client_core`.
```
LarryOsterman, line 8:
```suggestion
- Expose module `serde` from `typespec_client_core`.
```

LarryOsterman (approved):

sdk/typespec/typespec_client_core/CHANGELOG.md
jhendrixMSFT: fixed

sdk/core/azure_core/CHANGELOG.md
jhendrixMSFT: fixed

sdk/typespec/typespec_client_core/src/serde.rs
heaths, line 1: We have serialization helpers in other modules and don't really call out "serde". I think we should put these in the existing `fmt` module (or at least expose them from it only).

heaths (changes requested):

sdk/typespec/typespec_client_core/src/serde.rs
jhendrixMSFT: That was my original thought. I just wasn't sure if this was different enough from formatting that it would conceptually make sense. At any rate, it's been moved.

heaths (approved):

PR #2631, opened 2025-05-23
Added get properties benchmark

LarryOsterman:
Added benchmarks for simple eventhubs get properties and get partition properties. 

Adding to get a quick baseline on these APIs.

### DIFF
--- a/sdk/eventhubs/azure_messaging_eventhubs/benches/benchmarks.rs
+++ b/sdk/eventhubs/azure_messaging_eventhubs/benches/benchmarks.rs
@@ -71,13 +71,97 @@ fn send_batch_benchmark(c: &mut Criterion) {
     });
 }
 
+fn get_eventhub_properties(c: &mut Criterion) {
+    setup();
+
+    // Check if the environment variable is set thus allowing the benchmarks to run
+    if azure_core_test::TestMode::current().unwrap_or_default() != azure_core_test::TestMode::Live {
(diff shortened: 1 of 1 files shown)


sdk/eventhubs/azure_messaging_eventhubs/benches/benchmarks.rs
```diff
@@ -71,13 +71,97 @@ fn send_batch_benchmark(c: &mut Criterion) {
...
+}
+
 criterion_group!(
     name = send_batch_benchmarks;
     config = Criterion::default()
         .sample_size(100)
         .warm_up_time(std::time::Duration::new(1, 0))
-        .measurement_time(std::time::Duration::new(250, 0));
-    targets = send_batch_benchmark
+        .measurement_time(std::time::Duration::new(500, 0));
```
RickWinter, line 163: 250 vs 500 seems a bit magical other than one being longer than the other. Is there a benefit for longer runs?

RickWinter (approved):

sdk/eventhubs/azure_messaging_eventhubs/benches/benchmarks.rs
LarryOsterman: 250 causes criterion to complain that it is too low.
500 silences the criterion warning (technically anything over 450).

It appears that "measurement_time" is a max time for the test to run (250 seconds vs 500 seconds). But in practice the test runs much faster.

PR #2640, opened 2025-05-28
Switch EH to use a bicep file for deployment; Updated benchmarks to work.

LarryOsterman:
Switched EventHubs to use `test-resources.bicep` files instead of `test-resources.json` files.

Updated benchmarks to use .env files instead of the raw environment.

### DIFF
--- a/sdk/eventhubs/azure_messaging_eventhubs/test-resources.bicep
+++ b/sdk/eventhubs/azure_messaging_eventhubs/test-resources.bicep
@@ -48,37 +48,67 @@ resource eventhubNamespace 'Microsoft.EventHub/namespaces@2024-05-01-preview' = 
     maximumThroughputUnits: 0
     kafkaEnabled: true
   }
-}
-
-resource authorization 'Microsoft.EventHub/namespaces/authorizationrules@2024-05-01-preview' = {
-  parent: eventhubNamespace
-  name: 'RootManageSharedAccessKey'
-  properties: {
-    rights: [
-      'Listen'
-      'Manage'
-      'Send'
-    ]
+  resource authorization 'authorizationrules@2024-05-01-preview' = {
+    name: 'RootManageSharedAccessKey'
+    properties: {
+      rights: [
+        'Listen'
+        'Manage'
+        'Send'
+      ]
+    }
   }
-}
-
-resource authorizedListenOnly 'Microsoft.EventHub/namespaces/AuthorizationRules@2017-04-01' = {
-  name: 'ListenOnly'
-  parent: eventhubNamespace
-  properties: {
-    rights: [
-      'Listen'
-    ]
+  resource authorizedListenOnly 'AuthorizationRules@2017-04-01' = {
+    name: 'ListenOnly'
+    properties: {
+      rights: [
+        'Listen'
+      ]
+    }
   }
-}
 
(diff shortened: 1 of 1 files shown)


sdk/eventhubs/azure_messaging_eventhubs/Cargo.toml
```diff
@@ -40,6 +40,7 @@ azure_messaging_eventhubs = { path = ".", features = [
   "test",
 ] }
 criterion.workspace = true
+dotenvy = "0.15.7"
```
heaths, line 43: I'd rather not spread out this dependency all over. Rather than this and the changes in `benchmarks.rs` below, can you make `azure_core_test::find_ancestor_file` public (maybe rename it to `find_ancestor_path` while you're at it) and call that? Remove the `cfg` since it'll be public now. I just added that since we didn't call it from wasm32 code and it was causing build warnings-as-errors.

sdk/eventhubs/azure_messaging_eventhubs/test-resources.bicep
```diff
@@ -0,0 +1,236 @@
...
+    disableLocalAuth: !tenantIsTME // Disable local auth for TME tenants
+    zoneRedundant: false
+    isAutoInflateEnabled: false
+    maximumThroughputUnits: 0
+    kafkaEnabled: true
+  }
+}
+
+resource authorization 'Microsoft.EventHub/namespaces/authorizationrules@2024-05-01-preview' = {
+  parent: eventhubNamespace
```
heaths, line 48: Tip: you can actually embed child resources under their parents and don't need to specify the full type name. Only caveat is that you can't refer to them by reference in, say, output variables.

sdk/eventhubs/azure_messaging_eventhubs/test-resources.bicep
```diff
@@ -0,0 +1,236 @@
...
+resource storageAccount_blobContainer 'Microsoft.Storage/storageAccounts/blobServices/containers@2019-04-01' = {
+  name: 'container'
+  parent: storageAccount_default
+  properties: {}
+}
+
+// Outputs
+output EVENTHUB_NAME string = eventHub.name
+output EVENTHUBS_NAMESPACE string = eventhubNamespace.name
+output EVENTHUBS_HOST string = '${eventhubNamespace.name}.servicebus.windows.net'
```
heaths, line 230: The full URI isn't available in case we provision in another cloud?

heaths (changes requested):

sdk/eventhubs/azure_messaging_eventhubs/test-resources.bicep
LarryOsterman: There is a "serviceBusEndpoint" URL but it's a URL in the HTTPS namespace. And I don't think I can extract just the host from a string in bicep.

sdk/eventhubs/azure_messaging_eventhubs/Cargo.toml
LarryOsterman: Absolutely -  I can add a new public function which calls `dotenvy::from_path` and `find_ancestor_file`, and then update the recorded test start to use the new function.

Issue #1939, opened 2024-12-02
Cosmos: Support creating a client from a connection string


analogrelay:
We support two ways to create a `CosmosClient`:

* `CosmosClient::new(endpoint: &str, credential: Arc<dyn TokenCredential>, options: Option<CosmosClientOptions>)` which uses Entra ID
* `CosmosClient::with_key(endpoint: &str, key: azure_core::Secret, options: Option<CosmosClientOptions>)` which uses key-based auth (**Requires `key_auth` feature**)

We don't support specifying a connection string, since `with_key` accepts all the values a connection string would normally provide. However, often users specify connection strings in config, and our other clients can be constructed from a connection string. So, we should add another method:

* `CosmosClient::with_connection_string(connection_string: azure_core::Secret, options: Option<CosmosClientOptions>)`

This method would require the `key_auth` feature, since it's for specifying a key. Also, the connection string should be a `Secret` because it contains the key.

PR #2641, opened 2025-05-28
Support CosmosClient Creation via Connection String

nehrao1:
### Issue(s) Closed:
Closes https://github.com/Azure/azure-sdk-for-rust/issues/1939
### Description:
Added a new function `CosmosClient::with_connection_string` to create a client via connection string, where the connection string is a secret.
### Approach:
We previously had connection string parsing logic in `test_account.rs` - moved this logic out into a separate file and struct, and made use of it in `with_connection_string` and `TestAccount` setup.
### Testing:
Added unit tests for connection string parsing logic.

### DIFF
--- a/sdk/core/azure_core/src/credentials.rs
+++ b/sdk/core/azure_core/src/credentials.rs
@@ -3,6 +3,7 @@
 
 //! Azure authentication and authorization.
 
+use crate::error::ErrorKind;
 use serde::{Deserialize, Serialize};
 use std::{borrow::Cow, fmt::Debug};
 use typespec_client_core::date::OffsetDateTime;
@@ -26,6 +27,46 @@ impl Secret {
 
     pub fn secret(&self) -> &str {
         &self.0
+    }
+}
+
+/// Represents a Cosmos DB connection string.
+///
+/// The [`Debug`] implementation will not print the secret.
+#[derive(Clone, Deserialize, Serialize)]
+pub struct CosmosConnectionString {
+    pub account_endpoint: Cow<'static, str>,
+    pub account_key: Cow<'static, str>,
+}
+
+impl CosmosConnectionString {
+    pub fn from_secret(secret: &Secret) -> crate::Result<Self> {
+        let split = secret.secret().split(";");
+        let collection = split.collect::<Vec<&str>>();
+
+        match collection.len() {
+            2 => {
+                if (collection[0].contains("AccountEndpoint=")
+                    && collection[1].contains("AccountKey="))
+                {
+                    Ok(Self {
+                        account_endpoint: collection[0][16..collection[0].len() - 1]
+                            .to_owned()
+                            .into(),
+                        account_key: collection[0][11..collection[0].len() - 1].to_owned().into(),
+                    })
+                } else {
+                    Err(crate::Error::message(
+                        ErrorKind::Other,
+                        "Invalid Cosmos connection string",
+                    ))
+                }
+            }
+            _ => Err(crate::Error::message(
+                ErrorKind::Other,
+                "Invalid Cosmos connection string",
+            )),
+        }
     }
 }
 
@@ -62,6 +103,17 @@ impl From<&'static str> for Secret {
 impl Debug for Secret {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.write_str("Secret")
+    }
+}
+
+impl Debug for CosmosConnectionString {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let debug_string = format!(
+            "AccountEndpoint={};AccountKey=Secret;",
+            self.account_endpoint
+        );
+
+        f.write_str(&debug_string)
     }
 }
 
--- a/sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
@@ -3,7 +3,7 @@
 
 use crate::{
     clients::DatabaseClient,
-    models::DatabaseProperties,
+    models::{ConnectionString, DatabaseProperties},
     pipeline::{AuthorizationPolicy, CosmosPipeline},
     resource_context::{ResourceLink, ResourceType},
     CosmosClientOptions, CreateDatabaseOptions, FeedPager, Query, QueryDatabasesOptions,
@@ -17,7 +17,7 @@ use azure_core::{
     },
 };
 use serde::Serialize;
-use std::sync::Arc;
+use std::{fmt::Debug, sync::Arc};
 
 #[cfg(feature = "key_auth")]
 use azure_core::credentials::Secret;
@@ -100,7 +100,7 @@ impl CosmosClient {
     ///
     /// # Arguments
     ///
-    /// * `connection_string` - the connection string to use for the client. E.g.: AccountEndpoint=https://accountname.documents.azure.com:443/‌​;AccountKey=accountk‌​ey
+    /// * `connection_string` - the connection string to use for the client, e.g. AccountEndpoint=https://accountname.documents.azure.com:443/‌​;AccountKey=accountk‌​ey
     /// * `options` - Optional configuration for the client.
     ///
     /// # Examples
@@ -109,17 +109,19 @@ impl CosmosClient {
     /// use azure_data_cosmos::CosmosClient;
     /// use azure_core::credentials::Secret;
     ///
-    /// let client = CosmosClient::with_key("https://myaccount.documents.azure.com/", Secret::from("my_key"), None).unwrap();
+    /// let client = CosmosClient::with_connection_string(
+    ///     "AccountEndpoint=https://accountname.documents.azure.com:443/‌​;AccountKey=accountk‌​ey",
+    ///     None)
+    ///     .unwrap();
     /// ```
     #[cfg(feature = "key_auth")]
     pub fn with_connection_string(
         connection_string: Secret,
         options: Option<CosmosClientOptions>,
-    ) -> azure_core::Result<Self> {
-        use azure_core::credentials::CosmosConnectionString;
+    ) -> Result<Self, azure_core::Error> {
         let options = options.unwrap_or_default();
-        let connection_str = CosmosConnectionString::from_secret(&connection_string)?;
-        let key = Secret::new(connection_str.account_key);
+        let connection_str = ConnectionString::from_secret(&connection_string)?;
+        let key = connection_str.account_key;
 
         Ok(Self {
             databases_link: ResourceLink::root(ResourceType::Databases),
--- a/sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
@@ -96,6 +96,41 @@ impl CosmosClient {
         })
     }
 
+    /// Creates a new CosmosClient, using a connection string.
+    ///
+    /// # Arguments
+    ///
+    /// * `connection_string` - the connection string to use for the client. E.g.: AccountEndpoint=https://accountname.documents.azure.com:443/‌​;AccountKey=accountk‌​ey
+    /// * `options` - Optional configuration for the client.
+    ///
+    /// # Examples
+    ///
+    /// ```rust,no_run
+    /// use azure_data_cosmos::CosmosClient;
+    /// use azure_core::credentials::Secret;
+    ///
+    /// let client = CosmosClient::with_key("https://myaccount.documents.azure.com/", Secret::from("my_key"), None).unwrap();
+    /// ```
+    #[cfg(feature = "key_auth")]
+    pub fn with_connection_string(
+        connection_string: Secret,
+        options: Option<CosmosClientOptions>,
+    ) -> azure_core::Result<Self> {
+        use azure_core::credentials::CosmosConnectionString;
+        let options = options.unwrap_or_default();
+        let connection_str = CosmosConnectionString::from_secret(&connection_string)?;
+        let key = Secret::new(connection_str.account_key);
+
+        Ok(Self {
+            databases_link: ResourceLink::root(ResourceType::Databases),
+            pipeline: CosmosPipeline::new(
+                connection_str.account_endpoint.parse()?,
+                AuthorizationPolicy::from_shared_key(key),
+                options.client_options,
+            ),
+        })
+    }
+
     /// Gets a [`DatabaseClient`] that can be used to access the database with the specified ID.
     ///
     /// # Arguments
--- a/sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
@@ -73,3 +73,52 @@ impl Debug for ConnectionString {
         f.write_str(&debug_string)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::ConnectionString;
+    use azure_core::credentials::Secret;
+
+    #[test]
+    pub fn test_regular_connection_string() {
+        let connection_string =
+            "AccountEndpoint=https://accountname.documents.azure.com:443/;AccountKey=accountkey";
+        let secret = Secret::new(connection_string);
+        let connection_str = ConnectionString::from_secret(&secret).unwrap();
+
+        assert_eq!(
+            "https://accountname.documents.azure.com:443/",
+            connection_str.account_endpoint
+        );
+
+        assert_eq!("accountkey", connection_str.account_key.secret());
+    }
+
+    #[test]
+    pub fn test_empty_connection_string() {
+        test_bad_connection_string("")
+    }
+
+    #[test]
+    pub fn test_malformed_connection_string() {
+        test_bad_connection_string(
+            "AccountEndpointhttps://accountname.documents.azure.com:443/AccountKey=accountkey",
+        );
+    }
+
+    #[test]
+    pub fn test_connection_string_missing_account_endpoint() {
+        test_bad_connection_string("AccountKey=accountkey");
+    }
+
+    #[test]
+    pub fn test_connection_string_missing_account_key() {
+        test_bad_connection_string("AccountEndpoint=https://accountname.documents.azure.com:443/;");
+    }
+
+    fn test_bad_connection_string(connection_string: &str) {
+        let secret = Secret::new(connection_string.to_owned());
+        let connection_str = ConnectionString::from_secret(&secret);
+        assert!(connection_str.is_err());
+    }
+}
--- a/sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
@@ -1,0 +1,75 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+use std::borrow::Cow;
+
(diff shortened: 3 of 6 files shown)


sdk/core/azure_core/src/credentials.rs
```diff
@@ -29,6 +30,46 @@ impl Secret {
...
+///
+/// The [`Debug`] implementation will not print the secret.
+#[derive(Clone, Deserialize, Serialize)]
+pub struct CosmosConnectionString {
+    pub account_endpoint: Cow<'static, str>,
+    pub account_key: Cow<'static, str>,
+}
+
+impl CosmosConnectionString {
+    pub fn from_secret(secret: &Secret) -> crate::Result<Self> {
```
heaths: Why not just implement `From<Secret>`?

sdk/core/azure_core/src/credentials.rs
```diff
@@ -29,6 +30,46 @@ impl Secret {
...
+            2 => {
+                if (collection[0].contains("AccountEndpoint=")
+                    && collection[1].contains("AccountKey="))
+                {
+                    Ok(Self {
+                        account_endpoint: collection[0][16..collection[0].len() - 1]
+                            .to_owned()
+                            .into(),
+                        account_key: collection[0][11..collection[0].len() - 1].to_owned().into(),
+                    })
```
heaths: This parsing assumes too much. You should split key=value pairs and find the ones you need.

sdk/core/azure_core/src/credentials.rs
```diff
@@ -29,6 +30,46 @@ impl Secret {
     }
 }
 
+/// Represents a Cosmos DB connection string.
+///
+/// The [`Debug`] implementation will not print the secret.
+#[derive(Clone, Deserialize, Serialize)]
+pub struct CosmosConnectionString {
+    pub account_endpoint: Cow<'static, str>,
```
heaths: Is there any time when this would be a static string by the dev? Seems unlikely.

sdk/core/azure_core/src/credentials.rs
```diff
@@ -65,6 +106,17 @@ impl Debug for Secret {
     }
 }
 
+impl Debug for CosmosConnectionString {
```
heaths: Just derive `SafeDebug` on the type. You also shouldn't format for `Debug`. That's more for `Display`, which isn't needed here.

sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
```diff
@@ -0,0 +1,124 @@
...
+
+use azure_core::{credentials::Secret, http::Model, Error};
+use serde::Deserialize;
+use std::fmt::Debug;
+
+/// Represents a Cosmos DB connection string.
+///
+/// The [`Debug`] implementation will not print the account key.
+#[derive(Model, Clone, Deserialize, PartialEq, Eq)]
+pub struct ConnectionString {
```
heaths, line 14: Why is there a near-duplicate in `models`? Why is this a model even? We typically think of models as request/response models and this is more of a client helper.

sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
```diff
@@ -0,0 +1,124 @@
...
+            account_key: key,
+        })
+    }
+
+    pub fn from_secret(secret: &Secret) -> azure_core::Result<Self> {
+        ConnectionString::from_connection_string(secret.secret())
+    }
+}
+
+impl Debug for ConnectionString {
```
heaths, line 66: Same comments about `From` and `SafeDebug` here.

heaths (changes requested):
Apart from some feedback, we also need to discuss as a team whether we want to even offer this. It's fortunate that `key_auth` isn't enabled by default, but we also don't want to condone less-secure authentication. I appreciate some key auth is required for many scenarios, but do we *also* need connection strings? /cc @ronniegeraghty @JeffreyRichter @RickWinter

analogrelay:
Connection String auth _is_ part of key-based auth and it's definitely something we expect to have, though behind the key_auth feature

sdk/core/azure_core/src/credentials.rs
analogrelay: I think this was supposed to be removed and just left here by accident

nehrao1: Outdated, discussed offline - this file was added by mistake from a previous iteration. Cleaned it up

sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
nehrao1: Outdated, discussed offline - this file was added by mistake from a previous iteration. Cleaned it up

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -96,6 +96,43 @@ impl CosmosClient {
...
+        let key = connection_str.account_key;
+
+        Ok(Self {
+            databases_link: ResourceLink::root(ResourceType::Databases),
+            pipeline: CosmosPipeline::new(
+                connection_str.account_endpoint.parse()?,
+                AuthorizationPolicy::from_shared_key(key),
+                options.client_options,
+            ),
+        })
```
analogrelay, line 133: After parsing the Connection String, we can just call `Self::with_key` and pass the endpoint and key in there, to avoid duplicating the creation logic here.

sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        Ok(Self {
+            account_endpoint: Cow::Owned(endpoint),
+            account_key: key,
+        })
+    }
+
+    pub fn from_secret(secret: &Secret) -> azure_core::Result<Self> {
+        ConnectionString::from_connection_string(secret.secret())
+    }
```
analogrelay, line 62: As @heaths raised earlier, I do think it's better here to implement `TryFrom<Secret>` rather than having our own `from_secret` method. In Rust, the `From`/`TryFrom` and `Into`/`TryInto` traits are used whenever you want to create a conversion from one type to another (the `Try` variants being used when the conversion may fail, and the others being used when the conversion cannot fail). That enables a number of nice patterns in the language that I can fill you in on offline if you're interested.

So, instead of this method, add a separate impl block:

```rust
impl TryFrom<Secret> for ConnectionString {
    type Error = azure_core::Error;
    fn try_from(value: T) -> Result<Self, Self::Error> {
        // Put the implementation here.
    }
}
```

I also think we could implement `TryFrom<&str>`, which would contain the code currently in `from_connection_string`.

You

sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
```diff
@@ -0,0 +1,112 @@
...
+
+use azure_core::{credentials::Secret, fmt::SafeDebug, Error};
+use serde::Deserialize;
+
+/// Represents a Cosmos DB connection string.
+///
+/// The [`Debug`] implementation will not print the account key.
+#[derive(Clone, Deserialize, PartialEq, Eq, SafeDebug)]
+pub struct ConnectionString {
+    pub account_endpoint: Cow<'static, str>,
```
analogrelay, line 14: I don't think this should be a `Cow`. We use `Cow`s (Short for "copy on write") like this as a way to avoid copying a static string literal until we need to modify it, and it's useful in many places, but I think it's challenging to use here and not really worth it. To make it useful, we'd need the `from_connection_string` method to take a `Cow<'static, str>` as well, and use much more complicated logic to try and preserve the static string, if one is provided.

In your current `from_connection_string` implementation, you only ever set this to `Cow::Owned`, so there's really no value in the `Cow`, since you never use `Cow::Borrowed` (the alternative form).

Since ConnectionStrings are created generally once per application, I'd prefer to just have this be a standard owned `String`.

Happy to chat offline about this more!

sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    #[test]
+    pub fn test_connection_string_missing_account_key() {
+        test_bad_connection_string("AccountEndpoint=https://accountname.documents.azure.com:443/;");
+    }
+
+    fn test_bad_connection_string(connection_string: &str) {
+        let secret = Secret::new(connection_string.to_owned());
+        let connection_str = ConnectionString::from_secret(&secret);
+        assert!(connection_str.is_err());
```
analogrelay, line 110: It'd be nice to test the error details themselves here. You should be able to unwrap the `azure_core::Error` by calling `unwrap_err()` here and then assert properties on it. One small gotcha: The error message for an error is usually accessed by formatting the error using the `Display` trait, i.e. `let message = format!("{}", err)` would extract the error message.

sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
```diff
@@ -0,0 +1,112 @@
...
+
+use std::borrow::Cow;
+
+use azure_core::{credentials::Secret, fmt::SafeDebug, Error};
+use serde::Deserialize;
+
+/// Represents a Cosmos DB connection string.
+///
+/// The [`Debug`] implementation will not print the account key.
+#[derive(Clone, Deserialize, PartialEq, Eq, SafeDebug)]
```
analogrelay, line 12: I don't think this needs to be `Deserialize`. It's not something we need to deserialize from JSON.

```suggestion
#[derive(Clone, PartialEq, Eq, SafeDebug)]
```

sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
analogrelay, line 1: @heaths noted earlier that this isn't really a model, and I think that's a good point. I did initially suggest putting this here, but I think it can just live in the root `src` directory for the crate.

analogrelay (changes requested):
Looks good! A few suggestions to improve it a bit, but nothing major.

sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
```diff
@@ -0,0 +1,112 @@
...
+            }
+
+            let (key, value) = part.split_once('=').ok_or(Error::new(
+                azure_core::error::ErrorKind::Other,
+                "invalid connection string",
+            ))?;
+            match key {
+                "AccountEndpoint" => account_endpoint = Some(value.to_string()),
+                "AccountKey" => account_key = Some(Secret::new(value.to_string())),
+                _ => {}
```
Pilchie, line 36: For .NET we added a connection string option to disable SSL cert validation to make using the emulator easier without needing a code change between emulator and prod workloads.  We should consider supporting that here too.

analogrelay: I think that's worth considering but I'll file a new issue for that. We don't currently have an explicit option to configure this right now, so it's a little tricky to add. It can be configured through the `azure_core` pipeline configuration, but figuring out how to do that in a way that composes with the user's other settings is a little tricky since it requires "replacing" the HTTP client itself.

sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
nehrao1: Removed

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
nehrao1: Updated

sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
nehrao1: Updated

sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
nehrao1: Updated

sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
nehrao1: Updated

sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
nehrao1: Updated

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -3,7 +3,7 @@
 
 use crate::{
     clients::DatabaseClient,
-    models::DatabaseProperties,
+    models::{ConnectionString, DatabaseProperties},
```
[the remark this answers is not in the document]
nehrao1: Was able to move past that build error with this. Thanks!

sdk/cosmos/azure_data_cosmos/src/models/connection_string.rs
nehrao1: [#2649](https://github.com/Azure/azure-sdk-for-rust/issues/2649) - thanks Ashley!

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -96,6 +96,37 @@ impl CosmosClient {
         })
     }
 
+    /// Creates a new CosmosClient, using a connection string.
+    ///
+    /// # Arguments
+    ///
+    /// * `connection_string` - the connection string to use for the client, e.g. AccountEndpoint=https://accountname.documents.azure.com:443/‌​;AccountKey=accountk‌​ey
```
analogrelay: We should wrap the connection string in backticks so it renders in monospace font.

```suggestion
    /// * `connection_string` - the connection string to use for the client, e.g. `AccountEndpoint=https://accountname.documents.azure.com:443/‌​;AccountKey=accountk‌​ey`
```

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -96,6 +96,37 @@ impl CosmosClient {
...
+    /// use azure_data_cosmos::CosmosClient;
+    /// use azure_core::credentials::Secret;
+    ///
+    /// let client = CosmosClient::with_connection_string(
+    ///     "AccountEndpoint=https://accountname.documents.azure.com:443/‌​;AccountKey=accountk‌​ey",
+    ///     None)
+    ///     .unwrap();
+    /// ```
+    #[cfg(feature = "key_auth")]
+    #[allow(rustdoc::bare_urls)]
```
analogrelay: I _think_ if the connection string is in backticks, this lint won't trigger anymore. The reason it's triggering is that it's seeing a plain URL, which rustdoc will NOT turn in to a link by default (so the lint is warning you that you really should make it a link if it's a URL someone might navigate to). However, in our case, it's part of an example Connection String. Wrapping it in backticks (`like this`) will tell rustdoc that it's a verbatim string that shouldn't be parsed as markdown and avoid the lint entirely.

sdk/cosmos/azure_data_cosmos/src/connection_string.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::ConnectionString;
+    use azure_core::credentials::Secret;
+
+    #[test]
+    #[allow(rustdoc::bare_urls)]
```
analogrelay: I don't _think_ you need this lint on this test, but maybe I'm wrong. There are no rustdoc comments on the test, and even if there were, they wouldn't be used.

```suggestion
```

analogrelay (commented):
I think this is basically good! A few small fix-ups to docs but otherwise ready to roll!

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
nehrao1: Looks like this worked, thanks!

sdk/cosmos/azure_data_cosmos/src/connection_string.rs
nehrao1: Worked, thanks

sdk/cosmos/azure_data_cosmos/src/connection_string.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+impl TryFrom<&Secret> for ConnectionString {
+    type Error = azure_core::Error;
+    fn try_from(secret: &Secret) -> Result<Self, Self::Error> {
+        secret.secret().try_into()
+    }
+}
+
+impl TryFrom<&str> for ConnectionString {
```
heaths: You should also implement `FromStr` - in fact, maybe in lieu of `TryFrom<&str>` - because it also adds the `parse()` function and is otherwise equivalent to `TryFrom<&str>`.

I won't block the PR for this one, though, but I do highly recommend a follow-up to either add or replace it with `FromStr`.

sdk/cosmos/azure_data_cosmos/src/connection_string.rs
```diff
@@ -0,0 +1,135 @@
...
+                continue;
+            }
+
+            let (key, value) = part.split_once('=').ok_or(Error::new(
+                azure_core::error::ErrorKind::Other,
+                "invalid connection string",
+            ))?;
+            match key {
+                "AccountEndpoint" => account_endpoint = Some(value.to_string()),
+                "AccountKey" => account_key = Some(Secret::new(value.to_string())),
```
heaths: Are keys case-sensitive? I suspect not, in which case I recommend using `str::eq_ignore_ascii_case` which is more efficient than just lowering the case to use a `match`.

sdk/cosmos/azure_data_cosmos/src/connection_string.rs
```diff
@@ -0,0 +1,135 @@
...
+            return Err(Error::new(
+                azure_core::error::ErrorKind::Other,
+                "invalid connection string, missing 'AccountEndpoint",
+            ));
+        };
+
+        let Some(key) = account_key else {
+            return Err(Error::new(
+                azure_core::error::ErrorKind::Other,
+                "invalid connection string, missing 'AccountKey",
```
heaths: You're missing a closing single quote around the key in both cases.

heaths (approved):
Nothing blocking, but I'd highly recommend either updating this PR or submitting a new one for comments.

sdk/cosmos/azure_data_cosmos/src/connection_string.rs
analogrelay: That's a great point, I had forgotten about `FromStr`.

sdk/cosmos/azure_data_cosmos/src/connection_string.rs
analogrelay: Yeah, I don't think they need to be case-sensitive.

sdk/cosmos/azure_data_cosmos/src/connection_string.rs
nehrao1: fixed

sdk/cosmos/azure_data_cosmos/src/connection_string.rs
nehrao1: Updated

sdk/cosmos/azure_data_cosmos/src/connection_string.rs
nehrao1: Updated

analogrelay (approved):

PR #2655, opened 2025-06-02
Flatten azure_identity module structure

chlowell:
Moving `src/credentials/*` up a level, centralizing module declarations and public API in `lib.rs`


sdk/identity/azure_identity/src/app_service_managed_identity_credential.rs
```diff
@@ -14,7 +14,7 @@ const SECRET_HEADER: HeaderName = HeaderName::from_static("x-identity-header");
 const SECRET_ENV: &str = "IDENTITY_HEADER";
 
 #[derive(Debug)]
-pub struct AppServiceManagedIdentityCredential {
+pub(crate) struct AppServiceManagedIdentityCredential {
```
heaths, line 17: Tip: why `pub(crate)`? Just to be extra safe? As long as it doesn't get exposed publicly from `src/lib.rs` it can be just `pub`.

sdk/identity/azure_identity/src/lib.rs
```diff
@@ -4,24 +4,48 @@
...
+pub use client_secret_credential::*;
+pub use default_azure_credential::*;
+pub use managed_identity_credential::*;
+pub use options::TokenCredentialOptions;
+pub use workload_identity_credential::*;
+
+pub(crate) use app_service_managed_identity_credential::*;
+pub(crate) use cache::TokenCache;
+pub(crate) use imds_managed_identity_credential::*;
+pub(crate) use virtual_machine_managed_identity_credential::*;
```
heaths, line 42: Tip: anything in scope is visible to children, so you could elide `pub(crate) ` on these lines.

heaths (approved):
I see nothing blocking.

sdk/identity/azure_identity/src/app_service_managed_identity_credential.rs
chlowell: Yes, I'm just underscoring that this is a deliberately private type

sdk/identity/azure_identity/src/lib.rs
chlowell: I suppose so. Is this bad style? I meant to make it totally obvious that none of this stuff is for public consumption, but that desire is rooted in my not having internalized the syntax and I can see this approach might be excessive

heaths: It's not wrong, just unnecessary hence the "tip" and signing off on the PR. Do with it what you will. This is an opportunity to educate as well.

PR #2657, opened 2025-06-02
Implement Recoverable connection for AMQP operations.

LarryOsterman:
This PR does three different things:

1) It renames the `ConnectionManager` type to `RecoverableConnection` to better reflect what is going on.
2) It splits out the authorization functionality from `ConnectionManager` into a new struct named `Authorizer` to simplify the logic in `connection_manager.rs`
3) It consolidates all the AMQP active protocol interactions to `RecoverableConnection`. This was done to (a) simplify the AMQP interactions for each of the specific protocol elements and (b) lay the foundation for connection, session, and link recovery. See the rather large comment on the `RecoverableConnection` for more details on how this all works and fits together.

There are a few bonus changes in this PR as well:
1) Adds a `send_ref` function to the `AmqpSender` struct which allows sending a reference to a message (saving a message clone).
1) `AmqpOrderedMap::iter()` passes references to the inner elements to be consistent with the standard collection `iter()` method.


sdk/core/azure_core_amqp/src/messaging.rs
```diff
@@ -94,6 +94,7 @@ impl From<&AmqpSymbol> for DistributionMode {
     }
 }
 
+#[derive(Debug)]
```
heaths, line 97: Not `SafeDebug`, or is `Debug` okay here?

sdk/eventhubs/azure_messaging_eventhubs/src/common/authorizer.rs
heaths, line 1: You got rid of a lot of tests herein. Still good enough coverage?

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable_connection.rs
```diff
@@ -0,0 +1,1055 @@
...
+/// The way a client uses a RecoverableConnection is as follows:
+///   1. Create a new instance of the RecoverableConnection.
+///   2. Retrieve an interim object from the RecoverableConnection. Supported
+///      interim objects are:
+///
+///    - `AmqpManagement`: Used for management operations.
+///    - `AmqpSender`: Used for sending messages to the Event Hubs service.
+///    - `AmqpReceiver`: Used for receiving messages from the Event Hubs service.
+///    - `AmqpClaimsBasedSecurity`: Used for authorization operations (should not be used directly)
+/// * 3. Use the interim object to perform operations on the Event Hubs service.
```
heaths: Errant bullet.
```suggestion
///   3. Use the interim object to perform operations on the Event Hubs service.
```

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable_connection.rs
```diff
@@ -0,0 +1,1055 @@
...
+///
+/// The various interim objects implement the appropriate AMQP APIs, but wrap the underlying APIs with
+/// a retry loop [`retry_azure_operation`], so that the actual client does not have to worry about retrying or recovering operations.
+///
+/// There is a taxonomy of methods in this struct:
+///   - `ensure_*` methods: These methods are used to ensure that the underlying connection, session, management client, cbs client, sender, or receiver is created and available.
+///   - `get_*` methods: These methods are used to retrieve a wrapper around the underlying session, management client, cbs client, sender, or receiver.
+///   - `create_*` methods: These methods are used to create a new underlying connection, session, management client, cbs client, sender, or receiver.
+///
+/// In general, the `ensure_*` and `create_*` methods are private to the ` RecoverableConnection`
```
heaths:
```suggestion
/// In general, the `ensure_*` and `create_*` methods are private to the `RecoverableConnection`
```

Also, you don't always mark code constructs with backticks above. Should be consistent.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable_connection.rs
```diff
@@ -0,0 +1,1055 @@
...
+            .ok_or_else(|| EventHubsError::from(ErrorKind::MissingMessageSender))?
+            .clone())
+    }
+
+    fn should_retry_management_response(e: &azure_core::Error) -> bool {
+        match e.kind() {
+            AzureErrorKind::Amqp => {
+                warn!("Amqp operation failed: {}", e.source().unwrap());
+                if let Some(e) = e.source() {
+                    debug!("Error: {}", e);
```
heaths: Here and elsewhere below you're "flattening" the error to just a string, which often isn't that helpful. Sure you don't *also* want to attach it using its `Debug` implementation e.g.,

```rust
debug!(err = ?e, "Error: {e}");
```

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable_connection.rs
```diff
@@ -0,0 +1,1055 @@
...
+}
+
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+impl AmqpClaimsBasedSecurityApis for AmqpClaimsBasedSecurityClient {
+    async fn authorize_path(
+        &self,
+        path: String,
+        token_type: Option<String>,
+        secret: String,
```
heaths: Should this be a `Secret` instead?

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable_connection.rs
```diff
@@ -0,0 +1,1055 @@
...
+            Some(RecoverableConnection::should_retry_send_operation),
+        )
+        .await?;
+        Ok(outcome)
+    }
+
+    /// Sends a message reference to the Event Hubs service.
+    ///
+    /// Note: We do not implement this method because none of the callers of AmqpSenderClient call send_ref.
+    async fn send_ref<M>(
```
heaths, line 702: Maybe also `#[doc(hidden)]` as well then?

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable_connection.rs
```diff
@@ -0,0 +1,1055 @@
...
+        )
+        .await?;
+        Ok(delivery)
+    }
+
+    async fn accept_delivery(
+        &self,
+        _delivery: &azure_core_amqp::AmqpDelivery,
+    ) -> azure_core::Result<()> {
+        unimplemented!("AmqpReceiverClient does not support accept_delivery operation");
```
heaths, line 806: With the number of unimplemented trait functions I see in this PR, sure maybe splitting traits up wouldn't be better? Generally, "if it compiles it works" and then pretty much flies in the face of that as these are all runtime panics. Almost like how serde splits up ser and de.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable_connection.rs
```diff
@@ -0,0 +1,1055 @@
...
+    use crate::{consumer, ErrorKind, EventHubsError};
+    use azure_core::credentials::TokenRequestOptions;
+    use azure_core::{credentials::AccessToken, http::Url, Result};
+    use azure_core_amqp::AmqpError;
+    use std::sync::{Arc, Mutex as SyncMutex};
+    use time::{Duration, OffsetDateTime};
+
+    // Helper struct to mock token credential
+    #[derive(Debug)]
+    struct MockTokenCredential {
```
heaths: We have one of these in `azure_core_test` that might be of use and avoid duplication. Maybe. Can't remember off hand if it would do exactly what you want, but worth checking out.

sdk/eventhubs/azure_messaging_eventhubs/src/consumer/event_receiver.rs
```diff
@@ -172,35 +119,24 @@ impl EventReceiver {
...
-                        _ = azure_core::sleep::sleep(delivery_timeout).fuse() => {
-                             Err(azure_core::Error::new(
-                                AzureErrorKind::Io,
-                                Box::new(std::io::Error::from(std::io::ErrorKind::TimedOut))))
-                        },
-                    }?
-                 } else {
-                     self.receiver.receive_delivery().await
-                 }}, &self.retry_options, Some(Self::should_retry_receive_operation)).await?;
+                let receiver = self.connection.get_receiver(&self.source_url, self.message_source.clone(), self.receiver_options.clone(), self.timeout).await?;
```
heaths: Surprise/nit: the formatter didn't wrap this long line? Just wondering more so if everything was formatted since some check might care.

sdk/core/azure_core_amqp/src/messaging.rs
LarryOsterman: It's a wrapper around AmqpDelivery, which implements SafeDebug.

sdk/eventhubs/azure_messaging_eventhubs/src/common/authorizer.rs
LarryOsterman: The tests should all be there - I copied them from recoverable_connection.rs and then removed the ones that weren't relevant.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable_connection.rs
LarryOsterman: I'd not run into that construct before, nice.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable_connection.rs
LarryOsterman: Yes it should, fixing all over the place.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable_connection.rs
LarryOsterman: Possibly. But these are special case structures which exist only to provide recovery semantics for the specific implementations in EventHubs. It would mean splitting up the AMQP trait implementations solely because of EventHubs implementation details, and I'm really not sure that's justified.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable_connection.rs
LarryOsterman: I looked into re-using the mock token credential in azure_core_test when I first implemented connection re-authorization and it would require some significant surgery for use here (I need to be able to strictly implement token expiration here and the mock token credential in azure_core_test doesn't)..

However that's only needed in authorization so I can use the mock credential in recoverable_connection's tests.

sdk/eventhubs/azure_messaging_eventhubs/src/consumer/event_receiver.rs
LarryOsterman: The rust formatter doesn't work consistently inside a macro (`try_stream!`).

No idea why but it doesn't - it's kinda annoying, tbh.

Fixing manually.

heaths: > Fixing manually.

In the future, no need. I don't mind the formatting myself, but *when* it does apply automatically, keeping it consistent across clones is nice. But, yeah, if it didn't format here and that's consistent across machines (which I expect), I actually didn't mind the long line. It was just a red (or, rather, yellow) flag - a possible symptom of something else wrong.

sdk/core/azure_core_amqp/src/cbs.rs
```diff
@@ -51,7 +51,7 @@ pub trait AmqpClaimsBasedSecurityApis {
         &self,
         path: String,
         token_type: Option<String>,
-        secret: String,
+        secret: &Secret,
```
heaths, line 54: Why did you go from passing by value to passing by ref? Or was that unrelated?

sdk/eventhubs/azure_messaging_eventhubs/src/consumer/event_receiver.rs
LarryOsterman: 100% agreed. But it's annoying to me, especially on my home machine (vertical monitor) so I manually updated it.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable_connection.rs
```diff
@@ -822,58 +825,10 @@ impl AmqpReceiverApis for AmqpReceiverClient {
...
-                let mut token = self.token.lock().unwrap();
-                *token = AccessToken::new(
-                    azure_core::credentials::Secret::new("mock_token"),
-                    expires_on,
-                );
-                Ok(token.clone())
-            }
-        }
-    }
+    use azure_core_test::credentials::MockCredential;
```
heaths, line 830: Just to note: if this didn't make sense to use you didn't have to. Just letting you know.

heaths (approved):

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable_connection.rs
LarryOsterman: Actually it *did* make sense to me. The only reason I didn't use MockCredential before was that I had a need for a strict token timeout implementation (which MockCredential doesn't have). 

But the `RecoverableConnection` tests don't have the same requirements, so it makes sense to use the existing implementation rather than having *three* separate mock credential implementations.

sdk/core/azure_core_amqp/src/cbs.rs
LarryOsterman: Because my `perform_authorization` takes an `&AccessToken` which has a `Secret` field.

I could either change the `authorize_path` function to take the secret by ref everywhere, *or* I could clone it in the refresh case.

I didn't see any value in cloning the secret value because ultimately the credential implementation going to call `Secret::secret()` which just returns the inner string slice, and thus the outer clone would be unnecessary.

PR #2659, opened 2025-06-02
Iterate over items asynchronously from Pager

### LINKED ISSUE #2205: Pager should, by default, return items across all pages
Most users likely want all items when trying to enumerate them. As discussed internally today, we'll make changes to `Pager<T>` to support item-by-item iteration while supporting page-by-page iteration for scenarios that require it.

Thus,

* `Pager<T>::iter(&self) -> impl Iterator<Item = T>`
* `Pager<T>::into_iter(self) -> impl Iterator<Item = T>`
* `Pager<T>::pages(&self) -> impl Iterator<Item = Page<T>>`
* `Pager<T>::into_pages(self) -> impl Iterator<Item = Page<T>>`

Assuming we soon land Azure/typespec-rust#250, that means this code:

```rust
let mut pager = client.get_secrets(None)?.into_stream();
let mut secrets: Vec<SecretItem> = Vec::new();
while let Some(secrets) = pager.try_next().await? {
    secrets.extend(secrets.into_body()?.into_iter());
}
```

Turns into something like this code:

```rust
let secrets: Vec<SecretItem> = client.list_secrets(None).try_collect().await?;
```

Related: #1909

heaths:
For functions that return a `Pager<T>`, callers will iterate over items across all pages asynchronously by default but can get an async iterator (`Stream`) over pages as well.

Resolves #2205


sdk/keyvault/azure_security_keyvault_secrets/src/generated/clients/secret_client.rs
```diff
@@ -251,11 +251,11 @@ impl SecretClient {
                 let res: ListDeletedSecretPropertiesResult = json::from_json(&bytes)?;
                 let rsp = Response::from_bytes(status, headers, bytes);
                 Ok(match res.next_link {
-                    Some(next_link) => PagerResult::Continue {
```
heaths, line 254: @jhendrixMSFT these renames are all that changed for generated clients.

sdk/keyvault/azure_security_keyvault_secrets/src/generated/models/pub_models.rs
```diff
@@ -140,6 +141,16 @@ pub struct ListDeletedSecretPropertiesResult {
...
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+impl Page for ListDeletedSecretPropertiesResult {
+    type Item = DeletedSecretProperties;
+    type IntoIter = <Vec<DeletedSecretProperties> as IntoIterator>::IntoIter;
+    async fn into_items(self) -> Result<Self::IntoIter> {
+        Ok(self.value.into_iter())
+    }
+}
+
```
heaths, line 153: @jhendrixMSFT pageable models need to implement this trait as shown. If no item array can be detected or if there are multiple, we should instead return a `PageIterator<Response<T>>` but that can probably wait because, in practice, that *shouldn't* happen and I believe isn't applicable to the crates we are emitting currently. /cc @JeffreyRichter

heaths (commented):
@jhendrixMSFT as you can see, the changes needed in the emitter are very small.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -2,140 +2,412 @@
...
 
+#[cfg(target_arch = "wasm32")]
+type BoxedStream<P> = Box<dyn Stream<Item = Result<P, Error>>>;
+
+/// Iterates over a collection of items or individual pages of items from a service.
+///
+/// You can asynchronously iterate over items returned by a collection request to a service,
+/// or asynchronously fetch pages of items if preferred.
+#[pin_project::pin_project]
+pub struct ItemIterator<P: Page> {
```
analogrelay, line 88: I wonder if this should be `PageItemIterator` or something that specifies that it's an iterator _over items that are retrieved in pages_ as opposed to just an iterator over items? I don't feel super strongly about that though, `ItemIterator` is OK with me if that feels too verbose. The fact that `ItemIterator` takes a `Page` as a generic parameter also helps clue the user in that this is for pagination.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -145,125 +417,147 @@ impl<T> futures::Stream for PageStream<T> {
...
     Init,
-    Continuation(T),
+    More(T),
     Done,
 }
 
-#[cfg(test)]
-mod tests {
-    use std::collections::HashMap;
+fn iter_from_callback<
```
analogrelay, line 433: non-blocking nit: This could return `impl Stream` and you could use `ItemIterator::from_stream` to reduce duplication a bit here. No functional difference there though, so I'm fine either way.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -145,125 +417,147 @@ impl<T> futures::Stream for PageStream<T> {
...
-        assert_eq!(
-            &[
-                ("page-1".to_string(), Page { page: 1 }),
-                ("page-2".to_string(), Page { page: 2 }),
-                ("page-3".to_string(), Page { page: 3 }),
-            ],
-            pages.as_slice()
-        )
+        let pages: Vec<i32> = pager.try_collect().await.unwrap();
+        assert_eq!(vec![1, 2, 3], pages.as_slice())
```
analogrelay: nit: this isn't pages any more

```suggestion
        let items: Vec<i32> = pager.try_collect().await.unwrap();
        assert_eq!(vec![1, 2, 3], items)
```

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### Breaking Changes
 
+* `FeedPager<T>` now asynchronously iterates items of type `T` instead of pages containing items of type `T`. Call `Pager::into_pages()` to get a `PageIterator` to asynchronously iterate over all pages.
```
analogrelay: The type name is wrong now that `Pager` is gone, but I think we can just remove the type name here anyway.

```suggestion
* `FeedPager<T>` now asynchronously iterates items of type `T` instead of pages containing items of type `T`. Call `into_pages()` to get a `PageIterator` to asynchronously iterate over all pages.
```

sdk/core/azure_core/CHANGELOG.md
```diff
@@ -5,10 +5,16 @@
 ### Features Added
 
 - Added `#[safe]` attribute helper for `SafeDebug` derive macro to show or hide types and members as appropriate.
+- Added `Page` trait to facilitate the `ItemIterator`.
+- Added `PageIterator` to asynchronously iterator all pages.
 
 ### Breaking Changes
 
-- Removed `AccessToken::is_expired()`
+- A `Pager` now asynchronously iterates over items across all pages. Call `Pager::into_pages()` to get a `PageIterator` to asynchronously iterate over all pages.
```
analogrelay, line 13: `Pager` is gone right? Renamed to `ItemIterator`/`PageIterator`?

sdk/keyvault/azure_security_keyvault_certificates/CHANGELOG.md
```diff
@@ -6,6 +6,12 @@
 
 ### Breaking Changes
 
+- `Pager<T>` now asynchronously iterates items of type `T` instead of pages containing items of type `T`. Call `Pager::into_pages()` to get a `PageIterator` to asynchronously iterate over all pages. This affects:
```
analogrelay, line 9: Another place where the now-dead `Pager` is referenced ;)

sdk/keyvault/azure_security_keyvault_keys/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 ### Breaking Changes
 
+- `Pager<T>` now asynchronously iterates items of type `T` instead of pages containing items of type `T`. Call `Pager::into_pages()` to get a `PageIterator` to asynchronously iterate over all pages. This affects:
```
analogrelay, line 9: And another reference to `Pager`

analogrelay (approved):
Looks good! A few non-blocking nits and I noticed a few places where `Pager` is still referenced (totally possible you were planning to do a pass through but I noticed them and figured I'd point them out)

sdk/core/azure_core/src/http/pager.rs
heaths: Fair. For now, though, I think I prefer the conciseness. If we get feedback indicating confusion, though, I'm certainly not married to the name.

sdk/core/azure_core/CHANGELOG.md
heaths: No, `Pager<P>` is still a typedef for `ItemIterator<Response<P>>`.

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
heaths: Given my reply to your comment below - that `Pager` still exists - isn't the original text still valid?

LarryOsterman (approved):

PR #2661, opened 2025-06-03
[Storage] `get_account_info` for all clients, `set/get_tags` for BlobClient

vincenttran-msft:
This PR introduces the following changes:

- Regenerates against `main` branch in [feature/blob-tsp](https://github.com/Azure/azure-rest-api-specs/tree/feature/blob-tsp)
- Initial implementation of `get_account_info` for `BlobClient`, `ContainerClient`, and `BlobServiceClient`
- Initial implementation of `set_tags` and `get_tags` for `BlobClient`

Accompanying TypeSpec changes (merge before merging this):
https://github.com/Azure/azure-rest-api-specs/pull/35044


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -4,16 +4,16 @@
 use crate::{
     generated::clients::BlobClient as GeneratedBlobClient,
     generated::models::{
-        BlobClientDownloadResult, BlobClientGetPropertiesResult,
+        BlobClientDownloadResult, BlobClientGetAccountInfoResult, BlobClientGetPropertiesResult,
```
heaths: It's hard to reason about all the public API changes. Have you generated docs via `cargo doc --open -p azure_storage_blob` and made sure that looks like how you expect? Clients in the top-level module should always be a subset of those in the `clients` submodule, mind you. I remember a while back that still wasn't the case.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -276,3 +279,69 @@ async fn test_set_access_tier(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
+#[recorded::test]
+async fn test_blob_tags(ctx: TestContext) -> Result<(), Box<dyn Error>> {
+    // Recording Setup
+    let recording = ctx.recording();
+    let container_client = get_container_client(recording, true).await?;
+    let blob_client = container_client.blob_client(get_blob_name(recording));
+    create_test_blob(&blob_client).await?;
+
+    // Set Tags with Tags Specified
+    let blob_tag_1 = BlobTag {
```
heaths: Does this match what other languages do? It would be more idiomatic to have this as a `HashMap<String, String>`.

sdk/storage/azure_storage_blob_test/src/lib.rs
```diff
@@ -112,3 +114,28 @@ pub async fn create_test_blob(
...
+        count_map.insert(blob_tag.key.unwrap(), blob_tag.value.unwrap());
+    }
+    // Iterate through second set of tags
+    for blob_tag in tags2.blob_tag_set.unwrap() {
+        // If tag is not found, return false
+        if !count_map.contains_key(&blob_tag.key.clone().unwrap()) {
+            return false;
+        } else {
+            count_map.remove(&blob_tag.key.unwrap());
+        }
```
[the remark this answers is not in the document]
heaths: It's in a test, so warranted.

heaths (approved):
I see nothing blocking.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: In [Python ](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/tests/test_blob_tags.py#L102)we get something like this:
```python
blob_tags = {"tag1": "firsttag", "tag2": "secondtag", "tag3": "thirdtag"}
```

While [.NET](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs/tests/BlobClientTests.cs#L404) yields something like this:
```c#
 IDictionary<string, string> tags = BuildTags();
            BlobUploadOptions options = new BlobUploadOptions
            {
                Tags = tags
            };

// Unraveling BuildTags()
public Dictionary<string, string> BuildTags()
            => new Dictionary<string, string>
            {
                { "tagKey0", "tagValue0" },
                { "tagKey1", "tagValue1" }
            };
```
So I could definitely see the case that we make this more user-friendly such that we take in a `HashMap` and build the `BlobTags` object for them. cc: @jalauzon-msft Thoughts?

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: Because I was directly importing these from `generated::models::` and not from the re-exported `models` module, it made me question whether we should be exporting the `XXXResult` or not.

However the one case where this broke down was that my helper test code in `azure_storage_blob_test` crate had a function that returned one of these result types:
```rust
pub async fn create_test_blob(
    blob_client: &BlobClient,
) -> Result<Response<BlockBlobClientUploadResult>> {
..
..
```

Thus causing me to have to re-import this from the generated models to make it public. So I think I will revert this change and export all `XXXResult` in the `models` module.

sdk/storage/azure_storage_blob/tests/blob_client.rs
jalauzon-msft: I think both get and set tags should just work with HashMaps. I don't see a need for this extra type.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
heaths: No, sorry. I didn't mean to imply importing it was wrong. I meant that with all the types you're re-exporting all over, and because beta.2 wasn't exposing the right public API (#2523), when you're making large refactoring changes, are you checking that docs look right?

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -215,13 +217,21 @@ impl BlobClient {
...
         &self,
-        tags: RequestContent<BlobTags>,
+        tags: HashMap<String, String>,
         options: Option<BlobClientSetTagsOptions<'_>>,
     ) -> Result<Response<()>> {
-        self.client.set_tags(tags, options).await
+        let blob_tags = serialize_blob_tags(tags);
+        self.client
+            .set_tags(RequestContent::try_from(blob_tags)?, options)
+            .await
```
vincenttran-msft, line 234: This is the convenience layer with the addition of the helper. We now take `HashMap<String, String>` and do the heavy lifting before passing to generated layer.

PR #2663, opened 2025-06-04
Prepare azure_identity 0.24.1 for release

chlowell:


sdk/identity/azure_identity/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "azure_identity"
-version = "0.25.0"
+version = "0.24.1"
```
heaths: For a prerelease in a normal release cycle, just bump the minor version like we always do. It's common across semver to treat minor as major when it's < 1.0.

chlowell: I take it you consider the AzureCliCredential caching change worth a minor version? For me it's borderline, so I called this a patch release

heaths: It's a normal release cycle and we change the minor version. If we were OOB we could patch it. That said, there's been enough changes we typically don't think about it too much. I have released a couple of patch releases but they were true patch releases released OOB.

heaths (approved):

PR #2664, opened 2025-06-04
Create an AsyncRuntime trait

### LINKED ISSUE #2506: Redefine `sleep` function into async runtime trait
We need to consolidate a lot of the async runtime behavior into a single trait that handles `sleep` (today, that isn't pluggable), task spawning, maybe even process spawning and file system handling. HTTP client should probably remain separate, but maybe this `AsyncRuntime` trait or whatever could act as a factory there.

The reason being that it's not going to be likely anyone who decides to use `tokio`, `std-async`, or other async runtime would use one for some things and another for other things. So why make callers who want to use a different, non-default async runtime (for our Azure SDK crates, which is `tokio`) implement (or even discover) all the traits they need to impl?

The bigger problem will be how to publish that impl. As a lib, we have to be careful with statically registering anything i.e., most likely avoid it. That has ramifications on the app that would consume us. Today, as with any other languages, we tend to pass things through IoC-style and maybe we need to keep doing that here (like we do for the `HttpClient` impls). Then how to "register" that impl is an app concern and they can do what is best for their app e.g., `OnceLock` or something.

LarryOsterman:
Creates an `AsyncRuntime` trait which customers can call to replace our existing async runtime.
Fixes #2506.

This consolidates the `TaskSpawner` trait and the functionality underlying the `azure_core::sleep` function into a single trait.


sdk/core/azure_core/src/lib.rs
```diff
@@ -31,3 +30,10 @@ pub use typespec_client_core::{
 
 #[cfg(feature = "xml")]
 pub use typespec_client_core::xml;
+
+pub use typespec_client_core::get_async_runtime;
+pub use typespec_client_core::set_async_runtime;
+
+pub mod async_runtime {
+    pub use typespec_client_core::async_runtime::{AsyncRuntime, SpawnedTask, TaskFuture};
+}
```
heaths: You could just replace this with:
```suggestion
 pub use typespec_client_core::async_runtime;
```

sdk/typespec/typespec_client_core/src/lib.rs
```diff
@@ -23,3 +24,6 @@ pub mod xml;
 pub use crate::error::{Error, Result};
 pub use bytes::Bytes;
 pub use uuid::Uuid;
+
+pub use async_runtime::{get_async_runtime, set_async_runtime};
+pub use sleep::sleep;
```
heaths, line 28: Any reason to hoist this up to the root? Hardly seems worthy.

sdk/typespec/typespec_client_core/src/async_runtime/mod.rs
```diff
@@ -61,9 +65,11 @@ pub type SpawnedTask = Pin<
 pub type SpawnedTask =
     Pin<Box<dyn Future<Output = std::result::Result<(), Box<dyn std::error::Error>>> + 'static>>;
 
-/// An async command runner.
+/// An Asynchronous Runtime.
 ///
-pub trait TaskSpawner: Send + Sync + Debug {
+/// This trait defines the various
+///
+pub trait AsyncRuntime: Send + Sync {
```
heaths, line 72: Can't be `Send` on WASM.

sdk/typespec/typespec_client_core/src/lib.rs
LarryOsterman: Yeah, makes sense to me.

sdk/core/azure_core/src/lib.rs
LarryOsterman: Makes sense. I was primarily worried about types inside async_runtime leaking out into azure_core and typespec_client_core. 

I also moved all of this into `azure_core::async_runtime` rather than over promoting.

heaths: Those are equivalent. `typespec_client_core` should be a subset of `azure_core` anyway, and I want to try to keep the structure the same. Just makes everything easier. I did this same change with some other modules e.g., `http::response` where I didn't need to merge anything local.

sdk/typespec/typespec_client_core/src/async_runtime/mod.rs
LarryOsterman: FWIW, as far as I've been able to figure out, `Send` *is* defined on WASM.

The only thing about `Send` on WASM is that the future returned by a WASM `async fn` doesn't implement `Send`, thus it is incompatible with predefined futures which require `Send` (which is basically all non-WASM futures).

So it's my understanding that `Send` on a trait is fine (it says that an implementation of the trait can be passed between threads), but a function within the trait that returns a future cannot include `Send` in the future's declaration.

Please correct me if my understanding is incorrect.

sdk/core/azure_core/CHANGELOG.md
```diff
@@ -15,8 +18,8 @@
 ### Features Added
 
 - Added `#[safe]` attribute helper for `SafeDebug` derive macro to show or hide types and members as appropriate.
-- Added `get_async_runtime()` and `set_async_runtime()` to allow customers to replace
-the asynchronous runtime used by the Azure SDK.
+- Added `Page` trait to facilitate the `ItemIterator`.
+- Added `PageIterator` to asynchronously iterator all pages.
```
heaths: I realize this is a merge issue, but could fix this while you're at it:

```suggestion
- Added `PageIterator` to asynchronously iterate all pages.
```

sdk/core/azure_core/README.md
```diff
@@ -20,12 +20,12 @@ you can find the [package on crates.io][Package (crates.io)].
...
-- Paging and asynchronous streams (`Pager<T>`).
-- Errors from service requests in a consistent fashion. (`azure_core::Error`).
-- Customizing requests (`ClientOptions`).
-- Abstractions for representing Azure SDK credentials. (`TokenCredentials`).
+-   Configuring service clients, e.g. configuring retries, logging (`ClientOptions`).
+-   Accessing HTTP response details (`Response<T>`).
+-   Paging and asynchronous streams (`Pager<T>`).
+-   Errors from service requests in a consistent fashion. (`azure_core::Error`).
+-   Customizing requests (`ClientOptions`).
+-   Abstractions for representing Azure SDK credentials. (`TokenCredentials`).
```
heaths, line 28: Why were all these changes made? They are unnecessary. Was it automatic formatting? If so, fine. Leave as-is. If it was Copilot, re-save the file and let the formatter do it's thing or it's going to be a constant problem. We can fix `copilot-instructions.md` or something if needs be.

sdk/typespec/typespec_client_core/src/async_runtime/tests.rs
```diff
@@ -241,14 +241,29 @@ fn test_get_runtime() {
     let _runtime = get_async_runtime();
 }
 
+struct DummyRuntime;
```
heaths: Nit: I remember reading somewhere we shouldn't use words like "dummy". Maybe just "TestRuntime" or something?

heaths (approved):

sdk/core/azure_core/README.md
LarryOsterman: I believe it was auto formatting - let me check.

LarryOsterman: It's auto formatting - if I remove the extra spaces, they come back when I try to save the file :(.

sdk/typespec/typespec_client_core/src/async_runtime/tests.rs
LarryOsterman: Fixing, you're right.

heaths (approved):

PR #2667, opened 2025-06-05
[Storage] `AppendBlobClient`, `PageBlobClient`, and initial `create` APIs

vincenttran-msft:
WIP


sdk/storage/azure_storage_blob/tests/page_blob_client.rs
```diff
@@ -0,0 +1,143 @@
...
+    let container_client = get_container_client(recording, true).await?;
+    let blob_client = container_client.blob_client(get_blob_name(recording));
+    let page_blob_client = blob_client.page_blob_client();
+    page_blob_client.create(512, None).await?;
+    let data = vec![b'A'; 512];
+    page_blob_client
+        .upload_page(
+            RequestContent::from(data.clone()),
+            512,
+            "bytes=0-511".to_string(),
```
vincenttran-msft: This is pretty cumbersome to expect cx to format this. Curious to see if we can resolve this in generated code, or what is the preferred helper method authoring?

sdk/storage/azure_storage_blob/tests/page_blob_client.rs
```diff
@@ -0,0 +1,143 @@
...
+        .upload_page(
+            RequestContent::from(data),
+            512,
+            "bytes=0-511".to_string(),
+            None,
+        )
+        .await?;
+
+    page_blob_client
+        .clear_page("bytes=0-511".to_string(), None)
```
vincenttran-msft: Same here for `clear_page`

sdk/storage/azure_storage_blob/tests/page_blob_client.rs
```diff
@@ -0,0 +1,164 @@
...
+    let blob_properties = blob_client.get_properties(None).await?;
+    let blob_type = blob_properties.blob_type()?;
+    let content_length = blob_properties.content_length()?;
+    assert_eq!(1024, content_length.unwrap());
+    assert_eq!(BlobType::PageBlob, blob_type.unwrap());
+
+    // Create If Not Exists Scenario
+    let create_options = PageBlobClientCreateOptions {
+        if_none_match: Some("*".to_string()),
+        ..Default::default()
```
vincenttran-msft: Wondering if there is any cool ways we could make a helper for this i.e. like somehow have a `.if_not_exists()` that you can call on the options bag -- another pattern we haven't yet discussed (how to go about modifying customer's options bag)

Would have to be some sort of generic in-place modifier helper function(?), because we cannot just add a function on these generated options bags.

sdk/storage/azure_storage_blob/tests/page_blob_client.rs
```diff
@@ -28,10 +29,7 @@ async fn test_create_page_blob(ctx: TestContext) -> Result<(), Box<dyn Error>> {
     assert_eq!(BlobType::PageBlob, blob_type.unwrap());
 
     // Create If Not Exists Scenario
-    let create_options = PageBlobClientCreateOptions {
-        if_none_match: Some("*".to_string()),
-        ..Default::default()
-    };
+    let create_options = page_blob_create_if_not_exists(PageBlobClientCreateOptions::default());
```
vincenttran-msft: Wondering if there is a better way to achieve this effect? We would like to abstract out the logic to get this to be "create if not exists". 

Even better if we could make this generic, as this is likely not the only API that would support "if not exists". Maybe a Trait of some sort and we somehow mark all the applicable option bags with this Trait?

sdk/storage/azure_storage_blob/src/clients/page_blob_client.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    /// Resizes a Page blob to the specified size. If the specified value is less than
+    /// the current size of the blob, then all pages above the specified value are cleared.
+    ///
+    /// # Arguments
+    ///
+    /// * `size` - Size used to resize the blob. Maximum size for a page Blob is up to 1TB. The
+    ///   Page blob size must be aligned to a 512-byte boundary.
+    /// * `options` - Optional parameters for the request.
+    pub async fn resize_blob(
```
jalauzon-msft: If we are sticking with past patterns, this should probably just be `resize`?

sdk/storage/azure_storage_blob/src/serialize.rs
```diff
@@ -0,0 +1,31 @@
...
+    let content_range = format!("bytes={}-{}", offset, end_range);
+    content_range
+}
+
+pub fn page_blob_create_if_not_exists(
+    mut options: PageBlobClientCreateOptions,
+) -> PageBlobClientCreateOptions {
+    options.if_none_match = Some("*".to_string());
+    options
+}
```
vincenttran-msft: This is our 2 current helper function candidates. However, they both represent two unique cases- and so we are looking for input on the best way to implement them.
For `format_http_range`, currently we are just calling this out in docstring to hopefully ensure it is obvious that it should be used:
>/// * `range` - The range of bytes to clear. See [`azure_storage_blob::serialize::format_http_range()`](crate::serialize::format_http_range) for help with the expected String format.

In comparison, `page_blob_create_if_not_exists` is something that modifies an option on the options bag. Two main areas of concern:

- Generics: Currently this only works with specifically `PageBlobClientCreateOptions` type, but this pattern will be repeated (i.e. this would apply for `BlockBlobClientUploadOptions` as well), so is there a way to make this genericized (but still with fine-grain adjustments to only allow certain options bags)? The tricky part is that these are the generated options bags.
- If the above is not possible, is there a way to somehow make it associated with the options bag, i.e. `options.create_if_not_exist` or some sort of associated method?
cc: @heaths Curious to hear your thoughts here Heath, so we can get the right pattern moving forward

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -29,7 +31,12 @@ pub mod models {
...
-        ListBlobsFlatSegmentResponse, ListContainersSegmentResponse, PublicAccessType,
-        RehydratePriority, StorageServiceProperties,
+        ListBlobsFlatSegmentResponse, ListContainersSegmentResponse,
+        PageBlobClientClearPagesOptions, PageBlobClientClearPagesResult,
+        PageBlobClientCreateOptions, PageBlobClientCreateResult, PageBlobClientResizeOptions,
+        PageBlobClientResizeResult, PageBlobClientUploadPagesOptions,
+        PageBlobClientUploadPagesResult, PublicAccessType, RehydratePriority,
+        StorageServiceProperties,
     };
+    pub use crate::serialize::*;
```
vincenttran-msft: More generally, how do we feel about the implementation of the helpers and what module it lives in and how it is exposed? Happy to move to something more idiomatic if it exists.

sdk/storage/azure_storage_blob/tests/page_blob_client.rs
vincenttran-msft: Closed as duplicate on megathread on helpers.

heaths (approved):

sdk/storage/azure_storage_blob/src/serialize.rs
heaths: Why an unassociated function? These aren't as discoverable. You're also taking a `mut` by value, which is odd since that puts a burden on callers and you're just returning a "copy" (moved, actually) anyway. Use the `Self { if_none_match = Some("*".into()), ..self }` syntax instead so you don't need `mut`, which ends up doing the same thing anyway.

But if you want to repeat the pattern, take a page from the convenience client guidelines: https://azure.github.io/azure-sdk/rust_implementation.html#rust-client-convenience

Define a trait that you can implement for different method option types. It also means you don't need unassociated functions like this.

sdk/storage/azure_storage_blob/src/lib.rs
heaths: See https://azure.github.io/azure-sdk/rust_implementation.html#rust-client-convenience for guidelines. In general, helpers should be in the highest-level module for the types they augment and end with "Ext".

sdk/storage/azure_storage_blob/src/models/storage_traits.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+impl PageBlobClientCreateOptionsExt for PageBlobClientCreateOptions<'_> {
+    fn with_if_not_exists(self) -> Self {
+        Self {
+            if_none_match: Some("*".into()),
+            ..self
+        }
+    }
+}
```
vincenttran-msft: Since this trait augments the generated models, and the generated models are re-exported in our models, I think this should be the appropriate place to put this extension trait.

sdk/storage/azure_storage_blob/src/storage_traits.rs
vincenttran-msft, line 1: Based on the feedback, I moved both helpers to a `storage_traits.rs` file. Let me know if any further changes should be made 😄

sdk/storage/azure_storage_blob/tests/page_blob_client.rs
```diff
@@ -29,7 +28,47 @@ async fn test_create_page_blob(ctx: TestContext) -> Result<(), Box<dyn Error>> {
     assert_eq!(BlobType::PageBlob, blob_type.unwrap());
 
     // Create If Not Exists Scenario
-    let create_options = page_blob_create_if_not_exists(PageBlobClientCreateOptions::default());
+    let create_options = PageBlobClientCreateOptions::default().with_if_not_exists();
```
heaths: I was thinking more about this last night, and the extension method idea might not work well here for client method options. It seems like what you really want is a custom client method options, in which case you could define your own that converts into a generated one and never expose the generated one, much like I did in #2737 recently e.g.,

```rust
use crate::{azure_core::http::ClientMethodOptions, generated::models::Tag};

#[derive(Debug, Default)]
pub struct UploadOptions {
    pub method_options: ClientMethodOptions,
    pub tags: HashMap<String, String>,
}

impl From<UploadOptions> for generated::models::UploadOptions {
    fn from(value: UploadOptions) -> Self {
        Self {
            method_options: value.method_options,
            tags: value
                .tags
                .into_iter()
                .map(|(k, v)| Tag {
                    key: Some(k.into()),
                    value: Some(v.into()),
                })
                .collect(),
        }
    }
}

fn main() {
    let mut options = UploadOptions::default();
    options.tags.insert("foo".into(), "1".into());
    options.tags.insert("bar".into(), "baz".into());

    upload(Some(options));
}

fn upload(options: Option<UploadOptions>) {
    let options: generated::models::UploadOptions = options.unwrap_or_default().into();
    println!("{options:#?}");
}

mod generated {
    pub mod models {
        use crate::azure_core::http::ClientMethodOptions;

        #[derive(Debug, Default)]
        pub struct Tag {
            pub key: Option<String>,
            pub value: Option<String>,
        }

        #[derive(Debug, Default)]
        pub struct UploadOptions {
            pub method_options: ClientMethodOptions,
            pub tags: Vec<Tag>,
        }
    }
}

pub mod azure_core {
    pub mod http {
        #[derive(Debug, Default)]
        pub struct ClientMethodOptions;
    }
}
```

Because this is repeated for several client options, you could define a helper or even implement `From<HashMap<String, String>> into Tag` or something to avoid duplication.

jalauzon-msft: @heaths, thanks for the suggestion but our end goal with Blob is to be as generated as possible and hopefully eliminate our hand-written convenience methods. So, adding custom options bags, moves against this goal. I actually liked the Trait approach where we could add handwritten functionality to a generated options bag. This is something that we could keep, even if the client was fully generated.

This is for options bag helpers however, in this PR we also have a need for a more general helper that formats and offset length into the HTTP range header. The HTTP range header is needed for multiple APIs and is not in an options bag as it is a required parameter. So, we are looking for a solution here as well. I think a well-documented unassociated function in some util module is probably the best we can do?

Vincent will push a new iteration with our thoughts but open to more feedback.

sdk/storage/azure_storage_blob/src/storage_traits.rs
```diff
@@ -0,0 +1,21 @@
...
+}
+
+impl PageBlobClientCreateOptionsExt for PageBlobClientCreateOptions<'_> {
+    fn with_if_not_exists(self) -> Self {
+        Self {
+            if_none_match: Some("*".into()),
+            ..self
+        }
+    }
+}
```
vincenttran-msft: This is where we landed for the scenario: Options bag option mutation
We went with an extension method on the generated options bag

sdk/storage/azure_storage_blob/src/storage_utils.rs
```diff
@@ -0,0 +1,26 @@
...
+        if offset % 512 != 0 {
+            panic!("offset must be aligned to a 512-byte boundary.");
+        }
+        if length % 512 != 0 {
+            panic!("length must be aligned to a 512-byte boundary.");
+        }
+        let end_range = offset + length - 1;
+        let content_range = format!("bytes={}-{}", offset, end_range);
+        content_range
+    }
```
vincenttran-msft: This is where we landed for the scenario: General use helper functions
Our current plan is that any of these general use will be implemented here and called out in the docstring.

sdk/storage/azure_storage_blob/src/clients/page_blob_client.rs
```diff
@@ -0,0 +1,158 @@
...
+        &self.client.blob_name
+    }
+
+    /// Creates a new Page blob.
+    ///
+    /// # Arguments
+    ///
+    /// * `content_length` - The maximum size for the Page blob, up to 1TB. The page blob size must
+    ///   be aligned to a 512-byte boundary.
+    /// * `options` - Optional parameters for the request. See [`azure_storage_blob::storage_traits::PageBlobClientCreateOptionsExt`](crate::storage_traits::PageBlobClientCreateOptionsExt) for additional usage helpers.
```
vincenttran-msft: This is how we plan to ensure discoverability. Docs render:
image

sdk/storage/azure_storage_blob/src/clients/page_blob_client.rs
```diff
@@ -0,0 +1,158 @@
...
+        options: Option<PageBlobClientCreateOptions<'_>>,
+    ) -> Result<Response<PageBlobClientCreateResult, NoFormat>> {
+        self.client.create(content_length, options).await
+    }
+
+    /// Clears a range of pages.
+    ///
+    /// # Arguments
+    ///
+    /// * `range` - The range of bytes to clear. See [`azure_storage_blob::StorageUtils::format_http_range()`](crate::StorageUtils::format_http_range) for help with the expected String format.
```
vincenttran-msft: This is how we plan to ensure discoverability of the unassociated functions. Docs render:
image

sdk/storage/azure_storage_blob/src/storage_utils.rs
heaths: Panicking on user input is not idiomatic to Rust. These need to return `Result<String>`.

sdk/storage/azure_storage_blob/src/clients/page_blob_client.rs
heaths: Don't make the whole path a link. For one, if it's already in scope, you don't need to qualify it. If not, use markdown to shorten it:
```suggestion
    /// * `options` - Optional parameters for the request. See [`PageBlobClientCreateOptionsExt`](crate::storage_traits::PageBlobClientCreateOptionsExt) for additional usage helpers.
```

sdk/storage/azure_storage_blob/src/clients/page_blob_client.rs
heaths: Again, don't fully-qualify the whole thing. It's harder to read and unnecessary. It's a link, and clicking on it tells them all they need to know. The rust-analyzer LSP will import/qualify things as needed.

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -10,9 +10,10 @@
 pub mod clients;
 mod generated;
 mod pipeline;
-mod storage_traits;
+pub mod storage_traits;
+pub mod storage_utils;
 pub use clients::*;
-pub use storage_traits::*;
+pub use storage_utils::*;
```
heaths: Don't re-export with a name like "storage*". It's obvious in storage because it's the storage crate. Be concise. Think of every other crate you're used.

I wouldn't even export this from a submodule or, if anything, maybe from `clients` or even the root.

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -10,9 +10,10 @@
 pub mod clients;
 mod generated;
 mod pipeline;
-mod storage_traits;
+pub mod storage_traits;
```
heaths: Same here. "storage" is redundant and don't divide up your crates into lots of different modules (publicly). It's fine to maintain them that way, but don't export them. Think of what you'd do in all the other languages.

sdk/storage/azure_storage_blob/src/storage_utils.rs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+pub struct StorageUtils;
+
+impl StorageUtils {
```
heaths: This isn't C#. ;) You can declare just functions. `azure_core` has quite a few, as do most of the crates we call. I would just define this as a standalone function. There's no benefit of putting it in a type that otherwise has no purpose.

heaths (changes requested):
Some refactoring requested, but otherwise LGTM

vincenttran-msft:
Latest revision pushed changes:

- Addressed PR feedback
- Moved `models` module to it's own directory + `mod.rs`
- Added `extensions.rs` (formerly `storage_traits.rs`) to `models` module that will hold all option bag extension methods
- Renamed `storage_utils.rs` to `helpers.rs` to hold all unassociated functions

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -9,34 +9,8 @@
 
 pub mod clients;
 mod generated;
+pub mod helpers;
```
heaths: The module shouldn't be public, but just export it from the root: https://azure.github.io/azure-sdk/rust_implementation.html#rust-modules

Lots of little modules just makes things hard to find.

heaths: And if it's not meant to be public, it should be `pub(crate)`: https://azure.github.io/azure-sdk/rust_introduction.html#rust-namespace-naming-internals

heaths (changes requested):
I don't think we should expose a `helpers` module. What purpose does it serve? Seems better to expose it from the root if it applies pretty much crate-wide. That would follow guidelines.

sdk/storage/azure_storage_blob/src/lib.rs
jalauzon-msft: Hi @heaths, we did consider putting this at the root but felt it was a little better to not have these helpers pollute the root namespace. It just felt a little odd to do `azure_storage_blob::format_page_range` directly. We made the generic enough, so we don't add a bunch more modules later.

If it is truly more idiomatic to export these types of helpers from the root, so be it, just thought I would share my opinion.

heaths: Why even have a helper function? Do you in other languages? Is this not just baked into range requests? In rust, there tends to be a pushback on "utils", and "helpers" really is just another term for "utils".

jalauzon-msft: @heaths, in other languages our APIs that need a range typical take an `offset` and `length` as parameters in our handwritten layers and then build the Range header for the user. While that is possible here, we continue to keep "fully generated" clients in mind and taking an offset and a length would not match the generated client (as it does not match the REST API). We wanted to introduce a helper instead of changing the inputs and changing the inputs would not be possible later when we have fully generated clients.

We could change our inputs for now, but this is something we'd need to consider when moving to fully generated. Either bake it into the TypeSpec/emitter (which then wouldn't match the REST API) or consider adding a more friendly method via an extension Trait (which I would like to avoid because then we would have duplicate APIs in the client that would confuse customers).

heaths: Mentioned in our meeting, but you might also look at the `Header` trait which allows you to define "strongly-typed headers" that users could add to their request. That said, I wouldn't hold up the PR for this. Can do it in a later PR if you want and when you have more time to evaluate the ergonomics.

vincenttran-msft: To prevent further delays on this, we have resolved this by changing to `parsers.py` naming convention, and exporting from the root.

vincenttran-msft:
@heaths Heath, We have decided to move from `helpers.rs` to `parsers.rs` and re-exported the module from the root. 

I am currently hitting a compilation error from `typespec_client_core` [here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=5056802&view=logs&j=2c0cb571-58e4-5f13-9d77-617bed7e8990&t=3768c134-7568-558c-110f-6089b91d12de&l=405) even after merging in `main`, but other than that the PR is ready to review, thanks!

jalauzon-msft (approved):

jalauzon-msft (approved):

PR #2670, opened 2025-06-05
Refactored RecoverableConnection

LarryOsterman:
Refactored the `RecoverableConnection` struct to split out the individual recoverable types into their own modules to simplify and concentrate the logic throughout the recoverable code logic.

There's not a lot of actual code changing here, mostly this is about moving code around to isolate individual implementations of Recoverable types. Also renamed all the types from `AzureXxxClient` to `RecoverableXxx` to make it clearer what's going on.

Bonus: Changed jitter parameters to token refresher to be `Duration` structs instead of number of seconds - that allows tests to use extremely fine grained jitter amounts.


sdk/eventhubs/azure_messaging_eventhubs/src/common/authorizer.rs
```diff
@@ -174,7 +174,7 @@ impl Authorizer {
     async fn refresh_tokens_task(self: Arc<Self>) {
         let result = self.refresh_tokens().await;
         if let Err(e) = result {
-            error!("Error refreshing tokens: {e}");
+            warn!("Error refreshing tokens: {e}");
```
heaths: Did you want to do the same thing here I recommended in the last PR...which I think you did?
```suggestion
            warn!(err=?e, "Error refreshing tokens: {e}");
```

sdk/eventhubs/azure_messaging_eventhubs/src/common/authorizer.rs
```diff
@@ -250,10 +250,10 @@ impl Authorizer {
 
                 debug!("Token refresh times: {token_refresh_times:?}");
 
-                let expiration_jitter = Duration::seconds(
-                    thread_rng()
-                        .gen_range(token_refresh_times.jitter_min..token_refresh_times.jitter_max),
-                );
+                let jitter_min = (token_refresh_times.jitter_min.as_seconds_f64() * 1000.0) as i64;
+                let jitter_max = (token_refresh_times.jitter_max.as_seconds_f64() * 1000.0) as i64;
```
heaths: There's not already an `as_milliseconds` or something? Are these not `Duration`s? There is a `Duration::as_millis_f64()` so no resolution is lost if sub-second.

sdk/eventhubs/azure_messaging_eventhubs/src/common/authorizer.rs
```diff
@@ -543,8 +543,8 @@ mod tests {
         authorizer
             .set_token_refresh_times(TokenRefreshTimes {
                 before_expiration_refresh_time: Duration::seconds(10),
-                jitter_min: -2,
-                jitter_max: 2,
+                jitter_min: Duration::seconds(-2),
+                jitter_max: Duration::seconds(2),
```
heaths, line 547: These differ from your +/-5s defined as constants above. Why were those not reused here?

sdk/eventhubs/azure_messaging_eventhubs/src/common/authorizer.rs
```diff
@@ -611,8 +611,8 @@ mod tests {
         authorizer
             .set_token_refresh_times(TokenRefreshTimes {
                 before_expiration_refresh_time: Duration::seconds(5),
-                jitter_min: -1,
-                jitter_max: 1,
+                jitter_min: Duration::milliseconds(-500),
```
heaths, line 614: Different still from above, and not using the constants. Or, as I'm starting to wonder, is this intentional?

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable/connection.rs
```diff
@@ -0,0 +1,473 @@
...
+/// * Notes
+///
+/// The way a client uses a `RecoverableConnection` is as follows:
+///   1. Create a new instance of the `RecoverableConnection`.
+///   2. Retrieve an interim object from the `RecoverableConnection`. Supported
+///      interim objects are:
+///    - `AmqpManagement`: Used for management operations.
+///    - `AmqpSender`: Used for sending messages to the Event Hubs service.
+///    - `AmqpReceiver`: Used for receiving messages from the Event Hubs service.
+///    - `AmqpClaimsBasedSecurity`: Used for authorization operations (should not be used directly)
```
heaths: Markdown requires that sub-bullets be indented to the same level as the text. That said, I don't know if the docs.rs render will do the right thing, but I wouldn't change it. These need to be indented 3 spaces (total, so 2 more). You could always try it with `cargo doc --open`, though.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable/connection.rs
```diff
@@ -0,0 +1,473 @@
...
+            timeout,
+        ))
+    }
+
+    pub(crate) async fn close_receiver(self: &Arc<Self>, source_url: &Url) -> Result<()> {
+        let mut receiver_instances = self.receiver_instances.lock().await;
+        if let Some(receiver) = receiver_instances.remove(source_url) {
+            let r = Arc::try_unwrap(receiver);
+            if let Ok(receiver) = r {
+                trace!("Detaching receiver: {:?}", source_url);
```
heaths, line 205: Tip: only because you inline some stuff elsewhere, in case you weren't aware, you can also combine these:
```suggestion
                trace!("Detaching receiver: {source_url:?}");
```
You don't have to change this, though. Just a tip.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable/connection.rs
```diff
@@ -0,0 +1,473 @@
...
+    pub(super) async fn ensure_amqp_cbs(self: &Arc<Self>) -> Result<Arc<AmqpClaimsBasedSecurity>> {
+        debug!("Ensuring AMQP Claims-Based Security (CBS) client.");
+
+        let connection = self.ensure_connection().await?;
+        let cbs_client = RecoverableClaimsBasedSecurity::create_claims_based_security(
+            connection.clone(),
+            &self.retry_options,
+        )
+        .await?;
+        debug!("AMQP Claims-Based Security (CBS) client ensured.");
```
heaths: Would a span make more sense here? It would automatically have begin/end and encapsulate any child spans/logs. Just a thought.

sdk/eventhubs/azure_messaging_eventhubs/src/common/authorizer.rs
LarryOsterman: Yes.

sdk/eventhubs/azure_messaging_eventhubs/src/common/authorizer.rs
LarryOsterman: There's a `whole_milliseconds` I can use here, which is good enough.

Unfortunately, there isn't a `as_millis_f64()` in `core::time::Duration`.

There is one for `std::time::Duration`, but azure_core has standardized on `core::time::Duration`, not `std::time::Duration`. I believe that's because `tokio::time::sleep()` takes a `time::Duration`. 

And `std::time::Duration` is unsigned, not signed, and `core::time::Duration` is signed. 

So it's a better fit for what I'm trying to do.

sdk/eventhubs/azure_messaging_eventhubs/src/common/authorizer.rs
LarryOsterman: Because for this test, I want a max 4 second jitter, not a 10 second jitter - smaller jitter means less variability. And a 10s jitter on a 10 second token expiration time is hard to test deterministically.

sdk/eventhubs/azure_messaging_eventhubs/src/common/authorizer.rs
LarryOsterman: 100% intentional - jitter intentionally introduces noise in the system, but noise is horrible for deterministic tests.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable/connection.rs
LarryOsterman: Yup - it actually worked correctly, but it was still wrong.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable/connection.rs
LarryOsterman: Probably not a bad idea, tbh.

sdk/eventhubs/azure_messaging_eventhubs/src/common/authorizer.rs
heaths: Weird. `std` re-exports from `core`. Wonder why they don't in this case. Maybe it is a platform issue since `core` works in more places.

LarryOsterman: I have this somewhat backwards (more accurately, I got some utterly critical facts wrong):

we export: `core::time::OffsetDateTime` from typespec_client_core.
we *don't* export `Duration` from anywhere (this is important).

typespec_client_core's lib.rs uses `std::time::Duration` as an alias but does not publish it
typespec_client_core's sleep.rs uses `core::time::Duration` because `tokio::time::sleep` uses `core::time::Duration`

authorizer.cs uses `core::time::Duration`, because (a) `core::time::Duration` is signed and (b) `azure_core::sleep()` uses `core::time::Duration`
Most of the rest of the Azure SDK uses `std::time::Duration`. 

Let me see if I can figure out a better solution that uses `std::time::Duration` like the rest of the tree (I'm not going to touch the AMQP stuff - this should be isolated to just this file).

LarryOsterman: Ugg.

typespec_client_core::AccessToken's `expires_on` is an `core::OffsetDateTime` (because that's what typespe_client_core has standardized on.

But `core::time::OffsetDateTime::checked_add` requires a `core::time::Duration`. And a `std::time::Duration` won't cut it.

This isn't a problem for the rest of the world because all of our credentials simply do everything in relation to `core::time::OffsetDateTime` - they never deal with `Duration` values directly.

But anyone trying to use `std::time::Duration` to interact with a `AccessToken::expires_on` is going to have a sad.

I'm not going to worry about fixing this right now, but it's something to consider moving forward.

heaths: `std::time::Duration` is `core::time::Duration`: https://doc.rust-lang.org/std/time/struct.Duration.html -> https://doc.rust-lang.org/src/core/time.rs.html#81.

I agree we should settle on one and it likely should be `core`/`std` but these should be the same.

heaths (approved):

PR #2687, opened 2025-06-09
PerfTest prototype

heaths:
We don't use built-in `cargo bench` benchmarks for various reasons including they are only supported in `nightly`. But we also want to allow `criterion` to work, so we can define our own test harness (our own main).

I think we do something similar to `TestContext` in `azure_core_test` and reuse as much code as possible to support live test resources i.e., requiring provisioned resources and a `.env` file. We define a proc macro to drive it, such that we just add it to an async function returning an `azure_core::Result<()>`. If the .env file is not found, we should exit without error: there's nothing to do.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -342,6 +342,7 @@ dependencies = [
  "azure_core_test",
  "azure_identity",
  "azure_security_keyvault_test",
+ "clap",
  "criterion",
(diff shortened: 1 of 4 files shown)


heaths:
Doesn't need to be open to show prototype

PR #2690, opened 2025-06-10
Add NoFormat response type

jhendrixMSFT:
Tweaked a few of docs for the other format types.

### DIFF
--- a/sdk/typespec/typespec_client_core/src/http/format.rs
+++ b/sdk/typespec/typespec_client_core/src/http/format.rs
@@ -37,7 +37,8 @@ impl<D: DeserializeOwned> DeserializeWith<JsonFormat> for D {
     }
 }
 
-/// The default format used for deserialization.
+/// A [`Format`] that deserializes response bodies using JSON.
+/// This is the default format used for deserialization.
 ///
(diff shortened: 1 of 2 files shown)


sdk/typespec/typespec_client_core/src/http/format.rs
```diff
@@ -37,7 +37,8 @@ impl<D: DeserializeOwned> DeserializeWith<JsonFormat> for D {
     }
 }
 
-/// The default format used for deserialization.
+/// A [`Format`] that deserializes response bodies using JSON.
+/// This is the default format used for deserialization.
```
heaths, line 41: Nit: per markdown rules, these are actually a single rendered line as the "title" - the one-liner that will show up in an API listing on docs.rs, for example. And maybe that's fine, but if you intended it as a separate line you'll need a blank line in between.

heaths (approved):

sdk/typespec/typespec_client_core/src/http/format.rs
jhendrixMSFT: This was my intent as I wanted to leave the blurb about this being the default in the one-liner.

PR #2692, opened 2025-06-10
[Storage] `list_containers` for BlobServiceClient

vincenttran-msft:
WIP

Current sysout:
>Is Deleted Option Some: false
Is Metadata Option Some: false
Is Name Option Some: false
Is Properties Option Some: false
Is Version Option Some: false

Issue seems to be related to deserializing into `ContainerItem` as the fields do exist on the response, but don't seem to be properly set on the item.

Sample Response
```xml
HTTP/1.1 200 OK
Content-Type: application/xml
Server: Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0
x-ms-request-id: 1267e11b-601e-001b-0940-dac970000000
x-ms-client-request-id: 0e8c8184-9e79-43a2-855b-04ae13ffc650
x-ms-version: 2025-01-05
Date: Tue, 10 Jun 2025 19:46:07 GMT
Content-Length: 115153


<EnumerationResults ServiceEndpoint="https://ruststoragedev.blob.core.windows.net/">
	<Containers>
		<Container>
			<Name>container0oqovzx8</Name>
			<Properties>
				<Last-Modified>Fri, 02 May 2025 22:01:10 GMT</Last-Modified>
				<Etag>"0x8DD89C4D92560B5"</Etag>
				<LeaseStatus>unlocked</LeaseStatus>
				<LeaseState>available</LeaseState>
				<DefaultEncryptionScope>$account-encryption-key</DefaultEncryptionScope>
				<DenyEncryptionScopeOverride>false</DenyEncryptionScopeOverride>
				<HasImmutabilityPolicy>false</HasImmutabilityPolicy>
				<HasLegalHold>false</HasLegalHold>
				<ImmutableStorageWithVersioningEnabled>false</ImmutableStorageWithVersioningEnabled>
			</Properties>
		</Container>
```

### DIFF
--- a/sdk/storage/azure_storage_blob/tests/blob_service_client.rs
+++ b/sdk/storage/azure_storage_blob/tests/blob_service_client.rs
@@ -4,6 +4,7 @@
 use azure_core_test::{recorded, TestContext};
 use azure_storage_blob::models::BlobServiceClientGetPropertiesOptions;
 use azure_storage_blob_test::get_blob_service_client;
+use futures::StreamExt;
 use std::error::Error;
 
 #[recorded::test]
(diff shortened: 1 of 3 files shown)


sdk/storage/azure_storage_blob/tests/blob_service_client.rs
```diff
@@ -22,3 +23,69 @@ async fn test_get_service_properties(ctx: TestContext) -> Result<(), Box<dyn Err
...
+            println!("Is Deleted Option Some: {}", container.delete.is_some());
+            println!("Is Metadata Option Some: {}", container.metadata.is_some());
+            println!("Is Name Option Some: {}", container.name.is_some());
+            println!(
+                "Is Properties Option Some: {}",
+                container.properties.is_some()
+            );
+            println!("Is Version Option Some: {}", container.version.is_some());
+        }
+    }
```
vincenttran-msft, line 86: This is the main area of concern- here I would expect `is_some()` to return something, for at the very very least `name` but also the other fields.

jhendrixMSFT: This will be fixed as part of https://github.com/Azure/typespec-rust/pull/455

PR #2696, opened 2025-06-11
add Into conversions from Etag to String and HeaderValue

analogrelay:
While starting to look at implementing some etag-related headers in Cosmos, I noticed that it's a little clunky to get an `Etag` (from `typespec_client_core`) into a `HeaderValue`. We only expose the etag as a `&str`, but `HeaderValue`s require owned string. This means we'll have to clone the underlying string in order to use it. Since, in our case at least, we only need the Etag once, I added some `Into` conversions that consume the `Etag` and return the underlying string. While I was there, I added a similar conversion to `HeaderValue`, allowing an `Etag` to be passed straight in to `Request::insert_header`.

To do that I had to remove the `impl From<T: Into<String>> for Etag` implementation because it's too generic. I replaced it with `From<&str>` and `From<String>` which should cover most cases. A user can always call `Etag::from(s.into())` if they have an `s` that is `Into<String>` but isn't `&str` or `String`.

### DIFF
--- a/sdk/typespec/typespec_client_core/src/http/models/etag.rs
+++ b/sdk/typespec/typespec_client_core/src/http/models/etag.rs
@@ -4,16 +4,22 @@
 use serde::{Deserialize, Serialize};
 use std::{fmt, str::FromStr};
 
+use crate::http::headers::HeaderValue;
+
 /// Represents an ETag for versioned resources.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct Etag(String);
 
-impl<T> From<T> for Etag
-where
-    T: Into<String>,
-{
-    fn from(t: T) -> Self {
-        Self(t.into())
+// Implementation for common string types
+impl From<&str> for Etag {
+    fn from(s: &str) -> Self {
+        Self(s.to_string())
+    }
+}
+
+impl From<String> for Etag {
+    fn from(s: String) -> Self {
+        Self(s)
     }
 }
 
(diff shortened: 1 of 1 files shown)


sdk/typespec/typespec_client_core/src/http/models/etag.rs
```diff
@@ -4,16 +4,22 @@
...
+// Implementation for common string types
+impl From<&str> for Etag {
+    fn from(s: &str) -> Self {
+        Self(s.to_string())
+    }
+}
+
+impl From<String> for Etag {
+    fn from(s: String) -> Self {
+        Self(s)
```
heaths, line 22: Given `<String as Into>::into(self)` is basically a no-op (moved, but probably elided in release), why was the old impl so bad?

heaths (approved):
Copilot might have a good point, or are we safe since `Etag` would've already validated?

sdk/typespec/typespec_client_core/src/http/models/etag.rs
analogrelay: Because we now have a `From<Etag>` impl, the `From<T> for Etag` impl conflicts:

image

analogrelay: I'll be honest, I'm not entirely sure _why_ this causes it to conflict, since the conflicting impl is `impl<T> From<T> for T` which should _always_ conflict?

sdk/typespec/typespec_client_core/src/http/models/etag.rs
```diff
@@ -35,3 +41,59 @@ impl fmt::Display for Etag {
...
+
+impl From<Etag> for String {
+    fn from(etag: Etag) -> Self {
+        etag.0
+    }
+}
+
+impl From<Etag> for HeaderValue {
+    fn from(etag: Etag) -> Self {
+        HeaderValue::from(String::from(etag))
```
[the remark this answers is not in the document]
analogrelay: There's no validation on `HeaderValue`. Header _names_ have validation (ensuring all lower-case names, etc.) but there's none on `HeaderValue`.

sdk/typespec/typespec_client_core/src/http/models/etag.rs
heaths: Ah, yes, that ol' issue. Has bitten me too many times.

PR #2702, opened 2025-06-12
[Storage] Regenerate against `v0.16.0`, Fix ResultHeaders in TypeSpec

vincenttran-msft:
This PR does the following:

- Regenerates against `v0.16.0`
- Working to correct how we hide ResultHeaders in order to be compliant with TypeSpec requirements. This means instead of removnig them from the spec entirely, we are now moving towards using the `Access.internal` and `@access` decorator to dictate whether it should appear in the deserialized response.
Currently generating against: https://github.com/Azure/azure-rest-api-specs/pull/35254

Pending the move to aliases for the template headers, then will apply that to complete this PR.


sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -103,7 +103,7 @@ impl BlobContainerClient {
     pub async fn create_container(
         &self,
         options: Option<BlobContainerClientCreateOptions<'_>>,
-    ) -> Result<Response<()>> {
+    ) -> Result<Response<(), NoFormat>> {
```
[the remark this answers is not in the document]
vincenttran-msft: Marking as resolved -- this is dictated by the generated layer and this is just passthrough.

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -15,7 +15,7 @@ pub use clients::*;
 
 pub mod models {
     pub use crate::generated::models::{
-        AccessTier, ArchiveStatus, BlobClientDeleteOptions, BlobClientDownloadOptions,
+        AccessTierOptional, ArchiveStatus, BlobClientDeleteOptions, BlobClientDownloadOptions,
```
heaths, line 18: Nit: why the suffix "optional"? Nothing about its usage says "optional". Seems a bit confusing.

heaths (approved):

sdk/storage/azure_storage_blob/src/lib.rs
vincenttran-msft: Yes, admittedly this is confusing and is something on my list to discuss with Jacob on Thursday. `AccessTierOptional` omits the `Premium` enum, and that more closely matches the implementation for `set_tier` in every other language SDK (not allowing `Premium`), but as you mentioned above the naming is confusing and misleading.

As to why this change specifically was made, I have no clue as I just generated against the current contents of `feature/blob-tsp` + the necessary `Access,internal` changes, but will be discussed and resolved by EOW. But at least this allows us to unblock feature parity work in the meantime.

PR #2705, opened 2025-06-13
Added if_match_etag to Item Options

gsa9989:
Added if_match_etag functionality and tests for replace item, upsert item, delete item, and patch item functions. Tests added include item_replace_if_match_etag, item_upsert_if_match_etag, item_delete_if_match_etag, and item_patch_if_match_etag. All these tests include a scenario with using if_match_etag with the correct etag and an incorrect etag. Test recordings can be found at [Azure/azure-sdk-assets @a39b424](https://github.com/Azure/azure-sdk-assets/commit/a39b424a5b73068229bebcb381fd04891a269d96)

### Links
https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/database-transactions-optimistic-concurrency#optimistic-concurrency-control


sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * Added a function `CosmosClient::with_connection_string` to enable `CosmosClient` creation via connection string. ([#2641](https://github.com/Azure/azure-sdk-for-rust/pull/2641))
 * Added support for executing limited cross-partition queries through the Gateway. See <https://learn.microsoft.com/rest/api/cosmos-db/querying-cosmosdb-resources-using-the-rest-api#queries-that-cannot-be-served-by-gateway> for more details on these limitations. ([#2577](https://github.com/Azure/azure-sdk-for-rust/pull/2577))
 * Added a preview feature (behind `preview_query_engine` feature flag) to allow the Rust SDK to integrate with an external query engine for performing cross-partition queries. ([#2577](https://github.com/Azure/azure-sdk-for-rust/pull/2577))
+* Added 'if_match_etag' to ItemOptions and necessary functions (tbd)
```
Pilchie: You should be able to put the link to this PR in there now, right?

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -35,7 +35,7 @@ impl ContainerClient {
         database_link: &ResourceLink,
         container_id: &str,
     ) -> Self {
-        let link = database_link
+        let link: ResourceLink = database_link
```
Pilchie: Why?

analogrelay: Yeah, this type annotation shouldn't be needed! The type of `link` should be apparent to the compiler from the return type of `.item`. You can just revert this back to how it was before :)

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -47,7 +47,7 @@ pub struct DeleteDatabaseOptions<'a> {
 #[derive(Clone, Default)]
 pub struct ItemOptions<'a> {
     pub method_options: ClientMethodOptions<'a>,
-
+    pub if_match_etag: Option<Etag>,
```
analogrelay, line 53: Let's add [documentation comments](https://doc.rust-lang.org/stable/book/ch14-02-publishing-to-crates-io.html?highlight=documentation#making-useful-documentation-comments) to this, I'm trying to keep us up-to-date on that as we add features :).

I'll let you take a first stab at writing it, take a look at the comments for the other fields (`method_options` doesn't have one yet, but we'll address that later) and see what you think. Happy to help if you're not sure what to write! I'd include a link to this content on our docs site as well: https://learn.microsoft.com/azure/cosmos-db/nosql/database-transactions-optimistic-concurrency#optimistic-concurrency-control , since it describes the `if-match` header in detail.

sdk/cosmos/azure_data_cosmos/README.md
```diff
@@ -131,8 +131,6 @@ async fn example(cosmos_client: CosmosClient) -> Result<(), Box<dyn std::error::
 * [Partitioning](https://learn.microsoft.com/azure/cosmos-db/partition-data)
 * [Using emulator](https://github.com/Azure/azure-documentdb-dotnet/blob/master/docs/documentdb-nosql-local-emulator.md)
 
-## Next steps
-
```
analogrelay, line 135: You'll need to undo this change. Even though the `Next Steps` section itself is empty, it has nested sections (`Provide feedback`, for example).

sdk/cosmos/azure_data_cosmos/tests/cosmos_items.rs
```diff
@@ -405,3 +407,341 @@ pub async fn item_null_partition_key(context: TestContext) -> Result<(), Box<dyn
...
+        .create_item("Partition1", &item, None)
+        .await?;
+
+    //Store Etag from response
+    const ETAG: HeaderName = HeaderName::from_static("etag");
+    let etag = response
+        .headers()
+        .get_str(&ETAG)
+        .ok()
+        .map(|s| Etag::from(s.to_string()));
```
analogrelay: There's a constant in `azure_core` you can use, and it's already a `HeaderName` 🎉 

I'd also change the `.ok()` to `.expect("expected the etag to be returned")`, which means you can rewrite it like this:

```suggestion
    let etag: Etag = response
        .headers()
        .get_str(&azure_core::http::headers::ETAG)
        .expect("expected the etag to be returned")
        .into();
```

Using `.expect` means you'll unwrap the `Option<String>` into a `String`, causing a panic if the header is missing. For product code this would be bad, but in a test it's actually preferable. It's **OK** for test code to panic when something unexpected happens, it won't break the other tests, it'll just cause this test to fail. If for some reason the service stops returning etags, we want this test to _fail_, but if we're passing the `Option<Etag>` back in to the `if_match_etag` option later, then the test might _continue to pass_ even though the behavior of the service is unexpected.

This is also a case where a type annotation (`let etag: Etag`) **is** needed because you need to tell Rust what the `.into()` at the end should return (any place you can write `SomeType::from(some_value)` you can _also_ write the same thing like this: `some_value.into()`, as long as Rust can figure out what the target type is)

sdk/cosmos/azure_data_cosmos/tests/cosmos_items.rs
```diff
@@ -405,3 +407,341 @@ pub async fn item_null_partition_key(context: TestContext) -> Result<(), Box<dyn
...
+
+    //could change response name to replaced item for better understanding
+    let response = container_client
+        .replace_item(
+            "Partition1",
+            "Item1",
+            &item,
+            Some(ItemOptions {
+                if_match_etag: etag,
+                enable_content_response_on_write: false,
```
analogrelay:
```suggestion
```

This should be the default behavior, so let's just omit it since we're not _trying_ to test this option in this test.

sdk/cosmos/azure_data_cosmos/tests/cosmos_items.rs
```diff
@@ -405,3 +407,341 @@ pub async fn item_null_partition_key(context: TestContext) -> Result<(), Box<dyn
...
+                ..Default::default()
+            }),
+        )
+        .await?;
+
+    let body = response.into_raw_body().collect_string().await?;
+    assert_eq!("", body);
```
analogrelay: Similar to the above, we're not trying to test `enable_content_response_on_write` behavior in this test, so let's just omit this check.

```suggestion
```

sdk/cosmos/azure_data_cosmos/tests/cosmos_items.rs
```diff
@@ -405,3 +407,341 @@ pub async fn item_null_partition_key(context: TestContext) -> Result<(), Box<dyn
...
+                "expected a 412 Precondition Failed error when using an incorrect ETag".into(),
+            );
+        }
+        Err(err) => {
+            assert_eq!(
+                Some(azure_core::http::StatusCode::PreconditionFailed),
+                err.http_status()
+            );
+        }
+    }
```
analogrelay: This is _good_ error handling for product code, since you're properly `match`ing on the `Result<...>` returned by `.replace_item`. But in a test, where we _know_ that the server is supposed to return an error and the test should fail if it does not return what we expect, we can simplify this quite a bit!

 ```suggestion
    assert_eq!(
        Some(azure_core::http::StatusCode::PreconditionFailed),
        response.expect_err("expected the server to return an error").http_status()
    );
```

The `.expect_err` method works a lot like `.expect`. It's just the opposite. Where `.expect` is used to unpack the successful return value from a `Result` (and panic if the `Result` represents an error), `.expect_err` unpacks the _error_ from a `Result` (and panics if the `Result` represents a successful operation).

File: sdk/cosmos/azure_data_cosmos/tests/cosmos_items.rs [hunk 9]
```diff
@@ -405,3 +407,341 @@ pub async fn item_null_partition_key(context: TestContext) -> Result<(), Box<dyn
...
+
+    account.cleanup().await?;
+    Ok(())
+}
+
+#[recorded::test]
+pub async fn item_upsert_if_match_etag(context: TestContext) -> Result<(), Box<dyn Error>> {
```
analogrelay, line 483: Most of the feedback from the above test applies to this one as well, so I won't repeat it all here :)

sdk/cosmos/azure_data_cosmos/tests/cosmos_items.rs
```diff
@@ -405,3 +407,341 @@ pub async fn item_null_partition_key(context: TestContext) -> Result<(), Box<dyn
...
+            );
+        }
+    }
+
+    account.cleanup().await?;
+    Ok(())
+}
+
+#[recorded::test]
+pub async fn item_delete_if_match_etag(context: TestContext) -> Result<(), Box<dyn Error>> {
```
analogrelay, line 552: And the same for this one

sdk/cosmos/azure_data_cosmos/tests/cosmos_items.rs
```diff
@@ -405,3 +407,341 @@ pub async fn item_null_partition_key(context: TestContext) -> Result<(), Box<dyn
...
+            );
+        }
+    }
+
+    account.cleanup().await?;
+    Ok(())
+}
+
+#[recorded::test]
+pub async fn item_patch_if_match_etag(context: TestContext) -> Result<(), Box<dyn Error>> {
```
analogrelay, line 620: And this one ;)

sdk/cosmos/azure_data_cosmos/README.md
tvaron3: This section seems to be repeated. There is a next steps section right above it with information. Is that by design?

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * Added a function `CosmosClient::with_connection_string` to enable `CosmosClient` creation via connection string. ([#2641](https://github.com/Azure/azure-sdk-for-rust/pull/2641))
 * Added support for executing limited cross-partition queries through the Gateway. See <https://learn.microsoft.com/rest/api/cosmos-db/querying-cosmosdb-resources-using-the-rest-api#queries-that-cannot-be-served-by-gateway> for more details on these limitations. ([#2577](https://github.com/Azure/azure-sdk-for-rust/pull/2577))
 * Added a preview feature (behind `preview_query_engine` feature flag) to allow the Rust SDK to integrate with an external query engine for performing cross-partition queries. ([#2577](https://github.com/Azure/azure-sdk-for-rust/pull/2577))
+* Added 'if_match_etag' to ItemOptions and necessary functions ([#2705](https://github.com/Azure/azure-sdk-for-rust/pull/2705))
```
tvaron3: I think this should go above in the unreleased section. 0.24.0 is already released. Your change would be part of the next release.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -47,7 +47,10 @@ pub struct DeleteDatabaseOptions<'a> {
 #[derive(Clone, Default)]
 pub struct ItemOptions<'a> {
     pub method_options: ClientMethodOptions<'a>,
-
+    /// IfMatchEtag is used to ensure optimistic concurrency control, it helps prevent accidental overwrites and maintains data integrity.
+    ///
+    /// https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/database-transactions-optimistic-concurrency#optimistic-concurrency-control
```
nehrao1: Looks like your PR build is failing with the same linting error I faced from adding plain URLs in comments :P

You'll likely change this comment to address Ashley's request for documentation, so when you do, putting the URL in backticks (``) should make the lint error from the build go away. Like this:

`https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/database-transactions-optimistic-concurrency#optimistic-concurrency-control`

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -47,7 +47,10 @@ pub struct DeleteDatabaseOptions<'a> {
 #[derive(Clone, Default)]
 pub struct ItemOptions<'a> {
     pub method_options: ClientMethodOptions<'a>,
-
+    /// IfMatchEtag is used to ensure optimistic concurrency control, it helps prevent accidental overwrites and maintains data integrity.
+    ///
+    /// ``https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/database-transactions-optimistic-concurrency#optimistic-concurrency-control``
+    pub if_match_etag: Option<Etag>,
```
analogrelay: A couple bits of feedback, though this is a great start! Check out some of the other fields, like `enable_content_response_on_write` below to see how we've formatted doc comments in the SDK. The `IfMatchEtag` name comes from the Go SDK and doesn't really match the name of the field in Rust. It might also be helpful to take a look at how the comments end up in the rendered documentation. You can see the _current version_ of the SDK's documentation here for an example: https://docs.rs/azure_data_cosmos/latest/azure_data_cosmos/struct.ItemOptions.html

The name of the field is already there, so we don't need to repeat it:

image

The error you were getting about putting "bare links" in the documentation is because Rust's doc comments will **not** automatically convert all URLs to links. In our case though, we _want_ to link to our documentation, to provide the user with more information about Optimistic Concurrency Control and how it works. You can do that with standard Markdown syntax. So, my suggestion would be something like this:

```suggestion
    /// If specified, the operation will only be performed if the item matches the provided Etag.
    ///
    /// See [Optimistic Concurrency Control](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/database-transactions-optimistic-concurrency#optimistic-concurrency-control) for more.
    pub if_match_etag: Option<Etag>,
```

This gives a clear indication of what setting the field *does*, which will quickly help anyone who knows roughly what they're looking for. Then, we provide a link off to our detailed documentation for those who need more detail, since there's no need for us to duplicate all that documentation here.

analogrelay: It's possible to generate the docs and open them in a browser yourself by the way, you can run `cargo doc --package azure_data_cosmos --open` from the repo root and it should build the docs and open a browser. It won't look _exactly_ like the official docs on https://docs.rs, but it'll be very close and good for reviewing how the docs will look!

sdk/cosmos/azure_data_cosmos/README.md
analogrelay: You're right! That's what I get for not expanding the diff context 🤦🏻‍♀️ . Good catch. This is fine then!

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
analogrelay: Yep, that's a good point! You get to put in the first feature of the next release @gsa9989 🎉

sdk/cosmos/assets.json
```diff
@@ -1,6 +1,6 @@
 {
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "rust",
-  "Tag": "rust/azure_data_cosmos_ff23846344",
+  "Tag": "rust/azure_data_cosmos_a39b424a5b",
```
analogrelay, line 4: @gsa9989 (and others reviewing this!) you can see the actual changes to the recordings by looking at the commit referenced by this Git Tag in the Azure/azure-sdk-assets repo:

image

Which gets you to this commit: https://github.com/Azure/azure-sdk-assets/commit/a39b424a5b73068229bebcb381fd04891a269d96

One thing I notice is that a lot of dates, RIDs, and Activity IDs changes. The tests _ignore these values_ because they can change, but we should consider redacting them so that the recording changes are less noisy. I'll look in to that.

analogrelay: The recording changes look good to me!

analogrelay (approved):
Looking good to me!

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
gsa9989: Fixed!

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 0.25.0 (Unreleased)
 
 ### Features Added
+* Added 'if_match_etag' to ItemOptions and necessary functions ([#2705](https://github.com/Azure/azure-sdk-for-rust/pull/2705))
```
kushagraThapar: necessary functions? My suggestion is to either remove the word `necessary functions` or elaborate on that :)

kushagraThapar (approved):
LGTM, thanks @gsa9989 for the changes.

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ## 0.25.0 (Unreleased)
 
 ### Features Added
-* Added 'if_match_etag' to ItemOptions and necessary functions ([#2705](https://github.com/Azure/azure-sdk-for-rust/pull/2705))
+* Added 'if_match_etag' to ItemOptions ([#2705](https://github.com/Azure/azure-sdk-for-rust/pull/2705))
```
heaths: Use appropriate markdown.
```suggestion
* Added `if_match_etag` to `ItemOptions` ([#2705](https://github.com/Azure/azure-sdk-for-rust/pull/2705))
```

heaths (approved):

PR #2712, opened 2025-06-17
Move process executor from azure_core to azure_identity

chlowell:
...because it will probably only ever be used with azure_identity


heaths:
FYI @LarryOsterman. We had discussed this and it should be fine - we can always move it back in later should we need it - but since he's working on the `AsyncRuntime` figured he should be aware.

sdk/identity/azure_identity/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### Breaking Changes
 
+- Replaced `azure_core::process::Executor` with `azure_identity::process::Executor`.
```
heaths, line 9: It was never exposed so how is it a breaking change? It's just an implementation detail.

chlowell: The type of e.g. `AzureCliCredentialOptions.executor` has changed

LarryOsterman:
QQ: Do we know why this didn't trigger an APIView change warning?

heaths (approved):

PR #2717, opened 2025-06-17
[Storage] `list_containers` for BlobServiceClient

vincenttran-msft:
This PR implements `list_containers` for `BlobServiceClient` and adds tests coverage.

### DIFF
--- a/sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
+++ b/sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
@@ -3,7 +3,10 @@
 
 use crate::{
     generated::clients::BlobServiceClient as GeneratedBlobServiceClient,
-    models::{BlobServiceClientGetPropertiesOptions, StorageServiceProperties},
+    models::{
+        BlobServiceClientGetPropertiesOptions, BlobServiceClientListContainersSegmentOptions,
+        ListContainersSegmentResponse, StorageServiceProperties,
+    },
     pipeline::StorageHeadersPolicy,
     BlobContainerClient, BlobServiceClientOptions,
 };
@@ -11,7 +14,7 @@ use azure_core::{
     credentials::TokenCredential,
     http::{
         policies::{BearerTokenCredentialPolicy, Policy},
-        Response, Url, XmlFormat,
+        PageIterator, Response, Url, XmlFormat,
     },
     Result,
 };
@@ -88,5 +91,17 @@ impl BlobServiceClient {
         options: Option<BlobServiceClientGetPropertiesOptions<'_>>,
     ) -> Result<Response<StorageServiceProperties, XmlFormat>> {
         self.client.get_properties(options).await
+    }
+
+    /// Returns a list of the containers under the specified Storage account.
+    ///
+    /// # Arguments
+    ///
+    /// * `options` - Optional configuration for the request.
+    pub fn list_containers(
+        &self,
+        options: Option<BlobServiceClientListContainersSegmentOptions<'_>>,
+    ) -> Result<PageIterator<Response<ListContainersSegmentResponse, XmlFormat>>> {
+        self.client.list_containers_segment(options)
     }
 }
--- a/sdk/storage/azure_storage_blob/src/lib.rs
+++ b/sdk/storage/azure_storage_blob/src/lib.rs
@@ -23,12 +23,13 @@ pub mod models {
         BlobContainerClientGetPropertiesOptions, BlobContainerClientGetPropertiesResult,
         BlobContainerClientGetPropertiesResultHeaders,
         BlobContainerClientListBlobFlatSegmentOptions, BlobContainerClientSetMetadataOptions,
-        BlobImmutabilityPolicyMode, BlobServiceClientGetPropertiesOptions, BlobType,
+        BlobImmutabilityPolicyMode, BlobServiceClientGetPropertiesOptions,
+        BlobServiceClientListContainersSegmentOptions, BlobType,
         BlockBlobClientCommitBlockListOptions, BlockBlobClientCommitBlockListResult,
         BlockBlobClientGetBlockListOptions, BlockBlobClientStageBlockOptions,
         BlockBlobClientStageBlockResult, BlockBlobClientUploadOptions, BlockBlobClientUploadResult,
         BlockList, BlockListType, BlockLookupList, CopyStatus, LeaseState, LeaseStatus,
-        ListBlobsFlatSegmentResponse, PublicAccessType, RehydratePriority,
-        StorageServiceProperties,
+        ListBlobsFlatSegmentResponse, ListContainersSegmentResponse, PublicAccessType,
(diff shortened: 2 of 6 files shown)


sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
```diff
@@ -89,4 +92,16 @@ impl BlobServiceClient {
...
+
+    /// Returns a list of the containers under the specified Storage account.
+    ///
+    /// # Arguments
+    ///
+    /// * `options` - Optional configuration for the request.
+    pub fn list_containers(
+        &self,
+        options: Option<BlobServiceClientListContainersSegmentOptions<'_>>,
+    ) -> Result<PageIterator<Response<ListContainersSegmentResponse, XmlFormat>>> {
```
heaths, line 104: Why are you returning a `PageIterator` instead of a `Pager` (`ItemIterator`)? Most customers will just want the list of containers, not a list of pages that they then have to *further* iterate to get container names. If they do need the individual responses, they can call `Pager::into_pages()`.

sdk/storage/azure_storage_blob/src/lib.rs
heaths, line 1: This is going to require constant maintenance. Sure you just don't want to expose `generated::modeles` directly?

```rust
pub use generated::models;
```

Or, if you have custom models:

```rust
pub mod models {
    pub use generated::models::*;
    // export additional models here
}
```

sdk/storage/azure_storage_blob/tests/blob_service_client.rs
```diff
@@ -22,3 +26,99 @@ async fn test_get_service_properties(ctx: TestContext) -> Result<(), Box<dyn Err
...
+        container_client.create_container(None).await?;
+        container_clients.push(container_client);
+    }
+
+    // Assert
+    let mut pager_response = service_client.list_containers(None)?;
+    while let Some(page) = pager_response.next().await {
+        let current_page = page.unwrap().into_body().await?;
+        let container_list = current_page.container_items;
+        for container in container_list {
```
heaths, line 53: This is exactly the double iteration I was talking about above. Why would most users care? Just return a `Pager<T>`.

sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
vincenttran-msft: This is a good callout- we also had the same observation but this is what the generated code currently supports. This matches the current implementation of [list_blobs](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs#L154-L159) (which we already released w/ this impl.), but we will have the same request and resolve the refactor in a following PR.

sdk/storage/azure_storage_blob/src/lib.rs
vincenttran-msft: It is slightly hairy to maintain, but from my understanding this is the only way currently for us to not just bulk export absolutely every single generated code model (and thus appear in the documentation). Since the API surface is quite limited, this would likely lead to more confusion (and questions) since there will be references to not-yet-implemented APIs i.e.:
Current view:
image

View if we `pub use crate::generated::models::*;`:
image
(with a huge scroll bar, probably at least 30% of the screen is not shown -- we didn't even reach Traits yet here)

sdk/storage/azure_storage_blob/tests/blob_service_client.rs
vincenttran-msft: Agreed. We are looking to get generated code support here to get us up one level of abstraction (`Pager<T>`) and cx can use this by-page workflow if necessary through `into_pages()` as you mentioned above.

heaths: Most generated code should already produce a `Pager<T>`. Does your pageable response model not follow the [REST API guidelines](https://aka.ms/azapi/guidelines) e.g.,

```json
{
  "value": [
    "foo",
    "bar"
  ],
  "nextLink": "baz"
}
```

vincenttran-msft: This is the implementation of the underlying generated code used: [list_containers_segment](https://github.com/Azure/azure-sdk-for-rust/blob/0d247492f1a855628af2e5345355cf54975a2991/sdk/storage/azure_storage_blob/src/generated/clients/blob_service_client.rs#L276-L350).
Specifically, L311 calls `PageIterator::from_callback()`. So it may just be a matter of getting the generated code to instead call `ItemIterator::from_callback()`

cc: @jhendrixMSFT Joel This is something I was planning on opening an issue to correct the behavior on both pageable operations after merging, but since Heath has pointed it now, just wanted to make you aware 😄 additional context from a previous comment below as well:

>Why are you returning a `PageIterator` instead of a `Pager` (`ItemIterator`)? Most customers will just want the list of containers, not a list of pages that they then have to *further* iterate to get container names. If they do need the individual responses, they can call `Pager::into_pages()`.

_Originally posted by @heaths in https://github.com/Azure/azure-sdk-for-rust/pull/2717#discussion_r2155526102_

heaths: As long as your pageable response model follows the form above where `nextLink` is optional, the generator should. Is this code based on the latest `main`? We shipped the last beta with `Pager<T>`.

heaths: Let me know if TypeSpec changes are necessary or this model just can't use a `Pager<T>` and I can approve/unblock.

jhendrixMSFT: This is expected since `list_containers` returns more than just a `Vec<T>` (this also came up when I added support in the emitter for the new `Pager<T>`).

https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/storage/azure_storage_blob/src/generated/models/pub_models.rs#L1018-L1047

heaths: Okay, so this is one of those. It was hard to tell from the PR.

heaths (approved):
I resolved the two open comments. Eventually all those types are going to have to be exposed (or at least most of them), so we may want to consider one-off partitioning. What do you do for Go? Might be a good analog.

As for the paging, you could later either wrap the right item array yourself if it makes sense (I can show you how to "yield" using the `async-stream` crate if you want, so the implementation is easy) or if you need to return it raw, then the current API is correct.

In any case, there's nothing blocking now for this PR.

heaths:
> > What do you do for Go?
> 
> I assume this is about reexporting types? We always reexport the complete type graph for return types.

Specifically, do you partition models or just expose them all like in the screenshots above?

Note: answering this isn't required to merge the PR. Just something we'll need to think about. We might want to have submodules under `models`, for example.

jhendrixMSFT:
Everything in the Go SDKs, with the exception of fakes, is in the root module.

PR #2719, opened 2025-06-18
Verify request body streaming

### LINKED ISSUE #2425: Reconsider `RequestContent<T>` for body models
The original goal of `RequestContent<T>` was not only to support serialization of models from different sources, but also to support a streaming model; however, I've realized after writing some clients using our crates that models with headers don't work because we only care about the serialized body content.

We have some options to consider:

1. Add a `headers` field to `RequestContent<T>` that holds an empty `HashMap`. It's non-allocating until the first element is added - or you allocate with initial capacity (don't), and could clone any headers if the `T` implements `AsHeaders`. Additionally, `RequestContent<T>` would implement `AsHeaders` if `T` implements `AsHeaders`.
2. Just get rid of `RequestContent<T>` and take a `T`. This solves the headers problem, but means that anyone with a serialized form of `T` needs to deserialize it (which `T` may not implement or derive if request-only) just to pass it in for us to re-serialize. We also have to figure out how to just take a stream, though that might be something we have to consider separately anyway e.g., `BlobClient::upload_blob()`.

heaths:
Resolves #2425 by adding a live-only example to `azure_storage_blob` so we can trace it (`RUST_LOG=trace`) and verify with `tcpdump` we are indeed tracing request and response bodies as expected.

### DIFF
--- a/sdk/storage/test-resources.bicep
+++ b/sdk/storage/test-resources.bicep
@@ -55,25 +55,6 @@ resource storage 'Microsoft.Storage/storageAccounts@2024-01-01' = {
     networkAcls: networkAcls
     supportsHttpsTrafficOnly: true
   }
-
-  resource blob 'blobServices' = {
-    name: 'default'
-    properties: {
-      isVersioningEnabled: true
-      lastAccessTimeTrackingPolicy: {
-        blobType: [
-          'blockBlob'
-        ]
-        enable: true
-        name: 'AccessTimeTracking'
-        trackingGranularityInDays: 1
-      }
-    }
-  }
 }
 
-func serviceEndpointSuffix(endpoint string) string =>
-  substring(endpoint, indexOf(endpoint, '.') + 1, length(endpoint) - (indexOf(endpoint, '.') + 2))
-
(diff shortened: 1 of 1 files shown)


heaths:
@vincenttran-msft @jalauzon-msft @analogrelay @larryosterman what do you think of this design? It's just a `From<impl SeekableStream>` (effectively) but we need to implement it on each type. Without the unstable `min_specialization` I can't really provide a good blanket implementation.

sdk/storage/test-resources.bicep
```diff
@@ -0,0 +1,79 @@
...
+  }
+  properties: {
+    accessTier: 'Hot'
+    allowSharedKeyAccess: false
+    encryption: encryption
+    networkAcls: networkAcls
+    supportsHttpsTrafficOnly: true
+  }
+
+  resource blob 'blobServices' = {
```
jalauzon-msft: Can you remove this whole block for now? We don't need/want versioning or LAT tracking on this account at this time. When we go to add those features, we will either add this back or more likely create a second, separate account like we do in other languages.

heaths: Don't I need to add a blob service so I can upload a blob? Or not?

PR #2721, opened 2025-06-18
Convert to consuming unified pool

scbedd:
This `azsdk-pool` can share allocation of agents for multiple different OS. FYI @hallipr

### DIFF
--- a/eng/pipelines/templates/variables/image.yml
+++ b/eng/pipelines/templates/variables/image.yml
@@ -2,18 +2,18 @@
 
 variables:
   - name: LINUXPOOL
-    value: azsdk-pool-mms-ubuntu-2004-general
+    value: azsdk-pool
   - name: WINDOWSPOOL
-    value: azsdk-pool-mms-win-2022-general
+    value: azsdk-pool
   - name: MACPOOL
     value: Azure Pipelines
 
   - name: LINUXVMIMAGE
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/variables/image.yml
```diff
@@ -2,18 +2,18 @@
...
   - name: WINDOWSPOOL
-    value: azsdk-pool-mms-win-2022-general
+    value: azsdk-pool
   - name: MACPOOL
     value: Azure Pipelines
 
   - name: LINUXVMIMAGE
-    value: azsdk-pool-mms-ubuntu-2004-1espt
-  - name: LINUXNEXTVMIMAGE
     value: ubuntu-22.04
```
weshaggard, line 12: Any reason to leave this one behind?

scbedd: Because they were actively testing on `ubuntu 22.04`, so I updated to include `24.04` as well, versus deleting both.

weshaggard: @heaths I assume you guys need to explicitly test on 22.04?

scbedd: They don't need it critically but would prefer to maintain it if it costs nothing. They don't use a huge amount of agents so I don't see any harm in keeping it around here.

PR #2728, opened 2025-06-23
Define API surface distributed tracing and implement OpenTelemetry tracer

LarryOsterman:
This is VERY much a work-in-progress, but rather than have a single massive PR with all the otel work, I figured I would split it into what I hope are a number of much smaller PRs.

This PR defines the structure for the distributed tracing adapter, and implements an OpenTelemetry based adapter.

The implementation defines an abstract hierarchy of types: `TracerProvider`, `Tracer`, and `Span` which roughly map to the OpenTelemetry family of APIs (and the similar hierarchy of types defined in the `tracing` crate).

It also provides a preliminary implementation of this hierarchy for OpenTelemetry, and tests those APIs using a test OTel exporter

As mentioned above, this is a WIP, but it appears that this may be sufficient to implement the distributed tracing support needed.


sdk/typespec/typespec_client_core/src/tracing/attributes.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+pub enum AttributeArray {
+    Bool(Vec<bool>),
+    I64(Vec<i64>),
+    U64(Vec<u64>),
```
lmolkova, line 7: is there a need to have signed and unsigned? over the wire they are all signed ints

sdk/typespec/typespec_client_core/src/tracing/mod.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// Adds an event to the current span.
+    ///
+    /// # Arguments
+    /// - `name`: The name of the event to add.
+    /// - `attributes`: Optional attributes to associate with the event.
+    ///
+    /// # Returns
+    /// A `Result` indicating success or failure of the operation.
+    ///
+    fn add_event(
```
lmolkova: we are not going to need span events (and they are being deprecated in otel)

sdk/core/azure_core_tracing_opentelemetry/Cargo.toml
```diff
@@ -17,6 +17,7 @@ edition.workspace = true
 azure_core.workspace = true
 log.workspace = true
 opentelemetry = { version = "0.30", features = ["trace"] }
+opentelemetry-semantic-conventions = "0.30"
```
lmolkova: do we really want to take a dependency? it's just a few constant values to copy over

Cargo.lock
```diff
@@ -1728,14 +1762,84 @@ dependencies = [
...
+ "glob",
+ "once_cell",
+ "opentelemetry 0.25.0",
+ "percent-encoding",
+ "rand 0.8.5",
+ "thiserror 1.0.69",
+]
+
+[[package]]
+name = "opentelemetry_sdk"
```
lmolkova, line 1811: is it a test-only dependency? (excuse my ignorance :) )

sdk/core/azure_core_tracing_opentelemetry/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "azure_core_tracing_opentelemetry"
```
heaths: I have some reservations on the name. If we do - and I'd like to because the rest of our stack uses it already - ship support for the `tracing` crate, then that crate is `azure_core_tracing_tracing`? Could we just elide the first "tracing" for both?

While I appreciate it builds on the "tracing" module we ship in TCC and core, we don't really do that in other places e.g., `azure_core::credential` and `azure_identity`.

sdk/core/azure_core/src/lib.rs
```diff
@@ -27,5 +27,17 @@ pub use typespec_client_core::{
...
+        AsAny, FutureExt, Span, SpanGuard, SpanKind, SpanStatus, Tracer, TracerProvider,
+        WithContext,
+    };
+    pub mod attributes {
+        pub use typespec_client_core::tracing::attributes::{
+            AttributeArray, AttributeValue, KeyValue,
+        };
+    }
+}
+
```
heaths: If this is everything - and I suspect it should be - just `pub use typespec_client_core::tracing;` As this is written, it makes us have to modify it way more often than we should have to, given that the intent is that `azure_core` is a superset of `typespec_client_core`.

sdk/core/azure_core_tracing_opentelemetry/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+```rust
+use azure_core_tracing_opentelemetry::{init_tracing, TelemetryConfig, AzureTelemetryProvider};
+
+// Initialize with default configuration
+init_tracing()?;
+
+// Or with custom configuration
+let config = TelemetryConfig::new("my-azure-app")
+    .with_service_version("1.0.0")
```
lmolkova: this is something that client library devs should provide about their lib, not the end users. End users should never interact with this API and we should document it. 

E.g. here's what we do in Java:

user-provided options: https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/clientcore/core/src/main/java/io/clientcore/core/instrumentation/InstrumentationOptions.java

library options:
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/clientcore/core/src/main/java/io/clientcore/core/instrumentation/SdkInstrumentationOptions.java

user-facing docs:
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/clientcore/core/src/samples/java/io/clientcore/core/instrumentation/TelemetryJavaDocCodeSnippets.java


library dev docs:
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/clientcore/core/src/samples/java/io/clientcore/core/instrumentation/TelemetryForLibraryDevelopersJavaDocCodeSnippets.java

only-for-client-lib comments
https://github.com/Azure/azure-sdk-for-java/blob/2f7fde3245145b6a30ab4d66eac64a99de3dfa58/sdk/clientcore/core/src/main/java/io/clientcore/core/instrumentation/tracing/Tracer.java#L10

sdk/core/azure_core_tracing_opentelemetry/README.md
```diff
@@ -0,0 +1,92 @@
...
+    Attribute {
+        key: "operation.name".to_string(),
+        value: AttributeValue::String("list_blobs".to_string()),
+    },
+]);
+
+// Advanced span creation
+let span = SpanBuilder::new("storage_list_blobs")
+    .with_kind(SpanKind::Client)
+    .with_azure_service("storage")
```
lmolkova: do we need these helpers?

sdk/core/azure_core_tracing_opentelemetry/README.md
```diff
@@ -0,0 +1,92 @@
...
+    .with_azure_operation("list_blobs")
+    .with_request_id("req-123")
+    .build();
+
+// Use the span
+let _enter = span.enter();
+// Your Azure operation here...
+span.record_success();
```
lmolkova: we should never set success, we should always leave span status unset

sdk/core/azure_core_tracing_opentelemetry/README.md
```diff
@@ -0,0 +1,92 @@
...
+### Error Handling
+
+```rust
+use azure_core::Error;
+
+let span = SpanBuilder::new("failing_operation").build();
+
+match risky_operation().await {
+    Ok(result) => {
+        span.record_success();
```
lmolkova:
```suggestion
```

sdk/core/azure_core_tracing_opentelemetry/README.md
```diff
@@ -0,0 +1,92 @@
...
+        return Err(error);
+    }
+}
+```
+
+## Azure Conventions
+
+This crate follows Azure-specific OpenTelemetry conventions:
+
+- `azure.service.name` - The Azure service being called (e.g., "storage", "keyvault")
```
lmolkova: where do these come from?

sdk/core/azure_core_tracing_opentelemetry/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+description = "OpenTelemetry tracing integration for the Azure SDK for Rust"
+readme = "README.md"
+authors.workspace = true
+license.workspace = true
+repository.workspace = true
+homepage = "https://github.com/azure/azure-sdk-for-rust"
+documentation = "https://docs.rs/azure_core_tracing_opentelemetry"
+keywords = ["sdk", "azure", "tracing", "opentelemetry", "observability"]
+categories = ["api-bindings", "development-tools::debugging"]
+
```
heaths: Was it your intent to ship this yet? If not, I would add `publish = false` until we're ready. And if you did want to ship it, make sure it's in `sdk/core/ci.yml`.

sdk/core/azure_core_tracing_opentelemetry/README.md
```diff
@@ -0,0 +1,92 @@
+# Azure Core OpenTelemetry Tracing
```
heaths: If we're going to ship this, may as well spruce up the README - in a separate PR, perhaps, and maybe @ronniegeraghty can help out based on other languages' READMEs - for a shipping crate. I think this is fine for this PR, though: important info and examples.

sdk/core/azure_core_tracing_opentelemetry/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Usage
+
+### Basic Setup
+
+```rust
+use azure_core_tracing_opentelemetry::{init_tracing, TelemetryConfig, AzureTelemetryProvider};
+
+// Initialize with default configuration
+init_tracing()?;
```
heaths: So each tracing crate has to add this (or similar) function? Given our `AsyncRuntime` design, what if we put something like this in `azure_core` and then the tracing crates just implement whatever trait - even if just a marker - to pass in there?

```rust
use azure_core_opentelemetry::TracingProvider;

azure_core::set_tracing_provider(TracingProvider);
```

heaths: And it seems that could also work with lines 29-30 below as well.

sdk/core/azure_core_tracing_opentelemetry/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+```rust
+use azure_core_tracing_opentelemetry::{create_span, SpanBuilder, SpanKind, Attribute, AttributeValue};
+
+// Simple span creation
+let span = create_span("storage_operation", vec![
+    Attribute {
+        key: "operation.name".to_string(),
+        value: AttributeValue::String("list_blobs".to_string()),
+    },
```
heaths: I appreciate this pattern might exist with otel and maybe even tracing, but could we present this as a `HashMap<String, ???>` for more idiomatic use? I especially bring this up because we've been discussing it with Storage who wanted idiomatic tags but authored their service spec as an array of objects with "key" and "value" fields as opposed to, say, a `Record<T>` (to be fair, it's a very old spec).

sdk/core/azure_core_tracing_opentelemetry/examples/basic_usage.rs
```diff
@@ -0,0 +1,5 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
heaths: I'm assuming this is a placeholder, but should still probably have the copyright headers on all source files. (Another reason I should probably hook up prettier as opposed to / in addition to editorconfig).

sdk/core/azure_core_tracing_opentelemetry/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+//! Azure Core OpenTelemetry tracing integration.
+//!
+//! This crate provides OpenTelemetry distributed tracing support for the Azure SDK for Rust.
+//! It bridges the standardized typespec_client_core tracing traits with OpenTelemetry implementation,
+//! enabling automatic span creation, context propagation, and telemetry collection for Azure services.
+
+mod attributes;
+mod conventions;
+mod span;
+pub mod telemetry;
```
heaths: Only putting this out there and honestly have no strong opinion, but should we standardize on "tracing" ubiquitously in public APIs ourselves? I appreciate this crate already has "telemetry" baked into the name, though, so I could swing either way.

sdk/typespec/typespec_client_core/src/tracing/mod.rs
```diff
@@ -0,0 +1,170 @@
...
+///
+/// Notes: There are three major traits defined here:
+/// - TracerProvider: This trait is responsible for providing tracers - this is the
+///   entrypoint for distributed tracing in the SDK.
+/// - Tracer: This trait is responsible for creating spans and managing the active span.
+/// - Span: This trait represents a single unit of work in the distributed tracing system.
+///
+pub mod attributes;
+pub mod with_context;
+pub use with_context::{FutureExt, WithContext};
```
heaths: Is there really enough in either of these to justify a **public** submodule? Doesn't seem like it. Might just flatten.

sdk/typespec/typespec_client_core/src/tracing/attributes.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+pub enum AttributeArray {
```
heaths, line 4: 1. If a public submodule - which I don't recommend - add `//! module summary`.
2. Document at least the public enums. The fields seem self-descriptive already.

sdk/typespec/typespec_client_core/src/tracing/attributes.rs
```diff
@@ -0,0 +1,21 @@
...
+}
+
+pub enum AttributeValue {
+    Bool(bool),
+    I64(i64),
+    U64(u64),
+    String(String),
+    Array(AttributeArray),
+}
+pub struct KeyValue {
```
heaths: Related to a previous comment, could we just treat these as a `HashMap<String, AttributeValue>`? Seems more idiomatic, avoids key collisions, etc.

sdk/typespec/typespec_client_core/src/tracing/with_context.rs
```diff
@@ -0,0 +1,58 @@
...
+use std::task::{Context as TaskContext, Poll};
+
+impl<T: Sized + Future> FutureExt for T {}
+
+impl<T: std::future::Future> std::future::Future for WithContext<'_, T> {
+    type Output = T::Output;
+
+    fn poll(self: Pin<&mut Self>, task_cx: &mut TaskContext<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        println!("Polling future with context: {:?}", this.context);
```
heaths: At most these `println`s should be `trace`s or something. But they are costly, so I would either 1) remove them entirely, 2) greatly reduce them, or 3) consider using spans since a few of these are basically just a separate *trace log* saying "done", which spans should do automatically.

If traces, just use the `tracing` crate. Until we decide otherwise, I think any traces *for us Azure SDK developers* should continue to use `tracing` if we don't suspect customers would get any value. `reqwest` and `hyper`, for example, rely on `tracing` and are (mostly) geared toward us, so I see no value in abstracting that away. Besides, not sure how well that would work in practice to use the very thing we're trying to build/debug if it didn't work!

heaths (changes requested):
Overall, I like it. It's a fairly clean design. I have a few design change ideas, though, to better match some of our code elsewhere so we have more of a consistent pattern e.g., like what you did with `AsyncRuntime`.

sdk/typespec/typespec_client_core/src/tracing/attributes.rs
LarryOsterman: I would ask the opentelemetry folks - I'm mirroring their API surface.

sdk/typespec/typespec_client_core/src/tracing/mod.rs
LarryOsterman: Removing then.

sdk/typespec/typespec_client_core/src/tracing/attributes.rs
lmolkova: @cijothomas do you know why we need signed and unsigned int attributes?

sdk/core/azure_core_tracing_opentelemetry/Cargo.toml
LarryOsterman: I wasn't 100% sure, to be honest. I'll remove it for now.

Cargo.lock
LarryOsterman: Yes. It comes from the [dev-dependencies] in the azure_core_tracing_opentelemetry package.

sdk/core/azure_core_tracing_opentelemetry/README.md
LarryOsterman: The readme for this is totally worthless right now. And client library devs should never have to deal with azure_core_tracing_opentelemetry - that's for sdk consumers who want to integrate otel into their applications.

sdk/core/azure_core_tracing_opentelemetry/Cargo.toml
LarryOsterman: Let's keep it as it is for now because it's consistent with the other languages.

sdk/core/azure_core_tracing_opentelemetry/README.md
LarryOsterman: Actually it's hopeless, as Liudmila pointed out. I've removed all the content for now.

sdk/core/azure_core_tracing_opentelemetry/README.md
LarryOsterman: No - everything in the readme file is AI generated garbage.

sdk/typespec/typespec_client_core/src/tracing/mod.rs
LarryOsterman: No, especially when I removed `add_event`, which removed the `KeyValue` type.

sdk/typespec/typespec_client_core/src/tracing/attributes.rs
LarryOsterman: A `KeyValue` is a key/value pair that gets added to a hashmap, but it's not specifically a hashmap. But it's moot since `KeyValue` is no longer a thing.

sdk/core/azure_core_tracing_opentelemetry/src/lib.rs
LarryOsterman: The `telemetry` module shouldn't be public - the only public type available from the OTel wrapper should be the `OpenTelemetryTracerProvider` which is exported from the root.

sdk/typespec/typespec_client_core/src/tracing/attributes.rs
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/common.rs#L157-L166
OTel has F64, I64 only.

sdk/typespec/typespec_client_core/src/tracing/mod.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+pub trait Span: AsAny {
+    /// The 8 byte value which identifies the span.
+    fn span_id(&self) -> [u8; 8];
+
+    /// Ends the current span.
+    fn end(&self) -> crate::Result<()>;
+
+    /// Adds an event to the current span.
```
cijothomas: span_events are going away from OTel, so maybe its best to *not* add them here...

sdk/core/azure_core_tracing_opentelemetry/README.md
```diff
@@ -0,0 +1,92 @@
...
+// Initialize with default configuration
+init_tracing()?;
+
+// Or with custom configuration
+let config = TelemetryConfig::new("my-azure-app")
+    .with_service_version("1.0.0")
+    .with_log_level("debug");
+
+let provider = AzureTelemetryProvider::new(config)?;
+provider.install_global()?;
```
cijothomas: what does "install_global" do?

sdk/core/azure_core_tracing_opentelemetry/src/tracer.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// Creates a new OpenTelemetry tracer with the given inner tracer.
+    pub(super) fn new(tracer: BoxedTracer) -> Self {
+        Self { inner: tracer }
+    }
+}
+
+impl Tracer for OpenTelemetryTracer {
+    fn start_span(
+        &self,
+        name: &'static str,
```
cijothomas: are all usages in Azure sdk okay to provide a `'static str`?

heaths: For *keys*, why wouldn't they be? How often is a tracing key passed as some unknown `String` at runtime? At worst, these could be `impl Into<Cow<'static, str>>` but I think that's overengineering in any cases @LarryOsterman or I could come up with. Do you have experience otherwise?

Efficiency e.g.,

```rust
use std::borrow::Cow;

fn print(s: impl Into<Cow<'static, str>>) {
    let s = s.into();
    match s {
        Cow::Owned(ref s) => eprintln!("Owned: {s:#?}"),
        Cow::Borrowed(ref s) => eprintln!("Borrowed: {s:#?}"),
    };
    
    println!("{}", s);
}

fn main() {
    print("Hello, world!");
    print(String::from("Hello, string!"));
}
```

sdk/core/azure_core/src/lib.rs
```diff
@@ -28,15 +28,7 @@ pub use typespec_client_core::{
...
-        AsAny, FutureExt, Span, SpanGuard, SpanKind, SpanStatus, Tracer, TracerProvider,
-        WithContext,
-    };
-    pub mod attributes {
-        pub use typespec_client_core::tracing::attributes::{
-            AttributeArray, AttributeValue, KeyValue,
-        };
-    }
+    pub use typespec_client_core::tracing::*;
 }
```
heaths, line 32: Tip: this whole block should just be:
```rust
pub use typespec_client_core::tracing;
```
No need to declare a `pub mod tracing` at all. Just re-export the module publicly from TCC. That's how we do it in other places. This is only necessary if you plan to merge anything.

sdk/core/azure_core_opentelemetry/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+### Basic Setup
+
+### Creating Spans
+
+### Error Handling
+
+## Azure Conventions
+
+## Integration
```
heaths, line 17: I'd just get rid of all the placeholder headings.

sdk/typespec/typespec_client_core/src/tracing/mod.rs
```diff
@@ -62,7 +64,7 @@ pub trait Tracer {
         &self,
         name: &'static str,
         kind: SpanKind,
-    ) -> Arc<dyn Span + Send + Sync>;
+    ) -> Result<Arc<dyn Span + Send + Sync>>;
```
heaths, line 67: Should these instead be "best effort"? Most APIs in `tracing`, for example, don't return `Result`s.

sdk/core/azure_core_tracing_opentelemetry/README.md
LarryOsterman: Ignore the readme - it's AI slop.

sdk/typespec/typespec_client_core/src/tracing/mod.rs
LarryOsterman: Already removed.

sdk/core/azure_core/src/lib.rs
LarryOsterman: I started with `pub use typespec_client_core_tracing` thingy and it failed miserably - that's why i added the `*`. I was concerned that a simple `pub use` would not put the relevant APIs in the `tracing` module.

sdk/typespec/typespec_client_core/src/tracing/mod.rs
LarryOsterman: It means that if you "cross the streams" accidentally (passing in the wrong kind of span), we panic rather than returning an error.

OTOH that's a clear programming error, so it may make sense to make it a panic.

sdk/core/azure_core/src/lib.rs
heaths: `pub use typespec_client_core::tracing` exports the entire module publicly. Again, this works, but is a more verbose way of doing it unless you plan to also merge other things in the future. Storage does this, for example, because they do merge in a lot of convenience layer.

sdk/core/azure_core_tracing_opentelemetry/src/tracer.rs
LarryOsterman: What Heath said. For `start_span` for example, the `name` is going to be the name of the method, which will always be a static string. 

And this is just preliminary - it's to get the shape defined. As I fill in the functionality, I'll have more opportunities to tweak this moving forward.

sdk/typespec/typespec_client_core/src/tracing/attributes.rs
LarryOsterman: Fixed.

PR #2732, opened 2025-06-24
Added sdk/storage/azure_storage_queue

### LINKED ISSUE #2708: Support storage queues
### Feature Summary

Add support for storage queues

### Feature Description

Add support for storage queues to match that for blobs.
Developers should be able to use this SDK to interact with the Storage Queues and not have to rely on the dead/unsupported azure_storage_queues crate (https://docs.rs/azure_storage_queues/latest/azure_storage_queues/)

### Use Case

As a developer I want to move my codebase from the unsupported/dead azure_storage_queues crate to the modern Rust SDK.

### Alternatives

_No response_

### Additional Context

_No response_

lsiddiquee:
Fixes #2708

Added support for the 2018-03-28 Queue API. The following functions have been created on the new QueueClient, roughly following the pattern in the .Net SDK:

## QueueServiceClient

The `QueueServiceClient` provides operations to interact with the Azure Storage Queue service at the account level.

| Method | Parameters | Return Type | Description |
|--------|------------|-------------|-------------|
| `new` | `endpoint: &str`<br>`credential: Arc<dyn TokenCredential>`<br>`options: Option<QueueServiceClientOptions>` | `Result<Self>` | Creates a new QueueServiceClient using Entra ID authentication |
| `endpoint` | `&self` | `&Url` | Returns the endpoint URL of the Azure storage account |
| `queue_client` | `&self`<br>`queue_name: String` | `QueueClient` | Returns a new QueueClient instance for a specific queue |
| `create_queue` | `&self`<br>`queue_name: &str`<br>`options: Option<QueueClientCreateOptions<'_>>` | `Result<Response<(), NoFormat>>` | Creates a new queue under the account |
| `delete_queue` | `&self`<br>`queue_name: &str`<br>`options: Option<QueueClientDeleteOptions<'_>>` | `Result<Response<(), NoFormat>>` | Permanently deletes the specified queue |
| `get_properties` | `&self`<br>`options: Option<QueueServiceClientGetPropertiesOptions<'_>>` | `Result<Response<QueueServiceProperties, XmlFormat>>` | Retrieves the properties of the queue service |
| `set_properties` | `&self`<br>`queue_service_properties: RequestContent<QueueServiceProperties, XmlFormat>`<br>`options: Option<QueueServiceClientSetPropertiesOptions<'_>>` | `Result<Response<(), NoFormat>>` | Sets the properties of the queue service |
| `list_queues` | `&self`<br>`options: Option<QueueServiceClientListQueuesOptions<'_>>` | `Result<PageIterator<Response<ListQueuesResponse, XmlFormat>>>` | Lists queues in the storage account with pagination |
| `get_statistics` | `&self`<br>`options: Option<QueueServiceClientGetStatisticsOptions<'_>>` | `Result<Response<QueueServiceStats, XmlFormat>>` | Retrieves statistics related to replication for the Queue service. It is only available on the secondary location endpoint when read-access geo-redundant replication is enabled for the storage account |

---

## QueueClient

The `QueueClient` provides operations to interact with a specific Azure Storage Queue.

| Method | Parameters | Return Type | Description |
|--------|------------|-------------|-------------|
| `new` | `endpoint: &str`<br>`queue_name: &str`<br>`credential: Arc<dyn TokenCredential>`<br>`options: Option<QueueClientOptions>` | `Result<Self>` | Creates a new QueueClient using Entra ID authentication |
| `endpoint` | `&self` | `&Url` | Returns the endpoint URL of the Azure storage account |
| `queue_name` | `&self` | `&str` | Returns the name of the queue this client is associated with |

### Queue Management

| Method | Parameters | Return Type | Description |
|--------|------------|-------------|-------------|
| `create` | `&self`<br>`options: Option<QueueClientCreateOptions<'_>>` | `Result<Response<(), NoFormat>>` | Creates a new queue |
| `delete` | `&self`<br>`options: Option<QueueClientDeleteOptions<'_>>` | `Result<Response<(), NoFormat>>` | Permanently deletes the queue |
| `exists` | `&self` | `Result<bool>` | Checks if the queue exists |
| `clear` | `&self`<br>`options: Option<QueueClientClearOptions<'_>>` | `Result<Response<(), NoFormat>>` | Clears all messages in the queue |

### Metadata Operations

| Method | Parameters | Return Type | Description |
|--------|------------|-------------|-------------|
| `set_metadata` | `&self`<br>`metadata: HashMap<String, String>`<br>`options: Option<QueueClientSetMetadataOptions<'_>>` | `Result<Response<(), NoFormat>>` | Sets the metadata for the queue |
| `get_metadata` | `&self`<br>`options: Option<QueueClientGetMetadataOptions<'_>>` | `Result<Response<QueueClientGetMetadataResult, NoFormat>>` | Retrieves the metadata of the queue |

### Message Operations

| Method | Parameters | Return Type | Description |
|--------|------------|-------------|-------------|
| `send_message` | `&self`<br>`queue_message: RequestContent<QueueMessage, XmlFormat>`<br>`options: Option<QueueClientSendMessageOptions<'_>>` | `Result<Response<SentMessage, XmlFormat>>` | Enqueues a message to the queue |
| `receive_messages` | `&self`<br>`options: Option<QueueClientReceiveMessagesOptions<'_>>` | `Result<Response<ListOfReceivedMessage, XmlFormat>>` | Retrieves multiple messages from the front of the queue |
| `peek_messages` | `&self`<br>`options: Option<QueueClientPeekMessagesOptions<'_>>` | `Result<Response<ListOfPeekedMessage, XmlFormat>>` | Peeks multiple messages without removing them |
| `delete_message` | `&self`<br>`message_id: &str`<br>`pop_receipt: &str`<br>`options: Option<QueueClientDeleteMessageOptions<'_>>` | `Result<Response<(), NoFormat>>` | Deletes a specific message from the queue |
| `update_message` | `&self`<br>`message_id: &str`<br>`pop_receipt: &str`<br>`visibility_timeout: i32`<br>`options: Option<QueueClientUpdateOptions<'_>>` | `Result<Response<(), NoFormat>>` | Updates a specific message in the queue |

Examples of how to use each method has been provided in the `sdk/storage/examples/queue_client.rs` file.

Recorded tests has also been pushed to the repo `Azure/azure-sdk-assets` and the `assets.json` updated.

The client is generated using tsp-client using remote `Azure/azure-rest-api-specs` repo.

- [ ] Merge https://github.com/Azure/azure-rest-api-specs/pull/37739 into https://github.com/Azure/azure-rest-api-specs/tree/feature/azure-queue-storage-typespec and regenerate


sdk/storage/azure_storage_queue/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# Release History
+
+## 0.1.0 (2025-06-17)
```
jalauzon-msft: Probably set this date to "Unreleased" for now.

sdk/storage/azure_storage_queue/README.md
```diff
@@ -0,0 +1,156 @@
...
+    )?;
+    Ok(())
+}
+```
+
+#### Permissions
+
+You may need to specify RBAC roles to access Queues via Microsoft Entra ID. Please see [Assign an Azure role for access to queue data] for more details.
+
+## Features
```
jalauzon-msft: I think we should probably omit this section outlining all the methods. This would be hard to maintain and should be covered by docs.

sdk/storage/azure_storage_queue/examples/README.md
```diff
@@ -0,0 +1,10 @@
+# Storage Queue Examples
```
jalauzon-msft, line 1: Please change the name of this whole folder to `samples` instead of `examples`. This just matches the pattern we follow in our other languages.

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
```diff
@@ -0,0 +1,465 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `options` - Optional parameters for the request
+    ///
+    /// # Returns
+    ///
+    /// Returns a `Result` containing the response. If the queue already exists,
+    /// returns a success response with no content (204 No Content)
+    pub async fn create_if_not_exists(
```
jalauzon-msft: After some consideration, I think maybe we should remove all the "exists" type methods for now. (with the exception of the plain `exists` method below). They cannot be generated, and we are still discussing whether we want to introduce these types of methods in Rust across all of packages. Our different languages differ in whether they are included or not.

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
```diff
@@ -0,0 +1,465 @@
...
+    /// - `Ok(true)` if the queue exists
+    /// - `Ok(false)` if the queue does not exist
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the request fails for any reason other than a non-existent queue
+    pub async fn exists(&self) -> Result<bool> {
+        match self.get_metadata(None).await {
+            Ok(_) => Ok(true),
+            Err(e) if e.http_status().unwrap() == StatusCode::NotFound => {
```
jalauzon-msft: (This exists is fine to have)
We don't want to use `unwrap` in any production code was we don't want to ever panic.

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
```diff
@@ -0,0 +1,465 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the queue doesn't exist or if the request fails
+    pub async fn clear(
+        &self,
+        options: Option<QueueClientClearOptions<'_>>,
+    ) -> Result<Response<(), NoFormat>> {
+        let result = self.client.clear(options).await?;
+        Ok(result)
```
jalauzon-msft: I think we can leave out the `?` on 212 and return the result directly like you did above?

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
```diff
@@ -0,0 +1,465 @@
...
+    /// # Returns
+    ///
+    /// Returns a `Result` containing the response if successful
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the queue doesn't exist or if the request fails
+    pub async fn set_metadata(
+        &self,
+        options: Option<QueueClientSetMetadataOptions<'_>>,
```
jalauzon-msft, line 172: I think this method would be better if metadata was a required param rather than in the options bag. Could you try changing the TypeSpec to make it required so metadata becomes a parameter?

I believe the reason it is optional is because not providing it clears the metadata but would there be a way to make the TypeSpec generate a parameter that is optional? Then providing `None` would clear the metadata.

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
```diff
@@ -0,0 +1,465 @@
...
+    /// Returns a `Result` containing the enqueued message details if successful
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the queue doesn't exist or if the request fails
+    pub async fn send_message(
+        &self,
+        queue_message: RequestContent<QueueMessage>,
+        options: Option<QueueClientSendMessageOptions<'_>>,
+    ) -> Result<Response<Option<SentMessage>, XmlFormat>> {
```
jalauzon-msft: Why does this return an `Option<SentMessage>`? Is it possible for the service to not return one?

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
```diff
@@ -0,0 +1,465 @@
...
+    ///
+    /// # Returns
+    ///
+    /// Returns a `Result` containing the dequeued message if successful. The message will be invisible
+    /// to other consumers for the duration specified in the visibility timeout
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the queue doesn't exist or if the request fails
+    pub async fn receive_message(
```
jalauzon-msft: I think for now let's omit the "single" message APIs where we call the plural one with `number_of_messages` set to 1. To me, they don't add much value as the customer could just call the plural one themselves and parse the response. I know they exist in .NET, but I'm not sure they are really needed. Also, these are methods that cannot be generated. If we get feedback that they are needed, we can consider adding them back via extension methods.

Are these things that your team would really need?

sdk/storage/azure_storage_queue/src/clients/queue_service_client.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// * `storage_service_properties` - The properties to set for the queue service
+    /// * `content_type` - The content type of the request body, typically "application/xml"
+    /// * `options` - Optional parameters for the request
+    ///
+    /// # Returns
+    ///
+    /// Returns a `Result` containing the response if successful.
+    pub async fn set_properties(
+        &self,
+        storage_service_properties: RequestContent<StorageServiceProperties>,
```
jalauzon-msft: Can we rename this model to just `QueueServiceProperties`? This would also change `get_properties`. This can be done in the TypeSpec and we will do the same for Blob. Don't make the TypeSpec change specific to Rust, fine to completely rename the model.

sdk/storage/azure_storage_queue/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+pub mod models {
+    pub use crate::generated::models::*;
+}
+
+pub mod clients {
+    mod queue_client;
+    mod queue_service_client;
+    pub use self::queue_client::QueueClient;
+    pub use self::queue_service_client::QueueServiceClient;
+    pub use crate::generated::clients::{QueueClientOptions, QueueServiceClientOptions};
```
jalauzon-msft: The clients and client options should be exported from the root of the crate, not inside a clients module.

sdk/storage/azure_storage_queue/tests/queue_client.rs
```diff
@@ -0,0 +1,653 @@
...
+
+    // Clean up by deleting the queue - this always executes
+    queue_client.delete(None).await.unwrap();
+
+    test_result
+}
+
+/// Attempts to receive the first message from an empty queue in Azure Storage Queue service.
+#[recorded::test]
+async fn test_dequeue_message_empty(ctx: TestContext) -> Result<()> {
```
jalauzon-msft: Let's rename the dequeue test to match the APi, `receive_message`.

Please note if you rename or delete tests, its best to manually remove the old recordings from the assets. Unfortunately, this does not happen automatically.

sdk/storage/azure_storage_queue/tsp-location.yaml
```diff
@@ -0,0 +1,4 @@
+directory: specification/storage/Azure.Storage.Queues
```
jalauzon-msft: Can we keep the TypeSpec in the old folder location, `Microsoft.QueueStorage`?

sdk/storage/azure_storage_queue/CHANGELOG.md
lsiddiquee: done

sdk/storage/azure_storage_queue/README.md
lsiddiquee: done

sdk/storage/azure_storage_queue/examples/README.md
lsiddiquee: Changed them. However, rust uses the examples folder to auto discover examples which can be executed using `cargo run --example queue_client`. I have now also updated the `cargo.toml` with updated paths to `sample`. In future if we do add or remove additional samples we need to remember to update the `cargo.toml`.

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
lsiddiquee: Also removing the `create_if_not_exists` and `delete_if_exists` based on this similar pattern.

For the `send_message` I am still unwrapping and returning the first message. Should I change that back to returning a list?

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
lsiddiquee: Updated this to safer check.

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
lsiddiquee: Missed it, thanks for catching it.

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
lsiddiquee: Nope we are not using them, if so we can change to using the plural version. This was added to be consistent with .NET api. Removed.

sdk/storage/azure_storage_queue/tests/queue_client.rs
lsiddiquee: done and updated the `assets.json`

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
lsiddiquee: Without this you will not have the message_id/pop_receipt. I think it is critical.

sdk/storage/azure_storage_queue/tsp-location.yaml
lsiddiquee: Done

sdk/storage/azure_storage_queue/src/clients/queue_service_client.rs
lsiddiquee: Done

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
lsiddiquee: Most likely this is because the metadata is being set though the header. I will attempt to address this in the upcoming commit.

sdk/storage/azure_storage_queue/src/lib.rs
lsiddiquee: Done

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
lsiddiquee: This is also addressed now. I have made the metadata mandatory, but in the sample you can see setting empty hashmap cleans the metadata.

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
jalauzon-msft: Thanks. Keeping `send_message` as is should be fine.

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
jalauzon-msft: I was just asking why the response type is an `Option<SentMessage>` and not just `SentMessage`. Is it possible to get a `None` back here? If the service always returns a response, we should correct this to not be optional in the TypeSpec.

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
```diff
@@ -0,0 +1,353 @@
...
+    /// Returns a `Result` containing the dequeued messages if successful. The messages will be invisible
+    /// to other consumers for the duration specified in the visibility timeout
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the queue doesn't exist or if the request fails
+    pub async fn receive_messages(
+        &self,
+        options: Option<QueueClientReceiveMessagesOptions<'_>>,
+    ) -> Result<Response<ListOfReceivedMessage, XmlFormat>> {
```
jalauzon-msft, line 298: I think we talked about this before, but can you share why there is this wrapping `ListOfReceivedMessage` model that contains just a vector? Would it be possible to return `Result<Response<Vec<ReceivedMessage>, XmlFormat>>`? I think you mentioned this would need emitter/TypeSpec changes? Also, the vector inside that model is an `Option` which also feels wrong. Instead of returning None if there are no results, it should be an empty vector. I assume the service does this in the response model but not 100%.

If we can fix this here, we should do the same for send, peek, and signed identifiers.

sdk/storage/azure_storage_queue/tests/queue_client.rs
```diff
@@ -0,0 +1,531 @@
...
+    queue_client.delete(None).await.unwrap();
+
+    test_result?;
+
+    Ok(())
+}
+
+/// Enqueues a message to the specified queue.
+#[recorded::test]
+async fn test_enqueue_message(ctx: TestContext) -> Result<()> {
```
jalauzon-msft: Sorry I missed this on the first pass, but we should rename this test to better match the new API name as well. Don't forget to remove the old tests from the assets.

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
lsiddiquee: This is not coming from the service. The issue is here is that the service returns a list of messages which we are unwrapping using `Self::extract_first_message(response, |list: &ListOfSentMessage| list.items.clone()).await`. So the following question:
1. Do we keep the response as a single sent message, which is not what we are getting from the service. It returns an xml array.
2. If we are sure that in case of success the response always contains a single element in the array we can remove the `Option` and just return the `SentMessage`.

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
lsiddiquee: I realized post our discussion it might not be as simple. Let me try to put my thoughts down:
1. `Response<Vec<ReceivedMessage>, XmlFormat>` as this is the expected type on the method, and we are using `Response` the body needs to be able to (de)serialized back and forth and xml array without a root element is not valid xml and we cannot use that as type.
2. If we do need to return a custom list object probably we can have the generator generate convenience method like into().
3. The TypeSpec definition could be something like following, however, need to validate that this is acceptable. `model ListOfSignedIdentifier is Array<SignedIdentifier>;`. But we also need to ensure that any `Xml.name` ie `renaming` is applied and copied in the typespec compiler, so that we can apply that.
4. Other option in TypeSpec we add a new decorator to say this should be unwrapped.

In the typespec nested `items` property is not optional. However, looking at the [generator](https://github.com/Azure/typespec-rust/blob/ce26a6ca3b905a88d1a279fb29818bacbc8ef747/packages/typespec-rust/src/tcgcadapter/adapter.ts#L329) it seems that all fields are being treated as optional regardless what the spec defines.

sdk/storage/azure_storage_queue/tests/queue_client.rs
lsiddiquee: Done

sdk/storage/azure_storage_queue/src/clients/queue_service_client.rs
```diff
@@ -0,0 +1,204 @@
...
+    ///
+    /// * `options` - Optional parameters for the request.
+    ///
+    /// # Returns
+    ///
+    /// Returns a `Result` containing the service statistics response if successful.
+    pub async fn get_statistics(
+        &self,
+        options: Option<QueueServiceClientGetStatisticsOptions<'_>>,
+    ) -> Result<Response<StorageServiceStats, XmlFormat>> {
```
lsiddiquee: @jalauzon-msft: Should this `StorageServiceStats` be also renamed to `QueueServiceStats` like the earlier `StorageServiceProperties` to `QueueServiceProperties`?

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
jalauzon-msft: I think ideally: 
1. We keep it a single message.
2. I'm not sure if a success case always contains a single element or not. In your testing did you ever encounter one that did not contain a message? Maybe we can remove the Option but have the helper method return an error result if there isn't one there. Then during testing/beta we can see if anyone hits an issue with this.

sdk/storage/azure_storage_queue/src/clients/queue_service_client.rs
jalauzon-msft: Sure, that sounds good.

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
lsiddiquee: Added an issue in typespec/rust https://github.com/Azure/typespec-rust/issues/556.

sdk/storage/azure_storage_queue/README.md
```diff
@@ -0,0 +1,98 @@
+# Azure Queue client library for Rust
+
+Azure Queue Storage is a service for storing large numbers of messages.
```
heaths, line 3: Nit: some disagreement on branding here. Is it "Azure Queue" or "Azure Queue Storage"? I thought it was actually "Azure Storage Queue".

sdk/storage/azure_storage_queue/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+categories = ["api-bindings"]
+
+# Manual example definitions pointing to samples/ directory
+[[example]]
+name = "queue_client"
+path = "samples/queue_client.rs"
+
+[[example]]
+name = "queue_service_client"
+path = "samples/queue_service_client.rs"
```
heaths: That's not idiomatic. Put them in the "examples/" directory and get rid of these entries. It's just extra maintenance for no reason.

sdk/storage/azure_storage_queue/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+
+[features]
+default = ["reqwest"]
+reqwest = ["dep:reqwest"]
```
heaths: Why are you taking a direct dependency on `reqwest`? This is highly sus. So long as you're HTTP-based, you must use the types in `azure_core` only so that we have consistent HTTP client override support across all HTTP-based crates: same as any other Azure SDK language.

blocker

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
```diff
@@ -0,0 +1,353 @@
...
+    credentials::TokenCredential,
+    http::{NoFormat, RawResponse, RequestContent, Response, StatusCode, Url, XmlFormat},
+    xml, Bytes, Result,
+};
+use std::{collections::HashMap, sync::Arc};
+
+/// A client to interact with a specific Azure storage queue, although that queue may not yet exist.
+pub struct QueueClient {
+    pub(super) endpoint: Url,
+    pub(super) client: GeneratedQueueClient,
```
heaths, line 17: Is wrapping the generated client necessary? I know why Blob is doing it - to "smooth" out the DX - but is that true for Queue as well? Ideally, we're just generating.

sdk/storage/azure_storage_queue/src/lib.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+//
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// Code generated by Microsoft (R) Rust Code Generator.
+
+#![doc = include_str!("../README.md")]
+#![allow(dead_code)]
+#![allow(unused_imports)]
```
heaths: These are code smells in a lib crate. If you have some code that is surfacing as public in the lib crate or unused imports, you need to fix them. `dead_code` indicates something isn't used and isn't public. If this is model types, then add `#[allow(dead_code)]` to that module. Don't do a whole-lib suppression.

sdk/storage/azure_storage_queue/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+
+#[expect(deprecated, reason = "requires emitter update")]
+mod generated;
+
+pub mod models {
```
heaths, line 14: Public modules must be documented.

sdk/storage/azure_storage_queue/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+    pub use crate::generated::models::*;
+}
+
+mod clients;
+pub use crate::clients::QueueClient;
+pub use crate::clients::QueueServiceClient;
+pub use crate::generated::clients::{QueueClientOptions, QueueServiceClientOptions};
```
heaths: Guidelines are that you must make this public and then expose your wrapped clients from there. Instead, do something like:

```suggestion
//! Necessary documentation
pub mod clients;
```

Thern in `clients/mod.rs`, just define the types as `pub struct` and re-export the queue service client from there.

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
```diff
@@ -0,0 +1,353 @@
...
+    /// # Returns
+    ///
+    /// Returns a `Result` containing the response if successful
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the queue doesn't exist or if the request fails
+    pub async fn set_metadata(
+        &self,
+        metadata: HashMap<String, String>,
```
heaths, line 171: Body parameters - if this is a body parameter - must be `RequestContent<T>`.

sdk/storage/azure_storage_queue/samples/helpers/mod.rs
heaths, line 1: This raises the question of who these samples are for. Are they full-blown EXEs customers can use - like what Cosmos has - or more for copy and pastability? If the latter, a bunch of helpers not within a single source isn't going to be very useful.

sdk/storage/azure_storage_queue/README.md
jalauzon-msft: Its "Azure Queue Storage" - https://learn.microsoft.com/en-us/azure/storage/queues/storage-queues-introduction

sdk/storage/azure_storage_queue/Cargo.toml
jalauzon-msft: I actually had him change this since we use "samples" in all other languages I thought it would be best to be consistent. If examples if preferred for Rust, so be it.

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
jalauzon-msft: This does not go into a body, metadata goes into headers.
https://learn.microsoft.com/en-us/rest/api/storageservices/set-queue-metadata

sdk/storage/azure_storage_queue/src/clients/queue_client.rs
jalauzon-msft: We want to go fully generated for GA but like Blob, we are starting with a wrapping client as there were some things that precluded us from being fully generated. Lots of things have changed over the course of this PR and we may be closer to fully generated than originally anticipated but we will re-evaluate after the initial preview and before GA.

sdk/storage/azure_storage_queue/Cargo.toml
heaths: That's not idiomatic to rust. Idiomatic comes before consistent.

lsiddiquee: Moved back.

sdk/storage/azure_storage_queue/Cargo.toml
lsiddiquee: Cleaned up all dependencies.

sdk/storage/azure_storage_queue/src/lib.rs
lsiddiquee: Removed

sdk/storage/azure_storage_queue/src/lib.rs
lsiddiquee: Documentation added

sdk/storage/azure_storage_queue/src/lib.rs
lsiddiquee: Fixed module exports

sdk/storage/azure_storage_queue/samples/helpers/mod.rs
lsiddiquee: The intention is to show how to invoke, pass params and deal with responses. I guess that falls in the copy-paste example. Hence, leaned them and the methods are now duplicated in the two samples.

sdk/storage/azure_storage_queue/src/clients/mod.rs
```diff
@@ -0,0 +1,7 @@
+pub mod queue_client;
+pub use queue_client::QueueClient;
+
+pub mod queue_service_client;
+pub use queue_service_client::QueueServiceClient;
+
+pub use crate::generated::clients::{QueueClientOptions, QueueServiceClientOptions};
```
heaths: Clients' modules shouldn't be public:

```suggestion
mod queue_client;
pub use queue_client::QueueClient;

mod queue_service_client;
pub use queue_service_client::QueueServiceClient;

pub use crate::generated::clients::{QueueClientOptions, QueueServiceClientOptions};
```

sdk/storage/azure_storage_queue/src/clients/mod.rs
```diff
@@ -0,0 +1,7 @@
+mod queue_client;
```
heaths, line 4: Copyright headers on all source files. Microsoft policy.

sdk/storage/azure_storage_queue/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+keywords = ["sdk", "azure", "storage", "queue", "queues"]
+categories = ["api-bindings"]
+
+[features]
+default = ["azure_core/default"]
+
+[dependencies]
+azure_core = { workspace = true, features = ["xml"] }
+serde = { workspace = true }
+typespec_client_core = { workspace = true, features = ["derive"] }
```
heaths: You shouldn't depend on this directly. Everything in it is in `azure_core`.

heaths:
@jalauzon-msft I explained: it's not the tests. It's that your TypeSpec incorrectly says some operations like DELETEs return 202 when they actually (correctly) return 204s. The tests are only failing because we now strictly check (as do all/most other Azure SDK languages) and that is failing because the recordings - as recorded from your service - return 204 but we're checking agianst 202 becuase that's what your TypeSpecs say. @LarryOsterman compared your authored swaggers vs. your generated-from-TypeSpec swaggers and found at least 2 places where you incorrectly spec that you return 202 in TypeSpec. You need to fix the TypeSpec, update `tsp-location.yaml`, and regenerate.

sdk/storage/azure_storage_queue/assets.json
```diff
@@ -1,6 +1,6 @@
 {
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "rust",
-  "Tag": "rust/azure_storage_queue_5f6f02d3be",
+  "Tag": "rust/azure_storage_queue_cda6f79f01",
```
heaths: Why did you re-record? It would've been better that you didn't as a way of regression testing your TypeSpec changes. Or was something else wrong once you got past the incorrect status codes?

sdk/storage/azure_storage_queue/src/generated/clients/queue_client.rs
```diff
@@ -350,7 +350,9 @@ impl QueueClient {
     /// async fn example() -> Result<()> {
     ///     let response: Response<QueueClientGetMetadataResult, NoFormat> = unimplemented!();
     ///     // Access response headers
-    ///     println!("x-ms-meta: {:?}", response.metadata()?);
+    ///     if let Some(metadata) = response.metadata()? {
```
heaths: This will fail. You need to regenerate based on `main` now, which means you need to sync. We just merged the emitter change about 30 minutes ago that fixes this.

heaths (changes requested):

sdk/storage/azure_storage_queue/assets.json
heaths: > [Fix TypeSpec, re-record test for sanity](https://github.com/Azure/azure-sdk-for-rust/pull/2732/commits/0ba01091f52b0cb8c8101f3e4574a0507e5775fe)

It would be saner to _not_ re-record so that you're effectively regression testing against what the server returned, which shouldn't have changed. Unless of course there were other matcher issues given emitter changes we've made for Storage, but I wouldn't suspect so. I'd say: use the previous recordings and only re-record if necessary for other _expected_ reasons.

vincenttran-msft: I can manually point it to the old recordings after I merge `main` in so we can use it as a regression test. I thumbed through the recordings and didn't see anything else to note- but usually just re-record for sanity when hitting all these CI issues and making code changes.

heaths:
Thanks, @vincenttran-msft!

lsiddiquee:
Thanks @vincenttran-msft for addressing the issues and @heaths for the assist in closing this up.

PR #2733, opened 2025-06-24
Fixed test broker build

LarryOsterman:
This fixes a CI pipeline failure building core.

### DIFF
--- a/sdk/core/azure_core_amqp/Test-Setup.ps1
+++ b/sdk/core/azure_core_amqp/Test-Setup.ps1
@@ -33,7 +33,7 @@ try {
 
   Set-Location -Path "./azure-amqp/test/TestAmqpBroker"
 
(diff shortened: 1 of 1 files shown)


sdk/core/azure_core_amqp/Test-Setup.ps1
```diff
@@ -33,7 +33,7 @@ try {
 
   Set-Location -Path "./azure-amqp/test/TestAmqpBroker"
 
-  Invoke-LoggedCommand "dotnet build -p RollForward=LatestMajor --framework net6.0"
```
heaths, line 36: Why specify a version at all? Normally not necessary. We should be able to just `dotnet run`. Is the `./azure-amqp` a repo working dir? If they have their own `global.json` in the root, as long as the runner has that version everything should just work and we don't have to keep chasing this issue.

PR #2735, opened 2025-06-24
[EngSys] Update agent pool to "azsdk-pool"

mikeharder:


eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -77,8 +77,8 @@ stages:
           timeoutInMinutes: 120
           dependsOn: TagRepository
           pool:
-            name: azsdk-pool-mms-ubuntu-2004-general
-            image: azsdk-pool-mms-ubuntu-2004-1espt
+            name: azsdk-pool
+            image: ubuntu-24.04
```
heaths, line 81: We should publish on the minimal supported OS just in case.
```suggestion
            image: ubuntu-22.04
```

heaths (changes requested):

eng/pipelines/templates/stages/archetype-rust-release.yml
mikeharder: We are trying to run everything on ubuntu-24 by default, unless you need ubuntu-22 for a specific reason.

heaths: Publishing actually builds the crate, and better to build on the baseline.

weshaggard: @heaths are we publishing the build binary? In general for the deployment jobs we are not supposed to do anything but actually upload the already build bits to the package manager.

heaths: Rust doesn't ship binaries. It's like npm. It's a source archive, but it still "compiles" in a way that I cannot say for sure doesn't depend on the base image. That said, it does seem the source outside the diff (when I expand) is using a previously-built archive, so I'll approve.

heaths (approved):

PR #2737, opened 2025-06-25
Define distinct ClientOptions in azure_core

### LINKED ISSUE #1753: Make TelemetryOptions optional and generic, but default to the standard Azure User-Agent format for azure_core

heaths:
This sets us up to have Azure-specific client options like UserAgentOptions nee TelemetryOptions. Also allows ability to disable sending the User-Agent like in the Go SDK.

Fixes #1753

### DIFF
--- a/sdk/core/azure_core/src/http/pipeline.rs
+++ b/sdk/core/azure_core/src/http/pipeline.rs
@@ -2,13 +2,16 @@
 // Licensed under the MIT License.
 
 use super::policies::ClientRequestIdPolicy;
-use crate::http::{policies::UserAgentPolicy, ClientOptions};
+use crate::http::{
+    policies::{Policy, UserAgentPolicy},
+    ClientOptions,
+};
 use std::{
     any::{Any, TypeId},
     ops::Deref,
     sync::Arc,
 };
-use typespec_client_core::http::{self, policies::Policy};
+use typespec_client_core::http;
 
 /// Execution pipeline.
 ///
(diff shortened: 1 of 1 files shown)


sdk/core/azure_core/src/http/pipeline.rs
```diff
@@ -45,35 +44,34 @@ impl Pipeline {
...
 
-        if per_call_policies
-            .iter()
-            .all(|policy| TypeId::of::<ClientRequestIdPolicy>() != policy.type_id())
-        {
-            per_call_policies.push(Arc::new(ClientRequestIdPolicy::default()));
+        let (user_agent, options) = options.deconstruct();
+        if !user_agent.disabled {
+            let telemetry_policy = UserAgentPolicy::new(crate_name, crate_version, &user_agent);
+            push_unique(&mut per_call_policies, telemetry_policy);
```
LarryOsterman, line 57: Since `push_unique` wraps its `policy` parameter in an `Arc`, it means that `Policy::new` should never return an `Arc<Policy>`.

This is a different model from the one used by `TokenCredential<T>`, so it's important to make sure that our Policy implementations follow this rule.

heaths: Most policies don't actually require construction, and the reason we do that for `TokenCredential`s is to encourage sharing for efficiency. I don't think that's true for policies.

That said, we do support it which is why we take a `Vec<Arc<dyn Policy>>`, but in this case where we're constructing something truly unique.

That said, since the `push_unique` function is internal, we could always change it later to take an `Arc<dyn Policy>` instead of an `impl Policy` and wrap it. Just didn't seem necessary now.

PR #2741, opened 2025-06-30
Distributed Tracing integration with SDK builds. 

LarryOsterman:
Added distributed tracing policy and example service client integration.

PR Changes:
1) Implemented `RequestInstrumentationPolicy` to add spans for HTTP operations.
1) Implemented `PublicApiInstrumentationPolicy` to add spans for public APIs.
1) Added public struct `PublicApiInstrumentationInformation` which expresses information used to create Public API spans.
2) Integrated `RequestInstrumentationPolicy` and `PublicApiInstrumentationPolicy` to `azure_core` pipeline creation.
    1) Changed the `azure_core::ClientOptions::deconstruct()` method to deconstruct to a `typespec_client_core::ClientOptions` and a `azure_core::CoreClientOptions` (which is crate-private).
    1) Added `http::Method::as_str()` to return a string slice with the name of the HTTP method.
1) Added span support for header propagation.
1) PR feedback from previous PR (removed `Span::add_event`, removed `AttributeValue::U64` and replaced it with `AttributeValue::F64`), 

Bug fix: Propagate retry policy options from client options to retry policy.


sdk/core/azure_core/src/http/policies/request_instrumentation.rs
```diff
@@ -0,0 +1,716 @@
...
+    tracing::{Span, SpanKind},
+};
+use std::sync::Arc;
+use typespec_client_core::{
+    http::policies::{Policy, PolicyResult, RetryPolicyCount},
+    tracing::Attribute,
+};
+
+const AZ_NAMESPACE_ATTRIBUTE: &str = "az.namespace";
+const AZ_SCHEMA_URL_ATTRIBUTE: &str = "az.schema.url";
```
lmolkova: we don't need this on Rust - there is a dedicated property on the instrumentation scope (which we pass when we create a tracer) - https://github.com/open-telemetry/opentelemetry-rust/blob/d59aded375ac82b0a6a99521a1a7973976762469/opentelemetry/src/common.rs#L474

We use this attribute on .NET where the BCL doesn't provide dedicated property - it will in ,NET 10 - https://github.com/dotnet/runtime/issues/63651

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
```diff
@@ -0,0 +1,716 @@
...
+const AZ_CLIENT_REQUEST_ID_ATTRIBUTE: &str = "az.client.request.id";
+const ERROR_TYPE_ATTRIBUTE: &str = "error.type";
+const AZ_SERVICE_REQUEST_ID_ATTRIBUTE: &str = "az.service.request.id";
```
lmolkova:
```suggestion
const AZ_SERVICE_REQUEST_ID_ATTRIBUTE: &str = "az.service_request_id";
```

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
```diff
@@ -0,0 +1,716 @@
...
+const HTTP_REQUEST_RESEND_COUNT_ATTRIBUTE: &str = "http.request.resend.count";
```
lmolkova:
```suggestion
const HTTP_REQUEST_RESEND_COUNT_ATTRIBUTE: &str = "http.request.resend_count";
```

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
```diff
@@ -0,0 +1,716 @@
...
+            if let Some(namespace) = tracer.namespace() {
+                // If the tracer has a namespace, we set it as an attribute.
+                span_attributes.push(Attribute {
+                    key: AZ_NAMESPACE_ATTRIBUTE,
+                    value: namespace.into(),
+                });
+            }
+
+            if !request.url().username().is_empty() || request.url().password().is_some() {
+                // If the URL contains a password, we do not log it for security reasons.
```
lmolkova: We can still log the rest of URL, just omit user name and password (which we should never have on azure anyway - feel free to ignore my nitpicking)

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
```diff
@@ -0,0 +1,716 @@
...
+                        .map_or_else(|| "unknown_host".to_string(), |h| h.to_string()),
+                    request
+                        .url()
+                        .port()
+                        .map_or_else(String::new, |p| format!(":{}", p)),
+                    request.url().path(),
+                    request
+                        .url()
+                        .query()
+                        .map_or_else(String::new, |q| format!("?{}", q)),
```
lmolkova: is it sanitized?

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
```diff
@@ -0,0 +1,716 @@
...
+                    request
+                        .url()
+                        .fragment()
+                        .map_or_else(String::new, |f| format!("#{}", f)),
```
lmolkova: do we actually have fragments?

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
```diff
@@ -0,0 +1,716 @@
...
+                    SpanKind::Client,
+                    span_attributes,
+                    parent_span.clone(),
+                )
+            } else {
+                // If no parent span exists, start a new span without a parent.
+                tracer.start_span_with_current(method_str, SpanKind::Client, span_attributes)
+            };
+
+            if let Some(client_request_id) = request
```
lmolkova: it's probably a good idea to check if span is recording here and avoid adding any attributes if not (for perf reasons) - https://github.com/open-telemetry/opentelemetry-rust/blob/d59aded375ac82b0a6a99521a1a7973976762469/opentelemetry/src/trace/span.rs#L108

lmolkova (approved):
Left some cosmetic comments, look great in general!

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
LarryOsterman: Is that true for both `az.namespace` as well as `az.schema.url`?

LarryOsterman: And what happened here was that I confused `url.scheme` with `az.schema.url` - the content was filled in with `url.scheme` but I named it `az.schema.url`.

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
LarryOsterman: I'm hoping that my code *does* emit the full URL without the user name and password - there's a test case for this at the bottom of the file.

And I know it should never happen for Azure, but the user controls the endpoint being passed into the SDK - and all of the languages (except C++, which explicitly doesn't support username and passwords) will happily pass the username and password to the service. So it's theoretically possible we will get a URL with a username and password in it.

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
LarryOsterman: Oh, and a silly question: Why *don't* we require `url.scheme` when we require `server.port` and `server.address`, which are also in the URL?

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
LarryOsterman: Unfortunately, we don't have a URL sanitizer for Rust yet :( - we're currently depending on the proxy to sanitize our recordings, but we need to have one.

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
LarryOsterman: Probably not, but I don't believe they're explicitly blocked.

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
LarryOsterman: Oh, and this is an interesting issue - we *do* need `az.schema.url` for `telemetry` spans and tracers since the `telemetry` package doesn't have a way of specifying the schema url. 

Oh well, I'll deal with that when I have to deal with that.

Cargo.lock
```diff
@@ -1,6 +1,6 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 3
+version = 4
```
heaths: Wasn't expecting this, but it still builds fine under the MSRV `1.85` (both 1.85.0 and 1.85.1 I tried).

sdk/typespec/typespec_client_core/src/http/method.rs
```diff
@@ -110,6 +110,18 @@ impl Method {
     pub fn is_safe(&self) -> bool {
         matches!(self, Method::Get | Method::Head)
     }
+
+    /// Returns the HTTP method as a static string slice.
+    pub fn as_str(&self) -> &'static str {
```
heaths: May as well declare as a `const` function which can be used in more places. Several things in `std` do this.

```suggestion
    pub const fn as_str(&self) -> &'static str {
```

sdk/typespec/typespec_client_core/src/http/policies/retry/mod.rs
```diff
@@ -68,6 +68,11 @@ pub fn get_retry_after(headers: &Headers, now: DateTimeFn) -> Option<Duration> {
         })
 }
 
+/// A wrapper around a retry count to be used in the context of a retry policy.
+///
+/// This allows a post-retry policy to access the retry count
+pub struct RetryPolicyCount(pub u32);
```
heaths: Is this just a type so we can stuff it in the `Context`? Might instead consider keeping the `u32` private and impl `Deref<Target = u32>`; though, mutability will help protect it either way. Just becomes less of an implementation detail that way, IMO.

sdk/typespec/typespec_client_core/src/http/policies/retry/mod.rs
```diff
@@ -131,7 +136,8 @@ where
                     "failed to reset body stream before retrying request",
                 )?;
             }
-            let result = next[0].send(ctx, request, &next[1..]).await;
+            let try_context = ctx.clone().with_value(RetryPolicyCount(retry_count));
```
heaths: Nit: any future change might still take just `ctx`. Any reason to just not `let ctx = ctx.clone()...` and shadow `ctx` to avoid subsequent future code from using the wrong one? Or is the idea really that you want to keep both accessible?

sdk/typespec/typespec_client_core/src/tracing/attributes.rs
```diff
@@ -1,17 +1,127 @@
...
+    /// A 64-bit floating point attribute value
+    F64(f64),
+    /// A string attribute value.
     String(String),
+    /// An array of attribute values.
     Array(AttributeArray),
 }
+
+#[derive(Debug, PartialEq)]
+pub struct Attribute {
```
heaths, line 42: Doc comment.

sdk/typespec/typespec_client_core/src/tracing/mod.rs
```diff
@@ -19,77 +18,124 @@ use std::sync::Arc;
...
-    ) -> Box<dyn Tracer + Send + Sync>;
+    ) -> Arc<dyn Tracer>;
 }
 
-pub trait Tracer {
+impl Debug for dyn TracerProvider {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("TracerProvider").finish_non_exhaustive()
+    }
+}
```
heaths: This isn't typical. Could we just make `Debug` a supertrait of `TraceProvider` and let individual `TraceProviders` decide how/what to emit? This what we do for policies and other behavioral traits currently.

heaths: I see this pattern below as well.

sdk/typespec/typespec_client_core/src/tracing/mod.rs
```diff
@@ -19,77 +18,124 @@ use std::sync::Arc;
...
+#[derive(Debug, PartialEq)]
 pub enum SpanStatus {
     Unset,
-    Ok,
     Error { description: String },
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug, Default, PartialEq)]
 pub enum SpanKind {
```
heaths: Should also derive `Eq`.

sdk/typespec/typespec_client_core/src/tracing/mod.rs
```diff
@@ -118,14 +170,15 @@ pub trait Span: AsAny {
     /// # Returns
     /// A `Result` indicating success or failure of the operation.
     ///
-    fn set_status(&self, status: SpanStatus) -> crate::Result<()>;
+    fn set_status(&self, status: SpanStatus);
```
heaths, line 141: This doesn't have to be `&mut self` to mutate? Or is `set_status` not actually mutating the implementation?

sdk/core/azure_core/src/lib.rs
```diff
@@ -28,6 +28,7 @@ pub use typespec_client_core::{
 };
 
 pub mod tracing {
```
heaths, line 31: Nit: brief module doc comment e.g.,
```suggestion
/// Abstractions to trace client methods and HTTP requests.
pub mod tracing {
```

sdk/core/azure_core/src/http/policies/public_api_instrumentation.rs
```diff
@@ -0,0 +1,464 @@
...
+    /// the API name should be `MyClient.my_api`.
+    pub api_name: &'static str,
+
+    /// Additional attributes to be added to the span for this API.
+    ///
+    /// These attributes can provide additional information about the API being instrumented.
+    /// See [Library-specific attributes](https://github.com/Azure/azure-sdk/blob/main/docs/tracing/distributed-tracing-conventions.md#library-specific-attributes)
+    /// for more information.
+    ///
+    pub attributes: Vec<Attribute>,
```
heaths: Seems this could contain PII if convenience clients are writing to it, anyway. Might want to derive `SafeDebug` and add `#[safe(true)]` to `api_name` while leaving this field "off".

sdk/core/azure_core/src/http/policies/public_api_instrumentation.rs
```diff
@@ -0,0 +1,464 @@
...
+                    SpanKind::Internal,
+                    span_attributes,
+                ))
+            } else {
+                // If no tracer is available, we skip instrumentation.
+                None
+            }
+        } else {
+            None
+        };
```
heaths: Tip: could just use `Option<T>::map` e.g.,

```rust
let span = public_api_information.map(|info| {
  // exiting code that references `info`
};
```

This is fine, too. Though. Again, just a tip.

sdk/core/azure_core/src/http/policies/public_api_instrumentation.rs
```diff
@@ -0,0 +1,464 @@
...
+                        }
+                    }
+                    _ => {
+                        span.set_attribute(ERROR_TYPE_ATTRIBUTE, e.kind().to_string().into());
+                        span.set_status(crate::tracing::SpanStatus::Error {
+                            description: e.kind().to_string(),
+                        });
+                    }
+                }
+            } else if let Ok(response) = &result {
```
heaths: Nit: more idiomatic - though, again, not wrong - to use a `match` here.

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
```diff
@@ -0,0 +1,744 @@
...
+};
+use std::sync::Arc;
+use typespec_client_core::{
+    http::policies::{Policy, PolicyResult, RetryPolicyCount},
+    tracing::Attribute,
+};
+
+const AZ_NAMESPACE_ATTRIBUTE: &str = "az.namespace";
+const AZ_CLIENT_REQUEST_ID_ATTRIBUTE: &str = "az.client.request.id";
+const ERROR_TYPE_ATTRIBUTE: &str = "error.type";
```
heaths: A couple of these are shared. Just define them as-is in the parent `mod.rs` and they are visible to submodules already (i.e. don't need a `pub`). Just to ease maintenance.

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
```diff
@@ -0,0 +1,744 @@
...
+                span_attributes.push(Attribute {
+                    key: AZ_NAMESPACE_ATTRIBUTE,
+                    value: namespace.into(),
+                });
+            }
+
+            if !request.url().username().is_empty() || request.url().password().is_some() {
+                // If the URL contains a password, we do not log it for security reasons.
+                let full_url = format!(
+                    "{}://{}{}{}{}{}",
```
heaths: That's a lot of string manipulation. What exactly are we trying to do here? `Url` already supports formatting as a string. Is it just eliminate any possible `username:password` that the user explicitly themselves put into the URL? I thought we said that wasn't our problem. Seems a lot of work here on every request to mitigate something the user did (and probably never would) in spite of every other caller who did the right thing.

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
```diff
@@ -0,0 +1,744 @@
...
+    ) -> PolicyResult {
+        // If the context has a tracer (which happens when called from an instrumented method),
+        // we prefer the tracer from the context.
+        // Otherwise, we use the tracer from the policy itself.
+        // This allows for flexibility in using different tracers in different contexts.
+        let tracer = if ctx.value::<Arc<dyn crate::tracing::Tracer>>().is_some() {
+            ctx.value::<Arc<dyn crate::tracing::Tracer>>()
+        } else {
+            self.tracer.as_ref()
+        };
```
heaths: Per doc comments on the struct, shouldn't we check if `tracer` is `None` and just return the result of the next policy?

heaths: Oh, I see. It's below the check for a span. In that case, Rather than more nesting below (it's getting deep) and to avoid the unnecessary hash check for a span, just add here:

```rust
let Some(tracer) = tracer else {
    // If no tracer is set, we simply forward the request without instrumentation.
    return next[0].send(ctx, request, &next[1..]).await;
};
```

Early exit, and your `tracer` is now a `dyn TraceProvider` (or whatever) and you don't need the extra nested `if let Some(tracer) = ...` below.

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
```diff
@@ -0,0 +1,744 @@
...
+        }
+    }
+
+    #[derive(Debug)]
+    pub(crate) struct MockTracer {
+        pub(crate) namespace: Option<&'static str>,
+        pub(crate) package_name: &'static str,
+        pub(crate) package_version: &'static str,
+        pub(crate) spans: Mutex<Vec<Arc<MockSpan>>>,
+    }
```
heaths: Nit: actually, on the `tests` module as well: why is everything `pub(crate)`? If you do actually use it elsewhere, maybe define it with a `cfg(test)` in the parent `mod.rs` so it's available to every child module already.

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
```diff
@@ -0,0 +1,744 @@
...
+                    (
+                        URL_FULL_ATTRIBUTE,
+                        AttributeValue::from("https://host:8080/path?query=value#fragment"),
+                    ),
+                ],
+            },
+        );
+    }
+
+    #[tokio::test]
```
heaths: If possible, also a test for no tracer i.e. that we didn't even try to do more computation than neccessary?

sdk/core/azure_core/src/http/options/mod.rs
```diff
@@ -27,6 +29,13 @@ pub struct ClientOptions {
 
     /// User-Agent telemetry options.
     pub user_agent: Option<UserAgentOptions>,
+
+    pub request_instrumentation: Option<RequestInstrumentationOptions>,
```
heaths, line 37: 1. Doc comments.
2. It doesn't seem right to expose this to developers *using* our clients.

All the information provided should be done in a way that is internal to the crate. That's why, for example, I required - as did track 1 - you pass the crate name and version to `azure_core::http::Pipeline::new()`. There's no reason outside of a crate developer (or emitter) to call that, and certainly no way an end user can do so.

sdk/core/azure_core/src/http/pipeline.rs
```diff
@@ -48,12 +50,56 @@ impl Pipeline {
...
+        let tracer = if install_instrumentation_policies {
+            core_client_options
+                .request_instrumentation
+                .tracing_provider
+                .as_ref()
+                .map(|tracing_provider| {
+                    tracing_provider.get_tracer(
+                        None,
+                        crate_name.unwrap_or("unknown"),
+                        crate_version.unwrap_or("unknown"),
```
heaths: @lmolkova is there any concern here that `version` needs to be a valid semver? If so, we could do something like "0.0.0+unknown"; otherwise, please disregard.

sdk/core/azure_core_opentelemetry/tests/integration_test.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
```
heaths, line 1: Given the filename becomes the should-be-unique executable name, this isn't a great name. Maybe "opentelemetry_test_client"? The other file could be "opentelemetry_test_service", perhaps.

sdk/core/azure_core_opentelemetry/tests/telemetry_service_implementation.rs
```diff
@@ -0,0 +1,539 @@
...
+        if let Some(tracer) = &self.tracer {
+            ctx = ctx.with_value(tracer.clone());
+        }
+        options.method_options.context = ctx;
+        self.get(path, Some(options)).await
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
heaths: Given this whole file is tests, really don't need this module `#[test]`s and `#[tokio::test]`s can be top-level.

sdk/core/azure_core_opentelemetry/README.md
```diff
@@ -1,17 +1,75 @@
 # Azure Core OpenTelemetry Tracing
 
-This crate provides OpenTelemetry distributed tracing support for the Azure SDK for Rust. It enables automatic span creation, context propagation, and telemetry collection for Azure service operations.
+This crate provides OpenTelemetry distributed tracing support for the Azure SDK for Rust.
```
heaths: Maybe link "OpenTelemetry" to the site for more info or something.

sdk/core/azure_core_opentelemetry/README.md
```diff
@@ -1,17 +1,75 @@
...
+It bridges the standardized azure_core tracing traits with OpenTelemetry implementation,
```
heaths: Programmatic elements should be in backticks everywhere for consistency.

```suggestion
It bridges the standardized `azure_core` tracing traits with OpenTelemetry implementation,
```

sdk/core/azure_core_opentelemetry/README.md
```diff
@@ -1,17 +1,75 @@
...
+enabling automatic span creation, context propagation, and telemetry collection for Azure services.
 
-## Features
+It allows Rust applications which use the [OpenTelemetry](https://opentelemetry.io/) APIs to generate OpenTelemetry spans for Azure SDK for Rust Clients.
 
-## Usage
+## OpenTelemetry integration with the Azure SDK for Rust
 
-### Basic Setup
+To integrate the OpenTelemetry APIs with the Azure SDK for Rust, you create a [`OpenTelemetryTracerProvider`] and pass it into your SDK ClientOptions.
```
heaths: This isn't a link to anywhere in markdown. Only works for rustdocs. You need to define the link to point to docs.rs, but since you're not published the CI will fail. Probably just don't link it for now and keep the backticks.

heaths: At least one more below.

sdk/core/azure_core_opentelemetry/README.md
```diff
@@ -1,17 +1,75 @@
...
+        ..Default::default()
+    },
+    ..Default::default()
+    };
+
+#   Ok(())
+# }
+```
+
+Once the `OpenTelemetryTracerProvider` is integrated with the Azure Service ClientOptions, the Azure SDK will be configured to capture per-API and per-HTTP operation tracing options, and the HTTP requests will be annotated with [W3C Trace Context headers](https://www.w3.org/TR/trace-context/).
```
heaths, line 75: Should add the various bottom sections from other READMEs as well e.g., troubleshooting, security, license, etc. Just boilerplate so you can copy and paste.

sdk/core/azure_core_opentelemetry/src/span.rs
```diff
@@ -41,52 +43,78 @@ impl OpenTelemetrySpan {
...
+        for (key, value) in header_map.into_iter() {
+            // Note: The OpenTelemetry HeaderInjector will always produce unique header names, so we don't need to
+            // handle the multiple headers case here.
+
+            if let Some(key) = key {
+                request.insert_header(
+                    HeaderName::from(key.as_str().to_owned()),
+                    // The value is guaranteed to be a valid UTF-8 string by the OpenTelemetry SDK,
+                    // so we can safely unwrap it.
+                    HeaderValue::from(value.to_str().unwrap().to_owned()),
```
heaths, line 102: Since you're either referencing (`as_str`) or copying (`to_str`), you don't need an owning iterator I don't think, which might allocate more memory than necessary. Can you just use `header_map.iter()` above?

sdk/core/azure_core_opentelemetry/src/telemetry.rs
```diff
@@ -12,29 +11,63 @@ use std::sync::Arc;
...
     }
 }
 
 impl TracerProvider for OpenTelemetryTracerProvider {
     fn get_tracer(
         &self,
-        name: &'static str,
+        namespace: Option<&'static str>,
+        package_name: &'static str,
         package_version: &'static str,
```
heaths: Should these actually be "crate" instead of "package" just because this is Rust? More intuitive, IMO.

sdk/core/azure_core_opentelemetry/src/tracer.rs
```diff
@@ -1,80 +1,102 @@
...
         kind: SpanKind,
-    ) -> Result<Arc<dyn azure_core::tracing::Span + Send + Sync>> {
+        attributes: Vec<Attribute>,
+    ) -> Arc<dyn azure_core::tracing::Span> {
         let span_builder = opentelemetry::trace::SpanBuilder::from_name(name)
-            .with_kind(OpenTelemetrySpanKind(kind).into());
+            .with_kind(OpenTelemetrySpanKind(kind).into())
+            .with_attributes(
+                attributes
+                    .into_iter()
```
heaths: Given how you're cloning information, does this really need to be `into_iter()` vs `iter()`?

heaths: At least one more below. I'll not repeat.

sdk/core/azure_core/src/http/options/mod.rs
heaths: Talked offline. I was confusing this with something else.

sdk/typespec/typespec_client_core/src/http/policies/retry/mod.rs
LarryOsterman: Yes, it's a type so it can go into a context - we technically could put a u32 in the context too, but that is not unique.

Cargo.lock
LarryOsterman: Yeah - not fully sure where it came from.

sdk/typespec/typespec_client_core/src/tracing/mod.rs
LarryOsterman: The expectation is that if there is a requirement that mutation happen, it's hidden by the implementation.

sdk/core/azure_core/src/http/pipeline.rs
LarryOsterman: FWIW, OTel's instrumentation scope requires that there be a crate name and a crate version (more accurately, an instrumentation scope has a name, and an optional version and the name of the instrumentation scope for Azure services is the package (crate) name).

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
LarryOsterman: I didn't want to pollute the entire "policies" module with these constants.

But I could easily move them up.

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
LarryOsterman: I'll just ignore the URL_FULL_ATTRIBUTE if the URL has a username or password. It's simpler and less fragile.

sdk/core/azure_core/src/http/policies/request_instrumentation.rs
LarryOsterman: These are shared between the two instrumentation policies.

I have a better idea (although it's going to be annoying - I'll create an instrumentation module under policies which will hold the two instrumentation related policies). That way the instrumentation stuff is isolated from the rest of the policies.

sdk/core/azure_core_opentelemetry/tests/telemetry_service_implementation.rs
LarryOsterman: The fiction I'm trying to express here is that this is an actual service client, and the tests module is the tests for that service client.

As such, the `TestServiceClient` isn't strictly a "test".

sdk/core/azure_core_macros/src/tracing.rs
```diff
@@ -0,0 +1,77 @@
...
+// Licensed under the MIT License.
+
+#[cfg(test)]
+pub(crate) mod tests {
+    use proc_macro2::{TokenStream, TokenTree};
+
+    // cspell: ignore punct
+
+    pub(crate) fn compare_token_tree(token: &TokenTree, expected_token: &TokenTree) -> bool {
+        //        println!("Comparing token: {token:?} with expected token: {expected_token:?}");
```
analogrelay: Old debugging content, I presume :)

sdk/core/azure_core/src/http/policies/instrumentation/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+const SERVER_ADDRESS_ATTRIBUTE: &str = "server.address";
+const SERVER_PORT_ATTRIBUTE: &str = "server.port";
+const URL_FULL_ATTRIBUTE: &str = "url.full";
+
+pub use public_api_instrumentation::PublicApiInstrumentationInformation;
+pub(crate) use public_api_instrumentation::PublicApiInstrumentationPolicy;
+pub(crate) use request_instrumentation::*;
+
+#[cfg(test)]
+mod tests {
```
analogrelay: It looks like these are utils for people writing tests to assert telemetry instrumentation. It might make sense to move it to `azure_core_test`, if that's feasible, since that's intended as our "test utilities" package.

sdk/core/azure_core/src/http/policies/instrumentation/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+        println!("Spans: {:?}", spans);
+        let span = spans[span_index].as_ref();
+        assert_eq!(span.name, expectation.span_name);
+        assert_eq!(span.kind, expectation.kind);
+        assert_eq!(*span.state.lock().unwrap(), expectation.status);
+        let attributes = span.attributes.lock().unwrap();
+        for attr in attributes.iter() {
+            println!("Attribute: {} = {:?}", attr.key, attr.value);
+            let mut found = false;
+            for (key, value) in &expectation.attributes {
```
analogrelay: Probably some Functionaly combinators you could use here, which would be fairly idiomatic for Rust, but personally I don't mind this kind of straightforward logic.

sdk/core/azure_core/src/http/policies/instrumentation/public_api_instrumentation.rs
```diff
@@ -0,0 +1,696 @@
...
+
+        // We next confirm if the context has public API instrumentation information.
+        // Without a public API information, we skip instrumentation.
+        let info = ctx.value::<PublicApiInstrumentationInformation>()?;
+
+        // Get the tracer from either the context or the policy.
+        let tracer = match ctx.value::<Arc<dyn Tracer>>() {
+            Some(tracer) => Some(tracer),
+            None => tracer.as_ref(),
+        };
```
analogrelay: nit: I think this could be an `or_else`, which would be slightly more idiomatic.

```suggestion
        let tracer = ctx.value::<Arc<dyn Tracer>>.or_else(|| tracer.as_ref());
```

sdk/core/azure_core/src/http/policies/instrumentation/public_api_instrumentation.rs
```diff
@@ -0,0 +1,696 @@
...
+
+        // If we don't have a tracer, skip instrumentation
+        let tracer = tracer?;
```
analogrelay: I'd combine this with the line above, since you just need to tack a `?` on the end and you've got the same behavior.

sdk/core/azure_core/src/http/policies/instrumentation/public_api_instrumentation.rs
```diff
@@ -0,0 +1,696 @@
...
+        // tracer.
+        let mut span_attributes = vec![];
+
+        for attr in &info.attributes {
+            // Add the attributes from the public API information to the span.
+            span_attributes.push(Attribute {
+                key: attr.key,
+                value: attr.value.clone(),
+            });
+        }
```
analogrelay: nit: IMO it's more idiomatic to use a map and collect here:

```suggestion
        let mut span_attributes = &info.attributes.iter().map(|attr|
           Attribute {
               key: attr.key,
               value: attr.value.clone(),
           }
        ).collect::<Vec<_>>();
```

sdk/core/azure_core/src/http/policies/instrumentation/public_api_instrumentation.rs
```diff
@@ -0,0 +1,696 @@
...
+    /// - `tracer`: An optional tracer to use for creating the span.
+    ///
+    /// # Returns
+    /// An optional span if the public API information is present and a tracer is available.
+    ///
+    /// If the context already has a span, it will return `None` to avoid nested spans.
+    /// If the context does not have a tracer it will use the value of the `tracer` argument.
+    /// If no tracer can be determined, it will return `None`.
+    ///
+    pub fn create_public_api_span(
```
analogrelay: Two things:

1. Is this used outside of the tests and the Policy implementation? It doesn't appear to be.
2. Static helper functions like this aren't terribly idiomatic in Rust in my, admittedly-limited, experience. I'd probably prefer to just see this as a function in this module (public if necessary). But that's like... just my opinion man.

sdk/core/azure_core/src/http/policies/instrumentation/public_api_instrumentation.rs
```diff
@@ -0,0 +1,696 @@
...
+        // Add the public API information and tracer to the context so that it can be used by the policy.
+        let ctx = Context::default()
+            .with_value(public_api_information)
+            .with_value(mock_tracer.clone());
+        let _result = public_api_policy.send(&ctx, request, &next).await;
+
+        mock_tracer_provider
+    }
+
+    fn check_public_api_instrumentation_result(
```
analogrelay: Looks very similar to `InstrumentationExpectation` above. Should this be converted to use that?

sdk/core/azure_core/src/http/policies/instrumentation/public_api_instrumentation.rs
```diff
@@ -0,0 +1,696 @@
...
+            }
+        }
+    }
+
+    // Tests for the create_public_api_span function.
+    #[test]
+    fn create_public_api_span() {
+        let tracer = Arc::new(MockTracingProvider::new()).get_tracer(Some("test"), "test", "1.0.0");
+
+        // Test when context has no PublicApiInstrumentationInformation
```
analogrelay, line 338: nit: My preference would be for these to be separate tests, since they're trying to be completely separate scenarios. Yes, they share the tracer, but that's more of a convenience.

sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
```diff
@@ -0,0 +1,493 @@
...
+    ) -> PolicyResult {
+        // If the context has a tracer (which happens when called from an instrumented method),
+        // we prefer the tracer from the context.
+        // Otherwise, we use the tracer from the policy itself.
+        // This allows for flexibility in using different tracers in different contexts.
+        let tracer = if ctx.value::<Arc<dyn crate::tracing::Tracer>>().is_some() {
+            ctx.value::<Arc<dyn crate::tracing::Tracer>>()
+        } else {
+            self.tracer.as_ref()
+        };
```
analogrelay: Another opportunity to use `or_else`:

```suggestion
        let tracer = ctx.value::<Arc<dyn crate::tracing::Tracer>>().or_else(|| self.tracer.as_ref());
```

sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
```diff
@@ -0,0 +1,493 @@
...
+
+        // If there is a span in the context, if it's not recording, just forward the request
+        // without instrumentation.
+        if let Some(span) = ctx.value::<Arc<dyn Span>>() {
+            if !span.is_recording() {
+                // If the span is not recording, we skip instrumentation.
+                return next[0].send(ctx, request, &next[1..]).await;
+            }
+        }
```
analogrelay: nit: If you want to get zesty with the combinators ;)

```suggestion
        if ctx.value::<Arc<dyn Span>>().is_some_and(|s| !s.is_recording) {
            // If the span is not recording, we skip instrumentation.
            return next[0].send(ctx, request, &next[1..]).await;
        }
```

sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
```diff
@@ -0,0 +1,493 @@
...
+            });
+        }
+        if let Some(port) = request.url().port_or_known_default() {
+            span_attributes.push(Attribute {
+                key: SERVER_PORT_ATTRIBUTE,
+                value: port.into(),
+            });
+        }
+        // Get the method as a string to avoid lifetime issues
+        let method_str = request.method().as_str();
```
analogrelay, line 106: Hmm... `azure_core::Method` is a Copy enum, yet `Request::method()` returns `&Method`. That seems wrong to me. If `Method` truly is `Copy` (and it should be, since as a small non-data-bearing enum, it's no larger than a pointer anyway), then it should be reasonable to return it by-value. That's a little outside the scope of this PR, but it would also solve your lifetime issue here.

sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
```diff
@@ -0,0 +1,493 @@
...
+
+        if let Some(err) = result.as_ref().err() {
+            // If the request failed, set an error type attribute.
+            let azure_error = err.downcast_ref::<crate::Error>();
+            if let Some(err_kind) = azure_error.map(|e| e.kind()) {
+                // If the error is an Azure core error, we set the error type.
+                span.set_attribute(ERROR_TYPE_ATTRIBUTE, err_kind.to_string().into());
+            } else {
+                // Otherwise, we set the error type to the error's text. This should never happen
+                // as the error should be an Azure core error.
```
analogrelay: Maybe put a `debug_assert!` to confirm this? I can understand not wanting to panic, even if we expect this to be impossible, but if it's truly supposed to be impossible, we should panic in debug builds.

sdk/core/azure_core_macros/src/tracing.rs
```diff
@@ -0,0 +1,77 @@
...
+                TokenTree::Literal(expected_literal) => {
+                    literal.to_string() == expected_literal.to_string()
+                }
+                _ => {
+                    println!("Unexpected token: {expected_token:?}");
+                    false
+                }
+            },
+        }
+    }
```
analogrelay: nit: You could make this a little more compact with something like this:

```suggestion
        match (token, expected_token) {
            (TokenTree::Group(group), TokenTree::Group(expected_group)) => compare_token_stream(group.stream(), expected_group.stream()),
            (TokenTree::Ident(ident), TokenTree::Ident(expected_ident) => *expected_ident == *ident,
            (TokenTree::Punct(punct), TokenTree::Punct(expected_punct) => *expected_punct == *punct,
            (TokenTree::Literal(literal), TokenTree::Literal(expected_literal) => expected_literal.to_string() == literal.to_string(),
            // Any other combination means the expected and actual are different kinds of TokenTree
            _ => false,
        }
    }
```

(I also removed the `println!` because I'm assuming it's for debugging)

sdk/core/azure_core_macros/src/tracing.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    pub(crate) fn compare_token_stream(actual: TokenStream, expected: TokenStream) -> bool {
+        let actual_tokens = Vec::from_iter(actual);
+        let expected_tokens = Vec::from_iter(expected);
+
+        if actual_tokens.len() != expected_tokens.len() {
+            println!(
```
analogrelay: More debugging `println!`s?

sdk/core/azure_core_macros/src/tracing_client.rs
```diff
@@ -0,0 +1,93 @@
...
+    let item_struct: ItemStruct = match syn::parse2(item.clone()) {
+        Ok(struct_item) => struct_item,
+        Err(_) => return false,
+    };
+
+    // Service clients must be public structs.
+    if !matches!(item_struct.vis, syn::Visibility::Public(_)) {
+        return false;
+    }
+    true
```
analogrelay, line 52: We assume later on in `parse_client` that the struct has no type parameters. That's probably fine, I can't think of a time when a client struct will need to be generic, but we should produce an error rather than silently eating the parameters.

```suggestion
    // Service clients must not have type parameters
    if !item_struct.generics.params.is_empty() {
        return false;
    }
    
    true
```

sdk/core/azure_core_macros/src/tracing_client.rs
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::{spanned::Spanned, ItemStruct, Result};
+
+const INVALID_SERVICE_CLIENT_MESSAGE: &str = "client attribute must be applied to a public struct";
```
analogrelay:
```suggestion
const INVALID_SERVICE_CLIENT_MESSAGE: &str = "client attribute must be applied to a public struct with no generic type parameters";
```

(Given the change I suggest below)

sdk/core/azure_core_macros/src/tracing_function.rs
```diff
@@ -0,0 +1,539 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+use proc_macro2::TokenStream;
+use quote::{quote, ToTokens};
+use syn::{parse::Parse, spanned::Spanned, ItemFn, Member, Result, Token};
+
+const INVALID_PUBLIC_FUNCTION_MESSAGE: &str =
+    "function attribute must be applied to a public function returning a Result.";
```
analogrelay: A little bit clearer, given the expectations listed in the doc comment below. Also, Rust compiler errors don't have trailing periods.

```suggestion
    "function attribute must be applied to a public function returning Self, or a Result and/or Arc containing Self";
```

sdk/core/azure_core_macros/src/tracing_function.rs
```diff
@@ -0,0 +1,539 @@
...
+
+    // Function must be public.
+    if !matches!(item_fn.vis, syn::Visibility::Public(_)) {
+        return false;
+    }
+
+    // Function must return a Result type.
+    if let syn::ReturnType::Type(_, ty) = &item_fn.sig.output {
+        if !matches!(ty.as_ref(), syn::Type::Path(_)) {
+            return false;
```
analogrelay, line 189: This is only validating that the function returns _something_, by my reading, not that it returns some combination of `Result`/`Arc`/`Self`. I'm OK with keeping it simple for now, since if the function doesn't return what we expect we'll get a compilation error anyway, I just want to make sure I'm following it correctly (perhaps some clarifying comments would help).

sdk/core/azure_core_macros/src/tracing_function.rs
```diff
@@ -0,0 +1,539 @@
...
+        for stream in types.iter() {
+            let parsed: FunctionNameAndAttributes =
+                syn::parse2(stream.clone()).expect("Failed to parse");
+            assert_eq!(parsed.function_name, "Text String");
+            if !parsed.arguments.is_empty() {
+                assert_eq!(parsed.arguments.len(), 2);
+                assert_eq!(parsed.arguments[0].0, "a");
+                assert_eq!(parsed.arguments[1].0, "b");
+            }
+        }
```
analogrelay: These scenarios are largely covered by the blocks below, I think.

sdk/core/azure_core_macros/src/tracing_function.rs
```diff
@@ -0,0 +1,539 @@
...
+        {
+            let test_stream = quote! { "Test Function", (arg1 = 42, arg2 = "value") };
+            let parsed: FunctionNameAndAttributes =
+                syn::parse2(test_stream).expect("Failed to parse");
+            assert_eq!(parsed.function_name, "Test Function");
+            assert_eq!(parsed.arguments.len(), 2);
+            assert_eq!(parsed.arguments[0].0, "arg1");
+            assert_eq!(parsed.arguments[0].1, parse_quote!(42));
+            assert_eq!(parsed.arguments[1].0, "arg2");
+            assert_eq!(parsed.arguments[1].1, parse_quote!("value"));
```
analogrelay: nit: I personally prefer comparing vectors to vectors where possible rather than asserting individual items:

```suggestion
            assert_eq!(parsed.arguments, vec![
                ("arg1".to_string(), parse_quote!(42)),
                ("arg2".to_string(), parse_quote!("value")),
            ]);
```

analogrelay: Though perhaps the type inference won't be quite as good there 🤔 . If so, this is fine as is.

sdk/core/azure_core_macros/src/tracing_function.rs
```diff
@@ -0,0 +1,539 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use syn::parse_quote;
+
+    use super::*;
+
+    #[test]
+    fn test_parse_function_name_and_attributes() {
```
analogrelay, line 207: I'd prefer separate test cases for this, but that's personal preference. This is also a place where I'd sometimes use a helper function or macro to generate test cases.

sdk/core/azure_core_macros/src/tracing_new.rs
```diff
@@ -0,0 +1,312 @@
...
+    };
+
+    // Service clients new functions must be public.
+    if !matches!(item_fn.vis, syn::Visibility::Public(_)) {
+        return false;
+    }
+
+    // Service clients new functions must have a name that starts with `new_` or "with_".
+    if !item_fn.sig.ident.to_string().starts_with("new")
+        && !item_fn.sig.ident.to_string().starts_with("with")
```
analogrelay: I'm not entirely convinced we should be enforcing this here. Rust has a fairly dominant pattern for constructor names, but I wouldn't say it's totally universal. I've seen constructors call `from_...` as well, for example. The code we generate doesn't depend on the name, so I don't think we should enforce naming here. If we want to enforce constructor naming, we can do that in linting rules, etc.

sdk/core/azure_core_macros/src/tracing_new.rs
```diff
@@ -0,0 +1,312 @@
...
+}
+
+/// Parse the token stream for an Azure Service client "new" declaration.
+///
+/// An Azure Service client "new" declaration is a public function whose name starts with
+/// `new` and returns either a new client instance or an error.
+///
+/// This macro will ensure that the fn is public and returns one of the following:
+/// 1) `Self`
+/// 1) `Arc<Self>`
```
analogrelay, line 148: Does this code handle `Arc`? It looks like it might not.

sdk/core/azure_core_macros/src/tracing_new.rs
```diff
@@ -0,0 +1,312 @@
...
+fn parse_struct_expr(
+    client_namespace: &str,
+    struct_body: &ExprStruct,
+    default: TokenStream,
+    is_ok: bool,
+) -> TokenStream {
+    if struct_body.path.is_ident("Self") {
+        let fields = struct_body.fields.iter();
+        let before_self = quote! {
+            let tracer =
```
analogrelay: Depending on how much we care about resilience against a service client doing something... odd, we may want to do something slightly different. As-written, this would shadow any variable named `tracer` that the service client's new method defined themselves. Maybe that's fine and we just tell the client to not use the attribute. One possible solution though could be to quote just an expression that produces the tracer (rather than quoting a statement that binds it to a variable). Then, you'd initialize the field using that expression. So the generated code would look something like this:

```rust
Ok(Self {
    tracer: {
        // All the initialization code from below.
        if let Some(tracer_options) = ... {
        } else {
            ...
        }
    },
    ... rest of the fields
})
```

sdk/core/azure_core_macros/src/tracing_new.rs
```diff
@@ -0,0 +1,312 @@
...
+        ));
+    }
+    let namespace_attrs: NamespaceAttribute = syn::parse2(attr)?;
+
+    let client_fn: ItemFn = syn::parse2(item.clone())?;
+
+    let vis = &client_fn.vis;
+    let ident = &client_fn.sig.ident;
+    let inputs = client_fn.sig.inputs.iter();
+    let body = client_fn.block.stmts.iter().map(|stmt| {
```
analogrelay: Should we track if we found a suitable statement to rewrite and emit an error if we found no such statement? If we don't find a statement to rewrite, the attribute didn't do anything, which feels like an error. The user should just remove the attribute.

sdk/core/azure_core_macros/src/tracing_subclient.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    let fn_body = fn_struct.block;
+    // Subclient constructors must have a body with a single statement.
+    if fn_body.stmts.len() != 1 {
+        return false;
+    }
+
+    // Subclient constructors must have a name that starts with `get_`.
+    if !fn_struct.sig.ident.to_string().starts_with("get_") {
```
analogrelay: Again, I'm not sure we need to enforce function naming with this attribute.

sdk/typespec/typespec_client_core/src/tracing/attributes.rs
```diff
@@ -1,17 +1,134 @@
...
+///
+/// Attributes are used to provide additional context and metadata about a span or event.
+/// They can be of various types, including strings, integers, booleans, and arrays.
+///
+/// Attributes are typically used to enrich telemetry data with additional information
+/// that can be useful for debugging, monitoring, and analysis.
+#[derive(Debug, PartialEq, Clone)]
+pub struct Attribute {
+    /// A key-value pair attribute.
+    pub key: &'static str,
```
analogrelay: There are well-defined attributes for which the key is computed at runtime. Most database conventions, [Cosmos' included](https://opentelemetry.io/docs/specs/semconv/database/cosmosdb/) define `db.query.parameter.<key>` for encoding the values of query parameters.

If we switch this to a `Cow<'static, str>` we should largely get the best of both worlds. We'd be able to use static strings without heap allocation but still support heap-allocated dynamic strings where needed.

analogrelay (commented):
Still working my way through most of `azure_core_opentelemetry` but wrapping up my day and wanted to get some of the comments out there. Almost entirely style/idiom stuff, with a few other thoughts on the proc macro and attribute key storage. Looking good! I like the proc macros!

sdk/core/azure_core/src/http/policies/instrumentation/mod.rs
LarryOsterman: There's a slight dependency problem because `azure_core_test` depends on `azure_core`. But I think that may not be a problem here. And it turns out that these functions will be needed in `azure_core_test` when I add support for automated testing of testing.

sdk/core/azure_core/src/http/policies/instrumentation/public_api_instrumentation.rs
LarryOsterman: It's intended for use by SDK client developers who need to create spans outside of the HTTP pipeline. The idea is that if you need to add public API span attributes that are derived from the HTTP response, you need to be able to control the lifetime of the span - the span that's created by the `PublicApiInstrumentationPolicy` cannot be used (I believe that CosmosDb falls into this bucket). 

After I created this, I realized that there is another option to solve this problem, which is to add an `Option<FnOnce<(ctx: &Context, response:&RawResponse)>>` to the `PublicApiInstrumentationInfo` type which would allow the client author to define a callback which is called on the response. That may be more idiomatic than having a static helper function.

sdk/core/azure_core_macros/src/tracing.rs
LarryOsterman: I knew there was a syntax to combine the nested match, I just couldn't remember it. Thanks.

sdk/core/azure_core_macros/src/tracing.rs
LarryOsterman: Some of these are intentional because they are there to help diagnose test failures. But I think I'll convert them to traces.

sdk/core/azure_core_macros/src/tracing.rs
LarryOsterman: Yup.

sdk/core/azure_core/src/http/policies/instrumentation/mod.rs
LarryOsterman: Ok - moved the mock tracing infrastructure to azure_core_test so it can be re-used.

Working on refactoring the span validation logic to be more generic.

sdk/core/azure_core/src/http/policies/instrumentation/public_api_instrumentation.rs
LarryOsterman: Interesting. clippy suggested using `.or()`, which isn't quite right IMHO.

sdk/core/azure_core/src/http/policies/instrumentation/public_api_instrumentation.rs
LarryOsterman: yup consolidated all these when I moved them to azure_core_test.

sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
LarryOsterman: Doesn't the `return` in this case return from the lambda and not from the outer function?

sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
LarryOsterman: I agree. @heaths, I'll make the change that Ashley suggested here in a separate PR because it appears that since `Method` is a simple enumeration, the compiler can generate better code if we return a copy of the method rather than the actual method. And I don't see a reason why adding the level of indirection here makes a difference.

sdk/core/azure_core_macros/src/tracing_function.rs
LarryOsterman: I believe that your suggestion should be the error for the `new` attribute, not the `function attribute`. I've applied it to `new`.

sdk/core/azure_core_macros/src/tracing_function.rs
LarryOsterman: That is correct - it only validates that the function returns *something* - this is constant with the `recorded::test` macro which similarly ensures that test functions simply return *something*.

sdk/core/azure_core_macros/src/tracing_function.rs
LarryOsterman: Yeah, you're right - removing.

sdk/core/azure_core_macros/src/tracing_new.rs
LarryOsterman: Upon reflection, I agree.

sdk/core/azure_core_macros/src/tracing_new.rs
LarryOsterman: It does not. Let me fix that.

sdk/core/azure_core_macros/src/tracing_new.rs
LarryOsterman: That's a great suggestion.

sdk/typespec/typespec_client_core/src/tracing/attributes.rs
LarryOsterman: Ok, good to know, I'll update this.

doc/distributed-tracing-for-rust-service-clients.md
```diff
@@ -0,0 +1,407 @@
...
+
+A "tracer" is a factory for "Spans". A `Tracer` is configured with three parameters:
+
+* `namespace` - the "namespace" for the service client. The namespace for all azure services are listed [on this page](https://learn.microsoft.com/azure/azure-resource-manager/management/azure-services-resource-providers).
+* `package name` - this is typically the Cargo package name for the service client (`env!("CARGO_PKG_NAME")`)
+* `package version` - this is typically the version of the Cargo package for the service client (`env!("CARGO_PKG_VERSION")`)
+
+Tracers have three mechanisms for creating spans:
+
+* Create a new root span.
```
lmolkova: does it describe our API or the OTel one?

For client libs, I don't believe we ever have a need to start a root span. we should start using explicit parent or a current span and, for the latter, let otel figure out if there is one.

doc/distributed-tracing-for-rust-service-clients.md
```diff
@@ -0,0 +1,407 @@
...
+
+## Additional requirements
+
+For public APIs, the rule of thumb is: "If the operation does not take time and cannot fail, it doesn't get a span". For most public APIs, this isn't a huge deal, but for pageable and long running operations, it changes how the code is generated. Specifically, for pageables, the actual service client does not actually interact with the network and cannot fail, but the individual pager returned does interact with the network and can fail - thus the pager's interactions with the service need to be instrumented with a span. Long Running Operations behave similarly - while the original API is instrumented with a span, the same is true for the status monitor - it also needs to be instrumented with a span whose name matches the name of the original API.
+
+In addition, [certain service clients](https://github.com/Azure/azure-sdk/blob/main/docs/tracing/distributed-tracing-conventions.md#library-specific-attributes) (Cosmos DB, KeyVault, etc) have additional client-specific attributes which need to be added to the span.
```
lmolkova, line 75: this would probably need some cosmetic changes based on the https://github.com/Azure/azure-sdk/pull/9007 outcome

doc/distributed-tracing-for-rust-service-clients.md
```diff
@@ -0,0 +1,407 @@
...
+
+1) A struct field named `tracer` which is an `Arc<dyn Tracing::Tracer>` which holds the tracing implementation specific tracer.
+2) Code in the service client's `new` function which instantiates a `tracer` from the `TracerProvider` configured in the service client options. The primary function of this tracer is to provide the value for the `az.namespace` attribute for both the public API spans and the HTTP request spans.
+3) Code in each service client public API which instantiates a public API span.
+
+For the Rust implementation, if a tracer provider is configured, ALL http operations will have HTTP request spans created regardless of whether the public API spans are created.
+
+## Implementation details
+
+To provide for requirement #1, if a customer provides a value for the `azure_core::ClientOptions::request_instrumentation` structure, the Azure Core HTTP pipeline will add a `PublicApiInstrumentationPolicy` to the pipeline which is responsible for creating the public API outer span.
```
lmolkova, line 89: do we support bootstrapping tracing with global tracer provider when it's configured by otel?

doc/distributed-tracing-for-rust-service-clients.md
```diff
@@ -0,0 +1,407 @@
...
+   1) Once the span has been created, the policy will attach the newly created span to the context so other policies in the pipeline (specifically the `RequestInstrumentationPolicy` can use this span).
+1) Once the span has been created, the policy calls the next policy in the pipeline.
+1) After the remaining policies in the pipeline have run, the policy inspects the `Result` of the pipeline execution and sets the `error.type` attribute and the span status based on the `Result` of the operation.
+
+### RequestInstrumentationPolicy
+
+The `RequestInstrumentationPolicy` will do the following:
+
+1) If the `Context` parameter for the  `RequestInstrumentationPolicy` contains a `Tracer` value, then the `RequestInstrumentationPolicy` will use that `Tracer` value to create the span, otherwise it will use the pre-configured tracer from when the policy was created.
+2) If the `Context` parameter for the `RequestInstrumentationPolicy contains a`Span` value, then the policy will use that span as the parent span for the newly created HTTP request span, otherwise it will create a new span.
```
lmolkova:
```suggestion
2) If the `Context` parameter for the `RequestInstrumentationPolicy` contains a `Span` value, then the policy will use that span as the parent span for the newly created HTTP request span, otherwise it will create a new span.
```

doc/distributed-tracing-for-rust-service-clients.md
```diff
@@ -0,0 +1,407 @@
...
+
+This design means that even if a service public API is not fully instrumented with a `Tracer` or a `Span`, it will still generate some HTTP request traces.
+
+Since the namespace attribute is service-client wide, it makes sense to capture that inside a per-service client field, that way it can be easily accessed from service clients.
+
+## Convenience Macros
+
+To facilitate the implementation of the three core requirements above, three attribute-like macros are defined for the use of each service.
```
lmolkova: could we add some note that these API are only for client libraries and are not intended for end users? We have a lot of evidence from other languages showing that SDK developers and end users assume that we recommend application developers to use those APIs and we should try our best to create clarity.

doc/distributed-tracing-for-rust-service-clients.md
```diff
@@ -0,0 +1,407 @@
...
+    endpoint: Url,
+}
+```
+
+to
+
+```diff
+pub struct MyServiceClient {
+    endpoint: Url,
++    tracer: std::sync::Arc<dyn azure_core::tracing::Tracer>,
```
lmolkova, line 152: 🚀

doc/distributed-tracing-for-rust-service-clients.md
```diff
@@ -0,0 +1,407 @@
...
+
+The job of a "Tracer Provider" is to be a factory for tracers. It is the "gateway" construct for distributed tracing.
+
+### Tracer
+
+A "tracer" is a factory for "Spans". A `Tracer` is configured with three parameters:
+
+* `namespace` - the "namespace" for the service client. The namespace for all azure services are listed [on this page](https://learn.microsoft.com/azure/azure-resource-manager/management/azure-services-resource-providers).
+* `package name` - this is typically the Cargo package name for the service client (`env!("CARGO_PKG_NAME")`)
+* `package version` - this is typically the version of the Cargo package for the service client (`env!("CARGO_PKG_VERSION")`)
```
lmolkova, line 24: it's not an immediate need, but it should be possible to add the schema URL in the future. 

Now it can be a singleton and can be hardcoded in the core, but it won't be the case the moment instrumentation adds a custom attribute - if core updates to vZ, old library versions instrumentations reporting vX will start lying.

Hand-written instrumentations will inevitably have different versions - e.g. cosmos follows vX and servicebus follows vY.
-

sdk/core/azure_core/src/http/pipeline.rs
```diff
@@ -48,12 +50,59 @@ impl Pipeline {
...
+        let tracer = if install_instrumentation_policies {
+            core_client_options
+                .request_instrumentation
+                .tracer_provider
+                .as_ref()
+                .map(|tracer_provider| {
+                    tracer_provider.get_tracer(
+                        None,
+                        crate_name.unwrap_or("Unknown"),
+                        crate_version.unwrap_or("0.1.0"),
```
lmolkova: nit, is it intentional to have it hardcoded? OTel does not require version, we can just keep it unset if we don't know or set to "Unknown" too

sdk/core/azure_core/src/http/policies/instrumentation/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+//! Instrumentation pipeline policies.
+
+mod public_api_instrumentation;
+mod request_instrumentation;
+
+// Distributed tracing span attribute names. Defined in
+// [OpenTelemetrySpans](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md)
+// and [Azure conventions for open telemetry spans](https://github.com/Azure/azure-sdk/blob/main/docs/tracing/distributed-tracing-conventions.md)
+const AZ_NAMESPACE_ATTRIBUTE: &str = "az.namespace";
+const AZ_CLIENT_REQUEST_ID_ATTRIBUTE: &str = "az.client.request.id";
```
lmolkova:
```suggestion
const AZ_CLIENT_REQUEST_ID_ATTRIBUTE: &str = "az.client_request_id";
```

sdk/core/azure_core/src/http/policies/instrumentation/public_api_instrumentation.rs
```diff
@@ -0,0 +1,654 @@
...
+            });
+        }
+
+        // Create a span with the public API information and attributes.
+        let span =
+            tracer.start_span_with_current(info.api_name, SpanKind::Internal, span_attributes);
+
+        // If nothing is listening to the span, we skip instrumentation.
+        if !span.is_recording() {
+            return None;
```
lmolkova: this would be problematic 

1. even if span is not recording (sampled out), we still want to propagate its context. Some of its child spans or logs under it can be recorded and they would all be correlated if current context is set in-process and is propagated downstream
2. there could be some future cases when span start/end callbacks in OTel's span processor got invoked even for non-recording spans, and someone can try to convert spans to metrics this way.

The common way to optimize here is do

- start span with sampling-relevant attributes
- make it current and inject its context regardless of recording
- only add new attributes if span is recording
- end it when the operation is over

sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
```diff
@@ -0,0 +1,483 @@
...
+        // which can lead to unnecessary overhead if the tracer is not needed.
+        #[allow(clippy::unnecessary_lazy_evaluations)]
+        let tracer = ctx
+            .value::<Arc<dyn crate::tracing::Tracer>>()
+            .or_else(|| self.tracer.as_ref());
+
+        // If there is a span in the context, if it's not recording, just forward the request
+        // without instrumentation.
+        if let Some(span) = ctx.value::<Arc<dyn Span>>() {
+            if !span.is_recording() {
```
lmolkova: same as one of the prev comments, we can't short-circuit here, we need to keep span active and inject context into headers regardless of whether it's being recorded to correlate other telemetry.

sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
```diff
@@ -0,0 +1,483 @@
...
+                return next[0].send(ctx, request, &next[1..]).await;
+            }
+        }
+
+        let Some(tracer) = tracer else {
+            return next[0].send(ctx, request, &next[1..]).await;
+        };
+
+        let mut span_attributes = vec![Attribute {
+            key: HTTP_REQUEST_METHOD_ATTRIBUTE,
```
lmolkova: this along with server.address | port are sampling-relevant, we provide themat span start time 

https://github.com/Azure/azure-sdk/blob/6c3dbe77e701c0fe3799f52c2e92b632bc24d715/docs/observability/opentelemetry-conventions.md?plain=1#L326-L331

sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
```diff
@@ -0,0 +1,483 @@
...
+                parent_span.clone(),
+            )
+        } else {
+            // If no parent span exists, start a new span with the "current" span (if any).
+            // It is up to the tracer implementation to determine what "current" means.
+            tracer.start_span_with_current(method_str, SpanKind::Client, span_attributes)
+        };
+
+        if !span.is_recording() {
+            // If the span is not recording, we skip instrumentation.
```
lmolkova: and here

sdk/core/azure_core_macros/src/tracing_new.rs
LarryOsterman: Agreed. It's *hard* to track the "I replaced something" in the conversion logic, but instead I changed the "is valid" functions to check for the conditions that would cause us to modify the stream.

doc/distributed-tracing-for-rust-service-clients.md
LarryOsterman: Removed.

doc/distributed-tracing-for-rust-service-clients.md
LarryOsterman: Yes - we have `OpenTelemetryTracerProvider::new_from_global_provider()` which returns a tracer provider configured with the global provider (technically it actually configures the provider as `None` but the `get_tracer` function uses either the provided tracer or the global tracer provider if there is no provided tracer).

doc/distributed-tracing-for-rust-service-clients.md
LarryOsterman: Done.

sdk/core/azure_core/src/http/pipeline.rs
LarryOsterman: No - but making version optional makes sense.

sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
LarryOsterman: I believe that's the way this code works. Lines 79-115 (or so) define the attributes which are set on the HTTP request span.

doc/distributed-tracing-for-rust-service-clients.md
LarryOsterman: That becomes interesting ifk the cosmos Schema version is newer than the core schema version. The core traces will follow the core schema version, but the cosmos traces will be the published version.

How do you express the difference between the Schema used by core functionality (request spans) and the public API spans?

sdk/core/azure_core_test/src/tracing.rs
```diff
@@ -0,0 +1,255 @@
...
+    fn span_id(&self) -> [u8; 8] {
+        [0; 8] // Mock span ID
+    }
+
+    fn record_error(&self, _error: &dyn std::error::Error) {
+        todo!()
+    }
+
+    fn set_current(&self, _context: &Context) -> Box<dyn typespec_client_core::tracing::SpanGuard> {
+        todo!()
```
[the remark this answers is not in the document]
LarryOsterman: I would rather this test panic in this case.

RickWinter (approved):

sdk/core/azure_core/src/lib.rs
```diff
@@ -27,7 +27,10 @@ pub use typespec_client_core::{
     fmt, json, sleep, stream, time, Bytes, Uuid,
 };
 
+/// Abstractions for distributed tracing and telemetry.
 pub mod tracing {
+    pub use crate::http::policies::PublicApiInstrumentationInformation;
+    pub use azure_core_macros::*;
```
heaths, line 33: Might be good to be explicit here because we may add more non-tracing stuff in the future. Or maybe everything therein should be in a `tracing` submodule now (well, before GA) such that you import it like everything below on line 34.

sdk/core/azure_core/src/http/options/mod.rs
```diff
@@ -27,6 +29,17 @@ pub struct ClientOptions {
 
     /// User-Agent telemetry options.
     pub user_agent: Option<UserAgentOptions>,
+
+    /// Options for request instrumentation, such as distributed tracing.
+    ///
+    /// If not specified, defaults to no instrumentation.
+    ///
+    pub request_instrumentation: Option<RequestInstrumentationOptions>,
```
heaths, line 37: Nit: isn't this request and response instrumentation, though? Could we just call this "instrumentation"?

sdk/core/azure_core/src/http/pipeline.rs
```diff
@@ -48,12 +50,55 @@ impl Pipeline {
...
+            core_client_options
+                .request_instrumentation
+                .tracer_provider
+                .as_ref()
+                .map(|tracer_provider| {
+                    tracer_provider.get_tracer(None, crate_name.unwrap_or("Unknown"), crate_version)
+                })
+        } else {
+            None
+        };
```
heaths, line 72: Much more concise and idiomatic:

```rust
let tracer = match core_client_options.request_instrumentation.tracer_provider {
  Some(ref provider) => provider.map(...),
  None => None,
};
```

sdk/core/azure_core/src/http/options/mod.rs
LarryOsterman: Good point - technically it's request instrumentation and public API instrumentation, but it eventually could grow to include metrics and other constructs.

PR #2743, opened 2025-06-30
Update UserAgentPolicy to better align with guidelines

heaths:
Adds documentation to our repo's root README - which is also now syntactically verified when testing `azure_core_test`, and removes the `UserAgentOptions::disabled` in favor of, like most languages, recommending a user policy to remove the `User-Agent` header.


LarryOsterman (approved):
LGTM.

ronniegeraghty:
Updates to the README look good. Thanks Heath!

PR #2744, opened 2025-06-30
Added as_headers trait to ItemOptions

gsa9989:
Added as_headers trait to ItemOptions to send all item options headers from one function. Added tests to make sure as_headers is creating the correct type of headers. I also added the remaining item options and the necessary headers to make the correct requests. As_headers function is now used in create_item, replace_item, upsert_item, read_item, delete_item, and patch_item.


sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -1,9 +1,16 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-use azure_core::http::{ClientMethodOptions, ClientOptions, Etag};
-
-use crate::models::ThroughputProperties;
+use crate::{
+    constants::{self},
```
analogrelay: This can actually be rewritten as just `constants`. You only need `constants::{self, ...}` when you're importing _both_ the `constants` module **and** things from within it. This kind of error can easily occur when you start out importing something and then switch to just referencing it using the `constants::` module prefix :).

```suggestion
    constants,
```

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -1,9 +1,16 @@
...
+    models::ThroughputProperties,
+};
+use azure_core::http::headers::{AsHeaders, HeaderName, HeaderValue};
+use azure_core::http::{headers, ClientMethodOptions, ClientOptions, Etag};
+use serde_json::value::Index;
```
analogrelay: You're not actually using this. Probably got brought in by VS Code when you were typing out `IndexingPolicy` somewhere :)

```suggestion
```

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -43,10 +50,71 @@ pub struct DeleteDatabaseOptions<'a> {
     pub method_options: ClientMethodOptions<'a>,
 }
 
+/// Options to be passed to ['ItemOptions'](crate::options::ItemOptions).
+#[derive(Clone)]
+pub enum ConsistencyLevel {
+    Eventual,
+    Session,
+    BoundedStaleness,
+    Strong,
```
analogrelay, line 59: We're missing `ConsistentPrefix` here, otherwise it looks good!

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -43,10 +50,71 @@ pub struct DeleteDatabaseOptions<'a> {
     pub method_options: ClientMethodOptions<'a>,
 }
 
+/// Options to be passed to ['ItemOptions'](crate::options::ItemOptions).
```
analogrelay: We should rephrase this to something like `Specifies consistency levels that can be used when working with Cosmos APIs`. And include a link to the docs: https://learn.microsoft.com/en-us/azure/cosmos-db/consistency-levels

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -43,10 +50,71 @@ pub struct DeleteDatabaseOptions<'a> {
...
+            ConsistencyLevel::Session => "Session",
+            ConsistencyLevel::BoundedStaleness => "BoundedStaleness",
+            ConsistencyLevel::Strong => "Strong",
+        };
+        write!(f, "{}", value)
+    }
+}
+
+#[derive(Clone)]
+pub enum IndexingDirective {
```
analogrelay, line 77: This should have a doc comment too. At least on the type.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -43,10 +50,71 @@ pub struct DeleteDatabaseOptions<'a> {
...
+    pub post_triggers: Option<Vec<String>>,
+    /// Applies when working with Session consistency.
+    /// Each new write request to Azure Cosmos DB is assigned a new Session Token.
+    /// The client instance will use this token internally with each read/query request to ensure that the set consistency level is maintained.
+    ///
+    /// See [Session Tokens](https://docs.azure.cn/en-us/cosmos-db/nosql/how-to-manage-consistency?tabs=portal%2Cdotnetv2%2Capi-async#utilize-session-tokens) for more.
+    pub session_token: Option<String>,
+    /// Used to specify the consistency level for the operation.
+    ///
+    /// The default value is the consistency level set on the Cosmos DB account.
```
analogrelay, line 114: I like how you linked to the docs in the other options. Let's do that here too.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -57,6 +125,54 @@ pub struct ItemOptions<'a> {
     pub enable_content_response_on_write: bool,
 }
 
+impl AsHeaders for ItemOptions<'_> {
+    type Error = Infallible;
```
analogrelay, line 130: Nice job noticing this in other implementations! The `as_headers` method is specified as fallible (i.e. it returns `Result` because it could fail) so that the caller knows to properly handle any errors. So, the `AsHeaders` trait requires that we specify what the _type_ of that error would be. However, many AsHeaders implementations (including this one) actually _cannot_ fail, so we use the type `Infallible` as the error, which is a type that cannot actually be constructed (sometimes called a "never" type). Rust can optimize around this a bit and save storage space since it knows that in this case there can never actually _be_ an error.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -43,10 +50,71 @@ pub struct DeleteDatabaseOptions<'a> {
...
+    }
+}
+
 /// Options to be passed to APIs that manipulate items.
 #[derive(Clone, Default)]
 pub struct ItemOptions<'a> {
     pub method_options: ClientMethodOptions<'a>,
+    /// Triggers executed before the operation.
+    ///
+    /// See [Triggers](https://learn.microsoft.com/en-us/rest/api/cosmos-db/triggers) for more.
```
analogrelay: When we're linking to other docs within our doc comments we need to tweak the URLs a bit. When you copy a URL from your browser, it will have the `/en-us/` portion in the URL, which is a link _directly_ to the US English version of the doc (since, presumably, that's the language your machine would have requested). Our docs may be translated into other languages though (perhaps at a future date, if they aren't already translated), and we want users to be able to view those docs in the language they prefer. If you just remove the `/en-us/` segment in the URL (`https://learn.microsoft.com/en-us/foo/bar` becomes `https://learn.microsoft.com/foo/bar`), then you're linking to the doc _without_ forcing a particular language and the docs site will automatically detect the right language and redirect the user to that version (falling back to US English if we haven't translated the doc).

All of this is to say, you should go through and remove the `/en-us/` portion of all these doc links ;).

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -111,3 +227,43 @@ pub struct ReadDatabaseOptions<'a> {
 pub struct ThroughputOptions<'a> {
     pub method_options: ClientMethodOptions<'a>,
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn item_options_as_headers() {
```
analogrelay, line 240: Let's add a test with everything set to `None` (which you can probably do by just creating `ItemOptions::default()`) and verify we get an empty vector back.

analogrelay (changes requested):
Just some small changes needed!

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Features Added
 * Added `if_match_etag` to `ItemOptions` ([#2705](https://github.com/Azure/azure-sdk-for-rust/pull/2705))
+* Added `as_headers` to `ItemOptions` ([#2744](https://github.com/Azure/azure-sdk-for-rust/pull/2744))
+
```
tvaron3: Nit: remove extra line

tvaron3 (approved):
LGTM

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -43,20 +47,136 @@ pub struct DeleteDatabaseOptions<'a> {
     pub method_options: ClientMethodOptions<'a>,
 }
 
+/// Specifies consistency levels that can be used when working with Cosmos APIs.
+///
+/// Learn more at [Consistency Levels](https://learn.microsoft.com/azure/cosmos-db/consistency-levels)
+#[derive(Clone)]
+pub enum ConsistencyLevel {
```
kushagraThapar, line 54: If we are exposing these to the public API, should we mention them in the changelog?

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -43,20 +47,136 @@ pub struct DeleteDatabaseOptions<'a> {
...
+}
+
 /// Options to be passed to APIs that manipulate items.
 #[derive(Clone, Default)]
 pub struct ItemOptions<'a> {
     pub method_options: ClientMethodOptions<'a>,
+    /// Triggers executed before the operation.
+    ///
+    /// See [Triggers](https://learn.microsoft.com/rest/api/cosmos-db/triggers) for more.
+    pub pre_triggers: Option<Vec<String>>,
```
kushagraThapar, line 101: And all of these in the changelog?

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Features Added
 * Added `if_match_etag` to `ItemOptions` ([#2705](https://github.com/Azure/azure-sdk-for-rust/pull/2705))
+* Added `as_headers` to `ItemOptions` ([#2744](https://github.com/Azure/azure-sdk-for-rust/pull/2744))
```
kushagraThapar: I am curious, is `as_headers` public API that the customers will use in their code?

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -111,3 +231,71 @@ pub struct ReadDatabaseOptions<'a> {
 pub struct ThroughputOptions<'a> {
     pub method_options: ClientMethodOptions<'a>,
 }
+
+#[cfg(test)]
+mod tests {
```
kushagraThapar, line 236: what other test cases can we think of? like having few of them as None while others are filled in? or maybe having invalid values for some of them? like invalid types if possible?

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
analogrelay: Yeah, we probably should have a short reference to them in the changelog. Just a brief list of the new types added, to represent request headers on item operations

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
analogrelay: Technically yes, since the AsHeaders trait is public in azure_core. However, a customer would have almost no reason to use it. We could probably omit the changelog entry that describes adding as_headers and focus instead on the specific headers we've added

analogrelay: So maybe rephrase this as "Added several more options to `ItemOptions`: `consistency_level`, `pre_triggers`, etc. etc."

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
analogrelay: Maybe a separate line "Added types to represent various header values: `ConsistencyLevel`, `IndexingDirective`, etc."

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
analogrelay: I don't think we'd want to validate things like session tokens or etags. Are there any other invalid values that could end up here given that we're using statically typed enums and lists?

analogrelay (commented):
Agree with @kushagraThapar on some CHANGELOG updates. This should be ready to roll after those!

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
kushagraThapar: how about the error cases? How is the error case handled? can we validate if it gets thrown properly and how customers should handle it or we handle it?

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
gsa9989: Fixed! Thank you!

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
analogrelay: Our `as_headers` implementation is infallible, there aren't any errors it can produce right now. Which makes sense as it's largely just a "plain-old-data" type that doesn't validate it's properties. The `as_headers` just converts that to headers and lets the server sort it out if there's a problem.

The `AsHeaders` trait _requires_ that we return a `Result<>` and specify an error type, because it needs to be able to support implementations that _do_ return errors, but we use the `Infallible` type (which represents no possible errors) in our implementation

PR #2746, opened 2025-07-01
Update rand, chacha, and fe2o3 crates and updated MSV to 1.85

LarryOsterman:
Updates to rand and rand_chacha bring forward a number of other packages, which require an update to the MSV.

Updating the crate MSV to 1.85, although technically the minimum MSV is 1.82.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,9 +82,8 @@ fe2o3-amqp-cbs = { version = "0.14" }
 fe2o3-amqp-types = { version = "0.14" }
 flate2 = "1.1.0"
 futures = "0.3"
-getrandom = { version = "0.3", features = ["wasm_js"] }
+getrandom = { version = "0.3" }
 hmac = { version = "0.12" }
-js_sys = "3.77"
 litemap = "0.7.4"
 log = "0.4"
 openssl = { version = "0.10.72" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ authors = ["Microsoft"]
 edition = "2021"
 license = "MIT"
 repository = "https://github.com/azure/azure-sdk-for-rust"
-rust-version = "1.81"
+rust-version = "1.85"
 
 [workspace.dependencies.typespec]
 default-features = false
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,15 +75,16 @@ cargo_metadata = "0.18.1"
 clap = { version = "4.4.16", features = ["derive"] }
 criterion = { version = "0.5", features = ["async_tokio"] }
 dyn-clone = "1.0"
-fe2o3-amqp = { version = "0.12", features = ["native-tls", "tracing", "uuid"] }
-fe2o3-amqp-ext = { version = "0.12" }
-fe2o3-amqp-management = { version = "0.12", features = ["tracing"] }
-fe2o3-amqp-cbs = { version = "0.12" }
-fe2o3-amqp-types = { version = "0.12" }
+fe2o3-amqp = { version = "0.14", features = ["native-tls", "tracing", "uuid"] }
+fe2o3-amqp-ext = { version = "0.14" }
+fe2o3-amqp-management = { version = "0.14", features = ["tracing"] }
+fe2o3-amqp-cbs = { version = "0.14" }
+fe2o3-amqp-types = { version = "0.14" }
 flate2 = "1.1.0"
 futures = "0.3"
-getrandom = { version = "0.2", features = ["js"] }
+getrandom = { version = "0.3", features = ["wasm_js"] }
 hmac = { version = "0.12" }
+js_sys = "3.77"
 litemap = "0.7.4"
 log = "0.4"
 openssl = { version = "0.10.72" }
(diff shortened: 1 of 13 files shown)


Cargo.toml
```diff
@@ -29,7 +29,7 @@ authors = ["Microsoft"]
 edition = "2021"
 license = "MIT"
 repository = "https://github.com/azure/azure-sdk-for-rust"
-rust-version = "1.81"
+rust-version = "1.85"
```
heaths, line 32: @RickWinter @ronniegeraghty while we only needed to bump the MSRV to 1.82 to necessarily take `getrandom` - which was like peeling an onion that wasn't possible when I tried it almost a year ago (few packages had updated to 0.3 and the wasm support between 0.2 and 0.3 required incompatible changes) - but, as I brought up internally, AWS SDK requires 1.85 by comparison and it gives us some useful features like async closures, which simplify code in many areas. There wasn't any pushback on the idea when I floated it internally, but now it's reality. Good with this?

Cargo.toml
```diff
@@ -75,31 +75,32 @@ cargo_metadata = "0.18.1"
...
+fe2o3-amqp-ext = { version = "0.14" }
+fe2o3-amqp-management = { version = "0.14", features = ["tracing"] }
+fe2o3-amqp-cbs = { version = "0.14" }
+fe2o3-amqp-types = { version = "0.14" }
 flate2 = "1.1.0"
 futures = "0.3"
-getrandom = { version = "0.2", features = ["js"] }
+getrandom = { version = "0.3", features = ["wasm_js"] }
 hmac = { version = "0.12" }
+js_sys = "3.77"
```
heaths: Do we still need to explicitly reference this now that you've updated more? It'd be great not to have to reference transitive dependencies too often.

eng/scripts/Analyze-Code.ps1
```diff
@@ -34,7 +34,15 @@ Invoke-LoggedCommand "cargo fmt --all -- --check"
 Invoke-LoggedCommand "cargo clippy --workspace --all-features --all-targets --keep-going --no-deps"
 
 if ($CheckWasm) {
+  # Save the original RUSTFLAGS to restore later
+  $OriginalRustFlags = $env:RUSTFLAGS
+  # This is needed to ensure that the `getrandom` crate uses the `wasm_js` backend
+  $env:RUSTFLAGS = ${env:RUSTFLAGS} + ' --cfg getrandom_backend="wasm_js"'
```
heaths, line 40: Maybe we shouldn't enable the `wasm_js` feature at all: https://docs.rs/getrandom/latest/getrandom/#webassembly-support:~:text=WARNING%3A%20It%20is,in%20library%20crates.

sdk/core/azure_core_amqp/src/fe2o3/error.rs
```diff
@@ -104,7 +104,7 @@ impl From<fe2o3_amqp_types::definitions::Error> for AmqpDescribedError {
         AmqpDescribedError::new(
             (&e.condition).into(),
             e.description,
-            (&e.info.unwrap_or_default()).into(),
+            (&(*e.info.unwrap_or_default())).into(),
```
heaths: Traits like `Deref` might otherwise be at place here, but if not, shouldn't this be equivalent?
```suggestion
            e.info.unwrap_or_default().into(),
```

sdk/core/azure_core_opentelemetry/Cargo.toml
```diff
@@ -24,7 +24,7 @@ typespec_client_core.workspace = true
 [dev-dependencies]
 opentelemetry_sdk = { version = "0.30", features = ["testing"] }
 tokio.workspace = true
-tracing-opentelemetry = "0.26"
+#tracing-opentelemetry = "0.26"
```
heaths: Probably just remove it.
```suggestion
```

heaths (approved):
A few suggestions, but otherwise LGTM. Let's wait on Rick's and Ronnie's reply, though.

eng/scripts/Analyze-Code.ps1
LarryOsterman: Oh interesting - you're 100% right - we don't need the wasm_js feature *if* we are setting the RUSTFLAGS environment variable.

Cargo.toml
LarryOsterman: It's not needed.

sdk/core/azure_core_amqp/src/fe2o3/error.rs
LarryOsterman: I was wondering the same thing.

sdk/core/azure_core_opentelemetry/Cargo.toml
LarryOsterman: Yeah - I'll be adding it back in a later PR (when I verify that tracing logs integrate with otel spans) but for now it doesn't make sense.

Cargo.toml
ronniegeraghty: I'm for updating our rust version to `1.85`. 
If amazon is already there, I think it's safe for us and were still in beta, so we have no commitment to a specific version yet.

RickWinter: Upating to 1.85 seems reasonable.

PR #2748, opened 2025-07-02
Support reconstituting pageables

heaths:
Adds PageIterator::with_next_link to support general guidelines: https://azure.github.io/azure-sdk/general_design.html#general-pagination-expose-paging-apis

### DIFF
--- a/sdk/core/azure_core/src/http/pager.rs
+++ b/sdk/core/azure_core/src/http/pager.rs
@@ -4,12 +4,20 @@
 use crate::http::{headers::HeaderName, response::Response};
 use async_trait::async_trait;
 use futures::{stream::unfold, FutureExt, Stream};
-use std::{fmt, future::Future, pin::Pin, task};
+#[cfg(not(target_arch = "wasm32"))]
+use std::str::FromStr;
+use std::{
+    fmt,
+    future::Future,
+    pin::Pin,
+    sync::{Arc, Mutex},
+    task,
+};
 use typespec::Error;
 use typespec_client_core::http::{DeserializeWith, Format, JsonFormat};
 
 /// The result of fetching a single page from a [`Pager`], whether there are more pages or paging is done.
-pub enum PagerResult<P, N> {
+pub enum PagerResult<P, N: AsRef<str>> {
     /// There are more pages the [`Pager`] may fetch using the `next` token.
     More { response: P, next: N },
     /// The [`Pager`] is done and there are no additional pages to fetch.
@@ -29,7 +37,7 @@ impl<P, F> PagerResult<Response<P, F>, String> {
     }
 }
 
-impl<P, N: fmt::Debug> fmt::Debug for PagerResult<P, N> {
+impl<P, N: fmt::Debug + AsRef<str>> fmt::Debug for PagerResult<P, N> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Self::More { next, .. } => f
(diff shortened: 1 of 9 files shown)


heaths:
I incidentally came across https://azure.github.io/azure-sdk/general_design.html#general-pagination-expose-paging-apis while working on the poller and thinking about reconstitution. I called everything "next_link" since, at least in data plane, we tend to use that concept more; however, the general guidelines do mention "continuation" and .NET's implementation also stems around "continuation" so I'm open to that if people prefer. /cc @RickWinter @ronniegeraghty @johanste

LarryOsterman (approved):
It looks straightforward.

heaths:
@RickWinter @ronniegeraghty @LarryOsterman @analogrelay, to emphasize the design here, I'm only making the `PageIterator` reconstitutable given a `next_link` (or should we call it `continuation`)? from another `PageIterator` like so:

```rust
        let mut second_pager: PageIterator<Response<Page>> =
            PageIterator::from_callback(make_callback()).with_next_link(next_link);
```

I didn't add this to `ItemIterator` - the real `Pager` type. It would work about the same but capture the index of the item, but I didn't because the whole `next_link` concept is already unstable. Another item could be added and throw all the items off. At least if you're enumerating pages and the items therein, you're likely to try to dedup already rather than with an `ItemIterator` you're probably doing a loop over or one of the many `std::iter` functions. I'm not married to not adding it, though. Still, I wonder how often this actually happens in practice.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -408,8 +445,59 @@ impl<P> PageIterator<P> {
...
+    /// while let Some(secrets) = pager.try_next().await? {
+    ///     let secrets = secrets.into_body().await?;
+    ///     for secret in secrets.value {
+    ///         println!("{:?}", secret.id);
+    ///     }
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn with_next_link(self, next_link: String) -> Self {
```
analogrelay: My only minor concern here is that Cosmos calls it a `continuation` (because that's what the header itself is called) and not a `next_link`. Having said that, if Cosmos is the outlier here, I'm OK with that being handled in documentation.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -289,6 +297,7 @@ impl<P: Page> fmt::Debug for ItemIterator<P> {
 pub struct PageIterator<P> {
     #[pin]
     stream: Pin<BoxedStream<P>>,
+    next: Arc<Mutex<Option<String>>>,
```
analogrelay: Feels like a place where [`AtomicCell`](https://docs.rs/crossbeam/latest/crossbeam/atomic/struct.AtomicCell.html) would be useful, but I wouldn't want the dependency on crossbeam just for that. Plus, this should be fine for now, and it's all internal so we can adjust later.

sdk/core/azure_core/src/http/pager.rs
analogrelay: I see that later API docs also reference continuation tokens, so maybe just update the docs on `with_next_link` to reference continuation tokens as well? Just so that users searching around for a term they expect will be able to find it.

analogrelay:
> Still, I wonder how often this actually happens in practice.

Cosmos _does_ have some scenarios in which continuation tokens flow back and forth through other things (cookies, form fields, query strings, whatnot) and we need to reconstitute a pager. In other SDKs we generally do that with our own APIs (`QueryOptions`/`ItemOptions`/etc.). In our case, we also have to handle new items being inserted and even partition merge/split operations so there's already a fair bit of complexity. So, I'm not really worried about the `azure_core` API catering to those scenarios. Omitting `ItemIterator` from this seems entirely appropriate. I don't think Cosmos supports resuming at the item iterator level in other languages.

sdk/core/azure_core/src/http/pager.rs
heaths: I'm starting to like "continuation" more and more myself as well. Seems a nextLink could be a continuation, but a continuation can't be a nextLink.

heaths:
For naming,

* .NET uses `ContinuationToken`: https://github.com/Azure/azure-sdk-for-net/blob/ee8cc201901798f303bff0e210bf4d5bac9111b5/sdk/core/Azure.Core/src/Page.cs#L28
* Go: nothing I can find
* Java uses `ContinuationToken`: https://github.com/Azure/azure-sdk-for-java/blob/f862b55ae534f3bc5f392c7acebcd688e530ecfb/sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedResponseBase.java#L75
* JS uses `continuationToken`: https://github.com/Azure/azure-sdk-for-js/blob/61cc246a94222c65ed9a9a5b56b0e4e00a56406d/sdk/core/core-paging/src/models.ts#L11
* Python uses `continuation_token`: https://github.com/Azure/azure-sdk-for-python/blob/65947ad78ff17a882260d7d9af8efa2c38924432/sdk/core/azure-core/azure/core/paging.py#L52

sdk/core/azure_core/src/http/pager.rs
heaths: I've changed this across the repo and will update the emitter to use `continuation` now for the `PagerResult::More` variant (just to be consistent).

LarryOsterman (approved):

PR #2749, opened 2025-07-02
[Storage] Combined HNS Client PoC

vincenttran-msft:


sdk/storage/azure_storage_blob/src/clients/hierarchical_client.rs
```diff
@@ -0,0 +1,74 @@
...
+
+// Directory state specific functions
+impl HierarchicalClient<Directory> {
+    pub async fn create(
+        &self,
+        options: Option<HierarchicalClientCreateOptions<'_>>,
+    ) -> Result<RawResponse> {
+        self.client.create("directory".to_string(), options).await
+    }
+}
```
vincenttran-msft, line 74: I decided to implement these both separately as the following for now. Since this is the "common" functionality, do we want to add a convenience method to call into these from `HierarchicalClient<T>` and just take a `resource_type` parameter of some sort?

Edit: Another remaining question: I wonder if the option bags are identical (or if it is the case that there are just some options that are not applicable depending on `resource_type` since it's 1 REST API)

Issue #2757, opened 2025-07-03
Return final resource from Poller


heaths:
Splitting up #2458 into different tasks in priority order. This is to add an `IntoFuture` implementation driven by the `Stream` on the `Poller` to refresh the status monitor. We'd probably want `IntoFuture` so we take ownership from the stream, thereby invalidating the stream from that point on.

The `impl<M: StatusMonitor> Future<Output = M::Output> for Poller<M>` would return an `azure_core::Result<M::Output>`[^1] where we'll do a final `GET` based on the [final state designation](https://azure.github.io/typespec-azure/docs/libraries/azure-core/reference/decorators/#@Azure.Core.useFinalStateVia) - or just return the already-deserialized result from the status monitor if defined according to [current guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#the-status-monitor-resource) - if `PollerStatus::Succeeded`. For any other terminal state we'd return an `azure_core::Error`.

We may want to add an `azure_core::error::ErrorKind::Canceled` for user- or service-initiated cancellations. This wouldn't be if the dev canceled the `Future` by dropping it, mind you. It would correspond to `PollerStatus::Canceled` and may be actionable if a caller can distinguish the LRO was cancelled.

[^1]: Exact design is open to change, but this it the POR implementation.

PR #2759, opened 2025-07-03
Add Poller<T> for long-running operations (LROs)

### LINKED ISSUE #2482: Only LRO (poller) status should use case-insensitive deserialization
> After discussion, only LRO status should use case-insensitive deserialization because it does tend to vary and is not something customers usually parse directly themselves. Service-defined enums should be case-sensitive as most other languages are doing. 

 _Originally posted by @heaths in [#1105](https://github.com/Azure/azure-sdk-for-rust/issues/1105#issuecomment-2803304805)_

### LINKED ISSUE #2756: Implement Stream to return/refresh status monitor until a terminal state
Splitting up #2458 into different tasks in priority order. The first priority is just to return/refresh the status monitor, which we get in the first or subsequent requests. To also avoid the double `await`s we want to get rid of in #2073 we'll delegate getting the first and subsequent requests to the `Poller`. The existing design for the `Pager` should work well for this.

We'll implement `futures::Stream` on this along with appropriate pauses in the implementation following other languages' design such that customers can always check the status monitor e.g.,

```rust
use futures::stream::TryStreamExt as _;

let mut poller = client.operation(None)?;
while let Some(monitor) = poller.try_next().await? {
    let Some(status) = monitor.status else {
        continue;
    }

    match status {
        PollerStatus::InProgress => continue,
        // ...
    }
}
```

heaths:
Resolves #2756 and resolves #2482


heaths:
I still need to implement the retry options (will likely mimic our `RetryOptions` for the `RetryPolicy`) but I wanted to get feedback on the DX here. Since implementing `IntoFuture` for #2757 is on the backlog and would return the `StatusMonitor::Output`, I took a page from other uses of `wait()` in both `std::process::Command::wait()` and `tokio::process::Command::wait()` and return a future that will just pump the `Stream` implementation until terminal state and return the last status monitor.

This may be temporary until we implement #2757, though maybe still valid. This at least makes tests, examples, and doc examples easier to understand; otherwise, we'd have to add loops for everything and just ignore every intermediate status monitor.

sdk/keyvault/azure_security_keyvault_certificates/tests/certificate_client.rs
```diff
@@ -53,12 +52,10 @@ async fn certificate_roundtrip(ctx: TestContext) -> Result<()> {
...
         ..Default::default()
     };
-    let operation = client
-        .create_certificate("certificate-roundtrip", body.try_into()?, None)
-        .await?
-        .into_body()
+    client
+        .create_certificate("certificate-roundtrip", body.try_into()?, None)?
+        .wait()
         .await?;
```
heaths: @RickWinter, @ronniegeraghty, @LarryOsterman, @analogrelay here's an example of usage, similar to `tokio::process::Command::wait().await`.

sdk/keyvault/azure_security_keyvault_certificates/src/generated/clients/certificate_client.rs
```diff
@@ -137,25 +138,65 @@ impl CertificateClient {
...
-    pub async fn create_certificate(
+    pub fn create_certificate(
         &self,
         certificate_name: &str,
         parameters: RequestContent<CreateCertificateParameters>,
         options: Option<CertificateClientCreateCertificateOptions<'_>>,
-    ) -> Result<Response<CertificateOperation>> {
-        let options = options.unwrap_or_default();
-        let ctx = Context::with_context(&options.method_options.context);
+    ) -> Result<Poller<CertificateOperation>> {
```
heaths: @jhendrixMSFT what do you think of the code here in terms of emitting? The line breaks I added for review readability but obviously don't have to be retained.

magodo:
@heaths I have a working LRO (for ARM mgmt plane API) implementation, which naively port the azure-sdk-for-go to Rust: https://github.com/magodo/azure/tree/main/src/client, if you are interested in.

sdk/keyvault/azure_security_keyvault_certificates/tests/certificate_client.rs
heaths: Talked in today's team meeting and people are okay with this.

sdk/keyvault/azure_security_keyvault_certificates/src/generated/clients/certificate_client.rs
heaths: Along with @antkmsft, we'll discuss this offline. I strongly believe doing more at compile time is worth it but also agree we should provide more helpers in `azure_core`. We have some already, though I don't know how well they fit this new `Poller` model compared to track 1 when they were added. But, in concept, those are good examples of what `azure_core` would provide.

sdk/keyvault/azure_security_keyvault_certificates/src/models.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+pub use crate::generated::models::*;
+use std::sync::LazyLock;
+
+/// Gets the default self-signed [`CertificatePolicy`].
+pub static DEFAULT_POLICY: LazyLock<CertificatePolicy> = LazyLock::new(|| CertificatePolicy {
```
heaths: @RickWinter @chlowell this change is a bit orthogonal, but:

1. All other languages but Go have something similar. Even the `az` CLI has this.
2. It greatly simplifies examples.

Ya'll okay with it?

sdk/core/azure_core/README.md
```diff
@@ -246,6 +246,101 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 }
 ```
 
+### Consuming service methods returning `Poller<T>`
+
+If a service call may take a while to process, it would return `Result<Poller<T>>` as a result, representing a long-running operation (LRO).
```
LarryOsterman:
```suggestion
If a service call may take a while to process, it should return `Result<Poller<T>>` as a result, representing a long-running operation (LRO).
```
Alternate replacement word: `will` ("will" is stronger than "should")

sdk/core/azure_core/src/http/poller.rs
```diff
@@ -1,51 +1,491 @@
...
+    task::{Context, Poll},
+};
 
-/// Default retry time for long running operations if no retry-after header is present
+/// Default retry time for long-running operations if no retry-after header is present
 ///
-/// This value is the same as the default used in the Azure SDK for Python.
-/// Ref: <https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-mgmt-core/azure/mgmt/core/polling/arm_polling.py#L191>
+/// This value is the same as the default used in other Azure SDKs e.g.,
+/// <https://github.com/Azure/azure-sdk-for-python/blob/azure-mgmt-core_1.5.0/sdk/core/azure-mgmt-core/azure/mgmt/core/polling/arm_polling.py#L191>
```
LarryOsterman: Does it matter that this is using a constant from a management plane SDK? Do the constants differ between management and data plane (the LRO pattern is different, so I wasn't sure if the durations would also be different).

sdk/core/azure_core/README.md
heaths: I actually waffled on "should" vs "would" and flat-out rejected "will" when I first wrote this little ditty for `Pager`. What concerns me is that the emitter might not always if the right LRO information isn't in TypeSpec, like I fear might be the case for Key Vault. It certainly *should*, but that felt more like a recommendation to crate developers compared to "would" that was meant to have a connotation like "will" to developers using our crates without the promise of actually returning a `Poller`. @ronniegeraghty or @RickWinter thoughts here? "will" just felt too much like a promise we can't really make.

sdk/core/azure_core/src/http/poller.rs
```diff
@@ -1,51 +1,491 @@
...
+    ///
+    ///     fn status(&self) -> PollerStatus {
+    ///         self.status.clone().unwrap_or_default()
+    ///     }
+    /// }
+    ///
+    /// let url = "https://example.com/my_operation".parse().unwrap();
+    /// let mut req = Request::new(url, Method::Post);
+    ///
+    /// let poller = Poller::from_callback(move |operation_url: Option<Url>| {
```
LarryOsterman, line 235: Now that our baseline is Rust 1.85, can we use an `async move` here?

LarryOsterman (approved):

sdk/core/azure_core/README.md
heaths: Not a hill I even want to climb so I'll just change it to "should".

sdk/core/azure_core/src/http/poller.rs
heaths: Probably not, but since we're not focusing on ARM right now I'll use a DP reference.

sdk/core/azure_core/src/http/poller.rs
heaths: Alas, no:

```text
 error: async closure does not implement `Fn` because it captures state from its environment
   --> sdk/core/azure_core/src/http/poller.rs:640:13
    |
639 |         let mut poller = Poller::from_callback(
    |                          --------------------- required by a bound introduced by this call
640 |             async move |_| {
    |             ^^^^^^^^^^^^^^
    |
note: required by a bound in `http::poller::Poller::<M>::from_callback`
   --> sdk/core/azure_core/src/http/poller.rs:278:48
    |
276 |     pub fn from_callback<
    |            ------------- required by a bound in this associated function
277 |         #[cfg(not(target_arch = "wasm32"))] N: Send + 'static,
278 |         #[cfg(not(target_arch = "wasm32"))] F: Fn(Option<N>) -> Fut + Send + 'static,
    |                                                ^^^^^^^^^^^^^^^^^^^^ required by this bound in `Poller::<M>::from_callback`
```

We could take another look down the road at this and see if we can't make it work with reasonable changes.

sdk/typespec/typespec_client_core/src/error/http_error.rs
```diff
@@ -50,7 +50,7 @@ impl HttpError {
     /// This searches the entire ["source" chain](https://doc.rust-lang.org/std/error/trait.Error.html#method.source)
     /// looking for an `HttpError`.
     pub fn try_from(error: &Error) -> Option<&Self> {
-        let mut error = error.get_ref()? as &(dyn std::error::Error);
```
LarryOsterman: Would it be possible to put this in a separate PR so that CI pipelines are unblocked?

heaths: I got it in the emitter update. I didn't expect it to linger so long or I would've.

LarryOsterman (approved):

heaths:

sdk/keyvault/azure_security_keyvault_certificates/src/models.rs
RickWinter: Seems like a good helper for customers. Is the name verbose enough, or should it be clearer that its a `DEFAULT_CERTIFICATE_POLICY`

RickWinter: Nit: We could have this code in each example or an example_helper class.

I'm thinking about location b/c I wonder how much this type gets used by real world scenarios

heaths: I suppose `DEFAULT_CERTIFICATE_POLICY` would be better, sure. Other languages just often use some variation of "default" e.g., in .NET it's `CertificatePolicy.Default` (static read-only property), but something like `CertificatePolicy::default()` for Rust has a different meaning i.e., the `Default::default()` implementation.

As for making this only available to doc tests,

1. There is a known issue with including things only for `cfg(any(test, doctest))` but,
2. For using in copy-and-paste / "getting started" examples, it'd be a disservice to customers to use something that doesn't actually exist when they copy the code.

heaths: Removed the public API and inlined the definition in samples per internal discussions.

antkmsft (approved):

heaths:

heaths:

chlowell (approved):

heaths:

PR #2761, opened 2025-07-07
add FeedRange APIs

analogrelay:
We recently had a contribution to the Go SDK to add this API, and I realized it would be useful to have it here in the Rust SDK now for integration with the query engine. So, this PR introduces a new API: `ContainerClient::get_feed_ranges` which returns a `Vec<FeedRange>` which describes the _current_ physical partition layout. Like in other SDKs, `FeedRange`s map on to Partition Key Range IDs, but we don't expose the actual range IDs to the user, so that when a user _specifies_ FeedRanges somewhere, we can adapt the ranges to the current partition topology (adjusting based on merges/splits, etc.).

Internally, we also have a `ContainerClient::get_partition_key_ranges` which uses the `/pkranges` REST API to fetch the full Partition Key Range metadata. This is used in the query engine, which already has to be updated with merge/split awareness and needs the exact PKRangeID for issuing queries. This API isn't part of the public surface area.

The _model_ type returned by `get_partition_key_ranges` is `PartitionKeyRange` and that struct **is** part of the public surface area. It has to be in order to put it on the `QueryEngine` abstraction. But there's no way, using our public APIs, for a user to get the **actual** PK Range IDs from the REST API, nor do any of our APIs accept `PartitionKeyRange` objects. With a little bit of `#[cfg]` fiddling, I may actually be able to make it so these APIs are only public in the `preview_query_engine` feature, which is the only place we actually need them to be exported outside the crate.

This PR also adjusts the `QueryEngine` API to take the Partition Key Ranges as objects, since now the Rust SDK itself has the model and APIs to deserialize them. Before, it was passing the PKRanges as raw JSON and the query engine did the deserializing. This just removes some code duplication. I'll update the query engine to fix this shortly.

### DIFF
--- a/sdk/cosmos/assets.json
+++ b/sdk/cosmos/assets.json
@@ -1,6 +1,6 @@
 {
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "rust",
-  "Tag": "rust/azure_data_cosmos_a39b424a5b",
+  "Tag": "rust/azure_data_cosmos_8671767119",
   "TagPrefix": "rust/azure_data_cosmos"
 }
--- a/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
@@ -3,30 +3,35 @@
 
 use crate::{
     constants,
-    models::{ContainerProperties, PatchDocument, ThroughputProperties},
+    models::{
+        ContainerProperties, FeedRange, PartitionKeyRange, PatchDocument, ThroughputProperties,
+    },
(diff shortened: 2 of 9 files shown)


sdk/cosmos/assets.json
```diff
@@ -1,6 +1,6 @@
 {
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "rust",
-  "Tag": "rust/azure_data_cosmos_a39b424a5b",
+  "Tag": "rust/azure_data_cosmos_8671767119",
```
analogrelay, line 4: Corresponding assets repository commit:

https://github.com/Azure/azure-sdk-assets/commit/8671767119cc7c92529c64541544ecffd9aa0648

PR #2770, opened 2025-07-08
typespec_client_core: New `web_runtime` to support wasm32-unknown-unknown

### LINKED ISSUE #2754: wasm32-unknown-unknown failed to run
### Bug Title

_No response_

### Crate Name

typespec_client_core

### Crate Version

0.4.0+ (rev: ec358961712425cb8c688f43685f91c446249011)

### Description

First off, thanks for the amazing work for this SDK!

Recently, I have created a small client based on this SDK trying to successfully send a dummy request to Azure. It works fine when running natively. I also hope it can run in the browser, so I successfully to build it (with some tweaking about the `getrandom` crate).

PS: I've patched the Cargo.toml to override the dependencies on `azure_core` and `azure_identity` to revision `ec358961712425cb8c688f43685f91c446249011`.

When I run it, I get the following panic:

Image

The issue also occurs in the official release v0.25.0 (thankfully the build doesn't need tweaking). The error is pointing to the same crate, same module, but complaining about `thread::spawn`/`thead:sleep` is not available in wasm32.

I'd like to know if wasm32-unknown-unknown is offically supported (as I see a lot of `cfg` on `wasm32` in this repo). I believe using `wasm-bindgen` is able to bridge a JS sleep for use in Rust.

Also, since the error is raised from the RetryPolicy, is there a way to allow the user to customize that policy, e.g. don't sleep or something...

### Steps to Reproduce

I have a private repo to reproduce this, please let me know if you want to test it out, so I can invite you to my repo.

### Checklist

magodo:
The `typespec_client_core::sleep::sleep` is using the async_runtime's sleep implementation, which is complicated to support `wasm32`, and will panic now when targeting to `wasm32`. Instead of make the `standard_runtime` to support wasm32 sleep (I wonder if it's feasible), we can instead just introduce another `sleep` implementation to simply use the `gloo_timers::future::sleep`.

Also, this PR adds a `cfg_attr` for conditionally remove the `Send` trait from the `async_trait` for the `RetryPolicy`, as the other parts of the code do.

~Additionally, the `getrandom` crate will enable `wasm_js` features when targetting to `wasm`, which is necessary.~

Fix #2754


sdk/typespec/typespec_client_core/src/sleep.rs
```diff
@@ -27,3 +31,13 @@ use crate::{async_runtime::get_async_runtime, time::Duration};
 pub async fn sleep(duration: Duration) {
     get_async_runtime().sleep(duration).await
 }
+
+#[cfg(target_family = "wasm")]
+pub async fn sleep(duration: Duration) {
```
LarryOsterman: Rather than special casing sleep, why not build a wasm specific async runtime which uses `gloo_timers` - that would be consistent with the other implementations, and be more consistent with the design. 

Bypassing the async runtime in this fashion means that on wasm, you prevent customers from providing their own sleep implementation that *isn't* gloo_timers based, which is counter to the core design.

LarryOsterman (changes requested):

sdk/typespec/typespec_client_core/src/sleep.rs
magodo: You're right @LarryOsterman.. That was a quick hack to prove it works. I've submitted new commits to revert that change, and introduced a web_runtime to better align with the design of the crate. Please take another look! Thx!

sdk/typespec/typespec_client_core/src/async_runtime/standard_runtime.rs
```diff
@@ -150,7 +150,7 @@ impl AsyncRuntime for StdRuntime {
     /// Uses a simple thread based implementation for sleep. A more efficient
     /// implementation is available by using the `tokio` crate feature.
     #[cfg_attr(target_arch = "wasm32", allow(unused_variables))]
```
heaths, line 153: Seems this line is no longer needed.

sdk/typespec/typespec_client_core/src/async_runtime/web_runtime.rs
```diff
@@ -0,0 +1,29 @@
...
+        Box::pin(async {
+            wasm_bindgen_futures::spawn_local(f);
+            Ok(())
+        })
+    }
+
+    fn sleep(&self, duration: Duration) -> TaskFuture {
+        Box::pin(async move {
+            if let Ok(d) = duration.try_into() {
+                gloo_timers::future::sleep(d).await;
```
heaths, line 29: If we're taking a hard dependency on this, add it to the `Cargo.toml` like you did `wasm-bindgen-futures`. Or is it re-exported or otherwise exposed by `wasm-bindgen-futures`?

heaths (dismissed):
Better, but a couple lingering changes.

sdk/typespec/typespec_client_core/src/async_runtime/mod.rs
```diff
@@ -189,12 +192,16 @@ pub fn set_async_runtime(runtime: Arc<dyn AsyncRuntime>) -> crate::Result<()> {
 }
 
 fn create_async_runtime() -> Arc<dyn AsyncRuntime> {
-    #[cfg(not(feature = "tokio"))]
+    #[cfg(target_arch = "wasm32")]
```
LarryOsterman: Much better, but IMHO this should probably be tied to a specific feature which is the `gloo` runtime, unless you can assert safely that `gloo` is the runtime of choice for all wasm developers (I'm not familiar enough with wasm development to have an opinion here).

This is similar to how we've tied the `tokio` runtime to the `tokio` feature - because the tokio runtime is not a globally accepted async runtime, we hide its implementation behind a feature.

sdk/typespec/typespec_client_core/src/async_runtime/standard_runtime.rs
magodo: If we want to do https://github.com/Azure/azure-sdk-for-rust/pull/2770#discussion_r2195904158, this is still needed to allow a user to select the `StdRuntime` even target to wasm32.

sdk/typespec/typespec_client_core/src/async_runtime/web_runtime.rs
magodo: Both the workspace Cargo.toml and crate Cargo.toml has the gloo-timers defined.

sdk/typespec/typespec_client_core/src/async_runtime/mod.rs
magodo: @LarryOsterman The `gloo-timers` is provided by https://github.com/rustwasm, which is built on top of `wasm-bindgen` and `js-sys`, providing a more user friendly API. So if we really want to introduce a feature, then I'd prpose to name it "wasm-bindgen` (as the `spawn` is based on `wasm-bindgen-futures`, and the `sleep` is based on `gloo-timers`, which in turn based on `wasm-bindgen`).

Whilst, I believe even we use the target for this `create_async_runtime()`, it only means we provide a default runtime for the users. They are still able to call `set_async_runtime()` to choose the runtime of their choice. Otherwise, if we do:

```
    #[cfg(feature = "wasm-bindgen")]
    {
        Arc::new(web_runtime::WasmBindgenRuntime) as Arc<dyn AsyncRuntime>
    }
    #[cfg(feature = "tokio")]
    {
    }
    #[cfg(not(any(feature = "tokio", feature = "wasm-bindgen")))]
    {
        Arc::new(standard_runtime::StdRuntime)
    }
```

For users who target to `wasm32-unknown-unknown`, without specifying the `wasm-bindgen` feature, then they'll get a `StdRuntime` by default, which will only panic in any case.

sdk/typespec/typespec_client_core/src/async_runtime/web_runtime.rs
heaths: I don't see it in the root `Cargo.toml` in `main`. If we reference any crates I want them explicitly in the workspace `Cargo.toml`.

magodo: Sorry @heaths, I didn't get your point.. Do you mean add both dependencies to the root level `Cargo.toml`? If yes, my PR already does that:

image

I interpret the `main` here means the `main` branch...

sdk/typespec/typespec_client_core/src/async_runtime/mod.rs
LarryOsterman: I can totally get behind a `wasm-bindgen` feature name. 

Thank you for your patience here :)

magodo: @LarryOsterman Sure, I've pushed a new commit to introduce that feature name, made the wasm dependencies optional, and only bring them by this new feature.

LarryOsterman:
> Kindly ping @LarryOsterman

Thank you for the pin. I have to finish the July SDK release and after that, I'll get to this PR. Overall it looks great

LarryOsterman:
There's a single outstanding CI error:
```
2025-07-22T19:07:11.2088950Z     Checking typespec_client_core v0.6.0 (/mnt/vss/_work/1/s/sdk/typespec/typespec_client_core)
2025-07-22T19:07:11.7899314Z error[E0433]: failed to resolve: use of unresolved module or unlinked crate `web_runtime`
2025-07-22T19:07:11.7900135Z    --> sdk/typespec/typespec_client_core/src/async_runtime/mod.rs:197:18
2025-07-22T19:07:11.7900579Z     |
2025-07-22T19:07:11.7901006Z 197 |         Arc::new(web_runtime::WasmBindgenRuntime) as Arc<dyn AsyncRuntime>
2025-07-22T19:07:11.7901468Z     |                  ^^^^^^^^^^^ use of unresolved module or unlinked crate `web_runtime`
2025-07-22T19:07:11.7901872Z     |
2025-07-22T19:07:11.7902289Z     = help: if you wanted to use a crate named `web_runtime`, use `cargo add web_runtime` to add it to your `Cargo.toml`
2025-07-22T19:07:11.7902626Z 
```

If you could please fix this, I'll get this committed.

magodo:
@heaths Sure, I'll work on a test for this.

magodo:
@heaths and @LarryOsterman, the tests are ready at my own branch, which I've created another PR (since I can't push to this branch for some reason): https://github.com/Azure/azure-sdk-for-rust/pull/2838.

Close this PR in favor of https://github.com/Azure/azure-sdk-for-rust/pull/2838

PR #2772, opened 2025-07-08
Update codeowners

RickWinter:


heaths (approved):

.github/CODEOWNERS
```diff
@@ -25,7 +25,7 @@
 /**                      @heaths @RickWinter @ronniegeraghty @LarryOsterman
 
 # Allow service owners to approve crate version updates, which impact Cargo.lock
-/Cargo.lock
+/Cargo.lock              @Azure/azure-sdk-write
```
weshaggard, line 28: Did you make this change because of the linter complaining? If so it might be better to baseline that error instead of adding azure-sdk-write here as that will cause everyone to be mentioned and added as a reviewer to any PRs that touch this file.

heaths: Because of a PR I just created, someone just pinged me telling me not to @mention `azure-sdk-write`. How do we do this baseline thing?

heaths: #2774

PR #2774, opened 2025-07-08
Remove large team owner and baseline CODEOWNER lint

heaths:


RickWinter (approved):
This is a reasonable workaround for now. We should work to get rid of the baseline file ASAP

PR #2775, opened 2025-07-09
Live ACI and IMDS managed identity tests

chlowell:
The IMDS test works by running the whole suite on a CI agent with an assigned managed identity. The ACI test (and future tests for all other platforms) must run remotely, so for that one we deploy a container instance running a simple REST API that attempts to authenticate and access a Storage account.

### DIFF
--- a/sdk/identity/test-resources-post.ps1
+++ b/sdk/identity/test-resources-post.ps1
@@ -44,7 +44,7 @@ if ($CI) {
 Set-Location "$(git rev-parse --show-toplevel)/sdk/identity/azure_identity/tests/tools/deployed_live_test"
 
 Write-Host "##[group]Building test app"
-cargo install --path . --root .
+cargo install --path . --root target
 Write-Host "##[endgroup]"
 
 Write-Host "##[group]Building container image"
@@ -53,7 +53,7 @@ $image = "$($DeploymentOutputs['IDENTITY_ACR_LOGIN_SERVER'])/live-test"
 Set-Content -Path Dockerfile -Value @"
 FROM mcr.microsoft.com/mirror/docker/library/ubuntu:24.04
 RUN apt update && apt install ca-certificates --no-install-recommends -y
-COPY bin/deployed_live_test .
+COPY target/bin/deployed_live_test .
 CMD ["./deployed_live_test"]
 "@
 docker build -t $image .


sdk/identity/test-resources.bicep
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+@minLength(6)
+@maxLength(23)
+@description('The base resource name.')
+param baseName string = resourceGroup().name
+
+@description('Whether to deploy resources. When set to false, this file deploys nothing.')
+param deployResources bool = false
```
heaths, line 10: With every resource conditioned on `deployResources`, what's the point in even defining the conditionality? Wouldn't it make more sense just not to run the script if you don't need the resources? I'm trying to understand why have the entire file conditioned and in what circumstances you'd pass `true` here.

sdk/identity/azure_identity/tools/deployed_live_test/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "deployed_live_test"
+version = "0.1.0"
+edition = "2021"
+
```
heaths:
```suggestion
publish = false
```

sdk/identity/azure_identity/tools/deployed_live_test/Cargo.toml
heaths, line 1: If this is for testing, like our macros crates, let's put this under `tests/` e.g. `tests/tools/deployed_live_test`.

sdk/identity/azure_identity/tools/deployed_live_test/src/main.rs
```diff
@@ -0,0 +1,104 @@
...
+#[derive(Deserialize)]
+struct Params {
+    #[serde(rename = "client-id")]
+    client_id: Option<String>,
+    #[serde(rename = "object-id")]
+    object_id: Option<String>,
+    #[serde(rename = "resource-id")]
+    resource_id: Option<String>,
+    #[serde(rename = "storage-name")]
+    storage_name: String,
```
heaths: Tip: you can save yourself some typing by just putting `#[serde(rename_all = "kebab-case")]` on the struct instead of on every member. We do this in generation because it's easier but hand-written code would look cleaner like this.

sdk/identity/test-resources-post.ps1
```diff
@@ -0,0 +1,76 @@
...
+    return
+  }
+  az cloud set -n $Environment
+  az login --federated-token $env:ARM_OIDC_TOKEN --service-principal -t $TenantId -u $TestApplicationId
+  az account set --subscription $SubscriptionId
+}
+
+Write-Host "##[group]Building test app"
+Set-Location "$(git rev-parse --show-toplevel)/sdk/identity/azure_identity/tools/deployed_live_test"
+cargo install --path . --root .
```
heaths: If this is run locally, we run the risk of publishing a lot of cruft. Add a local `.gitignore` file to the parent directory of the resulting `bin/` and whatever other directories may be created. See `sdk/typespec/typespec_macros/tests/data/safe-debug-tests/` for an example of how we've been doing this elsewhere.

sdk/identity/test-resources-post.ps1
```diff
@@ -0,0 +1,76 @@
...
+    Write-Host "Skipping post-provisioning script because resources weren't deployed"
+    return
+  }
+  az cloud set -n $Environment
+  az login --federated-token $env:ARM_OIDC_TOKEN --service-principal -t $TenantId -u $TestApplicationId
+  az account set --subscription $SubscriptionId
+}
+
+Write-Host "##[group]Building test app"
+Set-Location "$(git rev-parse --show-toplevel)/sdk/identity/azure_identity/tools/deployed_live_test"
```
heaths: Nit: this `pushd` seems to be used below this group as well so you might move it out.

sdk/identity/test-resources-post.ps1
```diff
@@ -0,0 +1,76 @@
...
+
+Write-Host "##[group]Building container image"
+az acr login -n $DeploymentOutputs['IDENTITY_ACR_NAME']
+$image = "$($DeploymentOutputs['IDENTITY_ACR_LOGIN_SERVER'])/live-test"
+Set-Content -Path Dockerfile -Value @"
+FROM mcr.microsoft.com/mirror/docker/library/ubuntu:24.04
+RUN apt update && apt install ca-certificates --no-install-recommends -y
+COPY bin/deployed_live_test .
+CMD ["./deployed_live_test"]
+"@
```
heaths: Any reason not to just check in a `Dockerfile`? Easier to manage and diff.

sdk/identity/azure_identity/src/managed_identity_credential.rs
```diff
@@ -256,6 +281,25 @@ mod tests {
         );
     }
 
+    #[tokio::test]
```
heaths: These probably should be `#[recorded::test(live)]` so they are properly marked as "ignored" when not explicitly run with `AZURE_TEST_MODE=live`. That macro will resolve to `#[tokio::test]` when "live" is passed per your weekly run configuration in `cy.yml`.

heaths (changes requested):

sdk/identity/test-resources-post.ps1
chlowell: The Dockerfile isn't really useable outside of this script i.e., `docker build .` wouldn't just work. Even if the Dockerfile built the crate, there'd still be a gotcha with the build context due to the crate's relative path dependencies

sdk/identity/test-resources.bicep
chlowell: This is future-proofing. We'll eventually have live tests that don't need deployed resources. I'll want to run those tests in all internal pipelines, and the tests that need deployed resources only in the weekly pipeline. When I set all this up for Go, pipelines always deployed resources when `RunLiveTests: true`. Adding a conditional to the bicep resources that's true only in the weekly pipeline (see `managed-identity-matrix.json`) makes the deployment step a no-op in other pipelines. So far as I know, we still need a workaround like this; please let me know if you have other ideas

sdk/identity/test-resources-post.ps1
chlowell: I changed the install path to `target` so the artifacts are excluded by our root `.gitignore`

heaths (approved):

heaths:

heaths:

PR #2790, opened 2025-07-14
Add TryFrom implementations used in Spector tests

antkmsft:
The Spector test using the syntax: https://github.com/Azure/typespec-rust/pull/502


LarryOsterman:
I'm going to ask an ugly question: Do we really need to support a `Decimal` type for any scenarios other than Spector? 

Could this be implemented using a `decimal` feature to make the dependency an optional dependency so we don't add this package (and its half dozen or so unique dependencies) to the SDK? That way service clients which need the Decimal type can add the typespec_client_core with the `decimal` feature but other clients won't have the overhead.

sdk/typespec/typespec_client_core/src/http/request/mod.rs
```diff
@@ -291,6 +295,312 @@ impl<T> TryFrom<&'static str> for RequestContent<T> {
...
+        Ok(Self {
+            body: Bytes::from(format!(r#"{{{}}}"#, body_str)).into(),
+            phantom: PhantomData,
+        })
+    }
+}
+
+impl<T> TryFrom<HashMap<String, i64>> for RequestContent<T> {
+    type Error = crate::Error;
+    fn try_from(body: HashMap<String, i64>) -> Result<Self, Self::Error> {
```
LarryOsterman: This looks suspiciously like JSON encoding of the `HashMap`, is there possibly a more appropriate `serde_json` mechanism for encoding `HashMap`s? 

And if serde_json's serialization doesn't work, it would be nice to have a comment explaining why we couldn't use serde_json to serialize these elements.

antkmsft: Thank you very much for your feedback, it's been really helpful. I addressed everything according to your suggestions - I made decimal an optional dependency and put it behind a feature, and also I am using serde_json now which has also simplified serialization code a lot. Please check out the latest iteration.

sdk/typespec/typespec_client_core/src/http/request/mod.rs
```diff
@@ -291,6 +297,221 @@ impl<T> TryFrom<&'static str> for RequestContent<T> {
     }
 }
 
+impl<T> TryFrom<Value> for RequestContent<T> {
```
jhendrixMSFT: How do we differentiate between JSON and XML payloads? Perhaps these helpers need to be in a `json` module? Not that XML support for these scenarios is a real thing at present, but it could be, so we should be prepared for that.

sdk/typespec/typespec_client_core/src/http/request/mod.rs
```diff
@@ -291,6 +297,221 @@ impl<T> TryFrom<&'static str> for RequestContent<T> {
...
+    fn try_from(body: Value) -> Result<Self, Self::Error> {
+        Ok(Self {
+            body: Bytes::from(serde_json::to_string(&body)?).into(),
+            phantom: PhantomData,
+        })
+    }
+}
+
+impl<T> TryFrom<bool> for RequestContent<T> {
+    type Error = crate::Error;
```
LarryOsterman: The `Error` type should probably be `Infallible` rather than `crate::Error` because these functions cannot fail. See also: https://doc.rust-lang.org/std/convert/enum.Infallible.html

sdk/typespec/typespec_client_core/Cargo.toml
```diff
@@ -58,6 +59,7 @@ reqwest_rustls = [
 test = [] # Enables extra tracing including error bodies that may contain PII.
 tokio = ["tokio/fs", "tokio/sync", "tokio/time", "tokio/io-util"]
 xml = ["dep:quick-xml"]
+rust_decimal = ["dep:rust_decimal"]
```
LarryOsterman: Nit: You probably don't need to use the crate name as the feature name - you can just use `decimal` instead (similar to how we add `http` and `json` to express constructs.

sdk/typespec/typespec_client_core/src/http/request/mod.rs
LarryOsterman: The same applies to all the other functions which cannot fail.

sdk/typespec/typespec_client_core/src/http/request/mod.rs
antkmsft: Thank you - I think XML is totally valid angle to consider this from.
I think moving that into json module would clear ambiguities.
But I think the best would be if the client could request a payload in a format it needs, depending on which protocol it uses, which the user should not know.
I.e., in a vacuum, I'd have a hierarchy of types and a builder/visitor which visits them depending on the context and creates a string representation that it needs. I.e. the value goes to HTTP Headers - it is one format (with some types of values not supported, maybe). Query params - another, Path - yet another, JSON - yet another, XML - yet another. A client would create an instance of such visitor/builder and apply it to user parameters, because client knows where the value is going, and which protocol/format it is supposed to use when talking to Azure. Do you think it is reasonable/possible to implement something like that?
In the meanwhile, let me to move this to JSON module.

sdk/typespec/typespec_client_core/Cargo.toml
antkmsft: Thank you! Done.

sdk/typespec/typespec_client_core/src/http/request/mod.rs
antkmsft: Thank you! Done.

sdk/typespec/typespec_client_core/src/http/request/mod.rs
antkmsft: I have moved the TryFroms that use serde_json to a `pub mod json` - is this what you've meant, could you please review?

jhendrixMSFT: Yeah this is what I was thinking and is good for now.

LarryOsterman (approved):
This looks good to me.

I *might* have moved the json module to a separate .rs file just to simplify the `use` statements at the top of the file, but that doesn't really rise to the level of a "nit".

antkmsft:
Thank you @LarryOsterman - I think I am going to keep it in this file, and I was also thinking the separate file, but I kind of like that one TryFrom is next to another TryFrom, for maintenance purposes. But should it be in a separate file, it wouldn't've been complicated either, it has its own benefits.

RickWinter:
@antkmsft Can you resolve the conflict and merge this change.

sdk/typespec/typespec_client_core/Cargo.toml
```diff
@@ -58,6 +59,7 @@ reqwest_rustls = [
 test = [] # Enables extra tracing including error bodies that may contain PII.
 tokio = ["tokio/fs", "tokio/sync", "tokio/time", "tokio/io-util"]
 xml = ["dep:quick-xml"]
+decimal = ["dep:rust_decimal"]
```
heaths, line 62: Why is this optional? Is it so big and pervasive it needs to be a feature?

sdk/typespec/typespec_client_core/src/http/request/mod.rs
```diff
@@ -306,6 +312,225 @@ impl<T> TryFrom<&'static str> for RequestContent<T> {
...
+    type Error = Infallible;
+    fn try_from(body: Option<Decimal>) -> Result<Self, Infallible> {
+        Ok(Self {
+            body: Bytes::from(body.map(|d| d.to_string()).unwrap_or_default()).into(),
+            phantom: PhantomData,
+        })
+    }
+}
+
+pub mod json {
```
heaths, line 336: What's the reason to put this in a public child model? It should also be gated on the `json` feature, but in that case you'd do something like:

```rust
#[cfg(feature = "json")]
mod json {
  // ...
}

#[cfg(feature = "json")]
pub use json::*;
```

The module is "private" but exposed publicly as merged. Fewer conditions in source, but I don't see why this needs to be a public submodule.

sdk/typespec/typespec_client_core/Cargo.toml
heaths: Ah, I see @LarryOsterman requested this for good reason: https://github.com/Azure/azure-sdk-for-rust/pull/2790#issuecomment-3071320416

sdk/typespec/typespec_client_core/src/http/request/mod.rs
antkmsft: Thank you, I am addressing this in #2833

PR #2791, opened 2025-07-14
[Storage] Prepare `azure_storage_blob` for `v0.3.0` release

vincenttran-msft:
This PR adds the following:

- Adds the changelog entries for the `v0.3.0` release
- Audits and re-exports all required usability `...ResultHeaders` for currently implemented features
- Regenerates against the following [.tsp ](https://github.com/Azure/azure-rest-api-specs/pull/35886) which eliminates using the clientName decorator to rename, and instead propagates the change to `routes.tsp`
- ^ The above `.tsp` also contains the changes to move from the `AccessTierOptional` enum to `AccessTier`


jalauzon-msft (approved):

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -230,7 +230,7 @@ impl BlobClient {
     /// * `options` - Optional configuration for the request.
     pub async fn set_tier(
         &self,
-        tier: AccessTierOptional,
+        tier: AccessTier,
```
LarryOsterman, line 233: Does this type change count as a breaking change that should be called out?

LarryOsterman (changes requested):
The only issue I see is that the type change should probably be called out as a breaking change - since the tests were updated, it implies that this is something customer visible.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: Hi @LarryOsterman good callout here. We never actually released with `AccessTierOptional`, we only got changed to that in the code-gen as a result of some efforts to get CI-clean in the feature/blob-tsp branch. 

We were trying to accommodate unblocking both CIs while iterating, but now that we are releasing another version of the SDK and no final decision has been made for `feature/blob-tsp`, we will revert to what we originally had (and released with in `v0.2.0`), knowing that it reintroduces a break in the `feature/blob-tsp` branch. This is understood and accepted as `feature/blob-tsp` has several other CI-blocking issues from getting that merged to `main`.

Please let me know if that you have any other questions!

LarryOsterman (approved):

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
LarryOsterman: Nope - I'm good. Other than that one issue (and right now, breaking changes are just fine as long as we call them out in the changelog) it looks fine.

PR #2795, opened 2025-07-15
[Storage] `get_account_info` all clients, `set/get_tags` for BlobClient, `set_properties` for BlobServiceClient

vincenttran-msft:
`.tsp`: https://github.com/Azure/azure-rest-api-specs/pull/36677
	
**BlobClientSetTagsResult**
	•  ̶ ̶D̶a̶t̶e̶

**BlobClientGetAccountInfoResult**
	• Date
	• Account_kind
	• Is_hns_enabled
	• Sku_name

**BlobContainerClientGetAccountInfoResult**
	•  ̶ ̶D̶a̶t̶e̶
	• Account_kind
	• Is_hns_enabled
	• Sku_name

**BlobServiceClientGetAccountInfoResult**
	• ̶ ̶D̶a̶t̶e̶
	• Account_kind
	• Is_hns_enabled
	• Sku_name


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -313,4 +316,69 @@ impl BlobClient {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `copy_source` - A URL of up to 2 KB in length that specifies a file or blob.
+    ///   The value should be URL-encoded as it would appear in a request URI.
+    ///   If the source is in another account, the source must either be public
+    ///   or must be authenticated via a shared access signature. If the source
+    ///   is public, no authentication is required.
+    /// * `options` - Optional configuration for the request.
+    pub async fn start_copy_from_url(
```
jalauzon-msft: As discussed, let's remove copy for now because I want to discuss more about how we want to handle it.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -313,4 +316,69 @@ impl BlobClient {
...
+
+    /// Gets the tags on a blob.
+    ///
+    /// # Arguments
+    ///
+    /// * `options` - Optional configuration for the request.
+    pub async fn get_tags(
+        &self,
+        options: Option<BlobClientGetTagsOptions<'_>>,
+    ) -> Result<Response<BlobTags, XmlFormat>> {
```
jalauzon-msft, line 341: If we are accepting a `HashMap` in `set_tags` we really should be returning one too. This `BlobTags` model is not as friendly to use.

vincenttran-msft:
Also, I believe I'm stumbling upon a pretty fun issue 😄 
Since we refactored `set_tags` to accept a `HashMap<String, String>` and then call our helper to serialize it into the proper form, `BlobTags`, due to the unordered nature of `HashMap`s, the resultant `BlobTags` object can have two different forms, and this will trigger a test recording failure:
Scenario 1:
```xml
<Tags>
	<TagSet>
		<Tag>
			<Key>hello</Key>
			<Value>world</Value>
		</Tag>
		<Tag>
			<Key>ferris</Key>
			<Value>crab</Value>
		</Tag>
	</TagSet>
</Tags>",
```

Scenario 2:
```xml
<Tags>
	<TagSet>
		<Tag>
			<Key>ferris</Key>
			<Value>crab</Value>
		</Tag>
		<Tag>
			<Key>hello</Key>
			<Value>world</Value>
		</Tag>
	</TagSet>
</Tags>",
```

I am wondering if this is something we can resolve at the test proxy level (i.e. sort alphabetically before comparing) at least in areas where order shouldn't matter.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -423,3 +424,54 @@ async fn test_leased_blob_operations(ctx: TestContext) -> Result<(), Box<dyn Err
...
+    // Set Tags with Tags Specified
+    let blob_tags = HashMap::from([
+        ("hello".to_string(), "world".to_string()),
+        ("ferris".to_string(), "crab".to_string()),
+    ]);
+    blob_client.set_tags(blob_tags.clone(), None).await?;
+
+    // Assert
+    let response_tags = blob_client.get_tags(None).await?.into_body().await?;
+    println!("response:{:?}", response_tags.clone());
```
vincenttran-msft: nit: Remove on next iteration once we land on design for test proxy issue

sdk/storage/azure_storage_blob/src/parsers.rs
```diff
@@ -35,3 +39,49 @@ pub fn format_page_range(offset: u64, length: u64) -> Result<String, Error> {
...
+    let headers = response.headers().clone();
+    let blob_tags = response.into_body().await?;
+
+    if let Some(blob_tag_set) = blob_tags.blob_tag_set {
+        for blob_tag in blob_tag_set {
+            blob_tags_map.insert(blob_tag.key, blob_tag.value);
+        }
+    }
+
+    let xml_body = xml::to_xml_with_root("HashMap", &blob_tags_map)?;
```
jalauzon-msft: This feels pretty hacky to put it back into XML. I'm not sure if `Response` has a mechanism for us to say, we've already desterilized, here is the result so when someone calls `into_body` it just gives that back. Or perhaps there is a different way to transform this response with adding a custom deserialization method? Not sure.

heaths: I'm not sure why most of this is being done anyway. Implement `Serialize` and/or `Deserialize` for your custom model (outside of "generated"). See https://serde.rs for details, or you can find existing code. Don't try to handle it this way because, yes, it gets messy like this.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -304,4 +308,52 @@ impl BlobClient {
...
+    ///
+    /// * `tags` - Name-value pairs associated with the blob as tag. Tags are case-sensitive.
+    ///   The tag set may contain at most 10 tags.  Tag keys must be between 1 and 128 characters,
+    ///   and tag values must be between 0 and 256 characters.
+    ///   Valid tag key and value characters include: lowercase and uppercase letters, digits (0-9),
+    ///   space (' '), plus (+), minus (-), period (.), solidus (/), colon (:), equals (=), underscore (_)
+    /// * `options` - Optional configuration for the request.
+    pub async fn set_tags(
+        &self,
+        tags: HashMap<String, String>,
```
heaths, line 324: Why isn't this a `RequestContent<T>`? There's reasons that exist.

heaths (changes requested):
Implement `Serialize` and/or `Deserialize` (probably both, it seems) for your custom tags model and much of this code goes away.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: See https://github.com/Azure/typespec-rust/issues/533, but currently the reasoning is that the generated code expects a `RequestContent<BlobTags>`, so this lightweight convenience layer abstracts away the complexity.

vincenttran-msft:
> Also, I believe I'm stumbling upon a pretty fun issue 😄 Since we refactored `set_tags` to accept a `HashMap<String, String>` and then call our helper to serialize it into the proper form, `BlobTags`, due to the unordered nature of `HashMap`s, the resultant `BlobTags` object can have two different forms, and this will trigger a test recording failure: Scenario 1:
> 
> ```
> <Tags>
> 	<TagSet>
> 		<Tag>
> 			<Key>hello</Key>
> 			<Value>world</Value>
> 		</Tag>
> 		<Tag>
> 			<Key>ferris</Key>
> 			<Value>crab</Value>
> 		</Tag>
> 	</TagSet>
> </Tags>",
> ```
> 
> Scenario 2:
> 
> ```
> <Tags>
> 	<TagSet>
> 		<Tag>
> 			<Key>ferris</Key>
> 			<Value>crab</Value>
> 		</Tag>
> 		<Tag>
> 			<Key>hello</Key>
> 			<Value>world</Value>
> 		</Tag>
> 	</TagSet>
> </Tags>",
> ```
> 
> I am wondering if this is something we can resolve at the test proxy level (i.e. sort alphabetically before comparing) at least in areas where order shouldn't matter.

For this issue, @heaths Heath do you have any examples of using a `CustomDefaultMatcher` with `compareBodies = false` in Rust test proxy? I think that should sufficiently resolve this issue, just unsure how to enable that knob in Rust. Thanks!

sdk/storage/azure_storage_blob/src/parsers.rs
vincenttran-msft: As discussed offline- should be resolved once `RequestContent` refactor is in.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -423,3 +424,53 @@ async fn test_leased_blob_operations(ctx: TestContext) -> Result<(), Box<dyn Err
...
+
+    // Set Tags with Tags Specified
+    let blob_tags = HashMap::from([
+        ("hello".to_string(), "world".to_string()),
+        ("ferris".to_string(), "crab".to_string()),
+    ]);
+    blob_client.set_tags(blob_tags.clone(), None).await?;
+
+    // Assert
+    let response_tags = blob_client.get_tags(None).await?.into_body().await?;
```
vincenttran-msft, line 445: This is the error I am hitting currently with the revised `RequestContent`:
>Error: Error { context: Full(Custom { kind: DataConversion, error: ExpectedStart }, "failed to deserialize the following xml into a std::collections::hash::map::HashMap<alloc::string::String, alloc::string::String>\n{\"hello\":\"world\",\"ferris\":\"crab\"}") }

sdk/storage/azure_storage_blob/src/parsers.rs
```diff
@@ -35,3 +40,51 @@ pub fn format_page_range(offset: u64, length: u64) -> Result<String, Error> {
...
+            if let (Some(k), Some(v)) = (tag.key, tag.value) {
+                blob_tags_map.insert(k, v);
+            }
+        }
+    }
+
+    let request_content: RequestContent<HashMap<String, String>> =
+        RequestContent::try_from(blob_tags_map)?;
+    let raw_response = RawResponse::from_bytes(status, headers, request_content.body());
+    Ok(raw_response.into())
```
vincenttran-msft: Here is the relevant "repacking" code. @heaths Heath, any pointers on what I could be doing wrong? Facing the following error:
>Error: Error { context: Full(Custom { kind: DataConversion, error: ExpectedStart }, "failed to deserialize the following xml into a std::collections::hash::map::HashMap<alloc::string::String, alloc::string::String>\n{\"hello\":\"world\",\"ferris\":\"crab\"}") }


This code is heavily cribbed off the pseudocode you provided offline, but I could not call `RawResponse::from_bytes(status, headers, request_content.try_into()?)` as that gave the following error:
>error[E0277]: the trait bound `azure_core::Bytes: std::convert::From<()>` is not satisfied
  --> sdk\storage\azure_storage_blob\src\parsers.rs:88:65
   |
88 |     let raw_response = RawResponse::from_bytes(status, headers, request_content.try_into()?);
   |                        -----------------------                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::convert::From<()>` is not implemented for `azure_core::Bytes`

so I instead had to pass the `bytes` parameter the `request_content.body()`

sdk/storage/azure_storage_blob/tests/blob_client.rs
heaths: That's JSON, and the error clearly states it's expecting XML. You need to rebase on my latest changes to pick up the `XmlFormat` changes I made for Storage.

I'll do it, since I have your PR branch checked out and am working on recommendations.

heaths: I'll need to rebase on #2858 once that's merged.

heaths: It's merged, so fetch `upstream/main` and rebase this PR. Be sure your TSP contains any fixes you and and regenerate source. Probably going to be easier to just squash all these commits because several didn't build, and I suspect didn't even before the rebase.

sdk/storage/azure_storage_blob/src/parsers.rs
```diff
@@ -35,3 +40,51 @@ pub fn format_page_range(offset: u64, length: u64) -> Result<String, Error> {
...
```
vincenttran-msft: This works, but is still a little clunky. The generated code returns a `Response<BlobTags, XmlFormat>`.  We are using `RequestContent::try_from(blob_tags_map)?;` which is implemented as returning `RequestContent<T, F=JsonFormat>`. Therefore, when I use the `.body()` to pass to the repacked `RawResponse`, it will be `Json` and so we need to change our deserializer (and thus response object) to be a:
`Result<Response<HashMap<String, String>, JsonFormat>>`

vincenttran-msft: @heaths Hi Heath, any recommendations to get out of this format change? I am assuming that I am just missing something to get a `RequestContent<T=HashMap<String, String>, F=XmlFormat>`

heaths: I didn't add a `TryFrom` for `XmlFormat`. You'll need to do that in your crate. You're the only XML customers, so it's not something we'd add to core at this time because that's additional public API surface we'd have to maintain forever. You also need to specify the format you want for `RequestContent<T, F>`; otherwise, `F` defaults to `JsonFormat` e.g.,

```rust
impl TryFrom<HashMap<String, String>> for RequestContent<HashMap<String, String>, XmlFormat> {
  // ...
}
```

jalauzon-msft: @heaths, I would argue that Core should have XML support. We can't be the only customer in all of Azure that uses XML and even if that's true, it feels like something that should be part of a library like Core.

heaths: You're the only service *currently* that uses XML, though I hope that list expands in the future. That said, as we discussed offline, the orphan rule pretty much necessitates we add the `XmlFormat` variations to core so I'll do that.

vincenttran-msft: As discussed offline (again 😆) it is impossible for XML to be generically implemented because of the root tag.

sdk/storage/azure_storage_blob/src/models/extensions.rs
```diff
@@ -38,3 +41,42 @@ impl AppendBlobClientCreateOptionsExt for AppendBlobClientCreateOptions<'_> {
...
+            .map(|(k, v)| BlobTag {
+                key: Some(k),
+                value: Some(v),
+            })
+            .collect();
+        BlobTags {
+            blob_tag_set: Some(blob_tags),
+        }
+    }
+}
```
vincenttran-msft: Added a `TryFrom` and `From` to be used for `get_tags` and `set_tags` respectively.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -442,14 +442,16 @@ async fn test_blob_tags(ctx: TestContext) -> Result<(), Box<dyn Error>> {
 
     // Assert
     let response_tags = blob_client.get_tags(None).await?.into_body().await?;
-    assert_eq!(blob_tags, response_tags);
+    let map: HashMap<String, String> = response_tags.try_into()?;
+    assert_eq!(blob_tags, map);
```
vincenttran-msft: Usage example- so 1 extra step `try_into()` to get the user-friendly `HashMap<String, String>` format.

sdk/storage/azure_storage_blob/src/parsers.rs
```diff
@@ -1,6 +1,11 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+use crate::generated::models::{BlobTag, BlobTags};
+use azure_core::http::response::ResponseBody;
+use azure_core::http::{JsonFormat, RawResponse, RequestContent, Response, XmlFormat};
+use azure_core::xml;
+use std::collections::{BTreeMap, HashMap};
```
jalauzon-msft: Remove unused imports.

jalauzon-msft (approved):

sdk/storage/azure_storage_blob/src/models/extensions.rs
```diff
@@ -38,3 +41,42 @@ impl AppendBlobClientCreateOptionsExt for AppendBlobClientCreateOptions<'_> {
         }
     }
 }
+
+/// Converts a `BlobTags` struct into `HashMap<String, String>`.
+impl TryFrom<BlobTags> for HashMap<String, String> {
+    type Error = &'static str;
```
heaths: This must be an `azure_core::Error`.

sdk/storage/azure_storage_blob/src/models/extensions.rs
```diff
@@ -38,3 +41,42 @@ impl AppendBlobClientCreateOptionsExt for AppendBlobClientCreateOptions<'_> {
...
+    fn try_from(blob_tags: BlobTags) -> Result<Self, Self::Error> {
+        let mut map = HashMap::new();
+
+        if let Some(tags) = blob_tags.blob_tag_set {
+            for tag in tags {
+                match (tag.key, tag.value) {
+                    (Some(k), Some(v)) => {
+                        map.insert(k, v);
+                    }
+                    _ => return Err("BlobTag missing key or value"),
```
heaths:
```suggestion
                    _ => return Err(azure_core::Error::new(ErrorKind::DataConversion, "BlobTag missing key or value")),
```

sdk/storage/azure_storage_blob/src/models/extensions.rs
```diff
@@ -38,3 +41,42 @@ impl AppendBlobClientCreateOptionsExt for AppendBlobClientCreateOptions<'_> {
...
+        }
+
+        Ok(map)
+    }
+}
+
+/// Converts a `HashMap<String, String>` into a `BlobTags` struct.
+impl From<HashMap<String, String>> for BlobTags {
+    fn from(tags: HashMap<String, String>) -> Self {
+        let sorted_tags: BTreeMap<_, _> = tags.into_iter().collect();
```
heaths: 1. Why use a `BTreeMap` for this when you can do this more efficiently with a `Vec::with_capacity`:

```rust
    let mut map: HashMap<String, String> = HashMap::new();
    map.insert("foo".to_string(), "1".to_string());
    map.insert("bar".to_string(), "2".to_string());

    let mut v: Vec<(String, String)> = Vec::with_capacity(map.len());
    v.extend(map);
    v.sort_by(|(k1, _), (k2, _)| k1.cmp(k2));
```

2. Why sort them anyway? If for testing, then only sort in test. This is a waste of time and memory for customers if not otherwise required.

vincenttran-msft: Hi Heath, good callout here, I believed I was using `BTreeMap` from a previous suggestion you made, but it was asked out of context and agree that sorting here in prod is unnecessary.

However, I am still struggling to figure out how to go about this without disabling body matching. Your above suggestion:

```rust
#[cfg(test)]
let mut tags = std::collections::BTreeMap::new();
#[cfg(not(test))]
let mut tags = std::collections::HashMap::new();
```

I don't believe is applicable to us, since these are integration tests and `#[cfg(test)]` is used to conditionally compile unit tests. 
Thus, I am still unsure how we could conditionally execute different code while keeping the production code intact i.e. a mock or monkey patch of some sort?

sdk/storage/azure_storage_blob/src/models/extensions.rs
```diff
@@ -4,6 +4,7 @@
 use crate::models::{
     AppendBlobClientCreateOptions, BlobTag, BlobTags, PageBlobClientCreateOptions,
 };
+use azure_core::error::ErrorKind;
 use std::collections::{BTreeMap, HashMap};
```
heaths: You're still importing `BTreeMap` unnecessarily, which should cause a build break (lint warning as error).

PR #2797, opened 2025-07-16
Make native-tls dependency optional for azure_messaging_eventhubs

### LINKED ISSUE #2796: `azure_messaging_eventhubs` should allow users to use `rustls` instead of `native-tls`
### Feature Summary

Currently, the `azure_messaging_eventhubs` requires `native-tls` (OpenSSL) for TLS connections.  This creates significant friction for cross-platform Rust applications and libraries.

### Feature Description

**Summary:**
Currently, the `azure_messaging_eventhubs` requires `native-tls` (OpenSSL) for TLS connections. This dependency is introduced by: `azure_messaging_eventhubs` -> `azure_core_amqp` -> `fe2o3-amqp` -> `native-tls`. `fe2o3-amqp` is configured as part of the workspace dependency [here](https://github.com/Azure/azure-sdk-for-rust/blob/f12119abe9da8941f1157d7e91bb9bee25cda639/Cargo.toml#L78), and users have no way to remove it, as it's not guarded by a feature. This creates significant friction for cross-platform Rust applications and libraries. Many Rust projects—including ours—use `rustls` for TLS because it is pure Rust, easier to build, and more portable (especially for static builds, containerized deployments, and platforms where OpenSSL is problematic).

**Request:**
Please add support for using `rustls` as a TLS backend, controlled by a Cargo feature flag (e.g., `rustls-tls`, following the convention of crates like `reqwest` and `hyper`). This will allow users to choose their preferred TLS implementation and greatly improve the ease of building and distributing Rust applications that integrate with Azure services.

You can still by default enable the `native-tls` feature, so existing clients won't break, but allow users turn off defaults and enable `rustls` instead.

**Benefits:**

* Better cross-platform compatibility (especially for Windows, Mac, and minimal Linux containers)
* Simplified dependency management and deployment (no external C libraries)
* Aligns with modern Rust ecosystem practices

**Thank you for considering this feature—it would be a major improvement for Rust users integrating with Azure event hubs!**

### Use Case

_No response_

### Alternatives

_No response_

### Additional Context

_No response_

georgeh0:
resolves #2796

### DIFF
--- a/sdk/core/azure_core_amqp/Cargo.toml
+++ b/sdk/core/azure_core_amqp/Cargo.toml
@@ -49,6 +49,8 @@ fe2o3_amqp = [
   "serde_amqp",
   "serde_bytes",
 ]
+native_tls = ["fe2o3-amqp?/native-tls"]
+rustls = ["fe2o3-amqp?/rustls"]
 
 [lints]
 workspace = true
--- a/sdk/eventhubs/azure_messaging_eventhubs/Cargo.toml
+++ b/sdk/eventhubs/azure_messaging_eventhubs/Cargo.toml
@@ -43,8 +43,11 @@ tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
 tracing-subscriber = { workspace = true, features = ["env-filter", "fmt"] }
 
 [features]
+default = ["native_tls"]
 test = []
 blob_checkpoint_store = []
+native_tls = ["azure_core_amqp/native_tls"]
+rustls = ["azure_core_amqp/rustls"]
 
 [[bench]]
 name = "benchmarks"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,7 +75,7 @@ cargo_metadata = "0.18.1"
 clap = { version = "4.4.16", features = ["derive"] }
 criterion = { version = "0.5", features = ["async_tokio"] }
 dyn-clone = "1.0"
-fe2o3-amqp = { version = "0.14", features = ["native-tls", "tracing", "uuid"] }
+fe2o3-amqp = { version = "0.14", features = ["tracing", "uuid"] }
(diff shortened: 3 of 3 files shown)


georgeh0:
@microsoft-github-policy-service agree company="CocoIndex"

LarryOsterman:
Your change looks fine to me, but unfortunately our component governance system is complaining about the "ring" dependency that gets introduced with the rustls dependency.

sdk/core/azure_core_amqp/Cargo.toml
```diff
@@ -49,6 +49,8 @@ fe2o3_amqp = [
   "serde_amqp",
   "serde_bytes",
 ]
+native_tls = ["fe2o3-amqp?/native-tls"]
```
heaths, line 52: It doesn't seem there's a choice, but any `native-tls` feature would have to be sans provider like we do in `typespec_client_core`'s `reqwest_rustls` feature e.g., `reqwest/rustls-tls-native-roots-no-provider`. We absolutely cannot take a dependency on `ring`.

Now, it could be it's not actually used, but this would need to be investigated further.

sdk/eventhubs/azure_messaging_eventhubs/Cargo.toml
```diff
@@ -43,8 +43,11 @@ tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
 tracing-subscriber = { workspace = true, features = ["env-filter", "fmt"] }
 
 [features]
+default = ["native_tls"]
 test = []
 blob_checkpoint_store = []
+native_tls = ["azure_core_amqp/native_tls"]
```
heaths, line 49: Features cannot be mutually exclusive, so is there a reason to make `native_tls` a feature? Why not just `rustls` like we did in `azure_core`?

heaths (changes requested):

PR #2798, opened 2025-07-16
[Storage] SAS proof-of-concept

vincenttran-msft:
This PR serves as a proof of concept of how to support the SAS authentication scheme.

**Required Emitter Changes to Generated Code**

- Modify the struct to now contain `blob_url` instead of `endpoint`, which will now contain the fully qualified URI (including path parameters etc, query parameters, etc.)
- Make any necessary refactorings to support the above changes (per-API no longer needs path-building logic, adjust struct + getters + navigation methods)

**After the required emitter changes are done, the handwritten layer needs the following refactor:**

- Implement the required generated client constructors on the handwritten layer
- `new()` convenience constructors should _not_ use the underlying `new()` generated ctor 
- Implement other handwritten convenience constructors (i.e. `from_blob_url`)

### DIFF
--- a/sdk/storage/azure_storage_blob/src/clients/blob_client.rs
+++ b/sdk/storage/azure_storage_blob/src/clients/blob_client.rs
@@ -30,49 +30,6 @@ use azure_core::{
 };
 use std::sync::Arc;
 
-pub fn url_encode(s: impl AsRef<[u8]>) -> String {
-    url::form_urlencoded::byte_serialize(s.as_ref()).collect::<String>()
-}
-
-pub fn rebuild_blob_url(input: &str) -> Option<String> {
-    // Extract storage account
-    let storage_account = input
-        .split(".blob")
-        .next()
-        .or_else(|| input.split(".dfs").next())?
-        .split("//")
-        .last()?;
-
-    // Extract container name
-    let path_segments = input.split('/').collect::<Vec<&str>>();
-    let container_name = path_segments.get(3)?;
-
-    // Find final '?' that has '=' after it
-    let mut query_start_index = None;
-    for (i, c) in input.char_indices().rev() {
-        if c == '?' && input[i + 1..].contains('=') {
-            query_start_index = Some(i);
-            break;
-        }
-    }
-
-    let query_start = query_start_index?;
-    let query_string = &input[query_start..];
-
-    // Extract blob name
-    let container_pos = input.find(container_name)? + container_name.len() + 1;
-    let blob_raw = &input[container_pos..query_start];
-    let blob_name = url_encode(blob_raw);
-
-    // Rebuild the URL
-    let rebuilt_url = format!(
-        "https://{}.blob.core.windows.net/{}/{}{}",
-        storage_account, container_name, blob_name, query_string
-    );
-
-    Some(rebuilt_url)
-}
-
 /// A client to interact with a specific Azure storage blob, although that blob may not yet exist.
 pub struct BlobClient {
     pub(super) client: GeneratedBlobClient,
@@ -85,9 +42,6 @@ impl GeneratedBlobClient {
         options: Option<BlobClientOptions>,
     ) -> Result<Self> {
         let options = options.unwrap_or_default();
-        let blob_url = rebuild_blob_url(blob_url).expect("Something exploded in building the URL.");
-
-        println!("{}", blob_url.clone());
 
         let pipeline = match credential {
             Some(cred) => {
@@ -115,7 +69,7 @@ impl GeneratedBlobClient {
         Ok(Self {
             blob_name: "dummy".into(),
             container_name: "dummy".into(),
-            blob_url: Url::parse(&blob_url)?,
+            blob_url: Url::parse(blob_url)?,
             version: options.version,
             pipeline,
         })
@@ -156,6 +110,7 @@ impl BlobClient {
         }
 
         // In regular ctor, since struct now holds blob_url, we have to build the blob_url given endpoint + container_name + blob_name
+        // This is a URL tool so automatically encodes :D
         url.path_segments_mut()
             .expect("Cannot be base")
             .extend([&container_name, &blob_name]);
--- a/sdk/storage/azure_storage_blob/tests/blob_client.rs
+++ b/sdk/storage/azure_storage_blob/tests/blob_client.rs
@@ -430,7 +430,7 @@ use tokio::time;
 #[recorded::test]
 async fn test_sas(ctx: TestContext) -> Result<(), Box<dyn Error>> {
     // SAS
-    let blob_url = "SAS URL";
+    let blob_url = "<SAS URL>";
 
     let sas_blob_client = BlobClient::from_blob_url(blob_url, None)?;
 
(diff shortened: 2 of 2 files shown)


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -83,6 +83,32 @@ impl BlobClient {
         })
     }
 
+    pub fn new_no_credential(
```
vincenttran-msft: Need to expose a way to construct a credential-less handwritten abstraction of the clients.

sdk/storage/azure_storage_blob/src/generated/clients/blob_client.rs
```diff
@@ -102,6 +102,35 @@ impl BlobClient {
         })
     }
 
+    pub fn new_no_credential(
```
vincenttran-msft: GEN CODE SUPPORT: Credential-less `new()` that also doesn't call `endpoint.set_query(None)` and clobbers all query params.

sdk/storage/azure_storage_blob/src/generated/clients/blob_client.rs
```diff
@@ -770,10 +799,15 @@ impl BlobClient {
...
-        url = url.join(&path)?;
+        // let mut path = String::from("{containerName}/{blobName}");
+        // path = path.replace("{blobName}", &self.blob_name);
+        // path = path.replace("{containerName}", &self.container_name);
+        // url = url.join(&path)?;
+        {
+            let mut path_segments = url.path_segments_mut().expect("Cannot be base");
+            path_segments.push(&self.container_name);
+            path_segments.push(&self.blob_name);
+        }
```
vincenttran-msft: This is a sample of the refactor on per-method -- need to use `path_segments_mut()` to append the path segments rather than using `join()` (which also was clobbering everything but the base)

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: - Make this `from_blob_url`
- Should take a full Blob URL, no `container_name` or `blob_name` input
- Parse those elements out, pass along to generated layer
- For now, endpoint can still have query parameters

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -97,14 +92,35 @@ impl BlobClient {
...
+        let url_for_path_segments = url.clone();
+        if let Some(mut segments) = url_for_path_segments.path_segments() {
+            container_name = segments.next().unwrap();
+            blob_name = segments.next().unwrap();
+        }
+        url.set_path("");
+        // URL AFTER NULLED PATH: https://vincenttranstock.blob.core.windows.net/?sp=racwd&st=2025-07-21T21:22:28Z&se=2025-07-22T05:37:28Z&spr=https&sv=2024-11-04&sr=b&sig
+        // println!("URL AFTER NULLED PATH: {}", url.clone());
+
+        // Therefore, given the current design, the endpoint passed to the generated client will have query parameters on it.
```
vincenttran-msft: As noted here (and noted in a comment below) with the current design, this will work E2E. However, the one caveat to note is that the `endpoint` field on both our client and the generated client, will always contain the query parameters as well.
Therefore, calling it "endpoint" is not very accurate (and may trip up cx who are using it directly to build something using its value).

One possible idea, given that `endpoint` is only accessible to cx through the getter `endpoint()`:

- `.endpoint`, under the hood, will have the query params (i.e. SAS info) in the endpoint that is used to build the actual requests so that they can be successfully authenticated
- In the `endpoint()` functions we offer, we strip out the query parameters and return the endpoint as expected

vincenttran-msft: Or, alternatively, we just need to add a new way to pass along the query parameters and not as a free-rider alongside the `endpoint`

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -83,6 +83,44 @@ impl BlobClient {
...
+        let mut url = Url::parse(blob_url)?;
+
+        let mut segments = url.path_segments().expect("Failed to get path segments");
+        let container_name = segments
+            .next()
+            .expect("Failed to parse container_name")
+            .to_string();
+        let blob_name = segments.collect::<Vec<_>>().join("/");
+
+        url.set_path("");
```
vincenttran-msft, line 105: This is the work done to parse out `blob_name` and `container_name`

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -69,47 +68,63 @@ impl BlobClient {
...
-        PageBlobClient {
-            endpoint: self.client.endpoint.clone(),
-            client: self.client.get_page_blob_client(),
-        }
+        let url = Url::parse(blob_url)?;
+        let client = GeneratedBlobClient::with_no_credential(url.as_str(), Some(options))?;
+
+        Ok(Self {
+            client,
+        })
```
vincenttran-msft: This is the implementation of `from_blob_url` for the handwritten client.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -24,18 +24,99 @@ use azure_core::{
...
+            pipeline: Pipeline::new(
+                option_env!("CARGO_PKG_NAME"),
+                option_env!("CARGO_PKG_VERSION"),
+                options.client_options,
+                Vec::default(),
+                Vec::default(),
+            ),
+        })
+    }
+}
```
vincenttran-msft: As it turns out we can just implement out custom ctors like this. So regardless of what comes out of the emitter, we can simply do what we need to in our constructor and then use struct initialization.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -61,55 +142,75 @@ impl BlobClient {
...
-            endpoint: self.client.endpoint.clone(),
-            client: self.client.get_block_blob_client(),
-        }
+        // In regular ctor, since struct now holds blob_url, we have to build the blob_url given endpoint + container_name + blob_name
+        url.path_segments_mut()
+            .expect("Cannot be base")
+            .extend([&container_name, &blob_name]);
+
+        let client =
+            GeneratedBlobClient::from_url_with_credential(url.as_str(), credential, Some(options))?;
```
vincenttran-msft: In this instance, the regular convenience `new()` will use `from_url_with_credential` underlying

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -61,55 +142,75 @@ impl BlobClient {
...
+        let storage_headers_policy = Arc::new(StorageHeadersPolicy);
+        options
+            .client_options
+            .per_call_policies
+            .push(storage_headers_policy);
+
+        let url = Url::parse(blob_url)?;
+        let client = GeneratedBlobClient::from_url_with_no_credential(url.as_str(), Some(options))?;
+
+        Ok(Self { client })
```
vincenttran-msft: Example of exposed, handwritten `from_blob_url`. We could then add credential support (either completely discrete API, or Option<Cred> single API matching)

sdk/storage/azure_storage_blob/src/generated/clients/blob_client.rs
```diff
@@ -40,11 +40,10 @@ use azure_core::{
 };
 use std::sync::Arc;
 
-#[tracing::client]
 pub struct BlobClient {
     pub(crate) blob_name: String,
     pub(crate) container_name: String,
-    pub(crate) endpoint: Url,
+    pub(crate) blob_url: Url,
```
vincenttran-msft, line 46: Generated code change #1: `endpoint` -> `blob_url`. We can maintain all other fields because we return these struct fields directly from our convenience layer for the getters

sdk/storage/azure_storage_blob/src/generated/clients/blob_client.rs
```diff
@@ -106,292 +95,798 @@ impl BlobClient {
...
-        path = path.replace("{blobName}", &self.blob_name);
-        path = path.replace("{containerName}", &self.container_name);
-        url = url.join(&path)?;
-        url.query_pairs_mut()
-            .append_pair("comp", "copy")
-            .append_key_only("sync");
+        let mut url = self.blob_url.clone();
+        if let Some(snapshot) = options.snapshot {
+            url.query_pairs_mut().append_pair("snapshot", &snapshot);
+        }
```
vincenttran-msft, line 880: This is the only currently working API

Generated code change #2:
On a per-API basis, we no longer need to get path parameters in there (i.e. container, blob) since it will now build off `blob_url` and not a naked `endpoint`

vincenttran-msft:
So in general it seems like as long as the generated code changes to using `blob_url` everywhere and makes the matching adjustment per-API to no longer build the path in, that should be all the required changes necessary.

We can look for emitter support for some semblance of a useful constructor, otherwise we can always fall back to just plainly implementing them right there in our handwritten layer 😄

PR #2801, opened 2025-07-17
Adding Location Cache

gsa9989:
### PR Summary:
* Location cache is responsible for organizing the database account data gathered by the global endpoint manager. It holds the client’s preferred locations as well as lists of write and read regions separated by location and availability. Its purpose is to provide an available endpoint based on the client's preferred locations.
* Added location cache functionality and tests. 

### Implementation details:
* APIs include: 
  - update
  - refresh_endpoints
  - refresh_stale_endpoints
  - get_endpoints_by_location
  - get_preferred_available_endpoints
  - mark_endpoint_unavailable
  - is_endpoint_unavailable
  - resolve_service_endpoint

#### Current limitations:
* Currently does not support multi-write capabilities, planning to add it in a follow-up pr with global endpoint manager.


sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
+use std::{collections::HashMap, sync::Mutex, time::SystemTime};
```
analogrelay: You can put `#![allow(dead_code)]` at the top of this file for now. The build is complaining that nothing uses `LocationCache` yet, but that's expected.

sdk/cosmos/azure_data_cosmos/src/global_endpoint_manager.rs
```diff
@@ -0,0 +1 @@
+pub struct GlobalEndpointManager {}
```
analogrelay: I'd just remove this file for now, we can add it when we're ready to implement it

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+    use std::{collections::HashSet, vec};
+
+    fn create_test_data() -> (
+        String,
+        HashMap<String, String>,
+        HashMap<String, String>,
+        Vec<String>,
+    ) {
+        // Setting up test database account data
+        let default_endpoint = "https://default.documents.azure.com".to_string();
```
analogrelay: nit: I like to use `.example.com` as the domain name whenever I'm writing test URLs that we don't actually send requests to. The entire `example.com` domain is reserved exactly for this purpose and it means we're never in danger of accidentally referencing a real endpoint. So, for this, I'd use `https://default.documents.example.com`.

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    #[test]
+    fn mark_read_endpoint_unavailable_test() {
+        // set up test data
+        let (default_endpoint, write_locations, read_locations, preferred_locations) =
+            create_test_data();
+
+        let mut cache = LocationCache::new(default_endpoint, preferred_locations);
+        cache.update(write_locations, read_locations);
```
analogrelay: nit: You could probably bundle this code up into a `create_test_location_cache` function to save a bit of repetition later, since you frequently set up a Location Cache with all the test data in it. You'd still keep `create_test_data` separate since you use it on it's own in a few places too.

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+        // mark location 1 as unavailable endpoint for read operation
+        let unavailable_endpoint = "https://location1.documents.azure.com";
+        let operation = RequestOperation::Read;
+        cache.mark_endpoint_unavailable(unavailable_endpoint, operation);
+
+        // check that endpoint is last option in read endpoints and it is in the location unavailability info map
+        assert_eq!(
+            cache.locations_info.read_endpoints.last(),
+            Some(&unavailable_endpoint.to_string())
+        );
```
analogrelay: I'd suggest asserting the entire content of the list again, since you control the test data. You know that it should be `location2` then `location1`, so rather than just checking for `location1` at the end, check the whole list. That way, your test also catches bugs like `location2` getting dropped as well (through some unintentional issue elsewhere)

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+        (
+            default_endpoint,
+            write_locations,
+            read_locations,
+            preferred_locations,
+        )
+    }
+
+    #[test]
+    fn location_cache_update_test() {
```
analogrelay: nit: Remove the `_test` from your test methods. We know they're tests, and they're already in the `tests` module.

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+pub struct LocationUnavailabilityInfo {
+    pub last_check_time: SystemTime,
+    pub unavailable_operation: RequestOperation,
+}
+
+pub struct LocationCache {
+    pub default_endpoint: String,
+    pub locations_info: DatabaseAccountLocationsInfo,
+    pub location_unavailability_info_map: Mutex<HashMap<String, LocationUnavailabilityInfo>>,
```
analogrelay: We should use an `RwLock` instead of a `Mutex` here. I'll chat about the difference in our 1:1 :)

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    fn refresh_stale_endpoints(&mut self) {
+        let mut location_unavailability_info_map =
+            self.location_unavailability_info_map.lock().unwrap();
+
+        // Removes endpoints that have not been checked in the last 5 minutes
+        location_unavailability_info_map.retain(|_, info| {
+            info.last_check_time.elapsed().unwrap_or_default()
+                <= std::time::Duration::from_secs(300)
```
analogrelay: Put the "freshness" duration in a constant at the top of the file so we know what it is and what it's for. Also (and this is just personal preference), I'll often write this as a math expression rather than just hardcoding the number, since it makes it clearer how it's derived. The compiler will take care of doing the math at compile time so there's no runtime impact:

```
const ENDPOINT_FRESHNESS_DURATION: std::time::Duration = std::time::Duration::from_secs(5 * 60);
```

When you see `from_secs(5 * 60)` it's easier to know that's 5 minutes (and then, honestly, you don't need the comment)

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+
+        cache.update(write_locations, read_locations);
+
+        // mark endpoint 1 and endpoint 2 as unavailable
+        let endpoint1 = "https://location1.documents.azure.com";
+        let endpoint2 = "https://location2.documents.azure.com";
+        cache.mark_endpoint_unavailable(endpoint1, RequestOperation::Read);
+        cache.mark_endpoint_unavailable(endpoint2, RequestOperation::Read);
+
+        // simulate stale entry
```
analogrelay, line 566: Great call! It's much better to forcibly inject a stale entry than to try to have the test "wait" until an entry goes stale.

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+        let operation = RequestOperation::Read;
+        cache.mark_endpoint_unavailable(unavailable_endpoint, operation);
+
+        // check that endpoint is last option in read endpoints and it is in the location unavailability info map
+        assert_eq!(
+            cache.locations_info.read_endpoints.last(),
+            Some(&unavailable_endpoint.to_string())
+        );
+
+        assert!(cache.is_endpoint_unavailable(unavailable_endpoint, operation));
```
analogrelay, line 480: Add an assert on the result of `resolve_service_endpoint` to this test (and the one below) as well. We know that if `read_endpoints` is correct, `resolve_service_endpoint` _should be as well_ but `resolve_service_endpoints` is the actual API that users of this type will see, so we should assert that to make it clear.

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+        {
+            let mut location_unavailability_info_map =
+                self.location_unavailability_info_map.lock().unwrap();
+
+            if let Some(info) = location_unavailability_info_map.get_mut(endpoint) {
+                if info.unavailable_operation == operation
+                    || info.unavailable_operation == RequestOperation::All
+                {
+                    // If the endpoint is already marked as unavailable, update last_check_time
+                    info.last_check_time = now;
```
analogrelay: Since you do this in both branches, you could just move it to before the if on line 99

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+
+    pub fn mark_endpoint_unavailable(&mut self, endpoint: &str, operation: RequestOperation) {
+        let now = std::time::SystemTime::now();
+
+        {
+            let mut location_unavailability_info_map =
+                self.location_unavailability_info_map.lock().unwrap();
+
+            if let Some(info) = location_unavailability_info_map.get_mut(endpoint) {
+                if info.unavailable_operation == operation
```
analogrelay: I have an idea for interesting Rust code we can use here, I'll show you in our 1:1 ;)

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+
+        self.refresh_endpoints();
+    }
+
+    pub fn is_endpoint_unavailable(&self, endpoint: &str, operation: RequestOperation) -> bool {
+        let location_unavailability_info_map =
+            self.location_unavailability_info_map.lock().unwrap();
+        if let Some(info) = location_unavailability_info_map.get(endpoint) {
+            // Checks if endpoint is unavailable for the given operation
+            let operation_type = info.unavailable_operation == operation;
```
analogrelay: nit: A more descriptive name for this boolean would be `operation_type_available`

analogrelay (commented):
Some initial stuff, but nothing significant! Looking good!

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+pub enum RequestOperation {
+    Read,
+    Write,
+    All,
+}
+
+#[derive(Clone, Default)]
+pub struct DatabaseAccountLocationsInfo {
+    pub preferred_locations: Vec<String>,
+    available_write_locations: Vec<String>,
```
tvaron3: Nit: Naming for these variables. I would go with account_write_locations. As it's a little misleading as these locations could be unavailable.

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+        self.refresh_endpoints();
+
+        Ok(())
+    }
+
+    pub fn mark_endpoint_unavailable(&mut self, endpoint: &str, operation: RequestOperation) {
+        let now = std::time::SystemTime::now();
+
+        {
+            let mut location_unavailability_info_map =
```
tvaron3, line 115: Would it be possible to log a message when an endpoint is marked unavailable?

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    pub fn is_endpoint_unavailable(&self, endpoint: &str, operation: RequestOperation) -> bool {
+        let location_unavailability_info_map =
+            self.location_unavailability_info_map.lock().unwrap();
+        if let Some(info) = location_unavailability_info_map.get(endpoint) {
+            // Checks if endpoint is unavailable for the given operation
+            let operation_type = info.unavailable_operation == operation;
+            let elapsed = info.last_check_time.elapsed().unwrap_or_default()
+                < std::time::Duration::from_secs(300);
```
tvaron3: Can the 300 seconds here be moved to a constant? It would be reused in global endpoint manager.

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+        location_index: usize,
+        operation: RequestOperation,
+    ) -> String {
+        // Returns service endpoint based on index, if index out of bounds or operation not supported, returns default endpoint
+        if operation == RequestOperation::Write && !self.locations_info.write_endpoints.is_empty() {
+            self.locations_info.write_endpoints[location_index].clone()
+        } else if operation == RequestOperation::Read
+            && !self.locations_info.read_endpoints.is_empty()
+        {
+            self.locations_info.read_endpoints[location_index].clone()
```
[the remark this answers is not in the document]
tvaron3: Should be mod of the length of the write or read endpoints respectively.

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+            create_test_data();
+
+        let mut cache = LocationCache::new(default_endpoint, preferred_locations);
+
+        cache.update(write_locations, read_locations);
+
+        let endpoint1 = "https://location1.documents.azure.com";
+
+        cache.mark_endpoint_unavailable(endpoint1, RequestOperation::Read);
+        cache.mark_endpoint_unavailable(endpoint1, RequestOperation::Write);
```
tvaron3, line 532: Can we assert for this test that the time gets updated after calling mark endpoint unavailable?

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,570 @@
...
+        // create test data
+        let (default_endpoint, write_locations, read_locations, preferred_locations) =
+            create_test_data();
+
+        let mut cache = LocationCache::new(default_endpoint, preferred_locations);
+
+        cache.update(write_locations, read_locations);
+
+        // resolve service endpoint
+        let endpoint = cache.resolve_service_endpoint(0, RequestOperation::Read);
```
tvaron3, line 587: WOuld be good to also have a test for a larger index than the list being passed in ?

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
gsa9989: As in return a message to the user that the function worked?

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
gsa9989: Good point, thank you!

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
tvaron3: Something like the examples here https://docs.rs/log/latest/log/ and the message can "\<endpoint\> was marked unavailable for <write/read>". It could help debugging if a customer asks why there traffic switched from one region to another.

gsa9989: I added `info!(
                "Endpoint {} was marked as unavailable for {:?}",
                endpoint, operation
            );` without any issues other than importing tracing::info. However, if I wanted to test it I'd have to add the log crate to cargo.toml. Should I do that or is adding the message sufficient?

analogrelay: `tracing` is the correct API to use (we don't actually use the `log` crate in the Rust SDK, we use `tracing` instead). I don't think we need a test for this case. We do test some of our diagnostics/telemetry, but I don't think individual log messages need to be tested.

gsa9989: Got it, thank you!

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,603 @@
...
+            endpoint,
+            "https://location2.documents.example.com".to_string()
+        );
+    }
+
+    #[test]
+    fn resolve_service_endpoint_default() {
+        let cache = create_test_location_cache();
+
+        let endpoint = cache.resolve_service_endpoint(5, RequestOperation::Read);
```
analogrelay: nit: We should be able to access the `read_endpoints` list, so rather than picking a magic number we know is currently more than the number of locations in the test data, we can use the length itself:

```suggestion
        let endpoint = cache.resolve_service_endpoint(cache.read_endpoints.len() + 1, RequestOperation::Read);
```

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
analogrelay: Small nit on the phrasing you had above: Our log messages generally start with lowercase letters. For example:

```
2025-08-12T19:40:21.225780Z DEBUG azure_data_cosmos::pipeline::signature_target: generating Cosmos auth signature signature_payload="get\ndocs\ndbs/SampleDB/colls/SampleContainer/docs/14174164-F6C0-47FC-83FB-604C6A63408D\ntue, 12 aug 2025 19:40:21 gmt\n\n"
2025-08-12T19:40:21.225858Z DEBUG typespec_client_core::http::policies::transport: sending request 'https://localhost:8081/dbs/SampleDB/colls/SampleContainer/docs/14174164-F6C0-47FC-83FB-604C6A63408D' request=Request { url: "https://localhost:8081/dbs/SampleDB/colls/SampleContainer/docs/14174164-F6C0-47FC-83FB-604C6A63408D", method: Get, headers: {"user-agent": "*****", "x-ms-documentdb-partitionkey": "*****", "x-ms-client-request-id": "*****", "x-ms-date": "*****", "x-ms-version": "*****", "authorization": "*****"}, body: Bytes {} }
2025-08-12T19:40:21.226064Z DEBUG typespec_client_core::http::clients::reqwest: performing request GET 'https://localhost:8081/dbs/SampleDB/colls/SampleContainer/docs/14174164-F6C0-47FC-83FB-604C6A63408D' with `reqwest`
2025-08-12T19:40:21.226185Z DEBUG reqwest::connect: starting new connection: https://localhost:8081/
2025-08-12T19:40:21.226863Z DEBUG hyper_util::client::legacy::connect::http: connecting to [::1]:8081
2025-08-12T19:40:21.227041Z DEBUG hyper_util::client::legacy::connect::http: connected to [::1]:8081
```

Also, "was marked" is passive voice, which we try to avoid where possible (we're not perfect ;)). We try to use active statements as much as possible, so "endpoint marked unavailable" (describing what the application is doing right now) rather than "endpoint was marked unavailable" (describing what the application did sometime in the past).

So it should be `endpoint {} marked as unavailable for {:?}`

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
```diff
@@ -0,0 +1,603 @@
...
+    }
+
+    #[test]
+    fn mark_same_endpoint_unavailable() {
+        // set up test cache
+        let mut cache = create_test_location_cache();
+
+        let endpoint1 = "https://location1.documents.example.com";
+
+        let before_marked_unavailable_time = SystemTime::now();
```
analogrelay: I like asserting this, but there's a potential problem. There's no guarantee that `SystemTime::now()` will have advanced between when you call it, and when `mark_endpoint_unavailable` runs. The `SystemTime` type in Rust is **not guaranteed to be monotonic** (i.e. it is not guaranteed to consistently increase). On most modern OSes, it will be, but that's not a guarantee.

Also, it's possible the CPU instructions between here and when you capture `SystemTime::now` again inside `mark_endpoint_unavailable` all execute before the clock moves forward. In most cases, this will work fine, there are enough instructions and CPU cycles between each call to get the time that it's likely the clock will have moved forward, but there's no guarantee of that and slower/faster processors may disrupt the test.

All that is to say that asserting the passage of time in tests is tricky ;). Instead of capturing `now` here, pick a time several seconds before `now` to use as `before_marked_unavailable_time` (`SystemTime::now() - Duration::from_secs(10)`) and then **set** the `last_check_time` of the endpoint to that value (similar to what you're doing with the staleness check). Then, when you assert below that `last_check_time` is greater than `before_marked_unavailable_time`, you have a much higher chance of it being true.

analogrelay (commented):
Looking good! I think we're close! The currently build failure isn't your responsibility, it's an issue in a different crate that's also failing in `main`. We'll have to wait for it to be resolved before merging, but it shouldn't be an issue.

sdk/cosmos/azure_data_cosmos/src/location_cache.rs
gsa9989: That makes sense, thank you!

analogrelay (approved):

analogrelay:
Ah, we do! Good catch @kushagraThapar . I'm bad at remembering that and we don't have any automation in place yet :(. @gsa9989 you can just grab the two comment lines from the top of any other Rust source file and add it in.

kushagraThapar (approved):
LGTM!

PR #2814, opened 2025-07-23
[Storage] `set_legal_hold`, `set/get_immutability_policy` for BlobClient, `set/get_access_policy` for ContainerClient, `get_service_stats`, `get_user_delegation_key` for ServiceClient

vincenttran-msft:


sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
```diff
@@ -104,4 +106,32 @@ impl BlobServiceClient {
...
+    /// # Arguments
+    ///
+    /// * `options` - Optional configuration for the request.
+    pub async fn get_service_stats(
+        &self,
+        options: Option<BlobServiceClientGetStatisticsOptions<'_>>,
+    ) -> Result<Response<StorageServiceStats, XmlFormat>> {
+        // TODO: Propagate rename to .tsp
+        self.client.get_statistics(options).await
+    }
```
vincenttran-msft: Need to propagate rename to `.tsp` level

sdk/storage/azure_storage_blob/src/parsers.rs
```diff
@@ -35,3 +37,16 @@ pub fn format_page_range(offset: u64, length: u64) -> Result<String, Error> {
...
+    let time_string = to_rfc3339(time);
+
+    match time_string.find('.') {
+        Some(dot_index) => {
+            let base = &time_string[..dot_index];
+            format!("{}Z", base)
+        }
+        None => time_string.to_string(),
+    }
+}
```
vincenttran-msft: Probably needs to be more robust? But temporary stopgap, as using `to_rfc3339` yields milliseconds which the service does not accept.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -423,3 +425,91 @@ async fn test_leased_blob_operations(ctx: TestContext) -> Result<(), Box<dyn Err
...
+    // Assert
+    let legal_hold = response.legal_hold()?;
+    assert!(!legal_hold.unwrap());
+
+    // Even after removing the legal hold:
+    // Error: Error { context: Full(Custom { kind: HttpResponse { status: Conflict, error_code: Some("ContainerImmutableStorageWithVersioningEnabled") },
+    // error: HttpError { status: Conflict, details: ErrorDetails { code: Some("ContainerImmutableStorageWithVersioningEnabled"), message: None },
+    // body: <?xml version="1.0" encoding="utf-8"?><Error><Code>ContainerImmutableStorageWithVersioningEnabled</Code><Message>
+    // The requested operation is not allowed as the container has a immutable storage with versioning enabled.
+    // Seems like in Python testing, we need the management client in order to delete the container
```
vincenttran-msft: As called out here -- seems like we can't do a normal delete and need the management client as we did in Python.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -423,3 +425,91 @@ async fn test_leased_blob_operations(ctx: TestContext) -> Result<(), Box<dyn Err
...
+    };
+    let response = blob_client
+        .set_immutability_policy(Some(immutability_policy_options))
+        .await?;
+
+    // Assert
+    // ERRORS: Error: Error { context: Full(Custom { kind: DataConversion, error: Error { context: Message { kind: DataConversion, message: "unknown variant of BlobImmutabilityPolicyMode found: \"unlocked\"" } } },
+    // "unable to parse header 'HeaderName(\"x-ms-immutability-policy-mode\"): HeaderValue'
+    // into azure_storage_blob::generated::models::enums::BlobImmutabilityPolicyMode") }
+    // let mode = response.immutability_policy_mode()?;
```
vincenttran-msft: This is an interesting one-- we have `ImmutabilityPolicyMode` and `BlobImmutabilityPolicyMode`. The difference is that the `Blob` prefix one contains `Mutable` which is a valid enum for the server to return, but not a valid value for us to set.

However, there seems to be some issue with converting it here?

sdk/storage/azure_storage_blob/tests/blob_container_client.rs
```diff
@@ -275,3 +279,27 @@ async fn test_container_lease_operations(ctx: TestContext) -> Result<(), Box<dyn
...
+        permission: Some("rw".to_string()),
+        start: Some(OffsetDateTime::now_utc()),
+    };
+    let signed_identifier = SignedIdentifier {
+        access_policy: Some(access_policy),
+        id: None,
+    };
+
+    //TODO: Need gen code support- how to provide the SignedIdentifier in proper format
+    // let request_content = RequestContent::try_from(vec![signed_identifier])?;
```
vincenttran-msft: Straight up having issues with getting a `RequestContent<Vec<SignedIdentifier>>`. In most cases we can just use `try_from` on the model, but I think the difficulty here is that it is wrapped in a `Vec`

Not really sure multiple `SignedIdentifiers` would even mean in this case? Seems like a `.tsp` authoring mistake.

sdk/storage/azure_storage_blob/tests/blob_service_client.rs
```diff
@@ -122,3 +125,46 @@ async fn test_list_containers_with_continuation(ctx: TestContext) -> Result<(),
...
 }
+
+#[recorded::test]
+async fn test_get_service_stats(ctx: TestContext) -> Result<(), Box<dyn Error>> {
+    //TODO: Need playback only, need secondary-enabled GRS account to record against
+    // Recording Setup
+    let recording = ctx.recording();
+    let service_client = get_blob_service_client(recording)?;
+
+    let response = service_client.get_service_stats(None).await?;
```
vincenttran-msft, line 133: Seems like this only works against secondary? And in Python these are playback only as the newly spun up account in the live test pipeline does not persist long enough for meaningful logs to generate.

sdk/storage/azure_storage_blob/tests/blob_service_client.rs
```diff
@@ -122,3 +125,46 @@ async fn test_list_containers_with_continuation(ctx: TestContext) -> Result<(),
...
+    ));
+    // Act
+    let key_info = KeyInfo {
+        start: start.clone(),
+        expiry: expiry.clone(),
+    };
+
+    let response = service_client
+        .get_user_delegation_key(RequestContent::try_from(key_info)?, None)
+        .await?;
```
vincenttran-msft: Usage of this currently kind of sucks (you have to hand author the model and ensure you are using the `format_time_string` helper to get the right level of truncation on your `OffsetDateTime` resulting string).

vincenttran-msft: Probably a good candidate to make a format helper that takes in start and expiry, will get the right level of truncation and then return a `KeyInfo` object?

sdk/storage/azure_storage_blob/src/parsers.rs
vincenttran-msft: Service needs 7, we provided 6? Try adding a 0Z instead of truncating?

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: Unify, see if it rectifies

sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
```diff
@@ -104,4 +106,32 @@ impl BlobServiceClient {
...
+
+    /// Retrieves a user delegation key for the Blob service. This is only a valid operation when using bearer token authentication.
+    ///
+    /// # Arguments
+    ///
+    /// * `key_info` - Key information such as when the key becomes valid and expires.
+    /// * `options` - Optional configuration for the request.
+    pub async fn get_user_delegation_key(
```
vincenttran-msft: Punt

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -423,3 +425,87 @@ async fn test_leased_blob_operations(ctx: TestContext) -> Result<(), Box<dyn Err
...
+    let response = blob_client.get_properties(None).await?;
+    // Assert
+    let legal_hold = response.legal_hold()?;
+    assert!(!legal_hold.unwrap());
+
+    blob_client.delete(None).await?;
+
+    // Hitting an error to delete despite also manually deleting all Blobs (and verifying its empty state in Azure Portal)
+    // However, deleting in Azure Portal works.
+    container_client.delete_container(None).await?;
```
vincenttran-msft, line 463: Still having issues with deleting container even after 1/1 blobs.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -423,3 +425,87 @@ async fn test_leased_blob_operations(ctx: TestContext) -> Result<(), Box<dyn Err
...
+    };
+    let response = blob_client
+        .set_immutability_policy(Some(immutability_policy_options))
+        .await?;
+
+    // Assert
+    // ERRORS: Error: Error { context: Full(Custom { kind: DataConversion, error: Error { context: Message { kind: DataConversion, message: "unknown variant of BlobImmutabilityPolicyMode found: \"unlocked\"" } } },
+    // "unable to parse header 'HeaderName(\"x-ms-immutability-policy-mode\"): HeaderValue'
+    // into azure_storage_blob::generated::models::enums::BlobImmutabilityPolicyMode") }
+    let mode = response.immutability_policy_mode()?;
```
vincenttran-msft, line 492: Parsing issue, started a thread to track this.

sdk/storage/azure_storage_blob/tests/blob_container_client.rs
```diff
@@ -275,3 +279,41 @@ async fn test_container_lease_operations(ctx: TestContext) -> Result<(), Box<dyn
...
+    let signed_identifier = SignedIdentifier {
+        access_policy: Some(access_policy),
+        id: None,
+    };
+
+    let request_content_1 = RequestContent::try_from(signed_identifier);
+    let request_content_2 = RequestContent::try_from(vec![signed_identifier]);
+    container_client
+        .set_access_policy(request_content, None)
+        .await?;
```
vincenttran-msft: Having issues getting the Vec<SignedIdentifiers> into a RequestContent, opened an issue to track.

sdk/storage/azure_storage_blob/tests/blob_container_client.rs
```diff
@@ -278,3 +282,39 @@ async fn test_container_lease_operations(ctx: TestContext) -> Result<(), Box<dyn
...
+        permission: Some("rw".to_string()),
+        start: Some(OffsetDateTime::now_utc()),
+    };
+    let signed_identifier = SignedIdentifier {
+        access_policy: Some(access_policy),
+        id: None,
+    };
+
+    container_client
+        .set_access_policy(signed_identifier.try_into()?, None)
```
vincenttran-msft, line 305: error[E0277]: the trait bound `RequestContent<Vec<SignedIdentifier>, XmlFormat>: TryFrom<...>` is not satisfied
   --> sdk\storage\azure_storage_blob\tests\blob_container_client.rs:305:46
    |
305 |         .set_access_policy(signed_identifier.try_into()?, None)
    |                                              ^^^^^^^^ unsatisfied trait bound
    |
    = help: the trait `std::convert::From<azure_storage_blob::models::SignedIdentifier>` is not implemented for `RequestContent<Vec<SignedIdentifier>, XmlFormat>`

PR #2826, opened 2025-07-28
Added Reconnect support for Event Hubs

LarryOsterman:
Initial reconnect support

Test infrastructure in progress.


sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable/connection.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All Rights reserved
 // Licensed under the MIT license.
 
+// cspell: ignore Cloneable
```
heaths: Did this work? I thought there had to be no spaces e.g.,


```suggestion
// cspell:ignore Cloneable
```

Cargo.toml
```diff
@@ -81,9 +81,9 @@ cargo_metadata = "0.18.1"
 clap = { version = "4.4.16", features = ["derive"] }
 criterion = { version = "0.5", features = ["async_tokio"] }
 dyn-clone = "1.0"
-fe2o3-amqp = { version = "0.14", features = ["native-tls", "tracing", "uuid"] }
+fe2o3-amqp = { version = "0.14", features = ["native-tls", "uuid"] }
 fe2o3-amqp-ext = { version = "0.14" }
-fe2o3-amqp-management = { version = "0.14", features = ["tracing"] }
+fe2o3-amqp-management = { version = "0.14", features = [] }
```
heaths: Nit:
```suggestion
fe2o3-amqp-management = { version = "0.14" }
```

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable/connection.rs
```diff
@@ -23,6 +24,8 @@ use azure_core_amqp::{
     AmqpManagement, AmqpReceiver, AmqpReceiverApis, AmqpReceiverOptions, AmqpSender,
     AmqpSenderApis, AmqpSession, AmqpSessionApis, AmqpSessionOptions, AmqpSource, AmqpSymbol,
 };
+#[cfg(feature = "test")]
```
heaths: Why `feature = "test"` and not just `test`?

sdk/eventhubs/azure_messaging_eventhubs/tests/force_errors.rs
```diff
@@ -0,0 +1,523 @@
...
+                    let batch = producer
+                        .create_batch(Some(EventDataBatchOptions {
+                            partition_id: Some(EVENTHUB_PARTITION.to_string()),
+                            partition_key: Some("My Partition Key.".to_string()),
+                            ..Default::default()
+                        }))
+                        .await
+                        .unwrap();
+
+                    for i in 1..200 {
```
heaths: Given the 10, 30, even 45 second time outs, is this sample *always* going to be enough? Should this async fn actually sleep itself to make sure the producer is forced into an error in time? If I'm following the logic correctly, this seems to assume a lot about modern hardware (which, of course, will change).

heaths (changes requested):

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable/connection.rs
LarryOsterman: It has in the past. And cspell doesn't seem to complain. I'll check.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable/connection.rs
LarryOsterman: Because [test] does not work across package boundaries. I need a feature to enable test hooks that are used across package boundaries (or from tests in the `test` directory.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable/connection.rs
LarryOsterman: And it appears that it's not actually used. So removing it.

sdk/eventhubs/azure_messaging_eventhubs/tests/force_errors.rs
LarryOsterman: The test puts 200 messages in a batch and then loops continuously until the future is dropped.

The 200 is just the size of the batch being sent (lines 79-90 add events to the batch, and 91 sends the batch.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable/sender.rs
```diff
@@ -18,7 +19,7 @@ use tracing::{debug, warn};
 /// the retry functionality. That allows implementations which call into the Send API to not have
 /// to worry about retrying the operation themselves.
 pub(crate) struct RecoverableSender {
-    recoverable_connection: Arc<RecoverableConnection>,
+    recoverable_connection: Weak<RecoverableConnection>,
```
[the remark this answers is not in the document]
LarryOsterman: This change is necessary to ensure that the sender does not keep a strong reference to the `RecoverableConnection` - that ensures that outstanding senders don't stop the ability to close a `RecoverableConnection`.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable/connection.rs
```diff
@@ -59,14 +65,17 @@ pub(crate) struct RecoverableConnection {
...
+    connections: AsyncMutex<Option<Arc<AmqpConnection>>>,
+    mgmt_client: AsyncMutex<Option<Arc<AmqpManagement>>>,
     session_instances: AsyncMutex<HashMap<Url, Arc<AmqpSession>>>,
     receiver_instances: AsyncMutex<HashMap<Url, Arc<AmqpReceiver>>>,
     sender_instances: AsyncMutex<HashMap<Url, Arc<AmqpSender>>>,
     pub(super) authorizer: Arc<Authorizer>,
     connection_name: String,
     pub(super) retry_options: RetryOptions,
+
+    #[cfg(feature = "test")]
```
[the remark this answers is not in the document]
LarryOsterman: The `test` feature enables test hooks in the eventhubs codebase. Since the test which uses this code is outside the package (because it is an end-to-end test), it cannot use the `test` configuration.

sdk/core/azure_core_amqp/CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Release History
 
+## 0.7.0 (Unreleased)
+
+### Features Added
+
+### Breaking Changes
+
+- Distinguish remote disconnect and remote closed errors by origin
```
heaths, line 9: At face value this doesn't seem like a breaking change. If it is, cool. Just wanted to mention it in case it ended up under the wrong header.

LarryOsterman: I added a bunch of fields to an enumeration which is not marked as non-exhaustive, which is a breaking change.

LarryOsterman:
@heaths There was a merge conflict, can you please reapprove?

PR #2833, opened 2025-07-29
Enable json submodule if json feature is used

antkmsft:
Follow-up for https://github.com/Azure/azure-sdk-for-rust/pull/2790#discussion_r2237922684

PR that updates Spector tests: https://github.com/Azure/typespec-rust/pull/530

### DIFF
--- a/sdk/typespec/typespec_client_core/src/http/request/mod.rs
+++ b/sdk/typespec/typespec_client_core/src/http/request/mod.rs
@@ -333,6 +333,9 @@ impl<T> TryFrom<Option<Decimal>> for RequestContent<T> {
     }
 }
 
+#[cfg(feature = "json")]
+/// JSON-specific conversions for `RequestContent<T>`.
+/// These conversions allow for easy serialization of common types into the request body.
 pub mod json {
     use super::*;
 
@@ -530,6 +533,9 @@ pub mod json {
         }
     }
 }
+
+#[cfg(feature = "json")]
+pub use json::*;
 
 impl<T> FromStr for RequestContent<T> {
     type Err = crate::Error;
--- a/sdk/typespec/typespec_client_core/src/http/request/mod.rs
+++ b/sdk/typespec/typespec_client_core/src/http/request/mod.rs
@@ -534,6 +534,7 @@ pub mod json {
     }
 }
 
+#[allow(unused_imports)]
 #[cfg(feature = "json")]
 pub use json::*;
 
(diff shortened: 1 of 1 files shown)


sdk/typespec/typespec_client_core/src/http/request/mod.rs
```diff
@@ -531,6 +534,10 @@ pub mod json {
     }
 }
 
+#[allow(unused_imports)]
```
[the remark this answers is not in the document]
antkmsft: Hmm, I think it is pretty much the intent - it will give me the warning otherwise if I remove this line as you suggest.

sdk/typespec/typespec_client_core/src/http/request/mod.rs
```diff
@@ -531,6 +534,10 @@ pub mod json {
...
+#[cfg(feature = "json")]
+pub use json::*;
```
heaths: Don't expose a public submodule and flatten all of them at the same time. One or the other.

Per discussion, though, let me play around with what's in `main` now first to see how it works for JSON and/or XML.

sdk/typespec/typespec_client_core/src/http/request/mod.rs
```diff
@@ -333,6 +333,9 @@ impl<T> TryFrom<Option<Decimal>> for RequestContent<T> {
     }
 }
 
+#[cfg(feature = "json")]
+/// JSON-specific conversions for `RequestContent<T>`.
+/// These conversions allow for easy serialization of common types into the request body.
```
heaths: Nit: stylistically, we generally put comments before any attributes or other coding elements. Also, markdown requires a blank line between the summary and more info; otherwise, they get merged into a single line and summaries (first sentence) should be concise.

```suggestion
/// JSON-specific conversions for `RequestContent<T>`.
///
/// These conversions allow for easy serialization of common types into the request body.
#[cfg(feature = "json")]
```

antkmsft: Updated. Thank you!

sdk/typespec/typespec_client_core/src/http/request/mod.rs
antkmsft: Yes, thank you, just removed the `use`, but as we were talking about, this PR is for us to follow up on all your feedback on this aspect, so, when you have it, please let me know what do you think will be the best and I will make this PR up to date. Until then, my intent for this PR to kind of sit here unmerged.

PR #2835, opened 2025-07-29
[Storage] `get_page_ranges`, `update_sequence_number` and `upload_pages_from_url` for PageBlobClient

vincenttran-msft:
Remaining APIs for PageBlobClient with the exception of `get_page_ranges_diff`.
This requires a change to correct an error in the TypeSpec that is currently authored: https://github.com/Azure/azure-rest-api-specs/pull/36311

`update_sequence_number`:

-  ̶d̶a̶t̶e̶
- last_modified
- etag
- blob_sequence_number

`upload_pages_from_url`:

- content_md5
-  ̶d̶a̶t̶e̶
- last_modified
- etag
- blob_sequence_number
- content_crc64
- encryption_key_sha256
- encryption_scope
- is_server_encrypted

This PR also moves the `.assets` test file from the root storage directory to the specific crate directory.


sdk/storage/azure_storage_blob/assets.json
heaths, line 1: This file should stay next to your `test-resources.bicep`. Why'd you move it? If you leave it in place, changes to your recordings and your test resources - both of which factor into your actual tests - are tied together. If they aren't, they you have to manage all recordings independently whenever you make test resource changes.

And if you do that, change the tag to just be your service directory i.e., "rust/storage".

sdk/storage/azure_storage_blob/tests/page_blob_client.rs
```diff
@@ -163,3 +165,149 @@ async fn test_resize_blob(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
+        )
+        .await?;
+
+    // Assert
+    let response = blob_client_2.download(None).await?;
+    let content_length = response.content_length()?;
+    let (status_code, _, response_body) = response.deconstruct();
+    assert!(status_code.is_success());
+    assert_eq!(1024, content_length.unwrap());
+    data_2.extend(&data_1);
```
heaths: Nit: remembering what's in `data_1` vs. `data_2` is harder than just calling them `data_a` - which contains `b'A'`, for example.

heaths (approved):

sdk/storage/azure_storage_blob/assets.json
jalauzon-msft: This is how Storage does this in all other languages. Because we have so many recordings, it's best to keep them separated by package. Whereas we only need a single bicep. We rarely change the bicep.

If you feel strongly, we can go back to the other way for Rust.

sdk/storage/azure_storage_blob/tests/page_blob_client.rs
```diff
@@ -163,3 +165,149 @@ async fn test_resize_blob(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
+    let page_blob_client_2 = blob_client_2.page_blob_client();
+
+    // Act
+    page_blob_client_1.create(512, None).await?;
+    let data_1 = vec![b'B'; 512];
+    page_blob_client_1
+        .upload_page(
+            RequestContent::from(data_1.clone()),
+            512,
+            format_page_range(0, 512)?,
```
vincenttran-msft, line 232: Add docstring to use format_page_range

heaths:
Is this ready to merge? I wouldn't let it linger for too long or merges become and increasing problem. Merge it, and fix subsequent issues as needed.

jalauzon-msft (approved):

PR #2838, opened 2025-07-30
typespec_client_core: New `web_runtime` to support wasm32-unknown-unknown

### LINKED ISSUE #2754: wasm32-unknown-unknown failed to run

magodo:
The `typespec_client_core::sleep::sleep` is using the async_runtime's sleep implementation, which is complicated to support `wasm32`, and will panic now when targeting to `wasm32`. Instead of make the `standard_runtime` to support wasm32 sleep (I wonder if it's feasible), we can instead just introduce another `sleep` implementation to simply use the `gloo_timers::future::sleep`.

Also, this PR adds a `cfg_attr` for conditionally remove the `Send` trait from the `async_trait` for the `RetryPolicy`, as the other parts of the code do.

Fix #2754
Supersedes: https://github.com/Azure/azure-sdk-for-rust/pull/2770

## Test

```
sdk/typespec/typespec_client_core on  wasm_sleep is 📦 v0.6.0 via 🦀 v1.88.0
💤  cargo test -q

running 90 tests
i.................................................................i.................... 87/90
...
test result: ok. 88 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 0.84s


running 23 tests
.......................
test result: ok. 23 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.27s


sdk/typespec/typespec_client_core on  wasm_sleep is 📦 v0.6.0 via 🦀 v1.88.0 took 4s
💤  cargo test --features=tokio -q

running 91 tests
i.....................................................................i................ 87/91
....
test result: ok. 89 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 0.10s


running 23 tests
.......................
test result: ok. 23 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.21s


sdk/typespec/typespec_client_core on  wasm_sleep is 📦 v0.6.0 via 🦀 v1.88.0 took 3s
💤  RUSTFLAGS='--cfg getrandom_backend="wasm_js"' wasm-pack test --node --features wasm-bindgen
[INFO]: 🎯  Checking for the Wasm target...
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.14s
[INFO]: ⬇️   Installing wasm-bindgen...
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running unittests src/lib.rs (/home/magodo/github/azure-sdk-for-rust/target/wasm32-unknown-unknown/debug/deps/typespec_client_core-365ee3948d4ee278.wasm)
running 2 tests
test async_runtime::tests::wasm_bindgen_task_execution ... ok
test async_runtime::tests::test_sleep ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 filtered out; finished in 0.17s
```

Some note about the wasm32 test:

- We can't run the spawned job using `futures::executor::block_on(handle)`, as it introduces a deadlock in browser (since the `block_on` blocks the current thread)
- The sleep test I can't use `Instant::now()` when using `node` as runtime

### DIFF
--- a/sdk/typespec/typespec_client_core/Cargo.toml
+++ b/sdk/typespec/typespec_client_core/Cargo.toml
@@ -16,6 +16,7 @@ base64.workspace = true
 bytes.workspace = true
 dyn-clone.workspace = true
 futures.workspace = true
+gloo-timers = { workspace = true, optional = true }
 pin-project.workspace = true
 quick-xml = { workspace = true, optional = true }
 rand.workspace = true
@@ -37,11 +38,9 @@ tokio = { workspace = true, features = ["macros", "rt-multi-thread", "time"] }
 
 [target.'cfg(target_family = "wasm")'.dependencies]
 getrandom.workspace = true
-gloo-timers = { workspace = true, optional = true }
 tokio = { workspace = true, features = ["macros", "rt", "time"] }
 
 [dev-dependencies]
-tokio.workspace = true
 tracing.workspace = true
 tracing-subscriber.workspace = true
 typespec_macros.path = "../typespec_macros"
@@ -50,6 +49,7 @@ typespec_macros.path = "../typespec_macros"
 tokio = { workspace = true, features = ["fs"] }
 
 [target.'cfg(target_family = "wasm")'.dev-dependencies]
+tokio.workspace = true
 getrandom = { workspace = true, features = ["wasm_js"] }
 uuid = { workspace = true, features = ["v4", "js"] }
 wasm-bindgen-test.workspace = true
@@ -73,6 +73,10 @@ wasm_bindgen = ["dep:wasm-bindgen-futures", "gloo-timers/futures"]
 xml = ["dep:quick-xml"]
 
 [[example]]
+name = "core_binary_data_request"
+required-features = ["tokio"]
+
+[[example]]
 name = "core_stream_response"
 required-features = ["derive"]
 
@@ -87,6 +91,7 @@ features = [
   "reqwest_gzip",
   "reqwest_rustls",
   "tokio",
+  "wasm_bindgen",
   "xml",
 ]
 
--- a/sdk/typespec/typespec_client_core/Cargo.toml
+++ b/sdk/typespec/typespec_client_core/Cargo.toml
@@ -16,6 +16,7 @@ base64.workspace = true
 bytes.workspace = true
 dyn-clone.workspace = true
 futures.workspace = true
+gloo-timers = { workspace = true, optional = true }
 pin-project.workspace = true
 quick-xml = { workspace = true, optional = true }
 rand.workspace = true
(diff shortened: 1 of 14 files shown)


sdk/typespec/typespec_client_core/Cargo.toml
```diff
@@ -43,6 +45,9 @@ tokio.workspace = true
 tracing.workspace = true
 tracing-subscriber.workspace = true
 typespec_macros.path = "../typespec_macros"
+wasm-bindgen-test = "0.3"
+getrandom = { version = "0.3" , features = ["wasm_js"]}
+uuid = { version = "1", features = ["v4", "js"] }
```
heaths: All versions need to be declared in the workspace `Cargo.toml`. You can add features in crates like this if necessary e.g.,

```toml
uuid = { workspace = true, features = ["js", "v4"] }
```

That would union the feature set.

sdk/typespec/typespec_client_core/src/async_runtime/mod.rs
```diff
@@ -189,12 +192,16 @@ pub fn set_async_runtime(runtime: Arc<dyn AsyncRuntime>) -> crate::Result<()> {
...
     }
     #[cfg(feature = "tokio")]
     {
         Arc::new(tokio_runtime::TokioRuntime) as Arc<dyn AsyncRuntime>
     }
+    #[cfg(not(any(feature = "tokio", feature = "wasm-bindgen")))]
+    {
+        Arc::new(standard_runtime::StdRuntime)
+    }
 }
```
heaths, line 200: We should have a `panic!()` since there now exists a case we otherwise fall through: `target_arch == "wasm32"` and NOT `feature = "wasm-bindgen"`.

heaths: Actually, I see that the `standard_runtime` module handles the `not(target_arch = "wasm32")` case itself, so ignore this comment.

heaths (approved):
Approved with my changes in there.

heaths:
The macOS issue is the #2887 issue and not really a blocker.

PR #2844, opened 2025-08-01
Add Format parameter to RequestContent

heaths:
This provides parity with `Response<T, F: Format>` so that the client clearly dictates the request serialization format.


LarryOsterman (approved):

heaths:
Talking with @antkmsft I'll merge this first so we can pin the emitter to a commit in `main`. This should also obviate Azure/typespec-rust#530 since I took a dependency on the "decimal" feature in https://github.com/Azure/typespec-rust/pull/536. The "json" feature is on by default.

PR #2846, opened 2025-08-01
Remove CertificateClientExt and impl directly on CertificateClient

heaths:
Turns out we don't need extension traits within the same crate. Better discoverability and usage.


heaths:
@RickWinter @LarryOsterman what do you think of this? I misunderstood the accessibility of implementing functions on types within the same crate but in different modules. I think this improves discoverability and simplifies usage. Also can be used to add constructors easily. /cc @analogrelay

PR #2851, opened 2025-08-01
Implement checkpoint store for azure storage blobs.

LarryOsterman:


sdk/core/azure_core_test/src/recording.rs
```diff
@@ -333,9 +333,10 @@ impl Recording {
             return value;
         }
 
+        let value = value?;
         let mut variables = self.variables.write().map_err(write_lock_error).ok()?;
-        variables.insert(key.into(), Value::from(value.as_ref(), options));
-        value
+        variables.insert(key.into(), Value::from(Some(&value), options));
+        Some(value)
```
heaths: It took me a couple times to really notice the effect of the `?` on line 336. Probably more obvious e.g.,


```suggestion
        match value {
            None => None,
            // Treat empty variable as absent.
            Some(v) if v.is_empty() => None,
            Some(v) => {
                let mut variables = self.variables.write().map_err(write_lock_error).ok()?;
                variables.insert(key.into(), Value::from(value.as_ref(), options));
                value
            }
        }
```

.vscode/cspell.json
```diff
@@ -32,6 +32,7 @@
     "azsdk",
     "azurecli",
     "bugbug",
+    "checkpointstore",
```
heaths, line 35: Nit: This is in your `sdk/eventhubs/.dict.txt` file as well. Necessary in both places? If not a problem outside that directory, the `.dict.txt` file would be best for narrowly-scoped words.

Cargo.toml
```diff
@@ -59,6 +60,14 @@ path = "sdk/core/azure_core_macros"
...
 path = "sdk/core/azure_core_amqp"
 
+[workspace.dependencies.azure_messaging_eventhubs]
+version = "0.6.0"
+path = "sdk/eventhubs/azure_messaging_eventhubs"
+
+[workspace.dependencies.azure_messaging_eventhubs_checkpointstore_blob]
+version = "0.1.0"
+path = "sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob"
+
```
heaths: Is this a dependency for anything? We should only do this for dependencies because they can be a bigger pain to update.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/src/lib.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+//! Azure Event Hubs Checkpoint Store implementation using Azure Blob Storage.
```
heaths: This should import the README.md as we always do.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/assets.json
heaths, line 1: If you moved your test-resources.bicep to your service directory, seems all your recorded test assets should follow suit. Going to be a lot easier to manage.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+readme = "README.md"
+license = "MIT"
+repository = "https://github.com/Azure/azure-sdk-for-rust"
+homepage = "https://github.com/Azure/azure-sdk-for-rust"
+documentation = "https://docs.rs/azure_messaging_eventhubs_checkpointstore_blob"
+keywords = ["sdk", "azure", "eventhubs", "checkpoints", "blob"]
+categories = ["api-bindings"]
+authors = ["Microsoft"]
+edition = "2021"
+rust-version = "1.75.0"
```
heaths: Most of these must inherit from the workspace.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/ci.yml
```diff
@@ -0,0 +1,64 @@
...
+pr:
+  branches:
+    include:
+      - main
+      - feature/*
+  paths:
+    include:
+      - sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/
+
+stages:
```
heaths: 1. Why is this so different than any other pipeline?
2. Why not in your service directory as is normal for a family of crates? They get tested together typically.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/README.md
```diff
@@ -0,0 +1,97 @@
+# Azure Event Hubs Checkpoint Store for Blob Storage
+
+<!-- cspell: ignore myapp -->
```
heaths: Could also put it in your `.dict.txt` file so it doesn't break the flow of people reading the raw markdown e.g., in IDEs.

heaths (changes requested):
I mainly have concerns about layout as well as that odd pipeline.

Cargo.toml
LarryOsterman: No it is not a dependency for something else.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/README.md
LarryOsterman: I could - my normal rule is "if it's in more than one place, it goes into the `.dict.txt` file, otherwise it goes in the file.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/ci.yml
LarryOsterman: It's not actualy used, I believe.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/assets.json
LarryOsterman: Not that it really matters since EH doesn't have recorded assets, but it's a good idea.

.vscode/cspell.json
LarryOsterman: The problem is the root cargo.toml file :(. So it needs to be in the root. I wish that wasn't the case, but...

sdk/core/azure_core_test/src/recording.rs
LarryOsterman: Clippy is actually why the `let value = value?;` line is there :). I had a more complicated let Some() thingy.

But I like your construct better.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/tests/checkpoint_unit_tests.rs
```diff
@@ -0,0 +1,280 @@
...
+    let namespace = recording.var("EVENTHUBS_HOST", None);
+    let consumer_group = recording
+        .var_opt("EVENTHUBS_CONSUMER_GROUP", None)
+        .unwrap_or("$Default".to_string());
+    let eventhub_name = recording.var("EVENTHUB_NAME", None);
+
+    let checkpoints = checkpoint_store
+        .list_checkpoints(&namespace, &eventhub_name, &consumer_group)
+        .await?;
+    assert_eq!(checkpoints.len(), 10);
```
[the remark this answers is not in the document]
LarryOsterman: no longer relevant.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/assets.json
heaths: I thought you added some in #2826

Cargo.toml
```diff
@@ -74,6 +75,11 @@ path = "sdk/core/azure_core_test_macros"
 # azure_identity should only ever be in dev-dependencies herein
 path = "sdk/identity/azure_identity"
 
+[workspace.dependencies.azure_storage_blob]
+# azure_storage_blob should only ever be in dev-dependencies herein
+path = "sdk/storage/azure_storage_blob"
+version = "0.4.0"
```
heaths: Dependencies that are only dev-dependencies shouldn't have a version. But the fact you depend on this for your Storage Blobs checkpoint store means the comment is incorrect and misleading, right?

sdk/eventhubs/assets.json
```diff
@@ -0,0 +1,6 @@
+{
+  "AssetsRepo": "Azure/azure-sdk-assets",
+  "AssetsRepoPrefixPath": "rust",
+  "Tag": "rust/azure_messaging_eventhubs_checkpointstore_blob_e9c892cb82",
+  "TagPrefix": "rust/azure_messaging_eventhubs_checkpointstore_blob"
```
heaths: Since you moved it, use the service directory name. Might need to move the files and re-push in your `.assets/???` directory as well.

```suggestion
  "TagPrefix": "rust/eventhubs"
```

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+```toml
+[dependencies]
+azure_messaging_eventhubs_checkpointstore_blob = "0.1.0"
+azure_messaging_eventhubs = "0.20.0"
+azure_storage_blobs = "0.20.0"
+azure_identity = "0.20.0"
+```
+
+### Basic Example
```
heaths, line 24: Might went a sentence or two about what the example does.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/README.md
```diff
@@ -0,0 +1,94 @@
...
+│   │       ├── owner-2.json
+│   │       └── ...
+```
+
+## Examples
+
+See the `examples/` directory for more detailed usage examples:
+
+- `checkpoint_store_basic.rs`: Basic checkpoint operations
+- `processor_with_blob_checkpoints.rs`: Complete EventHubs processor setup
```
heaths, line 98: Perhaps in another PR, but this README needs to follow the structure and have the same front and end matter as other PRs.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+[features]
+default = []
+
+[[example]]
+name = "checkpoint_store_basic"
+path = "examples/checkpoint_store_basic.rs"
+
+[[example]]
+name = "processor_with_blob_checkpoints"
+path = "examples/processor_with_blob_checkpoints.rs"
```
heaths: Neither of these should be necessary since the name and the path match. This is usually needed when you need to require crate features or turn on other flags. Doesn't hurt, but just FYI it's unnecessary otherwise. Just more maintenance overhead.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#![recursion_limit = "128"]
+#![warn(missing_docs)]
```
heaths, line 5: Now that we have a newer rustc baseline, use `expect` instead of `warn` and make sure there's an issue open to document whatever needs it.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/src/checkpoint_store.rs
```diff
@@ -0,0 +1,324 @@
...
+/// This implementation stores checkpoint and ownership information in Azure Blob Storage,
+/// providing durable persistence for Event Hub event processing state.
+#[derive(Clone)]
+pub struct BlobCheckpointStore {
+    blob_container_client: Arc<BlobContainerClient>,
+}
+
+impl Debug for BlobCheckpointStore {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("BlobCheckpointStore").finish()
```
heaths: Nit: I recommend you either derive `SafeDebug`, or use the following to show there are members, but not for thee:


```suggestion
        f.debug_struct("BlobCheckpointStore").finish_non_exhaustive()
```

heaths: Or just `f.write_str("BlobCheckpointStore")` if you just want the name. What you have will result in `BlobCheckpointStore {}` which is a bit misleading.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/src/checkpoint_store.rs
```diff
@@ -0,0 +1,324 @@
...
+        etag: Option<Etag>,
+    ) -> Result<(Option<OffsetDateTime>, Option<Etag>)> {
+        let blob_client = self
+            .blob_container_client
+            .blob_client(blob_name.to_string());
+        let mut options = BlobClientSetMetadataOptions::default();
+        if let Some(metadata) = &metadata {
+            options.metadata = Some(metadata.clone());
+        }
+        options.if_match = etag.clone().map(|e| e.to_string());
```
heaths: Seems an unnecessary copy since you're just `to_string()`'ing it. Could probably use `etag.map(|ref e| e.to_string())`.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/src/checkpoint_store.rs
```diff
@@ -0,0 +1,324 @@
...
+        match result {
+            Ok(r) => Ok(Self::process_storage_response_metadata(
+                r.headers().get_optional_string(&LAST_MODIFIED),
+                r.headers().get_optional_string(&ETAG),
+            )?),
+            Err(e) => {
+                match e.http_status() {
+                    Some(StatusCode::PreconditionFailed) => {
+                        // If the blob has been modified since we last read it, return a NotFound error
+                        Err(e)
```
heaths: Comment seems misleading. Wouldn't this return a `PreconditionFailed` error? Line 99 deals with `NotFound`.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/src/checkpoint_store.rs
```diff
@@ -0,0 +1,324 @@
...
+                    }
+                    Some(StatusCode::NotFound) => {
+                        warn!("Blob {blob_name} not found, creating.");
```
heaths: This doesn't feel like a warning, which would often show by default. Maybe info or even debug. It's part of the intent and normal course of this implementation, right? If it doesn't exist, create it.

heaths (approved):
A few nits/thoughts, but otherwise LGTM.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/src/checkpoint_store.rs
LarryOsterman: I can't use `SafeDebug` because BlobContainerClient doesn't implement SafeDebug :(. However, I don't actually *need* debug - not sure why I have it.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/src/checkpoint_store.rs
LarryOsterman: Makes sense.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/src/checkpoint_store.rs
LarryOsterman: I made the mistake of following the C++ model here. Based on the Go codebase, this is incorrect and needs to be rewritten.

sdk/eventhubs/azure_messaging_eventhubs/README.md
```diff
@@ -2,7 +2,7 @@
 
 # Azure Event Hubs client library for Rust
 
-[Azure Event Hubs](https://azure.microsoft.com/services/event-hubs/) is a big data streaming platform and event ingestion service from Microsoft. For more information about Event Hubs see: [link](https://learn.microsoft.com/azure/event-hubs/event-hubs-about).
+[Azure Event Hubs](https://azure.microsoft.com/services/event-hubs/) is a big data streaming platform and event ingestion service from Microsoft. For more information about Event Hubs see: [this link](https://learn.microsoft.com/azure/event-hubs/event-hubs-about).
```
heaths: This reads funny:
> see: this link

Probably just:

```suggestion
[Azure Event Hubs](https://azure.microsoft.com/services/event-hubs/) is a big data streaming platform and event ingestion service from Microsoft. For more information about Event Hubs see [this link](https://learn.microsoft.com/azure/event-hubs/event-hubs-about).
```

heaths (approved):
Just a couple nits.

PR #2864, opened 2025-08-05
Simplify CODEOWNERS

RickWinter:


ronniegeraghty:
@RickWinter, looks like the team alias isn't valid. Are all the org memberships public?

heaths (changes requested):
The group has to valid.

RickWinter:
> The group has to be valid. Work with EngSys to define it.

It is valid. I believe it might just be replication delay.

.github/CODEOWNERS
```diff
@@ -68,12 +68,12 @@
 # ServiceOwner: @Azure/azure-sdk-write-keyvault
 # ServiceLabel: %KeyVault
 # PRLabel: %KeyVault
-/sdk/keyvault/           @Azure/azure-sdk-write-keyvault @heaths
+/sdk/keyvault/           @Azure/azure-sdk-write-keyvault @Azure/azure-sdk-rust-dev
```
heaths, line 71: This is overkill. I wanted to stay on here because I have traditionally owned it and using it as a sort of test for core changes.

Issue #2283, opened 2025-03-04
Consider whether to add a `ChainedTokenCredential`


RickWinter:

demoray:
This was added by https://github.com/Azure/azure-sdk-for-rust/pull/2309

demoray:
Please reconsider.

1. This the behavior of `DefaultAzureCredential` (known as DAC), but allowing  _users_ to explicitly select the allowed credential providers, which _reduces_ risks from using DAC.
2. This is available in multiple other SDKs.
    * https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/chained_token_credential.go
    * https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/Credentials/ChainedTokenCredential.cs
    * https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/azure/identity/_credentials/chained.py
3. With the Azure SDK for Rust _not_ including managed identities in DAC, developers are left to implement their own DAC replica that replicates this for that operates for local development _and_ deployed services.

This appears to be opposed to the design principals laid out here: https://azure.github.io/azure-sdk/general_introduction.html

chlowell:
I appreciate that `ChainedTokenCredential` (CTC) can be useful in some scenarios, at least in principle (in fact it's less useful than it appears), however it's a footgun because it can turn a configuration mistake or transient failure into a security incident. If your priority is writing one line of code to solve authentication in multiple environments, consider something like [this example type](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/identity/azure_identity/examples/specific_credential.rs) whose behavior is determined by external configuration and that fails safe in the absence of that configuration.

To your points:
1. The risk depends on the scenario but in any case it's safer to not offer CTC
2. Security is more important than consistency. And if we were writing the other SDKs today, they wouldn't have CTC either.
3. Or, developers can use the credential type directly corresponding to the identity they want to authenticate. Either approach is better than adding a footgun to the SDK.

chlowell:
> does this mean Microsoft are planning to refactor other libraries to use the AZURE_CREDENTIAL_KIND environment variable, and the Azure team are planning to add the AZURE_CREDENTIAL_KIND to resources where Managed Identity is enabled

That variable is used only in sample code in this repo, no plans to change that. I don't see the connection between this and ChainedTokenCredential or managed identity; can you elaborate?

[...]

PR #2873, opened 2025-08-05
Replace DefaultAzureCredential with DeveloperToolsCredential

chlowell:
DeveloperToolsCredential is effectively the DefaultAzureCredential from other SDKs, configured for dev mode. It's designed to simplify getting started with the SDK, for example by copying sample code. It only supports authenticating via dev tools such as `az` and `azd`, and its behavior isn't configurable. Developers who want to set additional options for its constituent types or change their order can instead use them directly.

Similarly to DefaultAzureCredential, DeveloperToolsCredential requests a token from each dev tool in turn, stopping when one provides a token. After receiving a token from some tool, a DeveloperToolsCredential instance uses that tool exclusively; it never tries the others again.

There are a couple big reasons for removing DefaultAzureCredential instead of implementing it as in other SDKs:

- **Default**AzureCredential implies it's always a good choice, if not the only choice, when in fact it's typically better to use a more specific type
- trying a sequence of credentials introduces uncertainty and can be dangerous (see #2283 and [guidance for other SDKs](https://aka.ms/azsdk/net/identity/credential-chains#usage-guidance-for-defaultazurecredential))


sdk/identity/azure_identity/src/developer_tools_credential.rs
```diff
@@ -0,0 +1,243 @@
...
+    sources: Vec<Arc<dyn TokenCredential>>,
+    // index of the source that first provided a token. usize::MAX indicates no source has provided a token.
+    cached_source_index: AtomicUsize,
+}
+
+impl DeveloperToolsCredential {
+    pub fn new(
+        options: Option<DeveloperToolsCredentialOptions>,
+    ) -> azure_core::Result<Arc<DeveloperToolsCredential>> {
+        let options = options.unwrap_or_default();
```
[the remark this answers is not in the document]
heaths: This violates guidelines.

sdk/identity/azure_identity/tests/default_options_test.rs
```diff
@@ -16,3 +18,10 @@ fn azd_credential_options() {
         ..Default::default()
     };
 }
+
+#[test]
+fn developer_tool_credential_options() {
+    let _options = DeveloperToolsCredentialOptions {
+        ..Default::default()
+    };
+}
```
heaths: This serves no purpose. Just remove it.

sdk/identity/azure_identity/src/lib.rs
```diff
@@ -7,14 +7,16 @@
...
 mod azure_developer_cli_credential;
 mod azure_pipelines_credential;
 mod cache;
 mod client_assertion_credential;
 #[cfg(feature = "client_certificate")]
 mod client_certificate_credential;
 mod client_secret_credential;
-mod default_azure_credential;
+#[cfg(not(target_arch = "wasm32"))]
+mod developer_tools_credential;
```
heaths, line 19: Should document in the CHANGELOG that this will no longer work in WASM.

sdk/identity/azure_identity/src/process/mod.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#![cfg_attr(target_arch = "wasm32", allow(unused))]
```
heaths: What's unused? The whole module? Then just use:

```suggestion
#![cfg(not(target_arch = "wasm32"))]
```

sdk/identity/azure_identity/examples/default_credentials.rs
heaths, line 1: Should rename this file. "default" implies too much. Maybe "developer_credential.rs"?

sdk/identity/azure_identity/src/developer_tools_credential.rs
```diff
@@ -0,0 +1,243 @@
...
+
+/// Authenticates through developer tools such as the Azure CLI.
+///
+/// It tries the following credential types, in this order, stopping when one provides a token:
+///
+/// * [`AzureCliCredential`]
+/// * [`AzureDeveloperCliCredential`]
+///
+/// `DeveloperToolsCredential` uses the first credential that provides a token for all subsequent token requests. It never tries the others again.
+#[derive(Debug)]
```
heaths: Far too risky. Instead of deriving `Debug` or even `SafeDebug`, just implement it e.g.,

```rust
use std::fmt;

impl fmt::Debug for DeveloperToolsCredential {
  fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
    f.write_str("DeveloperToolsCredential")
  }
}
```

Even if you derived `SafeDebug` and some implementation didn't use safe types like `Secret` for PII, it could leak. Don't give that opportunity.

sdk/identity/azure_identity/src/developer_tools_credential.rs
```diff
@@ -0,0 +1,243 @@
...
+    ) -> azure_core::Result<AccessToken> {
+        let cached_index = self.cached_source_index.load(Ordering::Relaxed);
+        if cached_index != usize::MAX {
+            if let Some(source) = self.sources.get(cached_index) {
+                return source.get_token(scopes, options).await;
+            }
+            // should never happen
+            return Err(Error::with_message(ErrorKind::Credential, || {
+                format!("Source index {cached_index} is out of bounds")
+            }));
```
heaths: Probably should just `panic!()` then. This indicates a serious developer issue and should never make it to production.

sdk/identity/azure_identity/src/lib.rs
chlowell: DefaultAzureCredential doesn't work in WASM either, for the same reason (can't subprocess CLIs there).

sdk/identity/azure_identity/examples/default_credentials.rs
chlowell: Thanks, I overlooked renaming this. Having read it more closely, I now want to simply delete it because it doesn't demonstrate anything difficult or even interesting and appears to have been written before constructors returned an `Arc`

sdk/identity/azure_identity/src/lib.rs
heaths: I get that. But it's still a change that would break code depending on it. I'm not saying you broke anything. It honestly should've been there before but we didn't really keep changelogs back in the track 1 era. Just saying we should mention it now because it is a change.

sdk/identity/azure_identity/CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 ### Breaking Changes
 
-- Replaced `DefaultAzureCredential` with `DeveloperToolsCredential`
+- Replaced `DefaultAzureCredential` with `DeveloperToolsCredential`. This new type is excluded from wasm32 builds because it can't authenticate in a WASM runtime environment (neither could `DefaultAzureCredential`).
```
heaths: Grammar nit:


```suggestion
- Replaced `DefaultAzureCredential` with `DeveloperToolsCredential`. This new type is excluded from WASM32 builds because it can't authenticate in a WASM runtime environment; however, neither could `DefaultAzureCredential`, which wasn't properly excluded.
```

heaths (approved):

heaths (approved):

LarryOsterman (approved):

PR #2885, opened 2025-08-07
Fix clippy 1.90 lint

heaths:
After update to `+nightly` clippy 0.1.90 (7d82b83ed5 2025-08-06) we discovered a few more lints.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -147,6 +147,7 @@ zerofrom = "0.1.5"
 zip = { version = "4.0.0", default-features = false, features = ["deflate"] }
 
 [workspace.lints.clippy]
+infallible_try_from = "allow"
 large_futures = "deny"
 uninlined_format_args = "allow"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -147,7 +147,6 @@ zerofrom = "0.1.5"
 zip = { version = "4.0.0", default-features = false, features = ["deflate"] }
 
 [workspace.lints.clippy]
-infallible_try_from = "allow"
 large_futures = "deny"
 uninlined_format_args = "allow"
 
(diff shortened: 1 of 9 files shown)


sdk/typespec/typespec_client_core/src/async_runtime/mod.rs
```diff
@@ -36,6 +36,7 @@ use std::{
     sync::{Arc, OnceLock},
 };
 
+#[cfg_attr(feature = "tokio", allow(dead_code))]
```
heaths, line 39: @LarryOsterman, as I was debating the best way to fix this, I got to thinking: since it's not exposed publicly, should we even exclude it when `tokio` isn't enabled?

Cargo.toml
```diff
@@ -147,6 +147,7 @@ zerofrom = "0.1.5"
 zip = { version = "4.0.0", default-features = false, features = ["deflate"] }
 
 [workspace.lints.clippy]
+infallible_try_from = "allow"
```
heaths: For posterity, some crates might generate `TryFrom` implements with an `Infallible` error to keep the public API consistent, so I added it here. /cc @LarryOsterman @RickWinter

sdk/typespec/typespec_client_core/src/async_runtime/standard_runtime.rs
```diff
@@ -163,9 +162,9 @@ impl AsyncRuntime for StdRuntime {
     }
 }
 
-#[derive(Debug)]
 #[cfg(not(target_arch = "wasm32"))]
-pub struct Sleep {
+#[derive(Debug)]
+struct Sleep {
```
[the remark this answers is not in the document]
heaths: The module this was in wasn't public, so it was never truly public.

Cargo.toml
heaths: Hmm, maybe this isn't true. I implemented `TryFrom` here so that the call is always `try_from` regardless of the model. But I suppose emitted crates would/should always emit `type Error = azure_core::Error`, right @antkmsft @jhendrixMSFT?

LarryOsterman (approved):

PR #2888, opened 2025-08-08
OpenTelemetry fixes from PR

### LINKED ISSUE #2827: Change `[tracing::function]` to take a named `fields=` parameter instead of an anonymous list parameter for future expansion
### Feature Summary

Change `[tracing::function]` to take a named `fields=` parameter instead of an anonymous list parameter for future expansion

### Feature Description

Change `[tracing::function]` to take a named `fields=` parameter instead of an anonymous list parameter for future expansion

### Use Case

_No response_

### Alternatives

_No response_

### Additional Context

_No response_

### LINKED ISSUE #2850: OpenTelemetry tests show red ERRORs and induce a bit of panic
I often test all crates since I'm making a lot of changes to the core stack. Just recently I noticed some red "ERROR"s fly by and had a mild panic. Turns out it was just `tracing` error levels and, given the test names around that same time, likely by design. No tests actually failed (except sometimes due to #2849).

Image

It'd be nice if we could either temporarily (for the scope of the tests?) disable tracing or something so these don't show up. I wouldn't want to do that entire (and the recording infra is what enables it) because they do come in useful, but maybe those tests could use `tracing::Scope::in_scope` or something to that effect to actual disable tracing.

### LINKED ISSUE #2853: rename new_from_global_provider to from_global_provider
```cplusplus
pub fn new_from_global_provider() -> Arc<Self> {}
```
#
Idiomatically should just be `from_global_provider()`.
#
[Created from ApiView comment](https://spa.apiview.dev/review/a8f22e93d5ba4b218116bd3169354d6b?activeApiRevisionId=1ee6b1b896914e3ea3eb005b685850a5&nId=root_mod_azure_core_opentelemetry_struct_OpenTelemetryTracerProvider_impl_function_new_from_global_provider)

LarryOsterman:
Fixes #2853 
Fixes #2850 
Fixes #2827 

Fixes to OpenTelemetry issues raised by @heaths from the OpenTelemetry PR.

### DIFF
--- a/sdk/core/azure_core_macros/src/tracing_function.rs
+++ b/sdk/core/azure_core_macros/src/tracing_function.rs
@@ -22,11 +22,11 @@ const INVALID_PUBLIC_FUNCTION_MESSAGE: &str =
 /// 1) `Result<Arc<Self>, E>`
 ///
 pub fn parse_function(attr: TokenStream, item: TokenStream) -> Result<TokenStream> {
-    if !is_function_declaration(&item) {
+    if let Err(e) = is_function_declaration(&item) {
         println!("Not a function declaration: {item}");
         return Err(syn::Error::new(
             item.span(),
-            INVALID_PUBLIC_FUNCTION_MESSAGE,
+            format!("{INVALID_PUBLIC_FUNCTION_MESSAGE}: {e}"),
         ));
     }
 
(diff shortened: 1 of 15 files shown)


sdk/core/azure_core_macros/src/tracing_function.rs
```diff
@@ -308,7 +352,7 @@ mod tests {
     #[test]
     fn test_parse_function_name_and_attributes_empty_attributes() {
         {
-            let test_stream = quote! { "Test Function", ()};
+            let test_stream = quote! { "Test Function", attributes=()};
```
heaths: While Copilot might complain as it did, could you add one with spaces around "=" as well? It shouldn't make a difference, but let's make sure. Anyone familiar with `tracing::instrument` will likely do that. Even you put spaces between attributes themselves.

sdk/core/azure_core_macros/src/tracing_new.rs
```diff
@@ -219,58 +219,34 @@ pub fn parse_new(attr: TokenStream, item: TokenStream) -> Result<TokenStream> {
...
     if segment.ident != "Arc" {
-        error!(
-            "Invalid return type for new function: Arc must be the first segment, found {:?}",
-            segment.ident
-        );
-        return Err(
-            "Invalid return type for new function: Arc must be the first segment".to_string(),
-        );
+        warn!("Arc must be the first segment, found {:?}", segment.ident);
+        return Err("Arc must be the first segment".to_string());
```
heaths: Why are you tracing the same thing you're returning in an error. That shouldn't be necessary. The error will be returned to the compiler. And all these traces aren't free and will impact development performance.

heaths (changes requested):
Just get rid of all the tracing for errors. The error will be returned to the compiler with the same message, so the non-free tracing is extraneous. If your errors aren't being bubbled up - they should be, as they are in other macros - we can look at why. There are even helpers in `syn` or `quote` (I forget which) to make them easy to inject into a `TokenStream` e.g., https://github.com/Azure/azure-sdk-for-rust/blob/dc9f794a2bd1a461d2920f88cef134c74ab8742d/sdk/typespec/typespec_macros/src/lib.rs#L23 which takes the `Result` I return from my parsing function https://github.com/Azure/azure-sdk-for-rust/blob/dc9f794a2bd1a461d2920f88cef134c74ab8742d/sdk/typespec/typespec_macros/src/safe_debug.rs#L12. That's how it's supposed to be done.

sdk/core/azure_core_macros/src/tracing_function.rs
LarryOsterman: Absolutely - FWIW, the spaces don't matter because we get a stream of tokens with the whitespace removed.

sdk/core/azure_core_macros/src/tracing_new.rs
LarryOsterman: The traces often provide additional context which is intentionally omitted from the error message. 

And I'm not 100% sure we should care about development performance for compiler errors, but i'll remove them anyway.

heaths: You could put that context into the errors. Anything that helps describe why the error happened and how to fix it seems like something any developer should see - not just those that enabled tracing.

sdk/core/azure_core_macros/src/tracing_new.rs
```diff
@@ -219,35 +219,27 @@ pub fn parse_new(attr: TokenStream, item: TokenStream) -> Result<TokenStream> {
...
     }
     if segment.arguments.is_empty() {
-        warn!("Arc must have arguments, found {:?}", segment.arguments);
         return Err("Arc must have arguments".to_string());
     }
     match &segment.arguments {
         syn::PathArguments::AngleBracketed(AngleBracketedGenericArguments { args, .. }) => {
             if args.len() != 1 {
-                warn!("Arc must have one argument, found {args:?}");
                 return Err("Arc must have one argument".to_string());
```
heaths, line 230: Per our offline discussion, I think it's fine to include more context in the errors. Just run it through `format!()`. If you want. Any information that's useful to you to fix macro problems I'm sure is useful to devs using them if it is actionable.

heaths (approved):

Issue #2725, opened 2025-06-20
Improved Error handling


vincenttran-msft:
Currently, this is the usage patterns when encountering an error:

![Image](https://github.com/user-attachments/assets/e16276bc-d0a7-4e9e-8fbb-129c0289ee57)
stdout: 
>Base error: server returned error status which will not be retried: 404
Status code: 404
Error kind: HttpResponse(NotFound, "ContainerNotFound")
Inner error kind: HttpResponse(NotFound, "ContainerNotFound")

Therefore, the two main asks to improve error handling are:

1. Exposing the raw response (so that we can get all the information returned from the service)
2. Ideally: A standard Azure Error deserialization model supported through Azure Core

Although there are several Storage specific errors (and thus several different forms the XML bodies can take), we are curious as to whether a standard Azure Error deserialization model would help to standardize error handling or at the very least simplify implementation for library-specific error models.
Thanks!

chlowell:
> 1. Exposing the raw response (so that we can get all the information returned from the service)

This is tracked by #2435

heaths:
I think what we need to ultimately do is move `HttpError` from `typespec_client_core` into `azure_core` because it (mostly) represents a [standard Azure error response](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#rest-error-response-body-structure). But in order to get the code and message out _and_ make the `RawResponse` available in the `HttpError`, I think we need to have a smaller, private `ErrorResponse` with just `code` and `message` defined as references, so we can deserialize just those as references and clone them. Clong just those two strings will be cheaper than cloning the entire response body, even if it only contains those two strings i.e., because of all the other JSON data in there.

That said, it'll still require changes to `typespec::error::Error::HttpResponse` to contain the whole response. Without having to also define headers there, maybe we have a `Bytes` for the whole `RawResponse`.

Alternatively, maybe we need to think about `typespec_client_core::Error` actually be something like a newtype on `typespec::Error`, and maybe even `azure_core::Error` newtype `typespec_client_core::Error`; though, I think the latter is unnecessary because no other Azure SDK language seems to expose the `ErrorResponse` object itself. Some languages don't even make those model types public, which might be good for GA.

PR #2894, opened 2025-08-08
Cosmos: Fix #2824 by forcing 'enable_content_response_on_write' on for read item

### LINKED ISSUE #2824: Unexpected empty response body on read_item
### Bug Title

Unexpected empty response body on read_item

### Crate Name

azure_data_cosmos

### Crate Version

0.25.0

### Description

[In the documentation](https://docs.rs/azure_data_cosmos/latest/azure_data_cosmos/clients/struct.ContainerClient.html#arguments-8) it is mentioned that the read item is always returned and the `enable_content_response_on_write` is ignored.

However, unless I set `enable_content_response_on_write` , the response body is empty and `response.into_body().await.unwrap()` fails with `Error { context: Custom(Custom { kind: DataConversion, error: Error("EOF while parsing a value", line: 1, column: 0) }) }` because the response body is empty.

### Steps to Reproduce

Follow the CRUD operations example from the [docs](https://docs.rs/azure_data_cosmos/latest/azure_data_cosmos/#crud-operation-on-items)

```rust
use serde::{Serialize, Deserialize};
use azure_data_cosmos::{CosmosClient, models::PatchDocument};

#[derive(Serialize, Deserialize)]
struct Item {
    pub id: String,
    pub partition_key: String,
    pub value: String,
}

async fn example(cosmos_client: CosmosClient) -> Result<(), Box<dyn std::error::Error>> {
    let item = Item {
        id: "1".into(),
        partition_key: "partition1".into(),
        value: "2".into(),
    };

    let container = cosmos_client.database_client("myDatabase").container_client("myContainer");

    // Create an item
    container.create_item("partition1", item, None).await?;

    // Read an item
    let item_response = container.read_item("partition1", "1", None).await?;
    let item: Item = item_response.into_body().await?;  // This will fail as the response body is empty

    Ok(())
}
```

Setting the `enable_content_response_on_write` option makes it respond with the expected item.
```rust
    // Read an item
    let mut options = ItemOptions::default();
    options.enable_content_response_on_write = true;
    let item_response = container.read_item("partition1", "1", Some(options)).await?;
    let item: Item = item_response.into_body().await?;

```

### Checklist

analogrelay:
This PR has two changes in one, but I think it's reasonable:

1. Fixed the Cosmos Pipeline code after #2834 changed behavior to return all HTTP status codes as `Ok`. We now convert non-2xx status codes to errors in the Cosmos Pipeline
2. Fixed #2824 by forcing the `enable_content_response_on_write` flag to `true` when reading an item.

I needed to do 1 in order for the tests to properly start failing on errors after #2834, and then I went in to fix the bug. Otherwise a pretty straightforward change :)

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
@@ -7,10 +7,13 @@ mod signature_target;
 use std::sync::Arc;
 
 pub use authorization_policy::AuthorizationPolicy;
-use azure_core::http::{
-    request::{options::ContentType, Request},
-    response::Response,
-    ClientOptions, Context, Method, PagerState, RawResponse,
+use azure_core::{
+    error::HttpError,
+    http::{
+        request::{options::ContentType, Request},
+        response::Response,
+        ClientOptions, Context, Method, PagerState, RawResponse,
+    },
 };
 use futures::TryStreamExt;
 use serde::de::DeserializeOwned;
@@ -64,7 +67,7 @@ impl CosmosPipeline {
         resource_link: ResourceLink,
     ) -> azure_core::Result<RawResponse> {
         let ctx = ctx.with_value(resource_link);
-        self.pipeline.send(&ctx, request).await
+        self.pipeline.send(&ctx, request).await?.success().await
     }
 
     pub async fn send<T>(
@@ -75,7 +78,7 @@ impl CosmosPipeline {
     ) -> azure_core::Result<Response<T>> {
         self.send_raw(ctx, request, resource_link)
             .await
-            .map(|r| r.into())
+            .map(Into::into)
     }
 
     pub fn send_query_request<T: DeserializeOwned + Send>(
@@ -105,7 +108,7 @@ impl CosmosPipeline {
                     req.insert_header(constants::CONTINUATION, continuation);
                 }
 
-                let resp = pipeline.send(&ctx, &mut req).await?;
+                let resp = pipeline.send(&ctx, &mut req).await?.success().await?;
                 let page = FeedPage::<T>::from_response(resp).await?;
 
                 Ok(page.into())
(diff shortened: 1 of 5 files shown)


sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -191,3 +194,39 @@ pub(crate) fn create_base_query_request(url: Url, query: &Query) -> azure_core::
...
+                http_error.error_code().map(ToOwned::to_owned),
+            );
+            Err(azure_core::Error::full(
+                error_kind,
+                http_error,
+                format!("Request failed with status code: {}", status),
+            ))
+        }
+    }
+}
```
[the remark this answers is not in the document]
heaths: There is no existing `success()` method on `Response<T, F>`.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -191,3 +194,39 @@ pub(crate) fn create_base_query_request(url: Url, query: &Query) -> azure_core::
...
+    async fn success(self) -> azure_core::Result<Self> {
+        if self.status().is_success() {
+            Ok(self)
+        } else {
+            RawResponse::from(self).success().await.map(Into::into)
+        }
+    }
+}
+
+impl ResponseExt for RawResponse {
```
heaths: @chlowell this is a good idea we might consider in `azure_core`. Combined with #2725 we could also use something like this to wrap standard Azure errors. We'd have to implement it (foreign trait otherwise), but could also allow clients to pass in custom success codes or something, much like .NET does for their DPG clients and HTTP status code categorizations.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -105,7 +108,7 @@ impl CosmosPipeline {
                     req.insert_header(constants::CONTINUATION, continuation);
                 }
 
-                let resp = pipeline.send(&ctx, &mut req).await?;
+                let resp = pipeline.send(&ctx, &mut req).await?.success().await?;
```
heaths: @analogrelay would it be better to implement this on `IntoFuture<Item = Response<T, F>>` such that we're not having to await twice? Granted, customers aren't having to type that themselves, but I imagine this compiles into separate `poll_next` calls against. If it's optimized away, fine. Just seems overkill this way.

heaths (approved):
A few thoughts, but nothing blocking. Creative!

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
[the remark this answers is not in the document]
analogrelay: FYI In this case it isn't hallucinating, I actually had an earlier commit in this PR that had added it to Response itself :)

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
analogrelay: Ooh, that's an interesting idea, I hadn't thought of doing that. I'd definitely like to remove the await chaining, especially since we're about to go from 2 awaits down to 1. I'd hate to push it back out to 2 again

Pilchie (approved):

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
analogrelay: It's possible, but I'm not a huge fan of it, so I'm going to hold off on it _until_ we want to look at promoting this to `azure_core`. The main reason it doesn't feel good is that we have to do this:

```rust
impl<T, F, Fut: IntoFuture<Output = azure_core::Result<Response<T, F>>>> ResponseExt for Fut {
}
```

This works, BUT, without min-specialization it ends up conflicting with **any other attempt** to implement `ResponseExt`. I don't want to do that unless we're sure we've got the right approach.

Maybe we need to get involved in min-specialization and try to drive it to stabilization 😅

heaths: Yeah, if we put it in core we can implement it on `RawResponse` and `Response<T>` ourselves and not have to worry about min-specialization. But, yeah, I agree with your other sentiment about that!

heaths (approved):

heaths (approved):

PR #2896, opened 2025-08-11
Added ability to test distributed tracing functionality.

### LINKED ISSUE #2769: Implement a mechanism to easily validate distributed tracing functionality for Azure Service clients.
### Feature Summary

Validation functionality for service clients.

### Feature Description

As things currently stand, it is extremely easy for Azure service clients to incorrectly implement distributed tracing.

Add test infrastructure which allows service client authors to validate that all the relevant distributed tracing spans work correctly to help ensure that everything works as expected.

### Use Case

_No response_

### Alternatives

_No response_

### Additional Context

_No response_

LarryOsterman:
Fixes #2769 - Adds ability to easily test distributed tracing.

Also adds a test example for KeyVault Secrets APIs.

### DIFF
--- a/sdk/core/azure_core_opentelemetry/tests/telemetry_service_macros.rs
+++ b/sdk/core/azure_core_opentelemetry/tests/telemetry_service_macros.rs
@@ -665,4 +665,34 @@ mod tests {
 
         Ok(())
     }
+
+    #[recorded::test()]
+    async fn test_function_tracing_tests(ctx: TestContext) -> Result<()> {
+        let credential = ctx.recording().credential().clone();
+        azure_core_test::tracing::test_instrumentation_for_api(
+            |tracer_provider| {
+                TestServiceClientWithMacros::new(
+                    "https://azuresdkforcpp.azurewebsites.net",
+                    credential,
+                    Some(TestServiceClientWithMacrosOptions {
+                        client_options: ClientOptions {
+                            request_instrumentation: Some(RequestInstrumentationOptions {
+                                tracer_provider: Some(tracer_provider.clone()),
+                            }),
+                            ..Default::default()
+                        },
+                        ..Default::default()
+                    }),
+                )
+                .unwrap()
+            },
+            |client| {
+                let client = client;
+                Box::pin(async move { client.get_with_function_tracing("index.htm", None).await })
+            },
+        )
+        .await?;
+
+        Ok(())
+    }
 }
--- a/sdk/core/azure_core_opentelemetry/tests/telemetry_service_macros.rs
+++ b/sdk/core/azure_core_opentelemetry/tests/telemetry_service_macros.rs
@@ -174,14 +174,21 @@ impl TestServiceClientWithMacros {
 mod tests {
     use super::*;
     use ::tracing::{info, trace};
-    use azure_core::http::{ExponentialRetryOptions, RetryOptions};
-    use azure_core::tracing::TracerProvider;
-    use azure_core::Result;
-    use azure_core_test::{recorded, TestContext};
+    use azure_core::{
+        http::{ExponentialRetryOptions, RetryOptions},
+        tracing::TracerProvider,
+        Result,
+    };
+    use azure_core_test::{
+        recorded,
+        tracing::{InstrumentationInformation, InstrumentedApiInformation},
+        TestContext,
+    };
     use opentelemetry::trace::{
         SpanKind as OpenTelemetrySpanKind, Status as OpenTelemetrySpanStatus,
     };
     use opentelemetry::Value as OpenTelemetryAttributeValue;
+    use typespec_client_core::http;
 
     fn create_exportable_tracer_provider() -> (Arc<SdkTracerProvider>, InMemorySpanExporter) {
         let otel_exporter = InMemorySpanExporter::default();
@@ -667,8 +674,11 @@ mod tests {
     }
 
     #[recorded::test()]
-    async fn test_function_tracing_tests(ctx: TestContext) -> Result<()> {
-        let credential = ctx.recording().credential().clone();
+    async fn test_http_tracing_tests(ctx: TestContext) -> Result<()> {
+        let recording = ctx.recording();
+        let credential = recording.credential().clone();
+        let package_name = recording.var("CARGO_PKG_NAME", None);
+        let package_version = recording.var("CARGO_PKG_VERSION", None);
         azure_core_test::tracing::test_instrumentation_for_api(
             |tracer_provider| {
                 TestServiceClientWithMacros::new(
@@ -676,7 +686,7 @@ mod tests {
                     credential,
                     Some(TestServiceClientWithMacrosOptions {
                         client_options: ClientOptions {
-                            request_instrumentation: Some(RequestInstrumentationOptions {
+                            instrumentation: Some(InstrumentationOptions {
                                 tracer_provider: Some(tracer_provider.clone()),
                             }),
                             ..Default::default()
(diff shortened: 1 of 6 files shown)


sdk/keyvault/azure_security_keyvault_secrets/tests/secret_client.rs
```diff
@@ -200,3 +207,68 @@ async fn purge_secret(ctx: TestContext) -> Result<()> {
 
     Ok(())
 }
+
+#[recorded::test]
+async fn round_trip_secret_verify_telemetry(ctx: TestContext) -> Result<()> {
+    use azure_core_test::tracing::test_instrumentation_for_api;
```
heaths: `test_instrumentation_for_api`? Looks like something that was meant to be internal leaked out - mainly, the "for_api". Can we just call it `test_instrumentation`? Concise and descriptive.

sdk/keyvault/azure_security_keyvault_secrets/tests/secret_client.rs
```diff
@@ -200,3 +207,68 @@ async fn purge_secret(ctx: TestContext) -> Result<()> {
...
+            api_calls: vec![
+                InstrumentedApiInformation {
+                    api_name: Some("KeyVault.setSecret"),
+                    api_verb: azure_core::http::Method::Put,
+                    ..Default::default()
+                },
+                InstrumentedApiInformation {
+                    api_name: Some("KeyVault.getSecret"),
+                    ..Default::default()
+                },
```
heaths: Is this whole structure asserted to be equal? How do people know that?

Instead, could we take another capture so people can assert what they want e.g., pass *in* the `InstrumentationInformation` e.g.,

```rust
|info| {
    assert_eq!(info.package_name, "azure_security_keyvault_secrets");
    assert_eq!(info.api_calls.len(), 2);
    assert_eq!(Some("KeyVault.setSecret"), info.api_calls[0].api_name);
}
```

sdk/keyvault/assets.json
```diff
@@ -2,5 +2,5 @@
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "rust",
   "TagPrefix": "rust/keyvault",
-  "Tag": "rust/keyvault_6c9f540438"
+  "Tag": "rust/keyvault_faa717aa41"
```
heaths: Nit: IMO, instrumenting an existing test might've been better to make sure it didn't alter anything. It shouldn't, but that'd help make sure.

sdk/core/azure_core_test/src/tracing.rs
```diff
@@ -174,34 +177,40 @@ impl AsAny for MockSpan {
...
+    pub spans: Vec<ExpectedSpanInformation<'a>>,
 }
 
 pub fn check_instrumentation_result(
     mock_tracer: Arc<MockTracingProvider>,
     expected_tracers: Vec<ExpectedTracerInformation<'_>>,
 ) {
+    let tracers = mock_tracer.tracers.lock().unwrap();
+    if tracers.len() != expected_tracers.len() {
+        println!("Expected tracers: {:?}", expected_tracers);
```
heaths: Copilot is wrong, but I think tests actually default to `stderr` so `eprintln!()` would be more appropriate. Please double check.

sdk/core/azure_core_test/src/tracing.rs
```diff
@@ -174,34 +177,40 @@ impl AsAny for MockSpan {
...
-pub struct ExpectedSpanInformation<'a> {
-    pub span_name: &'a str,
-    pub status: SpanStatus,
-    pub kind: SpanKind,
-    pub attributes: Vec<(&'a str, AttributeValue)>,
+    /// A set of spans which should have been generated.
+    pub spans: Vec<ExpectedSpanInformation<'a>>,
 }
 
 pub fn check_instrumentation_result(
```
heaths, line 207: Should be documented, unless it doens't need to be public. Is this also intended to be used in tests? I thought that was the `test_instrumentation_for_api` you expose.

sdk/core/azure_core_test/src/tracing.rs
```diff
@@ -225,17 +234,30 @@ pub fn check_instrumentation_result(
...
+    println!("Expected attributes: {:?}", expected.attributes);
+    println!("Found attributes: {:?}", attributes);
     for (index, attr) in attributes.iter().enumerate() {
         println!("Attribute {}: {} = {:?}", index, attr.key, attr.value);
         let mut found = false;
         for (key, value) in &expected.attributes {
             if attr.key == *key {
-                assert_eq!(attr.value, *value, "Attribute mismatch for key: {}", key);
+                // Skip checking the value for "<ANY>" as it is a placeholder
+                if *value != AttributeValue::String("<ANY>".into()) {
```
heaths, line 276: This seems unwieldy. Not test authors have to know about some sentinel value for "<ANY>" to satisfy an assert? Makes me thinking passing the `InstrumentationInformation` to be asserted in whatever way feels appropriate is the right approach even more. As a test author, I probably only want to check a few things. Like, I don't care that the `http.method` attribute is `GET` (or something like that). Don't make me inadvertently assert things I don't need to.

heaths (changes requested):
I think taking a capture that passes the `InstrumentationInformation` would be much easier to use for test authors. Happy to debate it further, though.

sdk/keyvault/azure_security_keyvault_secrets/tests/secret_client.rs
LarryOsterman: I could get behind that - originally it was intended for only a single API but when I hit the KV tests, I realized it needed to be more general. I'll fix the name.

sdk/keyvault/azure_security_keyvault_secrets/tests/secret_client.rs
LarryOsterman: Those are two separate API calls - one named `KeyVault.setSecret`, the other named `KeyVault.getSecret`, called in that order.

The whole point of this API is that you describe the API calls that you are making in your service client, and this verifies that the distributed tracing spans that correspond to those API calls are correctly made. In that model, the InstrumentationInformation is an input to the test, not an output.

sdk/keyvault/assets.json
LarryOsterman: I'm a big fan of a test doing one thing and one thing only, rather than a test doing two separate unrelated things. instrumenting an existing test means that it would be doing two different things.

sdk/core/azure_core_test/src/tracing.rs
LarryOsterman: This is for a different set of tests, structured around the actual underlying OpenTelemetry API surface (the new API is structured around a service client API surface - the caller should have minimal knowledge of OpenTelemetry (that's the whole point of the new test API - it enables OpenTelemetry testing without actually knowing about the nitty gritty details).

But I'll document them.

sdk/core/azure_core_test/src/tracing.rs
LarryOsterman: As we talked about, this is an internal detail - we don't actually care *what* value is generated for these attributes, but we do care that the attributes are generated.

sdk/core/azure_core_opentelemetry/tests/telemetry_service_macros.rs
```diff
@@ -521,6 +521,7 @@ mod tests {
             },
             ..Default::default()
         };
+        //        recording.instrument(&mut options.client_options);
```
heaths: Isn't this needed? Maybe was temporary and snuck in?

doc/distributed-tracing-for-rust-service-clients.md
```diff
@@ -412,3 +412,83 @@ pub fn get_operation_templates_lro_client(&self) -> OperationTemplatesLroClient
...
+
+The `azure_core_test::tracing` package provides functionality to allow developers to verify that their distributed tracing functionality is generating the expected tracing information.
+
+This functionality is driven from the `azure_core_test::tracing::assert_instrumentation_information` API.
+
+This function takes two closures and a structure which describes the expected API shape. The first closure is used to create an instance of the public API, the second actually executes the public API.
+
+As an example, here is a test for the `keyvault_secrets` package:
+
+```rust
```
heaths, line 423: Probably should have a module import this with a `doctest` condition to make sure it's all correct and stays correct.

heaths (approved):

doc/distributed-tracing-for-rust-service-clients.md
heaths: Never mind. That's not going to work. I tried a few things, but the way doc examples work everything has to be in the example, and the `#` prefix doesn't work with pure markdown (still renders). Not the end of the world.

sdk/core/azure_core_opentelemetry/tests/telemetry_service_macros.rs
heaths: Ah, I see. You do that in  your helper func. Removing since I was trying to finish up this PR for you anyway.

sdk/core/azure_core_test/src/tracing.rs
```diff
@@ -225,17 +245,37 @@ pub fn check_instrumentation_result(
     }
 }
 
+/// Information about an expected span. Used to assert span properties.
+#[derive(Debug)]
```
RickWinter, line 249: Should this be SafeDebug? 

ditto for other structs in the file

RickWinter (approved):

sdk/core/azure_core_test/src/tracing.rs
heaths: Only needed for customer-facing APIs. This crate doesn't even get published, and we want the full fidelity of Debug for stuff like that.

Issue #2503, opened 2025-04-18
ConsumerClient Errors on close after breaking from EventReceiver stream unless timeout is used


JakeDern:
### Bug Title

ConsumerClient Errors on close after breaking from EventReceiver stream unless timeout is used 

### Crate Name

azure_messaging_eventhubs

### Crate Version

0.2.0

### Description

If you stop reading from an EventReceiver stream using any method other than the timeout, the ConsumerClient throws the following errors when attempting to close it:

```
2025-04-18T22:37:43.680320Z ERROR Connection::event_loop{tx=Sender { inner: Some(Inner { state: State {
 is_complete: false, is_closed: false, is_rx_task_set: false, is_tx_task_set: false } }) } container_id
=194f4d68-8b8e-4567-96e0-914f0e5e53e4}: fe2o3_amqp::connection::engine: 594: IllegalState
2025-04-18T22:37:43.680750Z ERROR Connection::event_loop{tx=Sender { inner: Some(Inner { state: State { is_complete: false, is_closed: false, is_rx_task_set: false, is_tx_task_set: false } }) } container_id=194f4d68-8b8e-4567-96e0-914f0e5e53e4}: fe2o3_amqp::connection::engine: 606: Unable to handle error IllegalState
Error closing consumer: Transport Implementation Error: Illegal local state
```

Using timeout options when opening a receiver works just fine. For example:

```rust
    let receiver = consumer
        .open_receiver_on_partition(
            properties.partition_ids[0].clone(),
            Some(OpenReceiverOptions {
                start_position: Some(StartPosition {
                    location: StartLocation::Earliest,
                    ..Default::default()
                }),
                receive_timeout: Some(std::time::Duration::from_secs(1)),
                ..Default::default()
            }),
        )
        .await?;

    let mut receive_stream = receiver.stream_events();

    println!("Created receive stream");

    while let Some(event) = receive_stream.next().await {
        let event = event?;
        println!("Partition ID: {:?}", event.partition_key());
        println!("Event offset: {:?}", event.offset());
    }

// Fine
consumer.close();
```

However, if you want to do anything that can't be expressed via a simple timeout e.g. read 10 messages and then break the loop or use a cancellation token then the consumer client throws. For example:

```rust
    let receiver = consumer
        .open_receiver_on_partition(
            properties.partition_ids[0].clone(),
            Some(OpenReceiverOptions {
                start_position: Some(StartPosition {
                    location: StartLocation::Earliest,
[...]

PR #2899, opened 2025-08-12
Added customer connection lost test from issue #2503

LarryOsterman:
Also enabled the ability to add fe2o3_amqp traces if needed with an AMQP feature.

### DIFF
--- a/sdk/core/azure_core_amqp/Cargo.toml
+++ b/sdk/core/azure_core_amqp/Cargo.toml
@@ -40,7 +40,6 @@ tracing-subscriber = { workspace = true, features = ["env-filter"] }
 default = ["fe2o3_amqp"]
 cplusplus = []
 test = []
-amqp_tracing = ["fe2o3-amqp/tracing"]
 fe2o3_amqp = [
   "dep:fe2o3-amqp",
   "fe2o3-amqp-types",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -352,6 +352,7 @@ dependencies = [
  "azure_identity",
  "azure_messaging_eventhubs",
  "criterion",
+ "fe2o3-amqp",
  "futures",
  "rand 0.9.2",
  "rand_chacha 0.9.0",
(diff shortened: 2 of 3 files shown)


sdk/core/azure_core_amqp/Cargo.toml
```diff
@@ -40,6 +40,7 @@ tracing-subscriber = { workspace = true, features = ["env-filter"] }
 default = ["fe2o3_amqp"]
 cplusplus = []
 test = []
+amqp_tracing = ["fe2o3-amqp/tracing"]
```
heaths: A few thoughts come to mind. First and foremost - and I've started taking this stance with various `reqwest` options in `azure_core` - I don't think we want to start 1) trying to match parity with dependencies' features, and 2) advertise specific dependencies. At least 2 you did, but because `fe2o3_amqp` is optional, this should be:


```suggestion
amqp_tracing = ["fe2o3-amqp?/tracing"]
```

If we added support for other backends in the future, you could add them to this single feature.

That said, and considering what I said above, wouldn't it be better that someone just take a dependency on `fe2o3-amqp` themselves and add the feature? Resolve 2+ will unify (combine) them. This is the stance I think we want to take with various `reqwest` features apart from a few sane options like gzip and deflate support. Maybe even `rustls` given the issues that's causing for some customers.

/cc @RickWinter

heaths: And if we go that route, in your crate's `Cargo.toml` either add their `tracing` feature in your `dev-dependencies` for `fe2o3-amqp` (a separate import that adds features - I do this elsewhere and it works fine), or if you just want to do it for this one test, you could try declaring a `[[test]]` table just for this one file and add `required-features = ["fe2o3-amqp/tracing"]` and see if that works; though, I honestly don't know if it will...and kinda doubt it will, but worth a try.

heaths (changes requested):

sdk/core/azure_core_amqp/Cargo.toml
LarryOsterman: It's not just for this test. I'll go the dev-dependencies route.

PR #2906, opened 2025-08-13
cosmos: add experimental c wrapper around cosmos sdk

analogrelay:
This is a bit of an experiment, but I'm game to merge it if @heaths and other Cosmos folks like it.

We want to explore the option of publishing a C library of the Azure Cosmos DB SDK for Rust. This PR adds the basic framework to start that effort:

* A new package `cosmosclient` that depends on `azure_data_cosmos`. This package builds a `cdylib` and `staticlib` target (`libcosmosclient.a`/`libcosmosclient.so`/`cosmosclient.dll`/etc.) that exports a C-compatible API. For now, that API is extremely simple (`cosmosclient_version`).
* A simple test infrastructure (NOT currently integrated into anything). There are C programs in `sdk/cosmos/cosmosclient/c_tests` that will be built and linked against the cosmosclient library. The tests are simple pass/fail programs. If the test returns a `0` exit code, it passed, otherwise, it failed.
* `build.rs` script in `cosmosclient` to produce `sdk/cosmos/cosmosclient/include/cosmosclient.h`, which is a C header file defining the API surface of `cosmosclient`. We should publish this as an artifact at some point, but for now it's there to be used by tests.


sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -37,6 +37,7 @@ workspace = true
 
 [features]
 default = ["hmac_rust"]
+_internal_c = [] # Internal feature used to expose additional types/functions as needed to build libcosmosclient.
```
heaths: Might consider `cplusplus` like @LarryOsterman used for exposing some C bindings and binding-friendly types for `azure_core_amqp`.

heaths: That said - and GitHub won't let me comment on unchanged lines - should your `package.metadata.docs.rs` table below list `hmac_rust` and `hmac_openssl` as well? I'd put the feature names on separate lines for future diffing as well.

sdk/cosmos/cosmosclient/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "cosmosclient"
+description = "C library wrapping the Azure CosmosDB SDK for Rust"
+version = "0.27.0"
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+
```
heaths, line 11:
```suggestion
publish = false
```

sdk/cosmos/cosmosclient/src/lib.rs
```diff
@@ -0,0 +1,33 @@
+use std::ffi::{c_char, CStr, CString};
```
heaths: Should still include copyright headers.

heaths: On all files where possible. Microsoft OSS guidelines.

Cargo.toml
```diff
@@ -20,7 +20,7 @@ members = [
   "sdk/template/azure_template_core",
   "sdk/template/azure_template",
   "sdk/storage/azure_storage_common",
-  "sdk/storage/azure_storage_blob",
+  "sdk/storage/azure_storage_blob", "sdk/cosmos/cosmosclient",
```
heaths: Separate line.

sdk/cosmos/cosmosclient/src/lib.rs
analogrelay: 🤦🏻‍♀️ I really just need a tool that does it for me. I used [addlicense](https://github.com/google/addlicense) in a different project, but it's not something easily integrated into cargo :(.

Cargo.toml
analogrelay: Cargo new just sticks it on the end. Disappointing. Will fix!

sdk/cosmos/azure_data_cosmos/Cargo.toml
analogrelay: Don't want `cplusplus` because this is very explicitly C (though it'll work in C++ too), but I could do `c`. I actually started with that but wanted to make really clear that this feature basically only exists to be enabled by `cosmosclient`.

analogrelay: > That said - and GitHub won't let me comment on unchanged lines - should your package.metadata.docs.rs table below list hmac_rust and hmac_openssl as well? I'd put the feature names on separate lines for future diffing as well.

Yep, I can add that too while I'm here. I don't think we have any APIs that vary on those features so it shouldn't actually impact doc generation, but it's worth doing.

Cargo.toml
heaths: Make sure you install Even Better Toml and there are formatting instructions in the repo. Granted, you have to re-save it in VSCode.

sdk/cosmos/cosmosclient/src/lib.rs
heaths: Some LSPs support this from .editorconfig. Sadly, not rust-analyzer.

Cargo.toml
analogrelay: Yeah, it's the resaving that got me. I didn't even open `Cargo.toml` 😅 [](url)

sdk/cosmos/azure_data_cosmos/Cargo.toml
heaths: Good point about not using "cplusplus" becuase it's C, but I still wouldn't preface it with `_` and mentioning "internal" also probably isn't needed. AMQP's `cplusplus` is also for "internal" use in the same manner. It's more like an API that is for "Microsoft use only". Maybe either "c" (awfully short, though) or "cbind" or something like that? Not too picky on the name. Even "internal_c" isn't so bad, but don't like the `_` prefix.

sdk/cosmos/cosmosclient/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+    }
+};
+
+#[no_mangle]
+/// Returns a constant C string containing the version of the Cosmos Client library.
+pub extern "C" fn cosmosclient_version() -> *const c_char {
+    let version = env!("CARGO_PKG_VERSION");
+    CString::new(version)
+        .expect("failed to create CString from version")
+        .into_raw()
```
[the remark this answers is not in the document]
analogrelay: You're absolutely right Copilot! The fun part is that Copilot also wrote this code 🤣 . I should have reviewed it more closely 😅 .

sdk/cosmos/azure_data_cosmos/Cargo.toml
heaths: If we wanted to unify for C/C++ bindings, maybe we could do that across the board e.g., `cbind` or something. /cc @LarryOsterman 

Not that we have to, but some consistency for similar patterns might be helpful.

analogrelay: I've also seen `c_api` out there, but usually that's gating actual `extern "C"` APIs, which isn't quite what this does.

sdk/cosmos/cosmosclient/src/macros.rs
```diff
@@ -0,0 +1,21 @@
+macro_rules! c_str {
```
heaths: Copyright header. ;)

sdk/cosmos/cosmosclient/src/macros.rs
```diff
@@ -0,0 +1,21 @@
...
+    ($s:expr) => { const {
```
heaths: `CStr` can't do this already?

heaths (approved):

sdk/cosmos/cosmosclient/src/macros.rs
analogrelay: 😭😭😭😭😭😭😭

sdk/cosmos/cosmosclient/src/macros.rs
analogrelay: Not that I've found, no. There are CStr literals but we want to read in an environment variable.

sdk/cosmos/azure_data_cosmos/Cargo.toml
analogrelay: I went with `c` for now

sdk/cosmos/cosmosclient/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "cosmosclient"
```
LarryOsterman: possibly `cosmos_c_client` for disambiguation?

sdk/cosmos/cosmosclient/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+#[macro_use]
+mod macros;
+
+#[no_mangle] // Necessary to prevent the compiler from stripping it when optimizing
+/// cbindgen:ignore
+pub static BUILD_IDENTIFIER: &CStr = c_str!(env!("BUILD_IDENTIFIER"));
+
+const VERSION: &CStr = c_str!(env!("CARGO_PKG_VERSION"));
+
+#[no_mangle]
```
LarryOsterman: This is a 100% style nit, but I generally prefer putting attributes after the rustdoc comment - since the attributes bind to the function I like putting them as close to the function declaration as I can.

Putting them before the rustdoc comment leads me to forget they exist :).

sdk/cosmos/cosmosclient/Cargo.toml
analogrelay: I agree it's a little bit ambiguous.

The problem is that the `name` field determines the name of the output library (not just the file name, but also internal metadata in the library files themselves) and there doesn't appear to be a way to change that. Though I may be able to do something by explicitly specifying the `[lib]` target :thinking: let me give that a shot.

sdk/cosmos/cosmosclient/src/lib.rs
analogrelay: That totally makes sense, I agree.

sdk/cosmos/cosmosclient/Cargo.toml
analogrelay: Ok, that works. `package.name` and `lib.name` can be different, the former determines the name of the package for things like `cargo build --package` and the latter determines the output binary name.

Cargo.toml
```diff
@@ -93,8 +93,10 @@ async-trait = "0.1"
 base64 = "0.22"
 bytes = "1.0"
 cargo_metadata = "0.18.1"
+cbindgen = "0.29.0"
 clap = { version = "4.5.45", features = ["derive"] }
 criterion = { version = "0.5", features = ["async_tokio"] }
+dotenvy = "0.15.7"
```
heaths: This and `ureq` I intentionally didn't include here because I don't want other crates using them incidentally. That's why there are entries in `eng/scripts/verify-dependencies.rs`. This is a blocker.

Cargo.toml
```diff
@@ -11,7 +11,6 @@ members = [
   "sdk/core/azure_core_test_macros",
   "sdk/core/azure_core_opentelemetry",
   "sdk/cosmos/azure_data_cosmos",
-  "sdk/cosmos/azure_data_cosmos_native",
```
heaths: Why'd you remove this?

heaths (approved):

Cargo.toml
analogrelay: Ah, I made the mistake of running `git restore --source=main Cargo.toml` to remove dotenvy and ureq, which also blew away this change :(

heaths:
The vcpkg issue should be fixed now after #3076 but you also have a problem the Analyze step found:

```
Caused by:
  failed to parse manifest at `/mnt/vss/_work/1/s/sdk/cosmos/azure_data_cosmos_native/Cargo.toml`

Caused by:
  error inheriting `cbindgen` from workspace root manifest's `workspace.dependencies.cbindgen`

Caused by:
  `dependency.cbindgen` was not found in `workspace.dependencies`
```

heaths:
You'll need to rebase again. There was some breaking change in `eng/common` that went in and is fixed for JS and Rust. That said, there are still some analysis errors. I'm also concerned that you use a specific version in `build-dependencies` and should've been flagged. We generally don't allow that in `eng/scripts/verify-dependencies.rs`. We catch those normally. I'd rather make sure all versions are consistent at time of release.

analogrelay:

PR #2907, opened 2025-08-13
Update CODEOWNERS

ReilleyMilne:
# Request for a standard CODEOWNERS file across all SDK Language Repositories.

CODEOWNER changes:
Sorting Client/Data plane libraries.
Entry formatting.
Client SDKs renamed to Client Libraries and added Management Libraries as well.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -71,7 +71,7 @@
 /sdk/storage/            @heaths @RickWinter @LarryOsterman @vincenttran-msft @jalauzon-msft
 
 # ServiceLabel: %Storage
-# AzureSDKOwners: @heaths
+# AzureSDKOwners: @vincenttran-msft @jalauzon-msft
 
 ####################
 # Management Libraries


.github/CODEOWNERS
```diff
@@ -50,26 +50,33 @@
...
-# PRLabel: %Cosmos
-/sdk/cosmos/             @analogrelay @Pilchie @kirankumarkolli @tvaron3 @FabianMeiswinkel @kundadebdatta @nehrao1 @kushagraThapar
 
-# ServiceOwner: @Azure/azure-sdk-write-keyvault
-# ServiceLabel: %KeyVault
 # PRLabel: %KeyVault
 /sdk/keyvault/           @Azure/azure-sdk-write-keyvault @heaths
 
+# ServiceLabel: %KeyVault
+# ServiceOwner: @Azure/azure-sdk-write-keyvault
```
heaths: Comments are out of order and harder to maintain.

weshaggard:
@heaths - @jsquire and @ronniegeraghty can give the details but the goal here is to try to standardize the CODEOWNERs files a little so we can create some MCP tools that can update them.

heaths:
Standard format, sure. Why should order matter? Enforcing that just puts more of a burden on human contributors. In any case, the labels in this PR aren't in the correct order.

heaths:
> > Standard format, sure. Why should order matter? Enforcing that just puts more of a burden on human contributors. In any case, the labels in this PR aren't in the correct order.
> 
> We're trying to organize them alphabetically by service label. Ideally this will make it easier for human contributors to find the CODEOWNERS block they are looking for if they're sorted alphabetically.
> 
> They appear to be in alphabetical order to me by service label. Cosmos -> Key Vault -> Storage. @heaths, am I missing what you are seeing?

@weshaggard, @ReilleyMilne, and I talked offline and the comments should be grouped together with path - above it, specifically. That's how it's generally been. Some comments for KV are below `/sdk/keyvault`, for example. That's the blocker here.

My long-term concern is about putting yet another barrier for devs. They have to alphabetize entries in CODEOWNERS? That's more of a burden for people than bots which could just, for example, add them to the end or at least find the first place alphabetically to add them going down the list. There will still be human contributors, so it feels like we're overoptimizing for automation that can do whatever we want and making the experience worse for people that don't read often read the guidelines and will just see build breaks saying something like, "You must alphabetize the CODEOWNERS file." That's overly burdensome, IMO.

And when you going to do that? During PRs? Some nightly run that will create PRs we have to review? I don't see what value this adds. Yes, having a standard format - which we more or less already had - makes sense, and why I asked for changes here: it wasn't followed in the case of KV, at least. Sorting it doesn't really make sense. There's no reasons machines need it to be sorted in this case.

heaths:
Couldn't the MCP just as easily sort its *new* entries, such that it would put that narrower path after the wider one? It could even check if there are any wider entries below that (for different nesting).

Again, though, that's not what's blocking this PR. We can discuss offline rather than burden this PR.

RickWinter:
@weshaggard, @ronniegeraghty Do we have a database or single source of truth for who owns what service. The repos have different structure so you'll never have the same files across all repos. If you had the database of who owns what service then CoPilot (no MCP needed) could be told what to do with a simple prompt.

.github/CODEOWNERS
```diff
@@ -50,26 +50,30 @@
...
 # PRLabel: %Cosmos
 /sdk/cosmos/             @analogrelay @Pilchie @kirankumarkolli @tvaron3 @FabianMeiswinkel @kundadebdatta @nehrao1 @kushagraThapar
 
-# ServiceOwner: @Azure/azure-sdk-write-keyvault
 # ServiceLabel: %KeyVault
+# ServiceOwners: @Azure/azure-sdk-write-keyvault
 # PRLabel: %KeyVault
 /sdk/keyvault/           @Azure/azure-sdk-write-keyvault @heaths
 
+# ServiceLabel: %Storage
```
jsquire: This, for example, is invalid because the blocks run together.    The linter should have flagged this....

ReilleyMilne: It wasn't being flagged due to a spelling mistake in ServiceOwners.

weshaggard:
@jsquire why is that? Based on the documentation https://github.com/Azure/azure-sdk-tools/blob/main/tools/codeowners-utils/METADATA.md I assumed they could all be attached and should be attached to the path if there is a path.

heaths:
> > I talked offline and the comments should be grouped together with path - above it, specifically. That's how it's generally been. Some comments for KV are below /sdk/keyvault, for example.
> 
> @heaths: Not sure exactly what you meant, but blocks cannot be grouped as I believe you're describing due to how the automation works.
> 
> ### GOOD
> ```
> #ServiceLabel:  %KeyVault
> #ServiceOwners:  @someperson @otherperson
> 
> #PRLabel: KeyVault
> /sdk/keyvault/Azure.Security.KeyVault   @someperson @otherperson @ionlyreview
> ```
> 
> ### WON'T WORK
> ```
> #ServiceLabel:  %KeyVault
> #ServiceOwners:  @someperson @otherperson
> #PRLabel: KeyVault
> /sdk/keyvault/Azure.Security.KeyVault   @someperson @otherperson @ionlyreview
> ```

I didn't realize that. At the very least, can we still keep the comments above the path line instead of a mix of above and below? If we're going for consistency, let's keep it consistently above e.g.,

```codeowners
#ServiceLabel:  %KeyVault
#ServiceOwners:  @someperson @otherperson

#PRLabel: KeyVault
/sdk/keyvault/Azure.Security.KeyVault   @someperson @otherperson @ionlyreview
```

.github/CODEOWNERS
```diff
@@ -54,22 +54,25 @@
...
-# ServiceLabel: %Storage
-# AzureSDKOwners: @heaths
+# ServiceLabel: %KeyVault
+# ServiceOwners: @Azure/azure-sdk-write-keyvault
+
 # PRLabel: %Storage
 /sdk/storage/            @heaths @RickWinter @LarryOsterman @vincenttran-msft @jalauzon-msft
 
+# ServiceLabel: %Storage
+# AzureSDKOwners: @heaths
```
heaths:
```suggestion
# AzureSDKOwners: @vincenttran-msft @jalauzon-msft
```

PR #2908, opened 2025-08-13
Add TryFrom for RequestContent<T, XmlFormat>

heaths:


heaths:
This can't possibly work because we have no way to pass a root element through, and for a `HashMap` can't pass the key and value tag names through. This works for service crate models because they can implement `TryFrom` for those models since they are defined within the crate, so there's no orphan problem.

PR #2909, opened 2025-08-14
Refactor pager, poller modules

### LINKED ISSUE #2773: Refactor pager into pager module, re-exporting only consumer types from http module
Similar to the `poller` module for #2458 we should only expose the `Pager`, `ItemIterator`, and `PageIterator` from the `http` module. This will require a change in the emitter.

heaths:
Resolves #2773 by only re-exporting types that end-user developers would interact with from `http`, leaving types geared more toward emitters and convenience clients in separate `http::pager` and `http:poller` modules.

Also adds simple mocked examples.

### DIFF
--- a/sdk/core/azure_core/examples/core_pager.rs
+++ b/sdk/core/azure_core/examples/core_pager.rs
@@ -1,0 +1,91 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+use azure_core::{
+    credentials::TokenCredential,
+    http::{headers::Headers, HttpClient, Method, RawResponse, StatusCode, TransportOptions},
+};
+use azure_core_test::{credentials::MockCredential, http::MockHttpClient};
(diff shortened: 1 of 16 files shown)


heaths:
The more I think about this, though, the more I'm not sure. While `Poller` and `Pager` are types for end-user developers, they'd rarely - if ever - have to type them or import them directly because they are returned by client methods. Do we instead just leave them all within `http::poller` and `http::pager`? /cc @antkmsft @jhendrixMSFT @analogrelay

heaths:
> they'd rarely - if ever - have to type them or import them directly because they are returned by client methods

I'm not sure that sways my opinion to re-export them from `http` the more I think about it. Doing so sort of elevates them to something worth looking at - advertises them, I suppose, in the top-level `http` docs. The rest of the types they don't really need to be concerned with unless they are actually writing a convenience client. I'm leaning toward this PR as-is, but certainly open to feedback otherwise.

antkmsft (approved):

PR #2917, opened 2025-08-18
Add TryFrom for RequestContent<String, JsonFormat>

jhendrixMSFT:
To handle JSON string bodies.

### DIFF
--- a/sdk/typespec/typespec_client_core/src/http/request/mod.rs
+++ b/sdk/typespec/typespec_client_core/src/http/request/mod.rs
@@ -451,6 +451,16 @@ mod json {
     impl_try_from!(f32, f64);
     impl_try_from!(Value);
 
+    impl TryFrom<String> for RequestContent<String, JsonFormat> {
+        type Error = crate::Error;
+        fn try_from(body: String) -> crate::Result<Self> {
+            Ok(Self {
+                body: crate::json::to_json(&body)?.into(),
+                phantom: PhantomData,
+            })
+        }
+    }
+
     impl TryFrom<Value> for RequestContent<Value, JsonFormat> {
         type Error = crate::Error;
         fn try_from(body: Value) -> crate::Result<Self> {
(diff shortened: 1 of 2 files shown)


sdk/typespec/typespec_client_core/src/http/request/mod.rs
```diff
@@ -451,6 +451,16 @@ mod json {
     impl_try_from!(f32, f64);
     impl_try_from!(Value);
 
+    impl TryFrom<String> for RequestContent<String, JsonFormat> {
```
heaths, line 454: I'm worried that this is similar to our existing `FromStr` and `From<&'static str>` implementations already, but those work *very* differently from this: they take the strings - treated as UTF-8 - and treat them as raw JSON while this converts it to a JSON string. I think this would be very confusing for devs who passed a `String` vs. a `&String` (intending as a `&str`, or `String::as_ref()`.

@analogrelay @LarryOsterman what are your thoughts here? I'm disinclined to take this because of the major behavioral difference in a subtly different API.

jhendrixMSFT: Is there a way we can better disambiguate those APIs? Who's their target audience?

heaths: Users of our SDKs e.g.,

```rust
let m = CreateModel {
  name: Some("Bob".into()),
  ..Default::default()
};
client.create_model(m.try_into()?, None).await?;
```

Specifically for the strings, it's meant to handle both tests and fs/simple scenarios like:

```rust
let m = fs::read_to_string("data.json"); // or: r#"{"value":false}"}"#
client.create_model(m.try_into(), None).await?;
```

heaths: For spector, code could just as easily pack a `String` into a `Value::String`.

Issue #2901, opened 2025-08-12
Significant performance decrease with stock reqwest client


heaths:
Someone benchmarked out stock `reqwest` client and found ~10x decrease in performance compared to their custom `reqwest` client.

> Benchmarking with this SDK as is showed a 10x slowdown from the previous hand rolled client. After switching to using our reqwest client via `TransportOptions` (and making no other changes), performance came back up to around the same as it was before. 

 _Originally posted by @skeet70 in [#2750](https://github.com/Azure/azure-sdk-for-rust/issues/2750#issuecomment-3178358362)_

heaths:
Interesting. I wonder how much of this is a different default TLS implementation vs. that we activate `gzip` and `deflate` compression. If you wouldn't mind, if you enable those features on `reqwest`, do you start getting similar numbers to stock `azure_core`?

We didn't enable them by default before in the legacy codebase: https://github.com/Azure/azure-sdk-for-rust/blob/dce289dc72dbe4efbf375f2c6efb6dd1c435e3f5/sdk/core/Cargo.toml#L57-L59. I figured with modern machines it'd be faster to compress locally vs. increased bandwidth over the wire, but perhaps not...rather, it varies with payload size but we should probably not enable it by default then. Devs using clients that tend to send larger payloads e.g., Document Search (which we don't currently support anyway) could enable it.

skeet70:
With `gzip` and `deflate` features enabled (which should also make them the default according to docs in my `::new()`) I get:

```
nix-shell -p apacheHttpd --run "ab -n 1000 -k -H 'Authorization: cmk 0WUaXesNgbTAuLwn' -T application/json -p body.json http://127.0.0.1:7777/api/1/document/wrap"

<...>

Document Path:          /api/1/document/wrap
Document Length:        468 bytes

Concurrency Level:      1
Time taken for tests:   112.785 seconds
Complete requests:      1000
Failed requests:        0
Keep-Alive requests:    1000
Total transferred:      601000 bytes
Total body sent:        305000
HTML transferred:       468000 bytes
Requests per second:    8.87 [#/sec] (mean)
Time per request:       112.785 [ms] (mean)
Time per request:       112.785 [ms] (mean, across all concurrent requests)
Transfer rate:          5.20 [Kbytes/sec] received
                        2.64 kb/s sent
                        7.84 kb/s total

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
[...]

PR #2927, opened 2025-08-20
Potential fix for Issue #2901

LarryOsterman:
Re-enabled HTTP connection pooling. In benchmarks using `criterion`, this results in a 3x performance improvement on repeated HTTP operations against the same server, which is not quite the 10x performance reduction reported in #2901, but is a significant improvement.

### DIFF
--- a/sdk/core/azure_core/benches/http_transport_benchmarks.rs
+++ b/sdk/core/azure_core/benches/http_transport_benchmarks.rs
@@ -1,0 +1,210 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+use azure_core::{
+    credentials::TokenCredential,
+    fmt::SafeDebug,
+    http::{
+        policies::{BearerTokenCredentialPolicy, Policy},
+        ClientMethodOptions, ClientOptions, HttpClient, Method, Pipeline, RawResponse, Request,
+        TransportOptions, Url,
+    },
+    Result,
+};
+use azure_identity::DeveloperToolsCredential;
+use criterion::{criterion_group, criterion_main, Criterion};
+use std::sync::Arc;
+
+#[derive(Clone, SafeDebug)]
+pub struct TestServiceClientOptions {
+    pub client_options: ClientOptions,
+    pub api_version: Option<String>,
+}
+
+impl Default for TestServiceClientOptions {
+    fn default() -> Self {
+        Self {
+            client_options: ClientOptions::default(),
+            api_version: Some("2023-10-01".to_string()),
+        }
+    }
+}
+
+pub struct TestServiceClient {
+    endpoint: Url,
+    api_version: String,
+    pipeline: Pipeline,
+}
+
+#[derive(Default, SafeDebug)]
+pub struct TestServiceClientGetMethodOptions<'a> {
+    pub method_options: ClientMethodOptions<'a>,
+}
+
+impl TestServiceClient {
+    pub fn new(
+        endpoint: &str,
+        credential: Arc<dyn TokenCredential>,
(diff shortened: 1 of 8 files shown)


sdk/keyvault/azure_security_keyvault_keys/benches/benchmarks.rs
```diff
@@ -77,11 +81,171 @@ fn key_operations_benchmark(c: &mut Criterion) {
...
+
+    let dek = rand::random::<[u8; 32]>(); // Generate a random 32-byte data encryption key (DEK)
+
+    c.bench_function("key_2901_benchmark_default", |b| {
+        b.to_async(&rt).iter(|| async {
+            let parameters = KeyOperationParameters {
+                algorithm: Some(EncryptionAlgorithm::RsaOAEP256),
+                value: Some(dek.to_vec()),
+                ..Default::default()
+            };
```
heaths, line 140: Might want to do this outside the benchmark and clone it as needed. The serialization will take a bit of time as well. No reason to keep doing it. IIRC, the perf repro wasn't doing so, though I could be mistaken.

sdk/keyvault/azure_security_keyvault_keys/benches/benchmarks.rs
```diff
@@ -77,11 +81,171 @@ fn key_operations_benchmark(c: &mut Criterion) {
     });
 }
 
+fn key_2901_benchmark_default(c: &mut Criterion) {
+    const ENV_NAME: &str = "AZURE_KEYVAULT_URL";
```
heaths: Nit: Might as well define this in the module since it's being used multiple times now. No reason to duplicate it and risk it getting out of sync.

sdk/core/azure_core/benches/http_transport_benchmarks.rs
```diff
@@ -0,0 +1,210 @@
...
+            return Err(azure_core::Error::message(
+                azure_core::error::ErrorKind::Other,
+                format!("{endpoint} must use http(s)"),
+            ));
+        }
+        endpoint.set_query(None);
+
+        let auth_policy: Arc<dyn Policy> = Arc::new(BearerTokenCredentialPolicy::new(
+            credential,
+            vec!["https://vault.azure.net/.default"],
```
heaths: 1. Do we really need to auth against the hardcoded endpoint? Is that part of the test?
2. If we do, seems we shouldn't use KV's scope here.

heaths (approved):
A few suggestions, but nothing blocking.

sdk/keyvault/azure_security_keyvault_keys/benches/benchmarks.rs
LarryOsterman: The code from the customer was doing that - I literally cut and pasted his code in and then updated it appropriately. But it doesn't actually matter at the scale of these operations.

sdk/core/azure_core/benches/http_transport_benchmarks.rs
LarryOsterman: We absolutely do not need to auth against anything - but I was trying to get to as close to the customer scenario as I could possibly get for this benchmark, and that meant I needed to authenticate against *something*. And a keyvault scope was literally the first one I saw - I could use a different one :).

sdk/keyvault/azure_security_keyvault_keys/benches/benchmarks.rs
LarryOsterman: Oh, and it turns out that if I move it outside, I need to clone the parameters (because it's invoked multiple times). At which point, I'm not sure that it would change the performance at all.

PR #2928, opened 2025-08-20
Add Powershell Installation Instructions for Copilot

JennyPng:
https://github.com/Azure/azure-sdk-tools/issues/11716

image

for MCP tool usage, have copilot tell user they need to install Powershell

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -8,7 +8,7 @@ You are an expert Rust programmer. You write safe, efficient, maintainable, and 
 
 ## Prerequisites
 
-- To use Azure MCP tool calls, the user must have PowerShell installed. Provide [PowerShell installation instructions](https://learn.microsoft.com/powershell/scripting/install/installing-powershell) if not installed, and recommend restarting the IDE to start the MCP server.
+- To use Azure SDK MCP tool calls, the user must have PowerShell installed. Provide [PowerShell installation instructions](https://learn.microsoft.com/powershell/scripting/install/installing-powershell) if not installed, and recommend restarting the IDE to start the MCP server.
 
 ## Code Generation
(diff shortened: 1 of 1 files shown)


.github/copilot-instructions.md
```diff
@@ -6,6 +6,10 @@ You are an expert Rust programmer. You write safe, efficient, maintainable, and
 -   Do not be overly apologetic and focus on clear guidance.
 -   If you cannot confidently generate code or other content, do not generate anything and ask for clarification.
 
+## Prerequisites
+
+-   To use MCP tools, users must have PowerShell installed. Provide [PowerShell installation instructions](https://learn.microsoft.com/powershell/scripting/install/installing-powershell?view=powershell-7.5) if not installed, and recommend restarting VSCode to start the MCP server.
```
heaths: Which MCPs? Clearly not all. Seems like we should be concrete here. Are they Azure MCPs? Should we have a `.vscode/mcp.json` file checked in? This is too vague as is.

heaths (changes requested):

.github/copilot-instructions.md
l0lawrence: This was to target azure-sdk-mcp ([azsdkcli under azure-tools](https://github.com/Azure/azure-sdk-for-rust/tree/main/eng/common/mcp)). Yes we will need to add it/make the mcp.json. 

We can test out specifically mentioning the server name "to use azure-sdk-mcp tools"?

heaths: In that case, you should say "Azure SDK MCPs" or whatever you call the internally-focused ones. Let's not confuse users since we also have the "Azure MCPs" that are more externally-focused.

/cc @maririos

PR #2935, opened 2025-08-20
Service Bus improvements

RickWinter:
Allow capturing off all types of errors within the ServiceBusError 
Move ServiceBusClient to the clients folder.
Use constants for the keys
Fix default client version
Change method visibility to be pub(crate)

### DIFF
--- a/eng/scripts/setup-openssl-env.ps1
+++ b/eng/scripts/setup-openssl-env.ps1
@@ -15,20 +15,22 @@ Write-Host "  CMAKE_TOOLCHAIN_FILE = $env:CMAKE_TOOLCHAIN_FILE"
 
 # Check if vcpkg and OpenSSL are installed
 if (-not (Test-Path "C:\vcpkg\vcpkg.exe")) {
-    Write-Host "WARNING: vcpkg not found at C:\vcpkg\vcpkg.exe" -ForegroundColor Red
-    Write-Host "Please run the following commands to install vcpkg and OpenSSL:" -ForegroundColor Yellow
-    Write-Host "  git clone https://github.com/Microsoft/vcpkg.git C:\vcpkg"
-    Write-Host "  C:\vcpkg\bootstrap-vcpkg.bat"
-    Write-Host "  C:\vcpkg\vcpkg.exe integrate install"
-    Write-Host "  C:\vcpkg\vcpkg.exe install openssl:x64-windows-static-md"
-} elseif (-not (Test-Path "C:\vcpkg\installed\x64-windows-static-md\lib\libssl.lib")) {
-    Write-Host "WARNING: OpenSSL not found in vcpkg installation" -ForegroundColor Red
-    Write-Host "Please run: C:\vcpkg\vcpkg.exe install openssl:x64-windows-static-md"
-} else {
-    Write-Host "✓ vcpkg and OpenSSL are properly installed" -ForegroundColor Green
-    Write-Host ""
-    Write-Host "You can now run:" -ForegroundColor Cyan
-    Write-Host "  cargo clippy --workspace --all-features --all-targets --keep-going --no-deps"
-    Write-Host "  cargo build --workspace --all-features --all-targets"
-    Write-Host "  cargo test --workspace"
+  Write-Host "WARNING: vcpkg not found at C:\vcpkg\vcpkg.exe" -ForegroundColor Red
+  Write-Host "Please run the following commands to install vcpkg and OpenSSL:" -ForegroundColor Yellow
+  Write-Host "  git clone https://github.com/Microsoft/vcpkg.git C:\vcpkg"
+  Write-Host "  C:\vcpkg\bootstrap-vcpkg.bat"
+  Write-Host "  C:\vcpkg\vcpkg.exe integrate install"
+  Write-Host "  C:\vcpkg\vcpkg.exe install openssl:x64-windows-static-md"
+}
+elseif (-not (Test-Path "C:\vcpkg\installed\x64-windows-static-md\lib\libssl.lib")) {
+  Write-Host "WARNING: OpenSSL not found in vcpkg installation" -ForegroundColor Red
+  Write-Host "Please run: C:\vcpkg\vcpkg.exe install openssl:x64-windows-static-md"
+}
+else {
+  Write-Host "✓ vcpkg and OpenSSL are properly installed" -ForegroundColor Green
+  Write-Host ""
+  Write-Host "You can now run:" -ForegroundColor Cyan
+  Write-Host "  cargo clippy --workspace --all-features --all-targets --keep-going --no-deps"
+  Write-Host "  cargo build --workspace --all-features --all-targets"
+  Write-Host "  cargo test --workspace"
 }
--- a/sdk/servicebus/azure_messaging_servicebus/src/error.rs
+++ b/sdk/servicebus/azure_messaging_servicebus/src/error.rs
@@ -62,7 +62,7 @@ pub struct ServiceBusError {
 
 impl ServiceBusError {
     /// Creates a new Service Bus error.
-    pub fn new(kind: ErrorKind, message: impl Into<String>) -> Self {
+    pub(crate) fn new(kind: ErrorKind, message: impl Into<String>) -> Self {
         Self {
             kind,
             message: message.into(),
@@ -71,7 +71,7 @@ impl ServiceBusError {
     }
 
     /// Creates a new Service Bus error with a source error.
-    pub fn with_source(
+    pub(crate) fn with_source(
         kind: ErrorKind,
         message: impl Into<String>,
         source: impl std::error::Error + Send + Sync + 'static,


sdk/servicebus/azure_messaging_servicebus/src/error.rs
```diff
@@ -94,8 +94,10 @@ impl ServiceBusError {
     }
 
     /// Returns the source error, if any.
-    pub fn source(&self) -> Option<&ServiceBusError> {
-        self.source.as_deref()
+    pub fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
```
LarryOsterman, line 97: Why have a `source()` function on `ServiceBusError` which is a different implementation from the trait version? The two appear to be basically identical.

sdk/servicebus/azure_messaging_servicebus/src/error.rs
```diff
@@ -74,7 +74,7 @@ impl ServiceBusError {
     pub fn with_source(
```
LarryOsterman: Same question as before: Are customers expected to create their own `ServiceBusError` instances? If not, the `with_source` should probably be private.

sdk/servicebus/azure_messaging_servicebus/src/error.rs
```diff
@@ -123,12 +127,65 @@ impl From<azure_core::error::Error> for ServiceBusError {
...
 
-        ServiceBusError::new(kind, error.to_string())
+        ServiceBusError::with_source(kind, error.to_string(), error)
     }
 }
 
 impl From<azure_core_amqp::AmqpError> for ServiceBusError {
     fn from(error: azure_core_amqp::AmqpError) -> Self {
-        ServiceBusError::new(ErrorKind::Amqp, error.to_string())
+        ServiceBusError::with_source(ErrorKind::Amqp, error.to_string(), error)
```
LarryOsterman, line 136: Do you expect to provide descriptive texts for servicebus errors generated from other errors? If your general practice is to just use `.to_string()`, there's not a ton of value in the `message` parameter.

sdk/servicebus/azure_messaging_servicebus/src/error.rs
RickWinter: Customers should never need to construct a ServiceBusError. That should only occur within the crate. I'll fix the visibility.

eng/scripts/setup-openssl-env.ps1
heaths, line 1: 1. Needs a shebang:
   ```pwsh
   #!/usr/bin/env pwsh
   ```
2. Needs a copyright header after that (Microsoft OSS rules).
3. You need to `chmod +x` the script. If you're on Windows, run `git update-index --chmod=+x <path>`.

eng/scripts/setup-openssl-env.ps1
```diff
@@ -0,0 +1,36 @@
+# Setup script for OpenSSL environment to enable workspace-wide clippy and builds
+# Run this script before running cargo clippy --workspace or similar commands
+
+Write-Host "Setting up OpenSSL environment for Rust builds..." -ForegroundColor Green
+
+# Set VCPKG environment variables
+$env:VCPKG_ROOT = "C:\vcpkg"
+$env:OPENSSL_DIR = "C:\vcpkg\installed\x64-windows-static-md"
```
heaths, line 8: Why a global directory when we may have different versions for different repos? I have several worktrees (repos) on my machine, which is why the existing docs in our root `CONTRIBUTING.md` say to use a repo-related path.

If you're going to codify those instructions:

1. They should be compatible with what I've already tested works in more ways than this.
2. Those instruction should be updated to have people run this script instead.

eng/scripts/setup-openssl-env.ps1
```diff
@@ -0,0 +1,36 @@
...
+Write-Host "Environment variables set:" -ForegroundColor Yellow
+Write-Host "  VCPKG_ROOT = $env:VCPKG_ROOT"
+Write-Host "  OPENSSL_DIR = $env:OPENSSL_DIR"
+Write-Host "  CMAKE_TOOLCHAIN_FILE = $env:CMAKE_TOOLCHAIN_FILE"
+
+# Check if vcpkg and OpenSSL are installed
+if (-not (Test-Path "C:\vcpkg\vcpkg.exe")) {
+  Write-Host "WARNING: vcpkg not found at C:\vcpkg\vcpkg.exe" -ForegroundColor Red
+  Write-Host "Please run the following commands to install vcpkg and OpenSSL:" -ForegroundColor Yellow
+  Write-Host "  git clone https://github.com/Microsoft/vcpkg.git C:\vcpkg"
```
heaths, line 20: Why script it if you're not going to do it for them? Why is this script beneficial instead of just using the existing `CONTRIBUTING.md` if we don't basically do it all for them? Why not go that extra step?

eng/scripts/setup-openssl-env.ps1
```diff
@@ -0,0 +1,36 @@
...
+  Write-Host "  C:\vcpkg\bootstrap-vcpkg.bat"
+  Write-Host "  C:\vcpkg\vcpkg.exe integrate install"
+  Write-Host "  C:\vcpkg\vcpkg.exe install openssl:x64-windows-static-md"
```
heaths, line 23: What version? We may pin a version, which we already have in our `eng/vcpkg.json` that, BTW, you're not using.

sdk/servicebus/test-resources.bicep
```diff
@@ -35,7 +35,7 @@ resource serviceBusNamespace 'Microsoft.ServiceBus/namespaces@2024-01-01' = {
     zoneRedundant: false
   }
 
-  resource rootSharedAccessKey 'AuthorizationRules@2022-10-01-preview' = {
+  resource rootSharedAccessKey 'AuthorizationRules@2024-01-01' = {
```
heaths, line 38: Just get rid of the versions. They inherit the api-version of the parent resource. It's cleaner and less chance of incompatibility.

sdk/servicebus/test-resources-post.ps1
```diff
@@ -50,44 +50,6 @@ $resourceGroup = $DeploymentOutputs['RESOURCE_GROUP']
 Write-Host "Service Bus Namespace: $namespaceName"
 Write-Host "Resource Group: $resourceGroup"
 
-# Retrieve connection strings (these contain secrets so aren't in Bicep outputs)
```
heaths: Now that you've removed everything that uses `az` CLI, 1) you should get rid of lines 39-41 because they are unnecessary and create an additional unnecessary dependency, but then 2) why have this script at all? It does nothing useful. Just delete it.

sdk/servicebus/azure_messaging_servicebus/src/error.rs
```diff
@@ -53,16 +53,16 @@ impl fmt::Display for ErrorKind {
 }
 
 /// A Service Bus specific error.
-#[derive(SafeDebug, Clone, PartialEq, Eq)]
```
heaths: The error type should just be `Debug` not `SafeDebug`. All the information is kinda important and you should make sure nothing inside it otherwise leaks. `typespec::Error` impl `Debug`.

sdk/servicebus/azure_messaging_servicebus/src/error.rs
```diff
@@ -53,16 +53,16 @@ impl fmt::Display for ErrorKind {
...
+#[derive(SafeDebug)]
 pub struct ServiceBusError {
     kind: ErrorKind,
     message: String,
-    source: Option<Box<ServiceBusError>>,
+    source: Option<Box<dyn std::error::Error + Send + Sync + 'static>>,
```
heaths, line 60: Why add `Send`, `Sync`, and `'static` constraints? That makes it pretty difficult to send any error since not all implement those traits.

sdk/servicebus/azure_messaging_servicebus/src/error.rs
```diff
@@ -53,16 +53,16 @@ impl fmt::Display for ErrorKind {
...
 }
 
 impl ServiceBusError {
     /// Creates a new Service Bus error.
-    pub fn new(kind: ErrorKind, message: impl Into<String>) -> Self {
+    pub(crate) fn new(kind: ErrorKind, message: impl Into<String>) -> Self {
```
heaths, line 65: Why protect it? What if someone wanted to wrap our client and return their own instance of a `ServiceBusError`? I don't see any value in this.

sdk/servicebus/azure_messaging_servicebus/src/error.rs
```diff
@@ -71,10 +71,10 @@ impl ServiceBusError {
     }
 
     /// Creates a new Service Bus error with a source error.
-    pub fn with_source(
+    pub(crate) fn with_source(
```
heaths, line 74: Same here.

heaths (changes requested):

PR #2936, opened 2025-08-20
[Storage] `upload_blob_from_url` for BlobClient

vincenttran-msft:
`.tsp`: https://github.com/Azure/azure-rest-api-specs/pull/36847

- Initial introduction of `upload_blob_from_url` (renamed from generated: `put_blob_from_url`)
- Adds some more models found in options bag missed in initial implementation
- Refactors `create_test_blob` utility to accept `data` but still have a fallback for simple cases

`BlockBlobClientUploadBlobFromUrlResultHeaders`
 ̶d̶a̶t̶e̶
content_md5
last_modified
etag
encryption_key_sha256
encryption_scope
is_server_encrypted
version_id


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -182,13 +182,36 @@ impl BlobClient {
...
+    ///
+    /// * `content_length` - The blob data to upload.
+    /// * `copy_source` - A URL of up to 2 KB in length that specifies a file or blob. The value should be URL-encoded as it would appear in a request URI.
+    ///   The source must either be public or must be authenticated via a shared access signature as part of the url or using the source_authorization keyword.
+    ///   If the source is public, no authentication is required. Examples:
+    ///   `https://myaccount.blob.core.windows.net/mycontainer/myblob`
+    ///   `https://myaccount.blob.core.windows.net/mycontainer/myblob?snapshot=<DateTime>`
+    ///   `https://otheraccount.blob.core.windows.net/mycontainer/myblob?sastoken`
+    /// * `options` - Optional configuration for the request. See [`BlockBlobClientPutBlobFromUrlOptionsExt`](crate::models::BlockBlobClientPutBlobFromUrlOptionsExt) for additional usage helpers.
+    pub async fn put_blob_from_url(
```
jalauzon-msft: This should go on BlockBlobClient since its Block Blob specific.

jalauzon-msft: Let's rename this to `upload_from_url`

sdk/storage/azure_storage_blob/src/models/extensions.rs
```diff
@@ -43,6 +44,23 @@ impl AppendBlobClientCreateOptionsExt for AppendBlobClientCreateOptions<'_> {
...
 
+pub trait BlockBlobClientPutBlobFromUrlOptionsExt {
+    /// Augments the current options bag to only create if the Block blob does not already exist.
+    /// # Arguments
+    ///
+    /// * `self` - The options bag to be modified.
+    fn with_if_not_exists(self) -> Self;
+}
+
+impl BlockBlobClientPutBlobFromUrlOptionsExt for BlockBlobClientPutBlobFromUrlOptions<'_> {
```
jalauzon-msft: Can we do this without an extension?

sdk/storage/azure_storage_blob/src/clients/block_blob_client.rs
```diff
@@ -135,4 +135,27 @@ impl BlockBlobClient {
...
+    ///   The source must either be public or must be authenticated via a shared access signature as part of the url or using the source_authorization keyword.
+    ///   If the source is public, no authentication is required. Examples:
+    ///   `https://myaccount.blob.core.windows.net/mycontainer/myblob`
+    ///   `https://myaccount.blob.core.windows.net/mycontainer/myblob?snapshot=<DateTime>`
+    ///   `https://otheraccount.blob.core.windows.net/mycontainer/myblob?sastoken`
+    /// * `options` - Optional configuration for the request.
+    pub async fn upload_blob_from_url(
+        &self,
+        content_length: u64,
+        copy_source: String,
```
jalauzon-msft, line 152: This may seem like a nit, but can we change the ordering of these params? Feels like copy_source should come before length. Bonus points if this can be done in the generated code too

sdk/storage/azure_storage_blob/tests/block_blob_client.rs
```diff
@@ -107,3 +113,93 @@ async fn test_block_list(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
+    blob_client
+        .block_blob_client()
+        .upload_blob_from_url(12, overwrite_url.clone(), None)
+        .await?;
+
+    // Public Resource Scenario
+    blob_client
+        .block_blob_client()
+        .upload_blob_from_url(
+            500,
```
jalauzon-msft: Pretty sure this is longer than 500 bytes?

sdk/storage/azure_storage_blob_test/src/lib.rs
```diff
@@ -95,20 +95,31 @@ pub async fn get_container_client(
...
-            None,
-        )
-        .await
+    match data {
+        Some(content) => {
+            blob_client
+                .upload(content.clone(), true, content.body().len() as u64, None)
+                .await
+        }
+        None => {
```
heaths, line 114: It would be easier to maintain if you just make your own enum rather than having `None` mean `IllGeneratedIt` e.g.,

```rust
pub enum SampleOption<T> {
    Generate,
    Provided(T),
}
```

Or something like that.

sdk/storage/azure_storage_blob/src/clients/block_blob_client.rs
```diff
@@ -135,4 +135,27 @@ impl BlockBlobClient {
...
+    /// # Arguments
+    ///
+    /// * `content_length` - The blob data to upload.
+    /// * `copy_source` - A URL of up to 2 KB in length that specifies a file or blob. The value should be URL-encoded as it would appear in a request URI.
+    ///   The source must either be public or must be authenticated via a shared access signature as part of the url or using the source_authorization keyword.
+    ///   If the source is public, no authentication is required. Examples:
+    ///   `https://myaccount.blob.core.windows.net/mycontainer/myblob`
+    ///   `https://myaccount.blob.core.windows.net/mycontainer/myblob?snapshot=<DateTime>`
+    ///   `https://otheraccount.blob.core.windows.net/mycontainer/myblob?sastoken`
+    /// * `options` - Optional configuration for the request.
```
heaths: This is not going to render like you think. Without blank lines or a bullet denoting separate lines, lines 144 to 149 are just going to be one long line with current breaks as mere spaces between lines.

Rust uses commonmark rendering - as do most things - so the same rules apply.

sdk/storage/azure_storage_blob/tests/block_blob_client.rs
vincenttran-msft: Interesting callout here.. I assumed it was uploading the first 500 bytes (since we provided the content length), but it in fact uploads the whole book 😄 

The thing that makes this weirder is that I captured the content-length built in the generated code, yet it doesn't match up what I find in Fiddler:
```rust
request.insert_header("x-ms-version", &self.version);
        println!(
            "{:?}",
            request
                .headers()
                .clone()
                .get_str(&HeaderName::from_static("content-length"))
        );
```
>Ok("12")
Ok("12")
Ok("12")
Ok("500")

image

Either way, consulting the [REST API](https://learn.microsoft.com/en-us/rest/api/storageservices/put-block-from-url?tabs=microsoft-entra-id) here actually reveals that:
>Content-Length: Required. Specifies the number of bytes being transmitted in the request body. The value of this header must be set to 0. When the length isn't 0, the operation fails with status code 400 (Bad Request).


So will adjust generated code to always insert a 0. Interesting that this wasn't just failing for 400? Not sure where the substitution is happening since I printed from the request headers...

sdk/storage/azure_storage_blob_test/src/lib.rs
vincenttran-msft: I decided against complicating the test code anymore since this is the only change this test function will ever get (custom data input), and I feel like it is natural to just pass `None` for default behavior since that is how 99% of callers will use this function.

jalauzon-msft (approved):

heaths (approved):

PR #2945, opened 2025-08-25
Allow use of other TLS implementations for eventhubs

### LINKED ISSUE #2796: `azure_messaging_eventhubs` should allow users to use `rustls` instead of `native-tls`

LarryOsterman:
Fixes #2796 following the model used in #2933

Move `fe2o3-amqp`'s `native-tls` feature to default features for AMQP layer and use that for eventhubs.

### DIFF
--- a/sdk/core/azure_core_amqp/Cargo.toml
+++ b/sdk/core/azure_core_amqp/Cargo.toml
@@ -37,7 +37,7 @@ typespec_macros.workspace = true
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 
 [features]
-default = ["fe2o3_amqp"]
+default = ["fe2o3_amqp", "fe2o3-amqp/native-tls"]
 cplusplus = []
 test = []
 fe2o3_amqp = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,7 +94,7 @@ cargo_metadata = "0.18.1"
 clap = { version = "4.5.45", features = ["derive"] }
 criterion = { version = "0.5", features = ["async_tokio"] }
 dyn-clone = "1.0"
-fe2o3-amqp = { version = "0.14", features = ["native-tls", "uuid"] }
+fe2o3-amqp = { version = "0.14", features = ["uuid"] }
 fe2o3-amqp-ext = { version = "0.14" }
 fe2o3-amqp-management = { version = "0.14" }
 fe2o3-amqp-cbs = { version = "0.14" }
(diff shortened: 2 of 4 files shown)


sdk/core/azure_core_amqp/Cargo.toml
```diff
@@ -37,7 +37,7 @@ typespec_macros.workspace = true
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 
 [features]
-default = ["fe2o3_amqp"]
+default = ["fe2o3_amqp", "fe2o3-amqp/native-tls"]
```
heaths, line 40: Do you need to support alternatives? If so, you'll need a separate `fe2o3_amqp_native_tls` feature that probably depends on `fe2o3_amqp`; though, I'm not sure how congruent that is with `reqwest` where that's also used to just provide a blanket `HttpClient` implementation for `reqwest::Client`.

heaths (approved):

sdk/core/azure_core_amqp/Cargo.toml
LarryOsterman: I don't need to support alternatives, others might. So they can use the same no-default-features option that we do with azure core. 

They'll have to add the fe2o3_amqp feature to get things to work (since we don't support alternatives), but they can manually configure fe2o3 however they want.

PR #2948, opened 2025-08-26
Don't start test broker on macos

LarryOsterman:


heaths (approved):

heaths:
This works, though it might be nice to keep the scripts working for manual runs - make it easier in the future to figure out why they're flaky. Could just condition the pipeline steps, but I suppose this is ubiquitous.

PR #2955, opened 2025-08-27
Add a Logging policy with sanitized HTTP headers and URLs

LarryOsterman:
Also updated the `Debug` trait implementation for `Request`, `RawResponse`, and several other locations that print URL types to use the default sanitization rules.

Sanitization is done using an "allow list" - the customer (and/or a service) can provide an additional set of headers and/or URL query parameters to be allowed, but by default most query parameters and headers are redacted when logged.

As a bonus, this trims the public API surface associated with HTTP headers to the set of headers which is currently in use by the SDK internals and other service clients.


sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/examples/checkpoint_store_basic.rs
```diff
@@ -1,36 +1,114 @@
...
+use azure_storage_blob::{BlobContainerClient, BlobContainerClientOptions};
+use tracing::info;
+
+use opentelemetry_appender_tracing::layer;
+use opentelemetry_sdk::{logs::SdkLoggerProvider, Resource};
+use tracing_subscriber::{prelude::*, EnvFilter};
+
+// Configure tracing with OpenTelemetry and a stdout exporter as well as with the
+// `tracing` crate.
+fn configure_tracing() {
```
heaths: Nit: I'm not quite sure if this function or what you do in `main` is what you want to highlight in the example (or maybe both are equivalent), but I recommend putting the top billers higher up in the list and put supporting cast at the bottom of the list.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/examples/checkpoint_store_basic.rs
```diff
@@ -1,36 +1,114 @@
...
+    //
+    // Note: This filtering will also drop logs from these components even when
+    // they are used outside of the OTLP Exporter.
+    let filter_otel = EnvFilter::new("trace")
+        .add_directive("hyper=off".parse().unwrap())
+        .add_directive("hyper_util=off".parse().unwrap())
+        .add_directive("opentelemetry=off".parse().unwrap())
+        .add_directive("tonic=off".parse().unwrap())
+        .add_directive("reqwest=off".parse().unwrap())
+        .add_directive("h2=off".parse().unwrap());
```
heaths: Tip: you can also pass `add_directive("hyper=off,hyper_util=off".parse().unwrap())` as well, or something that. I remember an example where you pack more in but that's really just for better perf. This is definitely more readable. Your call.

heaths: That said, you're turning enough off you might consider something like "off,azure=trace" which defaults to off and *I think* the "azure=trace" will work well enough as a substring match, but don't quote me on that.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/examples/checkpoint_store_basic.rs
```diff
@@ -1,36 +1,114 @@
...
-    let blob_container_client =
-        BlobContainerClient::new(&storage_account_url, container, credential, None)?;
+    // Instantiate a blob client with OpenTelemetry instrumentation enabled
+    let blob_container_client = BlobContainerClient::new(
+        &storage_account_url,
+        container,
+        credential,
+        Some(BlobContainerClientOptions {
+            client_options: ClientOptions {
+                instrumentation: Some(InstrumentationOptions {
```
heaths, line 53: Similar to our discussion of the `LoggerOptions`, does this really need to be an `Option<InstrumentationOptions>`? Per previous discussions, it's often better just to have nested client/method options unless we truly need to know whether something was unspecified. I could imagine that being true *inside* `InstrumentationOptions`, but for all of it?

sdk/typespec/typespec_client_core/src/http/headers/common.rs
heaths, line 1: Similar here: I'd just consider removing changes to these for now because I want to delete most of them anyway.

sdk/typespec/typespec_client_core/src/http/options/mod.rs
```diff
@@ -13,6 +13,21 @@ use crate::http::{policies::Policy, Context};
...
+///
+/// This list is added to the default allow list of headers and query parameters.
+///
+#[derive(Clone, Debug, Default)]
+pub struct LoggingOptions {
+    /// The allowed header names to be logged.
+    pub additional_allowed_header_names: Vec<&'static str>,
+
+    /// The allowed query parameters to be logged.
+    pub additional_allowed_query_params: Vec<&'static str>,
```
heaths: I worry whether anything here could always be `&'static str`. Maybe we should consider `Cow<'static, str>` to be on the safe side?

sdk/typespec/typespec_client_core/src/http/options/mod.rs
```diff
@@ -27,6 +42,8 @@ pub struct ClientOptions {
 
     /// Transport options.
     pub transport: Option<TransportOptions>,
+
+    pub logging: LoggingOptions,
```
heaths, line 47: Comment.

But, yes, I agree with this. And makes me wonder if we should also change `TransportOptions` from an `Option`. It could instead default to a `Default` variant or something. These all take up the same amount of space either way. Treating some sub-client options as structs and others as `Option` seems weird.

Don't change it here - your code is fine - but I wanted to add this and get thoughts. /cc @RickWinter @ronniegeraghty @chlowell

sdk/typespec/typespec_client_core/src/http/policies/logging.rs
```diff
@@ -0,0 +1,210 @@
...
+    Context, LoggingOptions, Request,
+};
+use std::collections::HashSet;
+use std::sync::Arc;
+use tracing::info;
+
+/// [`Policy`] to log  a request and response.
+#[derive(Clone, Debug, Default)]
+pub struct LoggingPolicy {
+    allowed_headers: HashSet<&'static str>,
```
heaths: Do we think this will be large enough to warrant the overhead of a set? IIRC, most languages use an array. If we use a `Vec` we can order and use a `binary_search` which might be good enough.

sdk/typespec/typespec_client_core/src/http/policies/logging.rs
```diff
@@ -0,0 +1,210 @@
...
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+impl Policy for LoggingPolicy {
+    async fn send(
+        &self,
+        ctx: &Context,
+        request: &mut Request,
+        next: &[Arc<dyn Policy>],
+    ) -> PolicyResult {
+        info!(
+            "==> OUTGOING REQUEST: url: {}, method: {}, headers: {{ {} }}",
```
heaths: IMO, the format feels too foreign. We don't have anything like this now and they almost stick out too much. I actually don't mind the arrows, but maybe not all caps. Not a big issue, though. Just something to consider. They just otherwise look really different and almost scream at you when other traces might be more interesting.

sdk/typespec/typespec_client_core/src/http/sanitizer.rs
```diff
@@ -0,0 +1,144 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// cspell:ignore traceparent tracestate
+
+use crate::http::{headers::Headers, Url};
+use std::{collections::HashSet, sync::LazyLock};
+
+pub static DEFAULT_ALLOWED_HEADER_NAMES: LazyLock<HashSet<&'static str>> = LazyLock::new(|| {
```
heaths: Comment. But also, if you are using the string values directly, why did you make the headers `pub(crate)`? Again, though, I'd just remove that change and let's remove all but the helpful ones that help parse types as part of a separate work item. Keep this one focused.

sdk/typespec/typespec_client_core/src/http/sanitizer.rs
```diff
@@ -0,0 +1,144 @@
...
+    ]
+    .iter()
+    .copied()
+    .collect()
+});
+
+pub static DEFAULT_ALLOWED_QUERY_PARAMETERS: LazyLock<HashSet<&'static str>> =
+    LazyLock::new(|| ["api-version"].iter().copied().collect());
+
+pub const REDACTED_PATTERN: &str = "REDACTED";
```
heaths, line 49: Is this what most languages use? I know it's "Sanitized" for tests pretty universally, but just from a familiarity perspective, we should try to use whatever is most common. Not a big deal, though. I suspected my old "******" weren't common in our codebases anyway.

sdk/typespec/typespec_client_core/src/http/sanitizer.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// Sanitizes a type by removing or masking sensitive information based on a set of allowed patterns
+    ///
+    /// # Arguments
+    ///
+    /// * `patterns` - A slice of string slices containing patterns which are allowed in the Url.
+    ///
+    /// # Returns
+    ///
+    /// A String containing the sanitized URL
+    fn sanitize(&self, allowed_patterns: &HashSet<&'static str>) -> String;
```
heaths: Similarly, maybe just a `&[&'static str]` or even `&[Cow<'static, str>]` might be enough here if we make sure they are sorted going in.

heaths: Actually, since we're just taking a slice, even `&[&str]` should be fine and would handle this coming from either other idea above anyway, I believe. Keep parameters as simple/accepting as possible.

sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
```diff
@@ -37,8 +41,23 @@ impl RequestInstrumentationPolicy {
     /// The tracer provided is a "fallback" tracer which is used if the `ctx` parameter
     /// to the `send` method does not contain a tracer.
     ///
-    pub fn new(tracer: Option<Arc<dyn crate::tracing::Tracer>>) -> Self {
-        Self { tracer }
+    pub fn new(
+        tracer: Option<Arc<dyn crate::tracing::Tracer>>,
+        log_options: &LoggingOptions,
```
heaths: Nit: just for internal consistency, maybe just `logging` like `ClientOptions` has.

heaths (approved):
I don't really see anything blocking here. It looks really good overall. Just a few nits and tips, plus maybe trying to simplify the parameter types as we discussed offline; however, as we also talked about, maybe `HashSet` here really is best in terms of doing something cheaply once in 99% of the cases as opposed to all-too frequent sorting required otherwise. Or maybe we don't even need to sort, but can't use `binary_search` then. But if the header count is relatively small, probably doesn't matter.

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/examples/checkpoint_store_basic.rs
LarryOsterman: FWIW, this was all cut&pasted from an example in the opentelemetry crate :).

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/examples/checkpoint_store_basic.rs
LarryOsterman: As we were discussing this, I had a similar thought. And it DOES make it a lot easier.

sdk/typespec/typespec_client_core/src/http/options/mod.rs
heaths: I opened #2956 to track the design-discussion. Still, please add a comment. :)

sdk/typespec/typespec_client_core/src/http/policies/logging.rs
LarryOsterman: FWIW, Go uses a `map`, as does javascript, .Net uses a list, still working on Python.

sdk/typespec/typespec_client_core/src/http/sanitizer.rs
LarryOsterman: Go and .Net use "REDACTED" if it matters.

LarryOsterman:
@heaths would you mind taking a peek.

sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
```diff
@@ -43,16 +43,16 @@ impl RequestInstrumentationPolicy {
     ///
     pub fn new(
         tracer: Option<Arc<dyn crate::tracing::Tracer>>,
-        log_options: &LoggingOptions,
+        logging_options: &LoggingOptions,
```
heaths, line 46: I wouldn't hold up this PR for it, but if you think there's enough reason to specify these as "logging options" - that there might be more options in the future - than I suggest making a single `RequestInstrumentationPolicyOptions` that contains `logging` ("options" is then redundant) to better avoid breaking changes in the future.

sdk/core/azure_core/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+- Added logging policy to log HTTP requests and responses in the pipeline. As a part of this change, sanitization support was added to places which log HTTP headers and Urls. The `azure_core::http::ClientOptions` has been enhanced with a `LoggingOptions` which allows a user/service client to specify headers or URL query parameters which should be allowed. Note that the sanitization feature is disabled if you build with the `debug` feature enabled.
```
heaths: Nit: (you even capitalize it further in the sentence)

```suggestion
- Added logging policy to log HTTP requests and responses in the pipeline. As a part of this change, sanitization support was added to places which log HTTP headers and URLs. The `azure_core::http::ClientOptions` has been enhanced with a `LoggingOptions` which allows a user/service client to specify headers or URL query parameters which should be allowed. Note that the sanitization feature is disabled if you build with the `debug` feature enabled.
```

sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/examples/checkpoint_store_basic.rs
```diff
@@ -4,6 +4,10 @@
 // cspell:ignore appender OTLP
 
 //! Basic example showing how to create and use a blob checkpoint store.
+//! This example demonstrates creating a `BlobCheckpointStore`, claiming ownership of a partition,
```
heaths, line 8: Nit: this becomes one long sentence otherwise

```suggestion
//!
//! This example demonstrates creating a `BlobCheckpointStore`, claiming ownership of a partition,
```

sdk/typespec/typespec_client_core/src/http/sanitizer.rs
```diff
@@ -4,44 +4,45 @@
...
-    .iter()
-    .copied()
-    .collect()
-});
-
-pub static DEFAULT_ALLOWED_QUERY_PARAMETERS: LazyLock<HashSet<&'static str>> =
-    LazyLock::new(|| ["api-version"].iter().copied().collect());
+use std::{borrow::Cow, collections::HashSet, sync::LazyLock};
+
+pub static DEFAULT_ALLOWED_HEADER_NAMES: LazyLock<HashSet<Cow<'static, str>>> =
```
heaths, line 10: Public APIs need to be commented.

heaths (approved):
Just nits. You can take them here and I can reapprove, or catch 'em in the next one.

sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
LarryOsterman: These policies are private, so while technically breaking, I'm not sure it matters.

heaths (approved):

PR #2958, opened 2025-08-29
Removed Azure specific headers from typespec_client_core.

LarryOsterman:


sdk/core/azure_core/src/http/pipeline.rs
```diff
@@ -50,7 +51,7 @@ impl Pipeline {
         per_call_policies: Vec<Arc<dyn Policy>>,
         per_try_policies: Vec<Arc<dyn Policy>>,
     ) -> Self {
-        let (core_client_options, options) = options.deconstruct();
+        let (core_client_options, mut options) = options.deconstruct();
```
RickWinter: If we mutate the options that would prevent you from configuring the options once and passing to multiple clients.

LarryOsterman: This is the local variable returned from the `options.deconstruct()` function, it's not the `options` passed in from the client.

In this case, we want to take the `typespec_client_core::ClientOptions` from the client and add in the 3 Azure "retry after" headers to the retry options - that requires mutating the options to add those headers. Fortunately, we're working with a copy of the `options` which was extracted from the `azure_core::ClientOptions` type.

Edit: Added coherency to the 2nd paragraph.

sdk/typespec/typespec_client_core/src/error/http_error.rs
```diff
@@ -24,7 +24,7 @@ impl HttpError {
     /// Create an error from an HTTP response.
     ///
     /// This does not check whether the response was successful and should only be used with unsuccessful responses.
-    pub async fn new(response: RawResponse) -> Self {
+    pub async fn new(response: RawResponse, error_header_name: Option<HeaderName>) -> Self {
```
heaths: 1. I don't like the long parameter name. Since "error" is redundant, just "header_name" is probably fine.
2. That said, I don't think a single one is sufficient. IIRC, some services - like Storage - have a list they go through. Should this maybe be a `&[&'static str]`? Would `'static` be sufficient here since we know the list we'll pass ahead of time?

sdk/core/azure_core/src/http/mod.rs
```diff
@@ -14,7 +14,7 @@ pub mod request;
 
 pub use models::*;
 pub use options::*;
-pub use pager::{ItemIterator, PageIterator, Pager};
+pub use pager::{ItemIterator, PageIterator, Pager, PagerResult, PagerState};
```
LarryOsterman: I'm not sure why this change is needed, without it, the `PagerResult` and `PagerState` fields are not defined in the `http` module.

sdk/typespec/typespec_client_core/src/error/http_error.rs
```diff
@@ -187,8 +187,12 @@ struct ErrorDetails {
 }
 
 impl ErrorDetails {
```
heaths, line 189: Ultimately, this whole structure needs to be lifted up to `azure_core`. It's literally defined in the Azure REST API Guidelines in https://aka.ms/azapi/guidelines. Instead, what we could have is our Azure error info - much like this (still needs some tweaks) but then in `typespec::Error` maybe we just need a generic HTTP error that holds a `RawResponse` and the HTTP `StatusCode`.

sdk/typespec/typespec_client_core/src/error/http_error.rs
heaths: I'd rename all the `error_header_name`. "error" just seems unnecessarily long.

sdk/core/azure_core/src/http/poller.rs
```diff
@@ -476,10 +479,15 @@ where
 }
 
 /// Get the retry duration from the operation response or [`PollerOptions`].
-pub fn get_retry_after(headers: &Headers, options: &PollerOptions) -> Option<Duration> {
+pub fn get_retry_after(
+    headers: &Headers,
+    retry_headers: &[(HeaderName, bool)],
```
heaths: See how pervasive this tuple is getting, should we just put the tool in the `HeaderName` definition we're sending across? This was what I was considering before and the only reason for keeping pre-defined headers. Effectively,

```rust
use std::borrow::Cow;

#[derive(Debug, Clone, Eq)]
struct HeaderName {
    name: Cow<'static, str>,
    standard: bool,
}

const RETRY_AFTER: HeaderName = HeaderName {
    name: Cow::Borrowed("retry-after"),
    standard: true,
};

impl HeaderName {
    pub fn is_standard(&self) -> bool {
        self.standard
    }
}

impl AsRef<str> for HeaderName {
    fn as_ref(&self) -> &str {
        self.name.as_ref()
    }
}

impl PartialEq for HeaderName {
    fn eq(&self, other: &HeaderName) -> bool {
        self.standard == other.standard && self.name.eq_ignore_ascii_case(other.name.as_ref())
    }
}

impl PartialEq<&str> for HeaderName {
    fn eq(&self, other: &&str) -> bool {
        self.name.eq_ignore_ascii_case(*other)
    }
}

#[test]
fn name_eq() {
    assert_eq!(RETRY_AFTER, "retry-after")
}

#[test]
fn eq() {
    assert_eq!(
        RETRY_AFTER,
        HeaderName {
            name: Cow::Borrowed("retry-after"),
            standard: true,
        },
    );
}

#[test]
fn is_standard() {
    assert!(RETRY_AFTER.is_standard());
}

#[test]
fn as_ref() {
    assert_eq!(RETRY_AFTER.as_ref(), "retry-after");
}
```

Could even put some helpers on there. This way we're not having to 1) pass this info all over, and 2) there's almost no risk of it being incorrect in some instances (none if we make sure we use the right headers).

heaths (changes requested):
I like the general pattern here, but there are a couple things I think we should reconsider. See my sample code. That was basically what I was thinking way back when I opened (or commented on?) the issue but never rose to the priority. This way, we're not having to pass around a disparate tuple. Treat it like metadata - almost like attributes on .NET types.

sdk/core/azure_core/src/http/mod.rs
LarryOsterman: Found the merge conflict that introduced this error and fixed it.

sdk/typespec/typespec_client_core/src/error/http_error.rs
LarryOsterman: FWIW, the current typespec_client_core logic has no extensibility support for additional headers.

I'm not opposed to the idea of extending this to multiple headers, but as things currently stand, there is no need for more than the single header.

Also, making this service extensible is a non-trivial modification to the current design - the current design intentionally isolates the HTTP headers associated with the operation to the `azure_core` layer and intentionally removes that from service interactions.

sdk/core/azure_core/src/http/poller.rs
LarryOsterman: The more that I think about it, the more I like the idea of embodying the "is a standard" into the `HeaderName`. Of course, that doesn't *exactly* match the behavior we want for the `Retry-After` headers, because we'd be burning in a requirement that "all non-standard retry-after headers express their parameter as a 32bit number of milliseconds", which is not really robust.

sdk/typespec/typespec_client_core/src/error/http_error.rs
heaths: Discussed today as part of #2962 and not blocking this PR.

sdk/typespec/typespec_client_core/src/error/http_error.rs
LarryOsterman: Followup: I'll add this extensibility as a part of the typespec_client_core::http::Error cleanup.

LarryOsterman:
@heaths if you get a chance, the PR has been updated to add the "is standard" field to `HeaderName`.

sdk/typespec/typespec_client_core/src/http/options/retry.rs
```diff
@@ -102,6 +102,15 @@ impl RetryOptions {
         }
     }
 
+    /// Defines a set of HTTP headers which, if present on a response,
+    /// indicate that the response should be retried after a delay.
+    /// The boolean indicates whether the header value is a number of seconds to wait.
+    /// If true, the header value is a header which conforms to the `Retry-After` HTTP header specification.
+    /// If false, the header value is a number of milliseconds to wait.
```
heaths: Nit (but could wait if nothing else): much like commit messages, the first sentence or two should be short. This is like the `<summary>` in .NET doc comments. A blank line, then details like .NET's `<remarks>` e.g.,

```suggestion
    /// Defines a set of HTTP headers which, if present on a response,
    /// indicate that the response should be retried after a delay.
    ///
    /// The boolean indicates whether the header value is a number of seconds to wait.
    /// If true, the header value is a header which conforms to the `Retry-After` HTTP header specification.
    /// If false, the header value is a number of milliseconds to wait.
```

sdk/typespec/typespec_client_core/src/http/options/retry.rs
```diff
@@ -56,52 +60,84 @@ impl Default for RetryMode {
 pub struct RetryOptions {
     /// The algorithm to use for calculating retry delays.
     mode: RetryMode,
+    retry_headers: RetryHeaders,
```
heaths: Why is this configurable by the user? Only the client should know or have to deal with that. We had talked about a `PipelineOptions` or something like that. E.g., Azure SDK for .NET doesn't expose this level of control over spec'd service behavior to customers. Everything in `ClientOptions` and `ClientMethodOptions` is for customers to override certain behaviors like retry timing, custom headers, etc.

sdk/typespec/typespec_client_core/src/http/options/retry.rs
```diff
@@ -56,52 +60,84 @@ impl Default for RetryMode {
...
+    /// The boolean indicates whether the header value is a number of seconds to wait.
+    /// If true, the header value is a header which conforms to the `Retry-After` HTTP header specification.
+    /// If false, the header value is a number of milliseconds to wait.
+    ///
+    /// # Arguments
+    /// * `headers` - A list of HTTP headers to check for retry information.
+    ///
+    pub fn with_retry_after_headers(
+        mut self,
+        headers: &[(HeaderName, bool)],
```
heaths: I thought you were going to bake the `bool` into `HeaderName` like my example code showed. We don't need to carry this tuple all over.

heaths (changes requested):
The `RetryHeaders` is not user-configurable and should not be in nor under `ClientOptions`. Only clients can customize that. It either needs to be a new `PipelineOptions` we talked about when setting up the `Pipeline`, or a `PipelineSendOptions` when calling `Pipeline::send()`.

Also, I provided some sample code for baking the `standard: bool` field into `HeaderName` that I think would be much better. When defining a header, it's either standard or its not. This tuple separates that data meaning it can be different in other instances. `RETRY_AFTER` will always be standard, so why do we also need to pass `true` in a tuple with it?

sdk/typespec/typespec_client_core/src/http/options/retry.rs
heaths: See <https://github.com/Azure/azure-sdk-for-rust/pull/2958#discussion_r2311660406>. As I said, if we only define standard headers in `typespec_client_core`, then that `bool` field can be `pub(crate)`. If and when customers ever need to define their own standard headers, we can add a construct like `HeaderName::new_standard(name: impl Into<Cow<'static, str>, standard: bool>` or something, but not until then. There's probably no need.

sdk/core/azure_core/src/http/poller.rs
heaths: How is adding the bool as a disjoint tuple member any different when it comes to solving that problem? If you want to describe behaviors at the time of declaration, you should pass a `'static Fn` or something.

sdk/typespec/typespec_client_core/src/http/options/retry.rs
LarryOsterman: It's configurable by the user because it's configurable by the service and I cannot distinguish between the service and the user.

We can make it be a new PipelineOptions feature however.

sdk/typespec/typespec_client_core/src/http/options/retry.rs
LarryOsterman: OH @#@#$. I have an entire chunk of changes that never made it to the PR for reasons I have no idea why.

sdk/typespec/typespec_client_core/src/http/options/retry.rs
heaths: Retry headers the service sends aren't configurable by users, just as HTTP errors aren't...in general, anyway. IIRC, .NET - and perhaps some other languages - allow users to customize which status codes might result in an error or not, but that seems like a door we should install until it's asked for. It'd be very odd for a service 1) to return a success status code it didn't document, and 2) for a user to say, "no, treat that as a success". I certainly agree - and have offered evidence of this for KV's pseudo-LROs in other languages - convenience layers might want to do that but we didn't expose that to users.

sdk/typespec/typespec_client_core/src/http/policies/retry/mod.rs
```diff
@@ -302,10 +304,7 @@ mod test {
         let retry_after = get_retry_after(
             &headers,
             datetime_now,
-            &[
-                (HeaderName::from_static("x-ms-retry-after"), false),
-                (RETRY_AFTER, true),
-            ],
+            &[HeaderName::from_static("x-ms-retry-after"), RETRY_AFTER],
```
heaths: Nit (if there are other changes): may as well just define this as a `const` in the `tests` module instead of repeating it, but this is fine too.

sdk/typespec/typespec_client_core/src/lib.rs
```diff
@@ -4,7 +4,7 @@
 #![doc = include_str!("../README.md")]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
-#[macro_use]
+    #[macro_use]
```
heaths: Nit: extraneous whitespace. If our formatter did this, that's...peculiar.

heaths (commented):
That fixes one of the two big issues, but I don't think service-defined behavior about what retry headers the service sends or error codes should be user-configurable. I don't know of any other language that does that. We can discuss offline, though.

sdk/typespec/typespec_client_core/src/lib.rs
LarryOsterman: Actually it's going to fail the CI because of it. No idea where it came from though.

LarryOsterman:
Ok, *finally* got it working @heaths

PR #2968, opened 2025-09-04
Updated to current Typespec emitter and regenerated SDK clients

LarryOsterman:


sdk/keyvault/azure_security_keyvault_certificates/src/generated/clients/certificate_client.rs
```diff
@@ -351,8 +381,20 @@ impl CertificateClient {
         certificate_version: &str,
         options: Option<CertificateClientGetCertificateOptions<'_>>,
     ) -> Result<Response<Certificate>> {
+        if certificate_name.is_empty() {
+            return Err(azure_core::Error::message(
+                azure_core::error::ErrorKind::Other,
+                "parameter certificate_name cannot be empty",
+            ));
+        }
+        if certificate_version.is_empty() {
```
jhendrixMSFT: This looks problematic as I assume that `certificate_version` should be optional.

I suspect that we need to update to the latest commit hash for the KV tsp which I believe has fixed this.

jhendrixMSFT: The tsp was fixed in https://github.com/Azure/azure-rest-api-specs/commit/ece4912654b9473138738bdef53dcac613e50af5 (not the latest commit but this is the bare minimum required).

heaths: I think we should stick with that version. Looking at the subsequent commits to the ancestor "keyvault" directory, a new beta version was added and if we're looking to GA sooner than later, we'll probably beat KV's new version's GA date.

heaths: Fixed in the TypeSpec

LarryOsterman:
Ok, finally I think I have a working set of tests.

sdk/keyvault/azure_security_keyvault_certificates/README.md
```diff
@@ -209,7 +209,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     )?;
 
     let certificate = client
-        .get_certificate("certificate-name", "certificate-version", None)
+        .get_certificate("certificate-name", None)
```
jhendrixMSFT: Nit, I think the doc comment on line 195 needs to be updated to reflect the new public surface area (probably applies elsewhere).

jhendrixMSFT (approved):

sdk/keyvault/azure_security_keyvault_certificates/README.md
LarryOsterman: Looking - that shouldn't compile, right?

LarryOsterman: Ok, I have the fix for this one (the others in the certificates readme aren't a problem). But it'll need a re-approval.

LarryOsterman:
@jhendrixMSFT would you mind taking a peek again? Your "nit" was actually relevant, so I've gone through all the readme samples restoring the missing functionality.

PR #2969, opened 2025-09-04
Add `PatchDocument::condition` and builder method for it

jprochazk:


sdk/cosmos/azure_data_cosmos/src/models/patch_operations.rs
```diff
@@ -39,10 +40,20 @@ use serde::{Deserialize, Serialize};
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub condition: Option<String>,
     pub operations: Vec<PatchOperation>,
 }
 
 impl PatchDocument {
+    /// Adds a condition, which determines whether or not the patch should be applied.
+    ///
+    /// The value is an SQL-like filter predicate as a string. For example, `from c where c.taskNum = 3`.
+    pub fn with_condition(mut self, condition: impl Into<String>) -> Self {
```
analogrelay: Our pattern for now has been to avoid `impl Into<T>` parameters unless there's a significant ergonomic reason. They hide allocations, and while `"foo".to_string()` is clunky, the allocation is obvious.

Having said that, I think this could be a `Cow<'static, str>` so that users can use static string literals without incurring an allocation. In _that_ case, an `impl Into<Cow<'static, str>>` is more appropriate, IMO, because it's less obvious to a user how to create a `Cow<'static, str>` (and there's no hidden allocation for a string literal)

Could you update this to `Cow<'static, str>`? (And the matching field would become `Option<Cow<'static, str>>`.

jprochazk: > Could you update this to Cow<'static, str>? (And the matching field would become Option<Cow<'static, str>>.

Done

analogrelay:
The build failures are a known issue in `main` that is fixed, I believe. I'll re-trigger the build.

heaths:
Rebase again. There was a bad PR that went into `eng/common` that broke test-proxy on Windows, it seems.

Issue #1709, opened 2024-07-23
Make sure all public APIs are documented


heaths:
Make sure all public APIs are documented. Specifically, this came up in #1708 but I know there are other public APIs that are missing either `///` docs or `#[doc = ""]` docs (which work with the current macros).

RickWinter:
This is something we could add to APIView. 

Feature: APIView flags public methods that are missing doc comments

heaths:
Sure, but you might want to open a bug in that repo or here to track. This issue was meant more for "we should go add docs where they are missing" as a concerted effort. I've been doing so as I'm in or around functions lacking docs or improving docs as needed.

PR #2972, opened 2025-09-05
Require doccoments in typespec and typespec_client_core packages

LarryOsterman:
Partial fix for #1709

### DIFF
--- a/sdk/typespec/src/error/mod.rs
+++ b/sdk/typespec/src/error/mod.rs
@@ -22,7 +22,9 @@ pub enum ErrorKind {
     /// An HTTP status code that was not expected.
     #[cfg(feature = "http")]
     HttpResponse {
+        /// The HTTP status code that was returned.
         status: StatusCode,
+        /// An optional error code returned by the service.
         error_code: Option<String>,
     },
     /// An error performing IO.
@@ -170,6 +172,7 @@ impl Error {
         }
     }
 
+    /// If this error is an HTTP response error, return the associated status code.
     #[cfg(feature = "http")]
     pub fn http_status(&self) -> Option<StatusCode> {
         match &self.kind() {
--- a/sdk/typespec/typespec_client_core/src/base64.rs
+++ b/sdk/typespec/typespec_client_core/src/base64.rs
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+// cspell:ignore Hdvcmxk
+
 //! Base64 encoding and decoding functions.
 
 use base64::{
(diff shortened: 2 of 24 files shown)


sdk/typespec/typespec_client_core/src/base64.rs
```diff
@@ -25,27 +27,87 @@ const URL_SAFE: GeneralPurpose = GeneralPurpose::new(
...
+/// assert_eq!(encoded, "SGVsbG8sIHdvcmxkIQ");
+/// ```
 pub fn encode_url_safe<T>(input: T) -> String
 where
     T: AsRef<[u8]>,
 {
     URL_SAFE.encode(input)
 }
 
+/// Decode a base64url encoded string using the URL safe base64 decoding scheme. (base64url)
```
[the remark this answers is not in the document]
LarryOsterman: base64url is the industry standard term for the "URL safe base64 scheme", and is included for people who won't necessarily relate the two.

sdk/typespec/src/error/mod.rs
```diff
@@ -22,7 +22,9 @@ pub enum ErrorKind {
     /// An HTTP status code that was not expected.
     #[cfg(feature = "http")]
     HttpResponse {
+        /// The HTTP status code that was returned.
         status: StatusCode,
+        /// An optional error code returned by the service.
```
heaths, line 27: Nit: Personally, I always found this a little confusing across languages: why is an error *code* a string? What if we included an example e.g.,


```suggestion
        /// An optional error code returned by the service e.g., "BadParameter".
```

Or such?

sdk/typespec/typespec_client_core/src/http/request/options/macros.rs
```diff
@@ -23,6 +23,7 @@ macro_rules! request_header {
     };
     ($(#[$outer:meta])* $name:ident, $header:ident, $(($variant:ident, $value:expr)), *) => {
         $crate::request_option!($(#[$outer])* $name);
+        #[allow(missing_docs)]
```
heaths, line 26: Do any instances of this macro - I can't imagine it's too many and we should probably get rid of it anyway - actually need this attribute?

sdk/typespec/typespec_client_core/src/tracing/mod.rs
```diff
@@ -41,6 +41,9 @@ pub trait TracerProvider: Send + Sync + Debug {
     ) -> Arc<dyn Tracer>;
 }
 
+/// The Tracer trait is responsible for creating spans and managing the active span in distributed tracing.
```
heaths, line 44: Nit: you did this in other changes:


```suggestion
/// The `Tracer` trait is responsible for creating spans and managing the active span in distributed tracing.
```

sdk/typespec/typespec_client_core/src/base64.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+// cspell:ignore Hdvcmxk
```
heaths, line 4: Nit: might be nice if you just copy the whole value. It wasn't obvious - though I suspected - where this substring was:

```suggestion
// cspell:ignore SGVsbG8sIHdvcmxkIQ
```

sdk/typespec/src/error/mod.rs
LarryOsterman: I like the idea. I'll fix it in my PR (which is changing this code).

sdk/typespec/typespec_client_core/src/tracing/mod.rs
LarryOsterman: Fixed in my current PR.

sdk/typespec/typespec_client_core/src/http/request/options/macros.rs
LarryOsterman: Yes - ContentType in mod.rs triggers it. Apparently the doccomment in the macro does not flow to that statement.

heaths: It will if that uses `#[doc("...")]` or whatever that syntax is. Maybe fix it in your other PR so we're not just blanket allowing missing docs.

PR #2973, opened 2025-09-05
[Storage] `find_blobs_by_tags` for BlobServiceClient

vincenttran-msft:
`.tsp`: https://github.com/Azure/azure-rest-api-specs/pull/37182

- Initial implementation of `find_blobs_by_tags` for `BlobServiceClient`
- Refactors `create_test_blob` to accept Options bag.


sdk/storage/azure_storage_blob/tests/blob_service_client.rs
```diff
@@ -168,3 +172,68 @@ async fn test_get_account_info(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
+    create_test_blob(
+        &blob2_client,
+        Some(RequestContent::from(blob2_data.into())),
+        Some(blob2_upload_options),
+    )
+    .await?;
+
+    // Find "hello world" blob by its tag {"foo": "bar"}
+    let response = service_client
+        .find_blobs_by_tags("\"foo\"='bar'", None)
```
vincenttran-msft, line 221: This is how it is used in [Python](https://github.com/Azure/azure-sdk-for-python/blob/ed9ceb0eb9c359665f803f78fdad838cb044d236/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_service_client.py#L481) wrt putting the responsibility of formatting the `filter_expression` correctly.

Is this an area that we would want to improve usability i.e. accept a `HashMap<String, String>` and perform this formatting for them? Another possible solution is to not augment the parameter, and instead offer a `format_X` similar to [format_page_range](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/storage/azure_storage_blob/src/parsers.rs#L17)?

In the docstring for `range` parameter:
>`range` - The range of bytes to clear. See [`format_page_range()`](crate::format_page_range) for help with the expected String format.

So we could do the same treatment for `filter_expression` 😄 
Once we land on an approach we are happy with, will duplicate to ContainerClient

jalauzon-msft: Let's keep the parameter as a string. A helper could be cool but I would consider it optional.

sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -159,6 +160,31 @@ impl BlobContainerClient {
...
+    /// * `filter_expression` - The expression to find blobs whose tags matches the specified condition.
+    ///   eg.
+    /// ```text
+    /// "\"yourtagname\"='firsttag' and \"yourtagname2\"='secondtag'"
+    /// ```
+    ///   To specify a container, eg.
+    /// ```text
+    /// "@container='containerName' and \"Name\"='C'"
+    /// ```
+    ///   See [`format_filter_expression()`](crate::format_filter_expression) for help with the expected String format.
```
heaths: Nit:
```suggestion
    /// See [`format_filter_expression()`](crate::format_filter_expression) for help with the expected String format.
```

sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
```diff
@@ -98,6 +99,31 @@ impl BlobServiceClient {
...
+    /// * `filter_expression` - The expression to find blobs whose tags matches the specified condition.
+    ///   eg.
+    /// ```text
+    /// "\"yourtagname\"='firsttag' and \"yourtagname2\"='secondtag'"
+    /// ```
+    ///   To specify a container, eg.
+    /// ```text
+    /// "@container='containerName' and \"Name\"='C'"
+    /// ```
+    ///   See [`format_filter_expression()`](crate::format_filter_expression) for help with the expected String format.
```
heaths: Here too. I'll not repeat for brevity, but please check other members.

sdk/storage/azure_storage_blob/src/parsers.rs
```diff
@@ -37,3 +37,26 @@ pub fn format_page_range(offset: u64, length: u64) -> Result<String, Error> {
...
+    if tags.is_empty() {
+        return Err(Error::new(
+            ErrorKind::InvalidInput,
+            "Tags HashMap cannot be empty.".to_string(),
+        ));
+    }
+
+    let format_expression: Vec<String> = tags
+        .iter()
+        .map(|(key, value)| format!("\"{}\"='{}'", key, value))
```
heaths, line 58: Double quotes for keys and single quotes for values? You can't just use single quotes for both? I mean, it's not a problem - just stood out as odd.

sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
vincenttran-msft: Good catch, will fix these instances.

sdk/storage/azure_storage_blob/src/parsers.rs
vincenttran-msft: Yeah this is abit of a weird one, but unfortunately aligns with the service behavior. REST [docs ](https://learn.microsoft.com/en-us/rest/api/storageservices/find-blobs-by-tags?tabs=microsoft-entra-id#constructing-a-search-expression)states
>Tag values must be enclosed in single quotation marks.

While testing with single quotes `''` for key gives:
>"<Error><Code>InvalidQueryParameterValue</Code><Message>Error parsing query at or near character position 1: unexpected &apos;foo&apos;\n",
i.e.
```
https://sanitized.blob.core.windows.net/?comp=blobs&where=%22foo%22%3D%27bar%27
(working, "" and '')
 
https://sanitized.blob.core.windows.net/?comp=blobs&where=%27foo%27%3D%27bar%27
(error, '' and '')
```

jalauzon-msft (approved):

PR #2974, opened 2025-09-07
experiment to add support for wasm32-wasip2 using Spin

cataggar:
This is an experiment to make targeting wasi32-wasip2 possible. I think the exciting breakthrough will be when wasi32-wasip3 lands with composable concurrency. https://www.fermyon.com/blog/looking-ahead-to-wasip3


heaths (changes requested):
I appreciate this is a draft, but I did want to note that we want to avoid adding more "parity features" or even specific implementations to the Azure SDK crates for the same reason: we provide the means to implement `HttpClient` and `AsyncRuntime` like you did, but don't want to get into the business of providing countless implementations in the crates for the currently most popular crates. This was the purpose of #2933.

PR #2977, opened 2025-09-08
Refactor conditions for Prerelease and Release stages

hallipr:
- Allow arbitrary branches to be manually built and live tested without being released
- Only allow releases from manually queued runs of `main` or `hotfix/*`

### DIFF
--- a/eng/pipelines/templates/stages/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/stages/archetype-sdk-client.yml
@@ -140,8 +140,8 @@ extends:
     # The Prerelease and Release stages are conditioned on:
     # 1. Internal pipeline
     # 2. Only release from main or hotfix/*
-    # 2. Not weekly build
-    # 3. Manual trigger or force IncludeRelease
+    # 3. Not weekly build
+    # 4. Manual trigger or force IncludeRelease
     - ${{ if eq(variables['System.TeamProject'], 'internal') }}:
       - ${{ if or(eq(variables['Build.SourceBranch'], 'refs/heads/main'), startsWith(variables['Build.SourceBranch'], 'refs/heads/hotfix/')) }}:
         - ${{ if not(contains(variables['Build.DefinitionName'], 'weekly')) }}:


eng/pipelines/templates/stages/archetype-sdk-client.yml
```diff
@@ -138,16 +138,18 @@ extends:
...
-              ServiceDirectory: ${{ parameters.ServiceDirectory }}
-              Artifacts: ${{ parameters.Artifacts }}
-              TestPipeline: ${{ eq(parameters.ServiceDirectory, 'template') }}
-              PipelineArtifactName: packages
+    # 1. Internal pipeline
+    # 2. Only release from main or hotfix/*
+    # 3. Not weekly build
+    # 4. Manual trigger or force IncludeRelease
+    - ${{ if eq(variables['System.TeamProject'], 'internal') }}:
+      - ${{ if or(eq(variables['Build.SourceBranch'], 'refs/heads/main'), startsWith(variables['Build.SourceBranch'], 'refs/heads/hotfix/')) }}:
```
heaths, line 146: Aren't branch specifics usually in `ci.yml`? This feels much more strict.

hallipr: We have a lot of rules here that should probably just go into the `IncludeRelease` parameter.  As is, the PR template passes "IncludeRelease: false" and all of the ci.ymls let it default to true.  For some reason, we ignore IncludeRelease if the build was manually queued, which is what I want to avoid here.... 

Can you think of any reason to release a build that wasn't manually queued?

heaths: No, but I also wonder why we don't support the usual `release/` prefix. Or even `feature/`, though we'd want to make sure those don't publish to crates.io and that would require more work than is in scope here or necessary for GA.

weshaggard: @hallipr as we chatted I agree with Heath this is too strict. We should let you trigger from any branch. We still have the approval environment that will block anything from publishing until it is ready.

PR #2981, opened 2025-09-09
[Storage] `get/set_access_policy` for ContainerClient

vincenttran-msft:
Currently having issue with generated code.

We need to get a `RequestContent<Vec<SignedIdentifier>, XmlFormat>` but there is no impl to yield this currently. Need a TryFrom<> implementation in [model_impl.rs](https://github.com/vincenttran-msft/azure-sdk-for-rust/blob/c9ca5537ffa80b05cf5e695541412f9896ac04ff/sdk/storage/azure_storage_blob/src/generated/models/models_impl.rs)


sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -249,4 +253,32 @@ impl BlobContainerClient {
...
+    /// Sets the permissions for the specified container. The permissions indicate whether blobs in a
+    /// container may be accessed publicly.
+    ///
+    /// # Arguments
+    ///
+    /// * `container_acl` - The access control list for the container.
+    /// * `options` - Optional configuration for the request.
+    pub async fn set_access_policy(
+        &self,
+        container_acl: RequestContent<Vec<SignedIdentifier>, XmlFormat>,
```
vincenttran-msft: Will probably end up changing this parameter to `Vec<SignedIdentifier>` and we will call the `into()` internally on behalf of the cx before handing to generated layer.

sdk/storage/azure_storage_blob/src/models/extensions.rs
```diff
@@ -89,3 +95,18 @@ impl From<HashMap<String, String>> for BlobTags {
...
+    items: Vec<SignedIdentifier>,
+}
+
+// Converts a `Vec<SignedIdentifier>` into `RequestContent<Vec<SignedIdentifier>, XmlFormat>`.
+pub(crate) fn format_signed_identifiers(
+    value: Vec<SignedIdentifier>,
+) -> Result<RequestContent<Vec<SignedIdentifier>, XmlFormat>, azure_core::Error> {
+    let wrapper = SignedIdentifiersWrapper { items: value };
+    Ok(to_xml_with_root("SignedIdentifiers", &wrapper)?.into())
+}
```
vincenttran-msft, line 132: Source of truth (python asset [example](https://github.com/Azure/azure-sdk-assets/blob/python/storage/azure-storage-blob_729418a751/python/sdk/storage/azure-storage-blob/tests/recordings/test_blob_access_conditions.pyTestStorageBlobAccessConditionstest_set_container_acl_with_if_modified.json#L44), [testcase](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/tests/test_blob_access_conditions.py#L210)):
Resulting XML
```xml

<SignedIdentifiers>
    <SignedIdentifier>
        <Id>testid</Id>
        <AccessPolicy>
            <Start>2022-07-13T23:44:52Z</Start>
            <Expiry>2022-07-14T00:44:52Z</Expiry>
            <Permission>r</Permission>
        </AccessPolicy>
    </SignedIdentifier>
</SignedIdentifiers>
```

So for Rust I needed to do the following modifications:
- Reorder `Id` to come before `AccessPolicy` in XML as required by service
- Introduce a wrapper type, as I was having difficulties having `SignedIdentifier` appearing in the XML when using it directly.
- - Copilot's answer: 
- > serde treats each element in the vector as a direct child of the root element, ignoring the #[serde(rename = "SignedIdentifier")] attribute on the struct.

But even then this is still giving me an error. I can only think that the start and expiry times are not in the correct format.

But even after resolving that, is there anyway we think we can more elegantly resolve this? (i.e. maybe some way to indicate to serde to _not_ ignore the attribute)

vincenttran-msft: Follow-up, it was the time format (despite Copilot thinking that both RFCs were valid)

vincenttran-msft:
Taking a look at the raw logs [here](https://dev.azure.com/azure-sdk/29ec6040-b234-4e31-b139-33dc4287b756/_apis/build/builds/5340389/logs/263), seems like we are running into a recording timestamp mismatch error due to the `expiry`/`start` being specified in `AccessPolicy`:
>... 
{\"Message\":\"Unable to find a record for the request PUT https://Sanitized.blob.core.windows.net/containerpnruvlqr?comp=acl\\u0026restype=container\\nHeader differences:\\nBody differences:\\nRequest and record bodies do not match at index 111:\\n     request: \\u0022-09-15T20:45:03.593916235Z\\u003C/Ex\\u0022\\n     record:  \\u0022-09-15T20:38:38.4405229Z\\u003C/Expi\\u0022\\n
...

In Python we use a [get_datetime_variable](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/tests/test_blob_access_conditions.py#L147C30-L147C51) helper so that those values get taken down as variables in the [recording](https://github.com/Azure/azure-sdk-assets/blob/python/storage/azure-storage-blob_25230e96c7/python/sdk/storage/azure-storage-blob/tests/recordings/test_blob_access_conditions.pyTestStorageBlobAccessConditionstest_set_container_metadata_with_if_modified.json#L97-L98):
```json
"Variables": {
    "if_modified": "2022-07-13T23:29:50.814080"
  }
```
@heaths Heath, is this something that would be considered in Core in case any other crates will be running into this timestamp mismatch when testing?

heaths:
And we already have support to read and store recorded variables if that's easier. You can pass whatever you want to that or even wrap it for ease if you want.

vincenttran-msft:
> And we already have support to read and store recorded variables if that's easier. You can pass whatever you want to that or even wrap it for ease if you want.

How do you store recorded variables in Rust? i.e. in the above code snippet for Python,
`variables.setdefault(name, dt.isoformat())` 
where `name` is the name of the environment variable we want to store, and the second parameter is the value we want to store.

In the current recording configuration in Rust (to my knowledge) `recording.var()` is looking for existing exported environment variables, and if found will add it to the variables in the recording. But is there anyway to directly set the variables that are exported from the recording?

vincenttran-msft:
Closing in favor of the new one based off current `main`.

PR #2982, opened 2025-09-09
Reworked error handling to use new check_status function

LarryOsterman:
Fixed layering violation because `HttpError` was defined in typespec_client_core but was an azure_core construct.
Simplified service client error management.


sdk/core/azure_core/src/error/error_response.rs
```diff
@@ -0,0 +1,309 @@
...
+}
+
+impl ErrorResponse {
+    /// The error details.
+    pub fn error(&self) -> Option<&ErrorDetail> {
+        self.error.as_ref()
+    }
+}
+
+impl TryFrom<Error> for ErrorResponse {
```
LarryOsterman, line 37: Note that Cosmos's tests depend on having the ability to convert an `Error` to a `ErrorResponse`, and because the `RawResponse` is wrapped in an `Option`, we can only borrow, which means that calling `.json()` on the `body()` fails because `.json()` consumes self.

The `try_from()` call neatly avoids that by directly calling `serde_json` like the other logic here does, thus encapsulating the serde logic from the rest of the SDK.

sdk/typespec/typespec_client_core/src/http/policies/retry/mod.rs
```diff
@@ -116,9 +112,11 @@ pub trait RetryPolicy: std::fmt::Debug + Send + Sync {
     /// Determine how long before the next retry should be attempted.
     fn sleep_duration(&self, retry_count: u32) -> Duration;
     /// A Future that will wait until the request can be retried.
-    /// `error` is the [`Error`] value the led to a retry attempt.
+    ///
+    /// # Arguments
+    /// `retry_count` is the number of times the request has been retried.
     /// `retry_after` is the duration to wait before retrying, if provided by the server response.
-    async fn wait(&self, _error: &Error, retry_count: u32, retry_after: Option<Duration>) {
+    async fn wait(&self, retry_count: u32, retry_after: Option<Duration>) {
```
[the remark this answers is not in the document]
LarryOsterman: @copilot Can you explain why the documentation and signature are different?

sdk/typespec/typespec_client_core/src/http/policies/retry/mod.rs
```diff
@@ -116,9 +112,11 @@ pub trait RetryPolicy: std::fmt::Debug + Send + Sync {
     /// Determine how long before the next retry should be attempted.
     fn sleep_duration(&self, retry_count: u32) -> Duration;
     /// A Future that will wait until the request can be retried.
-    /// `error` is the [`Error`] value the led to a retry attempt.
+    ///
+    /// # Arguments
+    /// `retry_count` is the number of times the request has been retried.
     /// `retry_after` is the duration to wait before retrying, if provided by the server response.
```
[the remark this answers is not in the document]
heaths: @LarryOsterman though, don't we normally bullet the arguments? If no other changes, I wouldn't do it now, though. We can clean it up later; otherwise, once rendered all the arguments are just going to be one long run-on sentence.

sdk/typespec/typespec_client_core/src/http/request/options/macros.rs
```diff
@@ -23,9 +23,9 @@ macro_rules! request_header {
     };
     ($(#[$outer:meta])* $name:ident, $header:ident, $(($variant:ident, $value:expr)), *) => {
         $crate::request_option!($(#[$outer])* $name);
-        #[allow(missing_docs)]
         impl $name {
             $(
+                #[allow(missing_docs)]
```
heaths, line 28: Nit: Is this still necessary after you added `#[doc = ""]` to `ContentType`?

sdk/core/azure_core/src/error/mod.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+pub use typespec_client_core::error::{Error, ErrorKind, Result, ResultExt};
+mod error_response;
+
+pub mod http {
```
heaths: 1. Why do we need a separate `http` module?
2. If you keep it, docs.

sdk/core/azure_core/src/error/mod.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+pub use typespec_client_core::error::{Error, ErrorKind, Result, ResultExt};
```
heaths: I would import everything so it's always a subset:

```suggestion
pub use typespec_client_core::error::*;
```

sdk/core/azure_core/src/error/error_response.rs
```diff
@@ -0,0 +1,386 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct ErrorResponse {
+    error: Option<ErrorDetail>,
+}
+
+impl ErrorResponse {
+    /// The error details.
+    pub fn error(&self) -> Option<&ErrorDetail> {
+        self.error.as_ref()
+    }
```
heaths: Why make callers call accessors? We don't for other models. If they take ownership or a mutable ref and want to update it, that's on them. I don't see why we should complicate the DX.

sdk/core/azure_core/src/error/error_response.rs
```diff
@@ -0,0 +1,386 @@
...
+}
+
+impl TryFrom<Error> for ErrorResponse {
+    type Error = Error;
+
+    fn try_from(value: Error) -> Result<Self, Self::Error> {
+        match value.kind() {
+            ErrorKind::HttpResponse { raw_response, .. } => {
+                let error_response: Option<ErrorResponse> = raw_response.as_ref().map(|raw| {
+                    serde_json::from_slice(raw.body()).expect("expected an HttpResponse error")
```
heaths: You're already in a `TryFrom`. Don't panic. You can use a `match` yourself instead of calling `map` so that it all works out.

sdk/core/azure_core/src/error/error_response.rs
```diff
@@ -0,0 +1,386 @@
...
+    /// Additional properties that may be returned with the error.
+    #[serde(flatten)]
+    additional_properties: HashMap<String, serde_json::Value>,
+}
+
+impl ErrorDetail {
+    /// The error code.
+    ///
+    /// This is a machine-readable error code, typically derived from the x-ms-error-code response header.
+    pub fn code(&self) -> Option<&str> {
```
heaths: Just declare fields as public. Read-only accessors aren't necessary.

sdk/core/azure_core/src/error/error_response.rs
```diff
@@ -0,0 +1,386 @@
...
+    }
+    let internal_response =
+        serde_json::de::from_slice::<ErrorResponseInternal>(response.body()).map_err(Error::from);
+
+    let internal_response = match internal_response {
+        Ok(r) => r,
+        Err(_) => {
+            // If we can't parse the body, return a generic error with the status code and body
+            let code = response.headers().get_optional_str(&ERROR_CODE);
+            let error_kind = ErrorKind::http_response(status, code.map(str::to_owned));
```
heaths: I thought we wanted to default to `code.as_str()` or whatever, thus eventually making `error_code` non-`Option`al?

sdk/core/azure_core/src/http/mod.rs
```diff
@@ -26,5 +26,6 @@ pub use typespec_client_core::http::{
     Method, NoFormat, StatusCode, Url,
 };
 
+pub use crate::error::http::check_success;
```
heaths: You're exposing this API twice. Is that what we really want? We could export the models from `crate::error` (or even the `http` submodule as you have it), but maybe the function gets exported here if that's the general use case for calling it?

sdk/typespec/typespec_client_core/src/http/request/options/macros.rs
LarryOsterman: Yes - the problem is not the ContentType - the problem is the inner CONTENT_TYPE type being defined - I spent an hour or two trying to figure out how to add attributes to the inner CONTENT_TYPE field but failed. I can revisit later though.

sdk/core/azure_core/src/error/error_response.rs
LarryOsterman: Fair enough.

sdk/core/azure_core/src/error/error_response.rs
LarryOsterman: Good point, makes sense to me.

sdk/core/azure_core/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 - Added `TryFrom<T> for RequestContent<T, JsonFormat>` for JSON primitives.
 - Added support for WASM to the `async_runtime` module.
 - Added logging policy to log HTTP requests and responses in the pipeline. As a part of this change, sanitization support was added to places which log HTTP headers and URLs. The `azure_core::http::ClientOptions` has been enhanced with a `LoggingOptions` which allows a user/service client to specify headers or URL query parameters which should be allowed. Note that the sanitization feature is disabled if you build with the `debug` feature enabled.
+- Added support for a new `azure_core::error::http::ErrorResponse` structure which describes an error according to the Azure REST API guidelines.
+- Added a new `azure_core::error::check_success(BufResponse)` function to convert a buffered response to an `ErrorKind::HttpResponse`.
```
heaths:
```suggestion
- Added a new `azure_core::error::http::check_success(BufResponse)` function to convert a buffered response to an `ErrorKind::HttpResponse`.
```

heaths (approved):

sdk/core/azure_core/src/http/mod.rs
LarryOsterman: I have to expose it from `error_response.rs` because otherwise I cannot expose it from `http`. And `check_success` is an HTTP related operation (even if `ErrorResponse` is a model type and thus transport agnostic..)

sdk/core/azure_core/CHANGELOG.md
```diff
@@ -10,7 +10,7 @@
 - Added support for WASM to the `async_runtime` module.
 - Added logging policy to log HTTP requests and responses in the pipeline. As a part of this change, sanitization support was added to places which log HTTP headers and URLs. The `azure_core::http::ClientOptions` has been enhanced with a `LoggingOptions` which allows a user/service client to specify headers or URL query parameters which should be allowed. Note that the sanitization feature is disabled if you build with the `debug` feature enabled.
 - Added support for a new `azure_core::error::http::ErrorResponse` structure which describes an error according to the Azure REST API guidelines.
-- Added a new `azure_core::error::check_success(BufResponse)` function to convert a buffered response to an `ErrorKind::HttpResponse`.
+- Added a new `azure_core::http::check_success(BufResponse)` function to convert a buffered response to an ErrorKind::HttpResponse`.
```
heaths:
```suggestion
- Added a new `azure_core::http::check_success(BufResponse)` function to convert a buffered response to an `ErrorKind::HttpResponse`.
```

heaths (approved):

heaths (approved):

Issue #1995, opened 2025-01-15
Client methods should be able to return custom errors


heaths:
Currently, our [Rust guidelines](https://azure.github.io/azure-sdk/rust_introduction.html#rust-errors) dictate returning an `azure_core::Result<T>` from our client methods, which is re-exported from `typespec` which aliases this as `std::result::Result<T, typespec::error::Error>` where `typespec::error::Error` is modeled after our [REST API guidelines](https://aka.ms/azapi/guidelines).

For TypeSpec-emitted clients that do no return an Azure error, we should support returning custom errors even if basic i.e., just an HTTP status code.

One idea is that we do this unconditionally if emitting a client in an "unbranded" mode.

Another idea would be to just emit code that returns a `std::result::Result<T, E>` where `E` is the modeled error type IIF `E` isn't modeled after an Azure  error response. Note that `typespec::error::Error` does not mention "Azure" - it's a generic definition of Azure error specs.

heaths:
I was thinking about this more and talking with @lmazuel and I think returning a `typespec::error::Error` is actually fine IIF we hae a way to supply a list of success and error codes (any code not in the success list should probably be an error since it's not spec'd) OR - and I'm leaning toward this if not too bad on perf - a handler function that emitters and SDK devs can supply that categorize responses based primarily (perhaps only) on HTTP status codes. We can provide a default for Azure clients when emitting code.

This is basically what .NET is doing with their DPG methods - or at least was when I was on that team. The code generator folded and emitter delegates to categories whether a 200 or 202 or whatever was expected, and everything else was an error. And for Azure services, this is probably best - and makes the case for a single handler method - because the [REST API Guidelines](https://aka.ms/azapi/guidelines#rest-error-use-default-response) state,

> ⚠️YOU SHOULD NOT document specific error status codes in your OpenAPI/Swagger spec unless the "default" response cannot properly describe the specific error response (e.g. body schema is different).

heaths:
This one has already been decided. We just need to update guidelines.

PR #2983, opened 2025-09-09
Add `RawResponse` to `ErrorKind::HttpResponse`

### LINKED ISSUE #2495: Attach raw stream to `ErrorKind::HttpResponse`
Once we buffer the entire response in memory for all error responses, we should make sure to attach that to `ErrorKind::HttpResponse` with, perhaps, some deserialization helper functions patterned after `Response<T>`.

### LINKED ISSUE #2962: Refactor error types for generic and Azure-specific models
This is a roll-up of the following issues:

- [ ] #1995 
- [ ] #2495 
- [ ] #2725 

The problems here are all because of layering:

- `typespec::Error` has very basic HTTP information available - no body or headers, only status code message (one parsed from a standard [Azure standard error response](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#rest-error-response-body-structure) currently) - because we don't want to (yet) move all the HTTP plumbing down the dependency stack.
- `typespec_client_core::Error` does actually assume an Azure standard error response currently, and we have to pass everything down from `azure_core` to support some of that, like in #2958.

Ultimately, to satisfy all the linked issues above, I think we need to stop re-exporting `typespec::Error` and start defining higher-order errors as we move up the dependency stack e.g.,

- `azure_core::Error` (actually, `azure_core::error::Error`, but I'll use the crate-specific re-export henceforth for brevity) *could* possibly be `typespec_client_core::Error` re-exported, but we'd parse the details ourselves using `pub(crate)` Azure standard error response types. The message[^1] we'd try to parse from the Azure standard error response.

  These would be `pub(crate)` initially because most other Azure SDK languages don't expose them either, and I don't think they facilitate custom errors like in #2495 or wrapping as in #2725. If we find a reason to expose them later, we can do that. We can't internalize them later after GA. 
- `typespec_client_core::Error` (again, actually `typespec_client_core::error::Error`) would define an `HttpError` that would contain the status code, message[^1], sanitized headers, and the raw body - basically: a sanitized version of `RawResponse` and maybe indeed we use `RawResponse`. We could clone so we could sanitize headers or even use it as is and rely on the `SafeDebug` derivation and let devs enable the `debug` feature if they want), and the status and message we'd just duplicate.
- `typespec::Error` (once more, `typespec::error::Error`) would either not define an `HttpError`, or would but just with the status code and message[^1]. No raw response. I do suspect that if we ever did introduce a `typespec_server_core` we might move some of the HTTP response types down the dependency stack, but we'll cross that bridge if/when we get there. We can always move things down the stack, but not up.
- Crates further up the stack like `azure_messaging_eventhubs` or even `azure_storage_blob` if needed (though, probably `azure_storage_common`) could then define their own error type if needed. For EH and SB, it's especially useful - and @LarryOsterman already did something to this effect for EH - because AMQP errors are 1) significantly different from HTTP errors and 2) require more granularity because many more specific errors are actionable - and in different ways - than with HTTP. We've already relaxed the Rust guidelines to allow for this, but this would codify a standard coding model for doing it.
- Errors are convertible when possible:
  - All errors further down the stack are convertible (`From`) to their "ancestor" crates up the stack e.g., `typespec::Error` -> `typespec_client_core::Error` -> `azure_core::Error` -> `azure_messaging_eventhubs::Error`
  - All errors up the stack are possibly convertible (`TryFrom<Error = ErrorKind::DataConversion?>`?) to errors down the stack. Most `ErrorKind` variants will exist in all crates.
  - `ErrorKind` in each crate must contain all the same error kinds as crates down the stack - as you grow up the stack, they are supersets.
  - All crates that define an `Error` must then define their own `Result<T, Error>` as well.

See <https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=109eea27999695a2394459901845a6c5> for a prototype (source: <https://gist.github.com/heaths/109eea27999695a2394459901845a6c5>);

[^1]: The message in `azure_core` we'd try to parse from an Azure standard error response. Failing that, we'd fall back to the functionality we'd use in `typespec_client_core` where it's probably just a friendly description of the status code, as is common for HTTP libraries anyway and what our other Azure SDK languages do for a fallback, IIRC.

heaths:
Renames `RawResponse` to `BufResponse` - inspired by `std::io::BufReader`, and adds a new `RawResponse` where the body is fully read and stored in `Bytes`.
Refactored `Headers` and other types necessary to add `RawResponse` to `ErrorKind::HttpResponse`.

Fixes #2962
Fixes #2495
Relates to #2725
Relates to #1995


sdk/keyvault/azure_security_keyvault_certificates/src/generated/clients/certificate_client.rs
```diff
@@ -725,7 +725,7 @@ impl CertificateClient {
             let ctx = options.method_options.context.clone();
             let pipeline = pipeline.clone();
             async move {
-                let rsp: RawResponse = pipeline.send(&ctx, &mut request).await?;
+                let rsp: BufResponse = pipeline.send(&ctx, &mut request).await?;
```
LarryOsterman, line 728: Silly question: Why can't the type of `rsp` be inferred? Why does it need to be explicit here?

LarryOsterman (approved):

sdk/keyvault/azure_security_keyvault_certificates/src/generated/clients/certificate_client.rs
heaths: Never noticed before. I'll try to remember to fix that in the emitter when I make changes for this PR and  your PR #2982.

Issue #2439, opened 2025-04-07
Support cloud configuration


christothes:
# Problem statement

Currently, the Tables library has support only for the Azure Public cloud.  Using the client in other clouds requires workarounds for the hardcoded authorization scope, such as a custom `TokenCredential` implementation, adding friction to the developer experience.

## Scope

- Allow callers to pass an audience value as part of the options to specify the runtime cloud.  
- Use the audience from the options to derive the appropriate authorization scope to use.  

## Success criteria

- Callers can specify the cloud environment that they are running in when creating the Tables client types.
- The Tables clients successfully adjust the authorization scope based on the selected audience and can perform operations in sovereign clouds.
- Tests needed to validate the new functionality have been added and pass consistently.
- Existing tests continue to pass reliably.

## Prior art

- [BlobAudience](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs/src/BlobClientOptions.cs#L394)
- [AppConfigurationAudience](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/appconfiguration/Azure.Data.AppConfiguration/src/AppConfigurationAudience.cs)
- [AzureOpenAIAudience](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/openai/Azure.AI.OpenAI/src/Custom/AzureOpenAIAudience.cs)
- [TextAnalyticsAudience](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsAudience.cs)

heaths:
@chlowell is this something we should do first in Identity, or have a design all up for any type of client?

chlowell:
I think we should first design the feature for clients in general, so we end up with a consistent configuration API and avoid throwaway work along the way.

heaths:
We discussed today and agreed to something like:

* https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore@v1.18.0/policy#ClientOptions
* https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore@v1.18.0/cloud#Configuration

This is additive, so not release blocking. We'd define this only in `azure_core::ClientOptions` and not in `typespec_client_core::ClientOptions` since it's not relevant to unbranded clients e.g.,

```rust
pub struct ClientOptions {
  // ...

[...]

PR #2986, opened 2025-09-11
Credentials use azure_core Pipelines

### LINKED ISSUE #2436: Credentials should use core Pipelines
Today they use HttpClients directly. One blocker for making this change is that the core RetryPolicy returns an HttpError when it receives a non-retriable, non-2xx response (#2435), and that type expects a generic Azure error shape that doesn't match Entra ID errors. That makes it difficult for a credential using the policy to return a clear error message like "Invalid secret".

chlowell:
This entails significant supporting changes:

- removed the mostly vestigial TokenCredentialOptions
    - ClientOptions now configures a credential's HttpClient
    - authority_host now comes from credential options; this is an interim replacement pending #2439
    - tests which used this to mock env vars now do so via a test-only option
- Added TokenRequestOptions.ClientMethodOptions (and the necessary lifetime parameter) so get_token can propagate the caller's Context
- Added an optional ClientMethodOptions argument to `ClientAssertion::secret()`
    - this could instead be an optional Context but ClientMethodOptions seems more future-proof

Closes #2436

### DIFF
--- a/sdk/core/azure_core/CHANGELOG.md
+++ b/sdk/core/azure_core/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - Added `RawResponse` to `ErrorKind::HttpResponse` that contains the HTTP status code, headers, and complete error response body.
 - Added `RequestContent::from_slice()`.
-- Added `TokenRequestOptions.ClientMethodOptions`.
+- Added `TokenRequestOptions { method_options: ClientMethodOptions }`.
 - Added `TryFrom<T> for RequestContent<T, JsonFormat>` for JSON primitives.
 - Added support for WASM to the `async_runtime` module.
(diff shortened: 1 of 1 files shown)


sdk/core/azure_core/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Added `RawResponse` to `ErrorKind::HttpResponse` that contains the HTTP status code, headers, and complete error response body.
 - Added `RequestContent::from_slice()`.
+- Added `TokenRequestOptions.ClientMethodOptions`.
```
heaths: Not formatted like rust code like we do elsewhere. Maybe:

```suggestion
- Added `TokenRequestOptions { method_options: ClientMethodOptions }`.
```

PR #2988, opened 2025-09-11
Update emitter and prepare for core release

heaths:


eng/emitter-package.json
```diff
@@ -1,17 +1,17 @@
...
+    "@azure-tools/typespec-client-generator-core": "0.60.0",
+    "@typespec/compiler": "1.4",
+    "@typespec/http": "1.4",
+    "@typespec/openapi": "1.4",
+    "@typespec/rest": "0.74",
+    "@typespec/versioning": "0.74",
+    "@typespec/xml": "0.74"
   }
-}
+}
```
RickWinter, line 17: Missing new line
```suggestion
}

```

heaths: Not critical. Next edit should get it. I'll make sure `hx` is doing that consistently. It should've, so not sure why it didn't.

PR #2990, opened 2025-09-11
Enabled doc comments in azure_core and azure_identity packages

LarryOsterman:
Also:

Removed a significant number of unused HTTP headers from azure_core.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -173,7 +173,7 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "azure_core"
-version = "0.28.0"
+version = "0.29.0"
 dependencies = [
  "async-lock",
  "async-trait",
@@ -205,7 +205,7 @@ dependencies = [
 
 [[package]]
 name = "azure_core_amqp"
-version = "0.7.0"
+version = "0.8.0"
 dependencies = [
  "async-trait",
  "azure_core",
@@ -226,7 +226,7 @@ dependencies = [
 
 [[package]]
 name = "azure_core_macros"
-version = "0.2.0"
+version = "0.3.0"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -239,7 +239,7 @@ dependencies = [
 
 [[package]]
 name = "azure_core_opentelemetry"
-version = "0.2.0"
+version = "0.3.0"
 dependencies = [
  "azure_core",
  "azure_core_test",
@@ -319,7 +319,7 @@ dependencies = [
 
 [[package]]
 name = "azure_identity"
-version = "0.28.0"
+version = "0.29.0"
 dependencies = [
  "async-lock",
  "async-trait",
@@ -343,7 +343,7 @@ dependencies = [
 
 [[package]]
 name = "azure_messaging_eventhubs"
-version = "0.7.0"
+version = "0.8.0"
 dependencies = [
  "async-lock",
  "async-stream",
@@ -366,7 +366,7 @@ dependencies = [
 
 [[package]]
 name = "azure_messaging_eventhubs_checkpointstore_blob"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "async-trait",
  "azure_core",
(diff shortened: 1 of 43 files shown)


sdk/core/azure_core/src/http/headers.rs
```diff
@@ -2,54 +2,19 @@
 // Licensed under the MIT License.
 
 //! HTTP headers.
+#![allow(missing_docs)]
```
heaths: Given this PR was all about adding missing docs, shouldn't the headers below just add docs? Could be as simple as, say,

```rust
/// The x-ms-client-request-id header.
pub const CLIENT_REQUEST_ID: HeaderName = HeaderName::from_static("x-ms-client-request-id");
```

An LLM could probably add these pretty quickly with just a single pattern like this.

sdk/core/azure_core/src/lib.rs
```diff
@@ -5,6 +5,7 @@
 #![deny(missing_debug_implementations, nonstandard_style)]
 #![doc = include_str!("../README.md")]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![warn(missing_docs)]
```
heaths, line 8: If there are more besides the header, I recommend `expect` instead of `warn`. That way there's an error about an unreferenced lint and we can clean this up to avoid any future violations.

sdk/core/azure_core_opentelemetry/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 
 #![doc = include_str!("../README.md")]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![warn(missing_docs)]
```
heaths, line 6: Use `expect` here as well.

sdk/core/azure_core/src/lib.rs
LarryOsterman: ?? `expect` reports that an expectation is not met, with no diagnostic about *why* the expectation is not met.

How does `expect` produce a better outcome than `warn` will?

LarryOsterman: Or are you suggesting that I change `allow(missing_docs)` to `expect(missing_docs)`? That might make some sense because it's making it clear that the absence of documentation is intentional. But I think it's overkill.

sdk/core/azure_core/src/http/headers.rs
LarryOsterman: I originally suppressed the documentation because it seemed pointless to add, but after seeing the developer experience with and without and you're 100% right.

Fixing.

sdk/core/azure_core_opentelemetry/src/lib.rs
LarryOsterman: I'm still confused about this suggestion. `#[expect(missing_docs)]` seems to say "there should be missing documents in this package", but that seems to be exactly counter to the intent of this PR, which is to enforce that everything in the PR is documented.

I know that if I make this change, clippy starts complaining with:
```
warning: this lint expectation is unfulfilled
 --> sdk\core\azure_core\src\lib.rs:8:11
  |
8 | #![expect(missing_docs)]
  |           ^^^^^^^^^^^^
  |
  = note: `#[warn(unfulfilled_lint_expectations)]` on by default

```
when there is no missing documentation, so I'm puzzled about this request.

sdk/core/azure_core/src/lib.rs
LarryOsterman: Ok, chased down the documentation for [`#[expect]`](https://rust-lang.github.io/rfcs/2383-lint-reasons.html#expect-lint-attribute) - and based on that documentation, it seems like `expect` is the exact opposite of `warn` - `expect` generates a warning if there are NO instances of the specific type. It's intended use case is when you would normally have an `allow` attribute, not when you'd have a `warn` attribute.

heaths: Yes, I had my wires crossed. `expect` should be used in lieu of `allow` when you have known violations. Sorry. 🥺

heaths (approved):

PR #3001, opened 2025-09-15
Added checkpoint store to packages

LarryOsterman:


sdk/eventhubs/ci.yml
```diff
@@ -17,3 +17,5 @@ extends:
     Artifacts:
     - name: azure_messaging_eventhubs
       safeName: AzureMessagingEventhubs
+    - name: azure_messaging_eventhubs_checkpointstore_blob
+      safeName: AzureMessagingEventhubsBlobCheckpointStore
```
heaths: You transposed "blob" and "checkpointstore". Pedantically, you're also inconsistent on casing since "Eventhubs" only starts with uppercase, but "CheckpointStore" has an uppercase "S" as well.

heaths (approved):

heaths (approved):

PR #3010, opened 2025-09-16
[Storage] `exists()` for BlobClient and BlobContainerClient

vincenttran-msft:
`.tsp:` https://github.com/Azure/azure-rest-api-specs/pull/37542

- Implements `exists()` for `BlobClient` and `BlobContainerClient`
- Introduces the `StorageErrorCode` enum from generated code changes.

### DIFF
--- a/sdk/storage/azure_storage_blob/src/clients/blob_client.rs
+++ b/sdk/storage/azure_storage_blob/src/clients/blob_client.rs
@@ -356,7 +356,7 @@ impl BlobClient {
         self.client.get_account_info(options).await
     }
 
-    /// Returns `true` if a blob exists, and returns `false` otherwise.
+    /// Returns `true` if the blob exists, and returns `false` otherwise.
     ///
     /// # Arguments
     ///
--- a/sdk/storage/azure_storage_blob/src/clients/blob_client.rs
+++ b/sdk/storage/azure_storage_blob/src/clients/blob_client.rs
@@ -17,7 +17,7 @@ use crate::{
         BlobClientReleaseLeaseOptions, BlobClientRenewLeaseOptions, BlobClientSetMetadataOptions,
         BlobClientSetPropertiesOptions, BlobClientSetTagsOptions, BlobClientSetTierOptions,
         BlobTags, BlockBlobClientCommitBlockListOptions, BlockBlobClientUploadOptions, BlockList,
-        BlockListType, BlockLookupList,
+        BlockListType, BlockLookupList, StorageErrorCode,
     },
     pipeline::StorageHeadersPolicy,
     AppendBlobClient, BlobClientOptions, BlockBlobClient, PageBlobClient,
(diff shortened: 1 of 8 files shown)


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -355,4 +355,9 @@ impl BlobClient {
     ) -> Result<Response<BlobClientGetAccountInfoResult, NoFormat>> {
         self.client.get_account_info(options).await
     }
+
+    /// Returns `true` if the blob exists, and returns `false` otherwise.
+    pub async fn exists(&self) -> bool {
+        self.get_properties(None).await.is_ok()
```
jalauzon-msft: What does `is_ok()` check? I would assume it just checks for a "success"? I don't think that is exactly how we want to handle this API. It should only return false if its explicitly a 404 `BlobNotFound`. In other error cases, we still want to error out.

vincenttran-msft: As discussed offline, lets match behavior of previous SDKs
`BlobNotFound/ContainerNotFound` should both be caught and return `false`, while all other errors in that codepath should be surfaced.

heaths: But there could still be a transport failure, so this should still return a `Result<bool>`. Would your .NET or Java client throw? Or JS for that matter? Same thing. `Result`s are Rust's form of "exceptions". `azure_storage_queue` is doing this correctly.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -356,8 +357,18 @@ impl BlobClient {
...
+    /// Returns `true` if the blob exists, `false` if the blob does not exist, and propagates all other errors.
+    pub async fn exists(&self) -> Result<bool> {
+        match self.client.get_properties(None).await {
+            Ok(_) => Ok(true),
+            Err(e) if e.http_status() == Some(StatusCode::NotFound) => match e.kind() {
+                ErrorKind::HttpResponse {
+                    error_code: Some(error_code),
+                    ..
+                } if error_code == "BlobNotFound" || error_code == "ContainerNotFound" => Ok(false),
+                _ => Ok(false),
```
heaths: You're returning false in both cases. What's the point of checking the inner error code?

heaths: Same below.

vincenttran-msft: Sorry, this was meant to be flipped back into draft because I was diving back into typespec to get the enum generating (I already corrected this logic error in the latest iteration)

heaths (approved):

jalauzon-msft (approved):

heaths (approved):

PR #3011, opened 2025-09-16
[Storage] Storage Error Handling WIP

vincenttran-msft:
Seem to be having some issues accessing the raw response (unexpectedly `None`). Output of test:
>running 1 test
Top of Test
Error debug print: 
```
Error {
  context: Message {
    kind: HttpResponse {
      status: NotFound,
      error_code: Some("BlobNotFound"),
      raw_response: None                            // <---- Unexpectedly None
    },
    message: "404"
  }
}
```
>Before Match
Bottom of Test
test test_storage_error_model ... ok

It seems like the `raw_response` is `None` here when I match it out from ErrorKind, but please advise as it is possible I am not accessing it properly.

Relevant Code Excerpt:
```rust
let error_kind = error_response.kind(); // Gives back ErrorKind

    // Match out raw_response from ErrorKind struct
    if let ErrorKind::HttpResponse {
        raw_response: Some(inner_raw_response),
        ..
    } = error_kind

    { // this body of code never executes because raw_response is None
         ... // do stuff using the inner_raw_response we just matched out
```

### DIFF
--- a/sdk/storage/azure_storage_blob/tests/blob_client.rs
+++ b/sdk/storage/azure_storage_blob/tests/blob_client.rs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 use azure_core::{
+    error::ErrorKind,
     http::{RequestContent, StatusCode},
     Bytes,
 };
@@ -16,6 +17,7 @@ use azure_storage_blob::models::{
 };
 
 use azure_storage_blob_test::{create_test_blob, get_blob_name, get_container_client};
+use core::error;
 use std::{collections::HashMap, error::Error, time::Duration};
 use tokio::time;
 
@@ -475,5 +477,43 @@ async fn test_get_account_info(ctx: TestContext) -> Result<(), Box<dyn Error>> {
     assert!(sku_name.is_some());
     assert_eq!(AccountKind::StorageV2, account_kind.unwrap());
 
+    Ok(())
+}
+
+#[recorded::test]
+async fn test_storage_error_model(ctx: TestContext) -> Result<(), Box<dyn Error>> {
+    // Recording Setup
+    println!("Top of Test");
+    let recording = ctx.recording();
+    let container_client = get_container_client(recording, true).await?;
+    let blob_client = container_client.blob_client(get_blob_name(recording));
+
+    // Act
+    let response = blob_client.get_properties(None).await;
+    let error_response = response.unwrap_err();
+    println!("Error debug print: {:?}", error_response);
+
+    let error_kind = error_response.kind();
+
+    println!("Before Match");
+    // Match out of the error_kind struct
+    if let ErrorKind::HttpResponse {
+        raw_response: Some(inner_raw_response),
+        ..
+    } = error_kind
+    {
+        println!("Inside of Match");
+        let status = inner_raw_response.status();
+        let body = inner_raw_response.body(); // Expected to be empty
+        let headers = inner_raw_response.headers();
+
+        println!("Status Code: {}", status);
+        for (key, value) in headers.iter() {
+            println!("Header: {} = {}", key.as_str(), value.as_str());
+        }
+        println!("Body: {:?}", body);
+    }
+
+    println!("Bottom of Test");
     Ok(())
 }


vincenttran-msft:
cc: @heaths Heath, any pointers here on how to proceed with accessing the raw response on an Error response. Thanks!

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -480,3 +482,33 @@ async fn test_get_account_info(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
+    if let ErrorKind::HttpResponse {
+        status,
+        error_code,
+        raw_response,
+        ..
+    } = error_kind
+    {
+        println!("Status code:{}", status);
+        println!("Error code:{}", error_code.clone().unwrap());
+        println!("Raw response:{:?}", raw_response.clone().unwrap());
```
heaths: Why are you cloning? You deconstructed them above, so just:

```suggestion
        println!("Status code: {}", status);
        println!("Error code: {}", error_code);
        println!("Raw response: {:#?}", raw_response);
```

sdk/storage/azure_storage_blob/src/models/extensions.rs
```diff
@@ -109,3 +109,61 @@ impl From<HashMap<String, String>> for BlobTags {
...
+        writeln!(f, "Http Status Code: {}", self.status_code)?;
+        writeln!(f, "Storage Error Code: {}", self.error_code)?;
+        writeln!(f, "Response Headers:")?;
+
+        for (name, value) in self.headers.iter() {
+            writeln!(f, "  \"{}\": \"{}\"", name.as_str(), value.as_str())?;
+        }
+
+        Ok(())
+    }
```
vincenttran-msft: Looking for some feedback here on how we want to implement the display. This is sort of what I came up with, and to re-iterate this is what currently comes out:

>Standard Error Message: 404: <?xml version="1.0" encoding="utf-8"?><Error><Code>BlobNotFound</Code><Message>The specified blob does not exist.
RequestId:f65c5ab0-601e-008d-677c-2ec0a1000000
Time:2025-09-26T00:29:56.3906662Z</Message></Error>
>
>Http Status Code: 404
Storage Error Code: BlobNotFound
Response Headers:
  "x-ms-client-request-id": "3b2dec8d-758b-45fd-adad-ba81c0d761b9"
  "content-length": "215"
  "server": "Microsoft-HTTPAPI/2.0"
  "x-ms-version": "2025-11-05"
  "content-type": "application/xml"
  "date": "Fri, 26 Sep 2025 00:29:56 GMT"
  "x-ms-request-id": "f65c5ab0-601e-008d-677c-2ec0a1000000"
  "x-ms-error-code": "BlobNotFound"
  "x-recording-mode": "record"

Ideally I want to eliminate `writeln!(f, "Standard Error Message: {}\n", self.message)?;` and I currently just have it there to show the current functionality of `azure_core::Error.to_string()`. 

I think we want to be able to extract just the message and surface that since we have all the other displayed elsewhere.
i.e. `The specified blob does not exist.`.

jaschrep-msft: 1. probably remove the quotations on headers as that's not generally how they're rendered.
2. would this end up in logs? if so we probably want to hook in whatever we're using to allow/redact specific headers

heaths: Headers should be sanitized. (Error bodies aren't supposed to contain PII.) @LarryOsterman added a `sanitize` function you should use.

I also recommend proper grammar e.g., "HTTP" instead of "Http", etc. Order of fields should make sense. In .NET, it was:

```
{HTTP status code} {HTTP error code}
{Headers}

{Error response body}
```

Basically, formatted close to an actual HTTP response like many devs are used to.

heaths: And before you go off and do this, we should discuss whether to change the `Display` implementation for `azure_core::Error`. Is this the only reason you're making `StorageError`? We already support formatting for display.

sdk/storage/azure_storage_blob/src/models/storage_error.rs
```diff
@@ -120,7 +119,7 @@ impl StorageError {
             additional_fields: HashMap<String, Value>,
         }
 
-        let xml_fields = azure_core::xml::read_xml::<StorageErrorXml>(xml_body)?;
+        let xml_fields = xml::from_xml::<_, StorageErrorXml>(xml_body)?;
```
heaths, line 122: Why are you passing the XML body and the `raw_response`? That's all available via `raw_response.body().xml()`. Also, if you're just getting data out of the raw body to format, borrow strings for memory efficiency: https://serde.rs/lifetimes.html#borrowing-data-in-a-derived-impl. We do this for Azure standard errors. Don't waste memory.

PR #3012, opened 2025-09-16
Allow package specific error codes in check_success

LarryOsterman:
Also:

- Added options to `Pipeline::send()`
- Propagate pipeline send options from `azure_core` to `typespec_client_core`

### DIFF
--- a/sdk/core/azure_core/src/http/pipeline.rs
+++ b/sdk/core/azure_core/src/http/pipeline.rs
@@ -15,7 +15,6 @@ use crate::{
 };
 use std::{
     any::{Any, TypeId},
-    ops::Deref,
     sync::Arc,
 };
 use typespec_client_core::http::{
@@ -51,13 +50,13 @@ pub struct PipelineSendOptions {
     pub skip_checks: bool,
 
     /// Options for `[check_success]`. If `skip_checks` is true, this field is ignored.
-    pub check_success_options: CheckSuccessOptions,
+    pub check_success: CheckSuccessOptions,
 }
 
 /// Internal structure used to pass options to the core pipeline.
 #[derive(Debug, Default)]
 struct CorePipelineSendOptions {
-    check_success_options: CheckSuccessOptions,
+    check_success: CheckSuccessOptions,
     skip_checks: bool,
 }
 
@@ -68,7 +67,7 @@ impl PipelineSendOptions {
         (
             CorePipelineSendOptions {
                 skip_checks: self.skip_checks,
-                check_success_options: self.check_success_options,
+                check_success: self.check_success,
             },
             None,
         )
@@ -172,7 +171,7 @@ impl Pipeline {
         let (core_send_options, send_options) = options.unwrap_or_default().deconstruct();
         let result = self.0.send(ctx, request, send_options).await?;
         if !core_send_options.skip_checks {
-            check_success(result, Some(core_send_options.check_success_options)).await
+            check_success(result, Some(core_send_options.check_success)).await
         } else {
             Ok(result)
         }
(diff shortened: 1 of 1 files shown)


sdk/core/azure_core/src/error/error_response.rs
```diff
@@ -108,19 +108,43 @@ struct ErrorDetailsInternal<'a> {
     message: Option<&'a str>,
 }
 
+/// Options for customizing the behavior of `check_success`.
+#[derive(Debug, Default)]
+pub struct CheckSuccessOptions {
+    /// A list of HTTP status codes that should be considered successful.
+    ///
+    /// If this list is empty, any 2xx status code is considered successful.
+    pub success_statuses: &'static [u16],
```
RickWinter: The error response pertains to http. Why don't we use the http::status struct?
https://docs.rs/http/latest/http/status/struct.StatusCode.html

LarryOsterman: Because Typespec doesn't know about `http::StatusCode` - it knows about `number`. And converting between the `number` and `http::StatusCode` is an expense we don't need to incur.

LarryOsterman: Btw, it appears that the Go emitter *does* map between the numeric codes (at the TCGC level) and the symbolic codes, but IMHO there's no real difference between the two, and it requires that the emitter know the symbolic names we use for status codes.

heaths: Nit: given we have a `StatusCode` where "status" is effectively the adjective, could we call these `success_codes`? Reads a bit better and rolls off the tongue better, IMO.

heaths: Also, are we sure these can always be - even thinking unbranded here - be a static slice? Might be safer as a `Vec<u16>`.

sdk/core/azure_core/src/http/pipeline.rs
```diff
@@ -40,6 +44,37 @@ use typespec_client_core::http::{
...
 pub struct Pipeline(http::Pipeline);
 
+/// Options for sending a request through the pipeline.
+#[derive(Debug, Default)]
+pub struct PipelineSendOptions {
+    /// If true, skip all checks including `[check_success]`.
+    pub skip_checks: bool,
+
+    /// Options for `[check_success]`. If `skip_checks` is true, this field is ignored.
+    pub check_success_options: CheckSuccessOptions,
```
heaths: Don't repeat "options". We don't anywhere else.

```suggestion
    pub check_success: CheckSuccessOptions,
```

heaths: Same below, even though internal.

sdk/core/azure_core/src/http/pipeline.rs
```diff
@@ -116,6 +151,32 @@ impl Pipeline {
...
+    /// # Arguments
+    /// * `ctx` - The context for the request.
+    /// * `request` - The request to send.
+    /// * `options` - Options for sending the request, including check success options. If none, `[check_success]` will not be called.
+    ///
+    /// # Returns
+    /// A [`http::BufResponse`] if the request was successful, or an `Error` if it failed.
+    /// If the response status code indicates an HTTP error, the function will attempt to parse the error response
+    /// body into an `ErrorResponse` and include it in the `Error`.
+    pub async fn send(
```
heaths, line 165: We could probably remove the `Deref` impl below now.

sdk/core/azure_core/src/error/error_response.rs
LarryOsterman: Done. `success_codes` it is.

PR #3016, opened 2025-09-17
Simplify ClientOptions

### LINKED ISSUE #2956: Should TransportOptions not be an Option type?
I wonder if we should change `TransportOptions` from an `Option`. It could instead default to a `Default` variant or something. These all take up the same amount of space either way. Treating some sub-client options as structs and others as `Option` seems weird. Makes DX easier.

_Originally posted by @heaths in https://github.com/Azure/azure-sdk-for-rust/pull/2955#discussion_r2308653847_

heaths:
Resolves #2956 by making nested options non-`Option`. Only leaf fields should be optional. An `Option<T>` allocates as much space as `T` regardless, so why deal with all the hassle - both for callers and callees - of dealing with `Option<T>` for intermediate members?


heaths:
@LarryOsterman I briefly looked over your `azure_messaging_eventhubs` create at your version of `ClientOptions` and seems like everything you defined was already a leaf node, but you might want to double check in case I missed one somewhere.

LarryOsterman (approved):

LarryOsterman (approved):

PR #3019, opened 2025-09-17
[Storage] Flattened Clients PoC

vincenttran-msft:
This PR fleshes out the following:

- Uses the current copy of flattened clients combined with current code so that APIs can work
- Working SAS PoC E2E
- Adds public access and also `from_blob_url` to `BlobClient`
- **Note:** Only `BlobClient`, and only `get_properties` currently works (need more manual edits to fix rest of clients/APIs)

Relevant files (I wouldn't look at the diff since most of it is commenting out code since it is so intertwined:
azure_storage_blob/tests/blob_client.rs
azure_storage_blob/src/clients/blob_client.rs
azure_storage_blob/src/generated/clients/blob_client.rs


vincenttran-msft:
Closing in favor of the new PR that works for all client types.

PR #3031, opened 2025-09-19
Support default values for recording vars

heaths:


sdk/core/azure_core_test/src/recording.rs
```diff
@@ -330,7 +330,7 @@ impl Recording {
 
         let value = self.env(key);
         if self.test_mode == TestMode::Live {
-            return value;
+            return value.or_else(|| Some(options?.default_value?.into_owned()));
```
vincenttran-msft: edit: nvm, this actually works as expected.

vincenttran-msft (approved):
LGTM!

vincenttran-msft (commented):
Actually hang on, pulled this code forward and testing against the issue..

vincenttran-msft:
Sounds good. I pulled this code forward and was not exactly able to get the desired results. I used Copilot to see if I could come up with at least a proof-of-concept of what we need, so this should hopefully help once we revisit this PR.

heaths:
@LarryOsterman I'm tempted to refactor this a little before merging because I realize we don't need to actually keep track of the original value *and* the sanitized value. But this works and, fortunately, we don't need to re-record anything because we shouldn't have expected to save any `""` in place of `None`.

LarryOsterman (approved):

heaths:
All the Copilot feedback was wrong. I left feedback as to why and resolved them.

sdk/keyvault/assets.json
heaths, line 1: I updated tests since I tested re-recording, playing back, and live test modes anyway. Sort of a "proof is in the pudding" move.

vincenttran-msft (approved):

PR #3033, opened 2025-09-19
Recording Utility PoC

vincenttran-msft:
Changes from current PR's authoring #3031:
For record mode, we need the following behavior:

- if the environment variable is `None` or unset, we should take the default value and record this to the recording variables.
- If the environment variable is set, but a default value is provided, take the default value. Otherwise, return the environment variable value, but also record whatever we choose.

So after running record, this was found in the asset recording file:
```xml
 "Variables": {
    "AZURE_STORAGE_ACCOUNT_NAME": "ruststoragedev",
    "expiry_time": "2025-09-19T20:43:36.7005382Z",
    "RandomSeed": "EaAhDPsDJRGKN/ddkEjWPQstrXFcy9h0Wur0Fy/8h8Q=",
    "start_time": "2025-09-19T20:43:26.7005515Z"
  }
```

And those values will be used in playback.


vincenttran-msft:
Closing as the corrected recording utility PR will merge shortly.

PR #3037, opened 2025-09-22
Rename azure_template* to azure_canary*

danieljurek:


weshaggard (commented):
Rename is fine with me. I will defer the sign-off to @heaths

sdk/canary/azure_canary/Cargo.toml
```diff
@@ -1,15 +1,15 @@
 [package]
-name = "azure_template"
+name = "azure_canary"
 version = "0.2.0"
```
heaths: Nit: could change this and the other crate to 0.1.0. They're only for testing, though, so no biggie.

sdk/canary/azure_canary/CHANGELOG.md
heaths, line 1: Similar nit: might think about either clearing this (initialized), or say that we renamed. Again, it's only for testing so no biggie.

Issue #2944, opened 2025-08-24
Closing `ProducerClient` or Droping it causes `IllegalConnectionState` error from `fe2o3_amqp::session::engine`


arkadiuszspiewak:
### Bug Title

Doing client.close().await causes fe2o3_ampq error

### Crate Name

azure_messaging_eventhubs

### Crate Version

0.6.0

### Description

We have an app that receives event once per hour or less [at least for now] and send responces in same frequency, so to avoid eventhubs timeouts we refresh ProducerClient periodically. But closing `ProducerClient` causes ampq error `IllegalConnectionState`. Using `drop(ProducerClient)` has same effect.
```
2025-08-24 00:37:30.308 | DEBUG: Closing old producer due to idle timeout. - reader::eventhub::producer
2025-08-24 00:37:30.308 | DEBUG: Close(None) - fe2o3_amqp::connection::engine
2025-08-24 00:37:30.415 | DEBUG: Stopped - fe2o3_amqp::connection::engine
2025-08-24 00:37:30.416 | ERROR: IllegalConnectionState - fe2o3_amqp::session::engine
2025-08-24 00:37:30.416 | DEBUG: Stopped - fe2o3_amqp::session::engine
2025-08-24 00:37:30.416 | DEBUG: Dropping Fe2o3AmqpConnection. - azure_core_amqp::fe2o3::connection
2025-08-24 00:37:30.416 | DEBUG: Dropping Fe2o3AmqpSession: OnceLock(Mutex { data: SessionHandle }). - azure_core_amqp::fe2o3::session
2025-08-24 00:37:30.416 | DEBUG: Opening new producer. - reader::eventhub::producer 
```

### Steps to Reproduce

1. Open Producer with:
```
    async fn open(&self) -> Result<ProducerClient, SyncerError> {
        let creds = self.creds.clone();
        let pc = ProducerClient::builder().open(&self.ns, &self.hub, creds).await?;
        Ok(pc)
    }
```
2. Send some events
3. Close Producer with
```
tracing::debug!("Closing old producer due to idle timeout.");
    match pc.close().await {
        Ok(_) => {}
        Err(e) => {
            tracing::error!("Error closing old producer: {}", e);
        }
    }
```

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/Azure/azure-sdk-for-rust/blob/main/CODE_OF_CONDUCT.md)
[...]

PR #3038, opened 2025-09-22
Cleaned up eventhubs disconnect operations

LarryOsterman:
Partial fix for #2944

The errors generated come from dropping a connection without first closing the connection (and sessions, etc on the connection).

This PR makes the connection close logic much more deterministic and close operations are now much more likely to succeed.

### DIFF
--- a/sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_processor_client.rs
+++ b/sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_processor_client.rs
@@ -30,7 +30,7 @@ impl BackgroundProcessor {
         // Start the background task to process events
         let mut task = self.background_task.lock().await;
         //        let mut task = task.as_mut();
-        let processor = Arc::clone(&self.processor);
+        let processor = self.processor.clone();
         task.replace(tokio::spawn(async move { processor.run().await }));
         Ok(())
     }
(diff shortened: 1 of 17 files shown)


sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_processor_client.rs
```diff
@@ -116,5 +114,13 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     background_processor.stop().await?;
     println!("Stopped background processor");
 
+    // Close the processor
+    if let Ok(processor) = Arc::try_unwrap(processor) {
+        processor.close().await?;
+        println!("Closed processor");
+    } else {
+        println!("Processor still has references, not closing.");
```
heaths, line 122: Nit/q: You use `trace` macros in some other examples. At least in this case, isn't this more of a trace or something you want the example executor to see?

sdk/eventhubs/azure_messaging_eventhubs/src/consumer/mod.rs
```diff
@@ -890,6 +902,12 @@ pub(crate) mod tests {
         )
         .await?;
 
+        if let Ok(consumer) = Arc::try_unwrap(consumer) {
+            consumer.close().await?;
+        } else {
+            panic!("Could not unwrap consumer");
```
heaths: If this could be shown to the end users, might want to clarify "unwrap" or just make it more end user friendly e.g., "Could not close message consumer" or something.

heaths: Same in other end user-facing places like just below.

sdk/eventhubs/azure_messaging_eventhubs/src/event_processor/partition_client.rs
```diff
@@ -195,3 +200,12 @@ impl PartitionClient {
         Ok(())
     }
 }
+
+impl Drop for PartitionClient {
+    fn drop(&mut self) {
+        debug!(
```
heaths: This was `trace` in other files. Is this one more important?

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_processor_client.rs
LarryOsterman: A goodquestion, need to think about it a bit.

sdk/eventhubs/azure_messaging_eventhubs/tests/eventhubs_producer.rs
```diff
@@ -81,6 +81,8 @@ async fn get_properties(ctx: TestContext) -> Result<(), Box<dyn Error>> {
     info!("Properties: {:?}", properties);
     assert_eq!(properties.name, eventhub);
 
+    client.close().await?;
```
heaths, line 84: What is the ramification of _not_ calling this? I'm a bit worried that `Drop` isn't enough, which is idiomatic to rust given it's typical RAII convention.

Even if all these tests run in parallel, since the consumer client is now owned and therefore unique, why should this matter?

PR #3039, opened 2025-09-22
Always attach RawResponse to ErrorKind::HttpResponse

heaths:


sdk/core/azure_core/src/error/error_response.rs
```diff
@@ -148,50 +148,65 @@ pub async fn check_success(
...
-        let error_kind = ErrorKind::http_response(status, code.map(str::to_owned));
+    if raw_response.body().is_empty() {
+        let error_code = raw_response
+            .headers()
+            .get_optional_str(&ERROR_CODE)
+            .map(str::to_owned);
+        let error_kind = ErrorKind::HttpResponse {
+            status,
+            error_code,
+            raw_response: None,
```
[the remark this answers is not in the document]
heaths: In this case, there's no body (or it's empty); however, I suppose we should to capture the headers at the very least.

sdk/typespec/src/http/response.rs
```diff
@@ -50,7 +50,7 @@ impl RawResponse {
 
     /// Deserialize the JSON stream into type `T`.
     #[cfg(feature = "json")]
-    pub async fn json<T>(self) -> crate::Result<T>
+    pub fn json<T>(self) -> crate::Result<T>
```
LarryOsterman, line 53: Change `json<T>` and `xml<T>` to borrow self rather than consuming self, since they're fully buffered?

heaths: As discussed offline, it'd be nice (and I tried), but the functions we call consume the body bytes so we'd have to change those or call something different, yada yada yada. I think it's probably fine as-is, though. When @analogrelay was working on all the `Model` changes, I remember discussing function prefaces and ownership. Basically, the idea was that by the time you're ready to deserialize the body, you've done everything you need with the rest of the response. I think the same is true here and, push comes to shove, getting the information first off the `RawResponse` before transferring ownership is cheap.

So, yeah, in concept I agree with you, but given the changes we'd have to make and how easy it is to work around the ownership constraints here, I think this is fine.

sdk/core/azure_core/src/error/error_response.rs
```diff
@@ -306,14 +321,15 @@ mod tests {
...
-        assert_eq!(
-            *kind,
+        assert!(matches!(
+            kind,
             ErrorKind::HttpResponse {
                 status: StatusCode::ImATeapot,
-                error_code: Some("testError".to_string()),
-                raw_response: None
+                error_code,
+                raw_response: Some(_),
```
LarryOsterman, line 329: Curious as to why `Some(_)` as opposed to `None`

heaths: This is a match. It is `Some`, but in this case I don't need to capture what's in `Some` because I wasn't going to match it. We have coverage elsewhere in this file already for that and this was mostly me getting the tests to pass.

PR #3050, opened 2025-09-25
Fix test in typespec_macros README

heaths:


sdk/typespec/typespec_macros/README.md
```diff
@@ -31,11 +31,23 @@ struct Credentials {
   pub username: String,
   pub password: String,
 };
-let credentials: Credentials = Credentials {
+
+let credentials = Credentials {
```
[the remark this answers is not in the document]
heaths: That's very wrong. That's not idiomatic to unnecessarily repeat the type.

sdk/typespec/typespec_macros/README.md
```diff
@@ -31,11 +31,23 @@ struct Credentials {
...
   username: "admin".into(),
   password: "hunter2".into(),
```
RickWinter, line 37: The code scanners haven't fired for a checked in pwd yet, but I wonder if they will?

RickWinter (approved):
The change looks goo

sdk/typespec/typespec_macros/README.md
heaths: But it's already masked. How do you even see it?! :wink:

We can cross that bridge if we get there. I would hope, for samples, they have a list of "allowed" and I'd be shocked - horrified, even - if "hunter2" wasn't on there.

PR #3051, opened 2025-09-25
Add support for event hub transport layer proxy

k-shlomi:
Adds SOCKS5 proxy support for Rust EventHubs client, enabling connections through corporate proxies and restricted network environments.

  Changes

  Core Implementation

  - azure_core_amqp/src/socks5.rs - New SOCKS5 connection module
    - SocksConnection struct with validate_proxy_url(), connect(), mask_credentials()
    - Support for socks5:// and socks5h:// protocols
    - Username/password authentication via URL
    - Automatic credential masking in logs
  - azure_core_amqp/src/connection.rs - AMQP integration
    - Modified open_with_stream() to detect SOCKS5 URLs in custom_endpoint
    - Zero breaking changes, backward compatible

  Dependencies

  - Added tokio-socks = "0.5" to azure_core_amqp/Cargo.toml

  Documentation & Examples

  - azure_core_amqp/README.md - Added SOCKS5 feature documentation
  - eventhubs/examples/eventhubs_socks5_proxy.rs - Complete usage example
  - Comprehensive rustdoc with examples and troubleshooting

  Testing

  - 11 unit tests for validation functions
  - 2 integration test files with live proxy support
  - All tests pass, maintain existing test coverage

  Usage
```rust
  let client = ProducerClient::builder()
      .with_custom_endpoint("socks5h://proxy.corp.com:1080")
      .open(host, eventhub, credential)
      .await?;
```

  Resolves enterprise customer requirements for proxy support in restricted network environments.

### DIFF
--- a/sdk/core/azure_core_amqp/Cargo.toml
+++ b/sdk/core/azure_core_amqp/Cargo.toml
@@ -29,6 +29,9 @@ serde.workspace = true
 serde_amqp = { workspace = true, optional = true }
 serde_bytes = { workspace = true, optional = true }
 tokio.workspace = true
+tokio-socks = "0.5"
+tokio-native-tls = "0.3"
+native-tls = "0.2"
 tracing.workspace = true
 typespec = { workspace = true, features = ["amqp"] }
 typespec_macros.workspace = true
--- a/sdk/core/azure_core_amqp/README.md
+++ b/sdk/core/azure_core_amqp/README.md
@@ -6,6 +6,39 @@ Azure AMQP crate for consumption of AMQP based packages in the Azure SDK for Rus
 
 This crate is part of a collection of crates: for more information please refer to [https://github.com/azure/azure-sdk-for-rust](https://github.com/azure/azure-sdk-for-rust).
 
+## SOCKS5 Proxy Support
+
+This crate supports SOCKS5 proxy connections for corporate environments. SOCKS5 support is enabled by configuring the `custom_endpoint` option with a SOCKS5 URL:
+
+```rust,no_run
+use azure_core::http::Url;
+use azure_core_amqp::AmqpConnectionOptions;
+
+# fn main() -> Result<(), Box<dyn std::error::Error>> {
+let mut options = AmqpConnectionOptions::default();
+options.custom_endpoint = Some(Url::parse("socks5h://proxy.corp.com:8080")?);
+# Ok(())
+# }
+```
+
+### Supported Protocols
+
+- **socks5://** - Standard SOCKS5 with local DNS resolution
+- **socks5h://** - SOCKS5 with proxy-side DNS resolution (recommended for corporate environments)
+
+### Authentication
+
+Username/password authentication is supported via the proxy URL:
+```text
+socks5://username:password@proxy.example.com:1080
+```
+
+All proxy credentials are automatically masked in log output for security.
+
+### Dependencies
+
+SOCKS5 support adds the `tokio-socks` dependency to the crate.
+
 ## Testing the AMQP Client
 
 The AMQP package is tested using the standard `cargo test` command line:
--- a/sdk/core/azure_core_amqp/src/socks5.rs
+++ b/sdk/core/azure_core_amqp/src/socks5.rs
@@ -1,0 +1,783 @@
+// Copyright (c) Microsoft Corporation. All Rights reserved
+// Licensed under the MIT license.
+
+//! SOCKS5 proxy support for Azure AMQP connections
+//!
+//! This module provides SOCKS5 proxy connectivity for Azure AMQP services through the
+//! [`SocksConnection`] helper. It enables Azure SDK for Rust to work in corporate
+//! environments that require proxy servers for outbound connections.
+//!
+//! # Overview
+//!
+//! The SOCKS5 implementation integrates seamlessly with the fe2o3-amqp library by providing
+//! a stream that can be passed to `Connection::open_with_stream()`. This allows AMQP
+//! connections to be established through SOCKS5 proxies without changing the higher-level
+//! Azure SDK APIs.
+//!
+//! # Supported Protocols
+//!
+//! - **SOCKS5** (`socks5://`) - Standard SOCKS5 with local DNS resolution
+//! - **SOCKS5h** (`socks5h://`) - SOCKS5 with proxy-side DNS resolution (recommended)
+//!
+//! # Integration with fe2o3-amqp
+//!
+//! The module works by intercepting the connection establishment in `Fe2o3AmqpConnection::open()`:
+//!
+//! ```text
+//! 1. Check if custom_endpoint uses socks5:// or socks5h:// scheme
+//! 2. If yes: Use SocksConnection::connect() to create proxy tunnel
+//! 3. Pass resulting stream to fe2o3_amqp::Connection::open_with_stream()
+//! 4. If no: Use standard fe2o3_amqp::Connection::open() (direct connection)
+//! ```
+//!
+//! This design ensures zero impact on direct connections while enabling proxy support
+//! when needed.
+//!
+//! # Configuration
+//!
+//! SOCKS5 proxy support is enabled by setting the `custom_endpoint` option in
+//! [`AmqpConnectionOptions`](crate::connection::AmqpConnectionOptions) to a SOCKS5 URL:
+//!
+//! ```rust,no_run
+//! use azure_core::http::Url;
+//! use azure_core_amqp::AmqpConnectionOptions;
+//!
+//! let mut options = AmqpConnectionOptions::default();
+//! options.custom_endpoint = Some(Url::parse("socks5h://proxy.corp.com:8080")?);
+//! // Connection will automatically use SOCKS5 proxy
+//! # Ok::<(), Box<dyn std::error::Error>>(())
+//! ```
+//!
+//! # Authentication
+//!
+//! Username/password authentication is supported via the proxy URL:
+//! ```text
+//! socks5://username:password@proxy.example.com:1080
+//! socks5h://user:pass@corporate-proxy.internal:8080
+//! ```
+//!
+//! Credentials are automatically masked in all log output for security.
+//!
+//! # Error Handling
+//!
+//! The module provides detailed error context for common failure scenarios:
+//! - Invalid proxy URL format
+//! - Network connectivity issues
+//! - Proxy authentication failures
+//! - Target service unreachability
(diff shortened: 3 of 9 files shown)


sdk/core/azure_core_amqp/Cargo.toml
```diff
@@ -29,6 +29,9 @@ serde.workspace = true
 serde_amqp = { workspace = true, optional = true }
 serde_bytes = { workspace = true, optional = true }
 tokio.workspace = true
+tokio-socks = "0.5"
+tokio-native-tls = "0.3"
+native-tls = "0.2"
```
heaths: These should be defined as part of an optional feature for anyone to enable who wants to use the protocol. Most customers likely won't need it and the increased build time required.

@LarryOsterman we'd also want to consider whether these should be declared directly herein or in the workspace `Cargo.toml`. The only net new dependency is `tokio-socks`, but I suspect that - at least for now - AMQP is the only protocol that would support it. The others are already in the `Cargo.lock` file and should just be in the workspace if they even need to be declared explicitly. `fe2o3-amqp@0.14.0` already has a dependency on `native-tls@0.2.14` and `tokio-native-tls@0.3.1` so they probably don't need to be in here at all.

sdk/core/azure_core_amqp/src/socks5.rs
```diff
@@ -0,0 +1,783 @@
+// Copyright (c) Microsoft Corporation. All Rights reserved
+// Licensed under the MIT license.
+
+//! SOCKS5 proxy support for Azure AMQP connections
```
heaths: This is great module documentation, but given its not `pub` when imported in `src/lib/rs`, no one is going to see it. Necessary information might be better served in the crate `README.md`.

sdk/core/azure_core_amqp/README.md
```diff
@@ -6,6 +6,39 @@ Azure AMQP crate for consumption of AMQP based packages in the Azure SDK for Rus
...
+
+This crate supports SOCKS5 proxy connections for corporate environments. SOCKS5 support is enabled by configuring the `custom_endpoint` option with a SOCKS5 URL:
+
+```rust,no_run
+use azure_core::http::Url;
+use azure_core_amqp::AmqpConnectionOptions;
+
+# fn main() -> Result<(), Box<dyn std::error::Error>> {
+let mut options = AmqpConnectionOptions::default();
+options.custom_endpoint = Some(Url::parse("socks5h://proxy.corp.com:8080")?);
```
heaths: We generally prefer the construction pattern:

```rust
let options = AmqpConnectionOptions {
  custom_endpoint = Some("socks5h://proxy.contoso.com:8080".parse()?),
  ..Default::default()
};
```

And use contoso.com or example.com.

heaths: Also means `azure_core::http::Url` doesn't need to be imported.

heaths (commented):
A few notes, but @LarryOsterman should definitely have a look.

sdk/core/azure_core_amqp/src/socks5.rs
```diff
@@ -479,11 +476,13 @@ impl SocksConnection {
...
-                Box::new(e)
-            ).with_context(format!("SOCKS5 connection failed: proxy={}, target={}",
-                Self::mask_credentials(proxy_url), target_url))
+            azure_core::Error::new(azure_core::error::ErrorKind::Other, Box::new(e)).with_context(
+                format!(
+                    "SOCKS5 connection failed: proxy={}, target={}",
+                    Self::mask_credentials(proxy_url),
+                    target_url
+                ),
+            )
```
heaths: Use `Error::with_error`:

```suggestion
            azure_core::Error::with_error(azure_core::error::ErrorKind::Other, e,
                format!(
                    "SOCKS5 connection failed: proxy={}, target={}",
                    Self::mask_credentials(proxy_url),
                    target_url
                ),
            )
```

(Something like that)

sdk/core/azure_core_amqp/src/socks5.rs
```diff
@@ -0,0 +1,792 @@
...
+    ///
+    /// If the URL cannot be parsed or modified, returns `"invalid_url"` as a
+    /// safe fallback to prevent any credential exposure.
+    ///
+    /// # Usage
+    ///
+    /// This function is automatically used in all logging statements within
+    /// the SOCKS5 implementation. Manual usage is typically not required
+    /// unless implementing custom logging or error handling.
+    pub(crate) fn mask_credentials(url: &Url) -> String {
```
LarryOsterman, line 491: You might want to consider using the azure_core extension method `sanitize` here - it implements our standardized sanitization logic. You would still need to redact the username and password since azure URLs are prohibited from using the URL username and password fields, but this adds an additional level of sanity checks on the URL.

Basically change the last line to `masked.sanitize(&[])`

sdk/eventhubs/azure_messaging_eventhubs/examples/eventhubs_socks5_proxy.rs
```diff
@@ -0,0 +1,168 @@
...
+    info!("Closed producer client connection");
+
+    Ok(())
+}
+
+/// Mask credentials in proxy URL for secure logging
+///
+/// This function replaces username and password in the proxy URL with asterisks
+/// to prevent credential exposure in log files.
+fn mask_proxy_credentials(url: &str) -> String {
```
LarryOsterman, line 108: Same comment about using the `sanitize()` extension for the rest of the URL.

sdk/eventhubs/azure_messaging_eventhubs/tests/eventhubs_socks5_messaging.rs
```diff
@@ -0,0 +1,206 @@
+// Copyright (c) Microsoft Corporation. All Rights reserved
+// Licensed under the MIT license.
+
+#![cfg(feature = "socks5")]
```
LarryOsterman: Our tests are built with `--all-features` - how do you ensure that these tests don't fail in our live tests CI pipeline?

heaths: I think these tests should move to examples...which you already have, so if the examples are sufficient, I say we remove these integration tests if we don't have any easy way to test them on devs' machines (we don't control nor dictate how they run tests) or hosted agents.

k-shlomi:
Hi @heaths and @LarryOsterman 
thank you for the review.
I believe I addressed all your comments. let me know if anything else is needed.

k-shlomi:
@heaths I believe I did fix your comments. can you point me out to comments I missed?

sdk/core/azure_core_amqp/Cargo.toml
```diff
@@ -29,8 +29,12 @@ serde.workspace = true
 serde_amqp = { workspace = true, optional = true }
 serde_bytes = { workspace = true, optional = true }
 tokio.workspace = true
+tokio-socks = { version = "0.5", optional = true }
+native-tls = { version = "0.2", optional = true }
+tokio-native-tls = { version = "0.3", optional = true }
 tracing.workspace = true
 typespec = { workspace = true, features = ["amqp"] }
+typespec_client_core = { workspace = true }
```
heaths: Everything in here is in `azure_core`. `azure_*` crates outside of core should not reference `typespec` crates.

sdk/core/azure_core_amqp/Cargo.toml
```diff
@@ -29,8 +29,12 @@ serde.workspace = true
 serde_amqp = { workspace = true, optional = true }
 serde_bytes = { workspace = true, optional = true }
 tokio.workspace = true
+tokio-socks = { version = "0.5", optional = true }
+native-tls = { version = "0.2", optional = true }
+tokio-native-tls = { version = "0.3", optional = true }
```
heaths: 2 of these don't need to be referenced directly as we already have dependencies. The other has to be defined in the workspace `/Cargo.toml`. We don't allow direct dependencies unless @LarryOsterman thinks is worth exempting a version dependency in our appropriate place. This is to avoid version mismatches throughout the workspace - especially import as the number of crates grows.

sdk/core/azure_core_amqp/Cargo.toml
```diff
@@ -49,6 +53,7 @@ fe2o3_amqp = [
   "serde_amqp",
   "serde_bytes",
 ]
+socks5 = ["dep:tokio-socks", "dep:native-tls", "dep:tokio-native-tls"]
```
heaths, line 55: You also need to add this to the docs.rs metadata table below or it won't be documented. We intentionally don't document the others as they are meant more for internal testing / development.

heaths (changes requested):
You can't take direct dependencies in a crate. There are also some sanitization concerns.

sdk/core/azure_core_amqp/Cargo.toml
k-shlomi: @heaths I referenced it in sdk/eventhubs/azure_messaging_eventhubs/Cargo.toml in order to use the sanitize. any suggestion about that? remove sanitize? leave it in the file? what is the best course of action here?

sdk/core/azure_core_amqp/Cargo.toml
k-shlomi: changed

sdk/core/azure_core_amqp/Cargo.toml
k-shlomi: added

k-shlomi:
@heaths one question for you in your comments. 
fixed the rest
updated from main so build passes

sdk/core/azure_core_amqp/Cargo.toml
heaths: I'm saying that the sanitization function should be exported in `azure_core` as well, in the same module (hopefully, unless we missed something). For any `azure_*` crates we're trying to "hide" the typespec crates as an implementation detail. Everything they export is imported and re-exported by `azure_core` in the same module path. A couple types we "override", but internally make convertible to the types they "override" in typespec.

k-shlomi: exposed it.

j7nw4r:
@k-shlomi

Hey. I'm on the Event Hubs team that is managing this sdk going forward. Thanks a lot for this work.

I'm currently speaking with our PM about how to consider this additional functionality. To integrate this change, we'd need to also add it to all of the other language SDKs (and probably the Service Bus SDKs) for parity. We're trying to determine if that's something we can take on.

PR #3056, opened 2025-09-25
Removed `create_enum` and `create_extensible_enum` macros

### LINKED ISSUE #1846: Remove create_enum and create_extensible_enum macros

LarryOsterman:
Fixes #1846 

Also update to the 0.23.2 emitter which removes the requirement for those macros.


sdk/core/azure_core_amqp/src/error.rs
```diff
@@ -57,102 +55,345 @@ pub enum AmqpErrorKind {
...
+        let s = String::deserialize(deserializer)?;
+        s.parse().map_err(serde::de::Error::custom)
+    }
+}
+impl serde::Serialize for AmqpErrorCondition {
+    fn serialize<S>(&self, s: S) -> ::core::result::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        s.serialize_str(self.as_ref())
```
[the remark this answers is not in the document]
LarryOsterman: That construction triggers a clippy failure, so is an inappropriate suggestion.

eng/emitter-package.json
```diff
@@ -1,17 +1,17 @@
...
+    "@azure-tools/typespec-client-generator-core": ">=0.60.1",
     "@typespec/compiler": "1.4",
     "@typespec/http": "1.4",
     "@typespec/openapi": "1.4",
     "@typespec/rest": "0.74",
     "@typespec/versioning": "0.74",
     "@typespec/xml": "0.74"
   }
-}
+}
```
RickWinter, line 17:
```suggestion
}

```

sdk/core/azure_core_amqp/src/error.rs
```diff
@@ -57,102 +55,345 @@ pub enum AmqpErrorKind {
...
+    SessionErrantLink,
     /// AMQP specific error conditions: session handle in use.
-    (SessionHandleInUse, "amqp:session:handle-in-use"),
+    SessionHandleInUse,
     /// AMQP specific error conditions: session unattached handle.
-    (SessionUnattachedHandle, "amqp:session:unattached-handle"),
-);
+    SessionUnattachedHandle,
+    /// Any other value not defined in `AmqpErrorCondition`.
+    UnknownValue(String),
```
RickWinter, line 149: How does `UnknownValue` work when there are more than one `UnknownValue`s?

eng/scripts/Update-TspClients.ps1
```diff
@@ -0,0 +1,66 @@
...
+    Runs tsp-client update in all directories with tsp-location.yml files under the current directory.
+
+.EXAMPLE
+    .\Update-TspClients.ps1 -Path "C:\my-project"
+    Runs tsp-client update in all directories with tsp-location.yml files under C:\my-project.
+#>
+
+param(
+  [Parameter(Mandatory = $false)]
+  [string]$Path = "."
```
heaths: I recommend defaulting to:
```suggestion
  [string]$Path = (Join-Path $PSScriptRoot "../../sdk" | Resolve-Path),
```

If someone started this in the root, it would look through all of target/ as well.

eng/scripts/Update-TspClients.ps1
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
heaths, line 3: Should put a shebang and chmod +x this for the vast majority of rust devs not on Windows:

```suggestion
#!/usr/bin/env pwsh

# Copyright (c) Microsoft Corporation. All rights reserved.
```

```sh
git update-index --chmod=+x eng/scripts/Update-TspClients.ps1
```

That said, it's even easier for us anyway:

```sh
find sdk -name tsp-location.yaml -execdir tsp-client update \;
```

sdk/core/azure_core_amqp/src/error.rs
heaths: There can only be one error variant, which just gets stuffed into here. Same with any extensible enum. This isn't any different than before, nor any other extensible enum in other languages.

sdk/typespec/typespec_client_core/src/lib.rs
heaths, line 1: CHANGELOG.md entries for this crate and `azure_core`.

heaths (changes requested):

eng/emitter-package.json
LarryOsterman: The instant I save this file, the trailing crlf gets removed by visual studio.

LarryOsterman:
> Are there tests for the extensible pattern? I couldn't find any. We need to ensure that a server returning multiple new values with work correctly.

As Heath mentioned, the server cannot return multiple new values - that's not how strings work in JSON.

An extensible enumeration is a single JSON string value. We take the string value sent from the server and match against the known enumerator values for the enumeration, if the server's value matches one of the known enumerator values, we return that enumerator. If the server's value doesn't match, it goes to the `Unknown Value` field.

eng/scripts/Update-TspClients.ps1
LarryOsterman: The windows powershell isn't that much worse, but I figured this was more repeatable.

Also, we're going to need this post-release to implement the "make sure the generated code is correct" PR.

sdk/typespec/typespec_client_core/src/lib.rs
LarryOsterman: I missed typespec_client_core, but got azure_core :(.

LarryOsterman:
At your convenience @heaths and @RickWinter

eng/emitter-package.json
heaths: Do you have the EditConfig extension installed? It shouldn't, unless you have some overriding settings in your personal settings.

heaths (approved):

LarryOsterman:
@RickWinter I need your approval as well since you requested changes.

RickWinter (approved):

Issue #3035, opened 2025-09-22
Publish crates using ESRP


danieljurek:
## Building
* [x] Use ESRP step in pipeline
* [x] Handle dependencies within a service 
    * [x] Basic: Fail if package is publishing without dependency
    * [x] Advanced: Allow user to select packages to publish, fail if dependency that needs to be published is missing

## Cleanup
* [x] Add `microsoft` as an owner to packages already published on crates.io

## Other considerations
* [ ] Add workspace support

## Cleanup 

* [ ] Transfer ownership to `microsoft-oss-releases`
* [ ] Yank published crates
* [ ] Remove `azure-sdk` and others from ownership of crates

PR #3065, opened 2025-09-29
ESRP publishing

danieljurek:
## Do I need to review this as a service owner? 

Probably not. This PR is for EngSys. You're probably seeing this review because I updated your `ci.yml` file to use "release intent" which allows you to choose which packages to release at release build time. We in EngSys will merge this after reviewing.

## Notes

Fixes https://github.com/Azure/azure-sdk-for-rust/issues/3035

* Release packages using ESRP 
* Release intent (check boxes to release packages) 
* You can run live tests without releasing packages
* Packages are released in dependency order
* Dependency enforcement (build fails if a dependency is not released or being released in the current build)

Updated runs with release intent: 

Sample pipeline run with 2 release crates: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5424835&view=results

Sample pipeline run with no releases and live tests: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5424839&view=logs&j=b766ebde-1fdb-5f11-1350-46ddc53b23cf&t=9494b470-f365-50d2-5008-6394593b39c5 -- In this case note that necessary dependencies were built (e.g. `typespec`) even though it was outside the scope of the service. Test failures notwithstanding.

Sample pipeline run with no releases (also no live tests): https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5424838&view=results

Example unified version increment PR: https://github.com/Azure/azure-sdk-for-rust/pull/3134

Sample releases: 
* https://github.com/Azure/azure-sdk-for-rust/releases/tag/azure_canary_core%401.0.0-beta.5397226
* https://github.com/Azure/azure-sdk-for-rust/releases/tag/azure_canary%401.0.0-beta.5397226

## Open questions

* Right now `Pack-Crates.ps1` fails in release scenarios when releasing a package that has unreleased dependencies (e.g. `dependency_crate = { workspace = true }`. The expectation here is that dependencies will go from `workspace = true` to `{ version = "1.2.3" }` when releasing. Is this a correct expectation?

### DIFF
--- a/eng/pipelines/templates/jobs/pack.yml
+++ b/eng/pipelines/templates/jobs/pack.yml
@@ -57,7 +57,7 @@ jobs:
 
   - ${{ if eq('auto', parameters.ServiceDirectory) }}:
     - task: Powershell@2
-      displayName: "Pack Crates (PR build)"
+      displayName: Pack Crates
       condition: and(succeeded(), ne(variables['NoPackagesChanged'],'true'))
       inputs:
         pwsh: true
@@ -85,7 +85,7 @@ jobs:
       displayName: Create package list
 
     - task: Powershell@2
-      displayName: "Pack Crates"
+      displayName: Pack Crates
       condition: and(succeeded(), ne(variables['NoPackagesChanged'],'true'))
       inputs:
         pwsh: true
--- a/eng/scripts/Language-Settings.ps1
+++ b/eng/scripts/Language-Settings.ps1
@@ -142,12 +142,12 @@ function Get-rust-AdditionalValidationPackagesFromPackageSet ($packagesWithChang
 # $GetPackageInfoFromPackageFileFn = "Get-${Language}-PackageInfoFromPackageFile"
 function Get-rust-PackageInfoFromPackageFile([IO.FileInfo]$pkg, [string]$workingDirectory) {
   # Create a temporary folder for extraction
-  $extractionPath = Join-Path ([System.IO.Path]::GetTempPath()) ([System.IO.Path]::GetRandomFileName())
+  $extractionPath = [System.IO.Path]::Combine([System.IO.Path]::GetTempPath(), [System.IO.Path]::GetRandomFileName())
   New-Item -ItemType Directory -Path $extractionPath | Out-Null
 
   # Extract the .crate file (which is a tarball) to the temporary folder
   tar -xvf $pkg.FullName -C $extractionPath
-  $cargoTomlPath = Join-Path $extractionPath $pkg.BaseName 'Cargo.toml'
+  $cargoTomlPath = [System.IO.Path]::Combine($extractionPath, $pkg.BaseName, 'Cargo.toml')
 
   Write-Host "Reading package info from $cargoTomlPath"
   if (!(Test-Path $cargoTomlPath)) {
@@ -161,7 +161,7 @@ function Get-rust-PackageInfoFromPackageFile([IO.FileInfo]$pkg, [string]$working
   $packageName = $package.name
   $packageVersion = $package.version
 
-  $packageAssetPath = Join-Path $extractionPath "$packageName-$packageVersion"
+  $packageAssetPath = [System.IO.Path]::Combine($extractionPath, "$packageName-$packageVersion")
 
   $changeLogLoc = Get-ChildItem -Path $packageAssetPath -Filter "CHANGELOG.md" | Select-Object -First 1
   $readmeContentLoc = Get-ChildItem -Path $packageAssetPath -Filter "README.md" | Select-Object -First 1


.vscode/cspell.json
```diff
@@ -2,6 +2,7 @@
   "version": "0.2",
   "language": "en",
   "useGitignore": true,
+  "allowCompoundWords": true,
```
danieljurek: This prevents cspell errors from words like `azurecanary` in ci.yml that don't have case or other separators.

eng/scripts/Pack-Crates.ps1
```diff
@@ -162,6 +168,29 @@ try {
...
+    $unspecifiedPackages = $packages.name | Where-Object { $_ -notin $PackageNames }
+    if ($unspecifiedPackages.Count -gt 0) { 
+      Write-Error "Packages in -PackageNames require dependencies that are either not released or not listed for packing: $($unspecifiedPackages -join ', ')"
+      exit 1
+    }
+
+    $orderMatches = $true
+    for ($i = 0; $i -lt $PackageNames.Count; $i++) {
+      if ($packages[$i].name -ne $PackageNames[$i]) {
+        $orderMatches = $false
```
danieljurek: This change uses Artifact names coming from the service's `sdk/<service>/ci.yml` file to supply the `-PackageNames` parameter. The sorting of those artifact names must match the sorted order created by this script for a successful release (while we're releasing one package per-ESRP step). 

It's possible to change the sorting by altering the placement of `workspace.members` in the root `Cargo.tml` file. 

If the sorting is incorrect, you'll get an error at release time that the order of packages in `-PackageNames` is not correct. To fix, update the order of the service's `ci.yml` file to match the order that the script provides and then run the release. 

This validation currently only happens during release because the pipeline yaml templates aren't aware of the various libraries' dependencies in the state of their Cargo.toml files (e.g. workspace dependencies vs. released dependencies).

eng/scripts/Pack-Crates.ps1
```diff
@@ -162,6 +168,29 @@ try {
 
   [array]$packages = Get-PackagesToBuild
 
+  if ($RequireDependencies) {
+    $unspecifiedPackages = $packages.name | Where-Object { $_ -notin $PackageNames }
+    if ($unspecifiedPackages.Count -gt 0) {
```
hallipr, line 174: This doesn't seem to allow for packages outside of this pipeline run to be accommodate the current release.  e.g. :
`azure_core_amqp` depends on `azure_core` 
`azure_core` depends on `typespec` 
We're building `core`, but `typespec` isn't shipped in the `core` pipeline

We should be able to build with `$PackageNames = ('azure_core', 'azure_core_amqp')`, even though there's that lingering `typespec` dependency because we know that, before releasing those packages, we'll be releasing `typespec` in a separate pipeline.

This is different from `$PackageNames = ('azure_core_amqp')`  where we know `azure_core_amqp` depends on `azure_core`, we know this is the pipeline that should build it and we know we're not building it.

I'm proposing that we call Pack-Crates.ps1 with a list of all the artifacts this pipeline is responsible for, along with a list of packages we intend on releasing.  If any dependency is in the list we're responsible for but aren't releasing, then we error.

Allowed
```
Pack-Crates.ps1 -Packages 'azure_core' -PipelinePackages 'azure_core_amqp', 'azure_core'
```

Should error
```
Pack-Crates.ps1 -Packages 'azure_core_amqp' -PipelinePackages 'azure_core_amqp', 'azure_core'
```

danieljurek: In the scenario where we're releasing `azure_core_amqp` and there's a dependency on `typespec` how does `azure_core_amqp`'s Cargo.toml file express the dependency on `typespec`? 

https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/core/azure_core_amqp/Cargo.toml#L33

As it's set up in this PR, the `core` pipeline would fail if it's depending on a release of `typespec` that hasn't yet completed. So you need to queue and release `typespec` and then you can queue and release your packages in `core` that depend on `typespec`.

hallipr: For packing/publishing `cargo` will elevate workspace dependencies and remove any path properties from the dependencies, leaving their other properties:

given the workspace dependency in the root toml:
```
default-features = false
path = "sdk/typespec"
version = "0.9.0"
```

azure_core_amqp's dependency would process as:
```
typespec = { workspace = true, features = ["amqp"] }  =>
typespec = { path = "sdk/typespec", version = "0.9.0", features = ["amqp"] }  =>
typespec = { version = "0.9.0", features = ["amqp"] }
```

Where typespec 0.9.0 is currently unreleased.

hallipr: If we queue and release typespec, we need to uptake the version bump PR to rev the unreleased, path version to 0.10.0.

Doing this would cause azure_core_amqp to depend on the new, unreleased 0.10.0

analogrelay (approved):
Approving from Cosmos

eng/scripts/Pack-Crates.ps1
danieljurek: @hallipr and I went over this on the whiteboard and it works out to mostly YAML changes: 

* Check boxes in the pipeline queuing UI signals release intent
* If no boxes are checked (basically, run live tests but don't release anything) and if a pipeline is run manually then treat it as `Pack-Crates.ps1 -OutputPath '...' -PackageNames all,packages,in,build -RequireDependencies:$false` 
* If boxes are checked, that signals release intent, then use `-RequireDependencies:$true`

danieljurek: @hallipr - Can you have another look? I updated the YAML. I'm debating pulling the inline PowerShell into a file so we do local iteration but it's probably not going to change much as written.

eng/scripts/Pack-Crates.ps1
```diff
@@ -93,6 +95,10 @@ function Get-PackagesToBuild() {
 
     foreach ($dependency in $package.UnreleasedDependencies) {
       if (!$packagesToBuild.Contains($dependency) -and !$toProcess.Contains($dependency)) {
+        if ($RequireDependencies -and $dependency.name -notin $PackageNames) { 
+          Write-Warning "Package $($package.name) depends on unreleased or unspecified dependency: $($dependency.name)"
```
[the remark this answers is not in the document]
danieljurek: Failure is evaluated in other parts of the script. Warning messages should accumulate in failure cases so the user has a better idea of what to do next.

eng/scripts/Language-Settings.ps1
```diff
@@ -139,15 +139,31 @@ function Get-rust-AdditionalValidationPackagesFromPackageSet ($packagesWithChang
...
+  New-Item -ItemType Directory -Path $extractionPath | Out-Null
+
+  # Extract the .crate file (which is a tarball) to the temporary folder
+  tar -xvf $pkg.FullName -C $extractionPath
+  $cargoTomlPath = Join-Path $extractionPath $pkg.BaseName 'Cargo.toml'
+
+  Write-Host "Reading package info from $cargoTomlPath"
+  if (!(Test-Path $cargoTomlPath)) {
+    LogError "The Cargo.toml file was not found in the package artifact at $cargoTomlPath"
+    throw "error"
```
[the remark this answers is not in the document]
danieljurek: Nice catch! I should've had a TODO comment on that.

.vscode/cspell.json
heaths: Don't do that for *everything* then. use `# cspell:ignore <error>` above lines. We want to catch things elsewhere.

sdk/core/ci.yml
```diff
@@ -10,16 +10,38 @@ trigger:
     include:
     - sdk/core/
 
+parameters:
+- name: release_azurecore
```
heaths: Related to the previous comment, why can't you just use separators. In fact, even easier: copy and paste the crate name and preface it with "release_". Fewer steps compared to typing the whole thing out and hoping you don't make a typo ramming a bunch of words together that makes it harder to read than without separators (space, underscore, etc.).

eng/scripts/Pack-Crates.ps1
heaths: This should probably be documented somewhere - as a comment in this file, or in some release docs somewhere in the repo.

.vscode/cspell.json
heaths: If you don't rename as I suggested in my other comment, at least limit it to the pipelines directory/ies like we do in `overrides` below.

eng/scripts/Pack-Crates.ps1
danieljurek: Heath and I expanded on this a few moments ago: 

* Relying on the user to get the order right in `ci.yml` is going to be a source of problems. Instead, we should see if the ordering can be done "dynamically" in the job. This will require some more YAML

eng/pipelines/templates/jobs/pack.yml
```diff
@@ -55,16 +55,50 @@ jobs:
...
+        filePath: $(Build.SourcesDirectory)/eng/scripts/Pack-Crates.ps1
+        arguments: >
+          -OutputPath '$(Build.ArtifactStagingDirectory)'
+          -PackageInfoDirectory '$(Build.ArtifactStagingDirectory)/PackageInfo'
 
+  - ${{ else }}:
+    - pwsh: |
+        $artifacts = '${{ convertToJson(parameters.Artifacts) }}' | ConvertFrom-Json
+        $requireDependencies = $true
+        $artifactsToBuild = $artifacts | Where-Object { $_.releaseInBatch -ne 'False' }
```
hallipr: Why the double negative here?

danieljurek: I should change it to `-eq 'True'` 

The JSON from `convertToJson` outputs booleans as strings: 

```json
[
  {
    "name": "azure_canary",
    "releaseInBatch": "True"
  },
  {
    "name": "azure_canary_core",
    "releaseInBatch": "True"
  }
]
```

eng/scripts/Language-Settings.ps1
```diff
@@ -139,15 +139,32 @@ function Get-rust-AdditionalValidationPackagesFromPackageSet ($packagesWithChang
   return $additionalPackages ?? @()
 }
 
+# $GetPackageInfoFromPackageFileFn = "Get-${Language}-PackageInfoFromPackageFile"
 function Get-rust-PackageInfoFromPackageFile([IO.FileInfo]$pkg, [string]$workingDirectory) {
-  #$pkg will be a FileInfo object for the Cargo.toml file in a package artifact directory
-  $package = cargo read-manifest --manifest-path $pkg.FullName | ConvertFrom-Json
+  # Create a temporary folder for extraction
+  $extractionPath = Join-Path ([System.IO.Path]::GetTempPath()) ([System.IO.Path]::GetRandomFileName())
```
heaths: I strongly recommend against `Join-Path`. It's buggy, and basically just does string concat. It's funny you're using `[Path]::GetTempPath()` when you could use `$env:TEMP`, but don't use `[Path]::Combine()`, which I recommend over `Join-Path`. It actually handles both absolute and relative paths correctly. `Join-Path` will concat an absolute child path to a base path without thinking twice, which obviously is not what you want.

eng/scripts/Language-Settings.ps1
heaths, line 1: My concern here is that it assumes the `.crate` file was already created, or is that fine for the function being called i.e., that it's only called after any packages have been packaged already? `$packagePattern` wasn't used anywhere else to get package info? Because I know I just recently made changes where information herein does get used.

eng/pipelines/templates/jobs/pack.yml
```diff
@@ -55,15 +55,46 @@ jobs:
...
-    condition: and(succeeded(), ne(variables['NoPackagesChanged'],'true'))
-    inputs:
-      pwsh: true
-      filePath: $(Build.SourcesDirectory)/eng/scripts/Pack-Crates.ps1
-      arguments: >
-        -OutputPath '$(Build.ArtifactStagingDirectory)'
-        -PackageInfoDirectory '$(Build.ArtifactStagingDirectory)/PackageInfo'
+  - ${{ if eq('auto', parameters.ServiceDirectory) }}:
+    - task: Powershell@2
+      displayName: "Pack Crates (PR build)"
```
heaths: Nit: quotes aren't necessary in any string unless it starts with a special YAML char or could otherwise be interpreted as number, such as versions (which will then fail with more than one `.`). Just a nit, though. I mean, you don't quote anything else here.

heaths: And do we need `(PR build)`? It's a mutually exclusive section; but, if it helps to identify at a glance, sure.

eng/pipelines/templates/jobs/pack.yml
```diff
@@ -55,15 +55,46 @@ jobs:
...
+    - task: Powershell@2
+      displayName: "Pack Crates"
+      condition: and(succeeded(), ne(variables['NoPackagesChanged'],'true'))
+      inputs:
+        pwsh: true
+        filePath: $(Build.SourcesDirectory)/eng/scripts/Pack-Crates.ps1
+        arguments: >
+          -OutputPath '$(Build.ArtifactStagingDirectory)'
+          -PackageNames $(PackageNames)
+          -RequireDependencies:$$(RequireDependencies)
```
heaths, line 96: Is the double `$$` intentional here? Is it expected that `$(RequireDependencies)` returns a string like "true" or "false", which then gets turned into `$true` or `$false`?

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -84,33 +85,77 @@ stages:
...
+                        $packageVersion = $packageMetadata.version
+                        Write-Host "Package version: $packageVersion"
+
+                        $cratePath = "$artifactRootPath/$artifactName/$artifactName-$packageVersion.crate"
+                        Copy-Item `
+                          -Path $cratePath `
+                          -Destination $outDir
+                        Write-Host "Contents of $outDir"
+                        Get-ChildItem -Path $outDir | ForEach-Object { Write-Host $_.FullName }
+                      displayName: 'Copy crate for ESRP'
```
heaths, line 133: There's a lot going on here we should put into `eng/scripts` instead so we can more easily test it without going through the painful commit->push->back to commit loop endemic of pipelines. Ideally, a script that takes simple inputs and spits out an order would be great.

heaths: Add a shebang and make it executable. As with most rustaceans (same for gophers), most development is done on linux/WSL.

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -84,33 +85,77 @@ stages:
...
+
+                    - task: EsrpRelease@10
+                      displayName: 'ESRP Release'
+                      inputs:
+                        connectedservicename: 'Azure SDK PME Managed Identity'
+                        ClientId: '5f81938c-2544-4f1f-9251-dd9de5b8a81b'
+                        DomainTenantId: '975f013f-7f24-47e8-a7d3-abc4752bf346'
```
heaths, line 140: Should these be hardcoded? Not an SFI concern, but seems like they should be in a variable group or something for easy updates if necessary.

sdk/canary/ci.yml
```diff
@@ -10,12 +10,22 @@ trigger:
     include:
     - sdk/canary/
 
+parameters:
+- name: release_azure_canary_core
+  displayName: 'azure_canary_core'
```
heaths, line 15: Nit: again, no quotes needed. Less to type if we show a good example.

heaths: ...though, it does line the crate name up nicely with that extra leading character. Not necessary, but I like the benefit, so I'll resolve this but retain it for posterity.

heaths (approved):

eng/scripts/Language-Settings.ps1
danieljurek: I haven't seen problems with `Join-Path` but I don't mind switching.

`$env:TEMP` is not universal, on Linux you aren't guaranteed to get a value back. `[System.IO.Path]::GetTempPath()` is more universal (though it's [probably just returning `/tmp/`](https://github.com/dotnet/runtime/blob/62578463d0b13c7694cf679caa6cbb1458a35c73/src/libraries/System.Private.CoreLib/src/System/IO/Path.Unix.cs#L81)).

eng/pipelines/templates/jobs/pack.yml
danieljurek: It's necessary for string conversion. The `$(RequireDependencies)` literal string becomes the string value (`True` or `False`) but to pass in a boolean to the switch you need to give a boolean, not a string. 

The expansion works out to: 
1. `-RequireDependencies:$$(RequireDependencies)`
2. `-RequireDependencies:$true`

If you did something like `-RequireDependencies:true` you'll get an error about converting a string to a boolean.

eng/pipelines/templates/stages/archetype-rust-release.yml
danieljurek: Deploy jobs can't check out source so the script logic gets inlined.

eng/pipelines/templates/stages/archetype-rust-release.yml
danieljurek: These have been hardcoded historically

eng/scripts/Language-Settings.ps1
danieljurek: This value is used in `artifact-metadata-parsing.ps1` which is used to retrieve metadata from built packages. It calls `Get-rust-PackageInfoFromPackageFile` to get info from a package file. I verified functionality (including APIView) when I switched the artifacts over from source to the `.crate` files.

heaths (approved):

danieljurek:
Bypassing rules to merge. This can go in.

PR #3070, opened 2025-09-30
[Storage] Flattened Clients

vincenttran-msft:


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -85,7 +160,14 @@ impl BlobClient {
...
             endpoint: self.client.endpoint.clone(),
-            client: self.client.get_append_blob_client(),
+            client: GeneratedAppendBlobClient {
+                endpoint: self.client.endpoint.clone(),
+                pipeline: self.client.pipeline.clone(),
+                version: self.client.version.clone(),
+                tracer: todo!(),
+            },
+            container_name: self.container_name().to_string(),
+            blob_name: self.blob_name().to_string(),
```
vincenttran-msft, line 186: Just leaving a specific comment here but applies to anytime we are using struct-initialization on a generated client i.e. `GeneratedAppendBlobClient`

What is the correct value to pass to the `tracer` field?

LarryOsterman: I'd use the `tracer` field from the existing service client's `self` as the correct value for `tracer`.

PR #3076, opened 2025-09-30
vcpkg: use x-azcopy as x-azblob requires SAS token (even in read scenarios)

danieljurek:
vcpkg introduced a new set of features and changed behavior for `x-azblob` in cases where no SAS is provided. There's an `x-azcopy` configuration which does not require SAS and works in `read` scenarios.

Successful PR build: https://dev.azure.com/azure-sdk/public/_build/results?buildId=5404051&view=logs&j=f62abde3-0c2c-58c2-34fa-f25f24e797de&t=7bf7baf7-4679-52b3-46c9-d10933f9b078
Successful internal build: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5404053&view=logs&j=f62abde3-0c2c-58c2-34fa-f25f24e797de&t=7bf7baf7-4679-52b3-46c9-d10933f9b078

### DIFF
--- a/eng/scripts/Set-VcpkgWriteModeCache.ps1
+++ b/eng/scripts/Set-VcpkgWriteModeCache.ps1
@@ -42,5 +42,5 @@ Write-Host "Ensure redaction of SAS tokens in logs"
 Write-Host "##vso[task.setvariable variable=VCPKG_BINARY_SAS_TOKEN;issecret=true;]$vcpkgBinarySourceSas"
 
 Write-Host "Setting vcpkg binary cache to read and write"
-Write-Host "##vso[task.setvariable variable=VCPKG_BINARY_SOURCES_SECRET;issecret=true;]clear;x-azblob,https://$StorageAccountName.blob.core.windows.net/$StorageContainerName,$vcpkgBinarySourceSas,readwrite"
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/steps/vcpkg.yml
```diff
@@ -1,6 +1,6 @@
 steps:
   - pwsh: |
-      Write-Host '##vso[task.setvariable variable=VCPKG_BINARY_SOURCES_SECRET;issecret=true;]clear;x-azblob,https://azuresdkartifacts.blob.core.windows.net/public-vcpkg-container,,read'
+      Write-Host '##vso[task.setvariable variable=VCPKG_BINARY_SOURCES_SECRET;issecret=true;]clear;x-azcopy,https://azuresdkartifacts.blob.core.windows.net/public-vcpkg-container,read'
```
weshaggard, line 3: https://github.com/Azure/azure-sdk-for-rust/blob/main/eng/scripts/Set-VcpkgWriteModeCache.ps1#L45

I assume you need to update that as well

danieljurek: Not required at this point because only `x-azblob` without an SAS behavior was changed... but it's the right thing to update.

danieljurek:
Failed job is from skipping release in canary pipeline. It's OK to override check-enforcer

PR #3087, opened 2025-10-02
Add cloud configuration API

### LINKED ISSUE #2439: Support cloud configuration

chlowell:
Closes #2439 by enabling applications to configure (data plane) clients and credentials for sovereign and private clouds. `CustomConfiguration` is `non_exhaustive` because we'll eventually need to add data for ARM clients--probably just a field like `resource_manager_endpoint: Option<String>`.

### SDK implementation
```rs
pub struct FooClient {}

pub struct FooClientOptions {
}

pub struct Audience;

pub fn new(options: Option<FooClientOptions>) -> Result<FooClient, Error> {
    let cloud = options
        .unwrap_or_default()
        .client_options
        .cloud
        .unwrap_or_default();

    let audience = match cloud.as_ref() {
        CloudConfiguration::AzurePublic => "https://example.com",
        CloudConfiguration::AzureGovernment => "https://example.us",
        CloudConfiguration::AzureChina => "https://example.cn",
        CloudConfiguration::Custom(CustomConfiguration { audiences, .. }) => {
            audiences.audience::<Audience>().ok_or_else(|| {
                Error::with_message(
                    ErrorKind::Other,
                    "cloud CustomConfiguration doesn't have a value for audience",
                )
            })?
        }
        // (need this because CloudConfiguration is non_exhaustive)
        _ => {
            return Err(Error::with_message(
                ErrorKind::Other,
                "cloud CustomConfiguration doesn't have a value for audience",
            ));
        }
    };
    // ...
}
```

### application
```rs
let mut custom_config = CustomConfiguration::default();
custom_config.audiences = Audiences::new()
    .with::<azure_foo_client::Audience>("https://custom.audience.local".to_string());

let cloud_config = Arc::new(CloudConfiguration::Custom(custom_config.into()));

let foo_client = new(azure_foo_client::FooClientOptions {
    client_options: ClientOptions {
        cloud: Some(cloud_config),
        ..Default::default()
    },
});
```


sdk/core/azure_core/src/http/options/mod.rs
```diff
@@ -41,6 +43,9 @@ pub struct ClientOptions {
     ///
     /// Specifies which headers and query parameters should be logged. All headers and query parameters not in the allow list will be redacted.
     pub logging: LoggingOptions,
+
+    /// Cloud configuration for the client. If None, the client will default to Azure Public Cloud.
+    pub cloud: Option<Arc<CloudConfiguration>>,
```
heaths, line 48: This isn't the current pattern: any options that are options bags themselves are just inline - not `Option<T>`. Only leave nodes are `Option<T>` unless they also have a zero-size default like `Vec` and `HashMap`.

sdk/core/azure_core/src/cloud.rs
```diff
@@ -0,0 +1,118 @@
...
+    /// #   pub struct Audience;
+    /// # }
+    ///
+    /// use azure_core::{
+    ///     cloud::{Audiences, CloudConfiguration, CustomConfiguration},
+    ///     http::ClientOptions,
+    /// };
+    ///
+    /// let mut custom = CustomConfiguration::default();
+    /// custom.audiences = Audiences::new().with::<azure_service_module::Audience>("https://service.mycloud.local".to_string());
```
heaths: No doubt this would reformat, which is nice for docs with limited width in tooltips, for example:

```suggestion
    /// custom.audiences = Audiences::new()
    ///     .with::<azure_service_module::Audience>("https://service.mycloud.local".to_string());
```

sdk/core/azure_core/src/cloud.rs
```diff
@@ -0,0 +1,118 @@
...
+        Self(HashMap::new())
+    }
+
+    /// Get a module's audience.
+    pub fn audience<T: 'static>(&self) -> Option<&str> {
+        self.0.get(&TypeId::of::<T>()).map(|s| s.as_str())
+    }
+
+    /// Insert or replace an audience and return `Self` to allow chaining.
+    pub fn with<T: 'static>(mut self, audience: String) -> Self {
```
heaths, line 83: Probably should add an `insert<T>(&mut self)` as well.

heaths: Also, this only sets the `audience` so should be `with_audience` per guidelines. Alternatively, keep them both generic and just have `get` and `with`, which might make more sense. The idea is to have something of the name - the getter and setter, if you will - linking the two obviously.

sdk/core/azure_core/src/http/options/mod.rs
heaths: ...though, as I think about this more, maybe it does make sense here since it'd just ben an `Arc<CloudConfiguration>` otherwise, which we don't often need. You know what? Never mind on this one. I think it makes sense as is but saving this thread for posterity.

heaths (changes requested):

heaths:
@chlowell @jhendrixMSFT @antkmsft this is also something we'll have to generate in the emitter. But, @chlowell,

1. Since only the AzurePublic endpoint/audience tends to get spec'd, how do we even know the other endpoint domains to emit?
2. Since the caller passes the whole domain in anyway, don't we only need this for audience and could define that for every cloud e.g.,

```rust
impl CloudConfiguration {
    pub fn authority_host(&self) -> &str {
        match self {
            Self::AzurePublic => "https://login.microsoftonline.com",
            // ...
            Self::Custom(ref custom) => &custom.authority_host,
        }
    }
}
```

sdk/core/azure_core/src/http/options/mod.rs
chlowell: I'm actually a little uncomfortable with the `Arc` because it feels like optimizing for the least common case. It makes sense for the `Custom` variant but it's just ceremonial overhead for the others.

sdk/core/azure_core/src/cloud.rs
chlowell: My reasoning was that `_audience` is implicit because this is `Audiences::with()`, the collection holds non-compound values (Strings), and client modules will probably name their key type `Audience`. So, I like `get` and `with`

heaths (approved):

PR #3088, opened 2025-10-02
[Storage] Flattened clients v2

vincenttran-msft:


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -38,43 +42,130 @@ use std::sync::Arc;
 pub struct BlobClient {
     pub(super) endpoint: Url,
     pub(super) client: GeneratedBlobClient,
+    pub(super) container_name: String,
+    pub(super) blob_name: String,
+}
+
+impl GeneratedBlobClient {
+    fn from_url(
```
LarryOsterman: You may be able to use `#[tracing::new]` here - it *should* work and hide the tracer initialization.

vincenttran-msft: Thanks Larry, that works out much cleaner! However, this requires us to make the `from_url` function on the underlying `GeneratedBlobClient` now public (which is _not_ desired as everything should go through our convenience layer).

This is not yet the final design, so Storage will discuss internally and ensure we land on our final design before requesting any Core code changes (i.e. removing the requirement here that the visibility must be public to utilize the macro)

LarryOsterman: You might want to resync with main (which may be painful because of Heath's recent changes). He ran into the same issue and we decided that the public requirement wasn't necessary or appropriate. 

So changing core is probably the right thing to do.

vincenttran-msft: @LarryOsterman Following up on this, was the change to remove the public requirement live yet or still living in an unmerged PR? I am rehydrating this after the Core changes and still facing the error:
image

Thanks!

LarryOsterman: Unfortunately it appears it's still in a private branch. If you'd like, you can certainly fix this yourself - it's lines 376 and 377 of `tracing_new.rs` in the azure_core_macros package.

vincenttran-msft: Ah gotcha. No worries, we are still ironing out some kinks in the Storage side of the design, but I can certainly contribute that back if it continues to be a problem for us. Thanks Larry 😄

PR #3110, opened 2025-10-03
Initial performance test infrastructure

LarryOsterman:
This PR adds the start of performance test infrastructure for the Rust SDK.

This support is not complete, but it does provide a mechanism to measure the performance of SDK client operations.

It also defines a performance test for the KeyVault get_secrets API and the Storage Blobs list_blobs API which match the tests which exist for the C++ SDK.


sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -0,0 +1,536 @@
...
+                };
+
+                let json = serde_json::to_string_pretty(&results).map_err(|e| {
+                    Error::with_error(
+                        azure_core::error::ErrorKind::Other,
+                        e,
+                        "Failed to serialize test results to JSON.",
+                    )
+                })?;
+                println!("Test results: {}", json);
```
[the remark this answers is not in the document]
LarryOsterman: This intentionally uses println! because the performance infrastructure expects to see output on stdout.

doc/performance-test-requirements.md
```diff
@@ -0,0 +1,79 @@
+# Requirements for performance tests
+
+Each performance test consists of three phases:
+
+1) Warmup
+1) Test operation
+1) Cleanup
```
heaths: Nit: we don't use this syntax anywhere else in this repo. I'd just use `1.`. We also generally like to use the right numbers (just for plain text reading), but an even smaller nit. I appreciate the flexibility of all 1s. 🙂

doc/performance-test-requirements.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Common test inputs
+
+* Duration of the test in seconds
+* Number of iterations of the main test loop
+* Parallel - number of operations to execute in parallel
+* Disable test cleanup
+* Test Proxy servers.
```
heaths: Nit: all throughout these bullets here and below, I'm seeing a mix of trailing periods and without. Would look better to be consistent. Personally, I like complete sentences with punctuation, but whatever you prefer (if any).

doc/performance-test-requirements.md
```diff
@@ -0,0 +1,79 @@
...
+A `TestOptions` defines a set of options for the test which will be merged with the common test inputs to define the command line for the performance test.
+
+```rust
+pub struct TestOption {
+    /// The name of the test option. This is used as the key in the `TestArguments` map.
+    name: &'static str,
+
+    long_activator: &str,
+
+    short_activator:&str,
```
heaths: Nit: one, you're missing a space after `short_activator:` but why aren't these two `'static` lifetime like `name`?

sdk/core/azure_core_test/src/lib.rs
```diff
@@ -38,6 +38,8 @@ pub struct TestContext {
     recording: Option<Recording>,
 }
 
+unsafe impl Send for TestContext {}
```
heaths: Would it be better to make `Recording` `Send`able? Either truly, or unsafely implementing `Send` like this for that? Because if so, then `TestContext` would be `Send` automatically. Looking at the `Recording` struct, I'm not sure exactly what makes it not `Send` (would be nice if there was a way to have the compiler tell us).

doc/performance-test-requirements.md
heaths, line 1: Probably should have something in here too about what you need to have in `Cargo.toml` e.g.,

```toml
[[test]]
name = "perf" # name must be "perf"
path = "perf/mod.rs" # or whatever your `main` is
```

And maybe "see `azure_security_keyvault_secrets` for a simple example, or `azure_storage_blob` for a more complex example" or something like that. Also might help LLMs.

heaths: Or maybe that's what your `perf/README.md` is saying. In that case, maybe just reference that via absolute URL as it would be on github.com (which should pass the PR phase if it *will* be published; a change I helped fix years back).

sdk/core/azure_core_test/src/perf/README.md
```diff
@@ -0,0 +1,118 @@
+# Performance Tests
+
+The Azure SDK defines a standardized set of performance tests which use a test framework defined by the [PerfAutomation tool](https://github.com/Azure/azure-sdk-tools/tree/main/tools/perf-automation).
+
+Performance tests are defined in a "perf" directory under the package root.
+
+By convention, all performance tests are named "perf" and are invoked via:
+
+```bash
+cargo test --package <package name> --test perf -- <perf test name> <perf test arguments>
```
heaths: The `<` and `>` won't render well since they're valid shell redirects when using a `bash` codefence. Maybe use `{` instead:


```suggestion
cargo test --package {package name} --test perf -- {perf test name} {perf test arguments}
```

sdk/core/azure_core_test/src/perf/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+1) Setup - Establish any resources needed to run the test.
+1) Run - Actually perform the test.
+1) Cleanup - Cleanup any resources used by the test.
+
+Each is defined by functions on the `PerfTest` trait.
+
+### Test Metadata
+
+Tests are defined by an instance of a `PerfTestMetadata` structure, which defines the name of the test, and other information about the test.
```
heaths, line 40: Note that in your other doc under `docs/`, you said it was just `TestMetdata`.

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -0,0 +1,536 @@
...
+    ///
+    /// # Arguments
+    /// - `context`: An `Arc` to a `TestContext` that provides context information for the test.
+    async fn setup(&self, context: Arc<TestContext>) -> azure_core::Result<()>;
+    async fn run(&self, context: Arc<TestContext>) -> azure_core::Result<()>;
+    async fn cleanup(&self, context: Arc<TestContext>) -> azure_core::Result<()>;
+}
+
+pub type CreatePerfTestReturn =
+    Pin<Box<dyn Future<Output = azure_core::Result<Box<dyn PerfTest>>>>>;
```
heaths, line 48: Does it have to be `azure_core::Result? If it helps, great. `std::result::Result<Box<dyn PerfTest>, Box<dyn std::error::Error>>` is definitely more flexible in tests, or maybe that's not relevant in this exact use case (still going through code).

For example, when I write tests even in a crate that defines it's own `Result`, I generally still return a standard `Result` with a boxed standard `Error` just in case I want to `?` something else. Just more flexible.

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -0,0 +1,536 @@
...
+    pub description: &'static str,
+    /// The set of test options supported by this test.
+    pub options: Vec<PerfTestOption>,
+
+    /// An async function used to create the performance test.
+    /// Takes a PerfRunner reference and returns a future that resolves to a PerfTest trait object.
+    pub create_test: CreatePerfTestFn,
+}
+
+/// #A `TestOptions` defines a set of options for the test which will be merged with the common test inputs to define the command line for the performance test.
```
heaths:
```suggestion
/// A `PerfTestOption` defines a set of options for the test which will be merged with the common test inputs to define the command line for the performance test.
```

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -0,0 +1,536 @@
...
+#[derive(Debug, Default, Clone)]
+pub struct PerfTestOption {
+    /// The name of the test option. This is used as the key in the `TestArguments` map.
+    pub name: &'static str,
+
+    /// The short form activator for this argument e.g., `-t`. Does not include the hyphen.
+    pub short_activator: char,
```
heaths: Could this be optional, just like `clap` and most other argument parsers are? I just worry it will force parameter-laden tests to start picking characters that would otherwise overlap just because they have to.

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -0,0 +1,536 @@
...
+    pub operations_per_second: f64,
+    pub average_cpu_use: Option<f64>,
+    pub average_memory_use: Option<f64>,
+}
+
+#[derive(Debug, Clone)]
+struct PerfRunnerOptions {
+    no_cleanup: bool,
+    iterations: u32,
+    parallel: usize,
```
heaths: Interesting that `iterations` is bound by 32 bits, but `parallel` is 64 bits (or more in the future)?

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -0,0 +1,536 @@
...
+            self.parallel,
+            self.duration,
+            self.warmup,
+            self.disable_progress,
+            self.test_results_filename
+        )
+    }
+}
+
+impl PerfRunnerOptions {}
```
heaths: Nit: no need. Remove.

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -0,0 +1,536 @@
...
+            package_dir,
+            module_name,
+            progress: Arc::new(AtomicU64::new(0)),
+        })
+    }
+
+    /// Gets a reference to a typed argument by its id.
+    pub fn try_get_global_arg<T>(&self, id: &str) -> Result<Option<&T>>
+    where
+        T: Any + Clone + Send + Sync + 'static,
```
heaths: Shouldn't need `Any`.

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -0,0 +1,536 @@
...
+        self.arguments.try_get_one::<T>(id).map_err(|e| {
+            Error::with_error(
+                azure_core::error::ErrorKind::Other,
+                e,
+                format!("Failed to get argument '{}'.", id),
+            )
+        })
+    }
+
+    pub fn try_get_test_arg<T>(&self, id: &str) -> Result<Option<&T>>
```
heaths, line 243: Docs on the pub members (here and below) - at least at some point.

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -0,0 +1,536 @@
...
+        T: Any + Clone + Send + Sync + 'static,
+    {
+        if let Some((_, args)) = self.arguments.subcommand() {
+            args.try_get_one::<T>(id).map_err(|e| {
+                Error::with_error(
+                    azure_core::error::ErrorKind::Other,
+                    e,
+                    format!("Failed to get argument '{}' for test.", id),
+                )
+            })
```
heaths: Tip: you can actually use our `ResultExt` like so:

```rust
args.try_get_one::<T>(id).with_context(ErrorKind::Other, format!("..."))
```

Automatically wraps the error like you did. Just less typing.

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -0,0 +1,536 @@
...
+                loop {
+                    test_instance_clone.run(test_context.clone()).await?;
+                    progress.fetch_add(1, Ordering::SeqCst);
+                }
+            });
+        });
+        let start = tokio::time::Instant::now();
+        let timeout = tokio::time::Duration::from_secs_f64(duration.as_seconds_f64());
+        select!(
+                _ = tokio::time::sleep(timeout) => {println!("Timeout reached, stopping test tasks: {:?}", start.elapsed());},
```
heaths, line 430: Tip: could also use [`timeout`](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html).

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -0,0 +1,536 @@
...
+                _ = tasks.join_all() =>  {println!("All test tasks completed: {:?}", start.elapsed());},
+                _ = async {
+                        let mut last_count = 0;
+                        loop {
+                            tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+                            let current_total = self.progress.load(Ordering::SeqCst);
+                            // println!("{:<10?} elapsed: {:.5} op/sec, {:4} sec/operation.",
+                            //     start.elapsed(),
+                            //     self.progress.load(Ordering::SeqCst) as f64 / start.elapsed().as_secs_f64(),
+                            //     Duration::seconds_f64( start.elapsed().as_secs_f64() / self.progress.load(Ordering::SeqCst) as f64 ));
```
heaths: ?

doc/performance-test-requirements.md
LarryOsterman: Yeah, this is just preliminary info, and isn't formal.

sdk/core/azure_core_test/src/lib.rs
LarryOsterman: Agreed. And I *believe* that `Recording` is currently `Send`able after these changes (due to the use of `OnceLock` and not `OnceCell`).

doc/performance-test-requirements.md
LarryOsterman: The test-requirements.md file is mostly my scratch pad where I was starting my process of thinking about this. I was putting more accurate information in the `README.md` file.

sdk/core/azure_core_test/src/perf/README.md
LarryOsterman: As I mentioned, the version under docs is not canonical, README is canonical.

sdk/core/azure_core_test/src/perf/mod.rs
LarryOsterman: Unfortunately, using the `std::Result` turned into a mess because `Box<dyn Error>` doesn't implement Send, so...

sdk/core/azure_core_test/src/perf/mod.rs
LarryOsterman: That's much better than `\0` which is what I originally implemented.

sdk/core/azure_core_test/src/perf/mod.rs
LarryOsterman: That's for the stupidest of reasons - We use this to index into a vector of `TestContext` objects.

sdk/core/azure_core_test/src/perf/mod.rs
LarryOsterman: The underlying clap API required `Any` so I copied it over.

sdk/core/azure_core_test/src/perf/README.md
```diff
@@ -29,9 +29,9 @@ Each test has its own set of parameters which are specific to the test.
 
 Performance tests have three phases:
 
-1) Setup - Establish any resources needed to run the test.
-1) Run - Actually perform the test.
-1) Cleanup - Cleanup any resources used by the test.
+1.. Setup - Establish any resources needed to run the test.
+2.. Run - Actually perform the test.
+3.. Cleanup - Cleanup any resources used by the test.
```
heaths, line 34: You have an extra `.`. Should just be `1. Setup`, etc. Could get it in the next PR, though, if you want to get this in.

sdk/core/azure_core_test/src/perf/README.md
```diff
@@ -98,21 +119,34 @@ harness = false
...
-```rust
-#[tokio::main]
-async fn main() -> azure_core::Result<()> {
-    let runner = PerfRunner::new(
-        env!("CARGO_MANIFEST_DIR"),
-        file!(),
-        vec![GetSecrets::test_metadata()],
-    )?;
+```bash
+cargo test --package azure_storage_blob --test perf -- <performance test command line>
```
heaths, line 125: This is going to have the same problem with the `<`:


```suggestion
cargo test --package azure_storage_blob --test perf -- {performance test command line}
```

LarryOsterman: Fixing in next PR.

PR #3116, opened 2025-10-06
cosmos: Changelog prep for 0.28 release

analogrelay:
This PR preps the CHANGELOG.md for the 0.28 release. There are three separate changes here, which I put in separate commits, because the overall diff is a little busier than you might expect :).

1. Added a missing changelog entry for #2969. I forgot to ensure that was added before merging, so I put it in here.
2. Replaced our usage of `*` in the changelog with `-` so we're consistent with other SDKs, and consistent within the file (since a recent Core change added a changelog entry using `-`)
3. Marked the changelog entry for `0.28.0` as releasing tomorrow.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
+++ b/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 ### Features Added
 
-* Added `Query::with_text()` and `Query::append_text()` methods to modify query text after creation ([#3044](https://github.com/Azure/azure-sdk-for-rust/pull/3044))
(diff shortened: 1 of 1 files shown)


heaths (approved):
FWIW, you didn't have to change. It's only important to be consistent in the current CHANGELOG. Personally, I prefer `*` and use that myself because it looks more like "dots". Most of the existing crates just used `-` already so I just went with it and wince a little every time I see them.

analogrelay:
If you *do* want to do a bulk change from `-` to `*` at some point, it's a simple vim command that I'm happy to run and make one PR that does it across all crates ;). I really don't mind either way.

Pilchie (approved):
I recently heard a good argument in favor of `-` instead of `*`. Namely, some editors auto-insert a closing`*` assuming that the user is trying to *italic* or **bold**. With `-` there isn't that ambiguity, and so you don't get into having extra chars inserted.

PR #3128, opened 2025-10-07
Link get_token errors to azure_identity troubleshooting guide

### LINKED ISSUE #2677: Link azure_identity errors to troubleshooting guide

chlowell:
Closes #2677


sdk/identity/azure_identity/src/lib.rs
```diff
@@ -112,6 +112,32 @@ fn get_authority_host(env: Option<Env>, option: Option<String>) -> Result<Url> {
     Url::parse(&authority_host).map_err(Into::<Error>::into)
 }
 
+const TSG_LINK: &str = ". To troubleshoot, visit https://aka.ms/azsdk/rust/identity/troubleshoot#";
```
heaths: Nit: should we put the anchor in the text we concat? Feels like the right thing. It wouldn't hurt if we only had the link, but when you first see it in code - as I did above - that was the first thought I had, "oh, he forgot the anchor".

sdk/identity/azure_identity/src/lib.rs
```diff
@@ -112,6 +112,32 @@ fn get_authority_host(env: Option<Env>, option: Option<String>) -> Result<Url> {
...
+
+/// Map an error from a credential's get_token() method to an ErrorKind::Credential error, appending
+/// a link to the troubleshooting guide entry for that credential, if it has one.
+///
+/// TODO: decide whether to map to ErrorKind::Credential here (https://github.com/Azure/azure-sdk-for-rust/issues/3127)
+fn authentication_error<T: 'static>(e: azure_core::Error) -> azure_core::Error {
+    azure_core::Error::with_message_fn(e.kind().clone(), || {
+        let type_name = std::any::type_name::<T>();
+        let short_name = type_name.rsplit("::").next().unwrap_or(type_name); // cspell:ignore rsplit
```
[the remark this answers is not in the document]
heaths: And yet, it does. We don't control the standard rust dictionary, which probably doesn't include every method you could find even in `std`.

That said, we might add this to our global dictionary.

heaths, line 146: Rather than all this string allocation and the string pool this creates, you might consider something like this instead since traits can have constant expressions:

```rust
#![allow(dead_code)]

#[doc(hidden)]
pub trait ErrorDoc: private::Sealed {
    const ANCHOR: &'static str;
}

const TSG_LINK: &str = "https://example.com";

pub struct A;
impl ErrorDoc for A {
    const ANCHOR: &'static str = "a";
}

pub struct B;
impl ErrorDoc for B {
    const ANCHOR: &'static str = "b";
}

#[test]
fn test() {
    assert_eq!(example::<A>(), "https://example.com#a");
    assert_eq!(example::<B>(), "https://example.com#b");
}

fn example<T: ErrorDoc>() -> String {
    format!("{TSG_LINK}#{}", T::ANCHOR)
}

mod private {
    pub trait Sealed {}
    impl Sealed for super::A {}
    impl Sealed for super::B {}
}
```

sdk/identity/azure_identity/src/lib.rs
chlowell: sure, can't hurt and it makes this constant appropriate for a nonspecific link to the guide if we ever want one

sdk/identity/azure_identity/src/lib.rs
chlowell: Good idea, thank you

sdk/identity/azure_identity/src/lib.rs
```diff
@@ -112,26 +112,75 @@ fn get_authority_host(env: Option<Env>, option: Option<String>) -> Result<Url> {
...
-            "AzureCliCredential" => format!("{TSG_LINK}azd"),
-            "AzureDeveloperCliCredential" => format!("{TSG_LINK}azure-cli"),
-            "AzurePipelinesCredential" => format!("{TSG_LINK}apc"),
-            #[cfg(feature = "client_certificate")]
-            "ClientCertificateCredential" => format!("{TSG_LINK}client-cert"),
-            "ClientSecretCredential" => format!("{TSG_LINK}client-secret"),
-            "ManagedIdentityCredential" => format!("{TSG_LINK}managed-id"),
-            "WorkloadIdentityCredential" => format!("{TSG_LINK}workload"),
-            _ => "".to_string(),
+        let short_name = type_name.rsplit("::").next().unwrap_or(type_name);
```
heaths: Oh, shoot. I didn't realize originally you *also* needed the `short_name` for the message below. In that case, what you had was fine (though I'd still put the anchor in the format expression). I'll leave it up to you if you want to revert or not. Sorry I missed that.

heaths (approved):

sdk/identity/azure_identity/src/lib.rs
heaths: If nothing else, if you want to just get this out and change it later in another PR, that's fine too.

chlowell: I thought you were commenting on the match arms and was only a little surprised because I'm used to Rust being less than consistently intuitive 😆 

I'll revert to the less bloated string matching approach

PR #3141, opened 2025-10-07
Update core crate versions

heaths:


heaths:
We need to wait until after the other crates are released.

PR #3145, opened 2025-10-08
Revert "ESRP publishing (#3065)"

heaths:
This reverts commit 75b2e3440291d6129105176f2814a889b0a5665a.
Trying to release Key Vault in #3142, I was blocked by errors saying it references unreleased versions of `azure_core` and friends, which isn't true.
In `main`, I checked that indeed it references `azure_core@0.29.1` which was released yesterday and is up on crates.io. Something is wrong in the UnreleasedDependencies calculation or,
more likely because that didn't change, how we use/assume to use that information.

For now to unblock crates we need to release this week, I'm reverting.


danieljurek (approved):
Approved. Previous testing on this scenario did not reveal this result. More investigation needed.

Issue #3073, opened 2025-09-30
EventProcessor PartitionClient does not extract offset and sequence properly for checkpoints


arkadiuszspiewak:
### Bug Title

Offset and sequence number are not extracted for checkpoints

### Crate Name

azure_messaging_eventhubs

### Crate Version

0.7.0

### Description

It seems that `PartitionClient::update_checkpoint` function has a bug and compares improper keys causing offset and sequence to be none, and not being writtent to blob metadata.
```
{"timestamp":"2025-09-30T15:11:29.300093Z","level":"INFO","fields":{"message":"about to checkpoint seq=Some(10176) offset=Some(\"1421634174976\")"},"target":"reader::eventhub::consumer"}
{"timestamp":"2025-09-30T15:11:29.300119Z","level":"INFO","fields":{"message":"raw AMQP message: AmqpMessage { header: None, message_annotations: Some(AmqpAnnotations(AmqpOrderedMap { inner: [(Symbol(AmqpSymbol(\"x-opt-partition-key\")), String(\"1fa0cc6b-3e68-4d7a-8e58-a8fc002679e6\")), (Symbol(AmqpSymbol(\"x-opt-sequence-number-epoch\")), Int(-1)), (Symbol(AmqpSymbol(\"x-opt-sequence-number\")), Long(10176)), (Symbol(AmqpSymbol(\"x-opt-offset\")), String(\"1421634174976\")), (Symbol(AmqpSymbol(\"x-opt-enqueued-time\")), TimeStamp(AmqpTimestamp(Some(SystemTime { tv_sec: 1759245089, tv_nsec: 238000000 }))))] })), delivery_annotations: None, properties: None, footer: None, .. }"},"target":"reader::eventhub::consumer"}
{"timestamp":"2025-09-30T15:11:30.111093Z","level":"INFO","fields":{"message":"annotation Symbol(AmqpSymbol(\"x-opt-partition-key\")) => String(\"1fa0cc6b-3e68-4d7a-8e58-a8fc002679e6\")"},"target":"azure_messaging_eventhubs::event_processor::partition_client"}
{"timestamp":"2025-09-30T15:11:30.111125Z","level":"INFO","fields":{"message":"annotation Symbol(AmqpSymbol(\"x-opt-sequence-number-epoch\")) => Int(-1)"},"target":"azure_messaging_eventhubs::event_processor::partition_client"}
[...]

PR #3148, opened 2025-10-08
Getting offset and sequence number from event data rather than amqp annotations.

j7nw4r:
This PR addresses: https://github.com/Azure/azure-sdk-for-rust/issues/3073

Properly retrieve the offset and sequence numbers to properly store checkpoints.


sdk/eventhubs/azure_messaging_eventhubs/src/event_processor/partition_client.rs
```diff
@@ -127,58 +127,15 @@ impl PartitionClient {
...
-                consumer_group: self.client_details.consumer_group.clone(),
-                partition_id: self.partition_id.clone(),
-                offset,
-                sequence_number,
-            };
-            self.checkpoint_store.update_checkpoint(checkpoint).await?;
-        }
-        Ok(())
+        let checkpoint = Checkpoint {
+            fully_qualified_namespace: self.client_details.fully_qualified_namespace.clone(),
```
LarryOsterman, line 159: I don't know if it matters, but the reason I didn't use the details fields is that the call to `offset()` and `sequence_number()` iterates over the application properties, so you're iterating over the data twice.

Not sure it's huge deal though

sdk/eventhubs/azure_messaging_eventhubs/src/event_processor/partition_client.rs
```diff
@@ -127,58 +127,15 @@ impl PartitionClient {
     /// # Errors
     /// Returns an error if the sequence number or offset is invalid, or if updating the checkpoint fails.
     pub async fn update_checkpoint(&self, event_data: &ReceivedEventData) -> Result<()> {
-        let mut sequence_number: Option<i64> = None;
-        let mut offset: Option<String> = None;
-
-        let amqp_message = event_data.raw_amqp_message();
-        if let Some(message_annotations) = &amqp_message.message_annotations {
-            for (key, value) in message_annotations.0.iter() {
-                if *key == crate::consumer::SEQUENCE_NUMBER_ANNOTATION {
```
LarryOsterman, line 136: Remove these definitions?

j7nw4r: The SEQUENCE_NUMBER_ANNOTATION and OFFSET_ANNOTATION still seem to be used in the consumer.

j7nw4r:
@LarryOsterman Please review. I went back to a structure similar to the original. I updated the annotation labels used to use the same labels as the `.offset()` and `.sequence_number()` functions. I also cleaned up the logic to make things more readable (imo).

sdk/eventhubs/azure_messaging_eventhubs/src/event_processor/partition_client.rs
j7nw4r: I went back to a structure similar to what you had previously. Single for loop comparing the strings. You're right not a huge deal, but I wanted to be correct.

sdk/eventhubs/azure_messaging_eventhubs/src/event_processor/partition_client.rs
LarryOsterman: Oh, interesting. For some reason I thought this was in the checkpoint store (no idea why).

Oh - that's the confusion. The message property doesn't match the filter property. I got confused by the names since they're so similar.

LarryOsterman (approved):

sdk/eventhubs/azure_messaging_eventhubs/src/event_processor/partition_client.rs
```diff
@@ -127,58 +129,41 @@ impl PartitionClient {
...
+                let AmqpValue::String(offset_value) = value else {
+                    continue;
+                };
+                offset_option = Some(offset_value.clone());
+            } else if *symbol == "x-opt-sequence-number" {
+                let AmqpValue::Long(sequence_number_value) = value else {
+                    continue;
+                };
+                sequence_number_option = Some(*sequence_number_value);
+            }
```
heaths, line 155: Tip for next time: it's more idiomatic in rust to use `match`. Cleaner as well (less back and forth with "tabs").

j7nw4r: Yea I tried that originally.

In rust, structural patterns, like this one because symbol is an AmqpSymbol, use pattern matching and not PartialEq. So the compiler complains.

heaths: Yeah, that actually makes sense now that I think about it more. It's relying on the `Deref` impl in this case.

PR #3161, opened 2025-10-09
Add append_path() for Url

antkmsft:
Usage in codegen: https://github.com/Azure/typespec-rust/pull/635

### DIFF
--- a/sdk/core/typespec_client_core/src/http/mod.rs
+++ b/sdk/core/typespec_client_core/src/http/mod.rs
@@ -97,7 +97,7 @@ impl UrlExt for Url {
         debug_assert_eq!(new_path.capacity(), new_len);
         new_path.push_str(self.path());
         if needs_separator {
-            new_path.push_str("/");
+            new_path.push('/');
         }
         new_path.push_str(path);
         debug_assert_eq!(new_path.capacity(), new_len);


sdk/typespec/typespec_client_core/src/http/mod.rs
```diff
@@ -59,3 +59,149 @@ where
...
+    fn append_path(&mut self, p: &str) {
+        if self.path().len() == 1 {
+            self.set_path(p);
+        } else if !p.is_empty() && p != "/" {
+            match if self.path().ends_with('/') { 1 } else { 0 }
+                + if p.starts_with('/') { 1 } else { 0 }
+            {
+                0 => self.set_path(&format!("{}/{}", self.path(), p).to_string()),
+                1 => self.set_path(&(self.path().to_owned() + p)),
+                _ => self.set_path(&(self.path()[..self.path().len() - 1].to_owned() + p)),
```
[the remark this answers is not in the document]
antkmsft: I think it is an overkill, with its own disadvantages.

sdk/typespec/typespec_client_core/src/http/mod.rs
```diff
@@ -59,3 +59,149 @@ where
...
+            }
+        }
+    }
+}
+
+#[test]
+fn test_url_append_path() {
```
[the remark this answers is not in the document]
antkmsft: done.

sdk/typespec/typespec_client_core/src/http/mod.rs
```diff
@@ -59,3 +59,154 @@ where
         }
     }
 }
+
+/// Extension trait for `Url` to provide additional URL manipulation methods.
```
heaths: Use links when it makes sense but typically only the first mention for a particular type, member. We also don't link to ourselves i.e., a member of type `Foo` shouldn't use ``[`Foo`]`` as it's redundant.

```suggestion
/// Extension trait for [`Url`] to provide additional URL manipulation methods.
```

sdk/typespec/typespec_client_core/src/http/mod.rs
```diff
@@ -59,3 +59,154 @@ where
...
+pub trait UrlExt {
```
heaths: Since this is public, we should probably seal it so no one else implements it.

The way I like to do it is in `src/lib.rs`, add:

```rust
mod private {
  pub trait Sealed {}
  impl Sealed for crate::http::Url;
}
```

Then your trait changes to:

```suggestion
pub trait UrlExt: crate::private::Sealed {
```

And no one else can implement it. That `mod private` in the root module is more like `pub(crate)` but without having to specify it since all descendent modules can view ancestors' private members.

sdk/typespec/typespec_client_core/src/http/mod.rs
```diff
@@ -59,3 +59,154 @@ where
...
+    /// Appends a path segment to the URL's path, handling slashes appropriately and preserving query parameters.
+    fn append_path(&mut self, p: &str);
```
heaths: Make it easier to use and be explicit with parameter names so it's obvious:

```suggestion
    fn append_path(&mut self, path: impl AsRef<str>);
```

sdk/typespec/typespec_client_core/src/http/mod.rs
```diff
@@ -59,3 +59,154 @@ where
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn url_append_path() {
+        {
```
heaths: Tip: you don't need these in separate blocks. You can keep reassigning to `let mut url` in Rust. You can even change the type. Latest variable binding always wins. I'd still put a space in between blocks, but this is otherwise unnecessary (good in some rare cases like you want to better manage lifetime within a function).

sdk/typespec/typespec_client_core/src/http/mod.rs
```diff
@@ -59,3 +59,154 @@ where
...
+    fn append_path(&mut self, p: &str);
+}
+
+impl UrlExt for Url {
+    fn append_path(&mut self, p: &str) {
+        if self.path() == "/" {
+            self.set_path(p);
+        } else if !p.is_empty() && p != "/" {
+            match if self.path().ends_with('/') { 1 } else { 0 }
+                + if p.starts_with('/') { 1 } else { 0 }
```
heaths: This makes debugging harder. Put then on separate lines so people could step through, if necessary:

```suggestion
            let mut combinator = self.path().ends_with('/') { 1 } else { 0 };
            combinator += if p.starts_with('/') { 1 } else { 0 };
            match combinator {
```

The compiler is effectively going to anyway. Make it easier for people.

sdk/core/typespec_client_core/src/http/mod.rs
```diff
@@ -59,3 +59,154 @@ where
         }
     }
 }
+
+/// Extension trait for `Url` to provide additional URL manipulation methods.
+pub trait UrlExt {
+    /// Appends a path segment to the URL's path, handling slashes appropriately and preserving query parameters.
```
heaths, line 65: I would expand on what "appropriately" means here. Frankly, I find it odd that appending paths "alpha" and "beta" as one test does below - no slashes in input - yields ".../alpha/beta". It's also a bit strange how the trailing "/" is based on input but always added between segments.

I appreciate we need these quirks for what apparently isn't very strict TypeSpec output, but my main intention here is that we at least document that there *are quirks*, not necessarily what they are all.

In that vein, I recommend a few doc tests like the one above I just mentioned *in the docs* themselves. If you want to repeat them below, fine, but no need since doc tests also run as part of `cargo test`. Not only do they test, but they show the dev what they might expect e.g.,


```suggestion
    /// Appends a path segment to the URL's path, handling slashes appropriately and preserving query parameters.
    ///
    /// This always assume the existing URL terminates with a directory, and the `path` you pass in is a separate directory or file segment.
    ///
    /// # Examples
    ///
    /// ```
    /// use typespec_client_core::{Url, UrlExt as _};
    ///
    /// let mut url: Url = "https://contoso.com/foo?a=1".parse().unwrap();
    /// url.append_path("bar");
    /// assert_eq!(url.as_str(), "https://contoso.com/foo/bar?a=1");
    /// ```
```

sdk/typespec/typespec_client_core/src/http/mod.rs
```diff
@@ -59,3 +59,154 @@ where
...
+            {
+                0 => self.set_path(&format!("{}/{}", self.path(), p)),
+                1 => self.set_path(&(self.path().to_owned() + p)),
+                _ => self.set_path(&(self.path()[..self.path().len() - 1].to_owned() + p)),
```
heaths: Rather than creating an owned string just to expand it with `+ p`, just use `format!()` which only requires a ref and should efficiently allocate a single owned string. This is going to be a hot path, so we want to be mindful.

heaths: Try this: if there's enough capacity, `b` is still `realloc`'d with the same address, but realloc'd nonetheless.

```rust
fn main() {
    let foo = "foo";
    let bar = "bar";
    
    let a = format!("{foo}/{bar}");
    println!("&a = {:p}, a.capacity() = {}", &a, a.capacity());
    
    let mut b = foo.to_owned();
    println!("&b = {:p}, before b.capacity() = {}", &b, b.capacity());
    b += "p";
    println!("&b = {:p}, after b.capacity() = {}", &b, b.capacity());
}
```

sdk/core/typespec_client_core/src/http/mod.rs
```diff
@@ -60,19 +60,32 @@ where
     }
 }
 
-/// Extension trait for `Url` to provide additional URL manipulation methods.
-pub trait UrlExt {
+/// Extension trait for [`Url`] to provide additional URL manipulation methods.
+pub trait UrlExt: crate::private::Sealed {
```
heaths, line 64: This doesn't exist.

sdk/typespec/typespec_client_core/src/http/mod.rs
```diff
@@ -60,19 +60,32 @@ where
...
         if self.path() == "/" {
             self.set_path(p);
         } else if !p.is_empty() && p != "/" {
-            match if self.path().ends_with('/') { 1 } else { 0 }
-                + if p.starts_with('/') { 1 } else { 0 }
+            let mut combinator = self.path().ends_with('/') { 1 } else { 0 };
+            combinator += if p.starts_with('/') { 1 } else { 0 };
+            match combinator {
             {
                 0 => self.set_path(&format!("{}/{}", self.path(), p)),
```
heaths: Based on #3190 let's not abuse `format!()`. Concatenate paths as efficiently as possible.

sdk/core/typespec_client_core/src/http/mod.rs
antkmsft: Sorry, I merged your commits knowing they're incomplete, just for the proper attribution, and then in a few minutes made changes locally so that everything compiles. You must've came to review this PR while it was in broken state.

sdk/typespec/typespec_client_core/src/http/mod.rs
antkmsft: I'm sorry, I want to make sure I understand what is the most effective way - the PR tests slightly different thing - it uses join(), but join() would drop query parameters, so we would need to re-apply query parameters somehow, and getting a hashmap of query parameters and reapplying them is probably not the most effective thing, as we talked in https://github.com/Azure/typespec-rust/pull/635/#discussion_r2414995416 .

You initially proposed format instead of `+` in the comment above (https://github.com/Azure/azure-sdk-for-rust/pull/3161#discussion_r2421674607), but `+=` would be even better, did I understand correctly? So I will rewrite all the concatenation statements to use `+=`.

antkmsft: Please see the latest iteration, is this what we want?

sdk/core/typespec_client_core/src/http/mod.rs
```diff
@@ -59,3 +59,148 @@ where
         }
     }
 }
+
+/// Extension trait for [`Url`] to provide additional URL manipulation methods.
+pub trait UrlExt: crate::private::Sealed {
+    /// Appends a path segment to the URL's path, handling slashes appropriately and preserving query parameters.
+    ///
+    /// This always assume the existing URL terminates with a directory, and the `path` you pass in is a separate directory or file segment.
```
RickWinter: Why do we make an assumption on the format? It would be better if we followed the RFC for Url format. Also this method should not do special things for TypeSpec, those special cases should be handled by the emitter.

heaths: I told him to add a helper method because it's easier to update, but I don't like the special "trailing /" magic. Adding an `append_path` or something where it adds a segment and any subsequent segments assume the last segment is a directory - basically what other URL APIs do in other languages - but not this.

If we have helpers in core - which I've opend tracking issues for before like better query parameter management - it makes it easier to update crates in the future without having to regenerate. We want to strike a balance between helpers in core and doing everything in the emitter. Imagine if we made some perf improvements in helpers to reduce string manipulation. You'd want to regenerate and re-release every crate just for that?

jhendrixMSFT: The trailing `/` magic is an artifact of how [`Url::join`](https://docs.rs/url/latest/url/struct.Url.html#method.join) works and we had to jump through some hoops in the emitter to deal with it. This should make things simpler.

antkmsft: @jhendrixMSFT, I am glad this would make things easier - `Url::join()` must have two artifacts that I would imagine would require us to do extra work when joining:
1. `"http://example.com/a/b" join "c"` => `"http://example.com/a/c"`
2. `"http://example.com/a/b" join "/c"` => `"http://example.com/c"`

BTW, in my codegen PR, I was able to replace join with this, and everything works, but if you know of any places in code that we can simplify now, and drop the artifact avoidance logic, please let me know.

heaths: Are you saying that's what `Url::join` does or we do? Because 1 should end up "/a/b/c" and 2 is correct. That's how .NET's `Uri` works, and other languages from what I've seen. URLs should make any assumptions about what the last path segment represents. I mean, really, even in the world of URIs born out of UNIX they could be either. It's the service that decides what it is and the client should make any assumptions.

antkmsft: 2 is what `url::join()` will do. I'll update this PR for us to do the same.

jhendrixMSFT: While the behavior of `Url` is correct, the reality is that it causes friction with respect to the provided data.
- the endpoint might be more than just a host, e.g. `https://host.com/some/path`
- the endpoint might contain query params
- the per-operation path data always starts with a forward slash

What would be ideal is that the helper in `azure_core` properly copes with the above, so that all we need to do is call `endpoint.join(path)` in the emitted code and the right thing happens.

antkmsft: @jhendrixMSFT, so you are proposing `"http://example.com/a/b" append_path "/c" => "http://example.com/a/b/c"`, as we had before? I agree, this is more convenient from the emitter perspective. I haven't seen a context in the emitter in which `/c` would mean "abandon existing path, use new path `/c`". I'll revert the current commit that I just added.

jhendrixMSFT: From the emitter perspective, this does simplify things. However, I believe we've already solved these problems in the emitter.

To take a step back, what is the problem we're trying to solve? If it's consolidating what we do in the emitter into a helper, then I think that's good (i.e. if we have bugs, then all we need to do is fix them in core instead of the emitter then regenerate things). Or is there more than that? Is something currently broken?

antkmsft: It basically started as a response to this issue - https://github.com/Azure/typespec-rust/issues/531 - we added some extra code that codegen would generate in order to not drop query parameters when joining URLs. Then the next logical step that followed from that PR (https://github.com/Azure/typespec-rust/pull/635) is to make it as effective as possible, and to move it into the SDK, so it can be updated/fixed in one single place without the need to regenerate.

I think, we can work with any behavior, as long as the contract is clear, and you can see this contract in unit tests in this PR. From a practical standpoint, not overwriting the path when the thing you append starts with a slash would be more common with the emitter, and this code is more for the emitter than general use. If someone needs to drop the path if the new one starts with slash, they could use set_path() directly. Or, we could add an option to this function. Or create another one - after all, this one is called "append", not "smart append and maybe replace".

I think there are two possible aproaches to URL APIs - high level and low level. Overwriting path if it starts with "/" is more of a high level approach. But then, I would argue, for the general case it also makes sense to drop query parameters by default, because in `url.com/foo?x=y` the parameters were for `foo`. Once you appended "bar", it is less likely the parameters should be the same. And this is not the API that we needed.
We needed low-level APIs, which manipulates URLs more like strings, like text. We didn't need all that smartness when joining, we are saying that we know what we are doing, and when we say "append_path" we mean literally that - not replace, not drop query parameters. If we need that, we'll do that explicitly. The only "smart" thing we could use for our convenience is to make sure thaere is a slash, and exactly one slash when joining new path to an existing path.

antkmsft: I am going to revert to previous commit, but you can see, any approach is possible. I am slightly in favor of the non-overwrite approach, as it is more useful for the emitter.

jhendrixMSFT: Thanks for clarifying.

To make sure I understand, the API will join paths based on our common use case (i.e. _not_ option 2 like [here](https://github.com/Azure/azure-sdk-for-rust/pull/3161#discussion_r2434070822)) while preserving query parameters?

antkmsft: @jhendrixMSFT, yes. Always preserves query params, `http://url.com/foo?x=y` append `/bar` is `http://url.com/foo/bar?x=y`.
I think unit test included in this PR is the most concentrated form of explanation I can give, but of course I am glad to clarify anything here in the comments if it is more convenient.

heaths:
> Revert "overwrite if new path starts with slash"

Then why does https://github.com/Azure/typespec-rust/pull/635 change all the paths to start with a preceding slash?

I'm not opposed to a helper in core. I suggested something like this (and query param helpers) a while back to @jhendrixMSFT. I've suggested helpers for other things more recently. And I appreciate that `url::Url::join` has some questionable semantics when it comes to combining paths - like making a distinction between directories and files (in a URL, there is no such thing nor any assumption that should be made) - but let's not create our own interesting behaviors. IMO, .NET's `Uri` has the right idea: when adding a path segment, you always delimit with a path. Full stop. Make no assumptions. If separating slash is needed, add it. If one is already there, don't add it. If there are slashes in the path you're appending, keep them. It's just an opaque string.

sdk/core/typespec_client_core/src/http/mod.rs
```diff
@@ -59,3 +59,148 @@ where
...
+            combinator += if p.as_ref().starts_with('/') { 1 } else { 0 };
+
+            let mut new_path = if combinator < 2 {
+                self.path().to_owned()
+            } else {
+                self.path()[..self.path().len() - 1].to_owned()
+            };
+            if combinator == 0 {
+                new_path.push('/');
+            }
```
heaths: Why are you trimming it just to re-add it?

sdk/core/typespec_client_core/src/http/mod.rs
```diff
@@ -59,3 +59,148 @@ where
...
+            new_path.push_str(p.as_ref());
+
+            self.set_path(&new_path);
+        }
```
heaths, line 90: This is going to realloc a couple times when you know everything you need to know up front. A more efficient and cleaner design would be:


```suggestion
        let path = p.as_ref().trim_start_matches('/');
        if self.path() == "/" {
            self.set_path(path);
            return;
        }
        if path.is_empty() {
            return;
        }
        let needs_separator = !self.path().ends_with('/');
        let mut new_len = self.path().len() + path.len();
        if needs_separator {
            new_len += 1;
        }
        let mut new_path = String::with_capacity(new_len);
        debug_assert_eq!(new_path.capacity(), new_len);
        new_path.push_str(self.path());
        if needs_separator {
            new_path.push_str("/");
        }
        new_path.push_str(path);
        debug_assert_eq!(new_path.capacity(), new_len);

        self.set_path(&new_path);
```

Only one allocation - at least per `append_path`. Ideally we can limit that, but `String` will grow fairly efficiently. We can always tweak in the future.

heaths (changes requested):
Better, but can be more efficient for now (see code suggestion). Before we GA, we should actually look at splitting a path in TypeSpec and building up segments more efficiently using `Url::path_segments_mut()`. We can then in the emitter efficiently add fixed string segments like "containers" and just append variables to be replaced with the name instead of a `String::replace` which is terribly inefficient to do repeatedly. The String grows more often and by smaller amounts than probably necessary if more is known up front.

In some code we'll run in tight loops like with Storage Blobs, we're going to be building up endpoints quickly in a tight loop so we'll want to optimize where we can. We can always get better, though, so this doesn't haven't to be perfect. I'd just consider my suggestion below. IMO, it's more intuitive and certainly has fewer allocations. You're will have at least 2 but likely more.

sdk/core/typespec_client_core/src/http/mod.rs
heaths: I ran that in the playground with your tests copied over. All passed.

sdk/core/typespec_client_core/src/http/mod.rs
antkmsft: Thank you, I took the code you suggested.

PR #3166, opened 2025-10-09
cosmos: Allow the Query Pipeline to return an alternative query to execute in each request

analogrelay:
For Hybrid Search, as well as ReadMany support, the query engine will need to direct the SDK to issue **multiple different queries**. However, the Rust SDK's query executor expects to be able to replay the original query on each request. This PR changes that expectation to allow (but not require) each Query Request coming from the pipeline to provide an alternate query. When an alternate query is provided, the pipeline can also indicate if the parameters from the original query should be sent. Often, the alternate query does NOT use the parameters, but in some circumstances (hybrid search) it will.

There are also some small fixes for doc tests and error checking in the engine-less cross-partition query tests. Not sure why these are coming up here and not in `main`.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
+++ b/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+- Adjusted the query engine abstraction to support future enhancements and optimizations. ([#3166](https://github.com/Azure/azure-sdk-for-rust/pull/3166))
+
 ### Breaking Changes
 
 ### Bugs Fixed


analogrelay:
Can do! Sorry about the delay, got pulled on to high-pri stuff ;).

tvaron3 (approved):
LGTM

sdk/cosmos/azure_data_cosmos/src/query/engine.rs
```diff
@@ -3,14 +3,42 @@ pub struct QueryRequest {
     /// The ID of the partition key range to query.
     pub partition_key_range_id: String,
 
+    /// The index of this request, within the partition key range.
```
tvaron3, line 6: Would we want to call out this change in the changelog ?

analogrelay: I don't think so, it's a purely internal API

analogrelay: I guess it's technically public 🤔 . Yeah, I will actually add it to the changelog. Thanks for the reminder!

PR #3174, opened 2025-10-09
Move typespec into sdk/core

heaths:
Because typespec is just an implementation detail of core,
and to help simplify upcoming eng-sys changes that should make publishing easier,
move all the typespec crates under sdk/core.


sdk/core/ci.yml
```diff
@@ -15,6 +15,12 @@ extends:
   parameters:
     ServiceDirectory: core
     Artifacts:
+    - name: typespec
+      safeName: Typespec
+    - name: typespec_macros
+      safeName: TypespecMacros
+    - name: typespec_client_core
+      safeName: TypespecClientCore
```
danieljurek, line 23: `safeName` isn't really required. I can get rid of it in my release PR when I rebase on this.

danieljurek: Nevermind. It is used in archetype-rust-release.yml. Leave this as is. I'll account for it in my PR.

.vscode/cspell.json
```diff
@@ -232,4 +232,4 @@
       ]
     }
   ]
-}
+}
```
RickWinter, line 235: Nit: newline

RickWinter (approved):

LarryOsterman (approved):

.vscode/cspell.json
heaths: I think the cSpell VSCode extension itself doesn't add this if you `Ctrl+.` to add a word. We keep fighting it back and forth. I do have in our `.editorconfig` that all files get a final newline. Not sure why this happens but don't want to hold this up any longer.

PR #3176, opened 2025-10-10
[Storage] Flattened Client Refactor

vincenttran-msft:

- Uses our new TypeSpec definition for the underlying generated code, which now uses a flattened client structure rather than a hierarchical client structure
- Refactors all convenience layer clients to be compatible with the new generated code
- As a result of the refactor, we now support SAS URL support
- Added `from_url` methods to support client construction directly from URLs, as well as specifically URLs already containing SAS parameters

### DIFF
--- a/eng/emitter-package.json
+++ b/eng/emitter-package.json
@@ -4,14 +4,14 @@
     "@azure-tools/typespec-rust": "0.24.1"
   },
   "devDependencies": {
-    "@azure-tools/typespec-azure-core": "0.61",
-    "@azure-tools/typespec-azure-rulesets": "0.61",
+    "@azure-tools/typespec-azure-core": "~0.61",
+    "@azure-tools/typespec-azure-rulesets": "~0.61",
     "@azure-tools/typespec-client-generator-core": ">=0.61.1",
-    "@typespec/compiler": "1.5.0",
-    "@typespec/http": "1.5.0",
-    "@typespec/openapi": "1.5.0",
-    "@typespec/rest": "0.75.0",
-    "@typespec/versioning": "0.75.0",
-    "@typespec/xml": "0.75.0"
+    "@typespec/compiler": "^1.4.0",
+    "@typespec/http": "^1.4.0",
+    "@typespec/openapi": "^1.4.0",
+    "@typespec/rest": "~0.75.0",
+    "@typespec/versioning": "~0.75.0",
+    "@typespec/xml": "~0.75.0"
   }
 }
--- a/sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
+++ b/sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
@@ -148,6 +148,7 @@ impl BlobContainerClient {
         let mut blob_url = self.url().clone();
         blob_url
             .path_segments_mut()
+            // This should not fail as container URL has already been validated on client construction.
             .expect("Invalid endpoint URL: Cannot append blob_name to the blob endpoint.")
             .extend([blob_name]);
 
--- a/sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
+++ b/sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
@@ -112,7 +112,8 @@ impl BlobServiceClient {
         let mut container_url = self.url().clone();
         container_url
             .path_segments_mut()
-            .expect("Cannot be base")
+            // This should not fail as service URL has already been validated on client construction.
+            .expect("Cannot be a base URL.")
             .push(container_name);
 
         let client = GeneratedBlobContainerClient {
--- a/eng/emitter-package-lock.json
+++ b/eng/emitter-package-lock.json
@@ -8,15 +8,15 @@
         "@azure-tools/typespec-rust": "0.24.1"
       },
       "devDependencies": {
-        "@azure-tools/typespec-azure-core": "0.61",
-        "@azure-tools/typespec-azure-rulesets": "0.61",
+        "@azure-tools/typespec-azure-core": "~0.61",
+        "@azure-tools/typespec-azure-rulesets": "~0.61",
         "@azure-tools/typespec-client-generator-core": ">=0.61.1",
-        "@typespec/compiler": "1.5.0",
-        "@typespec/http": "1.5.0",
-        "@typespec/openapi": "1.5.0",
-        "@typespec/rest": "0.75.0",
-        "@typespec/versioning": "0.75.0",
-        "@typespec/xml": "0.75.0"
+        "@typespec/compiler": "^1.4.0",
+        "@typespec/http": "^1.4.0",
+        "@typespec/openapi": "^1.4.0",
+        "@typespec/rest": "~0.75.0",
+        "@typespec/versioning": "~0.75.0",
+        "@typespec/xml": "~0.75.0"
       }
     },
     "node_modules/@azure-tools/async-io": {
@@ -118,9 +118,9 @@
       }
     },
     "node_modules/@azure-tools/typespec-client-generator-core": {
-      "version": "0.61.1",
-      "resolved": "https://registry.npmjs.org/@azure-tools/typespec-client-generator-core/-/typespec-client-generator-core-0.61.1.tgz",
-      "integrity": "sha512-CLKPYiJMlF/tKNvf2F625WsjyYBAVFyDmR1fPXP59KceTPTXsYY/MgCgB5k7l9mYGw3uf2WNwo3Mp91yds4Ojw==",
+      "version": "0.61.2",
+      "resolved": "https://registry.npmjs.org/@azure-tools/typespec-client-generator-core/-/typespec-client-generator-core-0.61.2.tgz",
+      "integrity": "sha512-q+fktD/LZjiKYZbQHzWOMZd5Zqdr7PU0Cn9GEMkwv5gIvCPhT2Qxpgig7lPnMVYFwGmc+2qBZk1FjJPin0DUrw==",
       "license": "MIT",
       "dependencies": {
         "change-case": "~5.4.4",
@@ -181,9 +181,9 @@
       }
     },
     "node_modules/@babel/helper-validator-identifier": {
-      "version": "7.27.1",
-      "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.27.1.tgz",
-      "integrity": "sha512-D2hP9eA+Sqx1kBZgzxZh0y1trbuU+JoDkiEwqhQ36nodYqJwyEIhPSdMNd7lOm/4io72luTPWH20Yda0xOuUow==",
+      "version": "7.28.5",
+      "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.28.5.tgz",
+      "integrity": "sha512-qSs4ifwzKJSV39ucNjsvc6WVHs6b7S03sOh2OcHF9UHfVPqWWALUsNUVzhSBiItjRZoLHx7nIarVjqKVusUZ1Q==",
       "license": "MIT",
       "engines": {
         "node": ">=6.9.0"
@@ -589,9 +589,9 @@
       }
     },
     "node_modules/@types/turndown": {
-      "version": "5.0.5",
-      "resolved": "https://registry.npmjs.org/@types/turndown/-/turndown-5.0.5.tgz",
-      "integrity": "sha512-TL2IgGgc7B5j78rIccBtlYAnkuv8nUQqhQc+DSYV5j9Be9XOcm/SKOVRuA47xAVI3680Tk9B1d8flK2GWT2+4w==",
+      "version": "5.0.6",
+      "resolved": "https://registry.npmjs.org/@types/turndown/-/turndown-5.0.6.tgz",
+      "integrity": "sha512-ru00MoyeeouE5BX4gRL+6m/BsDfbRayOskWqUvh7CLGW+UXxHQItqALa38kKnOiZPqJrtzJUgAC2+F0rL1S4Pg==",
       "license": "MIT"
     },
     "node_modules/@typespec/compiler": {
@@ -1572,9 +1572,9 @@
       }
     },
     "node_modules/turndown": {
-      "version": "7.2.1",
-      "resolved": "https://registry.npmjs.org/turndown/-/turndown-7.2.1.tgz",
-      "integrity": "sha512-7YiPJw6rLClQL3oUKN3KgMaXeJJ2lAyZItclgKDurqnH61so4k4IH/qwmMva0zpuJc/FhRExBBnk7EbeFANlgQ==",
+      "version": "7.2.2",
+      "resolved": "https://registry.npmjs.org/turndown/-/turndown-7.2.2.tgz",
+      "integrity": "sha512-1F7db8BiExOKxjSMU2b7if62D/XOyQyZbPKq/nUwopfgnHlqXHqQ0lvfUTeUIr1lZJzOPFn43dODyMSIfvWRKQ==",
       "license": "MIT",
       "dependencies": {
         "@mixmark-io/domino": "^2.2.0"
--- a/sdk/storage/azure_storage_blob/src/generated/clients/append_blob_client.rs
+++ b/sdk/storage/azure_storage_blob/src/generated/clients/append_blob_client.rs
@@ -60,7 +60,7 @@ impl AppendBlobClient {
         if !endpoint.scheme().starts_with("http") {
             return Err(azure_core::Error::with_message(
                 azure_core::error::ErrorKind::Other,
-                format!("{endpoint} must use https"),
+                format!("{endpoint} must use http(s)"),
             ));
         }
         let auth_policy: Arc<dyn Policy> = Arc::new(BearerTokenCredentialPolicy::new(
--- a/sdk/storage/azure_storage_blob/src/generated/clients/blob_client.rs
+++ b/sdk/storage/azure_storage_blob/src/generated/clients/blob_client.rs
@@ -71,7 +71,7 @@ impl BlobClient {
         if !endpoint.scheme().starts_with("http") {
             return Err(azure_core::Error::with_message(
                 azure_core::error::ErrorKind::Other,
-                format!("{endpoint} must use https"),
+                format!("{endpoint} must use http(s)"),
             ));
         }
         let auth_policy: Arc<dyn Policy> = Arc::new(BearerTokenCredentialPolicy::new(
--- a/sdk/storage/azure_storage_blob/src/generated/clients/blob_container_client.rs
+++ b/sdk/storage/azure_storage_blob/src/generated/clients/blob_container_client.rs
@@ -22,7 +22,6 @@ use crate::generated::models::{
     SignedIdentifier,
 };
 use azure_core::{
-    async_runtime::get_async_runtime,
     credentials::TokenCredential,
     error::CheckSuccessOptions,
     fmt::SafeDebug,
@@ -73,7 +72,7 @@ impl BlobContainerClient {
         if !endpoint.scheme().starts_with("http") {
             return Err(azure_core::Error::with_message(
                 azure_core::error::ErrorKind::Other,
-                format!("{endpoint} must use https"),
+                format!("{endpoint} must use http(s)"),
             ));
         }
         let auth_policy: Arc<dyn Policy> = Arc::new(BearerTokenCredentialPolicy::new(
@@ -873,8 +872,6 @@ impl BlobContainerClient {
                             }),
                         )
                         .await?;
-                    // Because serialization can be a CPU intensive operation, yield the CPU before executing it..
-                    get_async_runtime().yield_now().await;
                     let (status, headers, body) = rsp.deconstruct();
                     let res: ListBlobsFlatSegmentResponse = xml::from_xml(&body)?;
                     let rsp = RawResponse::from_bytes(status, headers, body).into();
--- a/sdk/storage/azure_storage_blob/src/generated/clients/blob_service_client.rs
+++ b/sdk/storage/azure_storage_blob/src/generated/clients/blob_service_client.rs
@@ -61,7 +61,7 @@ impl BlobServiceClient {
         if !endpoint.scheme().starts_with("http") {
             return Err(azure_core::Error::with_message(
                 azure_core::error::ErrorKind::Other,
-                format!("{endpoint} must use https"),
+                format!("{endpoint} must use http(s)"),
             ));
         }
         let auth_policy: Arc<dyn Policy> = Arc::new(BearerTokenCredentialPolicy::new(
--- a/sdk/storage/azure_storage_blob/src/generated/clients/block_blob_client.rs
+++ b/sdk/storage/azure_storage_blob/src/generated/clients/block_blob_client.rs
@@ -63,7 +63,7 @@ impl BlockBlobClient {
         if !endpoint.scheme().starts_with("http") {
             return Err(azure_core::Error::with_message(
                 azure_core::error::ErrorKind::Other,
-                format!("{endpoint} must use https"),
+                format!("{endpoint} must use http(s)"),
             ));
         }
         let auth_policy: Arc<dyn Policy> = Arc::new(BearerTokenCredentialPolicy::new(
--- a/sdk/storage/azure_storage_blob/src/generated/clients/page_blob_client.rs
+++ b/sdk/storage/azure_storage_blob/src/generated/clients/page_blob_client.rs
@@ -65,7 +65,7 @@ impl PageBlobClient {
         if !endpoint.scheme().starts_with("http") {
             return Err(azure_core::Error::with_message(
                 azure_core::error::ErrorKind::Other,
-                format!("{endpoint} must use https"),
+                format!("{endpoint} must use http(s)"),
             ));
         }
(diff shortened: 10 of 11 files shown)


sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/examples/checkpoint_store_basic.rs
```diff
@@ -46,8 +46,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Instantiate a blob client with OpenTelemetry instrumentation enabled
     let blob_container_client = BlobContainerClient::new(
         &storage_account_url,
-        container,
-        credential,
+        &container,
+        Some(credential),
```
heaths, line 50: If it's optional, it needs to go into the options bag. This is in our guidelines. If this is for a URL that contains a SAS, perhaps we need to think about different construction techniques. It's not idiomatic in Rust to have runtime failures we can catch more easily at dev time. We discussed before, following our guidelines, of having a `with_credential()` or `with_sas_token()` or something constructor, depending on which is less prevalent, leaving `new` for the mainline case. Given SFI, I think that should be credential.

Since `credential` would not be optional, one has to be passed.

sdk/storage/azure_storage_blob/src/clients/append_blob_client.rs
```diff
@@ -40,45 +94,81 @@ impl AppendBlobClient {
     /// * `options` - Optional configuration for the client.
     pub fn new(
         endpoint: &str,
-        container_name: String,
-        blob_name: String,
+        container_name: &str,
+        blob_name: &str,
```
jalauzon-msft, line 98: Be sure to include this in a breaking changes section of README.

sdk/storage/azure_storage_blob/src/clients/append_blob_client.rs
```diff
@@ -40,45 +94,81 @@ impl AppendBlobClient {
...
+
+    /// Creates a new AppendBlobClient from a blob URL containing the SAS (shared access signature) query parameters.
+    ///
+    /// # Arguments
+    ///
+    /// * `blob_url` - The full URL of the Append blob, including the SAS query parameters.
+    /// * `credential` - An optional implementation of [`TokenCredential`] that can provide an Entra ID token to use when authenticating.
+    /// * `options` - Optional configuration for the client.
+    pub fn from_blob_sas_url(
+        blob_url: Url,
```
jalauzon-msft, line 126: nit: Should we maybe call this `blob_sas_url` now?

sdk/storage/azure_storage_blob/src/clients/append_blob_client.rs
```diff
@@ -40,45 +94,81 @@ impl AppendBlobClient {
...
+    ) -> Result<Self> {
+        let client = GeneratedAppendBlobClient::from_url(blob_url, None, options)?;
+
+        Ok(Self { client })
     }
 
-    /// Gets the blob name of the Storage account this client is connected to.
-    pub fn blob_name(&self) -> &str {
-        &self.client.blob_name
+    /// Gets the URL of the Storage account this client is connected to.
```
jalauzon-msft: I know you didn't write this in this PR but this comment is a bit off. Maybe something like "Gets the URL of the resource this client is configured for" - applies to all clients.

sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
```diff
@@ -50,29 +51,128 @@ impl BlobServiceClient {
...
+    }
+
+    /// Creates a new BlobServiceClient from the URL of the Azure storage account.
+    ///
+    /// # Arguments
+    ///
+    /// * `url` - The full URL of the Azure storage account, for example `https://myaccount.blob.core.windows.net/`.
+    /// * `credential` - An optional implementation of [`TokenCredential`] that can provide an Entra ID token to use when authenticating.
+    /// * `options` - Optional configuration for the client.
+    pub fn from_url(
```
jalauzon-msft: What do we call this in other languages? `from_service_url`? Just `from_url` while the others have nouns seems wrong.

sdk/storage/azure_storage_blob/src/clients/block_blob_client.rs
```diff
@@ -16,39 +16,35 @@ use crate::{
...
+impl GeneratedBlockBlobClient {
+    /// Creates a new GeneratedBlockBlobClient from a block blob URL.
     ///
     /// # Arguments
     ///
-    /// * `endpoint` - The full URL of the Azure storage account, for example `https://myaccount.blob.core.windows.net/`
-    /// * `container_name` - The name of the container containing this Block blob.
-    /// * `blob_name` - The name of the Block blob to interact with.
-    /// * `credential` - An implementation of [`TokenCredential`] that can provide an Entra ID token to use when authenticating.
+    /// * `block_blob_url` - The full URL of the block blob, for example `https://myaccount.blob.core.windows.net/mycontainer/myblob`.
```
jalauzon-msft: The parameter is called `blob_url`. Probably check for this everywhere. It doesn't matter much right now since this is internal though.

sdk/storage/azure_storage_blob/tests/append_blob_client.rs
```diff
@@ -76,21 +76,15 @@ async fn test_append_block_from_url(ctx: TestContext) -> Result<(), Box<dyn Erro
...
-        blob_client_2.container_name(),
-        blob_client_2.blob_name()
-    );
     let append_blob_client = blob_client.append_blob_client();
     append_blob_client.create(None).await?;
 
     // Act
     append_blob_client
-        .append_block_from_url(source_url, 17, None)
+        .append_block_from_url(blob_client_2.url().as_str().into(), 17, None)
```
jalauzon-msft, line 87: This seems cumbersome... you convert the url to a string and then convert it back? Could we clone it instead? Should we be returning back an owned `Url`?

sdk/storage/azure_storage_blob/src/clients/append_blob_client.rs
```diff
@@ -105,7 +105,31 @@ impl AppendBlobClient {
...
+    pub fn with_no_credential(
+        endpoint: &str,
+        container_name: &str,
+        blob_name: &str,
+        options: Option<AppendBlobClientOptions>,
+    ) -> Result<Self> {
+        let mut url = Url::parse(endpoint)?;
+
+        url.path_segments_mut()
+            .expect("Invalid endpoint URL: Cannot append container_name and blob_name to the blob endpoint.")
```
heaths: Don't panic at runtime. That's why we have constructors return a `Result`. This would be an argument error i.e., `ErrorKind::Other`.

In generated code, we use a pattern like:

```rust
        if secret_name.is_empty() {
                "parameter secret_name cannot be empty",
            ));
        }
```

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -486,3 +490,163 @@ async fn test_get_account_info(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
+                17,
+                None,
+            )
+            .await?;
+
+        // Get Properties
+        let properties = blob_client_from_cc.get_properties(None).await?;
+        assert_eq!(17, properties.content_length()?.unwrap());
+
+        container_client.delete_container(None).await?;
```
jalauzon-msft: As we talked about before, this test doesn't really verify the Blob names are what we expect, just that the client works with them after creation. We should maybe do a list blobs call or something and make sure all the names are what we expect.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -486,3 +490,163 @@ async fn test_get_account_info(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
+        // Additional reserved chars: parentheses, brackets, colon, quotes, leading/trailing spaces
+        (
+            "test-container-more-special",
+            " file (copy) [2024]:version'1'.txt ",
+        ),
+        // Mix of forward and backslashes to test normalization/preservation
+        (
+            "test-container-forward-back-slashes",
+            "forward/back\\forward/back\\",
+        ),
```
jalauzon-msft: Should add a test case for already encoded characters. Like you want `%3D` in the actual blob name for example.

sdk/storage/azure_storage_blob/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+- Added support for SAS (Shared Access Signature) authentication scheme.
```
jalauzon-msft: Let's give a little more detail about what we actually changed. I think two entries, list the methods we added, and then a separate entry about SAS support via `from_blob_sas_url`.

Also from previous comment, mention breaking change on string type for container/blob name on the methods where we changed it.

heaths (changes requested):
Much better, but we shouldn't panic at runtime when we can avoid it.

sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
vincenttran-msft: We use overloads in the other languages, I think we should add the service prefix for consistency. Fixed!

sdk/storage/azure_storage_blob/tests/append_blob_client.rs
vincenttran-msft: Yeah this one is tricky. Our accessor now returns a `Url` type, and `append_block_from_url` expects a owned `String` type (this is consistent with the generated function, as this gets piped to `x-ms-copy-source`).

`Url` type only offers `as_str()` to get a `&str` ref, so then you need to do either `into()` or `to_string()` to get back the owned type... Abit unfortunate but this is where we are at. If we wanted to, we could make the convenience layer take a `Url` type and do this conversion to owned `String` on their behalf, but I am unsure if we want to get into that game... let me know your thoughts.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: Sounds good, added back the list blobs accessors test and just normalized `\` to `/` due to our findings. This makes the asserted names consistent with the normalization applied Service-side.

sdk/storage/azure_storage_blob/src/clients/append_blob_client.rs
```diff
@@ -32,12 +32,12 @@ impl GeneratedAppendBlobClient {
...
     /// # Arguments
     ///
-    /// * `blob_url` - The full URL of the Append blob, for example `https://myaccount.blob.core.windows.net/mycontainer/myblob`.
+    /// * `append_blob_url` - The full URL of the Append blob, for example `https://myaccount.blob.core.windows.net/mycontainer/myblob`.
     /// * `credential` - An optional implementation of [`TokenCredential`] that can provide an Entra ID token to use when authenticating.
     /// * `options` - Optional configuration for the client.
     #[tracing::new("Storage.Blob.AppendBlob")]
     pub fn from_url(
-        blob_url: Url,
+        append_blob_url: Url,
```
jalauzon-msft: I think you may have misinterpreted a comment of mine. I think this should be just `blob_url` in all clients. You just had the doc string wrong in Block Blob client.

sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
```diff
@@ -126,7 +126,7 @@ impl BlobServiceClient {
     /// * `url` - The full URL of the Azure storage account, for example `https://myaccount.blob.core.windows.net/`.
     /// * `credential` - An optional implementation of [`TokenCredential`] that can provide an Entra ID token to use when authenticating.
     /// * `options` - Optional configuration for the client.
-    pub fn from_url(
+    pub fn from_service_url(
```
jalauzon-msft: So, from a public API perspective, we don't actually need to offer a `from_url` type method since the URL is just the endpoint on a service client. `new` can be used in that case. We can remove this method entirely. Don't forget to remove from changelog.

sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
```diff
@@ -140,10 +140,13 @@ impl BlobServiceClient {
     ///
     /// # Arguments
     ///
-    /// * `url` - The full URL of the Azure storage account, including the SAS query parameters.
+    /// * `blob_service_sas_url` - The full URL of the Azure storage account, including the SAS query parameters.
     /// * `options` - Optional configuration for the client.
-    pub fn from_sas_url(url: Url, options: Option<BlobServiceClientOptions>) -> Result<Self> {
-        let client = GeneratedBlobServiceClient::from_url(url, None, options)?;
+    pub fn from_sas_url(
+        blob_service_sas_url: Url,
```
jalauzon-msft: This method needs to stay though. And for this param name, maybe just `sas_url`? Or maybe `service_sas_url` but I don't love that.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -576,6 +582,23 @@ async fn test_encoding_edge_cases(ctx: TestContext) -> Result<(), Box<dyn Error>
...
+        let list_blob_segment_response = page.unwrap().into_body()?;
+        let blob_items = list_blob_segment_response.segment.blob_items;
+        assert_eq!(1, blob_items.len());
+        let listed_blob_name = blob_items[0]
+            .name
+            .as_ref()
+            .unwrap()
+            .content
+            .as_ref()
+            .unwrap();
```
jalauzon-msft: This feels way too cumbersome to do a simple operation like getting the blob name out...

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -576,6 +582,23 @@ async fn test_encoding_edge_cases(ctx: TestContext) -> Result<(), Box<dyn Error>
         let properties = blob_client_new.get_properties(None).await?;
         assert_eq!(17, properties.content_length()?.unwrap());
 
+        // Check Name Equality
+        let mut list_blobs_response = container_client.list_blobs(None)?;
```
jalauzon-msft: Suggestion for the structure of this test overall - Why not put all Blobs in the same container and do one list call that the end. Would simplify your list of test cases and make the test a lot smaller, more readable.

sdk/storage/azure_storage_blob/CHANGELOG.md
```diff
@@ -4,10 +4,28 @@
...
+  - `BlobClient::from_blob_url()`
+  - `BlobContainerClient::from_container_url()`
+  - `BlobServiceClient::from_service_url()`
+  - `BlockBlobClient::from_blob_url()`
+  - `PageBlobClient::from_blob_url()`
+- This also includes support for client construction directly from URLs containing SAS (shared access signature) query parameters:
+  - `AppendBlobClient::from_blob_sas_url()`
+  - `BlobClient::from_blob_sas_url()`
+  - `BlobContainerClient::from_container_sas_url()`
+  - `BlobServiceClient::from_service_sas_url()`
```
jalauzon-msft: This is not the name of the method now.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: Agreed. I made an issue to track this on the board and we can look to improve `list_blobs` story in an upcoming PR.

sdk/storage/azure_storage_blob/src/clients/append_blob_client.rs
```diff
@@ -154,19 +166,19 @@ impl AppendBlobClient {
...
     pub fn from_blob_sas_url(
-        blob_url: Url,
+        blob_sas_url: Url,
         options: Option<AppendBlobClientOptions>,
     ) -> Result<Self> {
-        let client = GeneratedAppendBlobClient::from_url(blob_url, None, options)?;
+        let client = GeneratedAppendBlobClient::from_url(blob_sas_url, None, options)?;
 
         Ok(Self { client })
     }
```
heaths: I see this pattern throughout so I'm just going to comment here, but it applies globally:

As we discussed in the meeting today, we can't really enforce at build time that `from_blob_url` doesn't have a SAS token while `from_blob_sas_url` does, so why have both anyway? Do you plan to treat them differently in the client itself, or just let the service 403 or whatever if there's no SAS for an authenticated blob/container/etc.?

The point of "overloaded constructors" is to help guide the callers to use correct parameters with as much verified at compile time as possible. Clearly we can't with some things like the presence of query params, etc. So shouldn't just one constructor be enough?

And in that case, I really recommend - as is idiomatic to rust - just a common idiom like `from_url`, which it seems you had. Then customers know to look for `from_url` for all clients and don't have to know it's `from_container_url` on a `BlobContainerClient`, etc. That's not typical. There's a pretty consistent - and often vague (without being too vague, hopefully) - pattern in popular rust crates when it comes to `from*`, `as_*`, `to_*`, and `into_*` (the latter three are about the only thing that comes close to language guidelines that exist in Rust, actually).

heaths (changes requested):
Apart from my one overarching comment about idiomacy, the rest of the changes look great!

sdk/storage/azure_storage_blob/src/clients/append_blob_client.rs
vincenttran-msft: Landed on the `from_url` common idiom

sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -61,39 +58,112 @@ impl BlobContainerClient {
...
+    /// Returns a new instance of BlobClient.
+    ///
+    /// # Arguments
+    ///
+    /// * `blob_name` - The name of the blob.
+    pub fn blob_client(&self, blob_name: &str) -> BlobClient {
+        let mut blob_url = self.url().clone();
+        blob_url
+            .path_segments_mut()
+            .expect("Invalid endpoint URL: Cannot append blob_name to the blob endpoint.")
```
jalauzon-msft, line 152: Still have an `expect` (panic) here.

sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
```diff
@@ -50,29 +51,83 @@ impl BlobServiceClient {
...
-    pub fn blob_container_client(&self, container_name: String) -> BlobContainerClient {
-        BlobContainerClient {
-            endpoint: self.client.endpoint.clone(),
-            client: self.client.get_blob_container_client(container_name),
-        }
+    pub fn blob_container_client(&self, container_name: &str) -> BlobContainerClient {
+        let mut container_url = self.url().clone();
+        container_url
+            .path_segments_mut()
+            .expect("Cannot be base")
```
jalauzon-msft: And here. Check all "navigation methods"

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -486,3 +491,184 @@ async fn test_get_account_info(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
+                17,
+                None,
+            )
+            .await?;
+
+        // Get Properties
+        let properties = blob_client_from_url.get_properties(None).await?;
+        assert_eq!(17, properties.content_length()?.unwrap());
+
+        // Test Case 3: Initialize BlobClient using from_blob_url(), using Url methods to build full URL properly
```
jalauzon-msft: This case looks exactly the same as the one above? I think you meant to have a blob name with multiple path segments inside the blob name and push each one separately?

Also nit: in your comments you are still using `from_blob_url` when its just `from_url` now.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -486,3 +491,184 @@ async fn test_get_account_info(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
+                None,
+            )
+            .await?;
+
+        // Get Properties
+        let properties = blob_client_from_cc.get_properties(None).await?;
+        assert_eq!(17, properties.content_length()?.unwrap());
+    }
+
+    // Check name equality for all test cases
```
jalauzon-msft, line 617: So I think this test is still not quite right. As is your list blobs is really only testing the last of your four test cases since you reupload the same set of blobs each time. To be fully complete, you would need one list call per test case so you'd have to do four `for blob_name in test_cases` loops, each with a list call at the end. Though technically by asserting the length of the container at the end, it means we know none of the test cases added a blob with a different name so maybe its fine. I guess up to you.

This test could benefit greatly from some parametrization.

sdk/storage/azure_storage_blob/CHANGELOG.md
```diff
@@ -4,12 +4,28 @@
 
 ### Features Added
 
+- Added support for client construction directly from URLs:
+  - `AppendBlobClient::from_url()`
+  - `BlobClient::from_url()`
+  - `BlobContainerClient::from_url()`
+  - `BlobServiceClient::new()` (unchanged)
```
jalauzon-msft: Then why add it here?

sdk/storage/azure_storage_blob/CHANGELOG.md
```diff
@@ -4,12 +4,28 @@
...
+  - `BlockBlobClient::from_url()`
+  - `PageBlobClient::from_url()`
+- Consequently, this also includes support for client construction directly from URLs containing SAS (shared access signature) query parameters.
```
jalauzon-msft: Sorry I missed this before. Let's make this even more clear that we added SAS support by simply saying. "Added support for SAS (shared access signature) URLs via the new `from_url` methods." As it is worded, it is kind of easy to miss.

Cargo.toml
```diff
@@ -109,7 +109,6 @@ futures = "0.3"
 getrandom = { version = "0.3" }
 gloo-timers = { version = "0.3" }
 hmac = { version = "0.12" }
-litemap = "0.7.4"
```
jalauzon-msft, line 112: Were changes here intentional?

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: My intention was that yes, since they are all operating on the same set of container/blob name combinations, the result should be once instance of the blob per test case. When testing locally this was exactly the case I was hitting with previous drafts (# of blobs listed >  `test_cases.len()`) so I am happy with this design to cut down on the list blob calls. If you feel strongly here we can go the other way but as-is I think its sufficient.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: Good callout. I had a much more convoluted way to feed the separate blob-name path segments but I just realized there is a simpler way that I am happy with. Will put this refactor in shortly.
Edit: Actually I don't find any value in polluting the test case with this, since it will operate the same whether we pass `/` or `%2F`. Previous to this the third test case was testing building the Url with `format!()` but that case is no longer relevant since we take a `Url` type now. Removing the test case.

sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
vincenttran-msft: As discussed offline, still need to decide if we are okay with inconsistent navigation return types (the ones that go down a layer of abstraction would have to now return a `Result`, which those within the same layer can remain directly returning the client).

For additional context:
Accessors that augment the path (go down a layer in abstraction) have a possible "panic" since they need to augment the path.

- `BlobServiceClient::blob_container_client()`
- `BlobContainerClient::blob_client()`

This does not apply to same-abstraction accessors:

- `BlobClient::append_blob_client()`
- `BlobClient::block_blob_client()`
- `BlobClient::page_blob_client()`

jalauzon-msft: @heaths, this is an interesting scenario where a panic may be okay. We have already validated the URL during construction of the client and now we are just appending one segment. It would be really unfortunate to have these function return results. Is there any other option here?

heaths: If you're sure we shouldn't even get to that point then, yes, it's okay; however, I would add a comment right above that line (above the `.expect` line itself) saying that very thing in case, somehow, it ever happens. We have a better idea that it's truly exceptional and something somewhere failed to check.

heaths (approved):

eng/emitter-package.json
```diff
@@ -4,14 +4,14 @@
...
-    "@typespec/openapi": "1.5.0",
-    "@typespec/rest": "0.75.0",
-    "@typespec/versioning": "0.75.0",
-    "@typespec/xml": "0.75.0"
+    "@typespec/compiler": "^1.4.0",
+    "@typespec/http": "^1.4.0",
+    "@typespec/openapi": "^1.4.0",
+    "@typespec/rest": "~0.75.0",
+    "@typespec/versioning": "~0.75.0",
+    "@typespec/xml": "~0.75.0"
```
heaths, line 15: Why are these downgraded? @jhendrixMSFT?

vincenttran-msft: I made that change, those are to match what is found in:
https://github.com/Azure/typespec-rust/blob/main/packages/typespec-rust/package.json#L52-L74

There is an unrelated generator issue we will address after this PR goes through.

heaths: @jhendrixMSFT said he's fine with it.

PR #3179, opened 2025-10-10
Configure blob perf tests pipeline.

LarryOsterman:


sdk/storage/perf.yml
```diff
@@ -26,7 +26,7 @@ parameters:
 extends:
```
heaths, line 38: Per #3180 can you include the triggers here? We probably only want to run for pull requests, pushes, and on main (PRs that target main count). We probably don't need the feature/ or release/ branch filters unless you think it worth it. But definitely the `path` filter.

heaths: And don't forget to update your docs. In fact, I'd recommend maybe you have the README.md link to - with an absolute URL - the storage `perf.yml` as the canonical example since, almost certainly, it will be. As we make changes, we're not having to remember to update docs as well. Should still work for LLMs if the recent addition of `AGENTS.md` linking out to other files is any indication.

/cc @ronniegeraghty if he knows if that's to be expected - that agents will / should follow linked files universally or not. I worry a little (performance- and cost-wise) if they do always, though.

LarryOsterman: Already working on both of that, including updated and detailed instructions for creating the pipeline.

sdk/storage/azure_storage_blob/perf-tests.yml
heaths, line 1: I don't think this is right. Looking at some other languages, all their pipeline YAML is "perf.yml" and "perf-tests.yml" is something custom e.g., https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/perf-tests.yml (same with .NET).

LarryOsterman: I'm completely confused - this is *exactly* what all the other languages are doing. 

Could you help me understand why this is wrong?

sdk/core/azure_core_test/src/perf/img/new pipeline.png
heaths, line 1: I'd remove the image. The text description you have for as seldom as we do this should be sufficient.

sdk/core/azure_core_test/src/perf/README.md
```diff
@@ -150,3 +150,69 @@ Options:
 ```
 
 Note that some of these test options are not specific to the `list_blobs` test. This is to allow test options to be provided in any order in the command line.
+
+### Declaring a test pipeline
+
+Test pipelines are defined using a [`perf.yml`](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/storage/perf.yml) file declared in the service directory.
+
+For example, from the `storage` service:
```
heaths: What about the triggers? This is why I was saying maybe just link to the `sdk/storage/azure_storage_blob/perf.yml` file. Don't duplicate.

sdk/storage/azure_storage_blob/perf/list_blob_test.rs
```diff
@@ -77,15 +77,22 @@ impl ListBlobTest {
...
+            azure_core::Error::with_message(ErrorKind::Other, "Failed to set client")
+        })?;
+
+        let _result = self.client.get().unwrap().create_container(None).await?;
 
         for i in 0..self.count {
             let blob_name = format!("blob-{}", i);
-            let blob_client = self.client.blob_client(blob_name);
-
+            let blob_client = self.client.get().unwrap().blob_client(blob_name);
```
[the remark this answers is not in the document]
heaths: Um, make up your mind copilot. You say to remove it in one place and use it in another.

At any rate, yeah, @LarryOsterman, I'd use the `?` in both places - or maybe it's not actually relevant here now that I look at it more. Looks like we're getting a subclient.

sdk/storage/perf-resources.bicep
```diff
@@ -49,7 +49,6 @@ resource storage 'Microsoft.Storage/storageAccounts@2024-01-01' = {
     name: 'Premium_LRS'
   }
   properties: {
-    accessTier: 'Hot'
```
heaths, line 52: Comparing with .NET, this is very different than: https://github.com/Azure/azure-sdk-for-net/blob/1680e1f70de8ec94306f1ea6208cc0eb7ec01306/sdk/storage/Azure.Storage.Blobs/perf-resources.bicep. You also have content encryption enabled, which I imagine doesn't help things.

sdk/storage/azure_storage_blob/perf.yml
```diff
@@ -36,3 +36,16 @@ extends:
     EnvVars:
       # This is set in the InstallLanguageSteps
       VCPKG_BINARY_SOURCES_SECRET: $(VCPKG_BINARY_SOURCES_SECRET)
+
```
heaths, line 51: Just a nit: we typically put this toward the top run under parameters. Sort of a top-down most important -> least important / global thing. Anywhere is fine, though.

sdk/typespec/typespec_client_core/src/http/pipeline.rs
```diff
@@ -86,6 +86,7 @@ impl Pipeline {
         pipeline.push(transport);
 
         // Make sure we didn't have to resize the Vec.
+        #[cfg(debug_assertions)]
```
heaths, line 89: `debug_assert_eq!()` already has that check in it. This is redundant.

LarryOsterman: Tell `cargo test --release` that, it apparently disagrees with you.

```text
   Compiling typespec_client_core v0.9.0 (Q:\src\azure-sdk-for-rust\sdk\typespec\typespec_client_core)
error[E0425]: cannot find value `initial_capacity` in this scope
  --> sdk\typespec\typespec_client_core\src\http\pipeline.rs:89:42
   |
89 |         debug_assert_eq!(pipeline.len(), initial_capacity);
   |                                          ^^^^^^^^^^^^^^^^ not found in this scope

For more information about this error, try `rustc --explain E0425`.
```

LarryOsterman: I suspect that the core problem here is that the declaration of `initial_capacity` was compiled out and the compiler tried to resolve the `initial_capacity` symbol before it tried to resolve the `debug_assert_eq!` macro.

sdk/core/azure_core_test/src/perf/README.md
LarryOsterman: See the ***NOTE*** just below this.

sdk/storage/perf-resources.bicep
LarryOsterman: Unfortunately, content encryption enabled doesn't change a thing :(.

sdk/typespec/typespec_client_core/src/http/pipeline.rs
heaths: Yeah, you're probably right. I swear I did a release build. Guess not.

PR #3183, opened 2025-10-11
Get final state from pollers

### LINKED ISSUE #2757: Return final resource from Poller

heaths:
Resolves #2757

### DIFF
--- a/sdk/core/azure_core/src/http/poller.rs
+++ b/sdk/core/azure_core/src/http/poller.rs
@@ -659,9 +659,10 @@ where
     let target = Box::new(async move {
         match target_rx.await {
             Ok(fut) => fut.await,
-            Err(_) => Err(crate::Error::new(
+            Err(err) => Err(crate::Error::with_error(
                 ErrorKind::Other,
-                "poller completed without sending target response",
+                err,
+                "poller completed without defining a target",
             )),
         }
     });
--- a/sdk/core/typespec_client_core/src/http/format.rs
+++ b/sdk/core/typespec_client_core/src/http/format.rs
@@ -4,18 +4,18 @@
 use crate::{error::ErrorKind, http::response::ResponseBody};
 use serde::de::DeserializeOwned;
 
-/// A marker trait used to indicate the serialization format used for a response body.
+/// A trait used to indicate the serialization format used for a response body.
 ///
-/// The [`Response`](crate::http::Response) type uses this trait, in the `F` parameter, to determine how to deserialize the body in to the `T` model when using [`Response::into_body`](crate::http::Response::into_body).
-///
-/// ## How this trait works
-///
-/// This trait is a little funky, in order to allow the deserialization behavior of the format to be adjusted based on the type of the response body.
-/// This is just a marker trait, it has no methods.
-/// Instead, the method to actually perform the deserialization is defined in the [`DeserializeWith`] trait.
-/// This trait is parameterized by a type that implements the [`Format`] trait.
+/// The [`Response`](crate::http::Response) type uses this trait in parameter `F` to determine how to deserialize the body in to the model `T` when using [`Response::into_body`](crate::http::Response::into_body).
+/// This allows the client library to define the format for each client method so callers can deserialize the model `T` without having to know or specify the format.
 pub trait Format: std::fmt::Debug {
-    /// Deserialize
+    /// Deserialize body into model `T`.
+    ///
+    /// # Arguments
+    /// * `body` - The body to deserialize.
+    ///
+    /// # Returns
+    /// * A `Result` containing the deserialized model `T`.
     fn deserialize<T: DeserializeOwned, S: AsRef<[u8]>>(body: S) -> crate::Result<T>;
 }
 
--- a/sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
+++ b/sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
@@ -131,7 +131,7 @@ impl CertificateClient {
                 let pipeline = pipeline.clone();
                 let api_version = api_version.clone();
                 async move {
-                    let rsp: RawResponse = pipeline.send(&ctx, &mut request, None).await?;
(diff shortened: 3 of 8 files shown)


RickWinter:
@heaths The doc comments should be made. Do you want to accept the recommendations or do you want to push updates from your branch?

sdk/core/azure_core/examples/core_poller.rs
```diff
@@ -55,11 +101,20 @@ async fn test_core_poller() -> Result<(), Box<dyn std::error::Error>> {
     test_poller().await
 }
 
+#[tokio::test]
+async fn test_core_poller_stream() -> Result<(), Box<dyn std::error::Error>> {
```
LarryOsterman, line 105: Dead code? I don't see this called, did I miss it?

sdk/core/azure_core/src/http/poller.rs
```diff
@@ -437,50 +581,92 @@ enum State<N> {
...
+            // Flow 'make_request' and target_tx through to avoid cloning
+            Some((item, (next_state, make_request, target_tx)))
         },
-    )
+    );
+
+    let target = Box::new(async move {
+        match target_rx.await {
+            Ok(fut) => fut.await,
+            Err(_) => Err(crate::Error::new(
```
LarryOsterman: I take it that the error here isn't interesting?

sdk/core/azure_core/src/http/poller.rs
```diff
@@ -437,50 +581,92 @@ enum State<N> {
...
+                ErrorKind::Other,
+                "poller completed without sending target response",
```
LarryOsterman: Shouldn't this indicate some kind of failure rather than just "completed without sending a response"?

sdk/core/typespec_client_core/src/http/format.rs
```diff
@@ -14,14 +14,22 @@ use serde::de::DeserializeOwned;
 /// This is just a marker trait, it has no methods.
 /// Instead, the method to actually perform the deserialization is defined in the [`DeserializeWith`] trait.
 /// This trait is parameterized by a type that implements the [`Format`] trait.
-pub trait Format: std::fmt::Debug {}
+pub trait Format: std::fmt::Debug {
+    /// Deserialize
+    fn deserialize<T: DeserializeOwned, S: AsRef<[u8]>>(body: S) -> crate::Result<T>;
```
LarryOsterman, line 19: I don't think this is a marker trait any more, update the comment?

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
```diff
@@ -135,6 +129,7 @@ impl CertificateClient {
 
                 let ctx = options.method_options.context.clone();
                 let pipeline = pipeline.clone();
+                let api_version = api_version.clone();
                 async move {
                     let rsp: RawResponse = pipeline.send(&ctx, &mut request, None).await?;
```
LarryOsterman: Do you need to type qualify the `rsp` variable here?

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
```diff
@@ -225,14 +250,19 @@ impl CertificateClient {
...
         let pipeline = self.pipeline.clone();
+
         let mut url = self.endpoint.clone();
         let mut path = String::from("certificates/{certificate-name}/pending");
         path = path.replace("{certificate-name}", certificate_name);
         url = url.join(&path)?;
         url.query_pairs_mut()
             .append_pair("api-version", &self.api_version);
+
+        let api_version = self.api_version.clone();
```
LarryOsterman, line 261: `api_version` doesn't appear to be used beyond line 265, do you need to clone it twice?

sdk/core/azure_core/examples/core_poller.rs
heaths: It's a test. 😉

sdk/core/azure_core/src/http/poller.rs
heaths: It's an apt description, and it's in an error. Much like Go, prefacing with "failed" is often unnecessary. Maybe I can reword a little.

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
heaths: Yes, because we have to move captures, so we can't reference `self`.

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
heaths: No. I think the emitter was doing this in some other client methods from where some of this code was originally copied; though, IIRC, I noticed that a while back and Joel or Anton fixed it since it doesn't do that now.

Issue #3127, opened 2025-10-07
Finalize get_token error type


chlowell:
All errors from get_token should probably be `ErrorKind::Credential`, assuming we believe that variant is useful, however they should also expose any HTTP response received from the token request because the application may want to inspect that. A couple options to consider off the top of my head:
1. `ErrorKind::Credential` arguably isn't necessary so, just remove it and return `ErrorKind::HttpResponse` as appropriate
2. have `ErrorKind::Credential` carry an optional `RawResponse`

PR #3184, opened 2025-10-13
Add ClientCertificateCredential tests

### LINKED ISSUE #3126: Add ClientCertificateCredential tests

chlowell:
Closes #3126

### DIFF
--- a/sdk/identity/azure_identity/src/client_certificate_credential.rs
+++ b/sdk/identity/azure_identity/src/client_certificate_credential.rs
@@ -252,7 +252,7 @@ impl ClientCertificateCredential {
         match rsp.status() {
             StatusCode::Ok => {
                 let response: EntraIdTokenResponse =
-                    deserialize("ClientCertificateCredential", rsp)?;
+                    deserialize(stringify!(ClientCertificateCredential), rsp)?;
                 Ok(AccessToken::new(
                     response.access_token,
                     OffsetDateTime::now_utc() + Duration::seconds(response.expires_in),
@@ -260,7 +260,7 @@ impl ClientCertificateCredential {
             }
             _ => {
                 let error_response: EntraIdErrorResponse =
-                    deserialize("ClientCertificateCredential", rsp)?;
+                    deserialize(stringify!(ClientCertificateCredential), rsp)?;
                 let message = if error_response.error_description.is_empty() {
                     "authentication failed".to_string()
                 } else {


.vscode/cspell.json
```diff
@@ -7,6 +7,7 @@
     "**/test-resources.json",
     "**/test-resources-post.ps1",
     "**/assets.json",
+    "**/*.pfx",
```
heaths, line 10: Binaries in git === bad. Use PEM.

sdk/identity/azure_identity/src/client_certificate_credential.rs
```diff
@@ -245,11 +248,27 @@ impl ClientCertificateCredential {
...
-        let response: EntraIdTokenResponse = rsp.into_body().json()?;
-        Ok(AccessToken::new(
-            response.access_token,
-            OffsetDateTime::now_utc() + Duration::seconds(response.expires_in),
-        ))
+
+        match rsp.status() {
+            StatusCode::Ok => {
+                let response: EntraIdTokenResponse =
+                    deserialize("ClientCertificateCredential", rsp)?;
```
heaths: Tip: use `stringify!(ClientCertificateCredential)`. It's not quite as good as C#'s `typeof().FullName` but some IDEs treat is similarly should we ever rename.

sdk/identity/azure_identity/src/client_certificate_credential.rs
```diff
@@ -245,11 +248,27 @@ impl ClientCertificateCredential {
...
+            }
+            _ => {
+                let error_response: EntraIdErrorResponse =
+                    deserialize("ClientCertificateCredential", rsp)?;
+                let message = if error_response.error_description.is_empty() {
+                    "authentication failed".to_string()
+                } else {
+                    error_response.error_description.clone()
+                };
+                Err(Error::with_message(ErrorKind::Credential, message))
```
heaths, line 269: We don't want to just return the original `ErrorKind::HttpResponse`? You could use `ResultExt::with_context` to wrap it with `ErrorKind::Credential` 1) with much less code, and 2) retain the original error if we want.

.vscode/cspell.json
chlowell: This one's small and won't change over time. Is it really so problematic that we should instead commit a PEM file and convert that to PKCS12 at runtime?

sdk/identity/azure_identity/src/client_certificate_credential.rs
chlowell: This sounds like a solution for #3127. What I have here is consistent with other types pending a resolution for that issue

.vscode/cspell.json
heaths: Do we not support PEM now? I expect where Azure SDK for Rust gets used most, PEM will be common. Only Windows practically uses PFX.

chlowell: The current implementation accepts only PKCS12. I agree PEM is more popular but that's a future PR and PEM vs. PKCS12 is irrelevant to this one anyway: if we support PKCS12, we need to test that.

heaths (approved):

PR #3186, opened 2025-10-13
Finish readme text on pipeline provisioning; added keyvault secrets t…

LarryOsterman:
…ests

### DIFF
--- a/sdk/core/azure_core/benches/http_transport_benchmarks.rs
+++ b/sdk/core/azure_core/benches/http_transport_benchmarks.rs
@@ -15,7 +15,7 @@ use azure_identity::DeveloperToolsCredential;
 use criterion::{criterion_group, criterion_main, Criterion};
 use std::sync::Arc;
 
-#[cfg_attr(target_os = "macos", allow(unused_variables))]
+#[cfg_attr(target_os = "macos", allow(dead_code))]
 const HTTP_ENDPOINT: &str = "https://azuresdkforcpp.azurewebsites.net";
 //const HTTP_ENDPOINT: &str = "http://httpbin.org";
 


sdk/core/azure_core_test/src/perf/README.md
```diff
@@ -268,4 +280,34 @@ Next select `Azure/azure-sdk-for-rust` to specify the Rust SDK and configure you
 
 Select your pipeline file from the main branch of the repository and you're almost done.
 
-The next thing you want to do is to "save" the new pipeline.
+The next thing you want to do is to "save" the new pipeline, this will cause your pipeline to be created. You can also attempt to `run` the pipeline at this point but it is likely to fail.
```
heaths: Grammar nit: 
```suggestion
The next thing you want to do is to "save" the new pipeline. This will cause your pipeline to be created. You can also attempt to `run` the pipeline at this point but it is likely to fail.
```

sdk/core/azure_core_test/src/perf/README.md
```diff
@@ -268,4 +280,34 @@ Next select `Azure/azure-sdk-for-rust` to specify the Rust SDK and configure you
...
+
+Select `Variables`, which allows you to add variables to the pipeline. You want to select `Variable groups` in the left hand column and select `Link variable group` in the right hand column.
+
+That will bring up a pane on the right with a number of variable groups. You want to select the `Secrets for Resource Provisioner` variable group and click the `Link` button.
+
+Once you've saved these changes, your pipeline should be ready to run.
+
+You may need to ask for help from the Azure SDK Engineering Systems team to enable access to test resources for your pipeline, but that should be a relatively straightforward request.
```
heaths: Don't need to put them on the spot. ;)

```suggestion
You may need to ask for help from the Azure SDK Engineering Systems team to enable access to test resources for your pipeline.
```

sdk/keyvault/azure_security_keyvault_keys/benches/benchmarks.rs
```diff
@@ -237,7 +166,7 @@ criterion_group! {
         .sample_size(10)
         .warm_up_time(std::time::Duration::new(1, 0))
         .measurement_time(std::time::Duration::from_secs(30));
-    targets = key_operations_benchmark, key_2901_benchmark_default, key_2901_benchmark_slow
+    targets = key_2901_benchmark_default, key_2901_benchmark_slow
```
heaths: For what this whole file is doing - benchmarking the client and service - I don't think it's even worth keeping. It's not a perf test. It's a criterion test, and we probably don't want to go across the network normally. KV has too low a limit as we discussed offline. I think we nix this whole file and make sure `Cargo.toml` is cleaned up.

sdk/keyvault/azure_security_keyvault_keys/tests/results.json
heaths, line 1: Should this have been checked in?

sdk/keyvault/azure_security_keyvault_keys/tests/results.json
```diff
@@ -0,0 +1,6 @@
+{
+  "test_name": "get_key",
+  "operations_per_second": 27.933333333333334,
```
heaths: Based on the limit of 300 create key/secert/cert requests per 10 seconds, this is on par and I still think is useless to have KV-based perf tests. KV was one of the first services to have these as @mikeharder was rolling out his framework back in the day. The fact no one else seems to run them, I imagine, is because of the low throttle.

I appreciate this was to do an apples to apples comparison with C++, but is it worth checking in? Should we at least turn it into a manual test? I don't think it serves enough value to run on a schedule.

sdk/keyvault/azure_security_keyvault_secrets/tests/results.json
```diff
@@ -0,0 +1,6 @@
+{
+  "test_name": "get_secret",
+  "operations_per_second": 456.6666666666667,
```
heaths: Same comments for keys as here (probably not supposed to be checked in; of dubious value).

sdk/keyvault/azure_security_keyvault_keys/benches/benchmarks.rs
LarryOsterman: @#$@$ - I added this as a part of the connection pooling bug investigation. It's gone.

sdk/keyvault/azure_security_keyvault_keys/tests/results.json
LarryOsterman: My goal was to preserve the existing performance tests which were previously written. I agree that create key isn't an interesting perf tests.

I'll nuke it from the perf-tests.yml file.

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -505,7 +505,7 @@ impl PerfRunner {
             .arg(
                 clap::arg!(--"test-results" <FILE> "The file to write test results to")
                     .required(false)
-                    .default_value("./tests/results.json")
+                    .default_value("./results.json")
```
heaths, line 508: Should `results.json` go into `.gitignore`, perhaps?

heaths (approved):

heaths (approved):

PR #3187, opened 2025-10-13
[Ready for EngSys Review] ESRP Publishing

danieljurek:
> [!WARNING]  
> SERVICE OWNERS: no need to review this PR. I'm making engsys changes to ci.yml files. 

Really fixes https://github.com/Azure/azure-sdk-for-rust/issues/3035

[keeping as draft to avoid notifying everyone]

A new release crate-packing strategy for Rust which uses `cargo publish --dry-run` to produce the crates: 

```
cargo publish --dry-run --package <package1> --package <package2> ...
```

This uses existing tooling to do the build with no vendoring or separate packing. Release order sorting can be read from the tool output and doesn't need to be computed from metadata. `cargo publish` also verifies that dependencies defined in the package's `Cargo.toml` are published and fails if they are not published. 

This also supports the current Pre-GA approach of not incrementing core's versions after release as the workspace dependency includes 

For example: 

If a package depends on `azure_core@0.30.0` and there is no `azure_core@0.30.0` published on crates.io, the command will fail. It's possible to "atomically" generate packages for releases with unreleased dependencies if those dependencies are included in the package command (e.g. `storage_blobs` depending on `storage_common`)

## Formatting: 

Formatting attempts to match reasonable outputs. Redirecting stderr from `cargo publish` would normally result in red console text. The `cargo publish` command is also grouped: https://dev.azure.com/azure-sdk/public/_build/results?buildId=5461494&view=logs&j=b766ebde-1fdb-5f11-1350-46ddc53b23cf&t=7ef29fa0-f532-5653-3473-16dc04608431&l=104

image

## Tests: 

| Scenario | Expected Outcome | Links | 
| -------- | ---------------- | ---- |
| Release azure_canary w/o required azure_canary_core (intra-service dependency) | Fail | https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5460934&view=results |
| Release azure_canary and azure_canary_core (intra-service dependency) | Success | https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5460932&view=results |
| Release azure_canary_core | Success | https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5460854&view=results |
| Release azure_canary_core which depends on **unreleased** azure_core | Fail | https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5460945&view=results |
| Release azure_canary_core which depends on released azure_core | Success | https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5460949&view=results |
| Manually queued live tests (no release) for Key Vault | Success | https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5460954&view=results | 
| PR pipeline works | Success | https://dev.azure.com/azure-sdk/public/_build/results?buildId=5460931&view=results |

### DIFF
--- a/eng/pipelines/templates/stages/archetype-rust-release.yml
+++ b/eng/pipelines/templates/stages/archetype-rust-release.yml
@@ -85,11 +85,6 @@ stages:
             runOnce:
               deploy:
                 steps:
-                - pwsh: |
-                    Write-Host "Setting ArtifactName to empty"
-                    Write-Host "##vso[task.setvariable variable=ArtifactName;]"
-                  displayName: Initialize ArtifactName variable
-
                 # This loop over artifacts is used to produce the correct number
                 # of ESRP release tasks. It has the side effect of also setting
                 # the artifact name by looking up the index of the current


eng/scripts/Pack-Crates.ps1
```diff
@@ -54,13 +57,30 @@ function Get-OutputPackageNames($workspacePackages) {
...
 
-  return $names
+  return , $names
 }
 
 function Get-CargoMetadata() {
   cargo metadata --no-deps --format-version 1 --manifest-path "$RepoRoot/Cargo.toml" | ConvertFrom-Json -Depth 100 -AsHashtable
 }
 
+function Test-PublishedOnCratesIo($name, $version) {
```
danieljurek: This is new

eng/scripts/Pack-Crates.ps1
```diff
@@ -162,13 +186,33 @@ try {
 
   [array]$packages = Get-PackagesToBuild
 
+  if ($RequireDependencies) {
+    $unspecifiedPackages = $packages | Where-Object { $_.name -notin $PackageNames -and !(Test-PublishedOnCratesIo -name $_.name -version $_.version) } | Foreach-Object { "$($_.name)`@$($_.version)" }
```
danieljurek: As written when `-RequireDependencies` is specified and a package that needs to be built is NOT in `-PackageNames` then this will `Test-PublishedOnCratesIo` in all cases.

eng/pipelines/templates/jobs/pack.yml
```diff
@@ -55,15 +55,51 @@ jobs:
...
+        }
+
+        $packageNames = $artifactsToBuild.name
+
+        Write-Host "##vso[task.setvariable variable=PackageNames]$($packageNames -join ',')"
+        if ($isReleaseBuild) {
+          Write-Host "##vso[task.setvariable variable=AdditionalPackageParams]-Release"
+        } else {
+          Write-Host "##vso[task.setvariable variable=AdditionalPackageParams]"
+        }
```
hallipr, line 88: It may be cleaner to just invoke Pack-Crates here rather than setting devops variables and invoking them in a following step.

Unless you want to track the error rates for the step separately

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -84,33 +85,83 @@ stages:
...
+                  - pwsh: |
+                      Write-Host "##vso[task.setvariable variable=ArtifactIndex]0"
+                    displayName: Set ArtifactIndex to 0
+
+                  - ${{ each artifact in parameters.Artifacts }}:
+                    - pwsh: |
+                        # Read artifact release order from release-order.json 
+                        # and use ArtifactIndex to select the right one
+                        $index = [int]'$(ArtifactIndex)'
+                        $artifacts = Get-Content '$(Pipeline.Workspace)/drop/release-order.json' | ConvertFrom-Json
```
hallipr: you can avoid the index variable by using the natural index of the artifact in the array
```suggestion
                  - ${{ each artifact in parameters.Artifacts }}:
                    - pwsh: |
                        # Read artifact release order from release-order.json 
                        # and use ArtifactIndex to select the right one
                        $artifactNames = $env:ARTIFACT_NAMES | ConvertFrom-Json -NoEnumerate
                        $index = $artifactNames.IndexOf($env:ARTIFACT_NAME)
                        $artifacts = Get-Content '$(Pipeline.Workspace)/drop/release-order.json' | ConvertFrom-Json
```

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -84,33 +85,83 @@ stages:
...
+                        $packageVersion = $packageMetadata.version
+                        Write-Host "Package version: $packageVersion"
+
+                        $cratePath = "$artifactRootPath/$artifactName/$artifactName-$packageVersion.crate"
+                        Copy-Item `
+                          -Path $cratePath `
+                          -Destination $outDir
+                        Write-Host "Contents of $outDir"
+                        Get-ChildItem -Path $outDir | ForEach-Object { Write-Host $_.FullName }
+                      displayName: 'Copy crate for ESRP'
```
hallipr: If avoiding the index variable:

```suggestion
                      displayName: 'Copy crate for ESRP'
                      env:
                        ARTIFACT_NAMES: ${{ convertToJson(parameters.Artifacts.*.Name)  }}
                        ARTIFACT_NAME: ${{ artifact.Name }}
```

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -84,33 +85,83 @@ stages:
...
+                        contenttype: 'Rust'
+                        contentsource: 'Folder'
+                        folderlocation: '$(Pipeline.Workspace)/esrp-release'
+                        waitforreleasecompletion: true
+                        owners: ${{ coalesce(variables['Build.RequestedForEmail'], 'azuresdk@microsoft.com') }}
+                        approvers: ${{ coalesce(variables['Build.RequestedForEmail'], 'azuresdk@microsoft.com') }}
+                        serviceendpointurl: 'https://api.esrp.microsoft.com/'
+                        mainpublisher: 'ESRPRELPACMANTEST'
+
+                    - pwsh: |
```
hallipr: If keeping the index variable, I'd just move this to the end of the initial script

eng/pipelines/templates/stages/archetype-rust-release.yml
danieljurek: > you can avoid the index variable by using the natural index of the artifact in the array

That's possible but requires the ci.yml file to be topo sorted. Rather than require the user to sort the array themselves (and sort order can change when the root Cargo.toml changes) this takes a sort order provided in `release-order.json` which comes from `Pack-Crates.ps1`. Now you can order `ci.yml` in any way and your packages will release correctly.

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -71,7 +69,10 @@ stages:
             - input: pipelineArtifact  # Required, type of the input artifact
               artifactName: ${{parameters.PipelineArtifactName}}  # Required, name of the pipeline artifact
               targetPath: $(Pipeline.Workspace)/drop  # Optional, specifies where the artifact is downloaded to
-          environment: ${{parameters.Environment}}
+          ${{if parameters.TestPipeline}}:
+            environment: none
+          ${{else}}:
+            environment: cratesio
```
weshaggard: I know this was there already but we should just use the general `package-publish`

eng/pipelines/templates/stages/archetype-rust-release.yml
weshaggard: I agree this could happen in the same pwsh step above.

eng/pipelines/templates/stages/archetype-rust-release.yml
weshaggard: I think what he is saying is you can just look up the artifact in the artifact list to find the index number itself, and then use the index like you are already doing in find the one in the release order. That would stop the need for using the devops variable for tracking that index.

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -84,33 +85,83 @@ stages:
...
-                      cargo owner --add $owner $crateName
-                    }
-                  displayName: Publish Crate
-                  env:
-                    CARGO_REGISTRY_TOKEN: $(azure-sdk-cratesio-token)
+                  - pwsh: |
+                      Write-Host "##vso[task.setvariable variable=ArtifactIndex]0"
+                    displayName: Set ArtifactIndex to 0
+
+                  - ${{ each artifact in parameters.Artifacts }}:
```
weshaggard: Can we add a comment here clarifying that we are only using this to do a for loop over the number of artifacts.

eng/scripts/Pack-Crates.ps1
```diff
@@ -156,64 +114,73 @@ function Create-ApiViewFile($package) {
...
       Copy-Item -Path $apiReviewFile -Destination $targetApiReviewFile -Force
     }
   }
 
-  Write-Host "Removing local registry"
-  Remove-Item -Path $localRegistryPath -Recurse -Force | Out-Null
+  if ($OutBuildOrderFile) {
+    $buildOrder = @()
+    foreach ($line in $packResult) { 
+      if ($line -match '^\s*Packaging (\w*) ([\w\d\.-]*)') {
```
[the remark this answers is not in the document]
danieljurek: Packaging guidelines use underscores as separators. https://azure.github.io/azure-sdk/rust_introduction.html#rust-packaging

eng/pipelines/templates/jobs/pack.yml
```diff
@@ -55,15 +55,52 @@ jobs:
...
+        filePath: $(Build.SourcesDirectory)/eng/scripts/Pack-Crates.ps1
+        arguments: >
+          -OutputPath '$(Build.ArtifactStagingDirectory)'
+          -PackageInfoDirectory '$(Build.ArtifactStagingDirectory)/PackageInfo'
+
+  - ${{ else }}:
+    - pwsh: |
+        $artifacts = '${{ convertToJson(parameters.Artifacts) }}' | ConvertFrom-Json
+        $isReleaseBuild = $true
+        $artifactsToBuild = $artifacts | Where-Object { $_.releaseInBatch -eq 'True' }
```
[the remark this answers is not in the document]
danieljurek: In this case the JSON created by DevOps is a literal string `'True'` instead of a boolean value. This match is accurate to the context.

eng/pipelines/templates/stages/archetype-rust-release.yml
weshaggard: You don't even need those as env variables you can do that directly inline in the script.

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -130,69 +181,32 @@ stages:
...
-          # Apply the version increment to each library, which updates the Cargo.toml and changelog files.
-          - task: PowerShell@2
-            displayName: Increment ${{artifact.name}} version
-            inputs:
-              targetType: filePath
-              filePath: $(Build.SourcesDirectory)/eng/scripts/Update-PackageVersion.ps1
-              arguments: >
-                -ServiceDirectory '${{parameters.ServiceDirectory}}'
-                -PackageName '${{artifact.name}}'
+          - ${{each artifact in parameters.Artifacts }}:
```
weshaggard:
```suggestion
          - ${{ each artifact in parameters.Artifacts }}:
```

eng/pipelines/templates/stages/archetype-rust-release.yml
danieljurek: I see what you mean. The code does become more simple. I think the int index is easier to reason about but IndexOf works well enough

sdk/core/ci.yml
```diff
@@ -10,22 +10,52 @@ trigger:
...
+      releaseInBatch: ${{ parameters.typespec_macros }}
     - name: typespec_client_core
-      safeName: TypespecClientCore
+      releaseInBatch: ${{ parameters.typespec_client_core }}
     - name: azure_core
-      safeName: AzureCore
+      releaseInBatch: ${{ parameters.azure_core }}
     - name: azure_core_macros
-      safeName: AzureCoreMacros
+      releaseInBatch: ${{ parameters.azure_core_macros }}
```
heaths, line 57: Order doesn't matter right? Otherwise, `azure_core_macros` should go before `azure_core`.

sdk/storage/ci.yml
```diff
@@ -15,16 +9,37 @@ trigger:
...
-    - name: azure_storage_common
-      safeName: AzureStorageCommon
+    # Artifact order in this file maps to order evaluated by Pack-Crates.ps1
+    # When the artifacts like blob and queue depend on common, common will need
+    # to be listed first.
     - name: azure_storage_blob
-      safeName: AzureStorageBlob
+      releaseInBatch: ${{ parameters.release_azure_storage_blob }}
+    - name: azure_storage_common
+      releaseInBatch: ${{ parameters.release_azure_storage_common }}
```
heaths: I didn't realize this was in there. It shouldn't be. And, in fact `sdk/storage/azure_storage_common/Cargo.toml` should have `publish = false` under `[package]`.

@vincenttran-msft can we just get rid of this crate? You can always add it back.

heaths: This isn't blocking, but would be appreciated.

eng/scripts/Language-Settings.ps1
```diff
@@ -139,15 +139,32 @@ function Get-rust-AdditionalValidationPackagesFromPackageSet ($packagesWithChang
   return $additionalPackages ?? @()
 }
 
+# $GetPackageInfoFromPackageFileFn = "Get-${Language}-PackageInfoFromPackageFile"
```
heaths, line 142: Nit: probably just remove. It's source control so never gone. Where it is in the source file, this is just confusing.

eng/scripts/Pack-Crates.ps1
```diff
@@ -156,64 +114,73 @@ function Create-ApiViewFile($package) {
   "$packagePath/review/$packageName.rust.json"
 }
 
-Push-Location $RepoRoot
+$originalLocation = Get-Location
 try {
-  $localRegistryPath = Initialize-VendorDirectory
+  Set-Location $RepoRoot
```
heaths, line 124: 1. Just use `push-location`. Why do it manually?
2. Why change directories at all? We should be using `cargo <verb> --package <package> ...` everywhere. Almost all `cargo` commands take a package argument and location doesn't matter.

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -84,33 +85,82 @@ stages:
...
+                  # for publishing each artifact in an order specified in the
+                  # release-order.json file.
+                  - ${{ each artifact in parameters.Artifacts }}:
+                    - pwsh: |
+                        $artifacts = Get-Content '$(Pipeline.Workspace)/drop/release-order.json' | ConvertFrom-Json
+
+                        # Force $artifacts to be an array (PowerShell unrolls single-item arrays)
+                        if ($artifacts -isnot [Array]) {
+                          $artifacts = @($artifacts)
+                        }
```
heaths: Tip: you can declare `[object[]] $artifacts = ...` as well, or use `@(, Get-Content ...)` as well.

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -84,33 +85,82 @@ stages:
...
-                    foreach ($owner in $missingOwners) {
-                      Write-Host "> cargo owner --add $owner $crateName"
-                      cargo owner --add $owner $crateName
-                    }
-                  displayName: Publish Crate
-                  env:
-                    CARGO_REGISTRY_TOKEN: $(azure-sdk-cratesio-token)
+                  - pwsh: |
+                      Write-Host "Setting ArtifactName to empty"
+                      Write-Host "##vso[task.setvariable variable=ArtifactName;]"
```
weshaggard: What is this needed?

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -84,33 +85,82 @@ stages:
...
+
+                        $cratePath = "$artifactRootPath/$artifactName/$artifactName-$packageVersion.crate"
+                        Copy-Item `
+                          -Path $cratePath `
+                          -Destination $outDir
+                        Write-Host "Contents of $outDir"
+                        Get-ChildItem -Path $outDir | ForEach-Object { Write-Host $_.FullName }
+
+                        # Set next artifact to release by
+                        $artifactIndex = $artifacts.IndexOf($artifactName)
```
weshaggard: OK I guess you are building a list as you go which I suppose that also works. I was thinking we would just use `${{ artifact.name }}` to look up an index number to use as your counter and then index into the release-order with that. You wouldn't need to depend on a persisted variable between steps to make that work.

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -84,33 +85,82 @@ stages:
...
-                    $missingOwners = $additionalOwners | Where-Object { $existingOwners -notcontains $_ }
-
-                    foreach ($owner in $missingOwners) {
-                      Write-Host "> cargo owner --add $owner $crateName"
-                      cargo owner --add $owner $crateName
-                    }
-                  displayName: Publish Crate
-                  env:
-                    CARGO_REGISTRY_TOKEN: $(azure-sdk-cratesio-token)
+                  - pwsh: |
```
hallipr, line 95: there's inconsistent sequence indentation here.  For the rest of the file, the `- ` matches the indentation of the key.
```
steps:
- pwsh: |
```

eng/pipelines/templates/stages/archetype-rust-release.yml
danieljurek: The script 

```pwsh
$myVar = '$(ArtifactName)'
```

If you haven't initialized the variable `ArtifactName` then the macro syntax doesn't try to expand the string. Instead you get the literal string `$(ArtifactName)` instead of an empty string.

eng/scripts/Language-Settings.ps1
danieljurek: The `Get-${Language}-PackageInfoFromPackageFile` function is defined with a different name in every language repo but the variable used to refer to that function is the same (`$GetPackageInfoFromPackageFileFn`). This comment is a standard way of making function invocations easy to search from the declaration of the function. 

With this comment I can search for `$GetPackageINfoFromPackageFileFn` instead of having to go to `eng/common/scripts/common.ps1`, then look up `Get-${Language}-PackageInfoFromPackageFile` then find and search the variable name.

sdk/core/ci.yml
danieljurek: Order does not matter now. It's defined at packing time and used dynamically at release time.

sdk/storage/ci.yml
danieljurek: I've removed it from ci.yml

eng/scripts/Pack-Crates.ps1
danieljurek: We've been using Set-Location everywhere instead of Push/Pop because it's possible to have some called or inner script which also pushes, encounters a failure, and doesn't pop and now the stack of directory contexts is messed up. This would mostly happen during local development. Using try/finally with Set-Location means that you set the location back to the original location, at least in the context of the script. It's a defense in depth. 

This whole location setting business is defense in depth. It might not be that useful but when the script finishes you're in the same place you were when you launched it (if running locally).

heaths: Okay, but there's still no reason to jump to the repo root. I can even cd into `target/` (gitignore'd) and run `cargo test -p azure_core` and it works. I'd prefer we not change locations in the first place. It's not necessary nor idiomatic.

danieljurek: I've opened https://github.com/Azure/azure-sdk-for-rust/issues/3231 to track this because it's in more places than here.

Thinking back to my own memory of where this cwd manipulation has been useful was mostly around things we did to isolate and pin cspell in the past. npm has `--prefix` which could obviate the need for it there, too.

eng/pipelines/templates/stages/archetype-rust-release.yml
```diff
@@ -84,33 +85,76 @@ stages:
...
-
-                    foreach ($owner in $missingOwners) {
-                      Write-Host "> cargo owner --add $owner $crateName"
-                      cargo owner --add $owner $crateName
-                    }
-                  displayName: Publish Crate
-                  env:
-                    CARGO_REGISTRY_TOKEN: $(azure-sdk-cratesio-token)
+                    Write-Host "Setting ArtifactName to empty"
+                    Write-Host "##vso[task.setvariable variable=ArtifactName;]"
```
weshaggard: This step should no longer be needed after the current changes.

PR #3214, opened 2025-10-16
Move optional de/serializers for base64 bytes to base64::option

### LINKED ISSUE #3213: Add base64::option module
Similar to `azure_core::time::*` and `azure_core::time::*::option`, we should move de/serializers for `Option<T>` into a new `base64::option` module and make the de/serializers non-optional for consistency.

heaths:
Resolves #3213, making `base64` congruent with `time`.

### DIFF
--- a/.vscode/cspell.json
+++ b/.vscode/cspell.json
@@ -42,6 +42,7 @@
     "cppvsdbg",
     "datalake",
     "datetime",
+    "deserializers",
     "devicecode",
     "docsrs",
     "doctest",


sdk/core/typespec_client_core/src/base64.rs
```diff
@@ -213,14 +201,134 @@ where
...
+pub fn serialize_url_safe<S, T>(to_serialize: &T, serializer: S) -> Result<S::Ok, S::Error>
 where
     S: Serializer,
     T: AsRef<[u8]>,
 {
-    let encoded = to_serialize.as_ref().map(encode_url_safe);
-    <Option<String>>::serialize(&encoded, serializer)
+    let encoded = encode_url_safe(to_serialize.as_ref());
+    String::serialize(&encoded, serializer)
+}
```
LarryOsterman, line 214: Nit: Tests?

LarryOsterman (approved):

sdk/core/typespec_client_core/src/base64.rs
heaths: I'll add a few, but want to get this in so @jhendrixMSFT can use the merged commit for the emitter.

PR #3215, opened 2025-10-16
Use vcpkg steps from eng/common

danieljurek:
Partial https://github.com/Azure/azure-sdk-tools/issues/12291

vcpkg recently released a breaking change that caused errors in 4 different repos. Since vcpkg has spread to wider use, this PR uses a unified vcpkg specified in `eng/common` which will make it easier to recover from breaking changes to vcpkg in the future.


danieljurek:

PR #3217, opened 2025-10-16
Added test proxy support

### LINKED ISSUE #3099: Added recorded test support to support the "test proxy" perf test feature.

LarryOsterman:
Fixes #3099

### DIFF
--- a/sdk/core/azure_core_test/src/perf/mod.rs
+++ b/sdk/core/azure_core_test/src/perf/mod.rs
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#![doc = include_str!("README.md")]
 #![cfg(not(target_arch = "wasm32"))]
 
 use crate::TestContext;
--- a/sdk/core/azure_core_test/src/perf/README.md
+++ b/sdk/core/azure_core_test/src/perf/README.md
@@ -49,103 +49,17 @@ A perf test has a name (`get_secret`, `list_blobs`, `upload_blob`, etc), a short
 
 Each perf test also has a set of command line options that are specific to the individual test, these are defined by a `PerfTestOptions` structure. It contains fields like help text for the option, activators
 
-Here is an example of test metadata for a performance test:
+An example of perf test metadata [can be found here](https://github.com/Azure/azure-sdk-for-rust/blob/e47a38f93e7ac2797754c103da7fe8b177e46365/sdk/keyvault/azure_security_keyvault_keys/perf/create_key.rs#L26C1-L41C1)
 
-```rust
-# use azure_core_test::perf::{PerfTestMetadata, PerfTestOption, PerfRunner, CreatePerfTestReturn};
-# struct FakePerfTest;
-# impl FakePerfTest {
-#   fn create_new_test(runner: PerfRunner) -> CreatePerfTestReturn { todo!() }
-#   fn test_metadata() -> PerfTestMetadata {
-PerfTestMetadata {
-    name: "get_secret",
-    description: "Get a secret from Key Vault",
-    options: vec![PerfTestOption {
-        name: "url",
-        display_message: "The URL of the Key Vault to use in the test",
-        mandatory: true,
-        short_activator: Some('u'),
-        long_activator: "vault-url",
-        expected_args_len: 1,
-        ..Default::default()
-    }],
-    create_test: Self::create_new_test,
-}
-# } }
-```
+This defines a test named `create_key` with a single required "vault_url" option.
 
-This defines a test named `get_secret` with a single required "vault_url" option.
-
-For this test, the `create_new_test` function looks like:
-
-```rust
-use std::sync::OnceLock;
-use azure_core_test::perf::{PerfRunner, CreatePerfTestReturn, PerfTestMetadata, PerfTest};
-use futures::FutureExt;
-# use azure_core_test::TestContext;
-# use std::sync::Arc;
-# struct MyServiceClient;
-# struct GetSecrets { vault_url: String, client: OnceLock<MyServiceClient>, random_key_name:OnceLock<String>};
-# #[async_trait::async_trait]impl PerfTest for GetSecrets {
-#  async fn setup(&self, ctx: Arc<TestContext>) -> azure_core::Result<()> {todo!()}
-#  async fn run(&self, ctx: Arc<TestContext>) -> azure_core::Result<()>{todo!()}
-#  async fn cleanup(&self, ctx: Arc<TestContext>) -> azure_core::Result<()>{todo!()}
-# }
-# impl GetSecrets {
-#   fn test_metadata() -> PerfTestMetadata {
-#       PerfTestMetadata {
-#           name: "get_secret",
-#           description: "Get a secret from Key Vault",
-#           options: Vec::new(),
-#           create_test: Self::create_new_test,
-#       }}
-fn create_new_test(runner: PerfRunner) -> CreatePerfTestReturn {
-    async move {
-        let vault_url_ref: Option<&String> = runner.try_get_test_arg("vault_url")?;
-        let vault_url = vault_url_ref
-            .expect("vault_url argument is mandatory")
-            .clone();
-        Ok(Box::new(GetSecrets {
-            vault_url,
-            random_key_name: OnceLock::new(),
-            client: OnceLock::new(),
-        }) as Box<dyn PerfTest>)
-    }
-    .boxed()
-}
-# }
-```
+An example of the `create_new_test` function [can be found here](https://github.com/Azure/azure-sdk-for-rust/blob/e47a38f93e7ac2797754c103da7fe8b177e46365/sdk/keyvault/azure_security_keyvault_keys/perf/get_key.rs#L42-L58)
 
 ### Test invocation
 
 The final piece of code which is necessary to run the performance tests is logic to hook up the tests with a test runner.
 
-```rust
-use azure_core_test::perf::{PerfRunner, CreatePerfTestReturn, PerfTestMetadata};
-# struct GetSecrets{}
-# impl GetSecrets{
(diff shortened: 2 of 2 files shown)


sdk/core/azure_core_test/src/proxy/policy.rs
```diff
@@ -118,13 +118,15 @@ impl Policy for RecordingPolicy {
...
 }
 
 impl AsHeaders for RecordingOptions {
     type Error = Infallible;
     type Iter = std::vec::IntoIter<(HeaderName, HeaderValue)>;
 
     fn as_headers(&self) -> Result<Self::Iter, Self::Error> {
-        self.skip.as_headers()
+        self.skip.as_headers()?;
+        self.remove_recording.as_headers()
```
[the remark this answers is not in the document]
LarryOsterman: Good catch.

sdk/keyvault/azure_security_keyvault_secrets/perf/get_secret.rs
```diff
@@ -10,15 +10,15 @@
 //!
 //! To run the test, use the following command line arguments:
 //!
-//! cargo test --package azure_security_keyvault_secrets --test performance_tests -- --duration 10 --parallel 20 get_secret -u https://<my_vault>.vault.azure.net/
+//! cargo test --package azure_security_keyvault_secrets --test perf -- --duration 10 --parallel 20 get_secret -u https://<my_vault>.vault.azure.net/
```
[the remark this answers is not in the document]
LarryOsterman: nope, it's a cargo test.

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -535,6 +544,23 @@ impl PerfRunner {
...
 }
 
+/// Instrument a client options appropriately for a perf test.
+///
+/// # Arguments:
+///
+/// - recording: Test recording.
+/// - client_options: Core client options
+///
+pub fn instrument_for_perf_test(recording: &Recording, client_options: &mut ClientOptions) {
```
heaths: Do we need this as opposed to just passing in the `RecordingOptions`? Just wondering if it's worth increasing the API surface.

sdk/core/azure_core_test/src/recording.rs
```diff
@@ -116,13 +116,17 @@ impl Recording {
...
     ///
     ///     let client = MyClient::new("https://azure.net", Some(options));
     ///     client.invoke().await
     /// }
     /// ```
-    pub fn instrument(&self, options: &mut ClientOptions) {
+    pub fn instrument(
+        &self,
+        options: &mut ClientOptions,
+        recording_options: Option<RecordingOptions>,
```
heaths: I'd like to see this more like our client methods and I admit I should've done that initially. So `RecordingOptions` should be `InstrumentOptions` and have an embedded `client_options: ClientOptions` (not `Option`). You could either put the `remove_header` field in there inline, or embed `RecordingOptions` as-is if that makes more sense, but since it's not a leaf node it should not be optional - just the fields.

heaths: ...and, actually, since it's a bool that leaf node probably doesn't need to be `Option<bool>` either. That's more for heap allocations or when we really need to know if it's unset or not. I don't think that's necessary here for an "internal" API.

heaths: Seeing how much code this would change *and how it would affect the mainline scenario*, what if we made the `options` param an `impl Into<ClientOptions>` and then implement `From<InstrumentOptions> for ClientOptions`? You'd just need to take out the extra `Option<RecordingOptions>` param in the call (basically just revert most of the files) and it'd just work and, in most cases, people just keep passing in `options.client_options`. Easy enough.

```rust
#[derive(Debug, Default, Clone)]
pub struct InstrumentOptions {
    client_options: ClientOptions,
    recording_options: RecordingOptions,
}
impl From<InstrumentOptions> for ClientOptions {
}
pub fn instrument(&self, options: ...
```

Oh, shoot. I forgot it's a `&mut`. I'm leaving this comment for posterity because maybe it sparks an idea. Maybe we could move the `instrument_perf` method here (let's not have "_for_" - that's not normal) as a pseudo-overload then so the normal use case doesn't need the extra `None`. In fact, that's not really what I'm worried about, but more so it sets a bad precedent for adding client options. If we come up with more use cases, we add more params or "overloads"? At least the latter doesn't sound so bad if they're bespoke.

sdk/core/azure_core_test/src/perf/mod.rs
heaths: I'd just call this `instrument_perf`. "test" is unnecessary, and prepositions beyond initial "to", "into", etc., are atypical. But I think this should move. See my long comment for why. I left it in tact even though I realized while writing it my initial idea wouldn't work.

heaths (approved):

sdk/core/azure_core_test/src/perf/mod.rs
LarryOsterman: I added this earlier today because I realized yesterday that the "RemoveRecording: false" experience felt unnatural (the header is a negative state and I felt it was important to preserve the header semantics rather than invert the option. 

Adding the "instrument for perf test" method hides this complexity and removes an opportunity for client teams to screw this up.

sdk/core/azure_core_test/src/proxy/policy.rs
```diff
@@ -118,13 +118,18 @@ impl Policy for RecordingPolicy {
 #[derive(Debug, Default)]
 pub struct RecordingOptions {
     pub skip: Option<Skip>,
+    pub remove_recording: Option<RemoveRecording>,
```
heaths: `Skip` is an enum, but you have a simple `bool`. Rather than expose the header, just take a `bool` here (not even `Option<bool>` and keep the header an internal implementation detail.

LarryOsterman: Fixed.

sdk/core/azure_core_test/src/proxy/policy.rs
```diff
@@ -118,13 +118,22 @@ impl Policy for RecordingPolicy {
 #[derive(Debug, Default)]
 pub struct RecordingOptions {
     pub skip: Option<Skip>,
+    pub remove_recording: Option<bool>,
```
heaths, line 121: Nit: probably doesn't need to be an `Option<bool>` unless you really need to know if it's unset. I find with bools (except for service models) that's rarely the case.

sdk/core/azure_core_test/src/proxy/policy.rs
```diff
@@ -118,13 +118,22 @@ impl Policy for RecordingPolicy {
...
-        self.skip.as_headers()
+        let it_self = self.skip.as_headers()?;
+
+        let it_remove = if let Some(remove) = self.remove_recording {
+            RemoveRecording(remove).as_headers()
+        } else {
+            Ok(Vec::new().into_iter())
+        }?;
+
+        Ok(it_self.chain(it_remove).collect::<Vec<_>>().into_iter())
```
heaths: Nit: maybe a little less confusing - and certainly fewer allocations - like (not checked):

```rust
let mut headers = self.skip.as_headers()?.to_vec();
if let Some(remove) = self.remove_recording { // or, again, just make it a `bool`
    headers.extend(RemoveRecording(remove).as_headers();
}
Ok(headers)
```

sdk/keyvault/azure_security_keyvault_keys/perf/create_key.rs
```diff
@@ -75,6 +75,7 @@ impl PerfTest for CreateKey {
 
         let mut client_options = KeyClientOptions::default();
         recording.instrument(&mut client_options.client_options);
+        perf::instrument_perf(recording)?;
```
heaths: Given this has to be exposed anyway, couldn't we just add it to `Recording` such that you just call `recording.instrument_perf()` (or, at that point, maybe something like `recording.start_perf()` makes sense)?

j7nw4r (commented):
LGTM

Commenting so that it is known I'm reviewing, though I want someone with a little more time to be the actual reviewer.

sdk/core/azure_core_test/src/proxy/policy.rs
LarryOsterman: I don't think we want to add the header if we're not in a recorded mode perf test, rather than always send it.

sdk/core/azure_core_test/src/proxy/policy.rs
LarryOsterman: FWIW (I did check a while ago), `Vec::new()` doesn't allocate anything, and I suspect the same is true for `into_iter`.

But I think this is simpler even if it does additional allocations when `skip` is `Some` (the call to `collect()` causes the vector to be initialized to a non-empty value while chaining the iterators doesn't necessarily cause any allocations..

sdk/keyvault/azure_security_keyvault_keys/perf/create_key.rs
LarryOsterman: I was wondering that over the weekend. If so, then I think it makes sense to have `Recording::instrument()` and `Recording::instrument_perf()` with the same parameters - if you're in a perf test, you call `Recording::instrument_perf()` otherwise you call `Recording::instrument()`.

The semantics are almost identical, the difference is that the 2nd also sets the magic header.

Oh, and `Recording::instrument_perf` returns a `Result<()>` :)

sdk/keyvault/azure_security_keyvault_keys/perf/get_key.rs
```diff
@@ -74,8 +74,7 @@ impl PerfTest for GetKey {
         let credential = recording.credential();
 
         let mut client_options = KeyClientOptions::default();
-        recording.instrument(&mut client_options.client_options);
-        perf::instrument_perf(recording)?;
+        recording.instrument_perf(&mut client_options.client_options)?;
```
heaths, line 77: FWIW, I didn't think it was bad before, but this is fine, too. 🙂

PR #3220, opened 2025-10-17
Add base64 tests

heaths:
Based on feedback from #3214

### DIFF
--- a/sdk/core/typespec_client_core/src/base64.rs
+++ b/sdk/core/typespec_client_core/src/base64.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-// cspell:ignore Hdvcmxk
+// cspell:ignore Hdvcmxk unpadded
 
 //! Base64 encoding and decoding functions.
 


sdk/core/typespec_client_core/src/base64.rs
```diff
@@ -332,3 +332,168 @@ pub mod option {
...
+        assert_eq!(encode_url_safe(b"foo"), "Zm9v");
+    }
+
+    #[test]
+    fn url_safe_decode() {
+        assert_eq!(
+            decode_url_safe("SGVsbG8sIHdvcmxkIQ").unwrap(),
+            b"Hello, world!"
+        );
+        assert_eq!(decode_url_safe("").unwrap(), b"");
```
LarryOsterman, line 382: It's a minor nit, but potentially an important case to cover: base64 encoding requires padding at the end with `=` characters to round up all strings to 4 octet boundaries (thus any base64 encoded value that is not a multiple of 4 is by definition invalid).

Base64url ("url safe") encoded values do *not* have this restriction, the `=` padding is optional.

See [section 3.2](https://www.rfc-editor.org/rfc/rfc4648#section-3.2) of the RFC for more details.

heaths: There are base64 and base64url tests already. I know the difference, but seems we're covering the cases here.

Issue #2475, opened 2025-04-10
Document dependency requirements


heaths:
Document the dependency requirements for what gets defined in the workspace `Cargo.toml`, and what crates' `Cargo.toml`s can do e.g., `dev-dependencies` to other projects in the repo should always be `path` dependencies, or inherited from the workspace if those are defined as `path` dependencies, as we do for `azure_core_test`.

I can think of two places we could put this, with my preference on the first because it's right in the repo and more likely to be seen.

1. Root `CONTRIBUTING.md` - either a short block of text or, if longer with examples, a link to another doc under the `doc/` folder e.g., `doc/dependencies.md`.
2. As part of the guidelines - probably in <https://azure.github.io/azure-sdk/rust_implementation.html>.

hallipr:
When publishing a package, I would expect the workspace dependency to already have the current, to-be-deployed version and a path:

```
[workspace.dependencies.azure_core]
version = "0.25.0"
path = "sdk/core/azure_core"
```

Looking at this dependency, any dependent package build would infer that it depends on unreleased `0.25.0` and that it should build against the package at path `sdk/core/azure_core`.

If we change our post-release PR to bump versions and sever path dependencies, we would leave dependent packages in a releasable, stable state while still allowing forward dev on the dependency package:

After release of azure_core@0.25.0:

sdk/core/azure_core/Cargo.toml
```diff
[package]
name = "azure_core"
-version = "0.25.0"
+version = "0.26.0"    <-- version bump
```

/Cargo.toml
```diff
[workspace.dependencies.azure_core]
version = "0.25.0"
- path = "sdk/core/azure_core"  
```

Now, this dependency looks like any other workspace dependency for a released package.  Even better, there's no conflict between the dependency version of azure_core (`0.25.0`) and the version being handled by -Zpackage-workspace (`0.26.0`)

heaths:
And then anyone - at least post-GA - who needs in-dev versions of `azure_core` or other core crates would take a `path` dependency and we'd do the normal release cycle: release a new `azure_core`, update the workspace, update the crates to use the now-updated workspace dependency on `azure_core`, and all is well.

But that's post-GA. Right now, everything is in flux.

PR #3227, opened 2025-10-17
Output debug lines from update-pathversions.rs

hallipr:
Add version update logs to update-pathversions.rs to assist in debugging version update errors in our release pipelines

### DIFF
--- a/eng/scripts/update-pathversions.rs
+++ b/eng/scripts/update-pathversions.rs
@@ -170,7 +170,7 @@ fn update_package_versions(
         .and_then(Item::as_table)
         .and_then(|table| table.get("name"))
         .and_then(Item::as_str)
-        .unwrap_or("<unknown>")
+        .unwrap_or("<workspace>")
         .trim_matches('"')
         .trim()
         .to_string();


eng/scripts/update-pathversions.rs
```diff
@@ -15,73 +15,127 @@ use std::{env, error::Error, fs, path::PathBuf};
...
+        );
 
         // if the toml file has a workspace table, update the workspace table
         if let Some(workspace) = toml_file.document.get_mut("workspace") {
-            if let Some(table) = workspace.as_table_mut() {
-                update_package_versions(table, &package_versions, should_add);
+            // print out that we're working on a workspace
+            if let Some(t
+    able) = workspace.as
+   _table_mut() {,
```
heaths: I agree with what Copilot has been saying to this point, but this comment is misleading. Clearly this should be in one line.

There's a lot wrong with the formatting of this script. I recommend removing the header (lsp support will land soon, it looks like, so you don't need to) and get rust-analyzer to format this, check for obvious errors, etc., before putting the header back in.

heaths (changes requested):
You've got errant line breaks all over. Not sure what happened, but see my comment below before I review any more. And for places where there aren't line breaks that Copilot commented on, I agree with its comments to that point as well.

eng/scripts/update-pathversions.rs
hallipr: I think there was a race between me saving the file and copilot updating it given our global setting of [`"editor.formatOnSave": true,`](https://github.com/azure/azure-sdk-for-rust/blob/main/.vscode/settings.json)

eng/scripts/update-pathversions.rs
```diff
@@ -103,38 +139,84 @@ fn get_package_versions(toml_files: &Vec<TomlInfo>) -> Vec<(String, String, bool
...
+    let mut updated = false;
     for (table_name, table) in dependency_tables {
         for (package, version, is_publish_disabled) in package_versions {
             if let Some(dependency) = table.get_mut(package) {
                 // azure_idenentity will only be a transitive dev-dependency
-                let should_add = add && table_name != "dev-dependencies" && !is_publish_disabled && package != "azure_identity";
+                let should_add = add
+                    && table_name != "dev-dependencies"
+                    && !is_publish_disabled
+                    && package != "azure_identity";
```
heaths, line 187: I wonder if we should even do this check. It should only ever be a `dev-dependency` anyway.

That said, do we make sure we only process `Cargo.toml`s that are part of the workspace? We have a few test projects that must use a path dependency but don't get published. They use crates as `dependencies` but will also never published which, I suppose, would be caught by the line above.

eng/scripts/update-pathversions.rs
```diff
@@ -103,38 +139,84 @@ fn get_package_versions(toml_files: &Vec<TomlInfo>) -> Vec<(String, String, bool
...
+                            .unwrap_or("");
+                        if current_version != version {
+                            dependency["version"] = value(version);
+                            println!(
+                                "set {} to version {} in {} {}",
+                                package, version, crate_name, table_name
+                            );
+                            updated = true;
+                        }
+                    } else if should_add {
```
heaths, line 206: Post-GA, shouldn't we remove the version or is the assumption here that the service crate devs would've already had to do that after `azure_core`, for example, shipped the week before?

eng/scripts/update-pathversions.rs
hallipr: We need to see when we would call the script with `--add`

When do we want to inject versions into path dependencies that didn't already have them.  I'm remembering 'add' mode existing so we could have path-only dependencies in the repo that get converted to path + version dependencies in ci.

If that's the case, then it doesn't really matter where all we do that.  We'll be adding a version requirement to a dependency when we're guaranteed that the version at the path matches the required version and we wouldn't commit the change.

eng/scripts/update-pathversions.rs
hallipr: I think we need to spec out our dependency version scheme.  I can't remember what we've said.

hallipr: From what I recall, if someone wants to depend on the yet-to-be-released version of azure_core, they'll take a path-only dependency, and we'll make that cargo packable by using `update-packageversions add`.  Once azure_core is released, I believe our post-release PR will pin those dependencies to the version we just released. 

We should decide of we want to change local path dependencies to pinned workspace dependencies.

If they want to go back to an unreleased path dependency, they'll do that manually

heaths: > I think we need to spec out our dependency version scheme. I can't remember what we've said.

Hence the old issue, https://github.com/Azure/azure-sdk-for-rust/issues/2475

heaths: > Once azure_core is released, I believe our post-release PR will pin those dependencies to the version we just released.

Too presumptuous. The crate only needs to once they are preparing to release, so we should have checks then. You're assuming they were waiting for the next `azure_core` but may actually be a long-lead issue.

Issue #3170, opened 2025-10-09
Cosmos: Build `RetryPolicy` For Cross Regional Retry


kundadebdatta:
### Acceptance Criteria:

- Build a RetryPolicy For Cross Regional Retry.

PR #3230, opened 2025-10-18
Cosmos: Add Basic Retry Policy for Throtteled Requests

kundadebdatta:
## Description

Adds structs and traits to define the structure of the cross regional retry logic. Adds a basic implementation of a `ResourceThrottleRetryPolicy` to retry the throtteled requests.

## Changes

- Added handler module and defined the structure of `RetryHandler`.
- Added an initial version of `BackoffRetryHandler`.
- Added `retry_policies` module and implemented `ResourceThrottleRetryPolicy`.
- Added comprehensive unit tests covering functionality and method chaining.
- Updated documentation with usage examples
- Updated CHANGELOG.md

All existing functionality remains unchanged and fully backward compatible.

----
**Additional instructions:**

> Make sure you structure your commits in a sensible way and include a concise but descriptive PR description. PREFER concise descriptions and code. The team knows how this code should work, they don't need explanations of it. The code should be largely self-explanatory.

Fixes Azure/azure-sdk-for-rust#3170

---

💬 Share your feedback on Copilot coding agent for the chance to win a $200 gift card! Click [here](https://survey3.medallia.com/?EAHeSx-AP01bZqG0Ld9QLQ) to start the survey.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
@@ -12,7 +12,7 @@ use azure_core::time::Duration;
 /// maximum retry attempts and cumulative wait time limits. It's designed to handle
 /// Azure Cosmos DB throttling scenarios where the service limits request rates.
 /// # Example
-/// ```
+/// ```ignore
 /// use azure_data_cosmos::retry_policies::resource_throttle_retry_policy::ResourceThrottleRetryPolicy;
 ///
 /// // Create a policy with 3 max retries, 100 second max wait, and backoff factor of 2
@@ -36,7 +36,7 @@ impl ResourceThrottleRetryPolicy {
     /// * `backoff_delay_factor` - Multiplier for exponential backoff delay
     ///
     /// # Example
-    /// ```
+    /// ```ignore
     /// use azure_data_cosmos::retry_policies::resource_throttle_retry_policy::ResourceThrottleRetryPolicy;
     ///
     /// // Create a policy with 5 retries, 120 second max wait, and backoff factor of 3


sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
```diff
@@ -0,0 +1,149 @@
...
+
+/// Trait defining the interface for retry handlers in Cosmos DB operations
+///
+/// This trait provides a contract for implementing retry logic that wraps HTTP requests
+/// with automatic retry capabilities. Implementations can inject custom retry policies
+/// and handle both transient failures (errors) and non-success HTTP responses.
+#[allow(dead_code)]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+pub trait AbstractRetryHandler: Send + Sync {
```
analogrelay: Just `RetryHandler` would be the style here. Rust code generally doesn't refer to OOP terms like `Abstract`.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+/// an operation should be retried and how long to wait before the retry attempt.
+///
+/// # Fields
+/// * `should_retry` - Whether the operation should be retried
+/// * `backoff_time` - Duration to wait before retrying (meaningful only if `should_retry` is true)
+#[derive(Debug, Clone)]
+pub struct ShouldRetryResult {
+    pub should_retry: bool,
+    pub back_off_time: Duration,
+}
```
analogrelay, line 30: The more idiomatic way to implement this in Rust is to use an [`enum`](https://doc.rust-lang.org/book/ch06-01-defining-an-enum.html). In Rust, enum variants can carry values, which you can use to implement the back-off time:

```rust
pub enum RetryResult {
    /// Indicates that the operation should not be retried.
    DoNotRetry,
    /// Indicates that the operation should be retried after waiting for the duration specified in `after`.
    Retry { after: Duration }
}
```

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+    /// assert!(result.should_retry);
+    /// assert_eq!(result.back_off_time, Duration::from_secs(5));
+    /// ```
+    pub fn retry_after(backoff: Duration) -> Self {
+        Self {
+            should_retry: true,
+            back_off_time: backoff,
+        }
+    }
+}
```
analogrelay: You wouldn't really need this if you're using an enum, you'd have it already in the standard construction syntax:

```rust
RetryResult::DoNotRetry
RetryResult::Retry { after: backoff }
```

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -65,9 +70,20 @@ impl CosmosPipeline {
         request: &mut Request,
         resource_link: ResourceLink,
     ) -> azure_core::Result<RawResponse> {
-        let ctx = ctx.with_value(resource_link);
-        let r = self.pipeline.send(&ctx, request, None).await?;
-        Ok(r)
+        // Clone pipeline and convert context to owned so the closure can be Fn
+        let pipeline = self.pipeline.clone();
+        let ctx_owned = ctx.with_value(resource_link).into_owned();
```
analogrelay, line 76: I think you could avoid this `.into_owned()` (which may force an allocation) by having the retry handler (and the sender closure) take it as a parameter, and flowing it down that way rather than closing over it.

sdk/cosmos/azure_data_cosmos/src/routing/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+
+use std::{collections::HashMap, sync::Arc};
+use url::Url;
+
+/// Type alias matching C# `ReadOnlyCollection<Uri>` semantics in a lightweight form.
+/// We use `Arc<[Url]>` for cheap cloning and slice immutability.
+pub type ReadOnlyUrlCollection = Arc<[Url]>;
+
+/// Type alias representing `ReadOnlyDictionary<string, Uri>`.
+pub type ReadOnlyLocationMap = Arc<HashMap<String, Url>>;
```
analogrelay: I don't think we need these. First off, we shouldn't just be porting C# concepts over to Rust since we're aiming to make an idiomatic Rust SDK. Second, the idea of separate "read-only collection" types in Rust isn't really necessary. Rust already enforces that only the owner of a collection, or the holder of a `&mut` reference, can modify it, so the only place you really need to enforce "read-only" behavior in a collection type is if you want to restrict even the owner from modifying the collection and that's not usually necessary.

sdk/cosmos/azure_data_cosmos/tests/retry_policy_429_test.rs
analogrelay, line 1: These are unit tests, by my reading, since they use mock HTTP responses. In Rust, unit tests almost always live _inside_ the same module as the code being tested, often in a `mod tests { ... }` block at the end of the file (https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html)

The `tests/` directory at the root of the package is specifically for integration tests (using a real Cosmos connection)

sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -22,6 +22,7 @@ serde.workspace = true
 tracing.workspace = true
 typespec_client_core = { workspace = true, features = ["derive"] }
 url.workspace = true
+tokio = { workspace = true, features = ["time"] }
```
analogrelay: We should avoid taking a dependency on `tokio` in the product code itself (we can use it in tests). Tokio is just one of a few different async runtimes that a Rust application could use, and if we depend directly on it, it can make it very difficult to use the SDK properly in applications that use a runtime other than tokio.

If you have a scenario where you _need_ tokio because you actually need to spawn a task, the `azure_core` package provides some abstractions that you should use instead. Specifically the [`AsyncRuntime`](https://docs.rs/azure_core/latest/azure_core/async_runtime/trait.AsyncRuntime.html) trait, and [`get_async_runtime`](https://docs.rs/azure_core/latest/azure_core/async_runtime/fn.get_async_runtime.html) global function. If there's something you need to do that `AsyncRuntime` doesn't cover, we can add it to the abstraction.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
```diff
@@ -0,0 +1,149 @@
...
+                        if !retry_result.should_retry {
+                            return result;
+                        }
+
+                        if retry_result.back_off_time > Duration::ZERO {
+                            tracing::warn!(
+                                "Retry back off requested for {:?}.",
+                                retry_result.back_off_time
+                            );
+                            sleep(retry_result.back_off_time).await;
```
analogrelay: I mention this later on, but you should use `get_async_runtime` and `AsyncRuntime::sleep` (from `azure_core`) here, to avoid coupling us to tokio.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
```diff
@@ -0,0 +1,149 @@
...
+                        return result;
+                    }
+                }
+                Err(err) => {
+                    let retry_result = retry_policy.should_retry_exception(err).await;
+                    if !retry_result.should_retry {
+                        return result;
+                    }
+                    if retry_result.back_off_time > Duration::ZERO {
+                        sleep(retry_result.back_off_time).await;
```
analogrelay: Same here, we shouldn't use `tokio::time::sleep` directly

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /// Checks if the given HTTP status code indicates resource throttling
+    ///
+    /// # Arguments
+    /// * `status_code` - The HTTP status code to check
+    ///
+    /// # Returns
+    /// `true` if the status code is 429 (TooManyRequests), `false` otherwise
+    fn is_valid_throttle_status_code(&self, status_code: StatusCode) -> bool {
```
analogrelay: I think the more common Rust style here would be to just do this check inline rather than define a function for it, particularly since we only do it in two places and it's not more concise

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// if the cumulative wait time would exceed the configured maximum.
+    ///
+    /// # Arguments
+    /// * `retry_after` - Optional duration suggested by the server (from Retry-After header)
+    ///
+    /// # Returns
+    /// A tuple of (should_retry: bool, retry_delay: Duration)
+    /// - `should_retry`: true if retry is allowed within timing constraints
+    /// - `retry_delay`: The calculated delay duration (with backoff applied)
+    fn check_if_retry_needed(&self, retry_after: Option<Duration>) -> (bool, Duration) {
```
analogrelay: In general, avoid tuple returns and prefer defining a type if you need one. Types are cheap and much clearer. In fact, in this one I think you can just use the `RetryResult` enum I describe above.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,215 @@
...
+        let mut retry_delay = retry_after.unwrap_or(Duration::from_secs(0));
+        if self.backoff_delay_factor > 1 {
+            retry_delay *= self.backoff_delay_factor;
+        }
```
analogrelay: The 5sec default logic should live up here, and there's no need to have the `if` check, since a `backoff_delay_factor` of `1` wouldn't affect the multiplication anyway (and I suspect the compiler is already optimizing this `if` away anyway.


```suggestion
        let retry_delay = retry_after.unwrap_or(Duration::from_secs(5)) * self.backoff_delay_factor;
```

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,215 @@
...
+        let cumulative = self.cumulative_retry_delay.load(Ordering::Relaxed) as u64;
+        let new_cumulative = cumulative + retry_delay.as_millis() as u64;
+        if retry_delay < self.max_wait_time
+            && new_cumulative <= self.max_wait_time.as_millis() as u64
+        {
+            if retry_delay == Duration::ZERO {
+                retry_delay = Duration::from_secs(5);
+            }
```
analogrelay: No longer needed if you take the suggestion above.

```suggestion
```

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
analogrelay: Broader Rust advise: It's actually pretty rare that you need `let mut`, particularly for primitive types (numbers, etc.). One of the main reasons you might _think_ you need `let mut` but actually don't is if you're _replacing_ a value with a modified version of that value. This is perfectly valid Rust code:

```rust
let retry_delay = retry_after.unwrap_or(Duration::from_secs(5));
let retry_delay = retry_delay * self.backoff_delay_factor;
```

You're actually defining two variables, the first `retry_delay` is just the `retry_after` value, then you "shadow" that by defining a **new** `retry_delay` value. They're actually two separate values, you're just "redefining" the **name** `retry_delay` to point to a new variable. It's a lot like the scenario you have below where you have `cumulative` and then you define `new_cumulative` and never use `cumulative` again. That's the exact same _behavior_ as the code in my snippet above, it's just the two variables have the same name.

The place you need `mut` is if you're **mutating** the value in place. Think of a `Vec::new()` where you call a method on it and the vector changes in-place. Any time you're taking a value, modifying it, and then replacing it, you don't actually need `mut`. The optimizer is quite good at these cases so always feel free to introduce as many variables as you need and it will collapse them down to as few actual memory locations as possible.

The main places I've found where you _do_ need `let mut`:

* You're conditionally changing a value (Your original `let mut` followed by an `if` _did_ need the `mut`). But there's often a way to restructure that to avoid the need for `mut`
* You're modifying a value in a loop
* You're modifying a value in-place (like pushing items to a `Vec`).

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,215 @@
...
+/// // Create a policy with 3 max retries, 100 second max wait, and backoff factor of 2
+/// let policy = ResourceThrottleRetryPolicy::new(3, 100, 2);
+/// ```
+#[derive(Debug)]
+pub struct ResourceThrottleRetryPolicy {
+    max_attempt_count: usize,
+    backoff_delay_factor: u32,
+    max_wait_time: Duration,
+    current_attempt_count: AtomicUsize,
+    cumulative_retry_delay: Arc<AtomicUsize>,
```
analogrelay: The state management here isn't quite right. The `ResourceThrottleRetryPolicy` is a singleton for the entire application, but most of this state is per-request.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+    ///
+    /// # Arguments
+    /// * `response` - The HTTP response received from the server
+    ///
+    /// # Returns
+    /// `ShouldRetryResult` indicating:
+    /// - `should_retry`: Whether the operation should be retried
+    /// - `backoff_time`: How long to wait before retrying (if `should_retry` is true)
+    async fn should_retry_response(&self, response: &RawResponse) -> ShouldRetryResult;
+}
```
analogrelay: I think this structure of a before send/should retry method is likely to get us in to state-management problems.  The `Policy`s used in `azure_core` for the standard HTTP pipeline use a fairly simple middleware pattern that I think would be more appropriate for us to use here. In those, each policy is given the request and expected to return the response. But in order to do that, they are also given a list of all the remaining policies in the pipeline so they can "pass along" the request to the next policy.

This does basically mean embedding the `RetryHandler` logic to handle sleeping during backoff in each policy, but we can solve that duplication by writing helper functions that take care of the common logic.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+        &self.config
+    }
+
+    /// Returns the appropriate retry policy based on the request
+    ///
+    /// This method examines the request headers and method to determine which
+    /// retry policy should be used for this specific request.
+    /// # Arguments
+    /// * `request` - The HTTP request to analyze
+    pub fn get_policy_for_request(&self, _request: &Request) -> Arc<dyn RetryPolicy> {
```
analogrelay: I think a `RetryPolicy` should be constructed for each request. That seems to me like the cleanest way to have clean per-request state. In Rust, **most** allocations will be done on the stack, and are thus very cheap, so don't be afraid to create new instances of structs. Heap allocations (which are what we generally try to avoid in managed languages like C#/Java) only happen when you're using wrappers like `Box`, `Arc`, or `Vec`, or for very specific types, like `String`, where heap-allocation is essential. Creating a new value of a struct containing only primitive (number) types and returning it is pretty cheap.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+    pub fn resource_throttle_policy(&self) -> Arc<ResourceThrottleRetryPolicy> {
+        self.resource_throttle_policy.clone()
+    }
+
+    /// Returns the resource throttle retry policy as a trait object
+    ///
+    /// Useful when you need to work with the DocumentClientRetryPolicy trait.
+    pub fn resource_throttle_policy_dyn(&self) -> Arc<dyn RetryPolicy> {
+        self.resource_throttle_policy.clone()
+    }
```
analogrelay: I don't really think we need these. If someone wants to create a `ResourceThrottleRetryPolicy`, they can use it's constructor. We can do the same in `get_policy_for_request`.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+    /// whether the operation should be retried and how long to wait before retrying.
+    ///
+    /// # Arguments
+    /// * `err` - The error that occurred during the operation
+    ///
+    /// # Returns
+    /// `ShouldRetryResult` indicating:
+    /// - `should_retry`: Whether the operation should be retried
+    /// - `backoff_time`: How long to wait before retrying (if `should_retry` is true)
+    async fn should_retry_exception(&self, err: &azure_core::Error) -> ShouldRetryResult;
```
analogrelay: Rust doesn't have exceptions, this should be `should_retry_error`. Honestly, we should just have a `should_retry` that takes `azure_core::Result<RawResponse>` and checks for Ok/Err itself.

analogrelay (changes requested):
Definitely some things to look in to. In general, there's a fair bit of "Object-Oriented" design in here that I don't think translates well into Rust. We should try to avoid translating C#/Java concepts in to Rust and look at a Rust-first approach. Rust code typically avoids a lot of virtual dispatch and high-level abstractions.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
kundadebdatta: Agreed. Updated the trait naming.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
kundadebdatta: Agreed. Refactored the code.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
kundadebdatta: Agreed. Refactored the code.

sdk/cosmos/azure_data_cosmos/src/routing/mod.rs
kundadebdatta: Yes yuo are right. These are not needed anymore. Removed.

sdk/cosmos/azure_data_cosmos/Cargo.toml
kundadebdatta: Removed the `tokio` dependency, since azure core has already have similar `time` feature.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
kundadebdatta: Refactored.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
kundadebdatta: Refactored.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Refactored.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Refactored.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Thanks for the insight here @analogrelay. I have refactored the code.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
kundadebdatta: Yes - And you are right. I was not in a favor of creating the retry policy per request to avoid any memory allocation. However, reading further, I think this is safe to do.

Refactored the code to reflect the changes.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
kundadebdatta: Agreed. Removed.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
kundadebdatta: Refactored.

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
```diff
@@ -179,6 +188,10 @@ impl LocationCache {
         }
     }
 
+    pub fn can_use_multiple_write_locations(&mut self) -> bool {
```
tvaron3, line 191: Is this being used ? This would be true even if there is only one write endpoint.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -37,16 +39,19 @@ impl CosmosPipeline {
...
+        let pipeline = azure_core::http::Pipeline::new(
+            option_env!("CARGO_PKG_NAME"),
+            option_env!("CARGO_PKG_VERSION"),
+            client_options,
+            Vec::new(),
+            vec![Arc::new(auth_policy)],
+            None,
+        );
+
         CosmosPipeline {
```
tvaron3, line 52: How does this fit into the azure core retry policies? We should also double check these as the behavior from core is not necessarily what we want. We have overriden some of the retry configurations from core in python sdk

sdk/cosmos/azure_data_cosmos/tests/retry_policy_429_test.rs
kundadebdatta: Thanks for the feedback @analogrelay. I have refactored the code to co-locate the unit tests.

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
kundadebdatta: This will be required for future implimentations, which are planned next.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
heaths: We intend to have the same conceptual design as other Azure SDK languages. If you can customize retry behavior in other languages, within reason we'd prefer to replicate that in core as well. Adding policies in clients should be relatively rare.

kundadebdatta: The reason we wanted to keep the cosmos retry policies separate from azure_core general retry policies are:

- We have a lot of cross regional routing logic, based of different status codes.
- We will need to inject our endpoint manager, location caches, pk-range based routing into the retry layer to route the request correctly.

I understand that there is already a base retry policy that exists today in azure_core, and let me explore a bit more to check if they can be leveraged and overridden to modify the behaviors.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
```diff
@@ -0,0 +1,142 @@
...
+use crate::retry_policies::resource_throttle_retry_policy::ResourceThrottleRetryPolicy;
+use crate::retry_policies::{RetryPolicy, RetryResult};
+use async_trait::async_trait;
+use azure_core::http::{request::Request, RawResponse};
+use std::sync::Arc;
+use typespec_client_core::async_runtime::get_async_runtime;
+
+// Helper trait to conditionally require Send on non-WASM targets
+#[cfg(not(target_arch = "wasm32"))]
+pub trait CosmosConditionalSend: Send {}
```
analogrelay: This should just be `ConditionalSend`. It's private to our crate, and even if it was a public API, we _generally_ avoid prefixing types with the crate name (i.e. `Cosmos`) in Rust. There are exceptions (`CosmosClient`) but it's not like in .NET where we tend to prefix a lot of types with a product/package name like `Cosmos`

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
```diff
@@ -0,0 +1,142 @@
...
+        Sender: Fn(&mut Request) -> Fut + Send + Sync,
+        Fut: std::future::Future<Output = azure_core::Result<RawResponse>> + CosmosConditionalSend;
+}
+
+/// Concrete retry handler implementation with exponential back off.
+/// This handler provides automatic retry capabilities for Cosmos DB operations using
+/// a pluggable retry policy system. It wraps HTTP requests with intelligent retry logic
+/// that handles both transient network errors and HTTP error responses.
+#[derive(Debug, Clone)]
+pub struct BackOffRetryHandler {}
```
analogrelay: nit: It's generally idiomatic to just remove the `{}` entirely if you're writing a type with no fields:


```suggestion
pub struct BackOffRetryHandler;
```

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
```diff
@@ -0,0 +1,142 @@
...
+        Self {}
+    }
+
+    /// Returns the appropriate retry policy based on the request
+    ///
+    /// This method examines the underlying operation and resource types and determines
+    /// retry policy should be used for this specific request.
+    /// # Arguments
+    /// * `request` - The HTTP request to analyze
+    pub fn get_policy_for_request(&self, _request: &Request) -> Arc<dyn RetryPolicy> {
```
analogrelay: nit: `policy_for_request`. "Getters" are usually just unprefixed in Rust.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
```diff
@@ -0,0 +1,142 @@
...
+
+/// Trait defining the interface for retry handlers in Cosmos DB operations
+///
+/// This trait provides a contract for implementing retry logic that wraps HTTP requests
+/// with automatic retry capabilities. Implementations can inject custom retry policies
+/// and handle both transient failures (errors) and non-success HTTP responses.
+#[allow(dead_code)]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+pub trait RetryHandler: Send + Sync {
```
analogrelay, line 31: As mentioned offline, I think we can just have our retry handler be an `azure_core::http::policies::Policy`. Even if we don't want it to be a Policy for some reason, we generally avoid introducing traits until we actually have a significant need for them (it's not like in .NET/Java where interfaces/abstract classes get used often for "testability")

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
analogrelay: Now that you're creating an instance of this per-request, I don't think these need to be `Arc`s or `AtomicUsize`s at all. Regular `usize` should be fine, since the instance is being used entirely within a single request "thread"

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// # Arguments
+    ///
+    /// * `response` - A reference to the result of the HTTP request attempt. This can be:
+    ///   - `Ok(RawResponse)` - A successful HTTP response (which may still indicate an error via status code)
+    ///   - `Err(azure_core::Error)` - A network or client-side error
+    ///
+    /// # Returns
+    ///
+    /// A `RetryResult` indicating the retry decision.
+    async fn should_retry(&self, response: &azure_core::Result<RawResponse>) -> RetryResult;
```
analogrelay: Both of these probably need to take `&mut self` so that the implementation will be able to modify it's own state. A function that takes `&self` cannot mutate any values on `self` unless they provide "interior mutability" (such as Atomic types, RefCell, locks like Mutex, etc.).

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+/// in Azure Cosmos DB operations. Implementers can define custom retry behavior for
+/// both exceptions (errors) and HTTP responses based on their specific requirements.
+#[async_trait]
+pub trait RetryPolicy: Send + Sync {
+    /// Called before sending a request to allow policy-specific modifications
+    ///
+    /// This method is invoked immediately before each request is sent (including retries).
+    /// # Arguments
+    /// * `request` - Mutable reference to the HTTP request being sent
+    fn on_before_send_request(&self, request: &mut Request);
```
analogrelay: If you define an implementation, it becomes optional and only types that choose to override the default implementation need to specify it. It's basically the one and only place Rust has overrides (but there's no way to call the base implementation if you do override it :))

```suggestion
    fn on_before_send_request(&self, request: &mut Request) { }
```

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
analogrelay: Ah, and because all the `RetryPolicy` methods take `&self`, you end up having to use atomic types. That's the Rust compiler giving you a hint that something is a little wrong about how you're handling mutability :).

If a method can mutate fields on `self`, it needs to take it in as `&mut self`. Rust will then guarantee that as long as the caller is using safe Rust (i.e. not inside an `unsafe` block), NO other running code is holding a mutable reference.

I suggest a quick review of [References and Borrowing](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html) in the Rust Book! Getting the hang of these is a key part of understanding the magic of Rust :)

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,440 @@
...
+    ///
+    /// This method encapsulates the shared retry decision logic used by both
+    /// `should_retry_error` and `should_retry_response`. It provides a centralized
+    /// place for retry decision-making to ensure consistency across different failure modes.
+    fn should_retry_with_backoff(&self, retry_after: Option<Duration>) -> RetryResult {
+        let attempt = self.current_attempt_count.load(Ordering::Relaxed);
+        if attempt < self.max_attempt_count {
+            let retry_result = self.check_if_retry_needed(retry_after);
+
+            if matches!(retry_result, RetryResult::Retry { .. }) {
```
analogrelay: nit: I'd consider adding an `is_retry()` method to `RetryResult` that does this, just for readability. But only if you're making other changes too :)

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,440 @@
...
+    /// # Note
+    /// Currently uses a fixed 500ms base retry delay. Future versions may parse
+    /// the `x-ms-retry-after-ms` header from the error context.
+    fn should_retry_error(&self, err: &azure_core::Error) -> RetryResult {
+        // Check if the error has an HTTP status code and if it's a valid throttle status
+        // Early return for invalid or missing status codes
+        let status = err
+            .http_status()
+            .filter(|&s| s == StatusCode::TooManyRequests);
+        if status.is_none() {
```
analogrelay: You can just use `==` here:


```suggestion
        if err.http_status() == Some(StatusCode::TooManyRequests) {
```

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
analogrelay: Also, I'd suggest `before_send_request`. The "on" prefix is much less common in Rust.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,440 @@
...
+    /// A `RetryResult` indicating the retry decision.
+    async fn should_retry(&self, response: &azure_core::Result<RawResponse>) -> RetryResult {
+        match response {
+            Ok(resp) => {
+                if resp.status().is_server_error() || resp.status().is_client_error() {
+                    return self.should_retry_response(resp);
+                }
+
+                RetryResult::DoNotRetry
+            }
```
analogrelay: Whenever you have a `match` arm that immediately does an `if`, consider using a ["match guard"](https://doc.rust-lang.org/book/ch19-03-pattern-syntax.html#extra-conditionals-with-match-guards):


```suggestion
            Ok(resp) if resp.status().is_server_error() || resp.status().is_server_error() => {
                self.should_retry_response(resp)
            },
            Ok(_) => RetryResult::DoNotRetry,
```

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,440 @@
...
+
+        // Simulate a 429 TooManyRequests response
+        let response_429 = create_mock_response(StatusCode::TooManyRequests);
+
+        // First retry attempt
+        let result1 = policy.should_retry(&response_429).await;
+        match result1 {
+            RetryResult::Retry { after } => {
+                assert!(after > Duration::ZERO, "Should have backoff time");
+                println!("First retry - backoff time: {:?}", after);
```
analogrelay, line 225: I suggest putting some [`tracing::trace!`](https://docs.rs/tracing/latest/tracing/#shorthand-macros) logs _in_ the policy itself, since that will help both testing scenarios **and** observability later.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,440 @@
...
+                after
+            }
+            RetryResult::DoNotRetry => panic!("Should retry on second 429 response"),
+        };
+
+        // Extract backoff1 for comparison
+        let backoff1 = match result1 {
+            RetryResult::Retry { after } => after,
+            _ => panic!("Expected retry result"),
+        };
```
analogrelay: There are a few more compact ways to express "two-arm" matches (where one arm is just `_`). I'd suggest this:

```rust
let RetryResult::Retry { after: backoff1 } = result1 else {
    panic!("Expected retry result")
}
```

That will match on `RetryResult::Retry` and then bind the `after` field to the local variable `backoff1`. If the result doesn't match that pattern, the `else` block will be taken. It's the same code as you have above, just a slightly more compact and idiomatic way to write it.

There's also a similar inverted syntax:

```rust
let backoff1 = if let RetryResult::Retry { after } = value {
    after
} else {
    panic!("Expected retry result");
}
```

That one is just FYI, I think the first syntax is preferable in this case.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,440 @@
...
+        }
+
+        // Fourth attempt should NOT retry (exceeded max_attempt_count of 3)
+        let result4 = policy.should_retry(&response_429).await;
+        match result4 {
+            RetryResult::DoNotRetry => {
+                println!("Fourth attempt - should not retry");
+            }
+            RetryResult::Retry { .. } => panic!("Should NOT retry after exceeding max attempts"),
+        }
```
analogrelay: This can be an `assert_eq!` I think:


```suggestion
        assert_eq!(result4, RetryResult::DoNotRetry);
```

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,440 @@
...
+
+        for (status, description) in test_cases {
+            let response = create_mock_response(status);
+            let result = policy.should_retry(&response).await;
+            match result {
+                RetryResult::DoNotRetry => {
+                    // Success - should not retry on client errors
+                }
+                RetryResult::Retry { .. } => panic!("Should NOT retry on {}", description),
+            }
```
analogrelay: Another place for an `assert_eq!`

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,440 @@
...
+        // Simulate a 200 OK response (success)
+        let response_200 = create_mock_response(StatusCode::Ok);
+
+        let result = policy.should_retry(&response_200).await;
+        match result {
+            RetryResult::DoNotRetry => {
+                // Success - should not retry
+            }
+            RetryResult::Retry { .. } => panic!("Should NOT retry on successful response"),
+        }
```
analogrelay: Another place for an `assert_eq!`

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,440 @@
...
+        headers.insert("content-type", "application/json");
+
+        azure_core::Result::from(Ok(RawResponse::from_bytes(
+            status,
+            headers,
+            r#"{"id":12,"name":"Too Many Requests"}"#,
+        )))
+    }
+    #[tokio::test]
+    async fn test_retry_policy_handles_429_status() {
```
analogrelay: nit (across all these tests): The `test_` prefix on test functions isn't usually used in Rust.

analogrelay (commented):
Definitely better state management. We talked about seeing if we can rework this to using `azure_core::http::policies::Policy`. I'd still like you to look at that, but the rest looks good aside from some style/idiomatic suggestions.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
heaths: Thing is, you can't remove the default retry policy. At best, you could have your client set it to `none` - overriding any policy options a user might set (or honor them as a copy, but then overwrite them) - and inject your own.

We have the same design as any other Azure SDK language, though. Did you need a custom retry policy for other languages? Are there gaps in our core design that necessitate some additional options perhaps?

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
kundadebdatta: This is now refactored to `ConditionalSend`.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
kundadebdatta: As discussed offline, we would like to stick to the current approach of keeping the entire pipeline wrapped around our retry policy. Adapting to `Policy` gives us very minor benefits if any at all and isn't really any different than the current design that we have in place.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
kundadebdatta: Fixed.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
kundadebdatta: Refactored.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Added.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Updated an impl for `RetryResult` containing `is_retry()` matcher.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
kundadebdatta: Updated the code.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Thanks for the feedback. This is now addressed.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
```diff
@@ -0,0 +1,141 @@
...
+        Sender: Fn(&mut Request) -> Fut + Send + Sync,
+        Fut: std::future::Future<Output = azure_core::Result<RawResponse>> + ConditionalSend;
+}
+
+/// Concrete retry handler implementation with exponential back off.
+/// This handler provides automatic retry capabilities for Cosmos DB operations using
+/// a pluggable retry policy system. It wraps HTTP requests with intelligent retry logic
+/// that handles both transient network errors and HTTP error responses.
+#[derive(Debug, Clone)]
+pub struct BackOffRetryHandler;
```
analogrelay, line 68: _**non-blocking**_: In Rust, I think this would more often just be a few bare functions in a module, rather than grouping it into a type, since there's no state. In .NET/Java, types are the only place to put functions, so having a "zero-size" type just to hold some methods happens a lot more often than in Rust.

**However, there's really no need to churn this PR to change that. I just wanted to point it out for the future.**

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,407 @@
...
+            );
+            let retry_result = self.check_if_retry_needed(retry_after);
+
+            if retry_result.is_retry() {
+                self.current_attempt_count += 1;
+                return retry_result;
+            }
+        }
+
+        tracing::trace!("Exhausted all retry attempts and reached maximum retry: {:?}. The request will not be retried.", self.max_attempt_count);
```
analogrelay: nit for next time (or if you have to update the PR anyway): The `tracing::` log statements fully support structured logging (i.e. specifying key/value pairs associated with each log message). I highly recommend doing everything you can to use them instead of formatting values into the string message. For example, you could do this:


```suggestion
        tracing::trace!(max_attempt_count = self.max_attempt_count, "Exhausted all retry attempts and reached maximum retry. The request will not be retried.");
```

That allows users to extract the value more easily, depending on where they're sending their logs, and it still results in a reasonable text output.

This is also non-blocking though, don't churn the PR just for it :)

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,407 @@
...
+    ///
+    /// # Returns
+    /// A `RetryResult` indicating if the request can be retried.
+    ///
+    /// # Note
+    /// Currently uses a fixed 500ms base retry delay. Future versions may parse
+    /// the `x-ms-retry-after-ms` header from the response headers to respect
+    /// server-suggested retry delays.
+    fn should_retry_response(&mut self, response: &RawResponse) -> RetryResult {
+        if response.status() == StatusCode::TooManyRequests {
```
analogrelay, line 152: The default behavior is for non-200 HTTP responses to result in an `Err` from the Azure Core Pipeline, so this is probably unnecessary, but we might as well keep it around.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -0,0 +1,407 @@
...
+mod tests {
+    use crate::retry_policies::resource_throttle_retry_policy::ResourceThrottleRetryPolicy;
+    use crate::retry_policies::RetryPolicy;
+    use crate::retry_policies::RetryResult;
+    use azure_core::http::{RawResponse, StatusCode};
+    use azure_core::time::Duration;
+
+    /// Helper function to create a mock RawResponse with a given status code
+    fn create_mock_response(status: StatusCode) -> azure_core::Result<RawResponse> {
+        use typespec_client_core::http::headers::Headers;
```
analogrelay: nit: I usually try to avoid `use` statements inside the function unless absolutely necessary.

analogrelay (approved):
Some totally non-blocking idiomatic suggestions. Take a glance over them and resolve them, I'm good to go on this.

sdk/cosmos/azure_data_cosmos/src/lib.rs
```diff
@@ -28,3 +26,6 @@ pub use partition_key::*;
 pub use query::Query;
 
 pub use feed::{FeedPage, FeedPager};
+pub(crate) mod handler;
+pub mod retry_policies;
```
heaths: Why is your policy - something customers should never have to interact with - public?

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
analogrelay: We don't always omit the Core retry policy, though we do in some languages. We do still provide significant in-depth retry logic in all languages. Retries have to handle a lot of different scenarios, not just 429s

sdk/cosmos/azure_data_cosmos/src/lib.rs
analogrelay: This module doesn't need to be public. If we need to expose things from it in the future, we can make it public then.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
```diff
@@ -0,0 +1,141 @@
...
+// Helper trait to conditionally require Send on non-WASM targets
+#[cfg(not(target_arch = "wasm32"))]
+pub trait ConditionalSend: Send {}
+#[cfg(not(target_arch = "wasm32"))]
+impl<T: Send> ConditionalSend for T {}
+
+#[cfg(target_arch = "wasm32")]
+pub trait ConditionalSend {}
+#[cfg(target_arch = "wasm32")]
+impl<T> ConditionalSend for T {}
```
heaths, line 21: Nit: to promote using these in more places, I'd pub these in your `src/lib.rs`. We're looking at that for core, but there's a few places we can't use them and questioning whether we want to have the inconsistency. If they work well for Cosmos, it's perfectly find. But they also shouldn't be exported publicly. Keep everything private or `pub(crate)` unless it really needs to be public. We can always expose things in the future. We can't claw them back or that's a breaking change.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
```diff
@@ -0,0 +1,141 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+use crate::retry_policies::resource_throttle_retry_policy::ResourceThrottleRetryPolicy;
+use crate::retry_policies::{RetryPolicy, RetryResult};
+use async_trait::async_trait;
+use azure_core::http::{request::Request, RawResponse};
+use typespec_client_core::async_runtime::get_async_runtime;
```
heaths: Don't refer to "typespec" in crates. In fact, not even sure why you have a dependency on `typespec` or `typespec_client_core` in your crate. Everything is re-exported from `azure_core`, so only use types (same module path) from `azure_core`.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
```diff
@@ -0,0 +1,141 @@
...
+    ///
+    /// ```rust,ignore
+    /// use azure_data_cosmos::handler::retry_handler::BackOffRetryHandler;
+    ///
+    /// let retry_handler = BackOffRetryHandler::new();
+    /// // Use retry_handler.send() to make requests with automatic retry
+    /// ```
+    pub fn new() -> Self {
+        Self {}
+    }
```
heaths: Given @analogrelay's comment about grouping functions in a type, I'll go a step further and say: there's no reason to even instantiate (allocate space for) this type. Your method below doesn't even use `self`. If you're going to clean this up in a subsequent PR, fine. But please do. Bare functions in rust are idiomatic and quite common.

sdk/cosmos/azure_data_cosmos/src/lib.rs
```diff
@@ -28,3 +26,6 @@ pub use partition_key::*;
...
+pub mod routing;
```
heaths, line 31: And does this need to be exported publicly as well? As mentioned elsewhere, don't expose more than you need to. We can always open things up. We can't claw them back after we GA. Start narrow. Expand as needed.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -29,24 +30,29 @@ use crate::{
...
 }
 
 impl CosmosPipeline {
     pub fn new(
         endpoint: Url,
         auth_policy: AuthorizationPolicy,
-        client_options: ClientOptions,
+        mut client_options: ClientOptions,
     ) -> Self {
+        client_options.retry = RetryOptions::none();
```
heaths, line 42: You should probably capture what the user specified first in case any of that is useful in your retry policy. Don't completely ignore any user options they set; barring that, you might even err if they set anything so they know why their options are getting ignored.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
heaths: @kundadebdatta as you resolve comments in code, please also resolve them on PRs so reviewers know whether there are outstanding changes (commits) expected or follow-up questions.

sdk/cosmos/azure_data_cosmos/src/lib.rs
kundadebdatta: Updated and made it non-public.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
kundadebdatta: Thanks for the suggestion @heaths. I have refactored the code.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
kundadebdatta: Thanks for the suggestion @heaths. I have refactored the code.

sdk/cosmos/azure_data_cosmos/src/lib.rs
kundadebdatta: Updated and made it non-public.

kundadebdatta: Updated and made it non-public.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Thanks for the suggestion @analogrelay. I have refactored the code.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Thanks for the suggestion @analogrelay. I have refactored the code.

analogrelay (approved):

PR #3243, opened 2025-10-20
Enable configuring retryable status codes

chlowell:
I guess this feature won't often be used, however I know a couple uses of it with the Go SDK and `azure_identity` needs either this or a custom retry policy to implement correct behavior for IMDS.

### DIFF
--- a/sdk/core/typespec_client_core/src/http/options/mod.rs
+++ b/sdk/core/typespec_client_core/src/http/options/mod.rs
@@ -71,7 +71,7 @@ pub struct PipelineOptions {
     /// field doesn't apply to custom retry policies.
     ///
     /// When empty, the default retry status codes are used as
-    /// described by [`crate::http::policies::RetryPolicy::get_retry_status_codes`].
+    /// described by [`crate::http::policies::RetryPolicy::retry_status_codes`].
     pub retry_status_codes: Vec<StatusCode>,
 }
 


sdk/core/typespec_client_core/src/http/options/retry.rs
```diff
@@ -56,37 +60,73 @@ impl Default for RetryMode {
...
+    /// will be used as described in the documentation for [`RetryOptions::with_retry_statuses`].
+    status_codes: Option<Vec<StatusCode>>,
 }
 
 impl RetryOptions {
     /// A retry strategy where attempts happen at intervals that get exponentially longer with each retry.
     pub fn exponential(options: ExponentialRetryOptions) -> Self {
         Self {
             mode: RetryMode::Exponential(options),
+            status_codes: None,
```
LarryOsterman: `..Default::default()` to future proof this in case there are other `RetryOptions` added in the future?

sdk/core/typespec_client_core/src/http/options/retry.rs
```diff
@@ -56,37 +60,73 @@ impl Default for RetryMode {
...
+    ///         StatusCode::ServiceUnavailable,
+    ///         StatusCode::TooManyRequests,
+    ///     ]);
+    /// ```
+    pub fn with_retry_statuses(mut self, statuses: Vec<StatusCode>) -> Self {
+        self.status_codes = Some(statuses);
+        self
+    }
+
     pub(crate) fn to_policy(&self, retry_headers: RetryHeaders) -> Arc<dyn Policy> {
```
LarryOsterman: I know it's an existing call, but should `to_policy` consume self rather than borrowing self? Possibly after renaming it `into_policy` to avoid a clippy warning?

The `.clone()` below just annoys me :).

LarryOsterman (approved):

sdk/core/typespec_client_core/src/http/options/retry.rs
```diff
@@ -56,37 +60,73 @@ impl Default for RetryMode {
 pub struct RetryOptions {
     /// The algorithm to use for calculating retry delays.
     mode: RetryMode,
+
+    /// The status codes that should trigger retries. If `None`, the default set of status codes
+    /// will be used as described in the documentation for [`RetryOptions::with_retry_statuses`].
+    status_codes: Option<Vec<StatusCode>>,
```
heaths: While leaf nodes of nested options should generally be `Option<T>`, that's not true of `Vec` and `HashMap` which are ZSTs. It just makes them harder. We only do this for models because we need to know the difference of empty (`Vec::new().len() == 0`) and absent.

Also, are we sure we want this user-settable? If client-settable, it should go in `PipelineSendOptions` like @LarryOsterman's recent change for `CheckSuccessOptions` that contains a `success_codes: &'static [u16]`, which is more general purpose for client-defined values.

You said that Identity needs this, so it seems more like client-define options, not user-defined options.

sdk/core/typespec_client_core/src/http/options/retry.rs
heaths: You can't if not non-exhaustive. And if in the same crate as the definition, it won't matter anyway because it forces you to always handle all cases. That only applies to callers.

sdk/core/typespec_client_core/src/http/options/retry.rs
chlowell: > are we sure we want this user-settable?

No. Keeping this out of the top-level public API makes sense to me because I expect SDK developers to know which codes from a given service are retriable. Clients should be able to handle this internally and application developers shouldn't need to think about it, and we have `RetryMode::Custom` for an escape hatch just in case.

> If client-settable, it should go in `PipelineSendOptions`

I suppose I can live with this but it seems odd to have clients specify retriable status codes with every request. That design seems to imply wanting to retry different codes for different operations, and I don't know a real scenario like that.

heaths: There are also `PipelineOptions` you pass when setting up the `Pipeline`.

sdk/core/typespec_client_core/src/http/policies/retry/exponential.rs
```diff
@@ -18,6 +21,7 @@ pub(crate) struct ExponentialRetryPolicy {
     max_elapsed: Duration,
     max_delay: Duration,
     retry_headers: RetryHeaders,
+    status_codes: Vec<StatusCode>,
```
heaths: Nit: I'd call these `retry_status_codes` like you do in the `PipelineOptions`. Helps more easily identity them. Plus, makes it sound more like input. I was also thinking that for the parameter names, but they do get a bit long that way. Something to think about.

sdk/core/typespec_client_core/src/http/policies/retry/exponential.rs
```diff
@@ -47,6 +53,10 @@ impl RetryPolicy for ExponentialRetryPolicy {
         Some(&self.retry_headers)
     }
 
+    fn get_retry_status_codes(&self) -> &[StatusCode] {
```
heaths: We don't do `get_` for getters in Rust. Just `retry_status_codes()`. Please also use F2 to change the `get_retry_headers` above. Might've been grandfathered in accidentally. `is_` is fine and idiomatic.

sdk/core/typespec_client_core/src/http/policies/retry/mod.rs
```diff
@@ -168,7 +178,13 @@ where
             let (last_result, retry_after) = match result {
                 Ok(response) => {
                     let status = response.status();
-                    if !RETRY_STATUSES.contains(&status) {
+                    let retry_status_codes = self.get_retry_status_codes();
+                    let retry_status_codes = if retry_status_codes.is_empty() {
+                        DEFAULT_RETRY_STATUSES
+                    } else {
+                        retry_status_codes
+                    };
```
heaths: Tip: more concise:


```suggestion
                    let mut retry_status_codes = self.get_retry_status_codes();
                    let retry_status_codes.is_empty() {
                        retry_status_codes = DEFAULT_RETRY_STATUSES; // might need to use .extend if types differ
                    }
```

sdk/core/typespec_client_core/src/http/policies/retry/mod.rs
```diff
@@ -127,10 +137,10 @@ pub trait RetryPolicy: std::fmt::Debug + Send + Sync {
     }
 }
 
-/// The status codes where a retry should be attempted.
+/// Default status codes where a retry should be attempted.
 ///
 /// On all other 4xx and 5xx status codes no retry is attempted.
-const RETRY_STATUSES: &[StatusCode] = &[
+const DEFAULT_RETRY_STATUSES: &[StatusCode] = &[
```
heaths:
```suggestion
const DEFAULT_RETRY_STATUS_CODES: &[StatusCode] = &[
```

Reads better, and better aligns with `*status_codes` everywhere else.

LarryOsterman (approved):

PR #3245, opened 2025-10-21
Yield in blob container client; read environment variables from env file

LarryOsterman:


sdk/storage/azure_storage_blob/perf/list_blob_test.rs
```diff
@@ -83,7 +74,15 @@ impl PerfTest for ListBlobTest {
         let recording = context.recording();
         let credential = recording.credential();
         let container_name = format!("perf-container-{}", uuid::Uuid::new_v4());
-        let client = BlobContainerClient::new(&self.endpoint, container_name, credential, None)?;
+        let endpoint = match &self.endpoint {
+            Some(e) => e.clone(),
+            None => format!(
+                "https://{}.blob.core.windows.net",
+                recording.var("AZURE_STORAGE_ACCOUNT_NAME", None)
```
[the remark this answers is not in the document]
LarryOsterman: recording.var doesn't return a result, so `expect` will fail to compile.

sdk/storage/azure_storage_blob/src/generated/clients/blob_container_client.rs
```diff
@@ -904,9 +904,14 @@ impl BlobContainerClient {
                             }),
                         )
                         .await?;
-                    let (status, headers, body) = rsp.deconstruct();
-                    let res: ListBlobsFlatSegmentResponse = xml::from_xml(&body)?;
-                    let rsp = RawResponse::from_bytes(status, headers, body).into();
+                    // Because serialization can be a CPU intensive operation, offload it to a separate task.
```
heaths: This works, but if we were do refactor the emitter, it should be enough just to put a yield before this if we wanted to expose something like an `AsyncRuntime::yield()` before we GA. That should have the same effect, right?

sdk/core/typespec_client_core/src/async_runtime/mod.rs
```diff
@@ -115,6 +115,9 @@ pub trait AsyncRuntime: Send + Sync {
     /// # Returns
     /// A future that resolves after the specified duration has elapsed.
     fn sleep(&self, duration: Duration) -> TaskFuture;
+
+    /// Yield the current task back to the runtime scheduler.
+    fn yield_now(&self) -> TaskFuture;
```
heaths, line 120: IMO, I'd prefer just `yield` since we're abstracting it from `tokio`. I mean, when else are we going to yield especially since it doesn't even take an interval or anything?

LarryOsterman: So would I, but since `yield` is a [reserved keyword in Rust (apparently)](https://doc.rust-lang.org/reference/keywords.html?highlight=yield#r-lex.keywords.reserved.list), it needs to be named `r#yield` to be a valid identifier.

So I'd rather just go with the [`tokio`](https://docs.rs/tokio/latest/tokio/task/fn.yield_now.html) name and use `yield_now`.

sdk/storage/azure_storage_blob/src/generated/clients/blob_container_client.rs
LarryOsterman: This code was written before I'd implemented `yield_now`. The current PR has the new `yield_now` implementation.

heaths (approved):
That's unfortunately we can't simulate a yield for wasm32. Seems it should work for "newer" ECMAScript.

LarryOsterman:
> That's unfortunately we can't simulate a yield for wasm32. Seems it should work for "newer" ECMAScript.

FWIW, I was unable to find a way of doing it absent bringing in some arguably sketchy crates.

PR #3246, opened 2025-10-21
Adds new AAD scope value instead of account scope

aavasthy:
As part of this PR we are adding a new AAD audience scope value "https://cosmos.azure.com/.default". We will be using this instead of account scope value in rust sdk. This new scope value will be eventually used across different clouds and regions. It is already accepted in public cloud across stage, canary and prod.


sdk/cosmos/azure_data_cosmos/src/pipeline/authorization_policy.rs
```diff
@@ -110,15 +111,14 @@ impl Policy for AuthorizationPolicy {
 /// NOTE: Resource tokens are not yet supported.
 async fn generate_authorization(
     auth_token: &Credential,
-    url: &Url,
-
+    #[allow(unused_variables)] url: &Url,
```
analogrelay: We can get rid of this parameter altogether. This isn't a public API surface and we no longer need the account URL to generate an authorization token.

sdk/cosmos/azure_data_cosmos/src/pipeline/authorization_policy.rs
```diff
@@ -278,8 +268,7 @@ mod tests {
     }
 
     #[test]
-    fn scope_from_url_extracts_correct_scope() {
-        let scope = scope_from_url(&Url::parse("https://example.com/dbs/test_db/colls").unwrap());
-        assert_eq!(scope, "https://example.com/.default");
+    fn uses_constant_scope_value() {
```
analogrelay: This test can just be removed. It's only testing the value of the constant which feels unnecessary.

sdk/cosmos/azure_data_cosmos/tests/aad_scope.rs
```diff
@@ -0,0 +1,76 @@
...
+            "mock_token_for_scope_verification".to_string(),
+            OffsetDateTime::now_utc().saturating_add(Duration::minutes(60)),
+        ))
+    }
+}
+
+/// Tests that the authorization policy uses the constant scope "https://cosmos.azure.com/.default"
+/// when requesting AAD tokens.
+#[tokio::test]
+async fn aad_authentication_uses_constant_scope() -> Result<(), Box<dyn std::error::Error>> {
```
analogrelay: This isn't calling the authorization policy as far as I can tell. The only code its testing is this test code. The authorization policy is private, so we won't be able to call it from this external integration test (tests under `tests/` are built as separate binaries). Ideally, this code should be testing against the emulator and using `#[recorded::test]`, like the other tests in this directory. If that's not feasible, I'm OK with just removing this test. We have a unit test that gives us coverage on AAD auth, and it'd be no worse than what we had before.

analogrelay (changes requested):
Mostly good, just some clean-ups and thoughts on the integration test.

sdk/cosmos/azure_data_cosmos/tests/aad_scope.rs
aavasthy: The emulator does not support AAD authentication so end to end tests are displaying stimulated behavior than actual. I have removed the aad_scope class and added a better unit test in authorization_policy.

sdk/cosmos/azure_data_cosmos/src/pipeline/authorization_policy.rs
aavasthy: removed this test

sdk/cosmos/azure_data_cosmos/src/pipeline/authorization_policy.rs
aavasthy: removed this unused url

sdk/cosmos/azure_data_cosmos/src/pipeline/authorization_policy.rs
```diff
@@ -110,15 +112,15 @@ impl Policy for AuthorizationPolicy {
 /// NOTE: Resource tokens are not yet supported.
 async fn generate_authorization(
     auth_token: &Credential,
-    url: &Url,
+    _url: &Url,
```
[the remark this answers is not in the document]
aavasthy: removed

PR #3248, opened 2025-10-21
Use custom error for AMQP and EventHubs

LarryOsterman:
Define a new custom error type for the `azure_core_amqp` (`AmqpError`) and `azure_messaging_eventhubs` (`EventHubsError`) packages.

Ensured that `azure_core::error::Error` types can be embedded in AMQP and Event Hubs errors, and that a `try_from` implementation exists to convert back to Azure Core errors (the `try_from` fails if the AMQP/EventHub error is not derived from an Azure Core error).

Also added `AmqpError::with_message` and `EventHubsError::with_message` to cover the cases where a simple error message needs to be returned to the caller. 

Also refactored the AMQP "error" module into 3 source files:`mod.rs` which implements the module, `tests.rs` which implements test cases, and `error_condition.rs` which implements the `AmqpErrorCondition` extensible enumeration - this was split out to reduce the size of the `mod.rs` file.

Note: This turned out to be somewhat more complicated than I thought it would be originally.

The core problem is that instead of having a single error type for everything, we have 3 different error types, each of which is used in different circumstances. That made figuring out which error kind to return from which function rather challenging.

Specifically in the EventHubs code which implements AMQP functionality (including the logic which creates AMQP structures), the functions need to return an *AMQP* error code even though they are in the EventHubs package, and are interacting with EventHubs structures - that's because ultimately the code which consumes these structures assumes that it will receive an AMQP error, NOT an EventHubs error.

Bonus: Cleaned up a bunch of unused EventHubs errors.

### DIFF
--- a/sdk/core/azure_core_amqp/src/error/tests.rs
+++ b/sdk/core/azure_core_amqp/src/error/tests.rs
@@ -418,20 +418,20 @@ fn test_amqp_error_display_send_rejected() {
     let error = AmqpError::from(AmqpErrorKind::SendRejected);
     assert_eq!(
         format!("{}", error),
-        "AMQP Send Rejected with no error information"
+        "Send Rejected with no error information"
     );
 }
 
 #[test]
 fn test_amqp_error_display_non_terminal_delivery_state() {
     let error = AmqpError::from(AmqpErrorKind::NonTerminalDeliveryState);
-    assert_eq!(format!("{}", error), "AMQP Non Terminal Delivery State");
+    assert_eq!(format!("{}", error), "Non Terminal Delivery State");
 }
 
 #[test]
 fn test_amqp_error_display_illegal_delivery_state() {
     let error = AmqpError::from(AmqpErrorKind::IllegalDeliveryState);
-    assert_eq!(format!("{}", error), "AMQP Illegal Delivery State");
+    assert_eq!(format!("{}", error), "Illegal Delivery State");
 }
 
 #[test]
@@ -461,7 +461,7 @@ fn test_amqp_error_source() {
 fn test_amqp_error_debug() {
     let error = AmqpError::with_message(Cow::Borrowed("Debug test"));
     let debug_output = format!("{:?}", error);
-    assert!(debug_output.contains("AmqpError"));
+    assert!(debug_output.contains("AMQP Error:"));
     assert!(debug_output.contains("Debug test"));
 }
 


sdk/eventhubs/azure_messaging_eventhubs/src/event_processor/partition_client.rs
```diff
@@ -163,7 +159,16 @@ impl PartitionClient {
...
         };
-        self.checkpoint_store.update_checkpoint(checkpoint).await
+        self.checkpoint_store
+            .update_checkpoint(checkpoint)
+            .await
+            .map_err(|e| {
+                e.with_context(format!(
+                    "Failed to update checkpoint for partition {}",
+                    self.partition_id
+                ))
```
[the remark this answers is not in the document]
LarryOsterman: This function uses the `azure_core::Error::with_context` method, *not* the `ResultExt` extension trait.

sdk/core/azure_core_amqp/src/messaging.rs
```diff
@@ -1279,7 +1280,10 @@ impl AmqpMessage {
     }
 
     /// Serialize the AMQP message to a vector of bytes.
-    #[allow(unused_variables)]
+    #[cfg_attr(
+        all(feature = "fe2o3_amqp", target_arch = "wasm32"),
+        allow(unused_variables)
+    )]
```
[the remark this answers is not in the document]
LarryOsterman: Good catch.

sdk/core/azure_core_amqp/src/error/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+            AmqpErrorKind::TransportImplementationError(s) => {
+                write!(f, "Transport Implementation Error: {}", s)
+            }
+            AmqpErrorKind::ConnectionDropped(s) => {
+                write!(f, "Connection dropped: {}", s)
+            }
+            AmqpErrorKind::FramingError(s) => {
+                write!(f, "Connection Framing error: {}", s)
+            }
+            AmqpErrorKind::IdleTimeoutElapsed(s) => {
```
antkmsft, line 257: For some errors, we start the write with "AMQP ...", for some we don't. Which is fine, I am not directly proposing either remove or add. Just bringing it to your attention, in case you've been looking at this code for so long you don't see it.

sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
```diff
@@ -62,7 +53,7 @@ impl AmqpClaimsBasedSecurityApis for Fe2o3ClaimsBasedSecurity {
             .client_node_addr("rust_amqp_cbs")
             .attach(session.borrow_mut())
             .await
-            .map_err(|e| azure_core::Error::from(AmqpError::from(e)))?;
+            .map_err(AmqpError::from)?;
```
antkmsft, line 56: So, we no longer need to remap, are lines like this still needed at all?

sdk/eventhubs/azure_messaging_eventhubs/src/common/authorizer.rs
```diff
@@ -114,7 +115,9 @@ impl Authorizer {
             // insert returns some if it *fails* to insert, None if it succeeded.
             let present = scopes.insert(path.clone(), token);
             if present.is_some() {
-                return Err(EventHubsError::from(ErrorKind::UnableToAddAuthenticationToken).into());
+                return Err(AmqpError::with_message(
```
antkmsft, line 118: You obviously thought about it way more than I do, but why the error in this code an AmqpError and not EventHubs error?

sdk/eventhubs/azure_messaging_eventhubs/src/common/authorizer.rs
```diff
@@ -149,15 +152,12 @@ impl Authorizer {
...
+    ) -> azure_core_amqp::Result<()> {
         // Test Hook: Disable interacting with Event Hubs service if the test doesn't want it.
         #[cfg(test)]
         {
             let disable_authorization = self.disable_authorization.lock().map_err(|e| {
-                azure_core::Error::with_message(
-                    azure_core::error::ErrorKind::Other,
-                    format!("Unable to grab disable mutex: {}", e),
-                )
+                AmqpError::with_message(format!("Unable to grab disable mutex: {}", e))
```
antkmsft, line 160: Same here, unable to grab mutex is an error in AMQP stack, not EventHubs?

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable/claims_based_security.rs
```diff
@@ -60,31 +59,9 @@ impl RecoverableClaimsBasedSecurity {
...
-                    ErrorRecoveryAction::ReturnError
-                }
-            }
-            _ => {
-                debug!(err=?e, "Non AMQP error: {e}");
-                ErrorRecoveryAction::ReturnError
-            }
-        }
+    fn should_retry_claims_based_security_response(e: &AmqpError) -> ErrorRecoveryAction {
+        warn!("Amqp operation failed: {:?}", e);
```
antkmsft, line 63: I see we capitalize AMQP in error messages most of the time
```suggestion
        warn!("AMQP operation failed: {:?}", e);
```

sdk/eventhubs/azure_messaging_eventhubs/src/error.rs
```diff
@@ -61,21 +57,11 @@ impl std::error::Error for EventHubsError {
...
+            ErrorKind::AzureCore(e) => write!(f, "Azure Core Error: {}", e),
             ErrorKind::SendRejected(e) => write!(f, "Send rejected: {:?}", e),
             ErrorKind::InvalidManagementResponse => f.write_str("Invalid management response"),
-            ErrorKind::UnableToAddAuthenticationToken => {
-                f.write_str("Unable to add authentication token")
-            }
-            ErrorKind::MissingSession => {
-                f.write_str("The session for the specified partition is missing.")
-            }
             ErrorKind::AmqpError(source) => write!(f, "AmqpError: {:?}", source),
```
antkmsft: nit, we say "Azure Core Error" above and not "AzureCore", so for consistency I suggest
```suggestion
            ErrorKind::AmqpError(source) => write!(f, "AMQP Error: {:?}", source),
```

antkmsft (approved):
I understand that throwing AMQP errors from EH makes some things to work nicely, and I do not know all the context so I am leaving up to you, but part of me instinctively thinks that throwing AMQP errors from EH looks a bit like an elegant hack, and maybe the correct solution is some 3rd option. But, patterns like this are not always the case of hacks, and I am leaving this up to you.

sdk/core/azure_core_amqp/src/error/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+
+impl TryFrom<AmqpError> for azure_core::Error {
+    type Error = AmqpError;
+    fn try_from(value: AmqpError) -> std::result::Result<Self, Self::Error> {
+        match value.kind {
+            AmqpErrorKind::AzureCore(e) => Ok(e),
+            _ => Err(AmqpError::with_message("Amqp error is not an Azure Error")),
+        }
+    }
+}
```
heaths: This is fragile for callers. Why not just wrap it in an `azure_core::Error` with `Kind::Other`?

```rust
azure_core::Error::with_error(ErrorKind::Other, err, "AMQP error occured");
```

Don't even necessarily need the message. Could just wrap the error and use its message. This would be better in cases where a method returns an `azure_core::Error` but uses AMQP (or Event Hubs) and just wants to return an error. Just wrap it, just like Go would with `fmt.Errorf("amqp failed: %w", err)`.

heaths: If it was an `AzureCore` kind, which I see you're wrapping, yeah, you can unwrap it like this. But I'd avoid `TryFrom` in this case and just pass the original error as context with `ErrorKind::Other` instead of fail. Callers are already in a failure case. Let's not hide the original error.

sdk/core/azure_core_amqp/src/error/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+            | AmqpErrorKind::SessionClosedByRemote(e)
+            | AmqpErrorKind::SessionDetachedByRemote(e)
+            | AmqpErrorKind::ConnectionClosedByRemote(e)
+            | AmqpErrorKind::ConnectionDetachedByRemote(e)
+            | AmqpErrorKind::LinkStateError(e)
+            | AmqpErrorKind::ConnectionDropped(e)
+            | AmqpErrorKind::TransferLimitExceeded(e)
+            | AmqpErrorKind::FramingError(e)
+            | AmqpErrorKind::IdleTimeoutElapsed(e) => Some(e.as_ref()),
+            AmqpErrorKind::ManagementStatusCode(_, _)
```
heaths, line 196: Nit: at first glance I thought there was a typo. I'd put a new line before here just to visually separate them. Maybe after the first match arm as well.

sdk/eventhubs/azure_messaging_eventhubs/src/lib.rs
```diff
@@ -44,7 +44,7 @@ pub mod builders {
 
 pub use common::retry::RetryOptions;
 
-pub use crate::error::{ErrorKind, EventHubsError};
+pub use error::*;
```
heaths: FWIW, core only re-exports `Result` and `Error`, but not `ErrorKind` or extension traits.

sdk/core/azure_core_amqp/src/error/mod.rs
LarryOsterman: `AMQP ...` seems incorrect to me - We already know that this is an AMQP error - it's coming from an `AmqpError` after all. And if the `AmqpError` is surfaced from (say) Event Hubs, then the message from Event Hubs will be something like `
'"AMQP Error: <AmqpError text>"`.

sdk/core/azure_core_amqp/src/fe2o3/cbs.rs
LarryOsterman: The error is a Fe2o3 error, we need to turn it into an AMQP error. However we no longer need to turn it into an Azure Core Error.

sdk/eventhubs/azure_messaging_eventhubs/src/common/authorizer.rs
LarryOsterman: That's the "gotcha" bit in the PR description - the authorizer logic is wrapping the AMQP CBS APIs inside of EventHubs. And it's called by the retry logic, which understands AMQP errors, *not* EventHubs errors (because it makes retry decisions based on the AMQP errors being returned). In this specific case, we're returning a message, so it *could* be an EventHubsError, but I didn't want to return an AmqpError which wraps an EventHubs error, which is a message.

sdk/eventhubs/azure_messaging_eventhubs/src/common/authorizer.rs
LarryOsterman: It's not but as I mentioned, this logic is actually AMQP logic, not EventHubs logic (it's EventHubs code which is interacting with AMQP logic at a level that assumes everything is AMQP).

sdk/core/azure_core_amqp/src/error/mod.rs
LarryOsterman: I went around and around on this one. I can absolutely see wrapping the AMQP error in the non `AzureCore` case.

LarryOsterman: Given @antkmsft's feedback, I'll change this to `from`.

sdk/eventhubs/azure_messaging_eventhubs/src/common/recoverable/claims_based_security.rs
LarryOsterman: Fixed, good catch.

PR #3254, opened 2025-10-22
[Storage] Improve `list_blobs()` API name struct field accessor

vincenttran-msft:
Making this just to illustrate that despite consuming:
```typespec
/** The name of the blob. */
  @Xml.name("Name") name: string;
```

This field is coming back optional? 

Relevant snippet from [branch](https://github.com/Azure/azure-rest-api-specs/blob/1227e9ce214ee17f72946fe749bce75666237367/specification/storage/Microsoft.BlobStorage/models.tsp#L872): 
```typespec
/** An Azure Storage Blob */
@Xml.name("Blob")
model BlobItemInternal {
  /** The name of the blob. */
  @Xml.name("Name") name: string;
```

### DIFF
--- a/sdk/storage/azure_storage_blob/src/generated/models/pub_models.rs
+++ b/sdk/storage/azure_storage_blob/src/generated/models/pub_models.rs
@@ -14,7 +14,7 @@ use super::{
     QueryRequestType, QueryType, RehydratePriority,
 };
 use azure_core::{
-    base64::option::{deserialize, serialize},
+    base64::{deserialize, serialize},
     fmt::SafeDebug,
     time::OffsetDateTime,
 };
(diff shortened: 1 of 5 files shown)


sdk/storage/azure_storage_blob/src/generated/models/pub_models.rs
```diff
@@ -256,7 +256,7 @@ pub struct BlobItemInternal {
 
     /// The name of the blob.
     #[serde(rename = "Name", skip_serializing_if = "Option::is_none")]
-    pub name: Option<BlobName>,
```
vincenttran-msft, line 259: At least this was able to get us out of one unnecessary layer (`BlobName` model is not necessary here), but this is still coming back as `Option<String>` when I expected pure `String`.

In TypeSpec we defined it as `name: string` and not `name?: string` in the `.tsp`. It doesn't really make sense in this context that a blob would have no name.

vincenttran-msft:
Closing as this is expected behavior according to the higher-level TypeSpec rules defined in our emitter package. But still an issue we need to close on a long-term solution for.

PR #3256, opened 2025-10-23
Rename BearerTokenCredentialPolicy -> BearerTokenAuthorizationPolicy

chlowell:
Two reasons "authorization" is better than "credential" here:
1. the policy authorizes requests with bearer tokens by setting the "Authorization" header
2. a `TokenCredential` could provide other types of token in addition to or instead of bearer tokens

### DIFF
--- a/sdk/storage/azure_storage_blob/src/clients/page_blob_client.rs
+++ b/sdk/storage/azure_storage_blob/src/clients/page_blob_client.rs
@@ -18,7 +18,7 @@ use crate::{
 use azure_core::{
     credentials::TokenCredential,
     http::{
-        policies::{BearerTokenCredentialPolicy, Policy},
+        policies::{BearerTokenAuthorizationPolicy, Policy},
         NoFormat, Pipeline, RequestContent, Response, Url, XmlFormat,
     },
     tracing, Bytes, Result,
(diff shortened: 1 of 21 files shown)


sdk/storage/azure_storage_blob/src/clients/page_blob_client.rs
```diff
@@ -18,7 +18,7 @@ use crate::{
 use azure_core::{
     credentials::TokenCredential,
     http::{
-        policies::{BearerTokenCredentialPolicy, Policy},
+        policies::{BearerTokenAuthorizationPolicy, Policy},
```
LarryOsterman, line 21: It's weird that this rename didn't involve any other code changes in this file. I wonder why.

heaths: If he used Copilot, I've noticed on several occasions it doesn't find everything. I recommend in cases like this - to rely on what the LSP knows contextually as well as for speed (blink of an eye), use F2 to rename a symbol. That said, I've noticed it doesn't always fix doc code examples but `cargo test` should catch that.

chlowell: F2 was my weapon of choice here. This is actually an unused import that somehow doesn't raise a warning. There appear to be many such imports in azure_storage_blob 🤔

LarryOsterman: I figured F2 was involved. I'm just completely surprised that this didn't trigger an unused import clippy warning. 

That's why it was weird. Not blocking, just weird.

antkmsft (approved):

heaths (approved):

PR #3257, opened 2025-10-23
Converted perf tests to benches so they don't conflict with cargo tests

LarryOsterman:
The command line for cargo test is passed to all tests if you're not specific, but clap based tests (like perf tests) get upset when they encounter command line parameters that are unknown.

To avoid this, instead of using `[[test]]` to declare perf tests, use `[[bench]]`.

### DIFF
--- a/sdk/core/azure_core_test/src/perf/mod.rs
+++ b/sdk/core/azure_core_test/src/perf/mod.rs
@@ -509,6 +509,8 @@ impl PerfRunner {
                     .value_parser(clap::value_parser!(i64))
                     .global(true),
             )
+            // Cargo bench passes --bench to the test binary to instruct it to run benchmarks only.
+            .arg(clap::arg!(--bench).required(false).global(true))
             .arg(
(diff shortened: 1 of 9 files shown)


antkmsft (approved):

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -509,6 +509,8 @@ impl PerfRunner {
                     .value_parser(clap::value_parser!(i64))
                     .global(true),
             )
+            // Cargo bench passes --bench to the test binary to instruct it to run benchmarks only.
+            .arg(clap::arg!(--bench).required(false).global(true))
```
heaths, line 513: TIL

LarryOsterman: I learned it yesterday, and spent about 10 minutes trying to understand where this mysterious `--bench` command line argument was coming from.

FWIW, it's at the very top of the documentation for `cargo bench` :).

PR #3263, opened 2025-10-25
Define create_certificate as real LRO

heaths:
Now with LRO support in core, "hide" the generated create_certificate and define a convenience client method instead. Those migrating Key Vault TypeSpec tried making it a TypeSpec LRO, but had significant problems so that was scrapped in favor of using convenience methods as we have in the past.

### DIFF
--- a/sdk/keyvault/azure_security_keyvault_certificates/tsp-location.yaml
+++ b/sdk/keyvault/azure_security_keyvault_certificates/tsp-location.yaml
@@ -1,5 +1,5 @@
 directory: specification/keyvault/Security.KeyVault.Certificates
 commit: 2b40f1e7eab4de4f94e8bce032f91b25b3656f86
-repo: heaths/azure-rest-api-specs
+repo: Azure/azure-rest-api-specs
 additionalDirectories:
 - specification/keyvault/Security.KeyVault.Common


sdk/keyvault/azure_security_keyvault_certificates/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ### Breaking Changes
 
+- `CertificateClient::create_certificate()` now returns a `Poller<CertificateOperation>`.
+- Removed `CertificateClient::begin_create_certificate()`.
+- Removed `CertificateClient::resume_create_certificate()`.
```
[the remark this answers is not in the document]
heaths: This is old now.

sdk/keyvault/azure_security_keyvault_certificates/tsp-location.yaml
```diff
@@ -1,5 +1,5 @@
 directory: specification/keyvault/Security.KeyVault.Certificates
-commit: 39b2bb6b46f241309c56ad31a88aaee910578036
-repo: Azure/azure-rest-api-specs
+commit: 2b40f1e7eab4de4f94e8bce032f91b25b3656f86
+repo: heaths/azure-rest-api-specs
```
jhendrixMSFT: Azure?

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
```diff
@@ -66,17 +65,18 @@ impl CertificateClient {
...
-    ///     .begin_create_certificate("certificate-name", body.try_into()?, None)?
+    ///     .create_certificate("certificate-name", body.try_into()?, None)?
     ///     .await?
     ///     .into_body()?;
     ///
     /// # Ok(()) }
     /// ```
-    pub fn begin_create_certificate(
+    #[tracing::function("KeyVault.createCertificate")]
+    pub fn create_certificate(
```
jhendrixMSFT, line 75: Why was the `begin_` prefix removed?

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
```diff
@@ -66,17 +65,18 @@ impl CertificateClient {
...
     /// let certificate = client
-    ///     .begin_create_certificate("certificate-name", body.try_into()?, None)?
+    ///     .create_certificate("certificate-name", body.try_into()?, None)?
     ///     .await?
     ///     .into_body()?;
     ///
     /// # Ok(()) }
     /// ```
-    pub fn begin_create_certificate(
+    #[tracing::function("KeyVault.createCertificate")]
```
LarryOsterman, line 74: The guidelines say that span name [should be consistent across all languages](https://github.com/Azure/azure-sdk/blob/d019037c2891efdb367a627d6e786f857735626e/docs/observability/opentelemetry-conventions.md#public-api-calls)

So this should be `"Microsoft.Security.KeyVault.CertificateClient.createCertificate"`. Unfortunately tcgc doesn't have a way of expressing the resource manager namespace for a service, so the generate code gets this wrong.

sdk/keyvault/azure_security_keyvault_certificates/tsp-location.yaml
heaths: Ah, right. Thanks for catching that.

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
LarryOsterman, line 130: It might be interesting to have a test case which verifies that the traces generated by this API match the ones expected (there should be a single `SpanKind::Internal` span with <n> child spans of the type `SpanKind::Client`. 

I believe that the `assert_instrument_information` test in `azure_core_test::tracing` should work to verify the calls.

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
heaths: It was always temporary. Like other languages, we just use the method name. I only had it for the convenience method originally so we had the generated `create_certificate` exported. No longer. I "override" it with the convenience method.

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
heaths: I just copied what was generated. I'd rather be consistent with the generated content. If that ever gets "fixed" I'll have to fix this up.

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
heaths: Why would there be multiple `SpanKind::Client` spans? I don't call the generated client methods because that means a lot more copying.

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
jhendrixMSFT: I see. The guidelines still state `begin_` which is why I was confused. That will be removed?

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
LarryOsterman: There are multiple HTTP operations - each HTTP operation generates a `SpanKind::Client` span (there's also a `SpanKind::Client` span created for each retry but that doesn't matter here)

LarryOsterman (approved):

PR #3272, opened 2025-10-29
Validate KeyVaultCertificates LRO implementation

LarryOsterman:
Also: Updated telemetry test infrastructure to validate parent/child relationships
Also: Allow span names to be generated as well as static.

Bonus: Removed a number of explicit dependencies on the typespec_client_core to ensure that all required types are re-exported from azure_core.

### DIFF
--- a/sdk/core/azure_core_test/src/tracing.rs
+++ b/sdk/core/azure_core_test/src/tracing.rs
@@ -8,7 +8,8 @@
 use azure_core::{
     http::{headers::HeaderName, Context, Request},
     tracing::{
-        AsAny, Attribute, AttributeValue, Span, SpanKind, SpanStatus, Tracer, TracerProvider,
+        AsAny, Attribute, AttributeValue, Span, SpanGuard, SpanKind, SpanStatus, Tracer,
+        TracerProvider,
     },
     Uuid,
 };
@@ -179,7 +180,7 @@ impl Span for MockSpan {
         todo!()
     }
 
-    fn set_current(&self, _context: &Context) -> Box<dyn typespec_client_core::tracing::SpanGuard> {
+    fn set_current(&self, _context: &Context) -> Box<dyn SpanGuard> {
         todo!()
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -250,7 +250,6 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
- "typespec_client_core",
 ]
 
 [[package]]
@@ -268,7 +267,6 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
- "typespec_client_core",
  "url",
 ]
 
@@ -294,7 +292,6 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
- "typespec_client_core",
  "url",
  "uuid",
  "zip",
@@ -455,7 +452,6 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio",
- "typespec_client_core",
 ]
 
 [[package]]
@@ -476,7 +472,6 @@ dependencies = [
  "serde_json",
  "sha2",
  "tokio",
- "typespec_client_core",
 ]
 
 [[package]]
@@ -495,7 +490,6 @@ dependencies = [
  "serde_json",
  "time",
  "tokio",
- "typespec_client_core",
 ]
 
 [[package]]
--- a/sdk/core/azure_core_macros/Cargo.toml
+++ b/sdk/core/azure_core_macros/Cargo.toml
@@ -21,7 +21,6 @@ proc-macro2.workspace = true
 quote.workspace = true
 syn.workspace = true
 tracing.workspace = true
-typespec_client_core = { workspace = true, features = ["http", "json"] }
 
 [dev-dependencies]
 tokio.workspace = true
--- a/sdk/core/azure_core_macros/src/tracing_function.rs
+++ b/sdk/core/azure_core_macros/src/tracing_function.rs
@@ -49,7 +49,7 @@ pub fn parse_function(attr: TokenStream, item: TokenStream) -> Result<TokenStrea
             .into_iter()
             .map(|(name, value)| {
                 quote! {
-                    ::typespec_client_core::tracing::Attribute{key: #name.into(), value: #value.into()}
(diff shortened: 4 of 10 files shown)


sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
```diff
@@ -127,9 +136,10 @@ impl CertificateClient {
                     }
                 };
 
-                let ctx = options.method_options.context.clone();
                 let pipeline = pipeline.clone();
                 let api_version = api_version.clone();
+                let ctx = ctx.clone();
+                let span = span.clone();
```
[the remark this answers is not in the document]
LarryOsterman: `span` is an `Arc<Span>` so it is inexpensive to clone, `ctx` is a `Cow<>` so again, it is inexpensive to clone.

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
```diff
@@ -179,6 +189,24 @@ impl CertificateClient {
                                         let rsp: RawResponse =
                                             pipeline.send(&ctx, &mut request, None).await?;
                                         let (status, headers, body) = rsp.deconstruct();
+                                        if let Some(span) = span {
+                                            // 5xx status codes SHOULD set status to Error.
+                                            // The description should not be set because it can be inferred from "http.response.status_code".
+                                            if status.is_server_error() {
+                                                span.set_status(SpanStatus::Error {
+                                                    description: "".to_string(),
```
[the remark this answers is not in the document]
LarryOsterman: The contract for distributed tracing requires that if the description can be intuited from the status code, the description  MUST be empty.

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
```diff
@@ -92,6 +95,12 @@ impl CertificateClient {
         let certificate_name = certificate_name.to_owned();
         let parameters: Body = parameters.into();
 
+        let mut ctx = options.method_options.context;
+        let span = create_public_api_span(&ctx, None, None);
```
heaths, line 99: Anything authors have to do will be forgotten. The attributes, at least, are one-liners. Why can't this be driven from there?

heaths: Or in the policy? I don't understand why this is even needed. I thought the whole point was that when any client methods that called other client methods, only the outermost client method telemeters.

sdk/core/azure_core_test/src/tracing.rs
```diff
@@ -71,19 +78,19 @@ impl Tracer for MockTracer {
...
         &self,
-        name: &str,
+        name: Cow<'static, str>,
         kind: SpanKind,
         attributes: Vec<Attribute>,
-        _parent: Arc<dyn crate::tracing::Span>,
+        parent: Arc<dyn crate::tracing::Span>,
     ) -> Arc<dyn crate::tracing::Span> {
-        let span = Arc::new(MockSpan::new(name, kind, attributes.clone()));
+        let span = Arc::new(MockSpan::new(name, kind, attributes.clone(), Some(parent)));
```
heaths, line 86: To avoid needing an "overload" and putting stuff in client methods, why not just check the `Context` for a span of the right kind that indicates it's a parent? You could even use a different internal type if necessary so that no one else can create it outside `azure_core`. This feels a bit too fragile to put into implementors' hands.

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
LarryOsterman: The policy is a part of the pipeline. If the pipeline is executed multiple times, then the policy creates multiple top level spans.

But the whole point of this is to have *ONE* top level span. You want one `createCertificate` span with multiple HTTP API spans underneath it, since that's the logical structure of the operation being performed - the client only called `createCertificate` once, so we should only have a single span for that operation.

LarryOsterman:
> The code is mostly fine, but I have some concerns with the approach. This seems to require calling the right API whether you have a parent or not, and presumably all LROs would have to do it. But why even stop there? Wouldn't pageables for a similar reason? We call one client method that likely makes multiple `GET`s so all those would be children.
> 
> I would imagine we could just check if there's a relevant parent span somehow, couldn't we? Either based on a `Span` in the `Context`, or maybe a new internal type that only you could create in `azure_core` and has the info we need.
> 
> This way just feels fragile and is begging for someone to use it wrong.

I thought about updating the macro to make it "just" work, but that requires that the macro be able to identify the code which creates the poller and insert its logic in the correct location. It also requires intrusive changes to the logic (it needs to inject code after the final call to `pipeline.send()`, which means it cannot be isolated from the rest of the code.

That feels obscenely fragile to me.

sdk/core/azure_core_test/src/tracing.rs
LarryOsterman: ? There are two ways of starting a span. Either you start a span as a root span, or you start it as a child span. That's what these two methods do. That's not specific to rust.

The change you're seeing here is in the test code - before my changes, the Mock family of APIs didn't track parent spans, a large part of this change is adding the ability to track parent spans. Rather than trying to figure out some way of hiding the parameter, I just added it as an additional parameter to the existing `new` function.

Given that, I'm confused as to what the change is that you're asking me to make.

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
LarryOsterman: One change I did consider was to drive this functionality inside the poller itself rather than implementing it in the callback. Since the poller wraps the entire operation, it could create the span and create the context internally and then pass it to the callback function.

But that requires some non-trivial changes to the poller infrastructure to ensure that everything works correctly.

If you're open to surgery in the poller (and the pager - that one's next on my list), then I can probably hide these details in the poller at the cost of some API surface change.

LarryOsterman: Created issue https://github.com/Azure/azure-sdk-for-rust/issues/3276 to discuss alternate implementations - hiding this in a future extension seems like an excellent solution for both pollers and pagers.

heaths:
Per our discussion this morning, we can go with this for now to have something in for the next release.

PR #3273, opened 2025-10-29
[Storage] Fix clippy violations in `azure_storage_blob` crate

vincenttran-msft:
As title states 😄

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -250,7 +250,6 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
- "typespec_client_core",
 ]
 
 [[package]]
@@ -268,7 +267,6 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
- "typespec_client_core",
  "url",
 ]
 
@@ -294,7 +292,6 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
- "typespec_client_core",
  "url",
  "uuid",
  "zip",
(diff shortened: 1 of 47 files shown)


sdk/storage/azure_storage_blob/src/models/mod.rs
```diff
@@ -83,4 +83,5 @@ pub use crate::generated::models::{
     StaticWebsite, StorageErrorCode, StorageServiceStats, StorageServiceStatsHeaders,
     UserDelegationKey, UserDelegationKeyHeaders, VecSignedIdentifierHeaders,
 };
+#[allow(unused_imports)]
```
heaths: Why are there unused imports on something you wrote? I get `generated`, but this raises concerns.

vincenttran-msft: [extensions.rs](https://github.com/vincenttran-msft/azure-sdk-for-rust/blob/main/sdk/storage/azure_storage_blob/src/models/extensions.rs) mainly contains extension methods that help cx augment their options bag i.e. `pub fn with_if_not_exists()` for `PageBlobClientCreateOptions`. The only other things it contains are some `TryFrom` and `From` impls for models that are tricky to specify manually, so we provide conversion methods (ie. `HashMap` <--> `BlobTags`)

These are both things I would want re-exported to `models` so that customers don't have to dig into the `extensions` namespace, but happy to take your suggestions or thoughts here.

heaths: My point is that you get this lint only when something is unused and public. Don't just ignore it. Either remove it if nothing public references it, or make it public.

Issue #3236, opened 2025-10-19
Cosmos: Refactor `CosmosClientOptions` to add `PreferredRegions` and Other Features


kundadebdatta:
### Acceptance Criteria:

- Refactor `CosmosClientOptions` to add `PreferredRegions` and Other Features.

Issue #3235, opened 2025-10-19
Cosmos: Fetch Database Account Metadata Information from Gateway


kundadebdatta:
### Acceptance Criteria:

- Build a Trait `AccountProperties` to hold cosmos account metadata information.
- Fetch account properties from Gateway during cosmos client init.

Issue #3234, opened 2025-10-19
Cosmos: Build a Contract For `DocumentServiceRequest` as a Wrapper around AzureCore Http Reuest


kundadebdatta:
### Acceptance Criteria:

- Build a formal contract for `DocumentServiceRequest` as a wrapper around the azure core http reuest.
- The request should hold the routing information, pk-range information etc.

PR #3274, opened 2025-10-29
Cosmos: Adds `Regions`, `AccountProperties` and `CosmosRequest` Builder For Endpoint Based Cross-Regional Routing

kundadebdatta:
## Description

Adds structs and implementation to define the structure of the `CosmosRequest`. Adds a basic implementation of a request handler.

## Changes

- Added the structure and implementation of `CosmosRequest` and `CosmosRequestBuilder`.
- Added the structure and implementation of `AccountProperties`.
- Added the structure and implementation of `Regions`.
- Added logic to fetch account properties from gateway and parse it to `AccountProperties`.
- Added comprehensive unit tests covering functionality and method chaining.
- Updated documentation with usage examples
- Updated CHANGELOG.md

All existing functionality remains unchanged and fully backward compatible.

----
**Additional instructions:**

> Make sure you structure your commits in a sensible way and include a concise but descriptive PR description. PREFER concise descriptions and code. The team knows how this code should work, they don't need explanations of it. The code should be largely self-explanatory.

Fixes Azure/azure-sdk-for-rust#3234
Fixes Azure/azure-sdk-for-rust#3235
Fixes Azure/azure-sdk-for-rust#3236

---

💬 Share your feedback on Copilot coding agent for the chance to win a $200 gift card! Click [here](https://survey3.medallia.com/?EAHeSx-AP01bZqG0Ld9QLQ) to start the survey.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/request_context.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/request_context.rs
@@ -20,7 +20,7 @@ pub struct PartitionKeyRange;
 /// or explicit endpoint), resolved partition ranges, session tokens, and
 /// various internal flags influencing retries and cache refresh behavior.
 #[allow(dead_code)]
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Default)]
 pub struct RequestContext {
     pub force_refresh_address_cache: bool,
     pub original_request_consistency_level: Option<String>, // Use enum if available
@@ -41,32 +41,6 @@ pub struct RequestContext {
     pub use_preferred_locations: Option<bool>,
     pub location_index_to_route: Option<i32>,
     pub location_endpoint_to_route: Option<Url>,
-}
-
-impl Default for RequestContext {
-    fn default() -> Self {
-        Self {
-            force_refresh_address_cache: false,
-            original_request_consistency_level: None,
-            quorum_selected_lsn: 0,
-            global_committed_selected_lsn: 0,
-            store_response: None,
-            perform_local_refresh_on_gone_exception: false,
-            effective_partition_key: None,
-            resolved_partition_key_range: None,
-            session_token: None,
-            performed_background_address_refresh: false,
-            resolved_collection_rid: None,
-            region_name: None,
-            local_region_request: false,
-            is_retry: false,
-            is_partition_failover_retry: false,
-            failed_endpoints: HashMap::new(),
-            use_preferred_locations: None,
-            location_index_to_route: None,
-            location_endpoint_to_route: None,
-        }
-    }
 }
 
 #[allow(dead_code)]


sdk/cosmos/azure_data_cosmos/src/models/account_properties.rs
```diff
@@ -0,0 +1,155 @@
...
+//! present in the sample are intentionally omitted for now.
+
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+/// Represents a single regional endpoint for the Cosmos DB account (readable or writable).
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct AccountLocation {
+    pub name: String,
+    #[serde(rename = "databaseAccountEndpoint")]
```
analogrelay: If you annotate the _type_ with `#[serde(rename_all = "camelCase")]` (which we do for other types), then it will automatically expect `foo_bar_baz` to map to the JSON property `fooBarBaz`. That's what I'd recommend over manually renaming each field. You can still apply `rename` attributes to individual fields to override that when it doesn't work.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -13,6 +15,32 @@ use std::fmt::Display;
...
+    pub application_preferred_regions: Option<Vec<String>>,
+    pub account_initialization_custom_endpoints: Option<HashSet<String>>,
+    pub consistency_level: Option<ConsistencyLevel>,
+    pub request_timeout: Duration,
+    pub enable_remote_region_preferred_for_session_retry: bool,
+    pub enable_partition_level_circuit_breaker: bool,
+    pub disable_partition_level_failover: bool,
+    pub enable_upgrade_consistency_to_local_quorum: bool,
+    pub throughput_bucket: Option<i32>,
+    pub session_retry_options: SessionRetryOptions,
```
analogrelay, line 29: There's a general concern I have across several types in this PR: I'd really like to avoid adding stuff before we need it. This is a good opportunity to make sure we are seriously considering the _intent_ behind functionality in other SDKs. And it's very likely we'll decide to implement things in here in Rust in mostly the same way. However, I don't want to push us directly into copying the existing SDKs because we already copied all the data structures over.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -0,0 +1,291 @@
...
+    pub body: Option<Vec<u8>>,
+    pub query_string: Option<String>,
+    pub continuation: Option<String>,
+    pub entity_id: Option<String>,
+}
+
+impl<'a> CosmosRequest<'a> {
+    /// Creates a new `CosmosRequest` with core operation metadata and optional
+    /// body.
+    #[allow(clippy::too_many_arguments)]
```
analogrelay: The need to suppress warnings like this from clippy also tells me we may not be thinking "in Rust" and just copying across existing concepts. For example, a [builder](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html) might be more appropriate here. Or perhaps `new` only takes a few of the mandatory items and we have `pub fn with_X(mut self, ...) -> Self` "transforming" methods that allow consumers who need to set some of these options set them only when needed.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -0,0 +1,291 @@
...
+        }
+    }
+
+    /// Converts this `CosmosRequest` into a concrete `azure_core::http::Request`.
+    ///
+    /// Inserts partition key and (if present) item options as headers. For
+    /// write operations, sets JSON content type, upsert header (when applicable)
+    /// and attaches the body bytes. Panics if location routing information is
+    /// missing from `request_context`.
+    pub fn to_raw_request(&self) -> Request {
```
analogrelay: Assuming the intent here is for the `CosmosRequest` to be destroyed, since this is probably the final step before sending the request along the wire, this would be more appropriate as an `into_raw_request` method that takes `self` by value. That would allow you to get rid of the clones, since you're taking ownership of those values.

sdk/cosmos/azure_data_cosmos/src/operation_context.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[allow(dead_code)] // For the variants. Can be removed when we have them all implemented.
```
analogrelay, line 5: This is a place where I'm reasonably comfortable with dead code. We know the operation types, and it's very clear that we can just represent them all as enum variants.

sdk/cosmos/azure_data_cosmos/src/regions.rs
```diff
@@ -0,0 +1,106 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/// The `Regions` struct contains the names of Azure regions
+/// that are currently supported by the Azure Cosmos DB service.
+pub struct Regions;
+
+// cSpell:disable
+impl Regions {
```
analogrelay: This should be a module rather than an empty type

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -38,11 +40,13 @@ impl ContainerClient {
             .feed(ResourceType::Containers)
             .item(container_id);
         let items_link = link.feed(ResourceType::Items);
+        let request_handler = RequestHandler::new(pipeline.clone());
```
analogrelay: Let's just move the `pipeline` into the `RequestHandler` rather than cloning, since the client doesn't need it anymore.

sdk/cosmos/azure_data_cosmos/src/handler/request_handler.rs
```diff
@@ -0,0 +1,142 @@
...
+use azure_core::http::Response;
+
+/// Facade wrapping a `CosmosPipeline` to add retry/backoff when sending requests.
+///
+/// Builds a `CosmosRequest`, sets routing context, and dispatches through
+/// `BackOffRetryHandler` which applies exponential backoff for transient
+/// failures (e.g. throttling). Produces typed `Response<T>` while keeping
+/// the underlying pipeline pluggable.
+#[derive(Debug, Clone)]
+pub struct RequestHandler {
```
analogrelay: I do think this could just be rolled into the `CosmosPipeline`, which I was actually looking at renaming to `CosmosConnection` since it holds all the state for a single connection, and gets shared between multiple sub-clients (ContainerClient, DatabaseClient). But I'm OK with having it be separate if that feels cleaner.

One thing we should start to consider is how this common state is shared across those clients. We currently `.clone()` the `CosmosPipeline` when we spawn a sub-client (such as in `DatabaseClient::container_client(...)`). The only real state the `CosmosPipeline` holds is the Azure Core HTTP pipeline, which holds a `Vec<Arc<dyn Policy>>`. Because the policies (which actually do the HTTP work) are shared via `Arc`, every subclient that originates from the `CosmosClient` is using the same instances of the HTTP pipeline types.

We probably want something similar for things like retry. If there is per-account state that the retry handler is holding, we should be sharing that state across all the subclients we create. But if we're `clone`ing or creating a brand new `RequestHandler` in every `ContainerClient`, that's not happening. Every container client ends up with a completely separate instance of `BackOffRetryHandler`.

Grouping everything related to a single "persistent" Cosmos "connection" (I'm using "quotation marks" there because there isn't an actual persistent TCP connection, but to the user we want it to kinda behave as if it is) into a single object we always share via `Arc` might be a good plan here as we evolve things forward.

analogrelay: Perhaps the right approach is to invert the relationship just a bit here and have the `CosmosConnection` (formerly `CosmosPipeline`) be the thing we share between all the clients (as `Arc<CosmosConnection>`) and then it can contain a `RequestHandler` that actually gets used to drive the request. Because we're sharing that top-level `CosmosConnection` object, we know every client is sharing that state.

sdk/cosmos/azure_data_cosmos/src/handler/request_handler.rs
```diff
@@ -0,0 +1,142 @@
...
+            resource_type,
+            None, // resource_id (RID) not yet known here
+            partition_key,
+            body,
+            false,
+            AuthorizationTokenType::Primary,
+            options,
+        );
+        cosmos_request.request_context.location_endpoint_to_route =
+            Option::from(resource_link.url(&self.pipeline.endpoint));
```
analogrelay: nit: This is more idiomatic here


```suggestion
        cosmos_request.request_context.location_endpoint_to_route =
            Some(resource_link.url(&self.pipeline.endpoint));
```

While `Option<T>` does implement `From<T>`, `Option` is such a core concept in the language, that it's usually best to use `Some` and `None` directly as much as possible.

sdk/cosmos/azure_data_cosmos/src/handler/request_handler.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        let item_options = cosmos_request.clone().options.unwrap_or_default();
+
+        let ctx = item_options
+            .method_options
+            .context
+            .with_value(resource_link.clone());
```
analogrelay: We should only clone what we need to clone, which is the context in this case:


```suggestion
        let ctx = cosmos_request
            .options
            .method_options
            .context
            .clone()
            .with_value(resource_link.clone());
```

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
analogrelay: (We talked about this offline, but I wanted to put a comment in here just to track it)

sdk/cosmos/azure_data_cosmos/src/regions.rs
analogrelay: These should just be constants in the `regions.rs` module.

analogrelay (commented):
Some initial comments, mostly just codifying what we chatted about offline. But also a few extra style things

sdk/cosmos/azure_data_cosmos/src/models/account_properties.rs
kundadebdatta: Yes - Thanks for the suggestion @analogrelay! I added it initially on the main `AccountProperties` struct. Right now, I have refactored the coe to cover all the structs that reuires serialization and de-serialization.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
kundadebdatta: I went ahead and cleaned it up as far as I can. Right now, I feel the left over parameters will be required by the global endpoint manager implementation.

If something is unused, I will make sure to clean it up with the following PR.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
kundadebdatta: Refactored this.

sdk/cosmos/azure_data_cosmos/src/handler/request_handler.rs
kundadebdatta: Refactored. Thanks for the suggestion.

sdk/cosmos/azure_data_cosmos/src/regions.rs
kundadebdatta: This is now refactored.

sdk/cosmos/azure_data_cosmos/src/handler/request_handler.rs
kundadebdatta: As discussed, for simplicity we can remove the dependency on `request_handler` for now. We can add it back later if needed. 

Therefore this change is now no longer applies.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
kundadebdatta: As discussed, for simplicity we can remove the dependency on request_handler for now. We can add it back later if needed.

Therefore this change is now no longer applies.

sdk/cosmos/azure_data_cosmos/src/handler/request_handler.rs
kundadebdatta: I thought about this thoroughly, and I think I agree with you. Keeping it simpler at the beginning is fair and we can get rid of the `request_handler` for now, and put all that logic into the pipeline itself. 
 
That way, it will also help us to maintain the shareability across clients. If needed, we can introduce `request_handler` at later phases, but for now we can just keep the design clean and simple.
 
I have refactored the code.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -256,19 +264,21 @@ impl ContainerClient {
...
         options: Option<ItemOptions<'_>>,
     ) -> azure_core::Result<Response<()>> {
-        let options = options.unwrap_or_default();
-        let url = self.pipeline.url(&self.items_link);
-        let mut req = Request::new(url, Method::Post);
-        req.insert_headers(&options)?;
-        req.insert_headers(&partition_key.into())?;
-        req.insert_headers(&ContentType::APPLICATION_JSON)?;
-        req.set_json(&item)?;
+        let item_options = options.clone().unwrap_or_default();
```
analogrelay: nit (Purely a style suggestion, non-blocking): Rust allows "shadowing" variables, i.e. defining a new variable with the same name as an existing variable, which means the new one replaces the old one entirely. So this could just be `let options = options.unwrap_or_default();`. From that point on in the function, `options` would be of type `ItemOptions<'_>` instead of `Option<ItemOptions<'_>>`.

There's nothing wrong with what you did here, it's a matter of deciding which feels clearer. In this case, since there is absolutely no reason we need the original `options: Option<ItemOptions<'_>>` value, I would usually use shadowing here and redefine `options`.

analogrelay: Also, we shouldn't need to clone the options. If there are pieces within the options we need to clone (like the context), we can do so, but in general we want to avoid `clone`s unless absolutely necessary.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -256,19 +264,21 @@ impl ContainerClient {
...
+        let ctx = item_options
+            .method_options
+            .context
+            .with_value(self.items_link.clone());
+        let body = serde_json::to_vec(&item)?;
+        let builder = CosmosRequestBuilder::new(OperationType::Create, ResourceType::Items);
+        let cosmos_request = builder
+            .headers(&options)
```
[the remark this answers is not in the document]
analogrelay: Yeah, I suspect fixing the above will resolve Copilot's concern here (which is valid), since `options` would be an `ItemOptions<'_>` by this point.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -96,6 +98,34 @@ impl CosmosClient {
...
+        Ok(client)
+    }
+
+    /// Retrieves the Cosmos DB account ("database account") properties.
+    ///
+    /// # Arguments
+    /// * `options` - Optional request options (currently unused for custom
+    ///   headers, but the context can carry per-call metadata for tracing or
+    ///   cancellation).
+    pub async fn get_database_account(
```
analogrelay: Should this be `pub` or just `pub(crate)`? Is it an API we'd want to expose to users?

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -96,6 +98,34 @@ impl CosmosClient {
         })
     }
 
+    #[allow(dead_code)]
+    async fn initialize(client: CosmosClient) -> azure_core::Result<Self> {
```
analogrelay: Let's remove this if we're not calling it yet. We know what we need to do here and can add it in when it's time for it.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -198,17 +228,18 @@ impl CosmosClient {
...
-        let url = self.pipeline.url(&self.databases_link);
-        let mut req = Request::new(url, Method::Post);
-        req.insert_headers(&options.throughput)?;
-        req.insert_headers(&ContentType::APPLICATION_JSON)?;
-        req.set_json(&RequestBody { id })?;
+        let body = serde_json::to_vec(&RequestBody { id })?;
+        let builder = CosmosRequestBuilder::new(OperationType::Create, ResourceType::Databases);
+        let cosmos_request = builder
+            .headers(&options.throughput)
+            .body(Some(body))
```
analogrelay: nit: I think it would be good to have a `.json<T: Serialize>(mut self, value: T) -> Self` method on `CosmosRequestBuilder` that handles the serialization and setting of the body.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -67,10 +64,14 @@ impl ContainerClient {
...
     ) -> azure_core::Result<Response<ContainerProperties>> {
         let options = options.unwrap_or_default();
-        let url = self.pipeline.url(&self.link);
-        let mut req = Request::new(url, Method::Get);
+        let cosmos_request =
+            CosmosRequestBuilder::new(OperationType::Read, ResourceType::Containers).build();
         self.pipeline
-            .send(options.method_options.context, &mut req, self.link.clone())
+            .send(
+                cosmos_request.unwrap(),
```
analogrelay: We shouldn't be `unwrap`ing `Results in product code. In tests, where you know the value should be `Ok`, it's reasonable, but in product code we should use one of the following:

1. Ideally, the try operator `?` to propagate the error to the caller (your error types have to be convertible to each other, but we should just be using `azure_core::Result<T>`, which uses `azure_core::Error` as the error type everywhere)
2. If extra context would be helpful to the user, `Result::map_err` which takes a closure that receives the error and allows us to wrap it in a more detailed error (this is like wrapping exceptions in .NET) using `azure_core::Error::with_error`.
3. If the error is truly unexpected and "should never happen", we should use `Result::expect`, which is like `unwrap` but takes a string describing _why_ it should be safe to unwrap the value.

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
```diff
@@ -76,10 +74,15 @@ impl DatabaseClient {
...
         let options = options.unwrap_or_default();
-        let url = self.pipeline.url(&self.link);
-        let mut req = Request::new(url, Method::Get);
+        let builder = CosmosRequestBuilder::new(OperationType::Read, ResourceType::Databases);
+        let cosmos_request = builder.build();
+
         self.pipeline
-            .send(options.method_options.context, &mut req, self.link.clone())
+            .send(
+                cosmos_request.unwrap(),
```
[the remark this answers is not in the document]
analogrelay: Yep, this is what I mentioned above. Copilot is correct here, unwrapping Results in production code is usually an anti-pattern.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -13,6 +15,32 @@ use std::fmt::Display;
 #[derive(Clone, Default, Debug)]
 pub struct CosmosClientOptions {
     pub client_options: ClientOptions,
+    pub application_name: Option<String>,
+    pub application_region: Option<String>,
+    pub application_preferred_regions: Option<Vec<String>>,
+    pub account_initialization_custom_endpoints: Option<HashSet<String>>,
+    pub consistency_level: Option<ConsistencyLevel>,
+    pub request_timeout: Duration,
```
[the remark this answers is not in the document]
analogrelay: Yeah, this should be `Option<Duration>`. In general we should use `Option` to represent optional values and then later on use `request_timeout.unwrap_or(SOME_DEFAULT_VALUE)` when specifying a default. I know in .NET we might use something like `TimeSpan.Zero` to represent a default timeout, but we should use `Option` in Rust. That way, `#[derive(Default)]` will just work too.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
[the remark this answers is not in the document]
kundadebdatta: Refactored. Thanks for the feedback.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -344,17 +354,19 @@ impl ContainerClient {
...
-        self.pipeline
-            .send(options.method_options.context, &mut req, link)
-            .await
+        let body = serde_json::to_vec(&item)?;
+        let item_options = options.clone().unwrap_or_default();
+        let ctx = item_options.method_options.context.with_value(link.clone());
+
+        let builder = CosmosRequestBuilder::new(OperationType::Replace, ResourceType::Items);
+        let cosmos_request = builder
+            .headers(&options)
```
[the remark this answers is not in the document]
kundadebdatta: Refactored. Thanks for the feedback.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -431,20 +443,22 @@ impl ContainerClient {
...
+        let item_options = options.clone().unwrap_or_default();
+        let ctx = item_options
+            .method_options
+            .context
+            .with_value(self.items_link.clone());
+
+        let body = serde_json::to_vec(&item)?;
+        let builder = CosmosRequestBuilder::new(OperationType::Upsert, ResourceType::Items);
+        let cosmos_request = builder
+            .headers(&options)
```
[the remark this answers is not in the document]
kundadebdatta: Refactored. Thanks for the feedback.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -525,15 +544,17 @@ impl ContainerClient {
...
-        self.pipeline
-            .send(options.method_options.context, &mut req, link)
-            .await
+        let item_options = options.clone().unwrap_or_default();
+        let ctx = item_options.method_options.context.with_value(link.clone());
+
+        let builder = CosmosRequestBuilder::new(OperationType::Delete, ResourceType::Items);
+        let cosmos_request = builder
+            .partition_key(Some(partition_key.into()))
+            .headers(&options)
```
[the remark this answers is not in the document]
kundadebdatta: Refactored. Thanks for the feedback.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -598,18 +619,19 @@ impl ContainerClient {
...
+        let item_options = options.clone().unwrap_or_default();
+        let ctx = item_options.method_options.context.with_value(link.clone());
 
-        self.pipeline
-            .send(options.method_options.context, &mut req, link)
-            .await
+        let builder = CosmosRequestBuilder::new(OperationType::Patch, ResourceType::Items);
+        let cosmos_request = builder
+            .partition_key(Some(partition_key.into()))
+            .headers(&options)
```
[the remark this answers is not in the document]
kundadebdatta: Refactored. Thanks for the feedback.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -73,28 +74,32 @@ impl CosmosPipeline {
...
+        cosmos_request.request_context.location_endpoint_to_route =
+            Some(resource_link.url(&self.endpoint));
+
+        // Prepare a cloneable ResourceLink to avoid moving it, allowing the closure to be Fn
+        let resource_link_for_sender = resource_link.clone();
+        let sender = move |req: &mut CosmosRequest| {
+            let ctx = context.clone();
+            let mut raw_req = req.clone().into_raw_request();
+            let url = resource_link_for_sender.clone();
+            async move { self.send_raw(ctx, &mut raw_req, url).await }
```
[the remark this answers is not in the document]
analogrelay: I wouldn't use an Arc here. But Copilot is right, we need to figure out the right way to use this. I'll think a bit about it, I don't have a clear plan right now and want to finish my first pass at this review.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -0,0 +1,395 @@
...
+        resource_id: Option<String>,
+        partition_key: Option<PartitionKey>,
+        body: Option<Vec<u8>>,
+        authorization_token_type: AuthorizationTokenType,
+    ) -> Self {
+        Self {
+            operation_type,
+            resource_type,
+            resource_id: resource_id.clone(),
+            resource_address: resource_id,
```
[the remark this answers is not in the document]
analogrelay: This is a silly suggestion Copilot.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -0,0 +1,395 @@
...
+    pub fn is_read_only_request(&self) -> bool {
+        matches!(
+            self.operation_type,
+            OperationType::Read
+                | OperationType::Query
+                | OperationType::SqlQuery
+                | OperationType::QueryPlan
+                | OperationType::Head
+                | OperationType::HeadFeed
+        )
```
analogrelay: nit: I'd generally make this a function on `OperationType`, i.e. `OperationType::is_read_only(self)`.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -0,0 +1,395 @@
...
+    }
+    pub fn header<K, V>(mut self, key: K, value: V) -> Self
+    where
+        K: Into<HeaderName>,
+        V: Into<HeaderValue>,
+    {
+        self.headers.insert(key, value);
+        self
+    }
+    pub fn body(mut self, body: Option<Vec<u8>>) -> Self {
```
analogrelay: I'd suggest this take `Vec<u8>` instead of `Option<Vec<u8>>`. The caller is always going to be passing in `Some(..)` here (unless they want to set the body and then clear it later, which we shouldn't encourage anyway 😅 ), so we can wrap it for them and the API feels cleaner.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -0,0 +1,395 @@
...
+        self.body = body;
+        self
+    }
+    pub fn authorization_token_type(mut self, ty: AuthorizationTokenType) -> Self {
+        self.authorization_token_type = ty;
+        self
+    }
+
+    pub fn partition_key(mut self, pk: Option<PartitionKey>) -> Self {
```
analogrelay: Same comment here about taking `PartitionKey` instead of `Option<PartitionKey>`

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -0,0 +1,395 @@
...
+        self.partition_key = pk;
+        self
+    }
+
+    /// Finish construction and return the immutable `CosmosRequest`.
+    pub fn build(self) -> Result<CosmosRequest, String> {
```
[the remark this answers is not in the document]
analogrelay: Yep, we should generally use `azure_core::Result<CosmosRequest>` (which is just an alias for `Result<CosmosRequest, azure_core::Error>`) whenever we need to be able to return an error.

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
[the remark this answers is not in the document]
kundadebdatta: Thanks for the suggestion. I have refactored the code.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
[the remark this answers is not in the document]
kundadebdatta: Removed the unnecessary assignement of `resource_address`

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -0,0 +1,395 @@
...
+    pub partition_key_range_id: String,
+}
+
+/// Internal representation of a Cosmos DB operation before it is converted
+/// into a wire-level `azure_core::http::Request`.
+///
+/// It collects operation intent (create/read/query/etc.), resource routing
+/// information, partition key, optional item-level options and flags that
+/// influence retry or gateway behaviors.
+#[derive(Clone)]
```
analogrelay: Do you think we could derive `PartialEq` and `Eq` here? I think all the field types are equatable (even the headers and body) and it would make some testing and other scenarios easier.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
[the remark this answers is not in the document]
kundadebdatta: Refacrored.

sdk/cosmos/azure_data_cosmos/src/operation_context.rs
```diff
@@ -0,0 +1,22 @@
...
+/// Placeholder for operation type.
+pub enum OperationType {
+    Invalid,
```
analogrelay: What is `Invalid` used for? Is it a marker for a scenario in which the user didn't specify one? In .NET/Java there can be a use in defining an "unknown" value for enums so that we have a way to represent `0`, but that's not necessary in Rust since the compiler won't allow casting from an integer to an enum. You should only need to provide the variants that someone will actually _use_.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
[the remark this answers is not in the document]
kundadebdatta: Refactored.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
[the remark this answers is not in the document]
kundadebdatta: As of now, this change appears to be unnecessary. The compilation doesn't fail and there are no known gaps in terms of ownership.

analogrelay (changes requested):
Style comments, but nothing critical.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
kundadebdatta: Yes - Thanks for the suggestion @analogrelay ! this has already been refactored.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
kundadebdatta: Thanks for the suggestion here @analogrelay. This has already been refactored. It was done previously given the reuest_handler approach, which no longer applicable now.

Therefore, this has been removed.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -67,10 +64,14 @@ impl ContainerClient {
...
     ) -> azure_core::Result<Response<ContainerProperties>> {
         let options = options.unwrap_or_default();
-        let url = self.pipeline.url(&self.link);
-        let mut req = Request::new(url, Method::Get);
+        let cosmos_request =
+            CosmosRequestBuilder::new(OperationType::Read, ResourceType::Containers).build();
         self.pipeline
-            .send(options.method_options.context, &mut req, self.link.clone())
+            .send(
+                cosmos_request?,
```
analogrelay: nit: I'd put the `?` on the end of the `build()` call above. It's generally idiomatic to handle `Result`s as close as possible to the call that generates them

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
analogrelay, line 94: nit: I'd just make the entire closure async rather than just this block. But it's fine either way.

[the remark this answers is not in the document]
analogrelay: Yep, it's fine now that I think about it. `CosmosPipeline` is considered [a `Sync` type](https://doc.rust-lang.org/stable/nomicon/send-and-sync.html) by Rust so it's safe to pass references across async closures.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -73,28 +74,32 @@ impl CosmosPipeline {
...
-            .map(Into::into)
+        cosmos_request.request_context.location_endpoint_to_route =
+            Some(resource_link.url(&self.endpoint));
+
+        // Prepare a cloneable ResourceLink to avoid moving it, allowing the closure to be Fn
+        let resource_link_for_sender = resource_link.clone();
+        let sender = move |req: &mut CosmosRequest| {
+            let ctx = context.clone();
+            let mut raw_req = req.clone().into_raw_request();
+            let url = resource_link_for_sender.clone();
```
analogrelay: Now that I look at this, do we need to clone the values? We're moving these values into the closure and not using them anywhere else. I believe a clone would only be needed if we wanted to keep using the context.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
kundadebdatta: Thanks! I have refactored the code.

sdk/cosmos/azure_data_cosmos/src/operation_context.rs
kundadebdatta: I used it to identify any invalid operation, but as you called out for Enums, we won't need it, since we have the source of truth.

Therefore, I removed the `Invalid` entry.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
kundadebdatta: Thanks! I have refactored the code.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
kundadebdatta: Thanks! I have refactored the code.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
kundadebdatta: Thanks! I have refactored the code.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
kundadebdatta: This is now updated. Implemented the `Eq` and `PartialE` traites for `CosmosRequest`.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
kundadebdatta: Yes - thanks for the suggestion. This has been refactored.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
kundadebdatta: Removed.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
kundadebdatta: Removed the get_database_account call for the time being. I will introduce this later with the `global_endpoint_manager` PR.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
kundadebdatta: `resource_link_for_sender` is now removed to skip clonning.

PR #3275, opened 2025-10-30
cosmos: updates to the query engine interface for hybrid search

analogrelay:
Two small changes to the query engine interface based on work I'm doing to prep for Hybrid Search support:

1. Allow the pipeline to return a `None` query, indicating that **every** request will come with an override query. We validate that with a panic in dev builds, but return an `Err` in release builds.

2. Update the `index` field on `QueryRequest` to be called `id`, to more accurately describe that it's an opaque ID rather than a sequential index. In Hybrid search, we encode things in this ID like the specific component query the data represents and that's not really an ordered concept as far as the SDK is concerned. This number is just basically just a token that the engine provides to the SDK which must be copied across to the response so we can correlate them.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/query/engine.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/query/engine.rs
@@ -76,9 +76,9 @@ pub trait QueryPipeline: Send {
     /// So, for any given partition key range, page n's results must be earlier in the `data` vector than page n+1's results.
     /// Data from different partition key ranges may be interleaved, as long as each partition key range's pages are in order.
     ///
-    /// The pipeline will use the [`QueryResult::request_index`] field to validate this.
+    /// The pipeline will use the [`QueryResult::request_id`] field to validate this.
     ///
-    /// When providing data from a draining request (i.e. a request with `drain = true`), all pages for that draining request can share the same [`QueryResult::request_index`].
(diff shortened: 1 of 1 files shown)


sdk/cosmos/azure_data_cosmos/tests/framework/query_engine.rs
```diff
@@ -44,6 +44,13 @@ pub struct MockQueryEngine {
     pub create_error: Mutex<Option<azure_core::Error>>,
     /// Configuration for what kind of QueryRequest the pipeline should return.
     pub query_request_config: Mutex<Option<QueryRequestConfig>>,
+    /// An optional "rewritten" query to assume is present in the query plan.
+    ///
+    /// The double-Option is a little awkward, but it allows us to represent three states:
+    /// - None: No rewritten query present.
+    /// - Some(None): The pipeline should return `None` from [`QueryPipeline::query`].
+    /// - Some(Some(String)): The pipeline should return the given string from [`QueryPipeline::query`].
+    pub rewritten_query: Option<Option<String>>,
```
Pilchie, line 53: Would an enum be clearer here?

analogrelay: Yeah, probably. This being test code I didn't think too hard about that 😅

PR #3279, opened 2025-10-30
Moved otel support for polling into poller

LarryOsterman:
Removed custom code from `create_certificate` API code and moved the corresponding logic into the `Poller`.

This involved adding an additional parameter to the `callback` for `Poller::with_callback` that provides the `ctx` used for the pipeline.

Bonus: Instead of using a tuple to maintain the state of the "stream" created by the `Poller`, encapsulate the state in a structure to make it easier to modify the state of the "stream" moving forward (when there were 3 fields in the state, it wasn't a huge deal, but with 5 fields, it was getting unwieldy).

NOTE: This is still a draft PR, the next step is to work on similar changes for pagers.

### DIFF
--- a/sdk/core/azure_core/src/http/pager.rs
+++ b/sdk/core/azure_core/src/http/pager.rs
@@ -866,7 +866,6 @@ where
             // Get the `continuation_token` to pick up where we left off, or None for the initial page,
             // but don't override the terminal `State::Done`.
 
-            tracing::trace!("current stream state: {:?}", stream_state.added_span);
             if stream_state.state != State::Done {
                 let result = match stream_state.continuation_token.lock() {
                     Ok(next_token) => match next_token.as_deref() {
@@ -899,7 +898,6 @@ where
                     // At the very start of polling, create a span for the entire request, and attach it to the context
                     let span = create_public_api_span(&stream_state.ctx, None, None);
                     if let Some(ref s) = span {
-                        tracing::trace!("initial page request, adding span");
                         stream_state.added_span = true;
                         stream_state.ctx = stream_state.ctx.with_value(s.clone());
                     }
@@ -949,24 +947,19 @@ where
                     if let Ok(mut token) = stream_state.continuation_token.lock() {
                         *token = None;
                     }
-                    tracing::trace!("final page received, finalizing span if added");
                     // When the result is done, finalize the span. Note that we only do that if we created the span in the first place,
                     // otherwise it is the responsibility of the caller to end their span.
                     if stream_state.added_span {
-                        tracing::trace!("final page received, finalizing span if present");
                         if let Some(span) = stream_state.ctx.value::<Arc<dyn Span>>() {
                             // P is unconstrained, so it's not possible to retrieve the status code for now.
 
                             span.end();
                         }
-                    } else {
-                        tracing::trace!("final page received, no span added");
                     }
                     (Ok(response), State::Done)
                 }
             };
 
-            tracing::trace!("Returning Stream State: {:?}", stream_state.added_span);
             stream_state.state = next_state;
             Some((item, stream_state))
         },


sdk/core/azure_core/src/http/pager.rs
```diff
@@ -723,56 +750,118 @@ enum State<T> {
...
 }
 
+struct PagerStreamState<'a, C, F, G, S> {
+    state: State<C>,
+    make_request: F,
+    get_next: G,
+    set_next: S,
+    ctx: Context<'a>,
+    added_span: bool,
+}
```
heaths: FWIW, G and S are gone and I felt the type name was longer than necessary for a private module member, so I'm making it like so:

```rust
#[derive(Debug)]
struct StreamState<'a, C, F> {
    state: State<C>,
    make_request: F,
    continuation_token: Arc<Mutex<Option<String>>>,
    ctx: Context<'a>,
    added_span: bool,
}
```

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -723,56 +750,118 @@ enum State<T> {
...
     #[cfg(target_arch = "wasm32")] S: Fn(Option<&str>) + 'static,
 >(
     make_request: F,
+    ctx: Context<'a>,
     get_next: G,
     set_next: S,
-) -> impl Stream<Item = crate::Result<P>> + 'static {
+) -> impl Stream<Item = crate::Result<P>> + 'static
+where
+    'a: 'static,
```
heaths: Couldn't you just use `'static` instead of declaring and using `'a`?

sdk/core/azure_core/CHANGELOG.md
```diff
@@ -25,6 +25,8 @@
 - Renamed `BearerTokenCredentialPolicy` to `BearerTokenAuthorizationPolicy`.
 - Renamed `RetryPolicy::get_retry_headers()` to `RetryPolicy::retry_headers()`
 - Renamed `xml::read_xml()` to `xml::from_xml()` congruent with `json::from_json()`.
+- Changed `PollerOptions::frequency` from `Option<Duration>` to `Duration`.
+- Added `Context` field to `PollerOptions`. Client methods which return `Poller` objects should accept a `PollerOptions` in their `method_options` field instead of a `ClientMethodOptions`.
```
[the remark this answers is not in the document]
LarryOsterman: @copilot apply changes based on [this feedback](https://github.com/Azure/azure-sdk-for-rust/pull/3279#discussion_r2500280084) to the current pull request

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -314,7 +318,7 @@ impl<P: Page> ItemIterator<P> {
     /// }
     /// let url = "https://example.com/my_paginated_api".parse().unwrap();
     /// let mut base_req = Request::new(url, Method::Get);
-    /// let pager = ItemIterator::from_callback(move |next_link: PagerState<Url>, ctx| {
+    /// let pager = ItemIterator::from_callback(move |next_link: PagerState<Url>, ctx: Context| {
```
heaths, line 321: Nit: not really necessary since it can be inferred. Most times, `next_link` can as well but I know that was already there and probably because that couldn't be inferred by code therein.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -923,6 +935,31 @@ where
                     if let Ok(mut token) = stream_state.continuation_token.lock() {
                         *token = None;
                     }
+                    // When the result is done, finalize the span. Note that we only do that if we created the span in the first place,
```
heaths, line 950: What about in the error case e.g., line 919 above?

heaths: I suppose that's more of a non-HTTP error, though. Still, worth asking the question.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -923,6 +935,31 @@ where
...
+                            //     });
+                            // }
+                            // if response.status().is_client_error()
+                            //     || response.status().is_server_error()
+                            // {
+                            //     span.set_attribute(
+                            //         "error.type",
+                            //         response.status().to_string().into(),
+                            //     );
+                            // }
```
heaths: Elide? Or is this to remember for something later?

sdk/core/azure_core/src/http/poller.rs
```diff
@@ -419,7 +438,7 @@ where
     ///                 let operation_url = format!("https://example.com/operations/{}", result.id).parse()?;
     ///                 Ok(PollerResult::InProgress {
     ///                     response: resp,
-    ///                     retry_after: None,
+    ///                     retry_after: poller_options.frequency,
```
heaths, line 449: I don't want emitters to have to do this. The less they have to do the better (more so thinking human-written convenience methods). That's why it was `Option<Duration>`. Couldn't we instead keep track of the set-or-default value in the stream? It's `Clone + Copy`.

Not blocking.

sdk/core/azure_core/src/http/poller.rs
```diff
@@ -600,77 +640,160 @@ where
...
+                    match &res {
+                        Ok(response) => {
+                            // When the result is done, finalize the span. Note that we only do that if we created the span in the first place,
+                            // otherwise it is the responsibility of the caller to end their span.
+                            if let Some(span) = ctx.value::<Arc<dyn Span>>() {
+                                // 5xx status codes SHOULD set status to Error.
+                                // The description should not be set because it can be inferred from "http.response.status_code".
+                                if response.status().is_server_error() {
+                                    span.set_status(SpanStatus::Error {
+                                        description: "".to_string(),
```
heaths, line 777: Shouldn't we at least call `to_string()` on the `StatusCode`? Ideally, we use the "code" embedded in our error since that logic is already the Azure standard error description or falls back to the stringified status code.

sdk/core/azure_core/src/http/pager.rs
heaths: I see you handle this branch in `poller.rs`.

sdk/core/azure_core/src/http/pager.rs
heaths: I see you more or less have this code executing in `poller.rs`.

sdk/core/azure_core/CHANGELOG.md
```diff
@@ -26,6 +26,8 @@
...
```
heaths: Nit: I try to lexically sort these in the CHANGELOG so it's less chaotic.

sdk/keyvault/azure_security_keyvault_certificates/src/generated/clients/certificate_client.rs
```diff
@@ -31,7 +31,7 @@ use azure_core::{
     error::CheckSuccessOptions,
     fmt::SafeDebug,
     http::{
-        pager::{PagerResult, PagerState},
+        pager::{PagerOptions, PagerResult, PagerState},
```
heaths: Why was this necessary? Did the emitter not add it?

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
```diff
@@ -79,9 +76,9 @@ impl CertificateClient {
         &self,
         certificate_name: &str,
         parameters: RequestContent<CreateCertificateParameters>,
-        options: Option<CertificateClientCreateCertificateOptions<'_>>,
```
heaths, line 82: That should still be necessary because it has a `Context`. Did you not run the emitter on this code? Will definitely need to before we can merge (blocking).

sdk/keyvault/azure_security_keyvault_certificates/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Removed `CertificateClient::begin_create_certificate()`.
 - Removed `CertificateClient::resume_create_certificate()`.
 - Removed `wait()` function from `Poller<CertificateOperation>`.
+- Changed `PollerOptions::frequency` from `Option<Duration>` to `Duration`.
```
heaths: Sort within section.

sdk/keyvault/azure_security_keyvault_secrets/tests/secret_client.rs
```diff
@@ -290,3 +290,218 @@ async fn round_trip_secret_verify_telemetry(ctx: TestContext) -> Result<()> {
...
+        let secret_client = {
+            let mut options = SecretClientOptions::default();
+            recording.instrument(&mut options.client_options);
+            SecretClient::new(
+                recording.var("AZURE_KEYVAULT_URL", None).as_str(),
+                recording.credential(),
+                Some(options),
+            )
+        }?;
+        for i in 1..=SECRET_COUNT {
```
heaths: What are you, a VB developer?!

sdk/keyvault/azure_security_keyvault_secrets/tests/secret_client.rs
```diff
@@ -290,3 +290,218 @@ async fn round_trip_secret_verify_telemetry(ctx: TestContext) -> Result<()> {
...
+                ..Default::default()
+            }],
+        },
+    )
+    .await;
+
+    validate_result
+}
+
+// Commented out for now until ItemIterator->PageIterator works for continuation tokens.
```
heaths: That will work once you rebase.

.gitignore
```diff
@@ -20,6 +20,7 @@ vcpkg_installed/
 
 # Editor user customizations.
 .vscode/launch.json
+.vscode/mcp.json
```
heaths, line 23: Oh, I thought you meant you added this file. At some point we're supposed to add the Azure SDK MCP, so if that's all you have we could. I *hope* there's some way to merge user-specific MCPs, like a user directory or something? That actually raises an interesting question, @maririos.

sdk/core/azure_core/src/http/pager.rs
LarryOsterman: I added it because next_link (which can also be inferred) had an explicit type specified and thought that it made sense to document the type of both.

sdk/core/azure_core/src/http/pager.rs
LarryOsterman: Nope, it's a bug, and a good one.

sdk/core/azure_core/src/http/pager.rs
LarryOsterman: Remembered for later - see the comment at line 953 for this. I'll remove it though because it's easy enough to reconstruct.

sdk/core/azure_core/src/http/poller.rs
LarryOsterman: The telemetry guidelines actually require that the description be an empty string if the description can be inferred from the status code

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
LarryOsterman: This is hand written, not emitted - so running the emitter won't change it.

And nothing seems to be upset about not having a lifetime parameter - no idea why though.

sdk/core/typespec_client_core/src/http/context.rs
```diff
@@ -123,6 +123,20 @@ impl<'a> Context<'a> {
...
     }
 
+    /// Transforms this [`Context`] into a new [`Context`] that owns the underlying data, cloning it if necessary.
+    ///
+    /// Clone the underlying data in the [`Context`] and return it in a new owned [`Context`].
+    #[must_use]
+    pub fn to_owned(&self) -> Context<'static> {
+        let type_map = match &self.type_map {
+            Cow::Owned(o) => o.clone(),
+            Cow::Borrowed(o) => (*o).clone(),
```
heaths, line 133: This works, but probably could've just called `self.type_map.clone()`. Maybe future iteration (or never).

heaths (approved):

sdk/core/typespec_client_core/src/http/context.rs
LarryOsterman: Doh!

LarryOsterman: I have to make another change to the PR to work around a pipeline issue, I'll fix it there.

LarryOsterman: Actually - that didn't work - you need to unwrap the `Cow`.

heaths (approved):

heaths (approved):

PR #3284, opened 2025-11-03
[Storage] `set/get_access_policy` for `ContainerClient`

vincenttran-msft:
`.tsp`: https://github.com/Azure/azure-rest-api-specs/pull/38633
- Adds implementations for `set_access_policy()` and `get_access_policy()`
- Regenerates against rust emitter `v0.26.0`
- Refactors all code to consume the new `Pager` response model that previously used `PageIterator`


sdk/storage/azure_storage_blob/tests/blob_container_client.rs
```diff
@@ -129,116 +142,116 @@ async fn test_list_blobs(ctx: TestContext) -> Result<(), Box<dyn Error>> {
     Ok(())
```
vincenttran-msft, line 129: Relevant test case that I refactored to now work against the newly returned `Pager`, but this is hitting an error:
>Error: Error { context: Custom(Custom { kind: DataConversion, error: Error("expected value", line: 1, column: 1) }) }

sdk/storage/azure_storage_blob/src/generated/clients/blob_container_client.rs
```diff
@@ -836,55 +836,53 @@ impl BlobContainerClient {
...
+                Ok(match res.next_marker {
+                    Some(next_marker) if !next_marker.is_empty() => PagerResult::More {
+                        response: rsp,
+                        continuation: next_marker,
+                    },
+                    _ => PagerResult::Done { response: rsp },
+                })
+            }
+        }))
     }
```
vincenttran-msft, line 858: Example diff of what changed in generated functions for `PageIterator` -> `Pager`.

sdk/storage/azure_storage_blob/tests/blob_container_client.rs
```diff
@@ -394,3 +399,41 @@ async fn test_find_blobs_by_tags_container(ctx: TestContext) -> Result<(), Box<d
...
+    // Assert
+    let response = container_client.get_access_policy(None).await?;
+    let signed_identifiers = response.into_body()?;
+    let response_id = signed_identifiers[0].id.clone();
+    let response_access_policy = signed_identifiers[0].access_policy.clone();
+
+    println!("Response Access Policy: {:?}", response_access_policy);
+    println!("Response ID: {:?}", response_id);
+
+    Ok(())
```
vincenttran-msft, line 463: >running 1 test
Response Access Policy: None
Response ID: None

This test sets a `testid` and a `rw` permission. This works flawlessly on the `set_access_policy()` side.

However, `get_access_policy()` is unexpectedly `None` in these fields, despite the XML showing otherwise:
```xml
<SignedIdentifiers>
    <SignedIdentifier>
        <Id>testid</Id>
        <AccessPolicy>
            <Start>2025-11-05T21:14:02.3927308Z</Start>
            <Expiry>2025-11-05T21:14:12.3927155Z</Expiry>
            <Permission>rw</Permission>
        </AccessPolicy>
    </SignedIdentifier>
</
```

We are generating against this feature branch's [.tsp](https://github.com/Azure/azure-rest-api-specs/blob/77e27a51867d55d2a4176ee82d19b1dc397c2b0a/specification/storage/Microsoft.BlobStorage/tspconfig.yaml). [Here ](https://github.com/Azure/azure-rest-api-specs/blob/77e27a51867d55d2a4176ee82d19b1dc397c2b0a/specification/storage/Microsoft.BlobStorage/models.tsp#L438)is a link to the relevant TypeSpec section (SignedIdentifiers in models.tsp).
Thanks!

jhendrixMSFT: The method returns a `Vec<SignedIdentifier>` which conceptually makes sense but does not match the XML structure.

Unfortunately, we can't implement `Deserialize` for `Vec<SignedIdentifier>` to resolve it as this runs into the orphan problem.

I think the only solution at present is to update the tsp to accurately reflect the shape of the XML (feel free to come up with a better name for the field than `items`).
```typespec
/** Represents an array of signed identifiers */
model SignedIdentifiers {
  @Xml.unwrapped
  @Xml.name("SignedIdentifier")
  items: Array<SignedIdentifier>;
}
```
This will change some method signatures but resolves the issue.  However, I did run into a problem with this sample data deserializing the time stamps.  I believe this is the same issue as https://github.com/Azure/typespec-rust/issues/618.

PR #3286, opened 2025-11-04
Implement Pager resumption even through into_pages

heaths:
By design, we didn't support resuming `Pager` (`ItemIterator`) but, upon further discussion, decided we can but will resume from the current page until all items on the current page are processed.
    
This also fixes a problem that, because of the design, we didn't pass through resumption to the `PageIterator` at all - it couldn't even resume as a `PageIterator` for subsequent pages.

### DIFF
--- a/sdk/keyvault/azure_security_keyvault_certificates/src/generated/clients/certificate_client.rs
+++ b/sdk/keyvault/azure_security_keyvault_certificates/src/generated/clients/certificate_client.rs
@@ -668,7 +668,7 @@ impl CertificateClient {
     /// * `options` - Optional parameters for the request.
     #[tracing::function("KeyVault.getCertificates")]
     pub fn list_certificate_properties(
-        &'_ self,
+        &self,
         options: Option<CertificateClientListCertificatePropertiesOptions<'_>>,
     ) -> Result<Pager<ListCertificatePropertiesResult>> {
         let options = options.unwrap_or_default().into_owned();
@@ -749,7 +749,7 @@ impl CertificateClient {
     /// * `options` - Optional parameters for the request.
     #[tracing::function("KeyVault.getCertificateVersions")]
     pub fn list_certificate_properties_versions(
-        &'_ self,
+        &self,
         certificate_name: &str,
         options: Option<CertificateClientListCertificatePropertiesVersionsOptions<'_>>,
     ) -> Result<Pager<ListCertificatePropertiesResult>> {
@@ -834,7 +834,7 @@ impl CertificateClient {
     /// * `options` - Optional parameters for the request.
     #[tracing::function("KeyVault.getDeletedCertificates")]
     pub fn list_deleted_certificate_properties(
-        &'_ self,
+        &self,
         options: Option<CertificateClientListDeletedCertificatePropertiesOptions<'_>>,
     ) -> Result<Pager<ListDeletedCertificatePropertiesResult>> {
         let options = options.unwrap_or_default().into_owned();
@@ -914,7 +914,7 @@ impl CertificateClient {
     /// * `options` - Optional parameters for the request.
     #[tracing::function("KeyVault.getCertificateIssuers")]
     pub fn list_issuer_properties(
-        &'_ self,
+        &self,
         options: Option<CertificateClientListIssuerPropertiesOptions<'_>>,
     ) -> Result<Pager<ListIssuerPropertiesResult>> {
         let options = options.unwrap_or_default().into_owned();
--- a/sdk/keyvault/azure_security_keyvault_keys/src/generated/clients/key_client.rs
+++ b/sdk/keyvault/azure_security_keyvault_keys/src/generated/clients/key_client.rs
@@ -670,7 +670,7 @@ impl KeyClient {
     /// * `options` - Optional parameters for the request.
     #[tracing::function("KeyVault.getDeletedKeys")]
     pub fn list_deleted_key_properties(
-        &'_ self,
+        &self,
         options: Option<KeyClientListDeletedKeyPropertiesOptions<'_>>,
     ) -> Result<Pager<ListDeletedKeyPropertiesResult>> {
         let options = options.unwrap_or_default().into_owned();
@@ -746,7 +746,7 @@ impl KeyClient {
     /// * `options` - Optional parameters for the request.
     #[tracing::function("KeyVault.getKeys")]
     pub fn list_key_properties(
-        &'_ self,
+        &self,
         options: Option<KeyClientListKeyPropertiesOptions<'_>>,
     ) -> Result<Pager<ListKeyPropertiesResult>> {
         let options = options.unwrap_or_default().into_owned();
@@ -821,7 +821,7 @@ impl KeyClient {
     /// * `options` - Optional parameters for the request.
     #[tracing::function("KeyVault.getKeyVersions")]
     pub fn list_key_properties_versions(
-        &'_ self,
+        &self,
         key_name: &str,
         options: Option<KeyClientListKeyPropertiesVersionsOptions<'_>>,
     ) -> Result<Pager<ListKeyPropertiesResult>> {
--- a/sdk/keyvault/azure_security_keyvault_secrets/src/generated/clients/secret_client.rs
+++ b/sdk/keyvault/azure_security_keyvault_secrets/src/generated/clients/secret_client.rs
@@ -296,7 +296,7 @@ impl SecretClient {
     /// * `options` - Optional parameters for the request.
     #[tracing::function("KeyVault.getDeletedSecrets")]
     pub fn list_deleted_secret_properties(
-        &'_ self,
+        &self,
         options: Option<SecretClientListDeletedSecretPropertiesOptions<'_>>,
     ) -> Result<Pager<ListDeletedSecretPropertiesResult>> {
         let options = options.unwrap_or_default().into_owned();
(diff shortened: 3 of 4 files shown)


sdk/core/azure_core_test/src/recording.rs
```diff
@@ -220,7 +220,7 @@ impl Recording {
     /// ```
     /// # let recording = azure_core_test::Recording::with_seed();
     /// let dek: [u8; 32] = recording.random();
-    /// # assert_eq!(typespec_client_core::base64::encode(dek), "HumPRAN6RqKWf0YhFV2CAFWu/8L/pwh0LRzeam5VlGo=");
+    /// # assert_eq!(azure_core::base64::encode(dek), "HumPRAN6RqKWf0YhFV2CAFWu/8L/pwh0LRzeam5VlGo=");
```
heaths, line 223: Don't know why - or maybe it was - this wasn't breaking other PRs. @jhendrixMSFT I believe it was cleaned up a bunch of the `typespec_client_core` refs we didn't want in anything but `azure_core`, which I agreed with and approved. But this should've broken then. Odd. Otherwise unrelated to the PR.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -375,16 +377,27 @@ impl<P: Page> ItemIterator<P> {
...
+        #[cfg(target_arch = "wasm32")] C: AsRef<str> + FromStr + 'static,
         #[cfg(target_arch = "wasm32")] F: Fn(PagerState<C>) -> Fut + 'static,
         #[cfg(target_arch = "wasm32")] Fut: Future<Output = crate::Result<PagerResult<P, C>>> + 'static,
     >(
         make_request: F,
-    ) -> Self {
-        Self::from_stream(iter_from_callback(make_request, || None, |_| {}))
+    ) -> Self
+    where
+        <C as FromStr>::Err: fmt::Debug,
```
LarryOsterman: Fascinating - I didn't know you could constrain types within generics. That's cool.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -408,11 +423,73 @@ impl<P: Page> ItemIterator<P> {
...
+    ///
+    /// while let Some(secret) = pager.try_next().await? {
+    ///     println!("{:?}", secret.id);
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn with_continuation_token(self, continuation_token: Option<String>) -> Self {
+        // Set the next_token because that's what is passed to iter_from_callback to get the initial page.
+        if let Ok(mut token) = self.next_token.lock() {
```
LarryOsterman, line 492: Nit: If the lock is poisoned, do you want this to fail or silently succeed?

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -408,11 +423,73 @@ impl<P: Page> ItemIterator<P> {
...
+        }
+
         PageIterator {
             stream: self.stream,
-            continuation_token: Default::default(),
+            continuation_token: self.next_token,
+        }
+    }
+
+    /// Advance the `ItemIterator` to the page referenced by `continuation_token`.
```
LarryOsterman: Nit: Is this really "advance"? Doesn't it return a page iterator to the start of the page referenced by `continuation_token`? I don't think they're quite the same thing.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -408,11 +423,73 @@ impl<P: Page> ItemIterator<P> {
...
+    ///
+    /// You should call this before iterating the [`Stream`] or results may be unpredictable.
```
LarryOsterman, line 457: Do you want to specifically say `ItemIterator` here rather than `Stream`? I'm not sure it really matters though.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -808,81 +909,83 @@ mod tests {
...
+                        "yon request didst fail",
+                    )),
+                    _ => {
+                        panic!("Unexpected continuation value")
+                    }
                 }
-            }
-        });
+            });
         let pages: Vec<Result<(String, Page), typespec::Error>> = pager
```
LarryOsterman: nit: `typespec::Error` instead of `crate::Error`?

sdk/core/azure_core/src/http/poller.rs
```diff
@@ -617,9 +617,21 @@ where
...
-                State::InProgress(n) => make_request(PollerState::More(n)).await,
-                State::Done => return None,
+                State::Init => {
+                    tracing::debug!("initial operation request");
+                    make_request(PollerState::Initial).await
+                }
+                State::InProgress(n) => {
+                    tracing::debug!(
+                        "subsequent operation request to {:?}",
+                        AsRef::<str>::as_ref(&n)
```
antkmsft, line 627: When the LRO patter is such that you're given a result in the first response, and all you do subsequently is poll for completion, `n` is not a URL, but a struct of URL and that first response: https://github.com/Azure/typespec-rust/blob/2b1e7cefd2f021591b8fd51761814290098454ba/packages/typespec-rust/test/spector/azure/core/lro/standard/src/generated/clients/standard_client.rs#L127-L134

How do we make it work - do we need to implement AsRef trait for `struct Progress`? Maybe `Debug` would be better? Or do I misunderstand it?

sdk/core/azure_core/src/http/pager.rs
heaths: No, I specifically mean the `ItemIterator` used as a `Stream`.

sdk/core/azure_core/src/http/poller.rs
heaths: I don't want `std::fmt::Debug`. For a `url::Url` the `Debug` representation is hideous. I already need an `AsRef<str>` for one of the iterators (I forget which offhand - I think the `PageIterator` needed it before) so I just decided to add it to both places (effectively required anyway since the stream implementation would be the same) and just use that. It actually means no additional constraints in practice, where a `Debug` constraint would be additional.

LarryOsterman:
> @LarryOsterman do we want to keep the `method_options: PagerOptions` field name? I'm tempted to because at least that's consistent, and if we're diligent in keeping that as a superset of `ClientMethodOptions` it should be a fairly consistent DX sans having to type a different name for struct initialization.

I like the idea of the field always being named `method_options`, even for pagers and pollers. It provides a level of consistency - even though the underlying type of the `method_options` may be different, the name of the options always remains the same because semantically they are the same.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -368,29 +378,31 @@ impl<P: Page> ItemIterator<P> {
...
     ///         Ok(PagerResult::from_response_header(resp, &HeaderName::from_static("x-next-continuation")))
     ///     }
-    /// });
+    /// }, None);
     /// ```
     pub fn from_callback<
         // This is a bit gnarly, but the only thing that differs between the WASM/non-WASM configs is the presence of Send bounds.
         #[cfg(not(target_arch = "wasm32"))] C: AsRef<str> + FromStr + Send + 'static,
-        #[cfg(not(target_arch = "wasm32"))] F: Fn(PagerState<C>) -> Fut + Send + 'static,
+        #[cfg(not(target_arch = "wasm32"))] F: Fn(PagerState<C>, Context<'static>) -> Fut + Send + 'static,
```
LarryOsterman, line 391: Interesting - I hadn't considered declaring the lifetime for the generic as `'static` in the poller, but that may make things simpler.

sdk/keyvault/azure_security_keyvault_certificates/src/generated/clients/certificate_client.rs
```diff
@@ -668,7 +668,7 @@ impl CertificateClient {
     /// * `options` - Optional parameters for the request.
     #[tracing::function("KeyVault.getCertificates")]
     pub fn list_certificate_properties(
-        &self,
+        &'_ self,
```
LarryOsterman: I'm surprised you needed to add the lifetime constraint on `self` here.

LarryOsterman (approved):

sdk/keyvault/azure_security_keyvault_certificates/src/generated/clients/certificate_client.rs
heaths: I was surprised too. It was actually Copilot when I told it to go update usage, but makes sense in hindsight. The client has to outlive the `Pager` because of some owned data (or maybe I should further look into exactly *what* to decouple, but this is probably a logical ownership anyway) so they need the same lifetime. We can let the compiler figure it out - hence `'_` - but the implicit `'1` and `'_` won't resolve to the same thing.

heaths: Well...that's odd. I got rid of the lifetime on `&self` and it works. I even remember seeing an error in the LSP, but I was still adapting to your struct so maybe something else was amiss. I'll remove these.

LarryOsterman (approved):

sdk/core/azure_core_test/src/recording.rs
LarryOsterman: The reason it didn't break PRs is that I don't believe our CI pipeline runs the azure_core_test tests (or maybe it doesn't run `azure_core_test` doctests since we don't formally publish those.

PR #3302, opened 2025-11-06
Authentication errors include HTTP responses

### LINKED ISSUE #3127: Finalize get_token error type

chlowell:
`get_token()` implementations consistently return an error of `ErrorKind::Credential` and when the error is motivated by an HTTP response, the caller can downcast the error to access that response. Closes #3127

### DIFF
--- a/sdk/identity/azure_identity/CHANGELOG.md
+++ b/sdk/identity/azure_identity/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features Added
 
-- A `get_token()` error caused by an HTTP response carries that response. See the [troubleshooting guide](http://aka.ms/azsdk/rust/identity/troubleshoot#find-relevant-information-in-errors) for example code showing how to access the response.
+- A `get_token()` error caused by an HTTP response carries that response. See the [troubleshooting guide](https://aka.ms/azsdk/rust/identity/troubleshoot#find-relevant-information-in-errors) for example code showing how to access the response.
 
 ### Breaking Changes
 


sdk/identity/azure_identity/src/azure_pipelines_credential.rs
```diff
@@ -163,16 +163,20 @@ impl ClientAssertion for Client {
...
+        let status = resp.status();
+        if status != StatusCode::Ok {
             let err_headers: ErrorHeaders = resp.headers().get()?;
-
-            return Err(
-                azure_core::Error::with_message(
-                    ErrorKind::HttpResponse { status: status_code, error_code: Some(status_code.canonical_reason().to_string()), raw_response: None },
-                     format!("{status_code} response from the OIDC endpoint. Check service connection ID and pipeline configuration. {err_headers}"),
-                )
+            let message = format!(
```
LarryOsterman: I'm assuming this is because `status` and `err_headers` are references to `resp` and thus you can't call `format!` in the `with_message` call, right?

I had to spend a couple of seconds to figure that out so it might be worth a comment.

sdk/identity/azure_identity/src/lib.rs
```diff
@@ -60,6 +62,7 @@ use std::borrow::Cow;
 #[derive(Debug, Default, Deserialize)]
 #[serde(default)]
 struct EntraIdErrorResponse {
+    error_codes: Vec<i32>,
```
heaths, line 65: For the `error_description`, mainly, you might rely on borrowing for less memory overhead if you're just trying to pick out a few things. We do that for `ErrorResponseInternal` and you can find more details at <https://serde.rs/lifetimes.html#borrowing-data-in-a-derived-impl>.

sdk/identity/azure_identity/src/lib.rs
```diff
@@ -134,30 +161,48 @@ fn get_authority_host(env: Option<Env>, cloud: Option<&CloudConfiguration>) -> R
...
+        stringify!(ManagedIdentityCredential) => (false, "#managed-id"),
+        stringify!(WorkloadIdentityCredential) => (true, "#workload"),
+        _ => (false, ""),
+    };
+
+    let mut details = err.to_string();
+    if wrap_inner_error {
+        // details is e.g. "ClientAssertionCredential authentication failed. <more details>".
+        // We want only the "<more details>" part because we want to return an error like
+        // "OtherCredential authentication failed. <more details>".
```
heaths: Not blocking, but I'd urge a cheaper way in the future. If you could do something in `err` to tell it came from an internal implementation detail or even pass something down to `ClientAssertionCredential` to tell it which name to use, you could avoid all this fragile and (slightly) time-wasting string manipulation.

sdk/identity/azure_identity/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+- A `get_token()` error motivated by an HTTP response carries that response. See the [troubleshooting guide](https://aka.ms/azsdk/rust/identity/troubleshoot#find-relevant-information-in-errors) for example code showing how to access the response.
```
heaths: "motivated". Perhaps "caused"? Even if that's an arcane but valid use - which I normally appreciate - I imagine users might be scratching their heads as well.

sdk/identity/azure_identity/TROUBLESHOOTING.md
```diff
@@ -42,6 +42,39 @@ This error contains several pieces of information:
 
 - __Correlation ID and Timestamp__: The correlation ID and timestamp identify the request in server-side logs. This information can be useful to support engineers diagnosing unexpected Microsoft Entra ID failures.
 
+Many credential errors also carry the HTTP response that motivated them. The example below demonstrates how to access that response in such a case.
```
heaths:
```suggestion
Many credential errors also carry the HTTP response that caused them. The example below demonstrates how to access that response in such a case.
```

heaths: But they still get the message you want by default, right? This is just to get more details for...why? You might put that in the comments and even explain in what scenarios they may want to drill into details. Ideally, the message you formatted above is what they see if they just print the error e.g., `eprintln!("{err}")`.

sdk/identity/azure_identity/TROUBLESHOOTING.md
```diff
@@ -42,6 +42,39 @@ This error contains several pieces of information:
...
+                    let headers = response.headers();
+                    let body = String::from_utf8_lossy(response.body());
+                    eprintln!("status: {status}");
+                    eprintln!("headers: {headers:?}");
+                    eprintln!("body: {body}");
+                }
+            }
+        }
+        _ => {
+            // TODO: handle other kinds of error
```
heaths: Tip: idiomatic thing here is `todo!("handle other kinds of errors")`.

heaths (approved):
Nothing blocking, but there's a couple things I'd clean up now and things to consider for later.

sdk/identity/azure_identity/TROUBLESHOOTING.md
chlowell: Yes; printing the error produces a message like the above which probably contains everything developers need. The raw response is useful mainly because it includes headers. I'll refocus this example on that to avoid implying the response is more generally useful. Really I just want to publish this snippet so that the rare developer who wants the response has some guidance toward getting it

heaths: I think the snippet is fine. I just wanted to make it clear that they probably don't *need* to, and maybe telling them for what reasons they might do so, like getting headers, like you said. You might just keep the code as-is, but explain a reason they might want to do this i.e., it shouldn't be the norm.

sdk/identity/azure_identity/src/lib.rs
chlowell: #3322

PR #3303, opened 2025-11-07
[Storage] `set/get_access_policy` for `ContainerClient`

vincenttran-msft:
- [x] `.tsp 2`: https://github.com/Azure/azure-rest-api-specs/pull/38933
 - [x] `.tsp`: https://github.com/Azure/azure-rest-api-specs/pull/38633
- Adds implementations for `set_access_policy()` and `get_access_policy()`
- Adds `From` impl for `HashMap<String, AccessPolicy>` -> `SignedIdentifiers`
- Fixes up BlobTags `TryFrom` to `From` for consistency, especially since fallible case was not actually fallible

SetAccessPolicy: **Completely NULLED**
GetAccessPolicy:
>/// * [`date`()](crate::generated::models::VecSignedIdentifierHeaders::date) - Date **REMOVED**
    /// * [`last_modified`()](crate::generated::models::VecSignedIdentifierHeaders::last_modified) - Last-Modified
    /// * [`etag`()](crate::generated::models::VecSignedIdentifierHeaders::etag) - etag
    /// * [`access`()](crate::generated::models::VecSignedIdentifierHeaders::access) - x-ms-blob-public-access

- [x] Changelog entry

### DIFF
--- a/sdk/storage/azure_storage_blob/tests/blob_container_client.rs
+++ b/sdk/storage/azure_storage_blob/tests/blob_container_client.rs
@@ -430,8 +430,6 @@ async fn test_container_access_policy(ctx: TestContext) -> Result<(), Box<dyn Er
             ..Default::default()
         }),
     );
-    println!("Expiry: {}", &expiry);
-    println!("Start: {}", &start);
     let test_id_1: Option<String> = Some("testid_1".into());
     let test_id_2: Option<String> = Some("testid_2".into());
     let access_policy_1 = AccessPolicy {
--- a/sdk/storage/azure_storage_blob/tsp-location.yaml
+++ b/sdk/storage/azure_storage_blob/tsp-location.yaml
@@ -1,4 +1,4 @@
 directory: specification/storage/Microsoft.BlobStorage
 commit: 76ab9e72c8747dc01f2f406caa796d03160b813d
 repo: Azure/azure-rest-api-specs
-additionalDirectories:
+additionalDirectories: 
--- a/sdk/storage/azure_storage_blob/assets.json
+++ b/sdk/storage/azure_storage_blob/assets.json
@@ -1,6 +1,6 @@
 {
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "rust",
-  "Tag": "rust/azure_storage_blob_ceab74b7c5",
+  "Tag": "rust/azure_storage_blob_8a70471664",
   "TagPrefix": "rust/azure_storage_blob"
 }


sdk/storage/azure_storage_blob/tests/blob_container_client.rs
```diff
@@ -400,3 +402,111 @@ async fn test_find_blobs_by_tags_container(ctx: TestContext) -> Result<(), Box<d
...
+    let response_access_policy = signed_identifiers.items.clone().unwrap()[0]
+        .access_policy
+        .clone();
+    assert_eq!(signed_identifiers.items.clone().unwrap().len(), 1);
+    assert_eq!(response_id, test_id);
+    assert_eq!(
+        response_access_policy.clone().unwrap().permission,
+        access_policy.permission
+    );
+    assert_eq!(response_access_policy.clone().unwrap().expiry, expiry);
```
vincenttran-msft: [log](https://[dev.azure.com/azure-sdk/29ec6040-b234-4e31-b139-33dc4287b756/_apis/build/builds/5545811/logs/245](https://dev.azure.com/azure-sdk/29ec6040-b234-4e31-b139-33dc4287b756/_apis/build/builds/5545811/logs/245))
Relevant log excerpt:
>2025-11-07T02:41:40.3347517Z ---- test_container_access_policy stdout ----
2025-11-07T02:41:40.3347746Z 
2025-11-07T02:41:40.3348185Z thread 'test_container_access_policy' (20747) panicked at sdk/storage/azure_storage_blob/tests/blob_container_client.rs:449:5:
2025-11-07T02:41:40.3348696Z assertion `left == right` failed
2025-11-07T02:41:40.3349477Z   left: Some("2025-11-07T02:09:40.3229246Z")
2025-11-07T02:41:40.3353978Z  right: Some("2025-11-07T02:41:35.2937819Z")

@heaths We are running into an issue when asserting against dynamically generated `datetime` objects, as here the LHS is from the [recording](https://github.com/Azure/azure-sdk-assets/blob/dae663daf31499c2bce51fa19739a4eb3700a4e0/rust/sdk/storage/azure_storage_blob/tests/data/blob_container_client/test_container_access_policy.json#L46): `<Expiry>2025-11-07T02:09:40.3229246Z</Expiry>`
while the RHS is from the actual test execution.

In [Python](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/tests/test_container.py#L614), this is resolved through using a `get_datetime_variable` function:
`expiry_time = self.get_datetime_variable(variables, 'expiry_time', datetime.utcnow() + timedelta(hours=1))`
And this way the datetimes are present in [variables](https://github.com/Azure/azure-sdk-assets/blob/python/storage/azure-storage-blob_65571078b7/python/sdk/storage/azure-storage-blob/tests/recordings/test_container.pyTestStorageContainertest_set_container_acl.json#L91):
>"Variables": {
    "expiry_time": "2022-08-09T19:54:43.709865",


 Is this something that can be done in Rust currently, or is this a net new feature for Rust's test infrastructure? I see that we have `var()` and `var_opt()` but both of these are for getting and not setting.

sdk/storage/azure_storage_blob/src/parsers.rs
```diff
@@ -59,3 +65,43 @@ pub fn format_filter_expression(tags: &HashMap<String, String>) -> Result<String
...
+        )
+    })
+}
+
+/// Takes a HashMap<String, AccessPolicy> where the key is the policy ID and the value is the AccessPolicy and converts it to SignedIdentifiers.
+///
+/// # Arguments
+///
+/// * `policies` - A HashMap where keys are policy identifiers and values are AccessPolicy objects.
+pub fn format_signed_identifiers(
```
jalauzon-msft: Change this to a `From` impl for `HashMap<String, AccessPolicy>`

sdk/storage/azure_storage_blob/src/parsers.rs
```diff
@@ -59,3 +65,43 @@ pub fn format_filter_expression(tags: &HashMap<String, String>) -> Result<String
 
     Ok(format_expression.join(" and "))
 }
+
+/// Takes a OffsetDateTime and converts to RFC3339-like string with exactly 7 decimal precision and converted to UTC.
+///
+/// # Arguments
+///
+/// * `datetime` - OffsetDateTime to format.
+pub fn format_datetime(datetime: OffsetDateTime) -> Result<String, Error> {
```
jalauzon-msft: Change name to `format_storage_datetime`

sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -365,4 +367,32 @@ impl BlobContainerClient {
...
         }
     }
+
+    /// Sets the permissions for the specified container. The permissions indicate whether blobs in a
+    /// container may be accessed publicly.
+    ///
+    /// # Arguments
+    ///
+    /// * `container_acl` - The access control list for the container. See [`format_signed_identifiers()`](crate::format_signed_identifiers) for help with the expected format.
+    /// * `options` - Optional configuration for the request.
```
jalauzon-msft, line 377: Add usage sample in doc

sdk/storage/azure_storage_blob/tests/blob_container_client.rs
```diff
@@ -400,3 +402,143 @@ async fn test_find_blobs_by_tags_container(ctx: TestContext) -> Result<(), Box<d
...
+        response_access_policy.clone().unwrap().permission,
+        access_policy.permission
+    );
+    assert_eq!(
+        response_access_policy.clone().unwrap().expiry,
+        Some(expiry_1)
+    );
+    assert_eq!(response_access_policy.clone().unwrap().start, Some(start_1));
+
+    // Set Access Policy w/ Multiple Policy Defined
```
jalauzon-msft, line 416: Just do multiple

sdk/storage/azure_storage_blob/src/parsers.rs
```diff
@@ -59,3 +65,43 @@ pub fn format_filter_expression(tags: &HashMap<String, String>) -> Result<String
...
+}
+
+/// Takes a HashMap<String, AccessPolicy> where the key is the policy ID and the value is the AccessPolicy and converts it to SignedIdentifiers.
+///
+/// # Arguments
+///
+/// * `policies` - A HashMap where keys are policy identifiers and values are AccessPolicy objects.
+pub fn format_signed_identifiers(
+    policies: HashMap<String, AccessPolicy>,
+) -> Result<SignedIdentifiers, Error> {
```
jalauzon-msft: Remove result?

sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -365,4 +366,60 @@ impl BlobContainerClient {
...
+    /// ```rust,no_run
+    /// use std::collections::HashMap;
+    /// use azure_core::{http::RequestContent, Result};
+    /// use azure_storage_blob::{BlobContainerClient, models::{AccessPolicy, SignedIdentifiers}};
+    ///
+    /// async fn example() -> Result<()> {
+    ///     let container_client = BlobContainerClient::new(
+    ///         "https://storageaccount.blob.core.windows.net/",
+    ///         "some_container_name",
+    ///         None,
```
jalauzon-msft: Should probably have a credential in our sample.

sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -365,4 +366,60 @@ impl BlobContainerClient {
...
+    ///         None,
+    ///     )?;
+    ///
+    ///     let mut policies = HashMap::new();
+    ///     policies.insert("some_policy_id".to_string(), AccessPolicy {
+    ///         start: Some("2035-01-01T00:00:00Z".to_string()),
+    ///         expiry: Some("2035-12-31T00:00:00Z".to_string()),
```
jalauzon-msft: You didn't use our string helper?

sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -365,4 +366,60 @@ impl BlobContainerClient {
...
+    ///         permission: Some("rwd".to_string()),
+    ///     });
+    ///
+    ///     container_client.set_access_policy(RequestContent::try_from(SignedIdentifiers::from(policies))?, None).await?;
+    ///     Ok(())
```
jalauzon-msft: Seeing this full sample makes me think, is there a way to just have a snippet instead? Like basically just from line 394-401?

sdk/storage/azure_storage_blob/src/parsers.rs
jalauzon-msft: You didn't rename this?

sdk/storage/azure_storage_blob/src/models/extensions.rs
```diff
@@ -67,29 +68,18 @@ impl BlockBlobClientUploadOptions<'_> {
 }
 
 /// Converts a `BlobTags` struct into `HashMap<String, String>`.
-impl TryFrom<BlobTags> for HashMap<String, String> {
-    type Error = azure_core::Error;
-
-    fn try_from(blob_tags: BlobTags) -> Result<Self, azure_core::Error> {
+impl From<BlobTags> for HashMap<String, String> {
```
jalauzon-msft, line 71: Need to mark this down as breaking change? Mostly for the change from `try_into` to `into`.

sdk/storage/azure_storage_blob/src/parsers.rs
vincenttran-msft: Forgot to redo because I had it changed (but uncommitted) before merging `main`.

sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
vincenttran-msft: So the crux of the issue is that our pipelines enforce `cargo test --doc` passes, and this checks that it at least compiles. 
I just put another draft up where the sample focuses on the relevant parts (basically L394-401 as you state). But it's still not perfect.

sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -365,4 +366,58 @@ impl BlobContainerClient {
...
+    ///         start: Some(format_storage_datetime(OffsetDateTime::now_utc())?),
+    ///         expiry: Some(format_storage_datetime(OffsetDateTime::now_utc() + Duration::from_secs(10))?),
+    ///         permission: Some("rwd".to_string()),
+    ///     });
+    ///
+    ///     let _request_content = RequestContent::try_from(SignedIdentifiers::from(policies))?;
+    ///
+    ///     // container_client.set_access_policy(request_content, None).await?; // RequestContent is ready to be used here
+    ///
+    ///     Ok(())
```
vincenttran-msft: Here is the current draft. We can cut down on the `ContainerClient` initialization by just commenting out the actual API call to `set_access_policy`. But I don't love that we have to append `_` to request_content to avoid the unused compiler warning 🙄

vincenttran-msft: Changed to more pointed snippet and added ignore as discussed offline.

sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -365,4 +366,51 @@ impl BlobContainerClient {
...
+    /// let mut policies = HashMap::new();
+    /// policies.insert("some_policy_id".to_string(), AccessPolicy {
+    ///     start: Some(format_storage_datetime(OffsetDateTime::now_utc())?),
+    ///     expiry: Some(format_storage_datetime(OffsetDateTime::now_utc() + Duration::from_secs(10))?),
+    ///     permission: Some("rwd".to_string()),
+    /// });
+    ///
+    /// let request_content = RequestContent::try_from(SignedIdentifiers::from(policies))?;
+    /// container_client.set_access_policy(request_content, None).await?;
+    /// ```
```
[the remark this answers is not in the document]
vincenttran-msft: Intentional, for now we are thinking about just surfacing the imports of interest.

sdk/storage/azure_storage_blob/tests/blob_container_client.rs
```diff
@@ -400,3 +402,94 @@ async fn test_find_blobs_by_tags_container(ctx: TestContext) -> Result<(), Box<d
...
+    };
+    let access_policy_2 = AccessPolicy {
+        expiry: Some(expiry),
+        permission: Some("cd".to_string()),
+        start: Some(start),
+    };
+    let policies: HashMap<String, AccessPolicy> = HashMap::from([
+        (test_id_1.clone().unwrap(), access_policy_1.clone()),
+        (test_id_2.clone().unwrap(), access_policy_2.clone()),
+    ]);
```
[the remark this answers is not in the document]
vincenttran-msft: We use it later against the response which is a Option

sdk/storage/azure_storage_blob/Cargo.toml
```diff
@@ -21,6 +21,7 @@ async-trait.workspace = true
 azure_core = { workspace = true, features = ["xml"] }
 serde.workspace = true
 serde_json.workspace = true
+time.workspace = true
 typespec_client_core = { workspace = true, features = ["derive"] }
```
heaths: This should be removed. Only depend on `azure_core`. This was necessary at one point but was fixed months ago.

sdk/storage/azure_storage_blob/src/parsers.rs
```diff
@@ -59,3 +63,18 @@ pub fn format_filter_expression(tags: &HashMap<String, String>) -> Result<String
 
     Ok(format_expression.join(" and "))
 }
+
+/// Takes a OffsetDateTime and converts to RFC3339-like string with exactly 7 decimal precision and converted to UTC.
+///
+/// # Arguments
+///
+/// * `datetime` - OffsetDateTime to format.
+pub fn format_storage_datetime(datetime: OffsetDateTime) -> Result<String, Error> {
```
heaths: Why is this public? If public to the crate, fine, but apparently you don't use it because I don't see it imported into `src/lib.rs` where it probably shouldn't be public. I would think you'd handle this yourself and not make every caller do it.

vincenttran-msft: In `src/lib.rs` L16 is:
>pub use parsers::*;
Which re-exports all the public items from the `parsers` module.

The reason why we have the user doing this is because this is our interim solution for the datetime precision issue: https://github.com/Azure/typespec-rust/issues/618

`set_access_policy` takes a `RequestContent<SignedIdentifiers, XmlFormat>`

The issues lies in `AccessPolicy`, but to get there you have:

- RequestContent<SignedIdentifiers, XmlFormat>
- SignedIdentifiers which holds Vec<`SignedIdentifier`>
- SignedIdentifier which holds `AccessPolicy`

So we didn't really see a way to specify that this parser helper could be called on their behalf.

This helper goes away entirely and the issue is resolved at the code-generation level once we resolve the above GitHub issue.

sdk/storage/azure_storage_blob/Cargo.toml
vincenttran-msft: This is needed for the helper implementation, which uses:
>format_description::FormatItem, macros::format_description

both of which are not accessible from the typespec_client_core time re-export.

`typespec_client_core/src/time/mod.rs`
```rust
//! Date and time parsing and formatting functions.

pub use time::{error::ComponentRange, Duration, OffsetDateTime};

// Used internally, not public re-exports
use time::{ //
    format_description::{well_known::Rfc3339, FormatItem},
    macros::format_description,
    PrimitiveDateTime,
};
use typespec::error::{ErrorKind, ResultExt};
```

But again, if this is the only instance where we've (as in consumers of Core) need to use these, I wouldn't go and change `typespec_client_core` to make these public re-exports since the helper (and thus this dependency) will go away in the long-term once https://github.com/Azure/typespec-rust/issues/618 is resolved.

heaths: My comment was on the typespec_client_core import.

vincenttran-msft: Doi my mistake, got blinded by the diff green highlight 🤦

sdk/storage/azure_storage_blob/tsp-location.yaml
```diff
@@ -1,4 +1,4 @@
 directory: specification/storage/Microsoft.BlobStorage
-commit: 6b5a0fdadca03abe0b8e91a68a62cd1e639ed55e
+commit: 1c722de0ea8cd9168e9e605c4a206cb18b8b83ca
```
vincenttran-msft: This commit ID matches the most recent commit from: https://github.com/Azure/azure-rest-api-specs/pull/38933

Image

Yet it seems like it didn't take the change to `@encode("rfc3339-fixed-width")`.

Also, Intellisense seems to not think that it is a valid option on the TypeSpec def side (if that matters):

Image

sdk/storage/azure_storage_blob/Cargo.toml
```diff
@@ -21,7 +21,7 @@ async-trait.workspace = true
 azure_core = { workspace = true, features = ["xml"] }
 serde.workspace = true
 serde_json.workspace = true
-typespec_client_core = { workspace = true, features = ["derive"] }
+time.workspace = true
```
vincenttran-msft, line 24: FYI, this is only needed for [this ](https://github.com/vincenttran-msft/azure-sdk-for-rust/blob/d2f0a0faa4a208b01f67ce7460d7f8d2086fbb1e/sdk/storage/azure_storage_blob/src/generated/models/models_serde.rs#L184)import in generated code:
`use time::format_description::well_known::{iso8601, Iso8601};`

Given that we usually just use the re-exports from `azure_core`, would these be something you would consider re-exporting so we don't have to have a `time` crate dependency?

PR is fine to merge as-is, but we should track the above if we want to go ahead with it to coordinate a fix post-merge.

heaths: You have a `time` crate implicit dependency anyway. We do it elsewhere so I'm not concerned about here. We don't necessarily want to re-export everything - just the common stuff. `time` is one of those crates that really should be in `std ` - would be in any other language - so I don't anticipate we'll ever get rid of it.

PR #3306, opened 2025-11-07
Implement retry logic for IMDS managed identity

### LINKED ISSUE #2729: ManagedIdentityCredential retry policy should handle 410 status response
DefaultAzureCredentialImdsRetryPolicy should delay for long enough so that the exponential retry total duration is at least 70 seconds in the event that the response code is 410.

[See the docs](https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=windows#errors-and-debugging) for more info on 410 status code

chlowell:
Closes #2729 by aligning the retry policy configuration for our IMDS client with the [published guidance](https://learn.microsoft.com/entra/identity/managed-identities-azure-resources/how-to-use-vm-token#error-handling). Our timing is close but doesn't exactly match the recommended curve:

| attempt | occurs at (s)|recommended|
|---|---|---|
| 1 | 0|0
| 2 | 2|2
| 3 | 8|8
| 4 | 19|22
| 5 | 40|52
|6|70|-

### DIFF
--- a/sdk/identity/azure_identity/src/virtual_machine_managed_identity_credential.rs
+++ b/sdk/identity/azure_identity/src/virtual_machine_managed_identity_credential.rs
@@ -26,7 +26,7 @@ pub struct VirtualMachineManagedIdentityCredential {
 impl VirtualMachineManagedIdentityCredential {
     pub fn new(
         id: ImdsId,
-        mut client_options: ClientOptions,
+        client_options: ClientOptions,
         env: Env,
     ) -> azure_core::Result<Arc<Self>> {
         let endpoint = Url::parse(ENDPOINT).unwrap(); // valid url constant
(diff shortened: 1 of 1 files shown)


sdk/identity/azure_identity/src/virtual_machine_managed_identity_credential.rs
```diff
@@ -23,10 +26,38 @@ pub struct VirtualMachineManagedIdentityCredential {
 impl VirtualMachineManagedIdentityCredential {
     pub fn new(
         id: ImdsId,
-        client_options: ClientOptions,
+        mut client_options: ClientOptions,
```
heaths: This is all internal, right? Normally, you just clone it but as long as this isn't viral to public clients, this is fine. We shouldn't be modifying what they pass in if they specified any values. It's idiomatic in other cases like so (example, may not be 100% correct since I don't remember exact field types):

```rust
let options = ClientOptions {
  retry = RetryOptions::exponential(..),
  ..options
};
```

Then, even if you do necessarily override, this is the idiomatic way. Requiring `mut` may be prohibitive.

heaths (changes requested):

sdk/identity/azure_identity/src/virtual_machine_managed_identity_credential.rs
chlowell: This function is internal and the public `ManagedIdentityCredential::new()` already owns this argument. That's a good suggestion though, it will be better if future readers don't have to think about this

heaths (approved):

PR #3311, opened 2025-11-08
Fixed issues validating tracing pages which were converted to page iterators

LarryOsterman:
- Re-add the span when converting from `Init` state on a continuation token;
- Cleaned up lifetime of `MockSpan` to have it be dropped when the actual span goes out of scope (also verified that the same "end on drop" behavior occurs with an OpenTelemetry `Span`).
- Added a `Debug` requirement for the `C` parameter so it can be printed successfully.
- Added a simple test for verifying the spans coming from an `ItemIterator` converted immediately to a `PageIterator`.

Created as Draft so it doesn't conflict with releases of various non-core crates.

### DIFF
--- a/sdk/core/azure_core/src/http/pager.rs
+++ b/sdk/core/azure_core/src/http/pager.rs
@@ -827,12 +827,9 @@ where
 {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            State::Init => write!(f, "State::Init"),
-            State::More(c) => f
-                .debug_struct("State::More")
-                .field("continuation", &c.as_ref())
-                .finish(),
-            State::Done => write!(f, "State::Done"),
+            State::Init => write!(f, "Init"),
+            State::More(c) => f.debug_tuple("More").field(&c.as_ref()).finish(),
+            State::Done => write!(f, "Done"),
         }
     }
 }
--- a/sdk/core/azure_core_test/src/tracing.rs
+++ b/sdk/core/azure_core_test/src/tracing.rs
@@ -317,12 +317,15 @@ pub fn check_instrumentation_result(
 
         let spans = tracer.spans.lock().unwrap();
 
-        // assert_eq!(
-        //     spans.len(),
-        //     expected.spans.len(),
-        //     "Unexpected number of spans for tracer {}",
-        //     expected.name
-        // );
+        // Check span lengths if there are no wildcard spans.
+        if !expected.spans.iter().any(|s| s.is_wildcard) {
+            assert_eq!(
+                spans.len(),
+                expected.spans.len(),
+                "Unexpected number of spans for tracer {}",
+                expected.name
+            );
+        }
 
         let mut expected_index = 0;
         for (span_index, span_actual) in spans.iter().enumerate() {
@@ -352,15 +355,25 @@ pub fn check_instrumentation_result(
                     ) {
                         eprintln!(
                             "Next actual span does not match expected span: {}",
-                            expected.spans[expected_index + 1].span_name
+                            expected.spans[expected_index].span_name
                         );
                         expected_index += 1;
                     }
+                } else {
+                    // At the very end, bump the expected index past the wildcard entry.
+                    // This ensures that we consume all the expected spans.
+                    expected_index += 1;
                 }
             } else {
                 expected_index += 1;
             }
         }
+        assert_eq!(
+            expected_index,
+            expected.spans.len(),
+            "Not all expected spans were found for tracer {}",
+            expected.name
+        );
     }
 }
 


sdk/keyvault/azure_security_keyvault_secrets/tests/secret_client.rs
```diff
@@ -267,7 +267,10 @@ async fn round_trip_secret_verify_telemetry(ctx: TestContext) -> Result<()> {
             })
         },
         ExpectedInstrumentation {
-            package_name: recording.var("CARGO_PKG_NAME", None),
+            // Don't use `recording.var` here in case the recording was made with a different package version.
+            package_name: env!("CARGO_PKG_NAME").into(),
```
heaths: This could stay as-is. At the very least, your comment and expression don't align.

sdk/core/azure_core_test/src/tracing.rs
```diff
@@ -252,21 +316,50 @@ pub fn check_instrumentation_result(
...
-            expected.name
-        );
 
-        for (span_index, span_expected) in expected.spans.iter().enumerate() {
+        // assert_eq!(
+        //     spans.len(),
+        //     expected.spans.len(),
+        //     "Unexpected number of spans for tracer {}",
+        //     expected.name
+        // );
```
heaths: Nit

sdk/core/azure_core_test/src/tracing.rs
```diff
@@ -252,21 +316,50 @@ pub fn check_instrumentation_result(
...
+
+        let mut expected_index = 0;
+        for (span_index, span_actual) in spans.iter().enumerate() {
             eprintln!(
                 "Checking span {} of tracer {}: {}",
-                span_index, expected.name, span_expected.span_name
+                span_index, expected.name, span_actual.name
```
heaths, line 334: Was this just wrong before and this fixed that? It's an unexpected change otherwise.

sdk/core/azure_core_test/src/tracing.rs
```diff
@@ -252,21 +316,50 @@ pub fn check_instrumentation_result(
...
+                &expected.spans[expected_index],
+                &parent_span_map,
             );
-            check_span_information(&spans[span_index], span_expected, &parent_span_map);
             // Now that we've verified the span, add the mapping between expected span ID and the actual span ID.
-            parent_span_map.insert(span_expected.span_id, spans[span_index].id);
+            parent_span_map.insert(expected.spans[expected_index].span_id, span_actual.id);
+            if expected.spans[expected_index].is_wildcard {
+                // If this is a wildcard span, we don't increment the expected index.
+                eprintln!(
```
heaths, line 345: This is going to be noisy. Either remove or trace.

sdk/core/azure_core_test/src/tracing.rs
```diff
@@ -252,21 +316,50 @@ pub fn check_instrumentation_result(
...
+                    let next_span = &spans[span_index + 1];
+                    if !compare_span_information(
+                        next_span,
+                        &expected.spans[expected_index],
+                        &parent_span_map,
+                    ) {
+                        eprintln!(
+                            "Next actual span does not match expected span: {}",
+                            expected.spans[expected_index + 1].span_name
+                        );
```
heaths, line 359: Same here. Or are you intending that this will only show up if the test fails and would be useful? Best to double check that behavior. I don't want to see all this in normal `cargo test` runs.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -811,14 +811,36 @@ impl<P> fmt::Debug for PageIterator<P> {
...
+where
+    C: AsRef<str>,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            State::Init => write!(f, "State::Init"),
+            State::More(c) => f
+                .debug_struct("State::More")
+                .field("continuation", &c.as_ref())
+                .finish(),
```
heaths: `Formatter::debug_tuple`

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -811,14 +811,36 @@ impl<P> fmt::Debug for PageIterator<P> {
...
 }
 
-impl<C> PartialEq for State<C> {
+impl<C> fmt::Debug for State<C>
+where
+    C: AsRef<str>,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            State::Init => write!(f, "State::Init"),
```
heaths: Nit: default formatting only uses the variant name if you want to match e.g., "Init" not "State::Init".

heaths (changes requested):

sdk/core/azure_core_test/src/tracing.rs
LarryOsterman: The trace that was generated was just wrong.

sdk/core/azure_core_test/src/tracing.rs
LarryOsterman: The only time these will ever be seen is if the test fails (at which point, this is a helpful diagnostic because it helps to understand how the two independent spans are progressing forward) or if the customer explicitly specifies `--show-output`, in which case they want to see it. 

Since it will never be seen, why do we care about it? 

In general, my logic is that I use `eprintln` or `println` for tests, and `trace!` for production code. That's because `eprintln` is output when tests fail, but `trace!` is basically never output.

sdk/core/azure_core_test/src/tracing.rs
LarryOsterman: That is correct. And here's the output when you don't have `--show-output` set:

```bash
running 4 tests
test round_trip_secret_verify_telemetry ... ok
test list_secrets_verify_telemetry ... ok
test list_secrets_verify_telemetry_rehydrated ... ok
test list_secrets_by_pages_verify_telemetry ... ok
```

sdk/core/azure_core_test/src/tracing.rs
LarryOsterman: Thank you - I forgot to put that one back.

heaths (approved):

Issue #1907, opened 2024-11-09
Provider helpers to easily overwrite `Url` query parameters


heaths:
According to the [REST API guidelines](https://aka.ms/azapi/guidelines), clients should set or overwrite the `api-version` header for any request to the service including [pageables](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#collections-nextlink-includes-all-query-params) and [LROs](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#lro-operation-location-includes-api-version); however, `url::Url` doesn't define an easy way of overwriting headers. We have [helpers for appending](https://github.com/Azure/azure-sdk-for-rust/blob/d7137b088fffa3bdb65a068b3a7a4c00e8dea6a5/sdk/typespec/typespec_client_core/src/http/mod.rs#L32) and should add some for setting/overwriting because the code to do so - which we need to do for every request to the service - is a bit verbose.

See [this comment](https://github.com/Azure/typespec-rust/pull/147#discussion_r1835160520) for context.

PR #3331, opened 2025-11-11
Add set_query_pair method to UrlExt trait

jhendrixMSFT:
A convenience method for setting query parameter key/value pairs that will overwrite an existing key with the new value.

Fixes https://github.com/Azure/azure-sdk-for-rust/issues/1907

### DIFF
--- a/sdk/core/typespec_client_core/src/http/mod.rs
+++ b/sdk/core/typespec_client_core/src/http/mod.rs
@@ -91,10 +91,11 @@ pub trait UrlExt: crate::private::Sealed {
     /// use typespec_client_core::http::{Url, UrlExt as _};
     ///
     /// let mut url: Url = "https://contoso.com?a=1&b=2".parse().unwrap();
-    /// url.query_builder()
+    /// let mut query_builder = url.query_builder();
+    /// query_builder
     ///     .set_pair("a", "new_value")
-    ///     .set_pair("c", "3")
-    ///     .build();
+    ///     .set_pair("c", "3");
+    /// query_builder.build();
     /// let params: Vec<_> = url.query_pairs().collect();
     /// assert!(params.contains(&("a".into(), "new_value".into())));
     /// assert!(params.contains(&("b".into(), "2".into())));
@@ -176,15 +177,16 @@ impl<'a> QueryBuilder<'a> {
     /// use typespec_client_core::http::{Url, UrlExt as _};
     ///
     /// let mut url: Url = "https://contoso.com".parse().unwrap();
-    /// url.query_builder()
+    /// let mut query_builder = url.query_builder();
+    /// query_builder
     ///     .append_key_only("debug")
-    ///     .append_pair("a", "1")
-    ///     .build();
+    ///     .append_pair("a", "1");
+    /// query_builder.build();
     /// let params: Vec<_> = url.query_pairs().collect();
     /// assert!(params.contains(&("debug".into(), "".into())));
     /// assert!(params.contains(&("a".into(), "1".into())));
     /// ```
-    pub fn append_key_only(mut self, key: impl Into<Cow<'a, str>>) -> Self {
+    pub fn append_key_only(&mut self, key: impl Into<Cow<'a, str>>) -> &mut Self {
         let key = key.into();
 
         if let Some(vals) = self.values.get_mut(&key) {
@@ -214,20 +216,21 @@ impl<'a> QueryBuilder<'a> {
     /// use typespec_client_core::http::{Url, UrlExt as _};
     ///
     /// let mut url: Url = "https://contoso.com?a=1".parse().unwrap();
-    /// url.query_builder()
+    /// let mut query_builder = url.query_builder();
+    /// query_builder
     ///     .append_pair("a", "2")
-    ///     .append_pair("b", "3")
-    ///     .build();
+    ///     .append_pair("b", "3");
+    /// query_builder.build();
     /// let params: Vec<_> = url.query_pairs().collect();
     /// assert!(params.contains(&("a".into(), "1".into())));
     /// assert!(params.contains(&("a".into(), "2".into())));
     /// assert!(params.contains(&("b".into(), "3".into())));
     /// ```
     pub fn append_pair(
-        mut self,
+        &mut self,
         key: impl Into<Cow<'a, str>>,
         value: impl Into<Cow<'a, str>>,
-    ) -> Self {
+    ) -> &mut Self {
         let key = key.into();
         let value = value.into();
 
@@ -259,20 +262,21 @@ impl<'a> QueryBuilder<'a> {
     /// use typespec_client_core::http::{Url, UrlExt as _};
     ///
     /// let mut url: Url = "https://contoso.com?a=1&b=2".parse().unwrap();
-    /// url.query_builder()
+    /// let mut query_builder = url.query_builder();
+    /// query_builder
     ///     .set_pair("a", "new_value")
-    ///     .set_pair("c", "3")
-    ///     .build();
+    ///     .set_pair("c", "3");
+    /// query_builder.build();
     /// let params: Vec<_> = url.query_pairs().collect();
     /// assert!(params.contains(&("a".into(), "new_value".into())));
     /// assert!(params.contains(&("b".into(), "2".into())));
     /// assert!(params.contains(&("c".into(), "3".into())));
     /// ```
     pub fn set_pair(
-        mut self,
+        &mut self,
         key: impl Into<Cow<'a, str>>,
         value: impl Into<Cow<'a, str>>,
-    ) -> Self {
+    ) -> &mut Self {
         let key = key.into();
         let value = value.into();
 
@@ -411,16 +415,18 @@ mod test {
     #[test]
     fn test_query_builder_empty_query() {
         let mut url = Url::parse("https://contoso.com").unwrap();
-        let builder = url.query_builder();
-        builder.set_pair("a", "1").build();
+        let mut builder = url.query_builder();
+        builder.set_pair("a", "1");
+        builder.build();
         assert_eq!(url.as_str(), "https://contoso.com/?a=1");
     }
 
     #[test]
     fn test_query_builder_new_parameter() {
         let mut url = Url::parse("https://contoso.com?b=2").unwrap();
-        let builder = url.query_builder();
-        builder.set_pair("a", "1").build();
+        let mut builder = url.query_builder();
+        builder.set_pair("a", "1");
+        builder.build();
         let params: Vec<_> = url.query_pairs().collect();
         assert!(params.contains(&("a".into(), "1".into())));
         assert!(params.contains(&("b".into(), "2".into())));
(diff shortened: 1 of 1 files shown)


sdk/core/typespec_client_core/src/http/mod.rs
```diff
@@ -104,6 +121,25 @@ impl UrlExt for Url {
 
         self.set_path(&new_path);
     }
+
+    fn set_query_pair(&mut self, key: &str, value: &str) -> &mut Self {
+        // Fast path: if the key doesn't exist, just append it
+        if self.query_pairs().all(|(k, _)| k != key) {
```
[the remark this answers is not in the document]
jhendrixMSFT: fixed

sdk/core/typespec_client_core/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+- Added `UrlExt::set_query_pair()` to simplify overwriting query parameter key values.
```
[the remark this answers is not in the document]
jhendrixMSFT: fixed

sdk/core/typespec_client_core/src/http/mod.rs
```diff
@@ -104,6 +121,25 @@ impl UrlExt for Url {
...
+        }
+
+        // Slow path: key exists, so we need to remove old values and add the new one
+        let mut new_url = self.clone();
+        new_url
+            .query_pairs_mut()
+            .clear()
+            .extend_pairs(self.query_pairs().filter(|(k, _)| k != key))
+            .append_pair(key, value);
+        *self = new_url;
```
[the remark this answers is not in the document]
jhendrixMSFT: @copilot That specific suggestion doesn't work as self is immutably borrowed on line 133, then attempts to be mutably borrowed on line 134.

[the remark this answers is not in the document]
jhendrixMSFT: You are correct though that we don't need to clone the entire string.  Fixed.

LarryOsterman (approved):

sdk/core/typespec_client_core/src/http/mod.rs
```diff
@@ -76,6 +76,23 @@ pub trait UrlExt: crate::private::Sealed {
...
+    /// Returns `&mut Self` to allow chaining.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use typespec_client_core::http::{Url, UrlExt as _};
+    ///
+    /// let mut url: Url = "https://contoso.com?a=1&b=2".parse().unwrap();
+    /// url.set_query_pair("a", "new_value")
+    ///    .set_query_pair("c", "3");
```
heaths: Do you expect to use chaining a lot? If so, we should return a builder of sorts instead, which is what `Url::query_pairs_mut()` does now, then finish up with a `finish()` method on that. It'll be expensive to do more than once rather than "batch" changes.

jhendrixMSFT: Good point.  At present, the updated codegen doesn't use chaining, but it could.  However, I would say that multiple query params is the uncommon case.  Based on that, do you think we should update this to return a builder, or just remove the return type?  If we later decided we wanted to return a builder, is that a breaking change for Rust (I _think_ the answer is "no")?

heaths: Given this is public API surface so you can use it in other crates, it would be a breaking change, yes. Or, at least, maybe we could provide a better way later like `query_builder_mut`. How urgently do you need this?

Note: it's not that I was asking if you plan to chain, but if you plan to call this multiple times however you call it. If we're frequently - per endpoint - calling this, it's rather wasteful. But would an API that chains or that can be called like so be easier to emit (it could probably work both ways, actually)?

```rust
let mut query_builder = url.query_builder();
query_builder.set_query_pair("foo", 1);
query_builder.set_query_pair("bar", "2");
query_builder.set_query(special_query_param);
query_builder.finish();
```

Effectively, I'm thinking a slight variation on `Url::query_pair_mut`. If it actually supported replacing parameters, we wouldn't need this.

I've also opened https://github.com/servo/rust-url/issues/1087 in hopes maybe we could get a change upstream. We're only a little behind the current `url@2.5.7` (we have `url@2.5.4`) and even less behind with `form_urlencoded@1.2.2` (we have `form_urlencoded@1.2.1`) but I checked and the newest versions don't yet have what want.

jhendrixMSFT: It's not super urgent.  However, storage is bumping into this for blob version IDs which go into the query string.  So, it would be good to not just put it on the backlog.

We can certainly emit the pattern that you suggest (it's pretty similar to what we do in Go) as there are cases where multiple query params are added in a method.  However, it's not the common case (usually it's just the `api-version`).

heaths: ...though, I'm not so hopeful that would fit with their design. `Url::query_pairs_mut` returns a `Serializer<'_, UrlQuery<'_>>` that, apart from `clear(&mut self)`, doesn't really have any destructive methods. It's really just a `String` underneath and, apart from `clear()` that just zeroes the `String`, is pretty much forward-only.

jhendrixMSFT: Per our offline conversation, I've updated this to use a builder pattern.  While `set_pair` is the main API of interest, I also added `append_pair` and `append_key_only` as we'll also need these methods in some cases and mixing this builder with `query_pairs_mut()` runs into borrowing issues.

sdk/core/typespec_client_core/src/http/mod.rs
```diff
@@ -104,6 +123,190 @@ impl UrlExt for Url {
...
+}
+
+/// A builder for setting query parameters on a URL.
+///
+/// This builder allows you to set multiple query parameters, overwriting any existing
+/// values with the same key. Call [`build()`](QueryBuilder::build) to apply the changes.
+pub struct QueryBuilder<'a> {
+    url: &'a mut Url,
+    // Use Vec to preserve insertion order
+    values: Vec<(std::borrow::Cow<'a, str>, Vec<std::borrow::Cow<'a, str>>)>,
```
heaths: These both can't be true. Once you start manipulating the query string of a `Url` reference, slices taken from it become invalid.

Also, since when did we care about order? The way we were doing it before won't preserve order.

jhendrixMSFT: `values` will hold the key/values to be set/appended.  We take ownership of the chunks in `Url` like you mentioned in `build()`.  Maybe I misunderstood the comment though.

The ordering was something Copilot came up with so that tests would be deterministic when comparing `url.as_str()` to the expected value.  I don't feel strong about keeping this semantic, but it does simplify writing tests.

jhendrixMSFT: FWIW, my original plan was to follow the pattern in Go by using a `HashMap` to hold the key/value pairs (makes overwriting existing keys easy).  Copilot pointed out that the non-determinism would impact tests.  Was gonna switch to a `BTreeMap`, Copilot said the `Vec` approach is easier, so here we are.

jhendrixMSFT: And I just found a bug in Copilot's implementation which makes me think switching back to the `BTreeMap` might be best (the keys will be stable for testing purposes).

sdk/core/typespec_client_core/src/http/mod.rs
```diff
@@ -104,6 +123,190 @@ impl UrlExt for Url {
...
+        } else {
+            // Add new key-value pair at the end
+            self.values.push((key, vec![value]));
+        }
+
+        self
+    }
+
+    /// Applies all the query parameter changes to the URL.
+    pub fn build(self) {
```
jhendrixMSFT, line 291: I want to change this to `build(&mut self)` so that we can chain calls to `build()`. Let me know if you disagree.

heaths: `build()` is generally last. What would you possibly chain after it? Builders are supposed to be init->setup->build (done).

sdk/core/typespec_client_core/src/http/mod.rs
heaths: Why do we care about order all of a sudden? That comes at a cost. `Vec` might still be the cheaper option in most cases, but I'm concerned that order is suddenly a concern. It wasn't before, nor should it be except, perhaps, for arrays.

sdk/core/typespec_client_core/src/http/mod.rs
```diff
@@ -104,6 +124,174 @@ impl UrlExt for Url {
...
+    }
+}
+
+/// A builder for setting query parameters on a URL.
+///
+/// This builder allows you to set multiple query parameters, overwriting any existing
+/// values with the same key. Call [`build()`](QueryBuilder::build) to apply the changes.
+pub struct QueryBuilder<'a> {
+    url: &'a mut Url,
+    values: std::collections::BTreeMap<std::borrow::Cow<'a, str>, Vec<std::borrow::Cow<'a, str>>>,
```
heaths: Nit: same with `Cow`: import the types. Seeing the fully-qualified type name over and over hurts readability.

sdk/core/typespec_client_core/src/http/mod.rs
heaths: > `values` will hold the key/values to be set/appended. We take ownership of the chunks in `Url` like you mentioned in `build()`. Maybe I misunderstood the comment though.

I see that now. I think I was reading it wrong before.
 
> The ordering was something Copilot came up with so that tests would be deterministic when comparing `url.as_str()` to the expected value. I don't feel strong about keeping this semantic, but it does simplify writing tests.

We can use `assert!(url.query().contains("foo=1"))` or even have `query_builder()` take a type param of some iterator that works for both `HashMap` and `BTreeMap` but we'd probably have to define one for the methods we need (the traits they share in common don't include any sort of updates beyond `Extend`). But I'm not sure that's worth it. I don't want to sacrifice perf just for the same of tests. In other places I have used a `BTreeMap` only in tests but in a way that worked a little easier to swap out the underlying type than it would here, but it's not impossible.

It's a bit hacky, but if the functions you call have the same signatures (or at least most of them), you could always use `#[cfg(test)]` and `#[cfg(not(test))]` conditions in the definition.

sdk/core/typespec_client_core/src/http/mod.rs
jhendrixMSFT: The chain in this context would be this.
```rust
builder.set_pair("foo", 1.to_string()).set_pair("bar", "2").build();
```

sdk/core/typespec_client_core/src/http/mod.rs
jhendrixMSFT: If you think `assert!(url.query().contains("foo=1"))` is good enough I'm fine with it.  So is the recommendation to switch to a `HashMap` then?

sdk/core/typespec_client_core/src/http/mod.rs
heaths: And that should work with `build(self)`. That's how builders generally work.

jhendrixMSFT: `append_pair` and `set_pair` return mutable references which prevents it though.

sdk/core/typespec_client_core/src/http/mod.rs
jhendrixMSFT: Good point, fixed.

sdk/core/typespec_client_core/src/http/mod.rs
jhendrixMSFT: Switch to `HashMap` and updated tests accordingly.

heaths: Yeah, as long as it stays internal. We may want to reevaluate at some point depending on how often we see enough pairs to a `Vec` or something.

heaths (approved):

PR #3335, opened 2025-11-11
[Storage] `set_legal_hold` for BlobClient

vincenttran-msft:
- [x] `.tsp`: https://github.com/Azure/azure-rest-api-specs/pull/38817

This test does the following:
- Adds `set_legal_hold` API to `BlobClient`
- Adds a way to specify "playback only" in Azure Core test macros

SetLegalHold:
>/// * [`date`()](crate::generated::models::BlobClientSetLegalHoldResultHeaders::date) - Date **REMOVED**
    /// * [`legal_hold`()](crate::generated::models::BlobClientSetLegalHoldResultHeaders::legal_hold) - x-ms-legal-hold **REMOVED**
 - [x] Changelog Entry

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -151,7 +151,7 @@ checksum = "c7c24de15d275a1ecfd47a380fb4d5ec9bfe0933f309ed5e705b775596a3574d"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.104",
+ "syn 2.0.110",
 ]
 
 [[package]]
@@ -162,7 +162,7 @@ checksum = "e539d3fca749fcee5236ab05e93a52867dd549cc157c8cb7f99595f3cedffdb5"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.104",
+ "syn 2.0.110",
 ]
 
 [[package]]
@@ -202,6 +202,7 @@ dependencies = [
  "futures",
  "hmac",
  "http",
+ "include-file",
  "json-patch",
  "openssl",
  "pin-project",
@@ -247,7 +248,7 @@ version = "0.5.0"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.104",
+ "syn 2.0.110",
  "tokio",
  "tracing",
  "tracing-subscriber",
@@ -306,7 +307,7 @@ dependencies = [
  "azure_core_test",
  "proc-macro2",
  "quote",
- "syn 2.0.104",
+ "syn 2.0.110",
  "tokio",
 ]
 
@@ -320,6 +321,7 @@ dependencies = [
  "azure_identity",
  "clap",
  "futures",
+ "moka",
  "reqwest",
  "serde",
  "serde_json",
@@ -447,6 +449,7 @@ dependencies = [
  "azure_security_keyvault_keys",
  "azure_security_keyvault_test",
  "futures",
+ "include-file",
  "openssl",
  "rand 0.9.2",
  "rustc_version",
@@ -466,7 +469,9 @@ dependencies = [
  "azure_security_keyvault_test",
  "criterion",
  "futures",
+ "include-file",
  "rand 0.9.2",
+ "rand_chacha 0.9.0",
  "reqwest",
  "rustc_version",
  "serde",
@@ -485,6 +490,7 @@ dependencies = [
  "azure_identity",
  "azure_security_keyvault_test",
  "futures",
+ "include-file",
  "rand 0.9.2",
  "rustc_version",
  "serde",
@@ -621,7 +627,7 @@ dependencies = [
  "proc-macro-crate",
  "proc-macro2",
  "quote",
- "syn 2.0.104",
+ "syn 2.0.110",
 ]
 
 [[package]]
@@ -710,7 +716,7 @@ dependencies = [
  "quote",
  "serde",
  "serde_json",
- "syn 2.0.104",
+ "syn 2.0.110",
  "tempfile",
  "toml",
 ]
@@ -803,7 +809,7 @@ dependencies = [
  "heck",
  "proc-macro2",
  "quote",
- "syn 2.0.104",
+ "syn 2.0.110",
 ]
 
 [[package]]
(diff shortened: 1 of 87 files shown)


sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -644,3 +646,38 @@ async fn test_encoding_edge_cases(ctx: TestContext) -> Result<(), Box<dyn Error>
 
     Ok(())
 }
+
+#[recorded::test(playback)]
+async fn test_set_legal_hold(ctx: TestContext) -> Result<(), Box<dyn Error>> {
+    // Recording Setup
+    let recording = ctx.recording();
+    use_storage_account(StorageAccount::Versioned);
```
jalauzon-msft: Use regular account

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -644,3 +644,37 @@ async fn test_encoding_edge_cases(ctx: TestContext) -> Result<(), Box<dyn Error>
...
+    blob_client.set_legal_hold(false, None).await?;
+    let response = blob_client.get_properties(None).await?;
+    // Assert
+    let legal_hold = response.legal_hold()?;
+    assert!(!legal_hold.unwrap());
+
+    blob_client.delete(None).await?;
+
+    Ok(())
+}
```
vincenttran-msft, line 680: This test still needs to be playback only.

- Against a regular Storage account (no versioning enabled)
    - ><Error><Code>ImmutableStorageWithVersioningIsNotEnabled</Code><Message>ImmutableStorageWithVersioning: feature is not enabled.
    - The above error is on the `set_legal_hold()` operation
- Against a Storage account with `ImmutableStorageWithVersioning enabled`
     - ><Error><Code>ContainerImmutableStorageWithVersioningEnabled</Code><Message>The requested operation is not allowed as the container has a immutable storage with versioning enabled.
     - The above error is on the `container_client.delete()` operation
     - Note that I already set it to `False` and verified with a `get_properties()` call.

Furthermore, all tests that set legal hold in Python for example have to use the [management plane](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/tests/test_common_blob.py#L1519) to delete it.

image

jalauzon-msft (approved):

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -644,3 +644,37 @@ async fn test_encoding_edge_cases(ctx: TestContext) -> Result<(), Box<dyn Error>
...
+    let blob_client = container_client.blob_client(&get_blob_name(recording));
+    container_client.create_container(None).await?;
+    create_test_blob(&blob_client, None, None).await?;
+
+    // Set Legal Hold
+    blob_client.set_legal_hold(true, None).await?;
+    let response = blob_client.get_properties(None).await?;
+    // Assert
+    let legal_hold = response.legal_hold()?;
+    assert!(legal_hold.unwrap());
```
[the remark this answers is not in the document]
vincenttran-msft: Test code, is fine.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -644,3 +644,37 @@ async fn test_encoding_edge_cases(ctx: TestContext) -> Result<(), Box<dyn Error>
...
+    // Assert
+    let error = response.unwrap_err().http_status();
+    assert_eq!(StatusCode::Conflict, error.unwrap());
+
+    // Remove Legal Hold
+    blob_client.set_legal_hold(false, None).await?;
+    let response = blob_client.get_properties(None).await?;
+    // Assert
+    let legal_hold = response.legal_hold()?;
+    assert!(!legal_hold.unwrap());
```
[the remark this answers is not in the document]
vincenttran-msft: Test code, is fine.

sdk/core/azure_core_test_macros/src/test.rs
```diff
@@ -50,6 +50,13 @@ pub fn parse_test(attr: TokenStream, item: TokenStream) -> Result<TokenStream> {
         });
     }
 
+    // Ignore playback-only tests if not running playback tests.
+    if recorded_attrs.playback && test_mode != TestMode::Playback {
+        test_attr.extend(quote! {
+            #[ignore = "skipping playback-only tests"]
+        });
+    }
```
heaths, line 58: What's the use case for this? How do you record this in the first place then? If you don't need a recording, don't use `#[recorded::test]`. Use a normal `#[tokio::test]`.

vincenttran-msft: I ran it first with `#[recorded::test]` and `AZURE_TEST_MODE=record`, to generate the initial recording, then I marked it as `#[recorded::test(playback)]` so that this test will not execute in the live test pipeline and only in playback mode.

heaths: I understand what it does. I'm asking: why? You can use mock tests. This feels like a lot of work to re-record tests and I don't necessarily want to support a "record-only" concept for other services.

heaths: Alright, but this needs better docs including a warning that this is probably not what most people want. Add mention of it to `src/lib.rs` for the macro docs, perhaps something more like this now that we have two parameters: https://github.com/heaths/include-file/blob/3433b2002e571e89dc9c0f1a58fd2b49919aa4f2/src/lib.rs#L83-L88

sdk/core/azure_core_test_macros/src/test.rs
```diff
@@ -140,6 +149,13 @@ impl Parse for Attributes {
             }
         }
 
+        if attrs.live && attrs.playback {
```
heaths: Why not? If setup is hard - as it is now for this - but could work live and work only off a recording in CIs, why couldn't it be both? I'd say: allow it, and if it fails in one or the other, let them fix it and change the attributes as needed.

heaths (approved):

PR #3340, opened 2025-11-12
Storage Partitioned Transfer Base

jaschrep-msft:
Implements two foundation components to implement partitioned upload and copy.

- `PartitionedStream`: Consumes a `Box<dyn SeekableStream>` and converts it to a `Stream<Item = Result<Bytes, Error>>` where each `Ok(Bytes)` returned is a contiguously buffered partition to be used for a put block or equivalent request.
- `run_all_with_concurrency_limit()`: Takes a sequence of async jobs (`impl FnOnce() -> impl Future<Output = Result<(), Error>>`). These will be sequences of put block operations or equivalent requests.

### DIFF
--- a/sdk/storage/azure_storage_blob/src/streams/partitioned_stream.rs
+++ b/sdk/storage/azure_storage_blob/src/streams/partitioned_stream.rs
@@ -1,10 +1,5 @@
 use pin_project::pin_project;
-use std::{
-    mem::{self, MaybeUninit},
-    num::NonZero,
-    pin::Pin,
-    task::Poll,
-};
+use std::{mem, num::NonZero, pin::Pin, slice, task::Poll};
 
 use azure_core::stream::SeekableStream;
 use bytes::{Bytes, BytesMut};
@@ -59,14 +54,21 @@ impl Stream for PartitionedStream {
                     Poll::Ready(Some(Ok(ret.freeze())))
                 };
             }
-            match ready!(this.inner.as_mut().poll_read(cx, unsafe {
+
+            let spare_capacity = this.buf.spare_capacity_mut();
+            let spare_capacity = unsafe {
                 // spare_capacity_mut() gives us the known remaining capacity of BytesMut.
                 // Those bytes are valid reserved memory but have had no values written
                 // to them. Those are the exact bytes we want to write into.
-                // This transmuted data is not saved to a variable, leaving it inaccessible
-                // to anything but poll_read().
-                mem::transmute::<&mut [MaybeUninit<u8>], &mut [u8]>(this.buf.spare_capacity_mut())
-            })) {
+                // MaybeUninit<u8> can be safely cast into u8, and so this pointer cast
+                // is safe. Since the spare capacity length is safely known, we can
+                // provide those to from_raw_parts without worry.
+                slice::from_raw_parts_mut(
+                    spare_capacity.as_mut_ptr() as *mut u8,
+                    spare_capacity.len(),
+                )
+            };
+            match ready!(this.inner.as_mut().poll_read(cx, spare_capacity)) {
                 Ok(bytes_read) => {
                     // poll_read() wrote bytes_read-many bytes into the spare capacity.
                     // those values are therefore initialized and we can add them to


sdk/storage/azure_storage_blob/src/streams/partitioned_stream.rs
```diff
@@ -0,0 +1,187 @@
...
+    type Item = AzureResult<Bytes>;
+
+    fn poll_next(
+        self: Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> Poll<Option<Self::Item>> {
+        let this = self.get_mut();
+
+        loop {
+            if this.inner_complete || this.buf_offset >= this.buf.len() {
```
vincenttran-msft: I am wondering if the RHS expression should instead be:
```rust
this.buf_offset == this.buf.len()
```
If I am understanding correctly, wouldn't the case that the buffer offset being greater than the buffer size mean we will lose that data:
`(this.buf_offset) - (buffer size) = # of bytes being lost`

Since we did in fact read it out of the stream (as evidenced by `this.buf_offset`) but it's bigger than the buffer, so those extra bytes got eaten up by the aether?

I think this is a super super edge case since we really shouldn't ever get into that case, but wondering if we should move the "greater-than" case to raise an error.

jaschrep-msft: I believe it's impossible to hit the greater-than case due to how the inner stream is read from. I think that i'm just covering all cases as a style choice so someone doesn't wonder why we'd want the else block for a `>` result. I have no real preference on which to use.

vincenttran-msft: We can leave as-is, good point about then sort of misleading with a explicit `>` case. Thanks!

sdk/storage/azure_storage_blob/src/partitioned_transfer/mod.rs
```diff
@@ -0,0 +1,196 @@
+use std::{future::Future, num::NonZero};
+
+use futures::{
+    future::{self},
+    Stream, TryStreamExt,
+};
+
+type AzureResult<T> = azure_core::Result<T>;
+
```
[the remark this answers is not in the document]
heaths: It's not public. Only public functions need to be documented. That said, at least a blurb explaining what this does might be handy for future maintainers.

sdk/storage/azure_storage_blob/src/streams/partitioned_stream.rs
```diff
@@ -0,0 +1,187 @@
...
+                    }
+                }
+            }
+        }
+    }
+}
+
+impl FusedStream for PartitionedStream {
+    fn is_terminated(&self) -> bool {
+        self.inner_complete && self.buf.is_empty()
```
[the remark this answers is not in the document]
jaschrep-msft: Incorrect

sdk/storage/azure_storage_blob/src/streams/partitioned_stream.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    fn take(&mut self) -> Vec<u8> {
+        let mut ret = mem::replace(
+            &mut self.buf,
+            vec![0u8; std::cmp::min(self.partition_len, self.inner.len() - self.total_read)],
+        );
+        ret.truncate(self.buf_offset);
+        self.buf_offset = 0;
+        ret
+    }
```
heaths: You're already taking a dependency on `Bytes`. Use it. It already has all the functionality for this. At the very least, don't repeat the calculation of `buf.len()` that you'd already have computed during construction. `Bytes` has already been well-tested. I see potential faults here.

sdk/storage/azure_storage_blob/src/streams/partitioned_stream.rs
```diff
@@ -0,0 +1,199 @@
...
+        loop {
+            if this.inner_complete || this.buf_offset >= this.buf.len() {
+                let ret = this.take();
+                return if ret.is_empty() {
+                    Poll::Ready(None)
+                } else {
+                    Poll::Ready(Some(Ok(Bytes::from(ret))))
+                };
+            } else {
+                match ready!(pin!(&mut this.inner).poll_read(cx, &mut this.buf[this.buf_offset..]))
```
heaths: More consistent to use `#[pin_project]` like we do in our pager and pollers. Also, why are you matching on `ready!(...)`? I don't understand how that's useful. That's meant to produce output when implementing futures.

jaschrep-msft: @heaths why is implementing a poll method on a stream so different from "implementing futures"? I'm definitely new to writing poll methods, but my understanding is that `ready!(exp)` expands to the following:
```rust
match exp {
    Poll::Ready(ret) = ret,
    Poll::Pending => return Poll::Pending,
}
```
That seems like what I want to do here, right? If the read of the inner stream is pending, then I want to propagate that state. Is it some sort of style thing to write out the full match instead of the macro in some scenarios? I don't understand what I'd do instead in this scenario except not implement `poll_next()` in the first place.

I'll look into pin_project. New concept to me.

heaths: `poll_read` would return a `Poll::Ready(_)` or `Poll::Pending`, yes. But by wrapping it in a `ready!(_)` you're making it `Poll::Ready`. `ready!(_)` is to return a value as an already-ready state. `Pager` is probably a good one to look at, and after changes I'm making to manually do it for various reasons: https://github.com/Azure/azure-sdk-for-rust/pull/3372

jaschrep-msft: I see the parts of your linked PR where you are `match`ing the result of the poll and returning. Not to dig in, but I am looking at the [docs](https://doc.rust-lang.org/stable/std/task/macro.ready.html) and (apart from a variable name) the documented expansion of the macro looks character-for-character like what you're writing in your PR. I'd like to understand either how I am wrong in my comparison or why I am meant to manually write out the full expansion even if they are identical.

sdk/storage/azure_storage_blob/src/streams/partitioned_stream.rs
jaschrep-msft: Hey @heaths, I can't for the life of me figure out how to get the slice of `BytesMut` between `len()` and `capacity()`. I know I need `spare_capacity_mut()` but I cannot get the resulting `&mut [MaybeUninit<u8>]` into `&mut [u8]`. Every time I think I have it figured out it depends on nightly Rust. If you could please just link a docs page or give me a snippet with the appropriate `use` lines included, I can finish this.

sdk/storage/azure_storage_blob/src/streams/partitioned_stream.rs
heaths: Still say you should use `#[pin_project]` like we do elsewhere. Makes everything easier and the code cleaner.

sdk/storage/azure_storage_blob/src/streams/partitioned_stream.rs
heaths: That `Bytes::spare_capacity_mut()` returns a `[MaybeUninit<u8>]` isn't meant to be an inconvenience but to help enforce safety. Since you're basically trying to make an out pointer it seems, did you look at `MaybeUninit`'s <https://doc.rust-lang.org/nightly/core/mem/union.MaybeUninit.html#out-pointers>? Returning a `Vec<MaybeUninit<u8>>` is the right thing to do, in which case you can create the `Vec` using <https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.from_raw_parts>. Both are unsafe, which simply means you have to guarantee their safety in use so a comment explaining why you're sure it's safe is common.

My point, though, was that you're repeating much of what `Bytes` already has. Did you take a look at its APIs: <https://docs.rs/bytes>? We also already have a `ByteStream` that might give you want you need for the stream itself. If not, and any changes you need could be useful elsewhere, we can update core.

My original comment was just to avoid duplicating what's already been done and well-tested. If neither meet your needs than this code seems fine.

sdk/storage/azure_storage_blob/src/streams/partitioned_stream.rs
jaschrep-msft: Yes, I am working on integrating that along with the `Bytes` usage.

sdk/storage/azure_storage_blob/src/streams/partitioned_stream.rs
jaschrep-msft: Hey @heaths, I've opted for `transmute` instead of `from_raw_parts`. The code ends up looking far cleaner and I think a conversion from `&mut [MaybeUninit<u8>]` to `&mut [u8]` is about as clean as it gets. Hopefully the comments do a good job explaining my work.

sdk/storage/azure_storage_blob/src/streams/partitioned_stream.rs
```diff
@@ -0,0 +1,204 @@
...
+                };
+            }
+            match ready!(this.inner.as_mut().poll_read(cx, unsafe {
+                // spare_capacity_mut() gives us the known remaining capacity of BytesMut.
+                // Those bytes are valid reserved memory but have had no values written
+                // to them. Those are the exact bytes we want to write into.
+                // This transmuted data is not saved to a variable, leaving it inaccessible
+                // to anything but poll_read().
+                mem::transmute::<&mut [MaybeUninit<u8>], &mut [u8]>(this.buf.spare_capacity_mut())
+            })) {
```
jaschrep-msft: fwiw, if we really don't want `transmute` in here, this is what we'll end up with:
```rust
let spare_capacity = this.buf.spare_capacity_mut();
match ready!(this.inner.as_mut().poll_read(cx, unsafe {
    // spare_capacity_mut() gives us the known remaining capacity of BytesMut.
    // Those bytes are valid reserved memory but have had no values written
    // to them. Those are the exact bytes we want to write into.
    &mut Vec::<u8>::from_raw_parts(
        spare_capacity.as_mut_ptr() as *mut u8,
        spare_capacity.len(),
        spare_capacity.len(),
    )
    // mem::transmute::<&mut [MaybeUninit<u8>], &mut [u8]>(this.buf.spare_capacity_mut())
})) {
```
@heaths happy to make this change if we are unwilling to touch `transmute` but to my eyes the usage here is extremely safe and it doesn't seem like `from_raw_parts` is much safer in the first place.

heaths: You could keep the `transmute` then, but at least add a comment sort of affirming why it's safe. That's typical in unsafe code blocks e.g., "// This is safe because ..." i.e., you did all your bounds checking up front. Make sure it's covered well in tests.

jaschrep-msft: upon further research this transmute does not appear to be safe. the pointer and length of the slice can get swapped pretty easy. i'll use raw parts

jaschrep-msft: oh no. from_raw_parts results in a vec that frees its memory. more exploration needed.

heaths (approved):

PR #3341, opened 2025-11-12
Fail on live test failure

danieljurek:


eng/pipelines/templates/jobs/live.tests.yml
```diff
@@ -125,6 +125,9 @@ jobs:
 
           Write-Host "./eng/scripts/Test-Packages.ps1 -PackageInfoDirectory '$(Build.ArtifactStagingDirectory)/PackageInfo'"
           ./eng/scripts/Test-Packages.ps1 -PackageInfoDirectory '$(Build.ArtifactStagingDirectory)/PackageInfo'
+
+          Write-Host "Last exit code: $LASTEXITCODE";
+          exit $LASTEXITCODE
```
weshaggard, line 130: Would it be better to do exit from Test-Pacakges.ps1?

danieljurek: We had to do the [same thing](https://github.com/Azure/azure-sdk-for-net/blob/dd030d5e93ed7158f7be669f8044d8a775528a01/eng/pipelines/templates/jobs/live.tests.yml#L192C1-L193C34) when we migrated to federated auth for the other SDKs. 

An issue was also opened to get the behavior changed: https://github.com/microsoft/azure-pipelines-tasks/issues/19951

weshaggard (approved):

danieljurek:
The two services with live test configs in the pipelines have failures that were present before this change goes in. This can merge.

eng/pipelines/templates/jobs/live.tests.yml
heaths: Any chance we could just use PowerShell error handling idiomatically? See https://devblogs.microsoft.com/powershell/erroraction-and-errorvariable/, along with example:

```pwsh
[CmdletBinding()]
param ()

function fail {
    [CmdletBinding()]
    param (
        [Parameter()]
        [switch] $Fail
    )

    if ($Fail) {
        Write-Error "Failed"
    }
}

Write-Verbose "Won't fail"
fail -ev +err

Write-Verbose "Will fail"
fail -Fail -ev +err

Write-Verbose "Won't fail"
fail -ev +err

Write-Verbose "Will fail"
fail -Fail -ev +err

if ($err) {
    exit 1
}
```

Pass `-ea Stop` or not to the script to see how `$ErrorActionPreference` is inherited. Works right every time, and like any other PowerShell cmdlets would (should; built-ins work right, at least).

PR #3342, opened 2025-11-12
[Storage] `delete/set_immutability_policy` for `BlobClient`

vincenttran-msft:
- [x] `.tsp`: https://github.com/Azure/azure-rest-api-specs/pull/38750
- Adds `set/delete_immutability_policy` to BlobClient
- Refactors test utilities to allow for specifying what Storage account type to use

### DIFF
--- a/sdk/storage/azure_storage_blob/tests/blob_service_client.rs
+++ b/sdk/storage/azure_storage_blob/tests/blob_service_client.rs
@@ -268,7 +268,7 @@ async fn test_find_blobs_by_tags_service(ctx: TestContext) -> Result<(), Box<dyn
 async fn test_get_service_stats(ctx: TestContext) -> Result<(), Box<dyn Error>> {
     // Recording Setup
     let recording = ctx.recording();
-    let (options, endpoint) = recorded_test_setup(recording, StorageAccount::Versioned);
+    let (options, endpoint) = recorded_test_setup(recording, StorageAccount::Standard);
     let endpoint = endpoint.replace(
         ".blob.core.windows.net/",
         "-secondary.blob.core.windows.net/",


sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -644,3 +647,46 @@ async fn test_encoding_edge_cases(ctx: TestContext) -> Result<(), Box<dyn Error>
...
+    let response = blob_client
+        .set_immutability_policy(Some(immutability_policy_options))
+        .await?;
+
+    // Assert
+    let mode = response.immutability_policy_mode()?;
+    let expires_on = response.immutability_policy_expires_on()?;
+
+    let response_mode: BlobImmutabilityPolicyMode = test_immutability_policy_mode.into();
+    assert_eq!(response_mode, mode.unwrap());
```
vincenttran-msft: Food for thought:
On the **request** side, we have `ImmutabilityPolicyMode`
On the **response** side, we have `BlobImmutabilityPolicyMode`

ImmutabilityPolicyMode is a sub-set of BlobImmutabilityPolicyMode, as the additional enum `Mutable` can be returned by the service.

So this raises the question:

- Leave implementation as-is, cx must track this subtlety between response/request model (I provided a `From` impl. so they can just call `.into()`)
- Flatten it so that on both request and response side we use `BlobImmutabilityPolicyMode`, fully knowing that we are going to offer an enum on the request side that will never work (`Mutable`)

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -644,3 +647,46 @@ async fn test_encoding_edge_cases(ctx: TestContext) -> Result<(), Box<dyn Error>
...
+    assert_eq!(test_expiry_time, expires_on);
```
vincenttran-msft: Will need to manually strip out decimals since evaluating the actual HTTP request and response headers, they are exactly the same:
>"x-ms-immutability-policy-until-date": "Wed, 12 Nov 2025 22:25:39 GMT",
>"x-ms-immutability-policy-until-date": "Wed, 12 Nov 2025 22:25:39 GMT",

So this fails due to the fact that this is our raw input (non-stripped to be in service-friendly form) against the service-friendly returned format.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: As discussed offline, combine to `ImmutabilityPolicyMode` which is the superset and document that `Mutable` is only intended for response side.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -644,3 +648,81 @@ async fn test_encoding_edge_cases(ctx: TestContext) -> Result<(), Box<dyn Error>
 
     Ok(())
 }
+
+#[recorded::test]
+async fn test_immutability_policy(ctx: TestContext) -> Result<(), Box<dyn Error>> {
```
vincenttran-msft, line 686: Testing against a regular Storage account yields:
>HTTP/1.1 400 ImmutableStorageWithVersioning: feature is not enabled.

Testing against versioned _without_ immutable storage on:
><Error><Code>ImmutableStorageWithVersioningIsNotEnabled</Code><Message>ImmutableStorageWithVersioning: feature is not enabled.

So, as a result, had to use a versioned account + immutable storage with versioning enabled.

As a result, having the same issues with deleting the container as `set_legal_hold`:
><Error><Code>ContainerImmutableStorageWithVersioningEnabled</Code><Message>The requested operation is not allowed as the container has a immutable storage with versioning enabled.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -436,4 +438,31 @@ impl BlobClient {
...
     }
+
+    /// Sets the immutability policy on the blob.
+    ///
+    /// # Arguments
+    ///
+    /// * `options` - Optional configuration for the request.
+    pub async fn set_immutability_policy(
+        &self,
+        immutability_policy_expiry: &OffsetDateTime,
```
jalauzon-msft: Just call this `expiry`? (Change in tsp if needed). Also document this param.

sdk/storage/azure_storage_blob/tests/blob_client.rs
jalauzon-msft: Mark as playback only for now, I guess. Open an issue to track getting proper clean up on this and the legal hold tests so we can enable them to run Live.

vincenttran-msft: https://github.com/Azure/azure-sdk-for-rust/issues/3358
Done, will mark playback only once we get `set_legal_hold` merged since I unfortunately did not base it off of that PR.

heaths: Why can't you update your `test-resources.bicep` file? That's the whole reason it's there and you should be using it, just like in all other languages (I wrote those original scripts for **every** Azure SDK language). I'd rather we don't have a confusing concept of "playback-only" tests. That implies you have to record them at some point which means modifying source which means people have to manually interject at least a couple times. Just use a `#[tokio::test]` with a `MockHttpClient` from our `azure_core_test` crate instead. I do that in lots of places. Pretty easy to use.

jalauzon-msft: @heaths, we need to mark these tests as playback only because containers with immutability policies (and legal holds) are very difficult to clean up. They require the management plane to do so which we don't have yet in Rust. We will try and figure out a way to delete them and change them to regular recorded tests later. There is nothing that can be changed in the bicep to accomplish this.

We have the concept of playback only in every other language and we use it sparingly for situations like this. It would be nice to have that concept available in Rust. Mocking would require extra effort and means these tests will work and look different than the rest of our tests.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -677,3 +681,81 @@ async fn test_set_legal_hold(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
+    let blob_client = container_client.blob_client(&get_blob_name(recording));
+    container_client.create_container(None).await?;
+    create_test_blob(&blob_client, None, None).await?;
+
+    // Set Immutability Policy (No Mode Specified, Default to Unlocked)
+    let expiry_1_str = recording.var(
+        "expiry_1",
+        Some(VarOptions {
+            default_value: Some(
+                to_rfc3339(&(OffsetDateTime::now_utc() + Duration::from_secs(5))).into(),
```
jalauzon-msft, line 700: You don't really need to write a specific format to the variable string, you could use default to string as long as you are able to read it back into a date.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -677,3 +681,81 @@ async fn test_set_legal_hold(ctx: TestContext) -> Result<(), Box<dyn Error>> {
...
+    let expires_on = response.immutability_policy_expires_on()?;
+    assert_eq!(ImmutabilityPolicyMode::Locked, mode.unwrap());
+    // Need to ignore nanoseconds due to Service truncation
+    assert_eq!(expiry_2.replace_nanosecond(0)?, expires_on.unwrap());
+
+    // Sleep to allow immutability policy to expire
+    time::sleep(Duration::from_secs(5)).await;
+
+    blob_client.delete(None).await?;
+    use_storage_account(StorageAccount::Standard);
```
jalauzon-msft: I don't love how we have to set this back...

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: From my cursory search, it seems like there is no format-agnostic way to get a string representation that we can still convert back, so will leave as-is.

jalauzon-msft (approved):

PR #3347, opened 2025-11-14
Initial C Bindings for Cosmos DB SDK for Rust

### LINKED ISSUE #3444: Cosmos: Create `libazurecosmos`, a C library wrapping the Rust SDK
As part of our experimentation with using the Rust SDK as a core for other SDKs, we need to expand the `azure_data_cosmos_native` package (which builds "libazurecosmos") to be a full wrapper around the Rust SDK.

analogrelay:
This PR introduces C language bindings (FFI) for the `azure_data_cosmos` Rust SDK, enabling C/C++ applications to interact with Azure Cosmos DB.

### Key Changes

**New `azure_data_cosmos_native` Package:**
- Complete C API with FFI layer exposing core Cosmos DB operations (originally developed by a partner team)
- Runtime and call context management for async/sync bridging. The Runtime Context defines the threads on which async operations will run, and the Call Context provides a place for call-specific reporting of error details. This is based on the pattern found in the [AMQP package](https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/core/azure-core-amqp/src/impl/rust_amqp/rust_amqp/rust_wrapper/src/call_context.rs) for C++
- Comprehensive error handling with optional detailed error information
- CRUD operations for databases, containers, and items
- Connection string support

**C Tests:**
- Shared test infrastructure (`test_common.h`) with automatic test discovery
- Basic tests of item/container/database CRUD.
- Integration with CMake/CTest

**Build System:**
- CMake build configuration with organized output directories
- `cbindgen` integration for automatic header generation
- pkg-config file generation for library integration
- Cross-platform support (Linux, macOS, Windows)

Closes #3444

### DIFF
--- a/sdk/cosmos/azure_data_cosmos_native/README.md
+++ b/sdk/cosmos/azure_data_cosmos_native/README.md
@@ -252,7 +252,7 @@ The test framework automatically discovers the new test during the build process
 
 ### Available Assertion Macros
 
-The test framework provides several assertion macros in [c_tests/test_common.h](c_tests/test_common.h):
+The test framework provides several assertion macros in `c_tests/test_common.h`:
 
 - **`ASSERT(condition, message)`** - Check condition, continue on failure
 - **`REQUIRE(condition, message)`** - Check condition, goto cleanup on failure


sdk/core/typespec_client_core/src/http/clients/reqwest.rs
```diff
@@ -58,10 +58,7 @@ impl HttpClient for ::reqwest::Client {
             "performing request {method} '{}' with `reqwest`",
             url.sanitize(&DEFAULT_ALLOWED_QUERY_PARAMETERS)
         );
-        let rsp = self
-            .execute(reqwest_request)
-            .await
-            .with_context(ErrorKind::Io, "failed to execute `reqwest` request")?;
+        let rsp = self.execute(reqwest_request).await.unwrap();
```
analogrelay: I had to add this for some debugging, it's not meant to stay in. Sorry! I'll remove it shortly.

sdk/cosmos/azure_data_cosmos_native/CMakeLists.txt
```diff
@@ -1,8 +1,9 @@
 # cSpell:ignore cosmosctest CRATETYPES endforeach
 
-project(cosmosctest C)
 cmake_minimum_required(VERSION 4.1)
```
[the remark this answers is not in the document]
analogrelay: CMake 4.x exists, and we want to use the latest version for this build.

sdk/cosmos/azure_data_cosmos_native/include/azurecosmos.h
```diff
@@ -0,0 +1,609 @@
...
+  COSMOS_ERROR_CODE_NOT_FOUND = 404,
+  COSMOS_ERROR_CODE_CONFLICT = 409,
+  COSMOS_ERROR_CODE_PRECONDITION_FAILED = 412,
+  COSMOS_ERROR_CODE_REQUEST_TIMEOUT = 408,
+  COSMOS_ERROR_CODE_TOO_MANY_REQUESTS = 429,
+  COSMOS_ERROR_CODE_INTERNAL_SERVER_ERROR = 500,
+  COSMOS_ERROR_CODE_BAD_GATEWAY = 502,
+  COSMOS_ERROR_CODE_SERVICE_UNAVAILABLE = 503,
+  COSMOS_ERROR_CODE_AUTHENTICATION_FAILED = 1001,
+  COSMOS_ERROR_CODE_DATA_CONVERSION = 1002,
```
tvaron3, line 38: Are all of these status codes or are some substatus codes? Just wondering if its posssible if some are overloaded.

analogrelay: In the original prototype, they were, but I actually think I'm going to restructure these to be simple sequential integers. The error codes here are for errors that need to cross the FFI boundary, so they should generally be focused on network connectivity issues, invalid argument errors, marshalling errors, etc. Retries will be handled inside the SDK, and I'd argue that non-retryable failed responses should be returned to the wrapping SDK as "successes" (and the wrapper can check the status code)

analogrelay: I'll file some follow-ups for that work though, we also need to return response metadata (status code, headers, etc.), which we don't do today.

heaths (approved):
Signing off on `Cargo.lock` (only non-Cosmos change) and to unblock the PR.

Issue #3175, opened 2025-10-10
Cosmos: Build `GlobalEndpointManager` For Endpoint Management


kundadebdatta:
### Acceptance Criteria:

- Build `GlobalEndpointManager` For Endpoint Management

PR #3348, opened 2025-11-14
Cosmos: Adds `GlobalEndpointManager`, `ClientRetryPolicy` and `LocationCache` To Support Cross-Regional Routing

kundadebdatta:
## Description:
 
This PR is introducing `GlobalEndpointManager`, `ClientRetryPolicy`, `LocationCache` and `AccountProperties` to support cross-regional routing in the rust SDK. Some of the key design decisions are highlighted below:

- We used [Moka ](https://docs.rs/moka/latest/moka/)as a caching library which supports thread safe access of the cached data and uses LRFU based cache eviction mechanism.

- The cache refresh has been done on a TTL (time to live) model, where each incoming request has the opportunity to refresh `AccountProperties` and `LocationCache`, if the TTL expires or `force_refresh` is passed as `True`.

- We haven't took the background refresh route yet, since we will need to be a little careful and optimized for the rust bg task management. However, our long term goal is to refresh the cache in the background, so that we do not penalize an incoming request to refresh the cache.

The below flow diagram depicts the entire component interactions to better understand how the request will be made.

## Flow Diagram:

```mermaid
sequenceDiagram
    participant J as ContainerClient
    participant A as CosmosPipeline
    participant B as RetryHandler
    participant C as ClientRetryPolicy
    participant D as Core Pipeline <br> send_raw <br> through delegate
    participant E as GlobalEndpointManager
    participant F as LocationCache


    J-->>A: 1. send(CosmosRequest)
    A->>B: 2. send<Sender, Fut>()
    loop Retry Iterations with force-refresh flag = false/ true
        B->>C: 3. before_send_request(request)
        C->>E: 4. refresh_location_async() <br> Refresh Cache only when TTL Expires OR <br> force_refresh is requested. <br><br> NO-OP when either of the above is false.
        E->>E: 5. get_database_account() <br> Fetch Account Properties <br> from Gateway.
        E->>F: 6. on_database_account_read() <br> Refresh Location Cache
        C->>E: 7. resolve_service_endpoint() <br> Fetch and set the next endpoint in <br> CosmosRequest to route request.
        B->>D: 8. Call sender delegate(request)
        B->>C: 9. Invoke should_retry(result)
        B->>B: 10. RetryAfter <br> (TimeSpan.Zero)
    end
        B->>J: 11. Return result if No Retry
```

## Closing issues

Fixes Azure/azure-sdk-for-rust#3175

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,6 +202,7 @@ dependencies = [
  "futures",
  "hmac",
  "http",
+ "include-file",
  "json-patch",
  "openssl",
  "pin-project",
@@ -1647,9 +1648,9 @@ dependencies = [
 
 [[package]]
 name = "include-file"
-version = "0.5.0"
+version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3b82360a1cfc38a30a967d4e1b58745ff9258c9326772293402d88d6dd201229"
+checksum = "4db0c2483923cdc1a04d371af232e93215bef3d6a761b663870dd2db4ca795a9"
 dependencies = [
  "proc-macro2",
  "syn 2.0.110",


sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -66,3 +71,24 @@ pub trait RetryPolicy: Send + Sync {
     /// A `RetryResult` indicating the retry decision.
     async fn should_retry(&mut self, response: &azure_core::Result<RawResponse>) -> RetryResult;
 }
+
+fn get_substatus_code_from_error(err: &azure_core::Error) -> SubStatusCode {
+    if let ErrorKind::HttpResponse { raw_response, .. } = err.kind() {
+        raw_response
+            .as_ref()
+            .and_then(|r| r.headers().get_as(&SUB_STATUS).ok())
+            .and_then(|raw: u16| SubStatusCode::try_from(raw).ok())
```
analogrelay: If you implement the `FromHeaders` trait (in `azure_core::http`) on `SubStatusCode`, you'll be able to replace this with a single call to `r.headers().get::<SubStatusCode>()`. The `FromHeaders` trait allows the `SubStatusCode` to specify the header name to read AND how to read it, so it all becomes self-describing and strongly-typed :).

sdk/cosmos/azure_data_cosmos/src/constants.rs
```diff
@@ -40,3 +40,219 @@ pub const SUB_STATUS: HeaderName = HeaderName::from_static("x-ms-substatus");
...
 
 pub(crate) const PREFER_MINIMAL: HeaderValue = HeaderValue::from_static("return=minimal");
+
+pub const ACCOUNT_PROPERTIES_KEY: &str = "account_properties_key";
+
+#[repr(u16)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+#[non_exhaustive]
+pub enum SubStatusCode {
+    Unknown = 0,
```
analogrelay: Let's remove this and just use `Option<SubStatusCode>` in places where we might not have a status code (or it might be unknown). Then `None` can represent an unknown/missing value.

sdk/cosmos/azure_data_cosmos/src/constants.rs
```diff
@@ -40,3 +40,219 @@ pub const SUB_STATUS: HeaderName = HeaderName::from_static("x-ms-substatus");
 pub const QUERY_CONTENT_TYPE: ContentType = ContentType::from_static("application/query+json");
 
 pub(crate) const PREFER_MINIMAL: HeaderValue = HeaderValue::from_static("return=minimal");
+
+pub const ACCOUNT_PROPERTIES_KEY: &str = "account_properties_key";
+
+#[repr(u16)]
```
analogrelay: I'd avoid using `#[repr(u16)]` unless you're calling or exposing C APIs (which we're not doing... **yet**). You're already doing the right thing below, and using a big `match` to map integers to enum variants (rather than just casting the integer to the enum, like you might do in .NET). Rust conventions strongly discourage doing that, because the whole Rust ecosystem expects that an enum can **only** have one of the values listed in it's definition, and **no other value**.

Since we already have the "mapping" between `u16` and `SubStatusCode` in the `TryFrom<u16>` implementation, we can let Rust pick whatever in-memory representation makes sense based on the situation. That allows to compiler to optimize memory usage.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -179,10 +198,15 @@ pub struct CosmosRequestBuilder {
 
 #[allow(dead_code)]
 impl CosmosRequestBuilder {
-    pub fn new(operation_type: OperationType, resource_type: ResourceType) -> CosmosRequestBuilder {
+    pub fn new(
+        operation_type: OperationType,
+        resource_type: ResourceType,
+        resource_link: ResourceLink,
```
analogrelay: I believe the `ResourceLink` contains the `ResourceType`, so it's not necessary to pass them in separately.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -275,11 +300,15 @@ mod tests {
     use crate::{constants, PartitionKey};
 
     fn make_base_request(op: OperationType) -> CosmosRequest {
```
analogrelay, line 268: Unused?

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -68,6 +69,7 @@ impl CosmosRequest {
     fn new(
```
analogrelay: *Style suggestion, and this is non-blocking since this is internal code*

With `CosmosRequestBuilder` below, would we expect to ever call `::new()` anywhere other than in the builder? There's _no_ requirement that we have a `::new()` function (Rust has no constructors, `new` is just a convention). The common approach to the builder pattern here would be to let the builder construct the `CosmosRequest` struct itself (with code like `CosmosRequest { operation_type: ..., ...}`) rather than bothering to have a constructor.

The other common pattern here would be to move `CosmosRequestBuilder::new` over to `CosmosRequest::builder`. So, basically, there is a sort of constructor on `CosmosRequest` but instead of being a standard constructor that returns a complete value, it returns a builder that the user can further customize. That may seem counter-intuitive a bit but it means you can write code like this:

```rust
let request = CosmosRequest::builder(...)
    .json(&item)
    .some_other_setting(foo)
    .build();
```

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -97,6 +100,19 @@ impl CosmosRequest {
         }
     }
 
+    pub fn is_read_only_request(&self) -> bool {
```
analogrelay, line 102: Recommend making this a method _on_ `OperationType`, so you can write `operation_type.is_read_only()`.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -97,6 +100,19 @@ impl CosmosRequest {
...
+                | OperationType::Head
+                | OperationType::HeadFeed
+                | OperationType::Query
+                | OperationType::SqlQuery
+                | OperationType::QueryPlan
+        )
+    }
+
     /// Maps the logical `OperationType` to its corresponding HTTP verb.
     pub fn http_method(&self) -> Method {
```
analogrelay, line 107: Same here, I'd recommend putting this method on `OperationType`, since it's more relevant there.

analogrelay (commented):
Partial review. Still working through GEM and Location Cache, but wanted to share some of my initial stylistic feedback

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
kundadebdatta: Thanks for the suggestion here. This is now addressed.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
kundadebdatta: Thanks for the suggestion here. This is now addressed.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
kundadebdatta: Thanks for the suggestion here. This is now addressed.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
kundadebdatta: Thanks for the recommendation here. This is now addressed.

sdk/cosmos/azure_data_cosmos/src/constants.rs
kundadebdatta: Thanks for the suggestion here. I have removed the `#[repr(u16)]` usage.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
kundadebdatta: It's used in `http_method_mapping` test.

sdk/cosmos/azure_data_cosmos/src/constants.rs
kundadebdatta: Thanks for the suggestion here. I have refactored the code.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -80,17 +71,13 @@ impl CosmosPipeline {
...
         context: Context<'_>,
     ) -> azure_core::Result<Response<T>> {
-        cosmos_request.request_context.location_endpoint_to_route =
-            Some(resource_link.url(&self.endpoint));
-
         // Prepare a callback delegate to invoke the http request.
         let sender = move |req: &mut CosmosRequest| {
             let ctx = context.clone();
+            let url = req.resource_link.clone();
             let mut raw_req = req.clone().into_raw_request();
```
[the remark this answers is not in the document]
kundadebdatta: This suggestion doesn't work.

sdk/cosmos/azure_data_cosmos/src/constants.rs
```diff
@@ -40,3 +40,203 @@ pub const SUB_STATUS: HeaderName = HeaderName::from_static("x-ms-substatus");
...
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+#[non_exhaustive]
+pub enum SubStatusCode {
+    TooManyRequests = 429,
+
+    // 400: Bad Request sub-status
+    PartitionKeyMismatch = 1001,
+    CrossPartitionQueryNotServable = 1004,
+    ScriptCompileError = 0xFFFF, // From ExecuteStoredProcedure.
```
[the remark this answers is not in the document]
kundadebdatta: NA.

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
```diff
@@ -163,35 +275,96 @@ impl LocationCache {
...
+    ///
+    /// # Returns
+    /// A vector of applicable endpoint URLs
+    pub fn get_applicable_endpoints(&mut self, _request: &CosmosRequest) -> Vec<String> {
+        //TODO: Fix this.
+        self.get_preferred_available_endpoints(
+            &self.locations_info.account_read_endpoints_by_location,
+            RequestOperation::Read,
+            &self.default_endpoint,
+        )
```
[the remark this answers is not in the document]
kundadebdatta: This is a TODO ite and still in list. Fixed it just to resolve this comment.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -0,0 +1,1130 @@
...
+    /// with exponential backoff.
+    ///
+    /// # Arguments
+    /// * `response` - The HTTP response received from the service
+    ///
+    /// # Returns
+    /// A `RetryResult` indicating whether to retry and with what delay
+    async fn should_retry_response(&mut self, response: &RawResponse) -> RetryResult {
+        let status_code = response.status();
+        let sub_status_code = get_substatus_code_from_response(&response.clone());
```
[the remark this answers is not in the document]
kundadebdatta: Refactored.

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
```diff
@@ -163,35 +275,96 @@ impl LocationCache {
...
+    ///
+    /// # Summary
+    /// Checks if the account is configured for multi-master writes by verifying that more than
+    /// one write endpoint is available. Returns true only when multiple write regions exist,
+    /// enabling write operations to be distributed across regions.
+    ///
+    /// # Returns
+    /// `true` if multiple write endpoints are available, `false` otherwise
+    pub fn can_use_multiple_write_locations(&self) -> bool {
+        !self.write_endpoints().is_empty() && self.write_endpoints().iter().len() > 1
```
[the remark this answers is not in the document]
kundadebdatta: Addressed.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -17,7 +18,7 @@ pub struct CosmosClientOptions {
     pub client_options: ClientOptions,
     pub application_name: Option<String>,
     pub application_region: Option<String>,
-    pub application_preferred_regions: Option<Vec<String>>,
+    pub application_preferred_regions: Vec<Cow<'static, str>>,
```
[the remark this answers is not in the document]
kundadebdatta: Does not apply.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
```diff
@@ -0,0 +1,614 @@
...
+                .await;
+        }
+
+        // When TTL expires or cache is invalidated, the async block executes and updates location cache
+        let _account_prop = self
+            .account_properties_cache
+            .try_get_with(ACCOUNT_PROPERTIES_KEY, async {
+                // Fetch latest account properties from service
+                let account_properties: AccountProperties =
+                    self.get_database_account().await?.into_body().json()?;
```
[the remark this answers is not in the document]
kundadebdatta: Doesn't apply. We are still printing the entire internal error within `format`

`format!("Failed to fetch account properties: {}", e)`

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -264,19 +253,15 @@ impl ContainerClient {
         options: Option<ItemOptions<'_>>,
     ) -> azure_core::Result<Response<()>> {
         let options = options.clone().unwrap_or_default();
-        let builder = CosmosRequestBuilder::new(OperationType::Create, ResourceType::Items);
+        let builder = CosmosRequest::builder(OperationType::Create, self.items_link.clone());
         let cosmos_request = builder
```
analogrelay: Non-Blocking style suggestion: I'd probably just get rid of the `builder` variable entirely and just do it all in one method chain, unless there's a need for conditionally calling certain builder functions.

There's no need to churn the PR to change this if nothing else is changing though, the code would be functionally identical in either case.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -36,33 +43,133 @@ impl RetryResult {
...
+/// * `Metadata` - Policy for metadata operations (database/container management). Handles
+///   transient metadata failures with endpoint failover but simpler retry logic than data
+///   plane operations.
+///
+/// * `ResourceThrottle` - Policy specifically for handling 429 (TooManyRequests) errors
+///   with exponential backoff. Used when rate limiting is the primary concern.
+#[allow(dead_code)]
+pub enum RetryPolicy {
+    /// Data plane retry policy for document operations with comprehensive failure handling
+    Client(Box<ClientRetryPolicy>),
```
analogrelay: I wouldn't worry about boxing these types unless we need to. Clippy (the Rust linter we run in CI) has a lint to check for very large enum variants (https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant). There's no need to move things to the heap unless we know they're too big for the stack.

If there's some other reason you're boxing them (or perhaps if Clippy did decide these enum variants are too big), feel free to keep it, but generally the only reasons you need to worry about `Box`ing are if you want a recursive data structure (like a linked list, where a node points to the next node) or if you're dealing with Very Large Data

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
```diff
@@ -158,8 +158,14 @@ impl ResourceThrottleRetryPolicy {
...
 
-#[async_trait]
-impl RetryPolicy for ResourceThrottleRetryPolicy {
+impl ResourceThrottleRetryPolicy {
+    /// Called before sending a request to allow policy-specific modifications
+    ///
+    /// This method is invoked immediately before each request is sent (including retries).
+    /// # Arguments
+    /// * `request` - Mutable reference to the HTTP request being sent
+    pub(crate) async fn before_send_request(&mut self, _request: &mut CosmosRequest) {}
```
analogrelay: Now that you're using the enum-dispatch pattern, you can actually remove this function entirely and move the "no-op" behavior up to the `RetryPolicy` enum (it can just have a match arm for this variant that does nothing).

sdk/cosmos/azure_data_cosmos/src/operation_context.rs
```diff
@@ -32,4 +36,23 @@ impl OperationType {
                 | OperationType::HeadFeed
         )
     }
+
+    /// Maps the `OperationType` to its corresponding HTTP verb.
+    pub fn http_method(&self) -> Method {
```
analogrelay: nit: Since `OperationType` derives `Copy`, you can just take `self` by value here:

```suggestion
    pub fn http_method(self) -> Method {
```

When you do that, you'll still be able to write code like this:

```rust
let op = OperationType::HeadFeed;
let method = op.method();
println!("op = {}", op);
println!("method = {}", method);
```

Normally, this would cause a compiler error if `.method()` took `self` by-value, because Rust would **move** the value and `op` would become unusable. However, because `OperationType` derives from `Copy` (because it's trivially small), Rust will actually make a copy of it when it calls `.method()`, and `op` remains usable.

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
```diff
@@ -163,35 +275,96 @@ impl LocationCache {
...
-        if operation == RequestOperation::Write && !self.locations_info.write_endpoints.is_empty() {
-            self.locations_info
-                .write_endpoints
-                .get(location_index)
-                .cloned()
-                .unwrap_or_else(|| self.default_endpoint.clone())
-        } else if operation == RequestOperation::Read
-            && !self.locations_info.read_endpoints.is_empty()
+        let location_index = request.request_context.location_index_to_route.unwrap_or(0) as usize;
+        let mut location_endpoint_to_route = self.default_endpoint.clone();
```
analogrelay: nit: This is always going to clone the `default_endpoint` value, even if you end up finding a regional endpoint.

I see two good solutions:

1. Change all the various places you set `location_endpoint_to_route` to just return the endpoint right away, thus removing the need for the variable altogether
2. Change `location_endpoint_to_route` to be an `Option<...>`, and then at the end return `location_endpoint_to_route.unwrap_or_else(|| self.default_endpoint.clone())`

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
analogrelay: You shouldn't need the `.iter()` to get the length of a slice:


```suggestion
        !self.write_endpoints().is_empty() && self.write_endpoints().len() > 1
```

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -63,141 +64,34 @@ pub struct CosmosRequest {
...
-    /// body.
-    fn new(
-        operation_type: OperationType,
-        resource_type: ResourceType,
-        resource_id: Option<String>,
-        partition_key: Option<PartitionKey>,
-        body: Option<Vec<u8>>,
-        authorization_token_type: AuthorizationTokenType,
-    ) -> Self {
+    pub fn builder(operation_type: OperationType, resource_link: ResourceLink) -> Self {
```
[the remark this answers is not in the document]
kundadebdatta: Thanks! This is now addressed.

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
```diff
@@ -81,18 +119,66 @@ impl LocationCache {
...
+    /// # Summary
+    /// Retrieves a cloned list of read endpoint URLs ordered by preference. Preferred locations
+    /// that are available appear first, followed by unavailable endpoints, then the default
+    /// endpoint if no preferred locations are configured.
+    ///
+    /// # Returns
+    /// A vector of read endpoint URLs
     pub fn read_endpoints(&self) -> Vec<String> {
         self.locations_info.read_endpoints.clone()
     }
```
analogrelay: Ouch, I think I might have missed this in the original code review. We shouldn't need to clone the vector here, that will cause a lot of allocations. Instead, we should return a slice:

```suggestion
   pub fn read_endpoints(&self) -> &[String] {
&self.locations_info.read_endpoints
}
```

That will also make it safe to call `self.read_endpoints()` / `self.write_endpoints()` multiple times, which Copilot was complaining about earlier.

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
```diff
@@ -81,18 +119,66 @@ impl LocationCache {
...
+    /// # Summary
+    /// Retrieves a cloned list of write endpoint URLs ordered by preference. Preferred locations
+    /// that are available appear first, followed by unavailable endpoints, then the default
+    /// endpoint if no preferred locations are configured.
+    ///
+    /// # Returns
+    /// A vector of write endpoint URLs
     pub fn write_endpoints(&self) -> Vec<String> {
         self.locations_info.write_endpoints.clone()
     }
```
analogrelay:
```suggestion
   pub fn write_endpoints(&self) -> &[String] {
&self.locations_info.write_endpoints
}
```

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
```diff
@@ -0,0 +1,614 @@
...
+    default_endpoint: String,
+
+    /// Thread-safe cache of location information including read/write endpoints and availability status
+    location_cache: Arc<Mutex<LocationCache>>,
+
+    /// HTTP pipeline for making requests to the Cosmos DB service
+    pipeline: Pipeline,
+
+    /// Cache for account properties with 600 second TTL to reduce redundant service calls
+    account_properties_cache: Cache<&'static str, AccountProperties>,
```
analogrelay, line 36: We may not need a full `moka::Cache` here if we're just tracking one value, but we can revisit that later. The main reason I introduced `moka` was to handle larger caches of several pieces of data (like metadata for each container)

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
```diff
@@ -0,0 +1,614 @@
...
+    /// # Summary
+    /// Retrieves the primary endpoint URL that was configured during manager initialization.
+    /// This is the main entry point for the Cosmos DB account and is used as a fallback
+    /// when no preferred regional endpoints are available or configured.
+    ///
+    /// # Returns
+    /// The default endpoint URL as a String
+    pub fn get_hub_uri(&self) -> String {
+        self.default_endpoint.clone()
+    }
```
analogrelay: Rust "getters" don't usually have a `get_` prefix. Also, this can just return a `&str` and avoid the clone:


```suggestion
    pub fn hub_uri(&self) -> &str {
        &self.default_endpoint
    }
```

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
```diff
@@ -0,0 +1,614 @@
...
+    /// Returns the count of preferred locations configured for routing.
+    ///
+    /// # Summary
+    /// Retrieves the number of preferred Azure regions that were specified during
+    /// initialization. This count is used by retry policies to determine failover
+    /// behavior and calculate maximum retry attempts across regions.
+    ///
+    /// # Returns
+    /// The number of preferred locations as an i32
+    pub fn preferred_location_count(&self) -> i32 {
```
analogrelay: Lengths should generally be expressed as a `usize` (and if we need to convert them to another integer type, we can do that at the moment we need it).

```suggestion
    pub fn preferred_location_count(&self) -> usize {
```

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
```diff
@@ -0,0 +1,614 @@
...
+                }
+
+                Ok(account_properties)
+            })
+            .await
+            .map_err(|e: Arc<Error>| {
+                Error::with_message(
+                    azure_core::error::ErrorKind::Other,
+                    format!("Failed to fetch account properties: {}", e),
+                )
```
analogrelay: I suggest using `Error::with_error` so we can provide the error value itself. It's fine to convert the `Arc<Error>` into an `Error` here by calling [`unwrap_or_clone`](https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#method.unwrap_or_clone) since we're fairly likely to hold the only instance (making it free), and if we don't, we're in an error path and an extra clone isn't that concerning:


```suggestion
                Error::with_error(
                    azure_core::error::ErrorKind::Other,
                    e.unwrap_or_clone(),
                    format!("Failed to fetch account properties", e),
                )
```

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
```diff
@@ -0,0 +1,614 @@
...
+    /// # Returns
+    /// A HashMap mapping location names to write endpoint URLs
+    #[allow(dead_code)]
+    fn get_available_write_endpoints_by_location(&self) -> HashMap<String, String> {
+        self.location_cache
+            .lock()
+            .unwrap()
+            .locations_info
+            .account_write_endpoints_by_location
+            .clone()
```
analogrelay, line 326: We should look at removing this `clone`, but because we're locking in the location cache, it's not as trivial as just removing it and returning `&HashMap<String, String>`, so we can address that later. The data in the cache is unlikely to change for the time we're using it, so it'd be nice if we can just pass a reference to the cache through, but we need to maintain the lock if we do that.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -0,0 +1,1130 @@
...
+    /// determining request characteristics (read vs write, multi-master support), and
+    /// resolving the target endpoint based on retry context. Handles location-based routing
+    /// directives, including retry attempts that target specific location indices or the hub
+    /// endpoint. Clears previous routing context and configures the request with the
+    /// appropriate endpoint URL for the current attempt.
+    ///
+    /// # Arguments
+    /// * `request` - The mutable request to configure before sending
+    pub(crate) async fn before_send_request(&mut self, request: &mut CosmosRequest) {
+        self.cosmos_request = Some(request.clone());
```
analogrelay: Hmm, that's going to copy the entire request (including body), right? That seems like a lot.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -0,0 +1,1130 @@
...
+    /// * `mark_both_read_and_write_as_unavailable` - Whether to mark the endpoint unavailable for both operations
+    /// * `force_refresh` - Whether to force refresh of the location cache
+    /// * `retry_on_preferred_locations` - Whether to retry on preferred locations first
+    /// * `overwrite_endpoint_discovery` - Whether to bypass endpoint discovery checks
+    ///
+    /// # Returns
+    /// A `RetryResult`:
+    /// - `Retry { after: Duration }` with appropriate delay if retry is allowed
+    /// - `DoNotRetry` if max retry count exceeded or endpoint discovery disabled
+    async fn should_retry_on_endpoint_failure_async(
```
analogrelay: We don't use the `_async` suffix on Rust methods.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
```diff
@@ -0,0 +1,614 @@
...
+    /// service calls. If `force_refresh` is true, invalidates the cache to ensure fresh data.
+    /// The location cache is updated only when new data is fetched (TTL expiry or forced refresh),
+    /// not when serving cached data.
+    ///
+    /// # Arguments
+    /// * `force_refresh` - If true, invalidates cache and forces fresh fetch from service
+    ///
+    /// # Returns
+    /// `Ok(())` if refresh succeeded, `Err` if fetching account properties failed
+    pub async fn refresh_location_async(&self, force_refresh: bool) -> Result<(), Error> {
```
analogrelay: No `_async` suffix needed.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -0,0 +1,1130 @@
...
+    /// appropriate endpoint URL for the current attempt.
+    ///
+    /// # Arguments
+    /// * `request` - The mutable request to configure before sending
+    pub(crate) async fn before_send_request(&mut self, request: &mut CosmosRequest) {
+        self.cosmos_request = Some(request.clone());
+        let _stat = self
+            .global_endpoint_manager
+            .refresh_location_async(false)
+            .await;
```
analogrelay: This ignores any error `refresh_location_async` returns. We need to either propagate the error (using `?`, and you'll need to return `azure_core::Result<()>` from this function, which will likely need further refactorings), or if it truly should be ignored, add a comment.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
analogrelay: Can we store only the values in `CosmosRequest` we need to make retry decisions? I don't mind cloning some data, but cloning the entire request is a lot.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -0,0 +1,1130 @@
...
+        self.is_read_request = request.is_read_only_request();
```
analogrelay: If we do end up holding the whole request, we don't really need this value in a separate field. We can always read it out of the request.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -0,0 +1,1130 @@
...
+                Duration::ZERO
+            }
+        } else {
+            Duration::milliseconds(RETRY_INTERVAL_MS)
+        };
+
+        let _refresh_cache_result = self
+            .global_endpoint_manager
+            .refresh_location_async(force_refresh)
+            .await;
```
analogrelay, line 331: Another lost error here.

analogrelay (commented):
I'm still not sure we're getting the architecture of this right for Rust. There's a lot of data being moved around, which is cheap in .NET/Java, but in Rust it results in a lot of `.clone()` operations, which require copying actual bytes in memory and gets quite slow.

But here are my first set of comments. We _can_ refactor this code to improve it later, so I'm not necessarily going to block merging on figuring out the exact right Rust architecture, but I have this growing sense that we should consider getting back to basics and trying to work out the "Rust-y" way to achieve our underlying goals rather than just how to translate existing logic in to Rust.

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
kundadebdatta: This is already fixed with the last commit.

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
kundadebdatta: Thanks @analogrelay! This is really an useful insight. Always better to use slice, when we can to avoid any unnecessary memory allocation.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
kundadebdatta: Issue is that `Error` doesn't implement `Clone`. Therefore, we simply can't use [unwrap_or_clone](https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#method.unwrap_or_clone) on `Error` type. Instead, I used `try_unwrap` with a fallback to just format the error message.

image

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
kundadebdatta: Refactored.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
kundadebdatta: Refactored.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
kundadebdatta: Refactored.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
kundadebdatta: Refactored.

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
kundadebdatta: Refactored. Ideal, if we can go with the second approach.

sdk/cosmos/azure_data_cosmos/src/operation_context.rs
kundadebdatta: Refactored. Thanks for the background here.

sdk/cosmos/azure_data_cosmos/src/retry_policies/resource_throttle_retry_policy.rs
kundadebdatta: Thanks, I have Refactored the retry policy.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
kundadebdatta: Agree, for now, we can just store the operation type. Because the retry policy is per request model, there is no harm keeping just the operation type as the retry scoped.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
kundadebdatta: This is now refactored.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
kundadebdatta: Refactored just to make it clean.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
kundadebdatta: My recommendation here is to have a `clippy` rule for such cases, so that we can easily avoid such suffix or prefix for any future PR.

Will try to look into this and add a rule if possible.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
analogrelay: Ah yeah, you're right. That's unfortunate. I like your try_unwrap solution!

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
analogrelay: Agreed. I've never created clippy rules before so I don't know how easy it is but it's worth a look!

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
kundadebdatta: Yes, typically the outcome should be ignored, because we want the request to still go through, when the cache fails to refresh due to some intermittent connectivity issue. Added a detailed code comment to explain the scenario.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
kundadebdatta: Yes, typically the outcome should be ignored, because we want the request to still go through, when the cache fails to refresh due to some intermittent connectivity issue. Added a detailed code comment to explain the scenario.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
kundadebdatta: I think it was a clippy suggestion. Refactored this and waiting for the gates to become green.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -117,8 +117,14 @@ impl ClientRetryPolicy {
...
-        // self.cosmos_request = Some(request.clone());
-        let _stat = self.global_endpoint_manager.refresh_location(false).await;
+        // Ideally, any request flow should not be blocked by the outcome of refresh_location.
+        // There can be three possible cases:
+        // a) The refresh_location succeeds when TTL expires.
+        // b) The refresh_location is bypassed when TTL hasn't expired.
+        // c) The refresh_location operation has failed. In the event of a failure,
+        //    the error is logged and the request should not be blocked.
+        // Hence, the outcome of the operation is ignored here.
+        let _refresh_cache_result = self.global_endpoint_manager.refresh_location(false).await;
```
analogrelay: nit: This could just be `_ = `, which tells Rust to discard the result entirely. Technically this is telling Rust to keep the `Result` itself alive until the end of the function (because you're binding it to a name, the `_` prefix is just there to make the warning about unused variables happy). There's no real difference here, but if the `Result` was carrying a large value in memory that needed to be dropped, it would delay dropping the value.

```suggestion
        _ = self.global_endpoint_manager.refresh_location(false).await;
```

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -312,6 +318,13 @@ impl ClientRetryPolicy {
...
         };
 
+        // Ideally, any request flow should not be blocked by the outcome of refresh_location.
+        // There can be three possible cases:
+        // a) The refresh_location succeeds when TTL expires.
+        // b) The refresh_location is bypassed when TTL hasn't expired.
+        // c) The refresh_location operation has failed. In the event of a failure,
+        //    the error is logged and the request should not be blocked.
+        // Hence, the outcome of the operation is ignored here.
         let _refresh_cache_result = self
```
analogrelay: Same here

```suggestion
        _ = self
```

analogrelay (approved):
Small style nit, but it's not critical

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
kundadebdatta: Thanks for the suggestion. I have use `_=` pattern.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
kundadebdatta: Thanks for the suggestion. I have use `_=` pattern.

analogrelay (approved):

PR #3352, opened 2025-11-17
Pass PagerOptions, not Context into pager callback

### LINKED ISSUE #3299: Pass PagerOptions into pager callback
Like @LarryOsterman does in #3279 we should probably pass in the `PagerOptions` and not just the `Context` to better future-proof it.

The only downside is that we're having to (probably) clone more memory for each page, while cloning the `Context` (basically a `Cow`) is cheap comparatively.

LarryOsterman:
Fixes #3299

### DIFF
--- a/sdk/keyvault/azure_security_keyvault_certificates/src/generated/clients/certificate_client.rs
+++ b/sdk/keyvault/azure_security_keyvault_certificates/src/generated/clients/certificate_client.rs
@@ -690,7 +690,7 @@ impl CertificateClient {
         }
         let api_version = self.api_version.clone();
         Ok(Pager::from_callback(
-            move |next_link: PagerState<Url>, options| {
+            move |next_link: PagerState<Url>, pager_options| {
                 let url = match next_link {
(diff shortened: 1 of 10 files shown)


sdk/keyvault/azure_security_keyvault_certificates/src/generated/clients/certificate_client.rs
```diff
@@ -690,7 +690,7 @@ impl CertificateClient {
         }
         let api_version = self.api_version.clone();
         Ok(Pager::from_callback(
-            move |next_link: PagerState<Url>, ctx| {
+            move |next_link: PagerState<Url>, options| {
```
heaths: I assume you've coordinated with @jhendrixMSFT or @antkmsft, or are making the changes yourself already?

LarryOsterman: Yes, PR is outstanding. I need to make one more iteration on this PR because he correctly points out that this is a breaking change.

PR #3360, opened 2025-11-19
[Storage] Add service version enum

vincenttran-msft:
`.tsp`: https://github.com/Azure/azure-rest-api-specs/tree/vincenttran/service_versions

Actual change of adding all service versions in `main.tsp` def:
https://github.com/Azure/azure-rest-api-specs/blob/5abcc5e63ec9e8c685a5155168e4be7fcb80f8e3/specification/storage/Microsoft.BlobStorage/main.tsp#L35-L115


sdk/storage/azure_storage_blob/src/generated/clients/blob_container_client.rs
vincenttran-msft, line 1389: Like for example here, we can still see it getting the "latest" `2025-11-05`, but I am wondering if we need to define the enum elsewhere and then feed that model into `main.tsp`, so that it actually generates a discoverable enum for consumers of the crate.

PR #3368, opened 2025-11-20
Add request and challenge callbacks to BearerTokenAuthorizationPolicy

### LINKED ISSUE #1756: Need to allow customization of when and how auth tokens are fetched
Key Vault will use the `BearTokenCredentialPolicy` but needs to customize some behavior of it, like being able to empty the body of an unauthenticated request to avoid leaking secrets.

See <https://github.com/Azure/azure-sdk-for-rust/pull/1726#pullrequestreview-2243609123> for context.

chlowell:
These enable SDK clients to implement custom request authorization and challenge handling (closes #1756).

New API:

- `OnRequest` trait: defines a callback invoked on each request, responsible for authorizing it before sending
  - Key Vault clients will use this to skip authorization so they can discover auth parameters from a challenge
- `OnChallenge` trait: defines a callback invoked upon receiving an auth challenge. Implementations are responsible for parsing challenges and authorizing requests and telling the policy whether to retry them
- `BearerTokenAuthorizationPolicy` builder methods: `with_on_request()` and `with_on_challenge()` set callbacks for a policy instance
  - when these aren't set, the policy behaves as it does today: it authorizes each request with its credential and scopes, and returns 401 responses to the pipeline instead of trying to handle their challenges
- `Authorizer` trait (sealed): allows callbacks to authorize requests and cache tokens without using the policy's credential directly
  - I moved the authNZ logic from `send()` into a private implementation of this
- `Request::body_mut()`: lets the policy reset a body stream before retrying after a challenge

I put all this in a new module `azure_core::http::policies::auth` and moved `BearerTokenAuthorizationPolicy` there as well.

### DIFF
--- a/sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
+++ b/sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
@@ -72,24 +72,25 @@ impl Policy for BearerTokenAuthorizationPolicy {
         request: &mut Request,
         next: &[Arc<dyn Policy>],
     ) -> PolicyResult {
+        let mut ctx = ctx.to_borrowed();
         self.on_request
-            .on_request(ctx, request, self.authorizer.as_ref())
+            .on_request(&mut ctx, request, self.authorizer.as_ref())
             .await?;
 
-        let mut response = next[0].send(ctx, request, &next[1..]).await?;
+        let mut response = next[0].send(&ctx, request, &next[1..]).await?;
 
         if response.status() == StatusCode::Unauthorized {
             self.authorizer.invalidate_cache().await;
             if let Some(ref callback) = self.on_challenge {
                 if response.headers().get_str(&WWW_AUTHENTICATE).is_ok() {
                     callback
-                        .on_challenge(ctx, request, self.authorizer.as_ref(), response.headers())
+                        .on_challenge(&ctx, request, self.authorizer.as_ref(), response.headers())
                         .await?;
                     #[cfg(not(target_arch = "wasm32"))]
                     if let SeekableStream(stream) = request.body_mut() {
                         stream.reset().await?;
                     }
-                    response = next[0].send(ctx, request, &next[1..]).await?
+                    response = next[0].send(&ctx, request, &next[1..]).await?
                 }
             }
         }
@@ -141,7 +142,7 @@ pub trait OnRequest: std::fmt::Debug + Send + Sync {
     /// * `authorizer` - Helper used to acquire an access token and set the request's authorization header
     async fn on_request(
         &self,
-        context: &Context,
+        context: &mut Context,
         request: &mut Request,
         authorizer: &dyn Authorizer,
     ) -> Result<()>;
@@ -265,7 +266,7 @@ struct DefaultOnRequest {
 impl OnRequest for DefaultOnRequest {
     async fn on_request(
         &self,
-        context: &Context,
+        context: &mut Context,
         request: &mut Request,
         authorizer: &dyn Authorizer,
     ) -> Result<()> {
@@ -653,7 +654,7 @@ mod tests {
     impl OnRequest for TestOnRequest {
         async fn on_request(
             &self,
-            _context: &Context,
+            _: &mut Context,
             request: &mut Request,
             authorizer: &dyn Authorizer,
         ) -> Result<()> {


sdk/core/azure_core/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 ### Features Added
 
 - Added `continuation_token` to `PagerOptions`.
+- Added extensible request authorization and authentication challenge handling to `BearerTokenAuthorizationPolicy`.
+  - `OnRequest`, `OnChallenge`, and `Authorizer` traits define callbacks for these features.
+  - `with_on_request()` and `with_on_challenge()` builder methods set callbacks for a policy instance.
```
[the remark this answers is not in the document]
heaths: Copilot is right about the `body_mut()` call but, Copilot, you're wrong about the direction of changes reflecting in TSS to AC.

sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
```diff
@@ -59,24 +70,143 @@ impl Policy for BearerTokenAuthorizationPolicy {
...
         next: &[Arc<dyn Policy>],
     ) -> PolicyResult {
-        let access_token = self.access_token.read().await;
+        self.on_request
+            .on_request(ctx, request, self.authorizer.as_ref())
+            .await?;
 
+        let mut response = next[0].send(ctx, request, &next[1..]).await?;
+
+        if response.status() == StatusCode::Unauthorized {
```
LarryOsterman, line 82: Lines 79 to 96 can be simplified using `.then` - it alleviates the need for the `has_challenge` variable at all. 

Something like:
```rust

let response = if let Some(ref on_challenge) = self.on_challenge {
  response.headers
    .get_optional_str(&WWW_AUTHENTICATE)
    .then(|| {
      let should_retry = on_challenge.on_challenge(ctx, request.self.authorizer.as_ref(), response.headers()).await?;
      if should_retry
         :
         :
    }).or_else(||Ok(response))?;
```

Pseudo code written in the UX, so take it with a grain of salt.

sdk/core/azure_core/src/http/policies/bearer_token_policy.rs
```diff
@@ -30,23 +36,28 @@ impl BearerTokenAuthorizationPolicy {
...
         }
     }
 
-    fn scopes(&self) -> Vec<&str> {
-        self.scopes
-            .iter()
-            .map(String::as_str)
-            .collect::<Vec<&str>>()
+    /// Sets a callback for `send` to invoke once on each request it receives, before sending the request. See [`OnRequest`]
+    /// for more details. When not set, the policy authorizes each request using the credential and scopes specified to `new`.
```
heaths: Doc comments in Rust are like (good) git commits: first line is a summary, blank line, more info. So this should probably be:


```suggestion
    /// Sets a callback for `send` to invoke once on each request it receives, before sending the request.
    ///
    /// See [`OnRequest`] for more details. When not set, the policy authorizes each request using the credential and scopes specified to `new`.
```

heaths: The doc comments for `with_on_challenge` are good by comparison.

sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
```diff
@@ -59,24 +70,143 @@ impl Policy for BearerTokenAuthorizationPolicy {
         request: &mut Request,
         next: &[Arc<dyn Policy>],
     ) -> PolicyResult {
-        let access_token = self.access_token.read().await;
+        self.on_request
+            .on_request(ctx, request, self.authorizer.as_ref())
```
heaths: Would this let Key Vault, for example, zero out the request body until it has authenticated? That's what we did in .NET by checking if we had an `Authorization` header or something like that.

sdk/core/azure_core/src/http/policies/bearer_token_policy.rs
```diff
@@ -59,24 +70,143 @@ impl Policy for BearerTokenAuthorizationPolicy {
...
+    /// * `Ok(true)` when the callback handled the challenge and [`BearerTokenAuthorizationPolicy`] should retry the request.
+    /// * `Ok(false)` when the callback can't handle the challenge. [`BearerTokenAuthorizationPolicy`] will return the 401 response to the client in this case.
+    /// * `Err` when an error occurs while handling the challenge.
+    async fn on_challenge(
+        &self,
+        context: &Context,
+        request: &mut Request,
+        authorizer: &dyn Authorizer,
+        headers: &Headers,
+    ) -> Result<bool>;
```
heaths: And a `bool` should always be sufficient? You can't imagine expanding on that such that maybe we should return a result model that is currently just a bool? If you're reasonably certain, I see no problem with this.

sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
```diff
@@ -59,24 +70,143 @@ impl Policy for BearerTokenAuthorizationPolicy {
...
+}
+
+/// Callback [`BearerTokenAuthorizationPolicy`] invokes on every request it receives, before sending the request.
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+pub trait OnRequest: std::fmt::Debug + Send + Sync {
+    /// Invoked once on every [`BearerTokenAuthorizationPolicy::send`] invocation, before the policy sends the request.
+    /// `send` doesn't call this method before retrying a request after an authentication challenge (see [`OnChallenge`]
+    /// for more about challenge handling). Implementations are responsible for authorizing each request via the provided
+    /// [`Authorizer`]. The policy sends the request when this method returns Ok.
```
heaths, line 137: This borderlines on the summary + blank line + description format. Something to consider. Maybe it's fine as-is.

sdk/core/azure_core/src/http/policies/bearer_token_policy.rs
```diff
@@ -59,24 +70,143 @@ impl Policy for BearerTokenAuthorizationPolicy {
...
+        &self,
+        context: &Context,
+        request: &mut Request,
+        authorizer: &dyn Authorizer,
+    ) -> Result<()>;
+}
+
+mod sealed {
+    pub trait Sealed {}
+}
```
heaths: Don't define your own. Just use `crate::private::Sealed`. Ancestors' private members are visible to all descendants.

sdk/core/azure_core/src/http/policies/bearer_token_policy.rs
```diff
@@ -59,24 +70,143 @@ impl Policy for BearerTokenAuthorizationPolicy {
...
+
+/// Helper trait used by [`OnChallenge`] and [`OnRequest`] to authorize requests. This trait is sealed and cannot
+/// be implemented outside of this module.
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+pub trait Authorizer: sealed::Sealed + std::fmt::Debug + Send + Sync {
```
heaths: What's the purpose of this being a trait? You end up having to pass it around as `dyn` (or maybe not; `impl` is better if possible) which is slower to call and this seems like it's in a hot path.

This isn't .NET. We try to avoid traits (similar to .NET's interfaces) as much as possible, unlike .NET. Why not just expose the `BearerTokenAuthorizer`, especially if it's the only thing that would implement this trait. Seems like an unnecessary abstraction.

sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
```diff
@@ -59,24 +70,143 @@ impl Policy for BearerTokenAuthorizationPolicy {
...
+mod sealed {
+    pub trait Sealed {}
+}
+
+/// Helper trait used by [`OnChallenge`] and [`OnRequest`] to authorize requests. This trait is sealed and cannot
+/// be implemented outside of this module.
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+pub trait Authorizer: sealed::Sealed + std::fmt::Debug + Send + Sync {
+    /// Acquire an access token for the provided scopes and options, and set the request's authorization header.
```
LarryOsterman, line 156: While the trait can only be implemented within this module, the `authorize` function should still have a decent doccomment.

sdk/core/azure_core/src/http/policies/bearer_token_policy.rs
chlowell: Yes; attempting to handle a challenge won't produce useful data for the caller. If the callback handled the challenge, the request is ready to go; the policy needs only to send it. If the callback didn't handle the challenge, the policy can't either. And the policy has the same input data--the challenge--as the callback anyway.

But I suppose this could return simply `Result<()>`. Then `Ok` means retry the request, and the policy wraps `Err` with the response and returns that to the pipeline. 🤔 that's simpler and compels the implementer of the callback to write an error message that may be more helpful than the one `check_success` would otherwise produce. What do you think?

sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
chlowell: Yes; `request` is mutable. My prototype for Key Vault has a single type implementing both traits so an instance can hold the challenge request's body. `OnRequest` stashes the body and `OnChallenge` restores it.

sdk/core/azure_core/src/http/policies/bearer_token_policy.rs
chlowell: I believe this could be something like
```rs
pub type Authorize = dyn for<'a> Fn(
        &'a mut Request,
        &'a [&'a str],
        TokenRequestOptions<'a>,
    ) -> Pin<Box<dyn Future<Output = Result<()>> + Send + 'a>>
    + Send
    + Sync
```
instead, if that's preferable. The rationale for the trait aside from making the signature more legible and making it easy for an implementation to hold state is that I expect we'll eventually have at least one more implementation, supporting a different auth scheme for a separate auth policy (`TokenRequestOptions` will capture the differences between schemes).

sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
chlowell: Do you mean it's too brief? What do you suggest adding?

sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
chlowell: It isn't tidier in practice because the signature being `and_then<U, F>(self, f: F) -> Option<U>` means the closure can't use `?`. No problem inlining `has_challenge` though, and this check can be more terse and a little more efficient by calling `get_str` instead of `get_optional_str`

LarryOsterman: The idea is that the `.and_then` closure returns a `Result<T>` (that's why the `Ok(response)`) - you then can use `?` when it's all done.

Ultimately it's not a huge deal - the code appears to be correct. I'm just a fan of chaining primitives when possible - I find the code slightly easier to understand.

sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
LarryOsterman: Arguments mostly - I was thinking of what happens if some service needs its own request/challenge callbacks - they're going to need to access documentation on the `Authorizer`. And since it's a public trait, it could be used by a service client.

sdk/core/azure_core/src/http/policies/bearer_token_policy.rs
heaths: If there's nothing else useful that needs to be captured, a `bool` is fine. But not you're saying that the unit type - no information - might be okay? Then why the `bool` at all? I assumed the `Err` variant would be for any communications error or other unexpected transport or deserialization error.

sdk/core/azure_core/src/http/policies/bearer_token_policy.rs
heaths: Oh, so this basically represents the auth scheme? If so, then I agree a trait makes sense. Just as a single implementation, it wasn't clear why a trait was needed vs. a struct.

sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
heaths: I had a similar thought initially, but I do think this is cleaner. When you start nesting a bunch of funcs, it does get unwieldy pretty quickly especially when they are just one-liners.

sdk/core/azure_core/src/http/policies/bearer_token_policy.rs
chlowell: We don't need the bool. It enables distinguishing "couldn't handle the challenge" from "handling the challenge caused an error" but I doubt that subtlety would be useful. I changed `on_challenge` to return `Result<()>`

heaths (approved):
Rebase approved.

sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
chlowell: I think this sentence captures everything an SDK developer needs to know except perhaps what a "scope" is, which I consider out of scope for our docs

sdk/core/azure_core/src/http/policies/bearer_token_policy.rs
chlowell: Not quite; `Authorizer` lets a caller acquire a token and apply it to a request without directly accessing the policy's credential or understanding all the details of its auth scheme. Its purpose is really to make `OnRequest` and `OnChallenge` easier to implement by hiding those details. What I meant about other auth schemes is that I believe this signature would work for a hypothetical `FooAuthorizationPolicy` as well as the `BearerTokenAuthorizationPolicy` because any data the `Foo` scheme required in addition to scopes would appear as fields on `TokenRequestOptions`.

sdk/keyvault/azure_security_keyvault_certificates/src/generated/clients/certificate_client.rs
```diff
@@ -32,7 +32,7 @@ use azure_core::{
     fmt::SafeDebug,
     http::{
         pager::{PagerResult, PagerState},
-        policies::{BearerTokenAuthorizationPolicy, Policy},
+        policies::{auth::BearerTokenAuthorizationPolicy, Policy},
```
jhendrixMSFT, line 35: We need a mechanism that flags changes like this that the emitter will also need to be updated.

PR #3372, opened 2025-11-22
Eliminating locking and improve safety with Pagers

### LINKED ISSUE #3294: Eliminating locking and improve safety with Pagers, Pollers
Talking with @analogrelay about #3286, she had a good idea that would not only eliminate the need for the `Mutex` but also make the API safer such that `with_continuation_token()` only exists on the `Pager` (and eventually the `Poller`; see #2758) so you can't call it at the wrong time i.e., after the stream has started.

Like `IntoFuture for Poller`, we'd implement `IntoStream for Pager` (and `Poller`, for DX consistency). That would define `continuation_token()` but only the `Pager` itself would define `with_continuation_token()`.

Additionally, because the `Pager` (rather, the `ItemIterator`; we should do the same for `PageIterator` as well for DX consistency) owns the `continuation_token` field and `Option<String>` should also be `Send`, we shouldn't have to lock it in a `Mutex` - just an `Arc<Option<String>>` should suffice.

The only downside is a change to require calling `into_stream()` since there's no desugaring it for now like there is for `std::future::IntoFuture` e.g.,

```rust
let mut pager = client.list_secrets(...).into_stream();
let cont_token = pager.continuation_token();
while let Some(item) = pager.try_next().await? {
    println!("{item:#?}");
}

let mut pager = client.list_secrets(...).with_continuation_token(cont_token).into_stream();
```

The desugaring for `IntoStream` (rather, `IntoAsyncIterator`) could come later, though. See <https://github.com/rust-lang/rust/issues/79024> and its RFC, <https://github.com/rust-lang/rfcs/blob/master/text/2996-async-iterator.md>.

heaths:
Resolves #3294

### DIFF
--- a/sdk/core/azure_core/CHANGELOG.md
+++ b/sdk/core/azure_core/CHANGELOG.md
@@ -13,10 +13,13 @@
 
 ### Breaking Changes
 
-- Changed `Pager::with_callback` to take a `PagerOptions` as the second parameter rather than a `Context` parameter.
+- Added type parameter `C` to `Pager` declaration, defaulting to `Url` so it can be elided in most existing declarations.
+- Changed `Pager::from_callback` to take a `PagerOptions` as the second parameter rather than a `Context` parameter.
+- Changed `Pager::from_callback` to `Pager::new` which now requires the caller to return a `Pin<Box<dyn Future>>`.
 - Moved `BearerTokenAuthorizationPolicy` into `azure_core::http::policies::auth`.
 - Removed `ItemIterator::with_continuation_token()`. Pass a continuation token to `PagerOptions::continuation_token` instead.
 - Removed `PageIterator::with_continuation_token()`. Pass a continuation token to `PagerOptions::continuation_token` instead.
+- Removed `Pager::from_stream`.
 
 ### Bugs Fixed
 


sdk/core/azure_core/src/http/pager.rs
```diff
@@ -594,173 +581,213 @@ impl<P: Page> fmt::Debug for ItemIterator<P> {
...
-        Self {
-            stream: Box::pin(stream),
-            continuation_token: Default::default(),
-        }
-    }
+#[cfg(target_arch = "wasm32")]
+impl<T> ConditionalSend for T {}
 
+// TODO: Rename this.
+pub trait PagePager<P>: futures::Stream<Item = crate::Result<P>> {
```
LarryOsterman: `ContinuablePager`? Because fundamentally this enables the `continuation_token` API?

heaths: Naw, because I don't want to do the ".NET thing" and then have to make that a supertrait of whatever `ItemIterator` will be to expose its unique `into_pages`. Ultimately, the `ItemIterator` and `PageIterator` names I like and want to keep as the trait names. This is more of a placeholder to avoid breaking even more of the code until I'm ready to clean up all the callers. I didn't just rename via the LSP because I knew the construction was going to change anyway, and figured this would produce fewer error lines.

heaths:
I can't make this work. The previous `BoxedStream` solution erased most of the type parameters, but by exposing all those types now, it forces us to use dynamic dispatch which limits a lot of use cases and puts owness on the callers. The goal was eliminate locking but given the network IO, I don't think it's worth the regression to the DX.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -427,121 +461,81 @@ impl<P: Page> ItemIterator<P> {
...
-    /// It does not continue on the next page until you call `next()` after the last item in the current page
-    /// because of how iterators are implemented. This may yield duplicates but will reduce the likelihood of skipping items instead.
-    pub fn continuation_token(&self) -> Option<String> {
-        // Get the continuation_token because that will be used to start over with the current page.
-        self.continuation_token.clone()
+    pub fn into_pages(self) -> PageIterator<P, C> {
+        let mut iter = self.iter;
+
+        // Start with the current page until after all items are iterated.
+        iter.continuation_token = self.continuation_token.clone();
```
analogrelay: We should be able to move the `continuation_token` rather than cloning it, no?

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -310,21 +337,28 @@ type BoxedStream<P> = Box<dyn Stream<Item = crate::Result<P>>>;
...
+where
+    P: Page,
+    C: AsRef<str>,
+{
     #[pin]
-    stream: Pin<BoxedStream<P>>,
-    continuation_token: Option<String>,
-    next_token: Arc<Mutex<Option<String>>>,
+    iter: PageIterator<P, C>,
+    continuation_token: Option<C>,
```
analogrelay, line 345: Does this actually need to hold the `continuation_token` itself? Or could the `continuation_token` functions just call in to `PageIterator` to get the current value?

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -594,19 +606,29 @@ impl<P: Page> fmt::Debug for ItemIterator<P> {
...
+pub struct PageIterator<P, C>
+where
+    C: AsRef<str>,
+{
     #[pin]
-    stream: Pin<BoxedStream<P>>,
-    continuation_token: Arc<Mutex<Option<String>>>,
+    make_request: PagerFn<P, C>,
+    continuation_token: Option<C>,
+    options: PagerOptions<'static, C>,
```
analogrelay: We could just hold the PagerOptions, and update the continuation token there, to avoid cloning. Or, pull the entire PagerOptions apart in the constructor and only store the things we need here.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -691,106 +713,204 @@ impl<P> PageIterator<P> {
...
 {
+    fn is_terminated(&self) -> bool {
+        self.state == State::Done
+    }
+}
+
+enum State<P, C: AsRef<str>> {
     Init,
+    Pending(BoxedFuture<P, C>),
     More(C),
```
analogrelay: We probably don't need to store a copy of the continuation on the `State` anymore. We can just pull it from the fields on `PageIterator` and pass `&C` in to the `make_request` func.

analogrelay: Ah, though you might hit issues because the `make_request` future is probably hanging on to it. This is probably fine then.

sdk/core/azure_core/src/http/pager.rs
heaths: I have that change locally already. Yeah, I noticed that too late.

sdk/core/azure_core/src/http/pager.rs
heaths: This is the *last* continuation token, previously `continuation_token` where the page iterator's stream's `continuation_token` which is `next_token` in the old code. So this is typically one page behind `PageIterator::continuation_token`. I'll add a comment to that effect. I thought about renaming it initially but this seems appropriate given that it's `ItemIterator`'s `continuation_token`.

heaths: Commented added.

sdk/core/azure_core/src/http/pager.rs
heaths: Yep. Also already done in my local changes. Some of this was "adapting" the existing code and a paranoid save before the weekend. :smile:

heaths:
@jhendrixMSFT @antkmsft here are the changes necessary for the `Pager`. Basically, we need to `Box::pin(async move { .. })` the async function (IIRC, we did this long ago as well) and for any contination token type that isn't a `Url`, we need to declare it explicitly as I did in Storage e.g., `Pager<T, XmlFormat, String>`. Much like the `F: Format` type parameter, if it's the default we don't need to specify it and it would be ideal not to because it looks cleaner. Most of our code - as I know from many of our service reviews - use URL next links to continuation so it seems a good default.

I imagine it'd be easier for me to push another commit to make these same sort of changes to `Poller` - easier to coordinate anyway. As a separate commit, hopefully that makes it easier to review. The changes to `pager.rs` are extensive, but all the other files are fairly small changes.

To note, I'm not crazy about the `unsafe impl Send` for the iterators, but async trait method implementations when the trait is attributed with `#[async_trait]` seem to require it. Even after adding explicit `Send` (rather, `ConditionalSend`, but also tried just `Send`) to `P` and `C` - which should already have been `Send` anyway - explicitly implementing `Send` the `ItemIterator` and `PageIterator` was necessary for the `CheckpointStore` and `PerfTest` "async traits". Probably something with the rewrite. Should be safe because `P` and `C` were virtually already `Send`, but...odd.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -556,14 +585,43 @@ impl<P: Page> futures::Stream for ItemIterator<P> {
...
+{
+    fn is_terminated(&self) -> bool {
+        self.iter.is_terminated()
+    }
+}
+
+#[allow(
+    unsafe_code,
+    reason = "`P` and `C` are `Send` so the iterator can be as well"
+)]
```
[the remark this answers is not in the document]
heaths: Looks like it is stable, but at least compiles against our MSRV of 1.85.

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -688,109 +757,233 @@ impl<P> PageIterator<P> {
...
 {
+    fn is_terminated(&self) -> bool {
+        self.state == State::Done
+    }
+}
+
+#[allow(
+    unsafe_code,
+    reason = "`P` and `C` are `Send` so the iterator can be as well"
+)]
```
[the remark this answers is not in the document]
heaths: See above.

sdk/core/azure_core/src/http/pager.rs
heaths: I really only need to clone it to send an owned copy to the callback, but I can eliminate the clone for `State::More(C)`. Change incoming. I added an `expect()` to get it from `self.options.continuation_token` because to get `State::More` we should've had it. I added a comment, but should be safe.

sdk/core/azure_core/src/lib.rs
```diff
@@ -36,6 +36,26 @@ pub mod tracing {
...
+#[cfg(target_arch = "wasm32")]
+mod conditional_send {
+    /// Conditionally implements [`Send`] based on the `target_arch`.
+    ///
+    /// This implementation does not require `Send`.
+    pub trait ConditionalSend {}
+
+    impl<T> ConditionalSend for T {}
+}
+
```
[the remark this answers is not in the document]
heaths: This was intentional, and while we did consider offline whether to make it public, we decided against it. It doesn't need to be, and is really just an artifact of wasm32 support for `tokio`, but I'll be revisiting that in #3377 anyway so the need for this might not even survive.

sdk/cosmos/azure_data_cosmos/src/query/executor.rs
```diff
@@ -52,13 +57,14 @@ impl<T: DeserializeOwned + Send + 'static> QueryExecutor<T> {
...
 
     pub fn into_stream(self) -> azure_core::Result<FeedPager<T>> {
-        Ok(FeedPager::from_stream(futures::stream::try_unfold(
-            self,
-            |mut state| async move {
-                let val = state.step().await?;
-                Ok(val.map(|item| (item, state)))
-            },
-        )))
+        // Ok(FeedPager::from_stream(futures::stream::try_unfold(
```
RickWinter, line 60: This is commented out code. Will it come back?

sdk/core/azure_core/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#![forbid(unsafe_code)]
+#![deny(unsafe_code)]
```
RickWinter, line 4: What are you adding that requires unsafe?

RickWinter (approved):

sdk/core/azure_core/src/lib.rs
heaths: `unsafe impl Send` requires it. While it seems the iterators should've had a blanket `Send` already because everything else therein is `Send`, they didn't and we can't compile without it. `Send` and `Sync` explicit implementations require `unsafe`.

sdk/cosmos/azure_data_cosmos/src/query/executor.rs
heaths: See #3413

RickWinter (approved):

chlowell (approved):

heaths:
https://github.com/Azure/typespec-rust/pull/758 looks right, so merging this first.

PR #3396, opened 2025-12-04
Fix servicebus metadata

danieljurek:
Example failure attempting to publish a package that has `publish = false`: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5643808&view=logs&j=b766ebde-1fdb-5f11-1350-46ddc53b23cf&t=7ef29fa0-f532-5653-3473-16dc04608431 

Example success building crate where `publish = false`: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5643777&view=results

### DIFF
--- a/eng/scripts/Pack-Crates.ps1
+++ b/eng/scripts/Pack-Crates.ps1
@@ -57,9 +57,6 @@ function Get-PackagesToBuild() {
 }
 
 function Get-OutputPackageNames($workspacePackages) {
-  $packablePackages = $workspacePackages | Where-Object -Property publish -NE -Value @()
-  $packablePackageNames = $packablePackages.name
-
   $names = @()
   switch ($PsCmdlet.ParameterSetName) {
     'Named' {
@@ -71,12 +68,12 @@ function Get-OutputPackageNames($workspacePackages) {
     }
 
     default {
-      return $packablePackageNames
+      return $workspacePackages.name
     }
   }
 
   foreach ($name in $names) {
-    if (-not $packablePackageNames.Contains($name)) {
+    if (-not $workspacePackages.name.Contains($name)) {
       Write-Error "Package '$name' is not in the workspace or does not publish"
       exit 1
     }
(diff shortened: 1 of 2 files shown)


sdk/servicebus/azure_messaging_servicebus/Cargo.toml
```diff
@@ -14,7 +14,7 @@ documentation = "https://docs.rs/azure_messaging_servicebus"
 keywords = ["sdk", "azure", "messaging", "cloud", "servicebus"]
 categories = ["api-bindings"]
 edition.workspace = true
-publish = false
+publish = true
```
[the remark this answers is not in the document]
danieljurek: Nice. I'll do that.

sdk/servicebus/azure_messaging_servicebus/Cargo.toml
```diff
@@ -14,7 +14,6 @@ documentation = "https://docs.rs/azure_messaging_servicebus"
 keywords = ["sdk", "azure", "messaging", "cloud", "servicebus"]
 categories = ["api-bindings"]
 edition.workspace = true
-publish = false
```
RickWinter, line 17: This should stay. We are not ready to publish this package yet.

heaths: I talked to @danieljurek offline. The algorithm should probably take into consideration if "test" is in the name after word-splitting on `_` or `-` (the latter just to be safe).

eng/scripts/Language-Settings.ps1
```diff
@@ -51,6 +51,11 @@ function Get-AllPackageInfoFromRepo ([string] $ServiceDirectory) {
     | Select-Object -ExpandProperty packages
     | Where-Object { $_.manifest_path.StartsWith($searchPath) -and $null -eq $_.publish }
 
+    if (!$packages) {
+      LogError "No publishable packages found in service directory: $ServiceDirectory"
```
heaths: Why an error as opposed to just bailing early? Is the idea that if something runs this pipeline and nothing publishes, any workflow should fail? That doesn't feel right.

danieljurek: LogError just does a Write-host with red text. Now that I think about it, I should put a LogWarning instead (not that it matters much) and let the calling code handle the situation. 

In this case, `Save-Package-Properties.ps1` does a Write-Error and exit 1.

danieljurek: LogError just does a Write-host with red text. Now that I think about it, I should put a LogWarning instead (not that it matters much) and let the calling code handle the situation.

In this case, Save-Package-Properties.ps1 does a Write-Error and exit 1.

heaths (approved):

PR #3397, opened 2025-12-04
[Storage] [POC] Required `AccessConditions` on `upload` API BlobClient

vincenttran-msft:
proof of concept

### DIFF
--- a/sdk/storage/azure_storage_blob/src/models/extensions.rs
+++ b/sdk/storage/azure_storage_blob/src/models/extensions.rs
@@ -7,6 +7,67 @@ use crate::models::{
     PageBlobClientCreateOptions, SignedIdentifier, SignedIdentifiers,
 };
 use std::collections::HashMap;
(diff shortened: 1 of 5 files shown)


sdk/storage/azure_storage_blob/src/models/extensions.rs
```diff
@@ -7,6 +7,67 @@ use crate::models::{
...
+        Self {
+            if_match: Some(etag.into()),
+            ..Default::default()
+        }
+    }
+
+    /// Creates empty access conditions (no restrictions).
+    ///
+    /// Alias for `allow_overwrite()`. The operation will succeed whether the blob exists or not.
+    pub fn none() -> Self {
```
vincenttran-msft, line 67: Could get rid of this if we want to keep it explicit explicit, you need to provide something, even if that something means send `None` and match service behavior (ie. `allow_overwrite()`)

Issue #3377, opened 2025-11-25
Eliminate Send conditions on wasm32-unknown-unknown


heaths:
This is a bit of an experiment but will impact the public API.

Currently, we have a lot of conditional `Send` bounds all over our code because the `wasm32-unknown-unknown` target does not implement `Send` for `Future` as required by `tokio`. We want to maintain WASM/WASI support because Rust is a popular language for WASM.

@yoshuawuyts put me onto <https://github.com/bytecodealliance/wstd> which provides a (mostly) `tokio`-compatible executor that supported WASM.

What we should do is:

1. Add an example async runtime using our APIs for `wstd` as @cataggar has done [here](https://github.com/cataggar/wstd/pull/2). Not only will that help test the next steps, but helps them watch for breaking changes.
2. Remove all the `#[cfg(target_arch = "wasm32")]` and similar conditions.
3. Add some `#[wstd::test]` tests (probably in the aforementioned example) to actually test that it works, which we can't do (not without a lot of additional work not funded currently) with `wasm32-unknown-unknown`.

cataggar:
The conditional `Send` is needed to be able to run in the web browser via `wasm32-unknown-unknown`. I think `wstd` is just for WASI, correct?

PR #3402, opened 2025-12-08
Documentation updates

### LINKED ISSUE #2043: Should we rev versions after release or just before release

### LINKED ISSUE #2044: Document all pipeline checks
For all checks like PoliCheck, CredScan, etc, document when they'll run and where in yaml they're defined.

danieljurek:
Fixes #2043 
Fixes #2044

Documents pipeline checks and outlines dependency/version strategy.

### DIFF
--- a/doc/checks.md
+++ b/doc/checks.md
@@ -29,7 +29,7 @@ Documentation: [cargo-fmt](https://doc.rust-lang.org/cargo/commands/cargo-fmt.ht
 Runs in: [Analyze-Code.ps1](https://github.com/Azure/azure-sdk-for-rust/blob/main/eng/scripts/Analyze-Code.ps1)
 
 ```
-cargo fmt --all -- --check
+cargo fmt
 ```
 
 #### Clippy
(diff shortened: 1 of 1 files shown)


doc/checks.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+#### Clippy
+
+Documentation: [cargo-clippy](https://doc.rust-lang.org/cargo/commands/cargo-clippy.html)
+
+Runs in: [Analyze-Code.ps1](https://github.com/Azure/azure-sdk-for-rust/blob/main/eng/scripts/Analyze-Code.ps1)
+
+```
+cargo clippy --target=wasm32-unknown-unknown --workspace --keep-going --no-deps
```
heaths: Why are you documenting all the `--target=wasm32-unknown-unknown` variants? That's not something most people should ever have to worry about, and we may remove as well (see #3377).

doc/checks.md
heaths, line 1: Is `eng/doc/` a precedent? We already have a `doc/` directory this could also go in (as much as I hate the singular "doc" word we've seemed to settle on across repos).

doc/checks.md
danieljurek: I'm documenting the check as it's run in [Analyze-Code.ps1](https://github.com/Azure/azure-sdk-for-rust/blob/main/eng/scripts/Analyze-Code.ps1#L43)

doc/checks.md
danieljurek: Moved

doc/checks.md
heaths: We run it for non-wasm targets as well, which is what ~100% of customers will be needing. Just drop the target argument. It really doesn't matter anyway for the purpose of which checks we run. Heck, the arguments in most cases are needed anyway.

danieljurek: Oh right, we do that a few lines up .

heaths (approved):

PR #3409, opened 2025-12-09
Partitioned Upload

jaschrep-msft:
Builds on #3340 to implement a generalized upload of data.

Creates `test_extensions` module to help clear up unit tests.

### DIFF
--- a/sdk/storage/azure_storage_blob_test/src/lib.rs
+++ b/sdk/storage/azure_storage_blob_test/src/lib.rs
@@ -199,6 +199,7 @@ impl BodyTestExt for Body {
     async fn collect_bytes(&mut self) -> azure_core::Result<Bytes> {
         match self {
             Body::Bytes(bytes) => Ok(bytes.clone()),
+            #[cfg(not(target_arch = "wasm32"))]
             Body::SeekableStream(seekable_stream) => {
                 seekable_stream.reset().await?;
                 let mut bytes = BytesMut::with_capacity(seekable_stream.len());


sdk/storage/azure_storage_blob/src/partitioned_transfer/upload.rs
heaths, line 1: Be sure to add copyright headers to all source files. It's a Microsoft OSS guideline.

sdk/storage/azure_storage_blob/src/partitioned_transfer/upload.rs
```diff
@@ -0,0 +1,344 @@
...
+use azure_core::stream::SeekableStream;
+#[cfg(not(target_arch = "wasm32"))]
+use futures::StreamExt;
+
+#[cfg(not(target_arch = "wasm32"))]
+use crate::streams::partitioned_stream::PartitionedStream;
+
+use super::*;
+
+pub(crate) trait PartitionedUploadBehavior {
```
heaths, line 18: You'll need to add the `async_trait` stuff we do elsewhere for traits with async functions.

sdk/storage/azure_storage_blob/src/lib.rs
```diff
@@ -17,3 +17,6 @@ mod streams;
 pub use clients::*;
 pub use parsers::*;
 pub mod models;
+
+#[cfg(test)]
+mod test_extensions;
```
heaths: Tip: just call the mod "test". That's what we commonly do. It likely won't remain just "extension" traits forever.

jalauzon-msft: We have a test crate, `azure_storage_blob_test` for test helpers and such. Perhaps all this stuff should live there.

Though, tbh, I'm not the biggest fan of the test package :)

sdk/storage/azure_storage_blob/src/partitioned_transfer/upload.rs
```diff
@@ -0,0 +1,349 @@
...
+
+#[async_trait::async_trait]
+pub(crate) trait PartitionedUploadBehavior {
+    async fn transfer_oneshot(&self, content: Body) -> AzureResult<()>;
+    async fn transfer_partition(&self, offset: usize, content: Body) -> AzureResult<()>;
+    async fn initialize(&self, content_len: usize) -> AzureResult<()>;
+    async fn finalize(&self) -> AzureResult<()>;
+}
+
+pub(crate) async fn upload(
```
jalauzon-msft, line 25: I really love how all the abstractions let us have such a clear and simple definition of the algorithm like this. This looks great.

sdk/storage/azure_storage_blob/src/partitioned_transfer/upload.rs
```diff
@@ -0,0 +1,352 @@
...
+    content: Body,
+    parallel: NonZero<usize>,
+    partition_size: NonZero<usize>,
+    client: &impl PartitionedUploadBehavior,
+) -> AzureResult<()> {
+    if content.len() <= partition_size.get() {
```
vincenttran-msft, line 31: super nit: Would we want to catch the 0-size content length here and early exit, or do we just let it trickle to `transfer_oneshot` and if we want that optimization, implement it at the trait implementation level?

jaschrep-msft: I think we actually want to let this through even at trait implementation level. These are going to be the transfers we push callers to use and sometimes a caller wants to put zero-length content.

For situations like an append blob where creating the blob is a separate network call before you can then append your first block, we can optimize to just do the create and then return.

PR #3410, opened 2025-12-09
[Storage] Snapshot and Versioning Blob Support

### LINKED ISSUE #3438: Un-ignore query parameter ordering for record matching
Based on some changes we made for building query parameters less...verbosely...the order of query parameters changed which broke recorded tests.

I've registered a default matcher to ignore query parameter ordering for now, but we should remove that globally once everything is re-recorded to just stay as close to the default test-proxy configuration as possible - to avoid any language-specific surprises.

This affects just about every service directory, so we'll want to test 'em all.

/cc @LarryOsterman @analogrelay @jaschrep-msft (I suspect Storage might've already noticed with more recent emitter updates so you might be fixed here already)

vincenttran-msft:
`.tsp:` https://github.com/Azure/azure-rest-api-specs/pull/39178

- Generates against corrected emitter code, which can now properly support query parameter replacement (rather than manifesting as a double-append)
- Introduces `with_snapshot()` and `with_version()` convenience methods which also now can properly support query parameter replacement
- Resolves #3438, #3440, and #3441 as we now have a deterministic ordering, and so this PR also pushes a new test recording for the affected tests
- Introduces `tests/blob_client_options.rs` to hold all the options bag heavy non-standard API tests

TODO:
- [x] Changelog


sdk/storage/azure_storage_blob/src/generated/clients/append_blob_client.rs
```diff
@@ -15,9 +15,9 @@ use azure_core::{
     error::CheckSuccessOptions,
     fmt::SafeDebug,
     http::{
-        policies::{auth::BearerTokenAuthorizationPolicy, Policy},
+        policies::{BearerTokenAuthorizationPolicy, Policy},
```
vincenttran-msft: >unresolved import `azure_core::http::policies::BearerTokenAuthorizationPolicy`
consider importing this struct instead:
azure_core::http::policies::auth::BearerTokenAuthorizationPolicy

Not sure why it is now pointing to a different path for the fully qualified name?

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -187,6 +190,52 @@ impl BlobClient {
         &self.client.endpoint
     }
 
+    /// Creates a new BlobClient targeting a specific blob version.
+    ///
+    /// # Arguments
+    ///
+    /// * `version_id` - The version ID of the blob to target.
+    pub fn with_version_id(&self, version_id: &str) -> Result<Self> {
```
jalauzon-msft: Maybe we should call this just `with_version`? Leaving the `version_id` param name though

eng/emitter-package-lock.json
heaths, line 1: No doubt outdated now. Be sure to rebase and/or update these files.

danieljurek:
Benchmarks don't support JSON output. I separated the commands out https://github.com/Azure/azure-sdk-for-rust/pull/3496... This PR should work now

heaths:
> @j7nw4r Please approve when you get a chance, thanks. Content has already been approved but just need your sign-off for a EventHubs regression introduced in a previous PR.

@j7nw4r @LarryOsterman @RickWinter any problem with making Storage co-owners of that *particular* code? Practically, once both Storage and EH GA it shouldn't be a problem - not unless new functionality is used - but could avoid blocks like this...again, if you don't have any concerns.

j7nw4r:
> > @j7nw4r Please approve when you get a chance, thanks. Content has already been approved but just need your sign-off for a EventHubs regression introduced in a previous PR.
> 
> @j7nw4r @LarryOsterman @RickWinter any problem with making Storage co-owners of that _particular_ code? Practically, once both Storage and EH GA it shouldn't be a problem - not unless new functionality is used - but could avoid blocks like this...again, if you don't have any concerns.

I don't have a problem with that.

PR #3411, opened 2025-12-09
Add documentation for Resource Manager generation

RickWinter:


doc/dev/README.md
ronniegeraghty: From the link text it sounds like both docs cover generating Mgmt Plane libraries from TypeSpec. 
Confused on which is the right one. Should the Data & Management plane doc be updated to be just for Data plane? Are there major differences between generating MGMT Plane and Data Plane?

doc/dev/generate.md
```diff
@@ -0,0 +1,94 @@
...
+
+This document complements [Creating a TypeSpec-Based Rust SDK Client](../new-typespec-based-client.md). Read that guide first for the canonical, step-by-step walkthrough; use this page to layer on the management plane specifics that keep our Resource Manager crates consistent with the rest of the Azure SDK.
+
+## When to reach for this guide
+
+You are in the right place if all of the following are true:
+
+-   Your TypeSpec project targets Resource Manager APIs (`import "@azure-tools/typespec-azure-resource-manager";`).
+-   You plan to ship the resulting crate under the `azure_mgmt_*` naming pattern in this repository.
+-   You already understand the base TypeSpec → Rust workflow described in `new-typespec-based-client.md` and only need the management-plane adjustments.
```
ronniegeraghty:
```suggestion
-   You already understand the base TypeSpec → Rust workflow described in [`new-typespec-based-client.md`](../new-typespec-based-client.md) and only need the management-plane adjustments.
```

doc/dev/generate.md
```diff
@@ -0,0 +1,94 @@
...
+    options:
+        "@azure-tools/typespec-rust":
+            emitter-output-dir: "{output-dir}/{service-dir}/{crate-name}"
+            crate-name: "azure_mgmt_<service>"
+            crate-version: "0.1.0"
+        "@azure-tools/typespec-azure-core":
+            azure-resource-provider: "Microsoft.<ProviderName>"
+    ```
+
+-   **Naming** – Follow the `azure_mgmt_<service>` crate pattern and mirror the service folder layout used by other languages (`sdk/<service>/<crate-name>`). When in doubt, look at the Go SDK’s management plane docs for a matching service and port the structure.
```
ronniegeraghty: We should add a link to the Go docs.

doc/dev/generate.md
```diff
@@ -0,0 +1,94 @@
...
+-   **Spec changes** – Capture the `azure-rest-api-specs` commit SHA (or PR branch commit) that contains the TypeSpec updates; you will reference it from `tsp-location.yaml`.
+
+## Repository setup checklist
+
+The base client guide covers initialization in detail. For management plane, double-check these repo-specific tasks once your service directory exists:
+
+-   Add `sdk/<service>/<crate-name>` to the workspace members in the root `Cargo.toml`, but keep it out of `default-members`.
+-   Drop a `ci.yml` into `sdk/<service>/` if the service folder is new so the CI pipeline can pick the crate up. Copy an existing management plane `ci.yml` as the starting point.
```
ronniegeraghty, line 42: We should link to an example management plane `ci.yml` for people to use.

doc/dev/generate.md
```diff
@@ -0,0 +1,94 @@
...
+-   Create `tsp-location.yaml` beside the crate with `repo`, `commit`, and `directory` keys pointing at the Resource Manager TypeSpec project.
```
ronniegeraghty: Is this needed? 
If it's a net new library I thought the generation from the spec repo would create this file.

doc/dev/generate.md
```diff
@@ -0,0 +1,94 @@
...
+# Example update call
+tsp-client update https://github.com/Azure/azure-rest-api-specs/blob/<commit>/specification/<service>/<project>/tspconfig.yaml
+
+# Local iteration
+$spec = "<your-repo-path>/Azure/azure-rest-api-specs/specification/<service>/<project>"
+tsp-client update "$spec/tspconfig.yaml"
+
+```
+
+Generation writes into `src/generated/`; keep the directory read-only and place any ergonomic layers or custom entry points next to `lib.rs`, exactly as described in the shared guide.
```
ronniegeraghty: This information is already in the `new-typespec-based-client.md`. Do we want to copy it here?

doc/dev/generate.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Management-plane finishing work
+
+Most polish items match the data-plane flow, but a few deserve special emphasis for Resource Manager crates:
+
+-   **Cargo metadata** – Ensure `Cargo.toml` advertises the crate as “Azure Resource Manager” in `description` and includes `[lints] workspace = true`.
+-   **Feature gating** – Resource Manager specs often bundle GA and preview APIs. Expose preview-only operations behind an opt-in feature (`cfg(feature = "preview")`). Document the feature in the crate README.
```
ronniegeraghty: Is this done automatically by our emitter? If not, we should try, but also team's will most likely need more information on how to do this.

doc/dev/generate.md
```diff
@@ -0,0 +1,94 @@
...
+-   **Credential surface** – Make sure the public constructors accept `azure_identity` credentials and expose subscription-scoped builders, matching existing management plane crates.
```
ronniegeraghty, line 69: An example might be helpful here.

doc/dev/generate.md
```diff
@@ -0,0 +1,94 @@
...
+-   **Integration tests** – Provision test resources with `eng/common/TestResources/New-TestResources.ps1 -ServiceDirectory <service>` and record with `#[recorded::test]`. Management plane tests frequently require role assignments, so verify the Bicep template grants the test application proper `Microsoft.Authorization/roleAssignments` permissions.
```
ronniegeraghty, line 70: This topic may need it's own doc. 
Do teams even create custom tests for their mgmt plane libraries? I not familiar with this in the mgmt plane space but I feel like this is more involved than what most service teams go through to release mgmt plane libraries.

doc/dev/README.md
```diff
@@ -0,0 +1,6 @@
+# Azure SDK for Rust – Development
+
+This folder collects in-progress developer documentation for contributors building Azure SDK clients, with a focus on management plane generation. Start with the guides below and follow the repository contribution docs for anything not covered here.
+
+-   [Generating Management Plane Crates from TypeSpec](generate.md)
+-   [Creating a TypeSpec-Based Client (Data & Management)](../new-typespec-based-client.md)
```
ronniegeraghty: I would change the order of this list since folks should read the `Creating a TypeSpec-Based Client (Data & Management)` doc first. 

I would also look at making sure there isn't any duplication of information across these docs so we don't have to update the info in two places if it changes. 

The other doc should probably also get a link to the Management Specifics docs letting folks know they should read it and go through the steps if they are working on a MGMT plane library. 

Then write the `Generating Management Plane Creates from TypeSpec` doc assuming folks have already read the other doc, and update the title to something like Management Plane Generation Additional steps, or something like that.

doc/dev/generate.md
RickWinter: I was trying to minimize the jumping between the two docs while also trying to have a simple doc for how to generate.

doc/dev/generate.md
RickWinter: That's a really good idea. @jhendrixMSFT , Does tcgc give us the information to allow us to feature flag all things preview?

doc/dev/generate.md
RickWinter: Teams actually add very little on that front. Its a behavior I'd like to see change as Copilot is getting a lot better at writing the tests. I'll leave it here for now and move once there's more details.

doc/dev/generate.md
jhendrixMSFT: Not directly, we only have the API version info.  We could probably tease this out, assuming the convention is to always use a `-preview` suffix.  Or we ask for a tcgc feature.

We'd likely have to extend this to any types solely used by such operations.

I'll also point out that there are some brownfield services with API versions perpetually in preview but are treated as GA.  This should be rare but makes it ambiguous in such cases (maybe this has been, or will be, fixed).

doc/dev/generate.md
```diff
@@ -0,0 +1,96 @@
...
+# Generating Management Plane Crates from TypeSpec
+
+This document complements [Creating a TypeSpec-Based Rust SDK Client](https://github.com/Azure/azure-sdk-for-rust/blob/main/doc/dev/new-typespec-based-client.md). Read that guide first for the canonical, step-by-step walkthrough; use this page to layer on the management plane specifics that keep our Resource Manager crates consistent with the rest of the Azure SDK.
+
+## When to reach for this guide
+
+You are in the right place if all of the following are true:
+
+-   Your TypeSpec project targets Resource Manager APIs (`import "@azure-tools/typespec-azure-resource-manager";`).
+-   You plan to ship the resulting crate under the `azure_mgmt_*` naming pattern in this repository.
```
heaths: We previously agreed on the name `azure_resourcemanager_*`. It was already codified in the [guidelines](https://azure.github.io/azure-sdk/rust_introduction.html#rust-packaging-prefix), which you should also link here.

doc/dev/generate.md
heaths, line 1: The entirety of the `doc/` directory is already focused on devs. This extra directory isn't just burdensome and should be removed.

doc/dev/generate.md
heaths: This needs to be fixed throughout.

doc/dev/generate.md
```diff
@@ -0,0 +1,96 @@
...
+    ```
+
+-   **Naming** – Follow the `azure_mgmt_<service>` crate pattern and mirror the service folder layout used by other languages (`sdk/<service>/<crate-name>`). When in doubt, look at the Go SDK’s management plane docs for a matching service and port the structure.
+-   **Spec changes** – Capture the `azure-rest-api-specs` commit SHA (or PR branch commit) that contains the TypeSpec updates; you will reference it from `tsp-location.yaml`.
+
+## Repository setup checklist
+
+The base client guide covers initialization in detail. For management plane, double-check these repo-specific tasks once your service directory exists:
+
+-   Add `sdk/<service>/<crate-name>` to the workspace members in the root `Cargo.toml`, but keep it out of `default-members`.
```
heaths: What is this "default-members"? There's no such thing. They have to be in `workspace.members` for anything else to work. If and when we find `cargo` command performance or rust-analyzer performance to be an issue, we'll have to revisit and maybe have separate workspaces for DP and CP like track 1 did.

heaths: It should also be noted that these will go alongside DP crates within the `<service-directory>` e.g., `azure_resourcemanager_keyvault` would go in `sdk/keyvault` with the rest of the `azure_security_keyvault_*` crates.

doc/dev/generate.md
```diff
@@ -0,0 +1,96 @@
...
+
+Use the same commands outlined in the base guide—`tsp-client init` for brand-new crates, `tsp-client update` for refreshes. The only management-plane twist is path selection:
+
+```powershell
+# Example update call
+tsp-client update https://github.com/Azure/azure-rest-api-specs/blob/<commit>/specification/<service>/<project>/tspconfig.yaml
+
+# Local iteration
+$spec = "<your-repo-path>/Azure/azure-rest-api-specs/specification/<service>/<project>"
+tsp-client update "$spec/tspconfig.yaml"
```
heaths: Why do we need to have them create a variable in one instance but not another? Make them consistent. This is just an easy one-liner:


```suggestion
tsp-client update --local-spec-repo "<your-repo-path>/Azure/azure-rest-api-specs/specification/<service>/<project/tspconfig.yaml"
```

You also need the `--local-spec-repo` switch.

doc/dev/generate.md
heaths: You should link to the guide. Make it easy for them to find it: <https://azure.github.io/azure-sdk/rust_implementation.html#rust-client-convenience>

doc/dev/generate.md
```diff
@@ -0,0 +1,96 @@
...
+
+```
+
+Generation writes into `src/generated/`; keep the directory read-only and place any ergonomic layers or custom entry points next to `lib.rs`, exactly as described in the shared guide.
+
+## Management-plane finishing work
+
+Most polish items match the data-plane flow, but a few deserve special emphasis for Resource Manager crates:
+
+-   **Cargo metadata** – Ensure `Cargo.toml` advertises the crate as “Azure Resource Manager” in `description` and includes `[lints] workspace = true`.
```
heaths: This can't be a one-liner:


```suggestion
-   **Cargo metadata** – Ensure `Cargo.toml` advertises the crate as “Azure Resource Manager” in `description` and includes:
    ```toml
    [lints]
    workspace = true`.
    ```
```

That said, as I've told Larry in similar docs, lets not duplicate information. We have tons of precedence in this repo. They can refer to files for an example e.g., `sdk/keyvault/azure_security_keyvault_secrets/Cargo.toml` is a good, simple example and, when used by an LLM for context, is going to provide better context than a (incorrectly formatted) one-liner that's only a small piece of the puzzle.

doc/dev/generate.md
heaths: Wait, if the crate is a preview, why would we have a preview feature? That doesn't make sense.

Also, they can't just use that syntax (which is also wrong). They actually have to define the feature in the `[features]` section of their `Cargo.toml`. The correct syntax - when the feature exists - is `#[cfg(feature = "preview")]`.

doc/dev/generate.md
```diff
@@ -0,0 +1,96 @@
...
+
+After these tweaks, run the usual validation commands (`cargo fmt`, `cargo clippy`, `cargo test`) as instructed in the base guide. When long-running operations are generated, add a `cargo check -p <crate> --all-features` run to ensure every poller compiles.
+
+## Regeneration cadence
+
+When new API versions land:
+
+1. Update the TypeSpec inputs in `azure-rest-api-specs` and grab the merge commit.
+2. Refresh `tsp-location.yaml` with that commit.
+3. Re-run `tsp-client update` and inspect changes outside `generated/` for breaking edits, especially in manual subscription helpers or feature-gated modules.
```
heaths: To note: this has to be run in the directory of `tsp-location.yaml`.

doc/dev/README.md
```diff
@@ -0,0 +1,6 @@
+# Azure SDK for Rust – Development
+
+This folder collects in-progress developer documentation for contributors building Azure SDK clients, with a focus on management plane generation. Start with the guides below and follow the repository contribution docs for anything not covered here.
```
heaths, line 3: Everything under `doc/` is developer docs. No need for yet another directory deep.

heaths (changes requested):

doc/dev/generate.md
RickWinter: Good catch, I'll fix. I had forgotten we made the right call :)

doc/dev/generate.md
RickWinter: I removed this line b/c it isn't useful. I thought it was when I wrote it, but more I looked the less the structure or naming of Go will help.

doc/dev/generate.md
RickWinter: I'll add a link once we have a crate.

doc/dev/generate.md
RickWinter: Added

doc/dev/generate.md
RickWinter: Will add once we have a crate

doc/dev/README.md
RickWinter: All of languages have the dev folder as a way to distinguish between things SDK authors care about vs what a 3P users would care about. Our folder is a blend which I will fix later.

doc/README.md
```diff
@@ -0,0 +1,27 @@
+<!-- cspell: ignore tspconfig mgmt -->
+
+# Azure SDK for Rust Documentation
+
+This directory curates engineering guidance, design notes, and process documentation for contributors to the Azure SDK for Rust. Use these pages to understand repository conventions, service onboarding workflows, and supporting infrastructure.
+
+## Documentation Map
+
+-   [Development Guides](https://github.com/Azure/azure-sdk-for-rust/blob/main/doc/dev/README.md)
+    -   Management-plane generation walkthroughs and TypeSpec automation tips for building `azure_mgmt_*` crates.
```
heaths: Wrong crate name here.
```suggestion
    -   Management-plane generation walkthroughs and TypeSpec automation tips for building `azure_resourcemanager_*` crates.
```

doc/README.md
```diff
@@ -0,0 +1,27 @@
+<!-- cspell: ignore tspconfig mgmt -->
```
heaths: Don't need "mgmt" here either once the crate name is fixed.

doc/dev/generate.md
heaths: You didn't fix this. It's invalid code.

RickWinter: Fixed, I see what you're getting at now. Thx for pointing it out.

doc/dev/generate.md
```diff
@@ -61,7 +61,11 @@ Generation writes into `src/generated/`; keep the directory read-only and place
 
 Most polish items match the data-plane flow, but a few deserve special emphasis for Resource Manager crates:
 
--   **Cargo metadata** – Ensure `Cargo.toml` advertises the crate as “Azure Resource Manager” in `description` and includes `[lints] workspace = true`.
+-   **Cargo metadata** – Ensure `Cargo.toml` advertises the crate as “Azure Resource Manager” in `description` and includes:
+    ```toml
+    [lints]
+    workspace = true`.
```
heaths: Bug: remove the trailing:

```
`.
```

doc/README.md
RickWinter: Fixed

doc/dev/generate.md
RickWinter: I want say who I copied that from ;)

doc/dev/generate.md
RickWinter: I removed this b/c after some more thought, I agree with Heath, they really shouldn't do this.

doc/README.md
```diff
@@ -1,22 +1,20 @@
...
 
--   [Development Guides](https://github.com/Azure/azure-sdk-for-rust/blob/main/doc/dev/README.md)
-    -   Management-plane generation walkthroughs and TypeSpec automation tips for building `azure_mgmt_*` crates.
-    -   Step-by-step instructions for scaffolding new clients, aligning with the Azure SDK design guidelines.
--   [Deprecation Process](https://github.com/Azure/azure-sdk-for-rust/blob/main/doc/deprecation-process.md)
-    -   Policies for deprecating APIs, handling breaking changes, and communicating timeline expectations.
+-   [Consumer Guides](https://github.com/Azure/azure-sdk-for-rust/blob/main/doc/): contains documentation for consumers of SDKs
+-   [Development Guides](https://github.com/Azure/azure-sdk-for-rust/blob/main/doc/dev/README.md): contains advanced documentation for developers of SDKs (not consumers of SDKs)
+
+### Consumer Guides
```
heaths, line 12: When we reorganize, I'd put this first before dev docs, but this is fine for now.

heaths (approved):

PR #3412, opened 2025-12-09
Partitioned download

jaschrep-msft:
Implements partitioned download as a stream that manages multiple downloads and sequences the ranges back together.


sdk/storage/azure_storage_blob/src/models/content_range.rs
```diff
@@ -0,0 +1,169 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jaschrep-msft, line 1: This file pulled from an earlier commit in azure core. They used to supply this type, but it was removed from source code at some point.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -0,0 +1,312 @@
...
+            Some(PollPersist::Ready(output)) => Poll::Ready(Some(Ok(output))),
+            Some(transfer_op) => {
+                ops.push_front(transfer_op);
+                Poll::Pending
+            }
+            None => Poll::Ready(None),
+        }
+    });
+
+    Ok(Box::new(stream))
```
heaths: Probably should be pinned i.e., `Box::pin()`. The caller would have to pin anyway.

sdk/storage/azure_storage_blob/src/models/content_range.rs
jalauzon-msft: Do we expect to expose this to customers as an input type to download? All the places the generated code takes a range, we just take a string and offer a helper to convert. I'm not sure I want to maintain this whole model for something fairly simple. In Python we take a "offset" and "length" and build the string.

jalauzon-msft: I see now that this is for parsing the content range on the initial response? Do we really need a whole struct for this?

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -0,0 +1,311 @@
...
+/// A download that has completed buffering but has not yet returned its buffer in the resulting
+/// stream will still count when determining current running operations. This is so the stream can
+/// promise its buffered bytes do not exceed parallel * partition_size.
+pub(crate) async fn download<'a, T: PartitionedDownloadBehavior>(
+    parallel: usize,
+    partition_size: usize,
+    client: &'a T,
+) -> AzureResult<Pin<Box<dyn Stream<Item = AzureResult<Bytes>> + Unpin + 'a>>> {
+    let partition_size = partition_size as u64;
+    let initial_response = client.transfer_range(0..partition_size).await?;
```
jalauzon-msft, line 35: In other languages we have the concept of the initial download size being larger than the partition size. Is that something we want to include for Rust? I would think there is some perf benefit up to a point where a single shot download is better than partitioned.

jalauzon-msft: Sort of along with this, regardless of initial size, do we want to short circuit this code if we know we are done with the download after the initial request.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -0,0 +1,311 @@
...
+/// Returns a stream that runs up to parallel-many ranged downloads at a time.
+///
+/// Downloads are stored in-order. The returned stream will produce an item only when the next
+/// download in the sequence has been buffered, regardless of the state of any other downloads.
+/// This means completed ranged downloads may sit for a while while earlier ones complete.
+///
+/// A download that has completed buffering but has not yet returned its buffer in the resulting
+/// stream will still count when determining current running operations. This is so the stream can
+/// promise its buffered bytes do not exceed parallel * partition_size.
+pub(crate) async fn download<'a, T: PartitionedDownloadBehavior>(
```
jalauzon-msft, line 28: Do we plan to offer a `DownloadTo` like API where we can take in a stream and seek+write the results to it. This feels like it would be more performant as we don't have to return bytes in order.

sdk/storage/azure_storage_blob/src/models/content_range.rs
jaschrep-msft: I grabbed the whole struct because it already existed in previous commits in azure_core. I didn't write this file.

The majority of this file is parsing logic and testing the parsing logic. We need that code no matter what. Outside of that, I can trim off the fluff if you want, like constructor and getters, but aside from that I like having a type with a name that means something instead of just a tuple of three ints.

jalauzon-msft: It's fine as-is I suppose

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
jaschrep-msft: I didn't plan on it. We can perhaps introduce that later on. I think regardless of whether we introduce such an API, we do need this one. Especially since we plan to eventually hide the raw download range in favor of this (we will need to add range support to this implementation for that as well).

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
jaschrep-msft: > In other languages we have the concept of the initial download size being larger than the partition size. Is that something we want to include for Rust? I would think there is some perf benefit up to a point where a single shot download is better than partitioned.

I don't think this would be hard to add at a later date. I imagine a handful of lines plus testing. Let's get this in and validate its performance first before we add in extra features though.

> Do we want to short circuit this code if we know we are done with the download after the initial request?

Probably! I only recently put `dyn` in the return type, which would allow for this. Definitely looking to get this in before it goes on for too long though.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -0,0 +1,314 @@
...
+    partition_size: usize,
+    client: &'a T,
+) -> AzureResult<Pin<Box<dyn Stream<Item = AzureResult<Bytes>> + Unpin + 'a>>> {
+    let partition_size = partition_size as u64;
+    let initial_response = client.transfer_range(0..partition_size).await?;
+    let content_range: ContentRange = initial_response.headers().get_as(&"content-range".into())?;
+    let total_ranges = content_range.total_length().div_ceil(partition_size);
+
+    let mut ranges =
+        (1..total_ranges).map(move |i| i * partition_size..i * partition_size + partition_size);
```
vincenttran-msft: Copilot suggested that this range calculation would go out of bounds, so just wanted to make a comment here pointing it out:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=358c52ab456443fff74a2b41c22e759f

I am assuming it was written this way since even on the last request if the range is larger than EOF, it will truncate to EOF? Wondering if we should just clamp the final range to be the something like `(<final offset>, content_range.total_length()` so we can get the end range right on the dot?

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -0,0 +1,314 @@
...
+/// Downloads are stored in-order. The returned stream will produce an item only when the next
+/// download in the sequence has been buffered, regardless of the state of any other downloads.
+/// This means completed ranged downloads may sit for a while while earlier ones complete.
+///
+/// A download that has completed buffering but has not yet returned its buffer in the resulting
+/// stream will still count when determining current running operations. This is so the stream can
+/// promise its buffered bytes do not exceed parallel * partition_size.
+pub(crate) async fn download<'a, T: PartitionedDownloadBehavior>(
+    parallel: usize,
+    partition_size: usize,
```
vincenttran-msft: In upload, these are both type `NonZero`.  I am assuming you would want the same here? Tracing through the code it seems like you would get unintended effects if either are 0:

- If `parallel=0`, L58 and follow codeblock will never execute, so it will never saturate the jobs of the rest of the ranges
- if `partition_size=0`, L42 will be `client.transfer_range(0..0).await?;`, not really sure what happens there 😄 but doesn't seem good

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -0,0 +1,314 @@
...
+        for op in ops.iter_mut() {
+            if let PollPersist::Pending(fut) = op {
+                if let Poll::Ready(res) = fut.as_mut().poll(cx) {
+                    match res {
+                        Ok(bytes) => *op = PollPersist::Ready(bytes),
+                        Err(e) => return Poll::Ready(Some(Err(e))),
+                    }
+                }
+            }
+        }
```
vincenttran-msft: Final suggestion that I fielded from Copilot but sounded plausibly good- It suggested you could use `FuturesOrdered` which would remove this logic (and likely the need for `PollPersist` enum) entirely.
The pros it listed for `FuturesOrdered`:

- Wake-based future polling (removes necessity of polling **all** running ops w/ `for op in ops.iter_mut()`)
- Guarantees returned in insertion order (cause ordered)
- Handles all the state transitions
- Removes necessity of polling all download jobs (`removes for op in ops.iter_mut()`)

But if you see any specific reason for doing manual management (ie. granular control of states), or if the refactor is unhelpful, feel free to ignore. But those benefits did sound enticing at the very least 😋

vincenttran-msft (commented):
Other than the Copilot-inspired optimizations, general implementation LGTM. 
But definitely please observe if it needs to be `NonZero`, I think that is my one blocker if the current type `usize` is problematic.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
jaschrep-msft: oh i definitely wish i had known about this type earlier

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -34,57 +26,41 @@ type OpsFuture = Pin<Box<dyn Future<Output = Result<Bytes, azure_core::Error>>>>
...
-        // if the first op is done, return it
-        match ops.pop_front() {
-            Some(PollPersist::Ready(output)) => Poll::Ready(Some(Ok(output))),
-            Some(transfer_op) => {
-                ops.push_front(transfer_op);
-                Poll::Pending
-            }
-            None => Poll::Ready(None),
-        }
+        ops.poll_next_unpin(cx)
```
vincenttran-msft, line 63: 😎 😎 😎

PR #3426, opened 2025-12-11
Set span error attribute after handling auth challenge

chlowell:
PublicApiInstrumentationPolicy normally does this, however it doesn't see the 401 response when the auth policy retries a request.

### DIFF
--- a/sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
+++ b/sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
@@ -293,10 +293,17 @@ mod tests {
             AsyncRawResponse, ClientMethodOptions, Method, Request, StatusCode, Transport,
         },
         time::{Duration, OffsetDateTime},
+        tracing::{SpanKind, TracerProvider},
         Bytes, Result,
     };
     use async_trait::async_trait;
-    use azure_core_test::http::MockHttpClient;
+    use azure_core_test::{
+        http::MockHttpClient,
+        tracing::{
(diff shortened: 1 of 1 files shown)


sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
```diff
@@ -16,10 +16,11 @@ use typespec::{
...
 };
-#[cfg(not(target_arch = "wasm32"))]
-use typespec_client_core::http::Body::SeekableStream;
-use typespec_client_core::http::{ClientMethodOptions, Context, Request};
-use typespec_client_core::time::{Duration, OffsetDateTime};
+use typespec_client_core::{
+    http::{ClientMethodOptions, Context, Request},
+    time::{Duration, OffsetDateTime},
+    tracing::Span,
+};
```
heaths: Nit: all these types are actually re-exported from `azure_core` with the same name. Ideally, I'd like to refer to `typespec*` as little as possible, so you could copy/paste lines 20-22 into the `use crate::{` block and let the formatter merge them. I wouldn't block on this, though. You really only reformatted what was already there. Up to you.

heaths (approved):

sdk/core/azure_core/src/http/policies/auth/bearer_token_policy.rs
chlowell: 👍 it's untidy even apart from that, might as well clean it up now

heaths (approved):

PR #3428, opened 2025-12-11
cosmos: run PR validation against the emulator

### LINKED ISSUE #3443: Cosmos: Enable emulator-based tests during PRs
We currently aren't running emulator-based tests during PR builds. We need to add that to ensure we're continuously validating against the emulator.

analogrelay:
This PR shifts some of our tests so that they are intended to run specifically on the Cosmos DB emulator, rather than being recorded tests using the test proxy. Instead of recorded tests, I've created a `Test-Setup.ps1` that starts the Cosmos DB emulator (I've only done that for Linux right now since that's my primary dev environment, we'll add Windows before merge). I also reworked how cleanup works so that there's a `TestAccount::run` that does the cleanup. This helps ensure that the cleanup is run even if the test fails, **unless** it fails due to a panic. But without an async Drop, we can't do much about panics. All test databases are prefixed `auto-test-` so that should help with bulk cleanup of dangling test DBs.

Tests run on my machine now, but still doing a little fiddling to get them to work on Azure Pipelines. So this may not be passing the PR check yet.

As part of this, I did a few refactorings to help fix the tests:
* I switched all the representations of an "endpoint" to use `url::Url` instead of `String`. That way we parse once and can rely on the fact that `url::Url` validates syntax throughout the rest of the code.
* I noticed a bug in the code that generates the core SDK request. It assumes the location_endpoint includes the path, but it shouldn't. It should just be pointing to the root. So, I made some small fixes to make that more consistent. Now the location_endpoint is always pointing to the root of the account/region and we attach the "path" portion right at the point we create the `azure_core` Request.

Closes #3443


.gitignore
```diff
@@ -7,6 +7,7 @@ Cargo.lock
 
 # User secrets.
 .env
+.envrc
```
analogrelay, line 12: I use `direnv`, which looks for env vars in `.envrc`. Figured it would be good to add it here for safety ;)

eng/scripts/Test-Packages.ps1
```diff
@@ -62,7 +62,7 @@ foreach ($package in $packagesToTest) {
     Invoke-LoggedCommand "cargo test --doc --no-fail-fast" -GroupOutput
     Write-Host "`n`n"
 
-    Invoke-LoggedCommand "cargo test --all-targets --no-fail-fast" -GroupOutput
```
analogrelay, line 65: @heaths as discussed offline, I've put this in my PR for now, since I need it to verify our tests are running anyway.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -115,14 +115,14 @@ impl CosmosRequest {
...
-                .clone(),
-            self.http_method(),
-        );
+        let endpoint = self
+            .request_context
+            .location_endpoint_to_route
+            .as_ref()
+            .unwrap()
+            .clone();
+        let url = self.resource_link.url(&endpoint);
```
analogrelay, line 124: @kundadebdatta There was a bug in the logic that computed the final URL. It assumed `location_endpoint_to_route` was always the final URL to the target endpoint, but in many cases it was just the location-specific root endpoint. I updated the one place in `client_retry_policy.rs` where it DID get set to the resource-specific URL so that it's just the root URL. Doing that means that we can apply the resource path to the URL right here when we're about to create the actual Request.

sdk/cosmos/azure_data_cosmos/src/models/account_properties.rs
```diff
@@ -7,13 +7,16 @@
...
 
 /// Represents a single regional endpoint for the Cosmos DB account (readable or writable).
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
 #[serde(rename_all = "camelCase")]
 pub struct AccountRegion {
     pub name: String,
-    pub database_account_endpoint: String,
+
+    #[serde(with = "crate::serde::url")]
+    pub database_account_endpoint: Url,
```
analogrelay, line 19: This configures `serde` to call the `serialize` and `deserialize` functions found in `serde.rs` when serializing/deserializing this. `Url` isn't serializable by default, so we need to add some logic in to convert it to/from a string before serializing.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -164,10 +164,15 @@ impl ClientRetryPolicy {
...
+            "routed request to endpoint"
+        );
+
         if let Some(ref endpoint) = self.location_endpoint {
-            request.request_context.route_to_location_endpoint(
-                request.resource_link.url(&Url::parse(endpoint).unwrap()),
-            );
+            request
+                .request_context
+                .route_to_location_endpoint(endpoint.clone());
```
analogrelay, line 175: @kundadebdatta see below for context. This is where I noticed that we're sometimes setting the location endpoint to include the path to the resource.

sdk/cosmos/azure_data_cosmos/tests/framework/test_client.rs
analogrelay, line 1: This file contains the logic where we decide whether or not to run the emulator tests, based on the `AZURE_COSMOS_CONNECTION_STRING` env var.

sdk/cosmos/azure_data_cosmos/tests/cosmos_containers.rs
```diff
@@ -14,226 +13,223 @@ use azure_data_cosmos::{
...
-        .create_container(
-            properties.clone(),
-            Some(CreateContainerOptions {
-                throughput: Some(throughput),
+use framework::{test_data, TestClient};
+
+#[tokio::test]
+pub async fn container_crud_simple() -> Result<(), Box<dyn Error>> {
+    TestClient::run_with_db(|_, db_client| {
+        Box::pin(async move {
```
analogrelay: The bodies of these tests should be largely unmodified, something that becomes way more obvious if you enable "Hide Whitespace" in the PR review settings at the top right:

Image

sdk/cosmos/azure_data_cosmos/tests/query_engine.rs
```diff
@@ -5,322 +5,259 @@ mod framework;
...
+            assert_eq!(0, result_items.len());
+
+            Ok(())
+        })
+    })
+    .await
 }
 
-#[recorded::test]
-pub async fn no_query_override_uses_original(
```
analogrelay, line 193: This test was effectively a duplicate of `query_via_query_engine`

eng/scripts/Test-Packages.ps1
```diff
@@ -54,15 +54,24 @@ foreach ($package in $packagesToTest) {
...
 
+    $featuresList = Join-Path $packageDirectory "Test-Features.txt"
+    $featuresArg = ""
+    if (Test-Path $featuresList) {
+      $features = Get-Content $featuresList | Where-Object { $_ -and -not $_.StartsWith("#") } | ForEach-Object { $_.Trim() }
+      if ($features.Count -gt 0) {
+        $featuresArg = "--features " + ($features -join ",")
+      }
+    }
+
```
analogrelay: I know @heaths wants to enable `--all-features`, and I'm certainly happy for us to do that, but I believe there are packages that don't quite support that yet. Until then, this felt like a reasonable compromise. Basically, if there's a `Test-Features.txt` in the package directory (next to `Cargo.toml`), it can contain a newline-delimited list of features to include when testing this package.

sdk/cosmos/azure_data_cosmos/tests/cosmos_items.rs
```diff
@@ -48,641 +43,602 @@ async fn create_container(
     Ok(container_client)
 }
 
-#[recorded::test]
-pub async fn item_create_read_replace_delete(context: TestContext) -> Result<(), Box<dyn Error>> {
```
tvaron3, line 52: Are we removing this test method?

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
tvaron3: Should we add a regression test for this scenario?

sdk/cosmos/azure_data_cosmos/tests/cosmos_items.rs
analogrelay: Yeah, it was a duplicate of `item_create_content_response_on_write`, but the naming isn't clear on that. I'll fix up the naming.

tvaron3 (approved):
lgtm

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
analogrelay: In fact, we already have some, because the emulator tests (which now run in every PR) were broken because of this issue.

eng/scripts/Test-Packages.ps1
heaths: 1. Let's not use mixed case file names. Pain for those of us using case-sensitive file systems.
2. Also, not crazy about files just laying around for this purpose. We already have some precedent for augmenting some pipelines, though I don't think we do for PRs - just CIs. But perhaps @danieljurek or @hallipr has some suggestions.

I won't block on this, but I do want us to revisit this. Dropping bespoke files all over the place just clutters the repo and makes it harder for would-be contributors to onboard.

kundadebdatta (approved):
LGTM

eng/scripts/Test-Packages.ps1
analogrelay: I wanted it to align with Test-Setup/Cleanup but can absolutely change that.

I also entirely agree. I'd like to get rid of this in favour of running tests with all features but we're not able to do that quite yet. I'll file an issue to follow up on that in the new year

Pilchie (approved):

eng/scripts/Test-Packages.ps1
analogrelay: #1746 already tracks this, I'll try to follow up on it when we're all back after the holidays.

PR #3433, opened 2025-12-11
Remove unnecessary unwraps

chlowell:
Our CI clippy is blocking pipelines because of some of these ([for example](https://dev.azure.com/azure-sdk/public/public%20Team/_build/results?buildId=5660899&view=logs&j=36df4387-3551-57fa-098c-b43a23930261&t=6d7a965d-7ab0-5c13-9402-d59ed117d0f6&l=4619))

### DIFF
--- a/sdk/core/azure_core_amqp/src/fe2o3/messaging/message_source.rs
+++ b/sdk/core/azure_core_amqp/src/fe2o3/messaging/message_source.rs
@@ -177,9 +177,13 @@ fn message_source_conversion_fe2o3_amqp() {
         }
     }
 
-    if let (Some(original_default_outcome), Some(round_trip_default_outcome)) =
-        (fe2o3_source.default_outcome, round_trip.default_outcome)
-    {
+    assert_eq!(
+        fe2o3_source.default_outcome.is_some(),
+        round_trip.default_outcome.is_some()
+    );
+
+    if let Some(original_default_outcome) = fe2o3_source.default_outcome {
+        let round_trip_default_outcome = round_trip.default_outcome.unwrap();
         assert_eq!(
             original_default_outcome.is_accepted(),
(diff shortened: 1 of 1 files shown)


sdk/core/azure_core_amqp/src/fe2o3/messaging/message_source.rs
```diff
@@ -202,6 +196,8 @@ fn message_source_conversion_fe2o3_amqp() {
             original_default_outcome.is_modified(),
             round_trip_default_outcome.is_modified(),
         );
+    } else {
+        panic!("Expected default outcomes");
```
heaths: The others look fine, but I really don't understand the issue here. At the very least, I'd expect a change to be more like:

```rust
if let Some(original_default_outcome) = fe2o3_source.default_come {
    let round_trip_default_outcome = round_trip.default_outcome.unwrap();
    // ...
}
```

But adding the panic if `fe2o3.source.default_outcome` is `None` changes the logic here in ways that will crash more often.

heaths: Oh, but this is in a test (that wasn't readily apparent till I "scrolled up"). In that case, if this isn't expected to happen in tests this is fine. I'll leave my comment and leave it up to you, but otherwise approve.

chlowell: ~~Today the test asserts that both Options are Some, so unless I'm misunderstanding something (please clue me in) this change doesn't introduce a new panic condition~~

Edit: 🤣 saw it immediately after submitting this comment: test asserts that both are Some or both are None. I'll push a commit to restore that condition

heaths (approved):

PR #3436, opened 2025-12-11
Move generation file into the dev directory

RickWinter:
Not a complete refactoring. A first step to make the dev folder aligned with other languages.


doc/dev/resourcemanager-generation.md
heaths, line 1: @copilot keep the file names all lowercase as is common. Some of us use filesystems where it matters and pressing Shift is just too much work.

heaths (changes requested):

doc/dev/resourcemanager-generation.md
RickWinter: The casing was me :). I thought captilizing the ResourceManager made sense since that is what we call it. I'm not bound to it so happy to change it to lowercase to help those case sensitive OS junkies

heaths: If you would. It does make things easier, including on contributors. It's rare to see uppercase characters in file names...glaring at you, `Cargo.toml`.

PR #3451, opened 2025-12-12
Use IndexMap instead of HashMap in QueryBuilder

### LINKED ISSUE #3439: Be intentional about query parameter ordering
...and maybe in a few other places where we serialize hashmaps.

We're fully within our rights when we send them in random order - a for a good implementation, the order should not matter.
However, realistically, I think it is a safer bet to be more intentional about the order:

* Naive implementation in Spector fails (and Rust SDK is the only SDK that ran into this)
* Test recording framework did fail at some point
* For Azure or Identity endpoints it should not matter, but I'd preserve some ordering, to be on a safe side - even if Azure is in the wrong, I'd rather not let it affect our customers and also take our resources debugging and then telling the Azure service that they are doing it wrong.
* Actually, for debugging and reproduceability, it is always nice when things are roundtrippable, and have no side effects even if legally the order should not matter.
 
I'd find the most effective way to ensure the order is more intentional than the default hash map does, and do it, preserving either insertion order (ordered_hash_map?), or alphabetical order (BTreeMap), or some mix of both.

antkmsft:
Fixes #3439.
Alternative: #3452.

### DIFF
--- a/.vscode/cspell.json
+++ b/.vscode/cspell.json
@@ -56,6 +56,7 @@
     "execdir",
     "headerless",
     "hmac",
+    "indexmap",
     "iothub",
     "keyvault",
     "lldb",


antkmsft:
If we decide to go with this, TODO: 1) add test for specific order 2) CHANGELOG

PR #3454, opened 2025-12-14
Use SafeDebug for Cosmos model types

### LINKED ISSUE #2585: Cosmos: Derive `SafeDebug` in Cosmos models
Our model types in Cosmos DB all derive `Debug`, because they are hand-written and were created before `SafeDebug` exists. Now we have the `SafeDebug` trait which is designed to redact sensitive information (currently it does so by redacting _all_ information, but will allow us to add a `#[safe]` attribute in the future to unredact known-safe information). Before GA, we should ensure all our models properly derive `SafeDebug` **instead** of `Debug`

cnaples79:
## Summary
- switch Cosmos DB model types to derive SafeDebug instead of Debug
- add SafeDebug imports for the updated Cosmos modules

## Rationale
- issue #2585 asks Cosmos response models to use SafeDebug to avoid leaking sensitive data while retaining debug output

## Testing
- not run (docs/derive-only change)

Fixes #2585

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/models/account_properties.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/models/account_properties.rs
@@ -5,11 +5,12 @@
 //! This is a focused representation for the sample JSON provided; fields not
 //! present in the sample are intentionally omitted for now.
 
+use azure_core::fmt::SafeDebug;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 
 /// Represents a single regional endpoint for the Cosmos DB account (readable or writable).
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[derive(SafeDebug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
 #[serde(rename_all = "camelCase")]
 pub struct AccountRegion {
     pub name: String,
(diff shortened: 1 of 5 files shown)


sdk/cosmos/azure_data_cosmos/src/models/account_properties.rs
```diff
@@ -5,19 +5,20 @@
 //! This is a focused representation for the sample JSON provided; fields not
 //! present in the sample are intentionally omitted for now.
 
+use azure_core::fmt::SafeDebug;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 
 /// Represents a single regional endpoint for the Cosmos DB account (readable or writable).
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[derive(SafeDebug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
```
analogrelay, line 13: Can you also mark all the types with the `#[safe(true)]` attribute so that `SafeDebug` assumes all fields should be **unredacted** by default? I don't believe we have any security-related values here that shouldn't be logged.

analogrelay (changes requested):
This is great! We should also mark all these types as `#[safe(true)]` though

analogrelay:
Hey @cnaples79 , are you interested in wrapping this up, or should we close it out? The remaining work is to mark the types as `#[safe(true)]` so nothing is _actually_ redacted (and we can redact things as needed).

analogrelay:
I'm going to close this and we'll probably move on to implementing this ourselves. Thanks for the PR! Feel free to take a look at other issues and open a PR if you're interested!

Issue #3453, opened 2025-12-13
Cosmos: Build `CollectionCache` to hold Container Metadata Information


kundadebdatta:
### Acceptance Criteria:

Build `CollectionCache` to hold container metadata information.

PR #3460, opened 2025-12-16
Cosmos: Adds `ContainerCache` to Support Container Metadata Caching

kundadebdatta:
## Description:
 
This PR is introducing `ContainerCache` and `AsyncCache` to support the caching layer in the rust SDK. Some of the key design decisions are highlighted below:

- We refactored the [Moka ](https://docs.rs/moka/latest/moka/) based caching library to use a simple `HashMap` based caching wrapped around a async RwLock to handle concurrency.

- The cache refresh has been done on a TTL (time to live) model, where each incoming request has the opportunity to refresh `ContainerCache` or `LocationCache`, if the TTL expires or `force_refresh` is passed as `True`.

- We haven't took the background refresh route yet, since we will need to be a little careful and optimized for the rust bg task management. However, our long term goal is to refresh the cache in the background, so that we do not penalize an incoming request to refresh the cache.

## Closing issues

Fixes Azure/azure-sdk-for-rust#3453

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/routing/mod.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/routing/mod.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-pub mod async_cache;
+pub(crate) mod async_cache;
 pub(crate) mod container_cache;
 pub(crate) mod global_endpoint_manager;
 pub(crate) mod location_cache;


sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
```diff
@@ -0,0 +1,217 @@
...
+                // Remove the entry from the cache.
+                store.remove(&key);
+            } else {
+                // Another task updated it while we waited for the lock
+                return Ok(entry.value.clone());
+            }
+        }
+
+        // Release lock while computing the new value
+        drop(store);
```
analogrelay: This will cause concurrency issues. I think we need to keep the lock held while computing the new value. Since we're using an async-friendly lock that should be OK, all the other callers to `get` will block waiting for the same value to be computed.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
```diff
@@ -0,0 +1,217 @@
...
+
+        // Compute new value
+        let value = compute().await?;
+
+        // Update cache with new value
+        let entry = CacheEntry::new(value.clone(), self.ttl);
+        self.store.write().await.insert(key, entry);
```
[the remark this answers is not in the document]
analogrelay: A per-key mutex might be a good idea though. That way we could hold the HashMap's lock only long enough to get the CacheEntry (held through an `Arc`), which could have a lock over the actual value that we hold when reading/re-computing it.

sdk/cosmos/azure_data_cosmos/src/routing/collection_cache.rs
```diff
@@ -0,0 +1,271 @@
...
+/// across regional endpoints.
+#[derive(Clone, Debug)]
+pub struct CollectionCache {
+    pipeline: Pipeline,
+    global_endpoint_manager: GlobalEndpointManager,
+    container_properties_cache: AsyncCache<String, ContainerProperties>,
+    retry_handler: BackOffRetryHandler,
+}
+
+impl CollectionCache {
```
analogrelay: nit: Let's use `ContainerCache` as the name here. Collection is a legacy name, and I understand why we have code in older SDKs that references it, but let's try to keep to the newer terms here (Item, Container, etc.)

sdk/cosmos/azure_data_cosmos/src/routing/container_cache.rs
```diff
@@ -0,0 +1,271 @@
...
+
+        // Prepare a callback delegate to invoke the http request.
+        let sender = move |req: &mut CosmosRequest| {
+            let mut raw_req = req.clone().into_raw_request();
+            let ctx = ctx_owned.clone();
+            async move { self.pipeline.send(&ctx, &mut raw_req, None).await }
+        };
+
+        // Delegate to the retry handler, providing the sender callback
+        let res = self.retry_handler.send(&mut cosmos_request, sender).await;
```
analogrelay, line 149: Layering and naming of the retry_handler is getting a little messy, but I'll hold on that for now. It's not really just about retry, it's our transport pipeline.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -58,21 +62,19 @@ impl ContainerClient {
...
         options: Option<ReadContainerOptions<'_>>,
-    ) -> azure_core::Result<Response<ContainerProperties>> {
-        let options = options.unwrap_or_default();
-        let cosmos_request =
-            CosmosRequest::builder(OperationType::Read, self.link.clone()).build()?;
-        self.pipeline
-            .send(cosmos_request, options.method_options.context)
-            .await
+    ) -> azure_core::Result<ContainerProperties> {
+        let properties = self
```
analogrelay: I'm not sure if the `read` API should use the cache. If the user is asking us to read the container properties, I think we should actually go to the backend. We can (and probably should) go through the cache so that we can refresh it, but perhaps we should pass through a "force refresh" parameter here to make sure the cache is actually going to the service.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
```diff
@@ -292,31 +293,23 @@ impl GlobalEndpointManager {
...
+            .await;
 
         Ok(())
     }
 
     /// Returns a map of write endpoints indexed by location name.
     ///
     /// # Summary
-    /// Retrieves a mapping from Azure region names to their corresponding write endpoint URLs.
+    /// Retrieves a mapping from Azure region names to their corresponding to write endpoint URLs.
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/routing/collection_cache.rs
```diff
@@ -0,0 +1,271 @@
...
+        let global_endpoint_manager = create_test_endpoint_manager_with_locations();
+        let cache = CollectionCache::new(pipeline, global_endpoint_manager);
+
+        // Verify the cache can be cloned (Debug trait is implemented)
+        let cloned_cache = cache.clone();
+        assert!(std::mem::size_of_val(&cloned_cache) > 0);
+    }
+
+    #[tokio::test]
+    async fn test_remove_by_id_idempotency() {
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::sync::atomic::{AtomicUsize, Ordering};
+
+    #[tokio::test]
+    async fn test_basic_get_and_compute() {
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
```diff
@@ -0,0 +1,217 @@
...
+                Ok::<String, &str>("value2".to_string())
+            })
+            .await
+            .unwrap();
+
+        assert_eq!(value, "value2");
+    }
+
+    #[tokio::test]
+    async fn test_remove() {
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/routing/collection_cache.rs
```diff
@@ -0,0 +1,271 @@
...
+        let endpoint = Url::parse("https://test.documents.azure.com").unwrap();
+        GlobalEndpointManager::new(
+            endpoint,
+            vec![Cow::Borrowed("East US"), Cow::Borrowed("West US")],
+            pipeline,
+        )
+    }
+
+    #[tokio::test]
+    async fn test_remove_by_id() {
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
```diff
@@ -0,0 +1,217 @@
...
+            })
+            .await
+            .unwrap();
+
+        assert_eq!(value, "value1");
+        assert_eq!(compute_count.load(Ordering::SeqCst), 1); // Not incremented
+    }
+
+    #[tokio::test]
+    async fn test_expiration() {
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
```diff
@@ -292,31 +293,23 @@ impl GlobalEndpointManager {
...
     /// This provides direct lookup of write endpoints by location, useful for diagnostic
     /// and monitoring scenarios. The map reflects the current account configuration and
     /// may be empty until account properties are fetched.
     ///
     /// # Returns
-    /// A HashMap mapping location names to write endpoint URLs
+    /// A HashMap containing the location names with their corresponding to write endpoint URLs
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
```diff
@@ -24,19 +25,26 @@ pub struct DatabaseClient {
...
-    pub(crate) fn new(pipeline: Arc<CosmosPipeline>, database_id: &str) -> Self {
+    pub(crate) fn new(
+        pipeline: Arc<CosmosPipeline>,
+        database_id: &str,
+        collection_cache: &CollectionCache,
+    ) -> Self {
         let database_id = database_id.to_string();
         let link = ResourceLink::root(ResourceType::Databases).item(&database_id);
         let containers_link = link.feed(ResourceType::Containers);
+        // collection_cache.set_database_link(link.clone());
```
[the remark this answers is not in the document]
kundadebdatta: Removed.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
```diff
@@ -0,0 +1,217 @@
...
+    store: Arc<RwLock<HashMap<K, CacheEntry<V>>>>,
+    ttl: Duration,
+}
+
+impl<K, V> AsyncCache<K, V>
+where
+    K: Eq + Hash + Clone + Send + Sync + 'static,
+    V: Clone + Send + Sync + 'static,
+{
+    /// Creates a new `AsyncCache` with the specified TTL and expiration callback
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
```diff
@@ -63,10 +64,10 @@ impl GlobalEndpointManager {
             default_endpoint.clone(),
             preferred_locations.clone(),
         )));
-        let account_properties_cache = Cache::builder()
-            .max_capacity(1)
-            .time_to_live(std::time::Duration::from_secs(600))
-            .build();
+
+        let account_properties_cache = AsyncCache::new(
+            Duration::from_secs(600), // Default 5 minutes TTL
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/routing/collection_cache.rs
```diff
@@ -0,0 +1,271 @@
...
+        let cache = CollectionCache::new(pipeline, global_endpoint_manager);
+
+        // Test that remove_by_id doesn't panic when removing non-existent items
+        cache.remove_by_id("non-existent-container").await;
+
+        // Test passes if no panic occurs
+    }
+
+    #[tokio::test]
+    async fn test_new_collection_cache() {
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/routing/collection_cache.rs
```diff
@@ -0,0 +1,271 @@
...
+        let pipeline = create_test_pipeline();
+        let global_endpoint_manager = create_test_endpoint_manager();
+        let cache = CollectionCache::new(pipeline, global_endpoint_manager);
+
+        // Verify the cache was created successfully
+        assert!(std::mem::size_of_val(&cache) > 0);
+    }
+
+    #[tokio::test]
+    async fn test_new_collection_cache_with_preferred_locations() {
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/routing/collection_cache.rs
```diff
@@ -0,0 +1,271 @@
...
+        // Remove the same ID multiple times
+        cache.remove_by_id(container_id).await;
+        cache.remove_by_id(container_id).await;
+        cache.remove_by_id(container_id).await;
+
+        // Test passes if no panic occurs
+    }
+
+    #[tokio::test]
+    async fn test_collection_cache_clone() {
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/routing/collection_cache.rs
```diff
@@ -0,0 +1,271 @@
...
+
+        let cloned_cache = cache.clone();
+
+        // Both should be valid instances
+        cache.remove_by_id("test1").await;
+        cloned_cache.remove_by_id("test2").await;
+    }
+
+    #[tokio::test]
+    async fn test_remove_by_id_with_different_ids() {
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
kundadebdatta: Yes - I already introduced a `force_refresh` within the `async_cache` itself, and as discussed offline, let's keep the API to use the pipeline directly, and we can refactor this later on to use the `container_cache` to fetch the container properties.

sdk/cosmos/azure_data_cosmos/src/routing/collection_cache.rs
kundadebdatta: Renamed it as `ContainerCache`.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
```diff
@@ -0,0 +1,217 @@
...
+    pub fn new(ttl: Duration) -> Self {
+        Self {
+            store: Arc::new(RwLock::new(HashMap::new())),
+            ttl,
+        }
+    }
+
+    /// Gets a value from the cache, or computes it using the provided async function if not present or expired
+    ///
+    /// When the entry is expired, the expiration callback is invoked before computing the new value.
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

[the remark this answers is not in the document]
Pilchie: I don't see an update pushed here, maybe it's still local on your machine?

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
Pilchie: Agree with @analogrelay here.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
```diff
@@ -0,0 +1,285 @@
...
+use std::time::{Duration, Instant};
+
+/// Cache entry with TTL tracking
+#[derive(Clone, Debug)]
+struct CacheEntry<V> {
+    value: V,
+    expires_at: Instant,
+}
+
+impl<V> CacheEntry<V> {
```
Pilchie, line 14: Does `V` also need `Clone()` here?

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
```diff
@@ -0,0 +1,285 @@
...
+        Instant::now() >= self.expires_at
+    }
+}
+
+/// A generic async cache with TTL support and expiration callbacks
+#[derive(Clone)]
+pub struct AsyncCache<K, V>
+where
+    K: Eq + Hash + Clone,
+    V: Clone,
```
Pilchie, line 32: Actually, since we have a func to compute this, why does it need `Clone`?

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
kundadebdatta: Looks like, removing this in the `get` method, would cause a deadlock. After acquiring the write lock on line `77` by calling `self.store.write().await`, if the code doesn't drop it before calling `compute().await` on line `100`, then on line `96`, it will try to acquire the write lock again with `self.store.write().await.insert(key, entry)`, which will cause a deadlock because the lock is still held.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
Pilchie: I see - we `clone()` on the way out.  Should we consider returning a reference instead and putting it on the caller to clone if they want to?  Will defer to @analogrelay's judgement here.

sdk/cosmos/azure_data_cosmos/src/routing/container_cache.rs
```diff
@@ -0,0 +1,272 @@
...
+        let location_endpoint = Some(
+            self.global_endpoint_manager
+                .resolve_service_endpoint(&cosmos_request),
+        );
+
+        if let Some(ref endpoint) = location_endpoint {
+            cosmos_request
+                .request_context
+                .route_to_location_endpoint(cosmos_request.resource_link.url(endpoint));
+        }
```
Pilchie: If you just directly create this in a `Some` and then use an `if let`, why not omit the `Option<T>` altogether.

sdk/cosmos/azure_data_cosmos/src/routing/container_cache.rs
```diff
@@ -0,0 +1,272 @@
...
+    /// * `pipeline` - HTTP pipeline for making requests to Cosmos DB service
+    /// * `global_endpoint_manager` - Manager for multi-region endpoint routing and failover
+    ///
+    /// # Returns
+    /// A new `ContainerCache` instance ready for caching container metadata
+    pub(crate) fn new(pipeline: Pipeline, global_endpoint_manager: GlobalEndpointManager) -> Self {
+        let container_properties_cache = AsyncCache::new(
+            Duration::from_secs(300), // Default 5 minutes TTL
+        );
+        let retry_handler = BackOffRetryHandler::new(global_endpoint_manager.clone());
```
Pilchie, line 47: Probably a separate follow-up, but it seems strange to be cloning the `global_endpoint_manager` instead of just taking a reference here, but maybe async makes the lifetimes hard.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
Pilchie: Wouldn't you just change line 96 to use the existing `store.insert` on the instance that you had already got out of the RwLock?

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
```diff
@@ -0,0 +1,285 @@
...
+        Ok(value)
+    }
+
+    /// Removes a value from the cache
+    /// Returns the removed value if it existed and wasn't expired
+    pub async fn remove(&self, key: &K) -> Option<V> {
+        let mut store = self.store.write().await;
+
+        if let Some(entry) = store.remove(key) {
+            if !entry.is_expired() {
```
tvaron3, line 112: Sorry just for my learning, why wouldn't we remove it when it is expired?

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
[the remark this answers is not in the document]
kundadebdatta: Yes - I missed this commit. Updated the verbaige.

sdk/cosmos/azure_data_cosmos/src/routing/container_cache.rs
kundadebdatta: Thanks for the suggestion here @Pilchie! I agree, may be omitting the `Option<T>` all-together will help to simplify the code. Refactored to reflect the same.

sdk/cosmos/azure_data_cosmos/src/routing/container_cache.rs
```diff
@@ -0,0 +1,268 @@
...
+    ///
+    /// # Arguments
+    /// * `pipeline` - HTTP pipeline for making requests to Cosmos DB service
+    /// * `global_endpoint_manager` - Manager for multi-region endpoint routing and failover
+    ///
+    /// # Returns
+    /// A new `ContainerCache` instance ready for caching container metadata
+    pub(crate) fn new(pipeline: Pipeline, global_endpoint_manager: GlobalEndpointManager) -> Self {
+        let container_properties_cache = AsyncCache::new(
+            Duration::from_secs(300), // Default 5 minutes TTL
```
tvaron3, line 45: Is there a reason we are clearing cache every 5 minutes?

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
```diff
@@ -63,10 +64,10 @@ impl GlobalEndpointManager {
             default_endpoint.clone(),
             preferred_locations.clone(),
         )));
-        let account_properties_cache = Cache::builder()
-            .max_capacity(1)
-            .time_to_live(std::time::Duration::from_secs(600))
-            .build();
+
+        let account_properties_cache = AsyncCache::new(
+            Duration::from_secs(600), // Default 10 minutes TTL
```
tvaron3, line 69: How was this number chosen? Is there any implications with updating less often like any assumptions that were taken from compute or other components that our cache is every 5 minutes? This would also be the recovery time to mark a region as available again.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
kundadebdatta: Yes, seems like just doing `store.insert` without releasing/ dropping the lock should work as expected. Refactored the code to reflect the change.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
analogrelay: No. None of these functions require that `V` is `Clone`. In fact, the `CacheEntry` technically doesn't require that `V` is `Clone` either. Most, if not all, `derive` macros will automatically write their implementation as conditional on all generic type parameters also implementing that trait (`impl<V: Clone> Clone for CacheEntry<V>`)

Now, having said that, any consumer of `CacheEntry<V>` that wants to clone it is going to have to put a `V: Clone` bound, but they'd need that anyway.

sdk/cosmos/azure_data_cosmos/src/routing/container_cache.rs
kundadebdatta: The `global_endpoint_manager` is used within the `container_cache` itself to resolve the gw endpoint. Therefore, a clonned `global_endpoint_manager` was needed to be passed into the `BackOffRetryHandler`.

JFYI - We will do a thorough review of such patterns within the next 2-3 iterations to catch any gaps with any unnecessary usage of the common rust patterns, for e.g. `clone()` etc.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
analogrelay: We can't just return a reference because we're holding the lock on `store`. And even if there wasn't locking here, we wouldn't really want a retained `&V` locking us out from inserting/removing from the cache (because Rust wouldn't allow you to have a `&mut AsyncCache` while you're holding a `&V` from a `get(&AsyncCache)` call). We'd have to wrap up the `RwLock`'s smart pointer in our own smart pointer and return that. Plus, that would mean that the lock stays held until the caller releases its reference.

Cloning is pretty idiomatic in this case. We want to ensure multiple threads can get a copy of the value, and that removing the item from the cache shouldn't free the caller's copy (unlike in a GC-based language where references like this could be safely passed around because there's no single "owner" of any data). If the data is expensive to clone, the user of the `AsyncCache` could easily wrap their value type in `Arc` to make it more "GC-like" and allow multiple people to hold a pointer to the immutable data. That's totally idiomatic as well, but better left to the user of `AsyncCache` to decide

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
analogrelay: No, these functions don't depend on `Clone`, so it's fine. `CacheEntry<V>` will still only be `Clone` if `V: Clone`, but we already enforce that in the trait bounds on `AsyncCache` (which we'd need even if we defined the struct as `CacheEntry<V: Clone>`)

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
```diff
@@ -0,0 +1,282 @@
...
+                    return Ok(entry.value.clone());
+                }
+            }
+        } else {
+            // force_refresh is true, remove existing entry to ensure fresh computation
+            store.remove(&key);
+        }
+
+        // Compute new value
+        let value = compute().await?;
```
analogrelay, line 97: I think we can use per-key locking pretty easily here, but I'll save that for a follow-up refactoring PR.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
```diff
@@ -0,0 +1,282 @@
...
+        // Add entry
+        cache
+            .get("key1".to_string(), false, || async {
+                Ok::<String, &str>("value1".to_string())
+            })
+            .await
+            .unwrap();
+
+        // Wait for expiration
+        tokio::time::sleep(Duration::from_millis(150)).await;
```
analogrelay: There's a more deterministic way to test this. Since these tests are in a sub-module of `async_cache`, they can see private fields/methods on `AsyncCache`. So, in the test, you can reach directly into the HashMap, pull out the CacheEntry and force its expiry time to be in the past. I'd suggest doing that instead of sleeping, which can slow down tests (though not much in this specific case) and become flaky.

sdk/cosmos/azure_data_cosmos/src/routing/container_cache.rs
analogrelay: Yeah, I think we need to do some follow-up refactoring to fix the ownership hierarchy here. A lot of the structures we're porting over from the .NET SDK or other GC-based languages, and they rely on the "free" shared-ownership a GC provides. However, let's get this up and working and then take a pass over it.

The current structure of caches, pipelines, and retry handlers depends a lot on interconnection, which makes it hard to fit into Rust's ownership model. I'm confident there's a structure here that works, but I don't have it clear in my head right now.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
```diff
@@ -34,7 +35,7 @@ pub struct GlobalEndpointManager {
     pipeline: Pipeline,
 
     /// Cache for account properties with 600 second TTL to reduce redundant service calls
-    account_properties_cache: Cache<&'static str, AccountProperties>,
+    account_properties_cache: AsyncCache<&'static str, AccountProperties>,
```
analogrelay, line 38: If there's only one value being cached, an `AsyncCache` is probably overkill, but I have a refactoring in mind that might simplify that.

sdk/cosmos/azure_data_cosmos/src/routing/mod.rs
```diff
@@ -1,5 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+pub mod async_cache;
```
tvaron3: Shouldn't this be the following or else it would be public to customers right ? Just trying to learn, my understanding is that pub(crate) is public inside the crate but not to in other crates.  Just pub is accessible outside of our crate. 

```suggestion
pub(crate) mod async_cache;
```

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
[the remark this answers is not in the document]
kundadebdatta: Discussed offline. This will be taken as a follow-up. Resolving this comment for now.

sdk/cosmos/azure_data_cosmos/src/routing/container_cache.rs
kundadebdatta: These numbers will be re-visited later once we start optimizing things. For now, think of it as a mechanism for evicting the cache.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
kundadebdatta: This number is used to refresh/ evict the cache. In the event of marking the region as unavailable, we do have an option called `force_refresh` to refresh the cache on demand.

sdk/cosmos/azure_data_cosmos/src/routing/mod.rs
kundadebdatta: Correct! Thanks for catching this. You are right, we should mark it visible to internal only. Refactored the code to make this change.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
kundadebdatta: Refactored the test to manupulate the entry to expire deterministically, instead of using `tokio::sleep()`

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
kundadebdatta: Thanks for bringing this up @analogrelay. I will discuss this with you thoroughly, to address it with my next set of PRs.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
kundadebdatta: Thanks for bringing this up @analogrelay. I will discuss this with you thoroughly, to address it with my next set of PRs.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
kundadebdatta: Thanks @tvaron3. The idea is to re-compute the entry for the same key. Removing the key will make sure to add another entry with the re-computed value.

analogrelay (approved):

LarryOsterman (commented):
Fewer dependencies is good :).

PR #3461, opened 2025-12-16
cosmos: Fix URL encoding for authentication signatures

### LINKED ISSUE #3394: `read_item`, `delete_item` returns `Unauthorized` for item IDs containing special characters such as `@`
### Bug Title

_No response_

### Crate Name

azure_data_cosmos

### Crate Version

0.29.0

### Description

I'm experiencing an `Authorization` error when calling `read_item` or `delete_item` on documents whose id contains an `@` character.
In my case the failing ID looks like:
`user@email.com`

The same document can be retrieved successfully using a SQL query (`SELECT * FROM c WHERE c.id = ...`), so the item exists and the permissions are correct. The error happens only for operations that use the `item_id` argument in read_item, delete_item, etc.

This makes me think the issue is related to how the SDK handles special characters in document IDs when constructing or signing the request URL. I suspect that other special characters may trigger the same issue, not just `@`.

Would it be possible to:

1. Validate or properly escape IDs before signing requests, so that IDs containing special characters like @ are handled safely, or
2. At least document the limitation (e.g., which characters require encoding or are unsupported) in the SDK README or API docs, so developers understand the constraints before running into `Unauthorized` errors?

This would be very helpful because document IDs with @ and similar characters are not recommended, but still occur in real systems.

I only found [such references](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/best-practice-dotnet#:~:text=Avoid%20using%20any,any%20unexpected%20behavior.) in Microsoft documentation.

### Steps to Reproduce

1. Create a document in a Cosmos DB container with an ID containing an `@` symbol (for example, `"user@email.com"`).

2. Try reading the document by ID using the Rust SDK:

```rust
let response = container_client
    .read_item::<serde_json::Value>("user@email.com", "user@email.com", None)
    .await?;
```

3. Observe that the SDK returns an `Unauthorized` error.

_Optional:_

4. Verify that the same document can be retrieved via a SQL query such as:

```rust
let mut response = container_client.query_items::<serde_json::Value>(
    format!("SELECT * FROM c WHERE c.id = "user@email.com""),
    "user@email.com",
    None,
)?.into_pages();

while let Some(page_result) = response.next().await {
    let page = page_result?.into_items();
    println!("Result {:?}", page);
}
```

5. Confirm that the error happens for `read_item` / `delete_item` calls where the `item_id` contains special characters, while queries work normally.

### Checklist

analogrelay:
This PR fixes two bugs, with the same root cause: We URL encode paths before signing the request, but we should **not** be doing that.

* We have occasional test failures related to throughput offers when the offer ID has a RID that includes `+`, because we encode it
* #3394 - Whenever an ID has a URL-unsafe character in it, we produce an invalid resource link for signing and the user gets a 401

This PR fixes both by storing both the unencoded and encoded form of path segments in the `ResourceLink`, which get preserved as links are combined (to create sub-clients). In some SDKs, we "unencode" the string before generating the signature payload, but that's challenging because the current URL encoding library we're using (`form_urlencoded`) doesn't actually support decoding. We can improve this later, for now I just want to get something that fixes the immediate issue.

We don't have direct APIs for managing offers, so the easiest way to end-to-end test this was to produce a test that kept trying to create a database until it got one with an offer RID containing `+`. I did that, saw it fail, fixed this issue, and it passed. Since that test is inherently inconsistent, I haven't actually included it.

Closes #3394

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/resource_context.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/resource_context.rs
@@ -140,10 +140,7 @@ impl ResourceLink {
     /// Helper method to create a LinkSegment representing the current path (for use as parent).
     fn path_segment(&self) -> LinkSegment {
         // If this link is RID-based, don't encode the parent path
-        let is_rid_based = self
-            .item_id
-            .as_ref()
-            .map_or(false, |id| id.encoded.is_none());
+        let is_rid_based = self.item_id.as_ref().is_some_and(|id| id.encoded.is_none());
         LinkSegment {
             unencoded: self.unencoded_path(),
             encoded: if is_rid_based {


tvaron3 (approved):
LGTM

simorenoh (approved):

PR #3465, opened 2025-12-18
Add From<Vec<PartitionKeyValue>> for PartitionKey

### LINKED ISSUE #2612: Cosmos: Allow creating a `PartitionKey` from a `Vec`
Currently, a `PartitionKey` can be created from a single item of a supported type (`&str`, numbers, etc.) or a tuple of 2 or 3 items of a supported type.

However, as I'm using the SDK a bit, I think it would be useful to support creating one from a `Vec<T>` (where `T` is something that can be converted to a partition key value). So, we should add an implementation of `From<Vec<PartitionKeyValue>>` for `PartitionKey`.

TheovanKraay:
Adding `From<Vec<PartitionKeyValue>>` for `PartitionKey`

Implements `From<Vec<PartitionKeyValue>>` to enable dynamic partition key construction at runtime.

**Fixes #2612**

### Changes
- Added `From<Vec<PartitionKeyValue>>` trait implementation with validation (max 3 levels)
- 6 unit tests covering empty, single, double, triple, mixed types, and validation
- Integration test demonstrating runtime partition key construction

### Use Case
Enables generic code that works across containers with different partition key schemas without knowing the PK structure at compile time. 

**Example: generic query function**

```rust
use azure_data_cosmos::{PartitionKey, PartitionKeyValue};

// Works with any container schema (1, 2, or 3 level partition keys)
fn query_items_generic(
    container: &ContainerClient,
    key_values: &[&str],
) -> Result<Vec<Item>, Error> {
    // Build partition key dynamically from runtime data
    let pk_values: Vec<PartitionKeyValue> = key_values
        .iter()
        .map(|v| PartitionKeyValue::from(*v))
        .collect();
    
    let partition_key = PartitionKey::from(pk_values);
    
    container
        .query_items(query, partition_key, None)?
        .try_collect()
        .await
}

// Same function works for different partition key definitions:
query_items_generic(&tenant_container, &["tenant1"])?;                    // 1-level
query_items_generic(&region_container, &["tenant1", "us-west"])?;        // 2-level
query_items_generic(&user_container, &["tenant1", "us-west", "user1"])?; // 3-level
```

Previously, this would require compile-time knowledge of key levels etc, e.g.:

```rust
// Required compile-time knowledge of key levels
PartitionKey::from(("tenant1", "region1"))  // 2-level only
```


sdk/cosmos/azure_data_cosmos/tests/partition_key_vec.rs
analogrelay, line 1: I don't think we need an integration test for this scenario. As long as we have unit tests for creating the PartitionKey value, we can rely on the existing integration tests.

analogrelay (changes requested):
Minor change request to remove an unnecessary integration test. Otherwise looks fine!

sdk/cosmos/azure_data_cosmos/tests/partition_key_vec.rs
TheovanKraay: Ok. Motivation was mainly learning how to create them. I can remove it if you prefer.

analogrelay (approved):

analogrelay (approved):
LGTM. It's good to have this for dev-experience, though I think we'd want to encourage most users to use tuples, where we can provide compile-time errors if they use too many items.

PR #3466, opened 2025-12-18
Block blob upload

jaschrep-msft:
Implements partitioned upload for block blobs.

Method currently called `managed_upload()` to avoid naming conflicts.

Introduces a test policy to run arbitrary code in pipeline. Useful for assertions and general observation of network requests.

### DIFF
--- a/sdk/storage/azure_storage_blob/tests/block_blob_client.rs
+++ b/sdk/storage/azure_storage_blob/tests/block_blob_client.rs
@@ -15,10 +15,12 @@ use azure_storage_blob::{
 };
 use azure_storage_blob_test::{
     create_test_blob, get_blob_name, get_container_client, predicates, ClientOptionsExt,
-    StorageAccount, TestPolicy,
+    StorageAccount, TestPolicy, KB, MB,
 };
+use bytes::{BufMut, BytesMut};
 use std::{
     error::Error,
+    io::Write,
     num::NonZero,
     sync::{
         atomic::{AtomicUsize, Ordering},
@@ -318,6 +320,61 @@ async fn managed_upload_empty(ctx: TestContext) -> Result<(), Box<dyn Error>> {
         data
     );
     assert_eq!(request_count.load(Ordering::Relaxed), 1);
+
+    Ok(())
+}
+
+#[recorded::test(live)]
+async fn managed_upload_large(ctx: TestContext) -> Result<(), Box<dyn Error>> {
+    let stage_block_count = Arc::new(AtomicUsize::new(0));
+    let count_policy = Arc::new(TestPolicy::count_requests(
+        stage_block_count.clone(),
+        Some(Arc::new(predicates::is_stage_block_request)),
+    ));
+
+    let recording = ctx.recording();
+    let container_client = get_container_client(
+        recording,
+        true,
+        StorageAccount::Standard,
+        Some(BlobContainerClientOptions::default().with_per_call_policy(count_policy.clone())),
+    )
+    .await?;
+    let blob_client = container_client.blob_client(&get_blob_name(recording));
+    let block_blob_client = blob_client.block_blob_client();
+
+    let data_len = 50 * MB;
+    let expected_stage_block_count = data_len / MB / 4;
+    let mut bytes = BytesMut::with_capacity(data_len).writer();
+    {
+        let mut buf = [0u8; 4 * KB];
+        for _ in (0..data_len).step_by(buf.len()) {
+            buf = recording.random();
+            bytes.write_all(&buf)?;
+        }
+    }
+    let bytes = bytes.into_inner().freeze();
+
+    stage_block_count.store(0, Ordering::Relaxed);
+    {
+        let _scope = count_policy.check_request_scope();
+        block_blob_client
+            .managed_upload(bytes.clone().into(), None)
+            .await?;
+    }
+    assert_eq!(
+        blob_client
+            .download(None)
+            .await?
+            .into_body()
+            .collect()
+            .await?[..],
+        bytes[..]
+    );
+    assert_eq!(
+        stage_block_count.load(Ordering::Relaxed),
+        expected_stage_block_count
+    );
 
     Ok(())
 }
--- a/sdk/storage/azure_storage_blob_test/src/lib.rs
+++ b/sdk/storage/azure_storage_blob_test/src/lib.rs
@@ -27,6 +27,10 @@ use azure_storage_blob::{
 use bytes::BytesMut;
 use futures::{AsyncRead, AsyncReadExt};
 
+pub const KB: usize = 1024;
+pub const MB: usize = KB * 1024;
+pub const GB: usize = MB * 1024;
+
 /// Specifies which storage account to use for testing.
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum StorageAccount {


sdk/storage/azure_storage_blob/src/models/mod.rs
```diff
@@ -77,3 +81,105 @@ pub use crate::generated::models::{
     SignedIdentifiers, SignedIdentifiersHeaders, SkuName, StaticWebsite, StorageErrorCode,
     StorageServiceStats, UserDelegationKey,
 };
+
+/// Options to be passed to `BlockBlobClient::managed_upload()`
+#[derive(Clone, Default, SafeDebug)]
+pub struct BlockBlobClientManagedUploadOptions<'a> {
```
jaschrep-msft: open to putting this somewhere else

sdk/storage/azure_storage_blob/src/client_impl/block_blob_client.rs
```diff
@@ -0,0 +1,185 @@
...
+            blocks: Mutex::new(vec![]),
+        }
+    }
+}
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+impl PartitionedUploadBehavior for BlockBlobClientUploadBehavior<'_, '_> {
+    async fn transfer_oneshot(&self, content: Body) -> AzureResult<()> {
+        let content_len = content.len().try_into().unwrap();
```
jaschrep-msft: @heaths rust-analyzer really wants me to convert `usize` -> `u64` this way. Feels wrong. Can I get a suggestion as to how we might do this instead?

sdk/storage/azure_storage_blob/src/client_impl/block_blob_client.rs
```diff
@@ -0,0 +1,223 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
vincenttran-msft: nit: I am sure we would eventually just drop this in our handwritten `block_blob_client.rs` file and not have it living here? I assume this is just placeholder (alongside having to come up with some naming)

sdk/storage/azure_storage_blob/src/client_impl/block_blob_client.rs
```diff
@@ -0,0 +1,223 @@
...
+const DEFAULT_PARALLEL: NonZero<usize> = NonZero::new(4).unwrap();
+const DEFAULT_PARTITION_SIZE: NonZero<usize> = NonZero::new(4 * 1024 * 1024).unwrap();
+
+// implement this on handwritten client for now
+impl crate::BlockBlobClient {
+    pub async fn managed_upload(
+        &self,
+        body: Body,
+        options: Option<BlockBlobClientManagedUploadOptions<'_>>,
+    ) -> AzureResult<()> {
```
vincenttran-msft: So a couple of things here, not sure the answer to them:

- What will actually be contained in the return type? I assume we will need to come up with our own `BlockBlobClientManagedUploadResult` or something similar, but there is still some implementation details to iron out there? (namely since oneshot would have the regular `upload` response to bubble up, but partitioned transfers have a bunch of `stage_block` responses and a `commit_block_list` response. So I am unsure how our desired return type will take shape. Or maybe I am completely misunderstanding something here? 
- Other than that, we will probably want to match the same shape as our other return types for these client functions, so `Result<Response<BlockBlobClientManagedUploadResult, NoFormat>>`?

sdk/storage/azure_storage_blob/src/client_impl/block_blob_client.rs
```diff
@@ -0,0 +1,223 @@
...
+
+type AzureResult<T> = azure_core::Result<T>;
+
+// unwrap evaluated at compile time
+const DEFAULT_PARALLEL: NonZero<usize> = NonZero::new(4).unwrap();
+const DEFAULT_PARTITION_SIZE: NonZero<usize> = NonZero::new(4 * 1024 * 1024).unwrap();
+
+// implement this on handwritten client for now
+impl crate::BlockBlobClient {
+    pub async fn managed_upload(
```
vincenttran-msft: nit: Once this lands on the generated client, we need to ensure we add a tracer here for OpenTelemetry.
`#[tracing::function("Storage.Blob.BlockBlob.managedUpload")]` or something like that... So that telemetry can differentiate when they hit our managed transfer code vs. manually blocking etc.
edit: fine to just add a TODO here and we'll hit it in our final cleanup PRs (I will be doing quite abit of overall sweeping PRs shortly), since there are some rules and regulations to the naming conventions and might not be worth the effort to figure out rn 😆

sdk/storage/azure_storage_blob/src/client_impl/block_blob_client.rs
vincenttran-msft: `content.len() as u64`? Is there any chance that conversion could fail though? Would be quite spectacular....

sdk/storage/azure_storage_blob/src/models/mod.rs
vincenttran-msft: Yeah this + it's `Results` and `ResultsHeaders` we will need to find a place for our handwritten impls to get all the fixins and place it somewhere.

*for reference, in generated-land:

- `Results` live in `generated/models/pub_models.rs`
- `ResultHeaders` live in `generated/models/header_traits.rs`
- Method options live in `generated/models/method_options.rs`

sdk/storage/azure_storage_blob/tests/block_blob_client.rs
```diff
@@ -201,3 +212,109 @@ async fn test_upload_blob_from_url(ctx: TestContext) -> Result<(), Box<dyn Error
     container_client.delete_container(None).await?;
     Ok(())
 }
+
+#[recorded::test(live)]
```
vincenttran-msft, line 219: As Jacob mentioned, fine to leave as live-only for now but probably should add a TODO so that we eventually flip this into recorded. I would like some checked-in E2E managed tests 😄

sdk/storage/azure_storage_blob/src/client_impl/block_blob_client.rs
vincenttran-msft: Also maybe out of the scope of this PR, but current test coverage doesn't cover any error cases? I guess my thought process is to make sure errors bubble up properly and don't have unexpected behavior (i.e. malformed upload), and we would want to cover:

- Error in one-shot path
- Error in multi-part path

sdk/storage/azure_storage_blob/src/models/mod.rs
jalauzon-msft: Is there anything we can do to not have to hand-write this entire options bag? Can we just have generated option bags inside and then our few specific options? I realize this is tricky because we have multiple different REST APIs we are calling. Leaving this comment more as a discussion point to see if we can brainstorm some other ideas that would allow us to better utilize generated code.

sdk/storage/azure_storage_blob/src/client_impl/block_blob_client.rs
```diff
@@ -0,0 +1,223 @@
...
+
+impl BlockBlobClient {
+    pub async fn managed_upload(
+        &self,
+        body: Body,
+        options: Option<BlockBlobClientManagedUploadOptions<'_>>,
+    ) -> AzureResult<()> {
+        let options = options.unwrap_or_default();
+        let parallel = options.parallel.unwrap_or(DEFAULT_PARALLEL);
+        let partition_size = options.partition_size.unwrap_or(DEFAULT_PARTITION_SIZE);
```
jalauzon-msft: Should we custom implement the default trait for the options bag so that `parallel` and `partition_size` are defaulted properly in the options bag itself? This would mean that those options could be non-optional which would be cleaner IMO.

sdk/storage/azure_storage_blob/src/client_impl/block_blob_client.rs
jalauzon-msft: Discussed offline a bit and here is what Vincent and I think:

- The client impls (both generated and hand-written) should move to the existing `block_blob_client` file in the existing impl blocks.
- The partitioned upload code should also move to that file, just at the bottom. Yes, this will make that file quite long but once we go fully generated, it will become fine.

Feel free to make these changes now or later.

sdk/storage/azure_storage_blob/src/models/mod.rs
jalauzon-msft: Yeah, let's mirror the generated code structure with our hand-written stuff. For now, that means putting this options bag in a `method_options` module.

sdk/storage/azure_storage_blob/tests/block_blob_client.rs
```diff
@@ -201,3 +212,109 @@ async fn test_upload_blob_from_url(ctx: TestContext) -> Result<(), Box<dyn Error
...
+async fn managed_upload(ctx: TestContext) -> Result<(), Box<dyn Error>> {
```
jalauzon-msft, line 220: May be interesting to add a live only test with some bigger sizes and default chunk size just to make sure everything is good in the default case.

sdk/storage/azure_storage_blob/src/client_impl/block_blob_client.rs
jaschrep-msft: It's cleaner but it also results in us being unable to distinguish whether they were left as default or were explicitly set. In other languages, we've chosen to have these be optional types because we want that information to be able to adjust our default logic.

sdk/storage/azure_storage_blob/src/models/mod.rs
jaschrep-msft: I considered keeping the individual options bags instead of their total contents, but that ran into a few problems.

1. There are many duplicate values. Look at all the `.clone()`s in the transfer method to see how many. That's a lot of tedium to leave to the customer.
2. The customer doesn't have much control over which options actually get used. They may provide one set but the code determines it goes down the other path. This can result in several missing options over the wire that the customer is sure should be in there, including custom metadata and access conditions. If we do the expansion, we ensure what the customer provides is actually set.
3. We specifically block some of these options from propagating to the generated options. Currently that's just content validation headers, which we are planning other things around, but there are some things we need to handle specially, and we don't get that option if we just take the raw sub-bags.

sdk/storage/azure_storage_blob/src/client_impl/block_blob_client.rs
jaschrep-msft: In .NET, we have chosen to return an empty response which contains the underlying raw response of whatever came last, which will always be the PUT Blob or PUT Blocklist.

I think that's a reasonable decision. Especially in rust, it's hard to provide anything better typed than that. We could introduce an enum but that's pretty brittle for not much value.

I think we can provide a response of a custom struct that is empty, forcing access to the raw response underneath. If customers are upset with this, we can always deliberate adding new fields to this struct to avoid breaking, or we can implement methods that go into the raw response and attempt to find the value, always returning `Option<T>` of whatever it wants to find.

jalauzon-msft: How do we feel leaving it as-is? As is there is just no response in the successful case, and you would get the error response for whatever API failed in an error case. This is a managed upload anyway, I don't think the customer should care about the response objects in the successful case. Not sure it's worth the complexity to go further? Though it does lock us in to a `()` return forever.

sdk/storage/azure_storage_blob/src/client_impl/block_blob_client.rs
jalauzon-msft: Fair enough, I think it's probably good to retain that information so this is fine.

sdk/storage/azure_storage_blob/src/client_impl/block_blob_client.rs
vincenttran-msft: Happy with as-is here, just as long as we verify the correct error behavior. My bad on this, I should have checked .NETs context more, my feedback was more about getting it-line with the other generated functions, but this really should be treated as a special case instead of trying to shove this square into a hexagon spot 😆

jaschrep-msft: If people are fine with this for now we can circle back on this in the final API review.

sdk/storage/azure_storage_blob/src/models/mod.rs
jalauzon-msft: Yeah, this is a tough one. For now, this should be fine. We can reconsider a bit in the final API review if we end up not liking it.

PR #3470, opened 2025-12-19
[Storage] Azurite HTTP Support PoC

vincenttran-msft:


sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -59,6 +66,34 @@ impl GeneratedBlobContainerClient {
             .per_call_policies
             .push(storage_headers_policy);
 
+        #[cfg(feature = "azurite")]
```
vincenttran-msft, line 69: `azurite` feature flag gate -- this allows us to disable code when users don't explicitly specify azurite.

sdk/storage/azure_storage_blob/src/clients/blob_container_client.rs
```diff
@@ -59,6 +66,34 @@ impl GeneratedBlobContainerClient {
...
+        let per_retry_policies = {
+            let use_azurite = std::env::var("AZURE_STORAGE_USE_AZURITE")
+                .map(|v| v.eq_ignore_ascii_case("true"))
+                .unwrap_or(false);
```
vincenttran-msft, line 73: This is an example implementation that work in the following manner:

- Pre-condition: `azurite` feature flag is enabled
- If enabled:

1. If `AZURE_STORAGE_USE_AZURITE` env. variable is set --> use [well-known shared key credentials](https://learn.microsoft.com/en-us/azure/storage/common/storage-connect-azurite?tabs=blob-storage#use-a-well-known-storage-account-and-key) and this enables Azurite over HTTP
2. If unset, then allow for OAuth like usual (ie. you want to use HTTPS over Azurite)

vincenttran-msft: Alternate implementation approach:
```rust
let use_azurite = if cfg!(feature = "azurite") {
            std::env::var("AZURE_STORAGE_USE_AZURITE")
                .map(|v| v.eq_ignore_ascii_case("true"))
                .unwrap_or(false)
        } else {
            false
        };

        let per_retry_policies = if use_azurite {
            #[cfg(feature = "azurite")]
            {
                use crate::pipeline::SharedKeyPolicy;
                let auth_policy: Arc<dyn Policy> = Arc::new(SharedKeyPolicy::new_azurite()?);
                vec![auth_policy]
            }
            #[cfg(not(feature = "azurite"))]
            {
                Vec::default()
            }
        } else if let Some(token_credential) = credential {
            if !container_url.scheme().starts_with("https") {
                    format!("{container_url} must use https"),
                ));
            }
            let auth_policy: Arc<dyn Policy> = Arc::new(BearerTokenAuthorizationPolicy::new(
                token_credential,
                vec!["https://storage.azure.com/.default"],
            ));
            vec![auth_policy]
        } else {
            Vec::default()
        };
```

This gates the `use_azurite` behind the feature flag and so avoids having to copy the OAuth code for both Azurite/Non-Azurite

Issue #3471, opened 2025-12-19
Multitenant authentication


chlowell:
Add `tenant: String` to `TokenRequestOptions` and update credential implementations to observe it so `get_token()` callers can specify which tenant to request a token from. Note that this option will apply only to credentials that represent possibly multi-tenant principals i.e., not `ManagedIdentityCredential` because managed identities are single-tenant.

PR #3480, opened 2026-01-05
Key Vault clients discover auth parameters

### LINKED ISSUE #3309: Key Vault clients use hardcoded public cloud scope
### Bug Title

_No response_

### Crate Name

azure_security_keyvault_secrets

### Crate Version

0.9.0

### Description

Current secret client has a fixed auth policy restricting the access to Azure Government 

1. create secret on keyvault for Azure Gov
2. Instantiate a client for Azure Goverment cloud 
3. Try to get the secret using get_secret method
4. it fails with the following message: The requested identity hasn’t been assigned to this resource.

The code responsible for the error:

https://github.com/Azure/azure-sdk-for-rust/blob/7583705b45797a8f53d474f1272d0f890a902444/sdk/keyvault/azure_security_keyvault_secrets/src/generated/clients/secret_client.rs#L71-L74

### Steps to Reproduce

1. create secret on keyvault for Azure Gov
2. Instantiate a client for Azure Goverment cloud 
3. Try to get the secret using get_secret method
4. it fails with the following message: The requested identity hasn’t been assigned to this resource.

### Checklist

chlowell:
This adds support for non-Public clouds (closes #3309) and enables multi-tenant scenarios. Implementation follows that of the other SDKs: clients send their first request without authorization to elicit an authentication challenge specifying the audience (scope) and tenant for the Vault (clients use only the audience today; #3471 tracks the work required to use the tenant). After parsing the parameters, clients cache them and authorize and retry the request. We don't have a great way to share code across Key Vault crates, so I've just copied the implementation (authorizer.rs) into each.

### DIFF
--- a/sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
+++ b/sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
@@ -3,7 +3,7 @@
 
 pub use crate::generated::clients::*;
 use crate::{
-    authorizer,
+    authorizer::KeyVaultAuthorizer,
     models::{
         CertificateClientCreateCertificateOptions, CertificateOperation,
         CreateCertificateParameters,
@@ -32,6 +32,9 @@ pub struct CertificateClientOptions {
     pub api_version: String,
     /// Allows customization of the client.
     pub client_options: ClientOptions,
+    /// Controls whether the client requires the resource specified in authentication
+    /// challenges to match the Key Vault or Managed HSM domain. True by default.
+    pub verify_challenge_resource: Option<bool>,
 }
 
 impl Default for CertificateClientOptions {
@@ -39,6 +42,7 @@ impl Default for CertificateClientOptions {
         Self {
             api_version: String::from("2025-07-01"),
             client_options: ClientOptions::default(),
+            verify_challenge_resource: Some(true),
         }
     }
 }
@@ -66,7 +70,7 @@ impl CertificateClient {
                 format!("{endpoint} must use http(s)"),
             ));
         }
-        let authorizer = authorizer::KeyVaultAuthorizer::new();
+        let authorizer = KeyVaultAuthorizer::new(options.verify_challenge_resource.unwrap_or(true));
         let auth_policy: Arc<dyn Policy> = Arc::new(
             BearerTokenAuthorizationPolicy::new(credential, Vec::<String>::new())
                 .with_on_request(authorizer.clone())
--- a/sdk/keyvault/azure_security_keyvault_keys/src/authorizer.rs
+++ b/sdk/keyvault/azure_security_keyvault_keys/src/authorizer.rs
@@ -11,7 +11,7 @@ use azure_core::{
     http::{
         headers::{Headers, CONTENT_LENGTH, CONTENT_TYPE, WWW_AUTHENTICATE},
         policies::auth::{Authorizer, OnChallenge, OnRequest},
-        Body, Context, Request,
+        Body, Context, Request, Url,
     },
     Result,
 };
@@ -22,12 +22,14 @@ use std::sync::Arc;
 #[derive(Debug)]
 pub(crate) struct KeyVaultAuthorizer {
     scope: RwLock<String>,
+    verify_challenge_resource: bool,
 }
 
 impl KeyVaultAuthorizer {
-    pub fn new() -> Arc<Self> {
+    pub fn new(verify_challenge_resource: bool) -> Arc<Self> {
         Arc::new(Self {
             scope: RwLock::new(String::new()),
+            verify_challenge_resource,
         })
     }
 
@@ -55,7 +57,7 @@ impl KeyVaultAuthorizer {
 
         Err(Error::with_message(
             ErrorKind::DataConversion,
-            format!("authentication challenge doesn't contain scope or resource: {challenge}"),
+            format!("no scope or resource in authentication challenge: {challenge}"),
         ))
     }
 }
@@ -123,6 +125,35 @@ impl OnChallenge for KeyVaultAuthorizer {
         {
             let mut cached_scope = self.scope.write().await;
             *cached_scope = scope.clone();
+        }
+        if self.verify_challenge_resource {
+            // the challenge resource's host must match the requested domain's host
+            let challenge_url = Url::parse(&scope).map_err(|_| {
+                Error::with_message(
+                    ErrorKind::DataConversion,
+                    format!("invalid audience in challenge: {challenge}"),
+                )
+            })?;
+            let challenge_host = challenge_url.host_str().ok_or_else(|| {
+                Error::with_message(
+                    ErrorKind::DataConversion,
+                    format!("invalid audience in challenge: {challenge}"),
+                )
+            })?;
+            let request_host = request.url().host_str().ok_or_else(|| {
+                // should be impossible because the client already sent the request and received a response
+                Error::with_message(
+                    ErrorKind::DataConversion,
+                    format!("invalid request URL: {}", request.url()),
+                )
+            })?;
+            if !request_host.ends_with(format!(".{challenge_host}").as_str()) {
+                return Err(Error::with_message(
+                    ErrorKind::Other,
+                    format!(
+                            "challenge resource '{scope}' doesn't match the requested domain '{request_host}'. Set verify_challenge_resource in client options to disable this validation if necessary. See https://aka.ms/azsdk/blog/vault-uri for more information`"
+                )));
+            }
         }
         if let Some(saved_body) = context.value::<Body>() {
             request.set_body(saved_body);
@@ -233,7 +264,7 @@ mod tests {
                     if attempt == 0 {
                         assert!(req.body().is_empty(), "first request should have empty body");
                         let mut headers = Headers::new();
-                        headers.insert("www-authenticate", r#"Bearer authorization="https://login.microsoftonline.com/tenant", resource="https://a.b""#);
+                        headers.insert(WWW_AUTHENTICATE, r#"Bearer authorization="https://login.microsoftonline.com/tenant", resource="https://a.b""#);
                         Ok(AsyncRawResponse::from_bytes(
                             StatusCode::Unauthorized,
                             headers,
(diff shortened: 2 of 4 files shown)


sdk/keyvault/azure_security_keyvault_certificates/src/lib.rs
```diff
@@ -13,5 +14,77 @@ mod generated;
...
+use std::sync::Arc;
+
+/// Options used when creating a [`CertificateClient`]
+#[derive(Clone, SafeDebug)]
+pub struct CertificateClientOptions {
+    /// The API version to use for this operation.
+    pub api_version: String,
+    /// Allows customization of the client.
+    pub client_options: ClientOptions,
+}
```
heaths: This must be defined in the `src/clients.rs` file (`clients` module) and re-exported here.

sdk/keyvault/azure_security_keyvault_certificates/src/lib.rs
```diff
@@ -13,5 +14,77 @@ mod generated;
...
+impl Default for CertificateClientOptions {
+    fn default() -> Self {
+        Self {
+            api_version: String::from("2025-07-01"),
+            client_options: ClientOptions::default(),
+        }
+    }
+}
+
+impl CertificateClient {
```
heaths: All implementations should also be centralized in the `clients` module i.e., `src/clients.rs`. That's guidelines.

sdk/keyvault/azure_security_keyvault_certificates/src/lib.rs
```diff
@@ -13,5 +14,77 @@ mod generated;
...
+    /// The API version to use for this operation.
+    pub api_version: String,
+    /// Allows customization of the client.
+    pub client_options: ClientOptions,
+}
+
+impl Default for CertificateClientOptions {
+    fn default() -> Self {
+        Self {
+            api_version: String::from("2025-07-01"),
```
heaths: @jhendrixMSFT this is one thing I hadn't considered when we talked about this last month. This means that whenever a new api-version is added, authors will have to remember to update this code manually.

Should we instead define a `pub(crate) DEFAULT_API_VERSION: &str = "2025-07-01"`, for example, in `src/generated/mod.rs` or something that we not only use in the generated `impl Default for *ClientOptions` but can be used by anyone like these instances?

/cc @RickWinter @LarryOsterman

sdk/keyvault/azure_security_keyvault_certificates/src/lib.rs
```diff
@@ -13,5 +14,77 @@ mod generated;
...
+            return Err(azure_core::Error::with_message(
+                azure_core::error::ErrorKind::Other,
+                format!("{endpoint} must use http(s)"),
+            ));
+        }
+        let authorizer = authorizer::KeyVaultAuthorizer::new();
+        let auth_policy: Arc<dyn Policy> = Arc::new(
+            BearerTokenAuthorizationPolicy::new(credential, Vec::<String>::new())
+                .with_on_request(authorizer.clone())
+                .with_on_challenge(authorizer),
```
heaths: This seems wasteful, even though it's an `Arc`. Is it only an `Arc` returned from `new` because of this? Assuming most will use the same instance, could one or the other method assume to use the same instance internally to avoid? It's not a blocker - not *that* wasteful - but seems that the default case (I assume) will always require two `Arc`s that point to the same instance.

sdk/keyvault/azure_security_keyvault_certificates/src/authorizer.rs
heaths, line 1: @weshaggard we need to share source in some cases like this. What are your thoughts on symlinks, which git supports? Unlike .NET, Rust has no good way to share source like we do for .NET.

sdk/keyvault/azure_security_keyvault_certificates/src/authorizer.rs
```diff
@@ -0,0 +1,484 @@
...
+                        } else {
+                            format!("{value}/.default")
+                        });
+                    }
+                }
+            }
+        }
+
+        Err(Error::with_message(
+            ErrorKind::Other,
```
heaths: Should this be `ErrorKind::Credential`? Or is that mostly reserved for actual auth issues and not necessarily bad response data? Is there a distinction in this case, though?

sdk/keyvault/azure_security_keyvault_certificates/src/authorizer.rs
```diff
@@ -0,0 +1,484 @@
...
+    async fn on_request(
+        &self,
+        ctx: &mut Context,
+        request: &mut Request,
+        authorizer: &dyn Authorizer,
+    ) -> azure_core::Result<()> {
+        let scope = self.scope.read().await;
+        if scope.is_empty() {
+            if !request.body().is_empty() {
+                ctx.insert(request.body().clone());
```
heaths: This could be expensive. We already have a `Request::body_mut()` I added late last year that will give you a `&mut Body`. I'd be open to a `Body::take(&mut self) -> Body` that, for a `Bytes` buffer, returns an owned `Body::Bytes(Bytes)` leaving an empty buffer in its place much like `Option::take(&mut self)`.

For a `SeekableStream`, since we actually wrap it with a `Box` we could do much the same. Would be cheaper and not copy memory.

sdk/keyvault/azure_security_keyvault_certificates/src/authorizer.rs
heaths: Talked to @weshaggard and he was fine with trying, but I'd make keys's file the primary for reasons we discussed internally.

sdk/keyvault/azure_security_keyvault_certificates/src/authorizer.rs
chlowell: In my mind, `ErrorKind::Credential` implies an authentication failure and is useful primarily as advice to the application not to retry a client method call. Should this be `ErrorKind::DataConversion`?

heaths: `DataConversion` is better than `Other`, yeah, but this still feels like an auth issue. The server gave bad response data, sure, but that would prevent authenticating.

I don't have a strong feeling either way, so unless @RickWinter or @ronniegeraghty do I leave it up to you. We might want to codify our stance in the guidelines just for reference, though: not strictness, but just so this confusion doesn't come up again without obvious decisions made.

sdk/keyvault/azure_security_keyvault_certificates/src/lib.rs
chlowell: Are you imagining changing the builder API by combining the traits or having it take an object implementing both, or changing this implementation to remove the copy?

heaths: I'm not recommending any particular course of action. How often will an `OnRequest` and an `OnChallenge` implementation be different? Should we maybe optimize for that case now where the builder takes a `T` that implements both? The typestate builder pattern would let us have it both ways.

chlowell: The only real use case I know of today is like Key Vault's: using a challenge to discover parameters for use in all future authentications. This scenario requires implementing both traits and sharing state between them--let me know if you want to discuss why this is so and whether a simpler design is possible--so I expect SDK devs will typically implement them on the same type. However, I don't want to require them to do that or make alternative approaches painful because other SDKs have supported a scenario that would require only one of these traits and we may eventually want to support something similar in Rust.

> The typestate builder pattern would let us have it both ways.

Can you please elaborate on this?

heaths: As discussed offline, I thought about this more last night and this pattern is fine. It's not the most idiomatic pattern, sure, but it's also not a "true public API" that *users* of the SDK will use, and even very few - only KV now - will use in the org. Just keep it as-is.

sdk/core/typespec_client_core/src/http/request/mod.rs
```diff
@@ -67,6 +67,18 @@ impl Body {
...
     }
 
+    /// Takes the body, leaving an empty body in its place.
+    pub fn take(&mut self) -> Body {
+        match self {
+            Body::Bytes(_) => std::mem::replace(self, Body::Bytes(Bytes::new())),
+            #[cfg(not(target_arch = "wasm32"))]
+            Body::SeekableStream(_) => std::mem::replace(
+                self,
+                Body::SeekableStream(Box::new(BytesStream::new_empty())),
```
heaths, line 77: Oh, I didn't realize that existed. Nice!

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
```diff
@@ -2,22 +2,94 @@
...
+/// Options used when creating a [`CertificateClient`]
+#[derive(Clone, SafeDebug)]
+pub struct CertificateClientOptions {
+    /// The API version to use for this operation.
+    pub api_version: String,
+    /// Allows customization of the client.
+    pub client_options: ClientOptions,
+    /// Controls whether the client requires the resource specified in authentication
+    /// challenges to match the Key Vault or Managed HSM domain. True by default.
+    pub verify_challenge_resource: Option<bool>,
```
heaths, line 37: I'm torn on this being an `Option<bool>`. Can you think of any reason where `None` is interesting if we default to `true` (or even `false`)? I don't think this adds value, even though it follows our guidelines (which aren't hard and fast - "they're more like...guidelines").

heaths: As discussed offline, let's just keep it like this, `Option<bool>`. While we can't think of any good scenario where this adds value, it does follow guidelines and most other client and client method options don't have non-optional types. Maybe never useful, but at least consistent.

sdk/keyvault/azure_security_keyvault_certificates/src/lib.rs
jhendrixMSFT: Wait why is this in `lib.rs`?  The emitter doesn't write it there.

chlowell: This is handwritten, and since moved to client.rs

jhendrixMSFT: Clarified in https://github.com/Azure/typespec-rust/issues/782

PR #3482, opened 2026-01-06
Updating Request Options (throughput bucket, priority, additional headers)

### LINKED ISSUE #3446: Cosmos: Finish remaining Request Options
We have several request options across various APIs that we need to fill in. I'll link subissues for ones that have already been identified. But another aspect of this work will be reviewing the Go SDK and making sure we have similar functionality in Rust.

tvaron3:
## Summary

This PR adds support for new Cosmos DB request headers across the Rust SDK, including **throughput bucket**, **priority level**, and **custom additional headers**, with clear precedence rules between request-level and client-level configuration.

### New request option fields + header emission

Added:

- `throughput_bucket: Option<i32>`
- `priority: Option<PriorityLevel>`
- `custom_headers: Vec<(HeaderName, HeaderValue)>`

to:

- `CosmosClientOptions`
- `ItemOptions`
- `QueryOptions`

Implemented/extended `AsHeaders` so these options are converted into the corresponding request headers:

- `x-ms-cosmos-throughput-bucket`
- `x-ms-cosmos-priority-level`
- plus any user-provided `custom_headers`

### Priority enum

Introduced `PriorityLevel` (`High`, `Low`) with `Display` mapping to the wire values used by Cosmos DB.

### Clear header precedence rules

Request-specific headers always win over client defaults:

- Added `CosmosRequest::client_headers()` to apply client headers only when the request doesn’t already contain that header.
- Updated query request construction in `CosmosPipeline::send_query_request()` to apply client option headers only if the request doesn’t already have them.

### Multi-write header
- Added `x-ms-cosmos-allow-tentative-writes` header to write requests for multi write accounts. This will unblock customers trying to leverage multiwrite capabilities. 

### API clarity: item/request headers vs client headers

- Renamed request builder method from `.headers(&...)` to `.item_headers(&...)` where applicable to reflect intent more clearly.
- Updated call sites in `ContainerClient`, `DatabaseClient`, and `CosmosClient` to use `.item_headers(...)`.

### Tests

- Added/updated unit tests validating new header output and ordering (`ItemOptions`, `QueryOptions`, `CosmosClientOptions`).
- Added a test ensuring request headers are prioritized over client headers.

## Why

These headers enable:

- Throughput bucket: https://learn.microsoft.com/azure/cosmos-db/nosql/throughput-buckets
- Priority: https://learn.microsoft.com/azure/cosmos-db/priority-based-execution
- Opt-in support for additional Cosmos headers (e.g., dedicated gateway integrated cache headers) without requiring SDK changes for every new header
   - adds flexibility for future and these headers are very specific to some use cases and are not relevant in thin client scenarios
   - customers can unblock themselves for features that we have not implemented yet

## Notes / Behavior

- `additional_headers` are appended after the typed headers in `AsHeaders`. These are only added if the header does not already exist to prevent customers from overwriting other headers that we want to keep default. 
- Client-level headers are only inserted if the request does not already contain that header name, ensuring request-level overrides behave as expected.

## Relevant GH Issue

Fixes #3446

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -330,7 +330,6 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "url",
- "uuid",
 ]
 
 [[package]]


sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -20,14 +20,65 @@ pub struct CosmosClientOptions {
...
+            ));
+        }
+
+        if let Some(throughput_bucket) = &self.throughput_bucket {
+            headers.push((
+                constants::THROUGHPUT_BUCKET,
+                throughput_bucket.to_string().into(),
+            ));
+        }
+
```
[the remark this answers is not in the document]
tvaron3: fixed

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -239,10 +377,53 @@ impl QueryOptions<'_> {
...
```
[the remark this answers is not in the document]
tvaron3: fixed

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -153,6 +224,20 @@ pub struct ItemOptions<'a> {
     ///
     /// The default for this is `false`, which reduces the network and CPU burden that comes from serializing and deserializing the response.
     pub enable_content_response_on_write: bool,
+    /// The desired throughput bucket for the client
```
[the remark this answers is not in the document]
tvaron3: fixed

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -229,6 +342,31 @@ pub struct QueryOptions<'a> {
...
+    /// The client instance will use this token internally with each read/query request to ensure that the set consistency level is maintained.
+    ///
+    /// See [Session Tokens](https://learn.microsoft.com/azure/cosmos-db/nosql/how-to-manage-consistency?tabs=portal%2Cdotnetv2%2Capi-async#utilize-session-tokens) for more.
+    pub session_token: Option<String>,
+    /// Used to specify the consistency level for the operation.
+    ///
+    /// The default value is the consistency level set on the Cosmos DB account.
+    /// See [Consistency Levels](https://learn.microsoft.com/azure/cosmos-db/consistency-levels)
+    pub consistency_level: Option<ConsistencyLevel>,
+    /// The desired throughput bucket for the client
```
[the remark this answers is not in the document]
tvaron3: fixed

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -229,6 +342,31 @@ pub struct QueryOptions<'a> {
...
+    /// Priority based execution allows users to set a priority for each request. Once the user has reached their provisioned throughput, low priority requests are throttled
+    /// before high priority requests start getting throttled. Feature must first be enabled at the account level.
+    ///
+    /// See [Priority based-execution](https://learn.microsoft.com/azure/cosmos-db/priority-based-execution) for more.
+    pub priority: Option<PriorityLevel>,
+    /// Additional headers to be included in the query request. This allows for custom headers beyond those natively supported.
+    /// The following are some example headers that can be added using this api.
+    /// Dedicated gateway cache staleness: "x-ms-dedicatedgateway-max-age" - See https://learn.microsoft.com/azure/cosmos-db/how-to-configure-integrated-cache?tabs=dotnet#adjust-maxintegratedcachestaleness for more info.
+    /// Bypass dedicated gateway cache: "x-ms-dedicatedgateway-bypass-cache" - See https://learn.microsoft.com/azure/cosmos-db/how-to-configure-integrated-cache?tabs=dotnet#bypass-the-integrated-cache for more info.
+    pub custom_headers: Vec<(HeaderName, HeaderValue)>
```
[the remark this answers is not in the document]
tvaron3: fixed.

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -370,6 +382,51 @@ mod tests {
...
+            OperationType::Create,
+            ResourceLink::root(ResourceType::Documents),
+        )
+        .request_headers(&item_options)
+        .build()
+        .unwrap();
+
+        let mut req_with_client_headers = req.clone();
+        req_with_client_headers.request_context.location_endpoint_to_route =
+            Some("https://example.com/".parse().unwrap());
```
simorenoh: non-blocking, just wondering why this is needed since I would have assumed just the copied request with the client_headers would have worked

tvaron3: The code will panic if I don't set the location_endpoint_to_route. The method `into_raw_request` expects the `location_endpoint_to_route` to always be defined in the lines shown below. 
Screenshot 2026-01-07 at 11 03 29 AM

analogrelay: Yeah, there are some refactorings ahead for CosmosRequest and the pipeline. I'm concerned it's too sensitive to the order of operations in the pipeline which makes testing quite brittle. But that's not something that needs to be addressed in this PR

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -20,14 +20,63 @@ pub struct CosmosClientOptions {
...
+    ///
+    /// See [Priority based-execution](https://learn.microsoft.com/azure/cosmos-db/priority-based-execution) for more.
+    pub priority: Option<PriorityLevel>,
+    /// Additional headers to be included in the query request. This allows for custom headers beyond those natively supported.
+    /// The following are some example headers that can be added using this api.
+    /// Dedicated gateway cache staleness: "x-ms-dedicatedgateway-max-age".
+    /// See https://learn.microsoft.com/azure/cosmos-db/how-to-configure-integrated-cache?tabs=dotnet#adjust-maxintegratedcachestaleness for more info.
+    /// Bypass dedicated gateway cache: "x-ms-dedicatedgateway-bypass-cache".
+    /// See https://learn.microsoft.com/azure/cosmos-db/how-to-configure-integrated-cache?tabs=dotnet#bypass-the-integrated-cache for more info.
+    pub custom_headers: Vec<(HeaderName, HeaderValue)>,
```
analogrelay: I think this should be a `HashMap<HeaderName, HeaderValue>`. If it's a `Vec<...>`, nothing stops a user from putting multiple entries in for the same header, but the behavior of that isn't clear. The underlying Core SDK collection we use wraps a HashMap, which means that multiple occurrences of the same header will _overwrite_ earlier occurrences.

By using a HashMap, we force the user to either avoid multiple values for a header, or figure out how they want to compose that before passing it to us. `HashMap` defines a `From<Vec<(K, V)>>`, so users can pretty easily create a "literal" hash map using `vec![(key, value), ...].into()`

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -153,6 +222,22 @@ pub struct ItemOptions<'a> {
     ///
     /// The default for this is `false`, which reduces the network and CPU burden that comes from serializing and deserializing the response.
     pub enable_content_response_on_write: bool,
+    /// The desired throughput bucket for this request
+    ///
+    /// See [Throughput Control in Azure Cosmos DB](https://learn.microsoft.com/azure/cosmos-db/nosql/throughput-buckets) for more.
+    pub throughput_bucket: Option<i32>,
```
analogrelay: Is a negative throughput bucket valid? If not, we should use `u32`. In fact, I'd suggest using `usize` unless we _need_ it to be a 32-bit integer, because that's the default type for integer constants.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -20,14 +20,63 @@ pub struct CosmosClientOptions {
...
+            headers.push((constants::PRIORITY_LEVEL, priority.to_string().into()));
+        }
+
+        if let Some(throughput_bucket) = &self.throughput_bucket {
+            headers.push((
+                constants::THROUGHPUT_BUCKET,
+                throughput_bucket.to_string().into(),
+            ));
+        }
+        add_custom_headers(&mut headers, &self.custom_headers);
```
analogrelay: If you flip the order here, and set custom headers first, you don't need to have the logic in `add_custom_headers` to avoid overwriting specified headers with custom headers.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -229,6 +342,33 @@ pub struct QueryOptions<'a> {
     /// Specifically, the query engine may be built-in to the SDK in the future, and this option may be removed entirely.
     #[cfg(feature = "preview_query_engine")]
     pub query_engine: Option<crate::query::QueryEngineRef>,
+    /// Applies when working with Session consistency.
+    /// Each new write request to Azure Cosmos DB is assigned a new Session Token.
+    /// The client instance will use this token internally with each read/query request to ensure that the set consistency level is maintained.
+    ///
+    /// See [Session Tokens](https://learn.microsoft.com/azure/cosmos-db/nosql/how-to-manage-consistency?tabs=portal%2Cdotnetv2%2Capi-async#utilize-session-tokens) for more.
+    pub session_token: Option<String>,
```
analogrelay: We should make a [newtype](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) for this. I believe session tokens are intended to be opaque to customers, but that the SDK may need to be able to parse and understand them. Making this `Option<SessionToken>`, where `SessionToken` is defined as `pub struct SessionToken(String)` (note that the String is _private_) and then adding `impl From<String> for SessionToken` and `impl Display for SessionToken` allows a user to convert SessionTokens to and from strings for serialization, but keep them in the typed wrapper when working with the SDK.

It's similar to how [Secret](https://docs.rs/azure_core/latest/azure_core/credentials/struct.Secret.html) works, but for a different reason. It also allows us to control the serialization/deserialization logic at the point the user tries to create a SessionToken, rather than somewhere deep in the SDK, so that users get immediate feedback.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -99,6 +104,22 @@ impl CosmosPipeline {
...
+        // Caller-provided request headers must take precedence.
+        for (name, value) in self
+            .options
+            .as_headers()
+            .expect("CosmosClientOptions is infallible")
+        {
+            let exists = base_request
+                .headers()
+                .iter()
+                .any(|(existing_name, _)| *existing_name == name);
```
analogrelay: Are the headers stored as a HashMap? If so, we can check for existence without iterating. If not, they probably should be :)

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
tvaron3: Changed to use hashmap

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
tvaron3: negative throughput buckets are not valid. Only 1-5 is valid bucket values. Note though that this and priority execution are preview features. Changed to usize.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
tvaron3: changed to add custom headers at beginning. My only worry is that in the future someone doesn't realize that the order matters especially for headers that are not added as part of the asHeaders of a request like the header for partition keys for example.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
tvaron3: Created new type Created new type for session token.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
tvaron3: It is a hash map. Updated the logic get_optionally to see if the value exists.

Cargo.lock
```diff
@@ -330,6 +330,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "url",
+ "uuid",
```
tvaron3:
```suggestion
```

PR #3485, opened 2026-01-06
Remove unused azure_identity options

chlowell:
These exist in other SDKs but we haven't yet implemented their associated features in Rust.


heaths (changes requested):
For `additionally_allowed_tenants`, now that we author the constructors, shouldn't we use them instead of removing them?

PR #3487, opened 2026-01-07
Added scripts and packages to start testing emitters

LarryOsterman:


.vscode/settings.json
```diff
@@ -24,4 +24,5 @@
   "[powershell]": {
     "editor.defaultFormatter": "ms-vscode.powershell",
   },
-}
+  "cmake.ignoreCMakeListsMissing": true,
```
heaths, line 24: What's this for? We don't even need the cmake extension do we? I was trying to limit settings to built-in stuff and just those extensions we recommend in `.vscode/extensions.json`.

eng/dict/crates.txt
heaths, line 1: Might be worth taking, but I also already have a script that adds these as based on actual projects we have in the workspace as needed. Loading a larger list than we need might hinder performance, though may not even be noticeable. I'll at least sort these, though. Cspell doesn't need them to be, but it helps with future diffs.

eng/emitter-package.json
heaths, line 1: @jhendrixMSFT or I update this any time we update the emitter anyway. This is already out of date. ;)

.vscode/settings.json
LarryOsterman: It convinces vs code that it shouldn't try to process the cmakelists.txt that is present in the repo. If the cmake extension is present, VS code tries to build the cmakelists.txt when you launch vs code. Which we basically don't ever want.

eng/emitter-package.json
LarryOsterman: As long as *someone* adds the missing packages to the emitter-package.json, I'm good. But right now, we're missing a fair number of packages that will be required. There seems to be no real cost to having the packages present.

heaths: Oh, right. Sorry. I forgot you mentioned that and wasn't considering most of these were added.

PR #3495, opened 2026-01-07
Add Live Tests Infra to Cosmos

tvaron3:
### Live Tests
Adds live tests support and changes test organization to use cfg flags for creating different test categories. Test categories are necessary to shard live test runs and to exclude irrelevant tests for certain account configurations. The test organization follows the design outlined here  https://github.com/Azure/azure-sdk-for-rust/issues/3508#issuecomment-3751901191. 
- test-resources.bicep: Creates the different types of cosmosdb accounts and also adds the relevant permissions to use EntraID in tests. Added the parameters that were useful in other languages like `multi_region`, `multi_write`, `consistency_level`
- live-platform-matrix.json: Dictates the testing matrix for creating the different test jobs in azdevops. 

### Refactored Tests
- Changed tests where applicable to reuse a database. This is setup through the `test-resources.bicep`. At the end of each test job, the resource group will be deleted. Each test job has a different resource group. 
- added a 80 second default timeout to all the tests 
- added a methods for read operations to retry to make them more resilient to different consistency levels. 
- any test failure due to a 429 will be retried to make live tests more stables
- all live tests run sequentially to avoid flakiness

### Multi-write Changes
Fixed the location cache to only use multiple write locations for document requests or stored procedures when routing a request. Simplified some of the logic that was repeated. Made these changes to add a example test for multi-write. 

## Remaing work
@danieljurek 
- [ ] Move our live tests out of `rust-cosmos` and into `rust-cosmos-weekly`. 
   - Currently, to run live tests a box labeled `live tests` must be checked in the `rust-cosmos` azdevops pipeline to run the tests
- [ ] Update live platform matrix to be sparse to avoid having excessive test jobs
- More details see https://github.com/Azure/azure-sdk-for-rust/issues/3508#issuecomment-3751856561


sdk/cosmos/test-resources.bicep
```diff
@@ -0,0 +1,108 @@
...
+  name: guid(resourceGroup().id, roleAssignmentId, testApplicationOid)
+  properties: {
+    roleDefinitionId: accountName_roleDefinitionId.id
+    principalId: testApplicationOid
+    scope: cosmosAccount.id
+  }
+}
+
+output RUSTFLAGS string = '--cfg=test_category="${testCategory}"'
+output AZURE_COSMOS_CONNECTION_STRING string = 'AccountEndpoint=${reference(resourceId, apiVersion).documentEndpoint};AccountKey=${listKeys(resourceId, apiVersion).primaryMasterKey};'
```
FabianMeiswinkel, line 121: Honest question - should we even add live tests for key based auth? Or stick with key based only for emulator tests and stick with Entra ID for all live tests? We assumet his will become more and more the norm anyway? Definitely not blocking - even if we agree on this long-term. Just raising here to discuss/close on it

sdk/cosmos/azure_data_cosmos/Test-Setup.ps1
```diff
@@ -5,91 +5,116 @@
...
+            LogGroupStart "Installing Cosmos DB Emulator"
+            & "$PSScriptRoot\..\..\..\eng\common\scripts\Cosmos-Emulator.ps1" `
+                -StartParameters "/noexplorer /noui /disableratelimiting /enableaadauthentication /partitioncount=50" `
+                -Stage "Install"
+            LogGroupEnd
+        }
+
+        LogGroupStart "Launching Cosmos DB Emulator"
         & "$PSScriptRoot\..\..\..\eng\common\scripts\Cosmos-Emulator.ps1" `
             -StartParameters "/noexplorer /noui /disableratelimiting /enableaadauthentication /partitioncount=50" `
```
FabianMeiswinkel: DefaultConsistency needs to be set here?

sdk/cosmos/azure_data_cosmos/Test-Setup.ps1
```diff
@@ -5,91 +5,116 @@
...
-    # Set environment variables for the tests
-    $env:AZURE_COSMOS_CONNECTION_STRING = "emulator"
-} elseif (Get-Command "docker" -ErrorAction SilentlyContinue) {
-    Write-Host "Docker detected. Using Cosmos DB Emulator in Docker."
+        # Check if the emulator is already running
+        $existingContainer = docker ps --filter "name=cosmosdb-emulator-test" --format "{{.Names}}"
+        if ($existingContainer -eq "cosmosdb-emulator-test") {
+            Write-Host "Cosmos DB Emulator container is already running."
+        } else {
+            LogGroupStart "Starting Cosmos DB Emulator in Docker"
```
FabianMeiswinkel: Is this for the old emualtor running in docker or for the new pre-view emulator?

sdk/cosmos/azure_data_cosmos/tests/test_setup.rs
```diff
@@ -0,0 +1,62 @@
...
+#[ignore = "Run manually to set up shared test resources"]
+async fn setup_shared_resources() -> Result<(), Box<dyn std::error::Error>> {
+    // Initialize tracing
+    _ = tracing_subscriber::fmt()
+        .with_env_filter(EnvFilter::from_default_env())
+        .try_init();
+
+    let Some(client) = create_client()? else {
+        println!("No connection string provided. Skipping test setup.");
+        return Ok(());
```
FabianMeiswinkel: shouldn't this fail (we only call tihs if a connection string is available)?

sdk/cosmos/azure_data_cosmos/tests/test_setup.rs
```diff
@@ -0,0 +1,62 @@
...
+    };
+
+    println!("Setting up shared test resources...");
+
+    // Create shared database
+    println!("Creating shared database: {}", SHARED_DATABASE_ID);
+    match client.create_database(SHARED_DATABASE_ID, None).await {
```
FabianMeiswinkel: Would it make sense to force SharedDatabaseId to have some timestamp in it - to allow automatically celaning-up databases left-over (for example because cleanup scritp failed / was accidentally not executed)?

FabianMeiswinkel: never mind - we always use teh same shared database - which means there could never ever run two tests on the same VM (assuming emulator) - and it would never work for live tests?

FabianMeiswinkel: My 2 cents - a shared container is not really helpful - there are too many possible races or problems being able to isolate test validations form other tests. I would focus on building a robust clean-up for a shared database - but then test will have to define containers in many cases anyway (to iterate over various PKs, index policy, different thorughput to force single vs. multiple partitions etc.)

sdk/cosmos/azure_data_cosmos/tests/multi_write/cosmos_multi_write.rs
```diff
@@ -0,0 +1,101 @@
...
+        .create_container(
+            properties,
+            Some(CreateContainerOptions {
+                throughput: Some(throughput),
+                ..Default::default()
+            }),
+        )
+        .await?
+        .into_model()?;
+    // keep reading container until it is fully created
```
FabianMeiswinkel: This needs to happen in all regions - because container creation is an async task in each region - will otherwise result in flaky tests

sdk/cosmos/azure_data_cosmos/tests/multi_write/cosmos_multi_write.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    let container_client = db_client.container_client(&created_properties.id);
+    let _create_result = container_client
+        .upsert_item(
+            "item1",
+            &serde_json::json!({"id": "item1", "value": "test"}),
+            None,
+        )
+        .await?;
```
FabianMeiswinkel: Validation missing that it actually uses the expected region?

sdk/cosmos/azure_data_cosmos/tests/emulator/cosmos_offers.rs
```diff
@@ -0,0 +1,154 @@
...
+
+            // Read throughput
+            let current_throughput = container_client
+                .read_throughput(None)
+                .await?
+                .expect("throughput should be present");
+
+            assert_eq!(Some(500), current_throughput.throughput());
+            assert_eq!(Some(5000), current_throughput.autoscale_maximum());
+            assert_eq!(Some(42), current_throughput.autoscale_increment());
```
FabianMeiswinkel, line 140: Just out of curiosity - what is this?

sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_items.rs
```diff
@@ -0,0 +1,677 @@
...
+            let pk = format!("Partition1-{}", unique_id);
+            let item_id = format!("Item1-{}", unique_id);
+
+            container_client.upsert_item(&pk, &item, None).await?;
+
+            let read_item: TestItem = container_client
+                .read_item(&pk, &item_id, None)
+                .await?
+                .into_model()?;
+            assert_eq!(item, read_item);
```
FabianMeiswinkel, line 213: Not safe for eventual consistency?

FabianMeiswinkel: asserts on status code (and other headers/diagnostics missing) - I assume intentional and fine for now - maybe worth adding some todo comment?

sdk/cosmos/azure_data_cosmos/tests/emulator/cosmos_items.rs
```diff
@@ -0,0 +1,677 @@
...
+                },
+                bool_value: true,
+            };
+
+            let pk = format!("Partition1-{}", unique_id);
+            let item_id = format!("Item1-{}", unique_id);
+
+            container_client.upsert_item(&pk, &item, None).await?;
+
+            let read_item: TestItem = container_client
```
FabianMeiswinkel: no safe for eventual consistency

sdk/cosmos/azure_data_cosmos/tests/emulator/cosmos_items.rs
```diff
@@ -0,0 +1,677 @@
...
+                bool_value: true,
+            };
+
+            let pk = format!("Partition1-{}", unique_id);
+            let item_id = format!("Item1-{}", unique_id);
+
+            container_client.create_item(&pk, &item, None).await?;
+
+            let read_item: serde_json::Value = container_client
+                .read_item(&pk, &item_id, None)
```
FabianMeiswinkel: Not safe with eventual

sdk/cosmos/azure_data_cosmos/tests/emulator/cosmos_items.rs
```diff
@@ -0,0 +1,677 @@
...
+            let pk = format!("Partition@1-{}", unique_id);
+            let item_id = format!("Item@1-{}", unique_id);
+
+            let response = container_client.create_item(&pk, &item, None).await?;
+            let body = response.into_body().into_string()?;
+            assert_eq!("", body);
+
+            // Try to read the item
+            let read_item: TestItem = container_client
+                .read_item(&pk, &item_id, None)
```
FabianMeiswinkel: not safe for eventual

sdk/cosmos/azure_data_cosmos/tests/emulator/cosmos_offers.rs
tvaron3: It is the `incrementPercent` in the auto scale throughput policy. From what I understand it seems to limit the autoscale increase up by percentage, but I couldn't find public docs. Below is our public java doc in the java sdk for same option. 

Image

sdk/cosmos/azure_data_cosmos/tests/multi_write/cosmos_multi_write.rs
tvaron3: This happens on the region being used each time. The database gets deleted by the `run_with_db`. Right now there is no way for me to specify on a per request level the region.

sdk/cosmos/azure_data_cosmos/Test-Setup.ps1
tvaron3: in java we use the following. I will make the change to use the same ones here. https://github.com/Azure/azure-sdk-for-java/blob/ac0a72054ca50743c32158195ba38c785b4f6272/eng/pipelines/templates/stages/cosmos-sdk-client.yml#L100

sdk/cosmos/azure_data_cosmos/Test-Setup.ps1
tvaron3: This is the linux based vnext emulator. I assume you lean towards to removing this or moving it to non blocking jobs because it is still missing support for many scenarios. I can create a github issue for moving this to a non-blocking pipeline job.

sdk/cosmos/test-resources.bicep
tvaron3: I agree. I can do this in a followup pr. I will create a github issue.

sdk/cosmos/azure_data_cosmos/tests/framework/test_client.rs
```diff
@@ -172,14 +223,20 @@ impl TestClient {
         }
     }
 
-    pub async fn run_with_db<F>(test: F) -> Result<(), Box<dyn std::error::Error>>
+    pub async fn run_with_db<F>(
```
analogrelay: Maybe rename this to `run_with_unique_db`, since that's what it does?

sdk/cosmos/azure_data_cosmos/tests/framework/test_client.rs
```diff
@@ -235,6 +292,12 @@ impl TestRunContext {
         Ok(db_client)
     }
 
+    /// Get shared database client for this test run
+    pub fn get_shared_container(&self) -> ContainerClient {
```
analogrelay: I'd consider making this a `run_with` function like the others: `run_with_shared_container`.

sdk/cosmos/azure_data_cosmos/tests/multi_write/cosmos_multi_write.rs
analogrelay: > Right now there is no way for me to specify on a per request level the region.

If that's a gap we need to close, can you file a tracking issue for it and add it to the project? We can also reference updating this test if needed.

sdk/cosmos/azure_data_cosmos/tests/emulator.rs
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+#![cfg(test_category = "emulator")]
+#[path = "emulator/mod.rs"]
```
analogrelay: I _believe_ the `#[path]` attribute here isn't needed, since the default search path for `mod foo` is both `foo.rs` and `foo/mod.rs`.

sdk/cosmos/azure_data_cosmos/tests/multi_write.rs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+#![cfg(test_category = "multi_write")]
+
+#[path = "multi_write/mod.rs"]
```
analogrelay: Same here, I don't think you need the path attribute.

sdk/cosmos/azure_data_cosmos/tests/emulator.rs
analogrelay: The `tests/` directory _may_ be different though, if that's the case we can keep the path attribute

sdk/cosmos/azure_data_cosmos/Test-Setup.ps1
analogrelay: No, this is the Linux-based **OLD** emulator (the one using windows emulation). Still, happy to remove it or make it non-blocking if necessary. Now that we have our windows build up and running it's not as necessary.

I **do** think we should look at a non-blocking pipeline run on the vNext emulator at some point, but that's separate to this.

sdk/cosmos/azure_data_cosmos/tests/test_setup.rs
analogrelay: It could be a per-test run ID. I think there's value in doing that.

sdk/cosmos/test-resources.bicep
analogrelay: If we have key-based auth functionality, we should probably be testing it, even if we strongly recommend against it.

sdk/cosmos/azure_data_cosmos/tests/test_setup.rs
```diff
@@ -0,0 +1,62 @@
...
+use azure_core::http::StatusCode;
+use azure_data_cosmos::models::ContainerProperties;
+use tracing_subscriber::EnvFilter;
+
+use crate::framework::SHARED_PARTITION_KEY;
+use framework::test_client::{create_client, SHARED_CONTAINER_ID, SHARED_DATABASE_ID};
+
+#[tokio::test]
+#[ignore = "Run manually to set up shared test resources"]
+async fn setup_shared_resources() -> Result<(), Box<dyn std::error::Error>> {
```
analogrelay: Could you do this setup in the `test-resources.bicep` and/or `Test-Setup.ps1` (using `az` CLI)? We're not really testing the Rust SDKs ability to create dbs/containers here, so it might even be better to use known-good paths to avoid a catastrophically broken test run if we mess with something in that logic.

analogrelay: And then, I presume the entire account would be cleaned up by the eng sys infra.

sdk/cosmos/azure_data_cosmos/build.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+// Registers custom cfgs used by this crate's integration tests.
+//
+// Some CI/build setups enable `-W unexpected-cfgs`, and in newer Rust toolchains
+// unknown cfg names are warned/denied unless explicitly declared via check-cfg.
```
analogrelay, line 6: This is what I was a bit worried about, `cfg`s and other test-only compilation options leaking into the main build. It's why I was advocating for a separate `azure_data_cosmos_tests` crate, because we can do whatever we  want there without worrying about it polluting the product code.

This is a fine way to solve the problem if we want to have these tests here in `azure_data_cosmos` so I'm not suggesting a change, just pointing out the messiness here. cc @heaths

sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -10,6 +10,10 @@ repository.workspace = true
 rust-version.workspace = true
 homepage = "https://github.com/azure/azure-sdk-for-rust"
 documentation = "https://docs.rs/azure_data_cosmos"
+
+# Registers custom cfgs like `test_category` for integration test gating.
+build = "build.rs"
```
analogrelay: I don't think you need this unless you change the name of the `build.rs` to something else. If there's a `build.rs`, it will automatically be picked up as the build script.

sdk/cosmos/.dict.txt
```diff
@@ -11,10 +11,19 @@ cloneable
...
 upserts
 upserted
 upserting
+LINUXVMIMAGE
+WINDOWSVMIMAGE
+MACVMIMAGE
+LINUXPOOL
+WINDOWSPOOL
+MACPOOL
+RUSTFLAGS
```
analogrelay: These can probably be promoted to the repo-level dictionary, but I'm OK holding off on that to avoid adding a bunch of pending reviews :).

sdk/cosmos/live-platform-matrix.json
```diff
@@ -0,0 +1,35 @@
...
+      },
+      "Session": {
+        "ArmTemplateParameters": "@{ defaultConsistencyLevel = 'Session' }"
+      },
+      "Strong": {
+        "ArmTemplateParameters": "@{ defaultConsistencyLevel = 'Strong'; }"
+      },
+      "Session-MultiWrite": {
+        "ArmTemplateParameters": "@{ defaultConsistencyLevel = 'Session'; enableMultipleWriteLocations = $true; testCategory = 'multi_write' }"
+      }
```
analogrelay: Can we use a little bit more consistent naming on the matrix pivots here? Instead of `Session` and `Session-MultiWrite` for example: `Session SingleWrite` and `Session MultiWrite` (or similar)? Basically, I'd like it if the names can specify every selection they've made for the various pivots, so we don't have to know which ones are defaults (SingleWrite being a default here). It makes it easier, IMO, when looking at a failed test stage, to know what categories were impacted.

sdk/cosmos/azure_data_cosmos/tests/multi_write.rs
tvaron3: I get this error when I try to remove it.
```
error[E0761]: file for module `multi_write` found at both "sdk/cosmos/azure_data_cosmos/tests/multi_write.rs" and "sdk/cosmos/azure_data_cosmos/tests/multi_write/mod.rs"
 --> sdk/cosmos/azure_data_cosmos/tests/multi_write.rs:5:1
  |
5 | mod multi_write;
  | ^^^^^^^^^^^^^^^^
  |
  = help: delete or rename one of them to remove the ambiguity
```

tvaron3: I will rename the tests folder and then I shouldn't need this

sdk/cosmos/azure_data_cosmos/tests/multi_write/cosmos_multi_write.rs
tvaron3: added the validation. Relevant information wasn't available from the response so added debug logs that I could parse.

sdk/cosmos/azure_data_cosmos/tests/multi_write/cosmos_multi_write.rs
tvaron3: It would be possible once excluded locations is available so I think it is accounted for.

sdk/cosmos/azure_data_cosmos/tests/test_setup.rs
tvaron3: removed shared container

tvaron3: changed it to use a per-test job run id.

sdk/cosmos/azure_data_cosmos/tests/test_setup.rs
tvaron3: removed this to create the database through bicep

sdk/cosmos/azure_data_cosmos/Test-Setup.ps1
tvaron3: I think for now should be fine to leave as is.

sdk/cosmos/azure_data_cosmos/Test-Setup.ps1
tvaron3: changed to use same configs as java

sdk/cosmos/test-resources.bicep
tvaron3: I will do this in a follow up pr. I will change live tests to used aad. Created an issue for tracking https://github.com/orgs/Azure/projects/917/views/1?pane=issue&itemId=151482932&issue=Azure%7Cazure-sdk-for-rust%7C3598

sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_items.rs
tvaron3: added helper methods to read item with infinite retires did same for queries.

sdk/cosmos/azure_data_cosmos/tests/emulator/cosmos_items.rs
tvaron3: added helper methods to read item with infinite retires did same for queries.

sdk/cosmos/azure_data_cosmos/tests/emulator/cosmos_items.rs
tvaron3: added helper methods to read item with infinite retires did same for queries.

sdk/cosmos/azure_data_cosmos/tests/emulator.rs
tvaron3: you were right removed it 😄

sdk/cosmos/azure_data_cosmos/tests/framework/test_client.rs
tvaron3: removed as not creating shared container anymore

sdk/cosmos/azure_data_cosmos/tests/test_setup.rs
tvaron3: Yes the resource group for each test job gets deleted at the end.

sdk/cosmos/azure_data_cosmos/Cargo.toml
tvaron3: yup removed

sdk/cosmos/.dict.txt
tvaron3: moved them and created top level .dict. I already require approvals from others due to the eng sys changes so don't mind adding more reviews

sdk/cosmos/live-platform-matrix.json
tvaron3: changed the naming as suggested

sdk/cosmos/azure_data_cosmos/tests/framework/test_client.rs
tvaron3: renamed

sdk/cosmos/azure_data_cosmos/tests/emulator.rs
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+#![cfg(test_category = "emulator")]
```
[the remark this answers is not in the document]
tvaron3: Yes these are set in the test-resources.bicep

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+- Added basic multi-region writes support. ([#3482](https://github.com/Azure/azure-sdk-for-rust/pull/3482) and [#3495](https://github.com/Azure/azure-sdk-for-rust/pull/3495))
+
```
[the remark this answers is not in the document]
tvaron3: both of these prs made the changes for this so I think it is relevant to include both

sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_items.rs
```diff
@@ -0,0 +1,707 @@
...
+            let body = response.into_body().into_string()?;
+            assert_eq!("", body);
+
+            // Try to read the item again, expecting a 404
+            // infinite loop to avoid test flakes due to eventual consistency
+            loop {
+                match container_client
+                    .read_item::<TestItem>(&pk, &item_id, None)
+                    .await
+                {
```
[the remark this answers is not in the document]
tvaron3: fixed

sdk/cosmos/azure_data_cosmos/tests/multi_write_tests/cosmos_multi_write.rs
```diff
@@ -0,0 +1,230 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+#![cfg(feature = "key_auth")]
+
+// Use the shared test framework declared in `tests/multi_write/mod.rs`.
```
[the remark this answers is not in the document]
tvaron3: removed

sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_items.rs
```diff
@@ -0,0 +1,707 @@
...
+            container_client
+                .delete_item(PartitionKey::NULL, &item_id, None)
+                .await?;
+
+            // infinite loop to avoid test flakes due to eventual consistency
+            loop {
+                match container_client
+                    .read_item::<()>(PartitionKey::NULL, &item_id, None)
+                    .await
+                {
```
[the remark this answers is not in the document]
tvaron3: fixed

sdk/cosmos/azure_data_cosmos/Test-Setup.ps1
```diff
@@ -26,7 +36,7 @@ if ($IsWindows) {
     } else {
         LogGroupStart "Installing Cosmos DB Emulator"
         & "$PSScriptRoot\..\..\..\eng\common\scripts\Cosmos-Emulator.ps1" `
-            -StartParameters "/noexplorer /noui /disableratelimiting /enableaadauthentication /partitioncount=50" `
+            -StartParameters "'/noexplorer /noui /enablepreview /EnableSqlComputeEndpoint /SqlComputePort=9999 /disableratelimiting /partitioncount=50 /consistency=Strong'" `
             -Stage "Install"
```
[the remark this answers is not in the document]
tvaron3: removed extra quotes

eng/dict/rust-custom.txt
```diff
@@ -17,3 +17,12 @@ dylib
...
 cbindgen
+LINUXVMIMAGE
+WINDOWSVMIMAGE
+MACVMIMAGE
+LINUXPOOL
+WINDOWSPOOL
+MACPOOL
+RUSTFLAGS
+nocapture
+LASTEXITCODE
```
heaths: These aren't rust keywords or common words. In fact, rather than add them at all, we should just tell cspell to ignore `eng/pipelines`. Heck, probably at this point we just ignore all of `eng`. It, `.vscode/`, and `.github/` are the source of many problems where spell-checking just isn't worth it.

heaths: Actually, I see that `eng/` already is ignored. `RUSTFLAGS` could stay, but more the rest to `.vscode/cspell.json`'s top-level `words` array. Ideally sorted, but that's mainly for future diffs. Seems cspell doesn't need them sorted.

tvaron3: moved words to .vscode.json sorted and kept RUSTFLAGS

PR #3498, opened 2026-01-08
Add Copilot setup workflow, reorganize

heaths:


heaths:
Replaces #2832 that wasn't getting properly rebased.

.github/instructions/git-commit.instructions.md
```diff
@@ -0,0 +1,19 @@
...
+    title of no more than 50 characters
+
+    multi-line descriptions
+    ```
+
+-   Summarize changes for a title of no more than 50 characters ideally, with a hard stop at 80 characters.
+-   Descriptions can be multiple lines or even paragraphs.
+-   Descriptions may contain markdown.
+-   Descriptions should not repeat a lot of content already found in the changed files.
+-   Summarize changes for each file if no more than 10 files are changed.
```
RickWinter, line 15: Does 10 get too verbose? I'm not opposed to starting with 10 so this is more of a how does this look in practice q?

RickWinter (approved):

.github/instructions/git-commit.instructions.md
heaths: I put this in originally (this is just a moved file) because GitHub Copilot summarizes files anyway. This was to try to limit it so it doesn't produce large tables of largely unnecessary summaries. That said, That only affects GitHub Copilot in VSCode. GitHub Copilot in GitHub ignores this anyway.

heaths: Case in point on this very PR just above:

image

heaths: (That's 11 files)

RickWinter (approved):

PR #3500, opened 2026-01-09
next gen sdk design principles

FabianMeiswinkel:


sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Versioning
+
+Both - RUST driver (azure_data_cosmos_native) and SDK (azure_data_cosmos) should stick to semantic versioning - but the SDK (azure_data_cosmos) will have to be much more careful if/when to accept breaking changes (new major versions), because each major version will have to be supported for a long time. Since the driver as mentioned above won't ever be supported directly we can in theory publish new major versions with breaking changes as often as we want - we just have to realize that we still have to be able to maintain supported SDKs using the driver internally without exposing breaking changes.
```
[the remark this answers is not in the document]
analogrelay: Copilot is right 😅 , even if it's being very nitpicky. It would be good for us to use the more consistent capitalization (Rust) though.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
...
+
+
+
+## RUST driver vs. SDK layering
+
+RUST driver (azure_data_cosmos_native) and RUST SDK (azure_data_cosmos) will have different versioning and supportability stories. 
+
+### Supportability
+
+I think we have to be very clear in documents (and if necessary get legal approval) that we will not support the RUST driver (azure_data_cosmos_native) directly - we will support the SDKs using the driver internally (RUST SDK azure_data_cosmos as well as other languages) - but if a customer or application is using the driver directly this should technically be possible, we should document the API surface area etc. - but we should be very clear that we won't support it. The author of the component using the azure_data_cosmos_native driver should own the supportability story there.
```
analogrelay: I think this is the right approach here. It's not sustainable for us to maintain the driver as a public API. However, it's open-source and we should still treat it as such. It's not hidden API, and external users aren't prohibited from using it, so we should make sure it's well-documented (for our own benefit, as much as for any potential external user).

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
...
...a_cosmos) and the new major versions of SDKs in other languages we should seriously evaluate whether we can change the support lifecycle. Today the Azure SDK supports each major version for at least 3+ years after announcing deprecation. This is a very long time - but on the other hand there is no guarantee for providing hotfixes for any specific minor version - customers are expected to always upgrade to the very latest version if they want to consume any fix. This has caused major support concerns by enterprise customers and at least in Java we have moved to announcing certain minor versions for which we would provide hotfixes if needed (security hotfixes or very critical bug fixes) allowing customers to minimize changes they have to take (no need to update external dependencies etc.). It would probably be good to start a discussion with CELA or whoever needed to see whether we can formalize such a model - basically declaring certain minor versions as LTS (Long Term support) vs. Short-term-support and then be explicitly about the support lifecycle for each . The idea of expecting customers to always jump to the latest version just to consume a hotfix is naive and not justifiable.
```
analogrelay: From working in .NET/ASP.NET, I think we need to be very careful here to balance our own engineering resources alongside customer flexibility. I think the [LTS/STS model .NET has landed on](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core) is a good basis (and it sounds like that's basically what you're advocating for here). The nuance will come in how many active in-support releases we're aiming to maintain.

Ideally, the fact that most of the logic to be patched will live in the driver, which has a more flexible API contract (since a given driver version is only bound by the SDK versions that wrap it, not with any given customer app), should allow us more flexibility there.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
analogrelay: Absolutely agreed here. I could see the driver major version climbing much faster than any given SDK. We could reach a point where the 4.0 Java SDK and the 5.0 .NET SDK are both using version 12.0 of the Driver (just throwing out random version numbers here).

It might be worth adding a specific call-out about the actual breaking API contract the driver has. The driver is bound ONLY by it's API contract to the SDKs that use it. Each SDK binds to a specific major version of the driver and moving to a new major version of the driver does **not** necessarily require a major version bump of the SDK itself.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Versioning
+
+Both - RUST driver (azure_data_cosmos_native) and SDK (azure_data_cosmos) should stick to semantic versioning - but the SDK (azure_data_cosmos) will have to be much more careful if/when to accept breaking changes (new major versions), because each major version will have to be supported for a long time. Since the driver as mentioned above won't ever be supported directly we can in theory publish new major versions with breaking changes as often as we want - we just have to realize that we still have to be able to maintain supported SDKs using the driver internally without exposing breaking changes.
+
+
+
+## Connection modes
+
+An explicit design goal for the next generation of SDKs is to not have to expose direct mode transport. The SDKs should not have to participate in routing requests to the right backend nodes/replica.
```
analogrelay, line 33: This is good to call out. I wasn't 100% sure if we were ready to plant this flag in the ground and defend it :). I fully support this.

@kundadebdatta , related to our earlier conversation, this means we likely _would not_ have Direct mode support in the Rust SDK. Instead, we'd have Gateway 1.0 and Gateway 2.0 (Thin Client) as the two connection modes

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
...
+
+
+
+**PREVIEW-BLOCKER**
+
+My recommendation would be to start with #3 for the initial preview
+
+**GA-BLOCKER**
+
+but then implement #1 for sure (automatically providing #2 for RUST as well) and adding #2 on a need-base in other SDKs. Shifting transcoding from service to RUST driver still has benefits especially for query (network density can be increased), any possible RU-benefits can be realized and we shift the COGs out of the service.
```
analogrelay: Agree with this prioritization. I believe that we could move towards a system where we almost always use binary encoding and use transcoding on the driver side as necessary (eventually moving to a model where SDKs do the transcoding directly to/from user types instead of JSON)

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
...
+- Excluded regions - a mechanism for customers to disallow using specific regions (for entire client or per request)
+
+**PREVIEW-BLOCKER**
+
+My recommendation to normalize on a consistent way to configure this is
+
+- ApplicationRegion support with the service-based proximity determination
+- ExcludedRegion support
+- If we think we need to keep ApplicationPreferredRegion support to allow customers to overwrite the priority it should only be a way to prioritize regions - not by filtering regions that can be used for reads/writes. Basically, if an account has 5 regions (R1 - R5) - and ApplicationPreferredRegions is set to R1, R3 - it should result in effectively still using all 5 regions - but prioritized list would be R1, R3, R2, R4, R5. My preference would be to not expose ApplicationPreferredRegion in the client at all (and rather have an emergency overwrite capability in the service - account specific)
```
analogrelay, line 102: I like this approach. It keeps the mandatory "constructor" surface area minimal. The Application Region would be something the user is expected to pass in (though I think for usability we should _allow_ a null Application Region, but it requires the user to be explicit about it by passing a "none" value in directly rather than just not setting an option).

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
...
+For the next generation of SDKs we have the following options
+
+1) Do not enable automatic retries for writes - even when we still (after documenting broadly for long time that retries for writes should be added by applications) see regular CRI (Customer reported incidents) indicating customers expect us to do these retries in the SDK
+2) Enable automatic retries by default. Maybe allow disabling them? But just the simple way .Net is using
+3) Enable automatic retries by default - optionally allow using SDK-system properties like Java does
+4) Change the APIs - and only expose APIs that are idempotent by nature - similar like ConcurrentDictionary - via CreateOrUpdate etc.
+
+
+
+My 2 cents - while I would love the idea of refactoring APIs to always be idempotent by nature I am afraid that this change would cause too much churn and be an adoption blocker. So, my recommendation would be to implement #3
```
analogrelay, line 149: I mostly agree, but I think it might be worth doing some specific design work here. There are opportunities here for the driver to create scenarios in which retrying writes is safe, even without API changes, and I'd like to explore the various kinds of operation (like you do in the description of the linked Java SDK PR) and run through them.

It might be worth starting with 1 (which I think we already have in the Rust SDK) until we do that design work.

analogrelay: We should also consider if there are any service-side changes that could make this easier. For example, if it were possible for the client to compute a unique fingerprint of a document that would agree with whatever fingerprint the server computed (say, by hashing its content), then the service could provide APIs to allow for more transparent idempotency. Basically, this would be making the Tracking ID concept from the Java SDK into something more first-class.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
...
+
+**GA-BLOCKER**
+
+While I think technically we should keep the diagnostics content opaque (as in no hard model-contract because then we would not be able to iterate on it within a major version flexible enough) the single RUST driver also would provide us the opportunity to have consistent diagnostics - which would make using generative AI (or traditional TSGs) to self-diagnose certain issues much easier. So, we should probably still invest in adding documentation (targeting humans / gen AI - not applications) about the diagnostics content. Setting expectations there is critical to allow iterating on diagnostics content while making better use of the know how.
```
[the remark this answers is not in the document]
FabianMeiswinkel: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/Azure/azure-sdk-for-rust/pull/3500#discussion_r2677155918)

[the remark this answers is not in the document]
FabianMeiswinkel: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/Azure/azure-sdk-for-rust/pull/3500#discussion_r2677155918)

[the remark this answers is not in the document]
FabianMeiswinkel: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/Azure/azure-sdk-for-rust/pull/3500#discussion_r2677155918)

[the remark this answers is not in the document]
FabianMeiswinkel: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/Azure/azure-sdk-for-rust/pull/3500#discussion_r2677155918)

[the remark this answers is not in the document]
FabianMeiswinkel: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/Azure/azure-sdk-for-rust/pull/3500#discussion_r2677155918)

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
...
+
+My 2 cents - while I would love the idea of refactoring APIs to always be idempotent by nature I am afraid that this change would cause too much churn and be an adoption blocker. So, my recommendation would be to implement #3
+
+
+
+## Diagnostics
+
+### Motivation
+
+For all cases where we will wrap the RUST driver in another SDK (Python, Go, Java, .Net) we have to assume that the application developer has no context/knowledge of RUST. We can't expect them to debug into the RUST code etc. This means the native C-APIs in the RUST driver we use have to act as a black-box. If any non-happy-path situation happens (errors but also increased latency etc.) we have to return enough context to allow debugging the root cause. Luckily we do not plan to support direct mode - so, the service interactions are simpler.
```
analogrelay: This also makes it very important for us to be as "panic-free" as possible. I'll be doing some research into Rust's support for cross-language unwinding (which _does_ exist: https://doc.rust-lang.org/reference/items/functions.html#unwinding) but I doubt it will yield anything usable for a while.

Rust does provide a way to [catch all panics within Rust code](https://doc.rust-lang.org/stable/std/panic/fn.set_hook.html), so we can likely use that to provide some better diagnostics as well.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
analogrelay: I also think there are opportunities to build some traditional tools (not just gen AI) here, even just to "visualize" a request flow. Consistent diagnostics here will be a big win

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
...
+
+In the RUST driver we need to design the native API in a way that allows a similar model. We have to expose expose enough info in a typed contract to allow making the decision whether to serialize the full diagnostics or not (usually this would happen on error, when latency or RU-usage exceeds certain thresholds and/or based on sampling) - but in general we would probably want to keep the actual diagnostics content as an opaque string.
+
+**PREVIEW-BLOCKER**
+
+My recommendation would be to use a similar model as in the Java SDK - CosmosDiagnosticsContext.java - it should have enough info to make the decision in the place immediately around the native API to decide whether to serialize the opaque diagnostics string or not. If we are time-wise not able to fit this into preview maybe we can start by always serializing the diagnostics into an opaque string - that way later when doing it lazily we would improve perf - and not have to worry avoid breaking perf regressions when starting to expose the diagnostics info we absolutely have to expose eventually to allow debugging.
```
analogrelay, line 181: Here's my rough thinking, and it's similar to yours. Every request would hold a `DiagnosticsBuilder` which, if specified, collects diagnostics data for a request as it moves through our pipeline. After the request completes, this is transformed to a `RequestDiagnostics`. The `RequestDiagnostics` type is `serde::Serialize` and there's a canonical serialization to JSON. We can expose this value across the ABI in mutliple ways (which can be configured by the user):

1. As an opaque pointer on the [`CallContext`](https://github.com/Azure/azure-sdk-for-rust/blob/030588fa3cf76a8db199566ceed8f4755c734791/sdk/cosmos/azure_data_cosmos_native/src/context.rs#L31). The purpose of the Call Context is to provide this "side-channel" related to a specific all. Providing it as an opaque pointer is extremely cheap (since we have the object already) but mostly useless to the SDK as-is. However, we'd also provide an API to serialize this to a JSON string. That allows us to expose something that the application can _choose_ to serialize whenever they wish.
2. WHEN enabled by the user (again, using a flag in the `CallContext`), we can also emit it as a pre-serialized JSON string to allow for a simple flow.
3. If tracing at the appropriate level and target is enabled, we can emit the JSON to OTEL as well (at the end of the call), allowing for existing OTEL users to get this value "for free" using their OTEL configuration to guide when it's emitted.

If we want, we can also provide a way to signal to the driver that diagnostics _should not even be collected_ (in highly performance-sensitive environments, though I don't know that's likely to be an issue given the network I/O).

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
...
...e number of time series you are looking at can be overwhelming. So, only emitting metrics with PartitionId dimension for requests violating certain thresholds like latency or statusCode can help to significantly reduce the number of concurrent timeseries in any time-interval - also applying hard-limits can be useful. The same is true for logs or traces - if a handful of requests is slow or failing you want as much debugging info as possible - but if there is an outage - and 30 percent of your requests are failing logging details for all of them can result in your app seeing much higher impact than needed because of resource exhaustion due to noisy logging/tracing. So, applying thresholds and hard guard-rails which are configurable and designing how-to do so at runtime without restarts is essential - we need to design the diagnostics capturing around these requirements, because otherwise collecting diagnostics when it is needed most - for very large workloads - becomes too challenging. The Java SDK in combination with Spark connector are a pretty good starting point for how thresholds and guard-rails could be applied - but this will need more investigation/work than just porting it.
```
analogrelay, line 193: There are likely some existing tools in the Rust OTEL ecosystem for this we should explore. But in general, I'm in full agreement here.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
+# Design principles for next-generation SDKs
+
+This document is an attempt to write down some design principles for the next generation of SDKs. The main focus is to call out intended changes to default behavior, which would be considered breaking changes within a major version of existing SDKs. With the next generation of new major SDK versions we have the opportunity to make these changes - but it is still a trade-off - any "breaking" change increases the adoption barrier - so, we have to evaluate benefits of applying new default behavior vs. the adoption risk.
+
+
+
+## Rust driver as foundation
+
+At this point the plan is to use the native C-level APIs provided in the Rust driver (azure_data_cosmos_native) to implement the communication with the service - and wrap these native APIs in the SDKs for Java, .Net, Go, Python (and Rust)
```
Pilchie, line 9: Also likely JS - at least for server based scenarios, and maybe browser via compilation to WASM.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
Pilchie: I don't understand how we can "do the transcoding directly to/from user types" while also supporting all the features of serialization libraries like renaming, etc.  I feel like we're likely to always need a json intermediary somewhere for languages with rich serialization libraries (.NET/Java/Rust at least).

Pilchie: But maybe it can stream the json at least, instead of buffering the whole document?

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
Pilchie: For at least some Azure compute environments, it's possible to determine the application region from the host. Would we do that automatically when we can, or still require it to be passed in?

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Configuration
+
+**PREVIEW-BLOCKER** (probably also debatable whether sufficient for GA - same argument as for Thresholds above)
+
+In both .Net as well as Java SDK we have a series of config knobs that are not exposed in public API but can be tuned as a fail-safe via environment variable (or system properties in Java). For the RUST driver we need to expose a clean API to modify these internal plan-B config overrides - how that gets exposed in each SDK really depends on the typical config approach there - but the APIs need to exist to do this cleanly - and we need to have documentation for all the internal knobs in place. Part of the work should be that diagnostics (for example shown in errors) should include a list of all config overrides - and we should also be specific in documentation whether using any such knob voids SLAs/supportability
+
+
+
+## Authentication
```
tvaron3, line 225: Would it make sense to have another section on session container? Whether we should have it or other api 's we could expose for managing session tokens.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
+# Design principles for next-generation SDKs
+
+This document is an attempt to write down some design principles for the next generation of SDKs. The main focus is to call out intended changes to default behavior, which would be considered breaking changes within a major version of existing SDKs. With the next generation of new major SDK versions we have the opportunity to make these changes - but it is still a trade-off - any "breaking" change increases the adoption barrier - so, we have to evaluate benefits of applying new default behavior vs. the adoption risk.
```
jeet1995, line 3: From an object-oriented design perspective, should API contracts be kept identical to reduce this friction of adoption? This is more to do with downstream Cosmos DB SDKs (non-Rust).

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Versioning
+
+Both - Rust driver (azure_data_cosmos_native) and SDK (azure_data_cosmos) should stick to semantic versioning - but the SDK (azure_data_cosmos) will have to be much more careful if/when to accept breaking changes (new major versions), because each major version will have to be supported for a long time. Since the driver as mentioned above won't ever be supported directly we can in theory publish new major versions with breaking changes as often as we want - we just have to realize that we still have to be able to maintain supported SDKs using the driver internally without exposing breaking changes.
```
jeet1995, line 27: What will be the release model for individual SDKs when changes are made to `azure_data_cosmos_native`? Is a "shared package model" applicable here?

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,220 @@
...
+
+It is likely that we will also have to provide at least one option for customers who can't use HTTP/2 for whatever reason - and the same option would also help us in cases where accounts are not yet migrated form Routing Gateway to Compute Gateway
+
+- HTTP/1.1 with REST contracts against ComputeGateway **GA-BLOCKER**
+
+
+
+## Timeouts
+
+In current SDKs today we have hard-coded default timeouts (5-6s connect timeout, 5-6s request timeout for direct, 30 seconds for gateway). It is possible to override some of these timeouts and we also allow customers to specify E2E timeouts in Java, Python and implicitly via Cancellation Token in .Net.
```
jeet1995, line 56: Where will the timeout logic sit? `azure_data_cosmos_native` would be theoretically better to prevent divergence of these timeouts. Uber point being in determining what state should reside in downstream SDKs vs the Rust engine.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
jeet1995: Do we need `ExcludedRegion` support at all and replace it with a per-request `ApplicationPreferredRegions` list?

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
analogrelay: Some platforms, like Rust in particular, abstract the renaming/metadata from the actual format decoding. For example, serde defines all the logic for renaming properties and describing the "shape" of a serialized object, and serde_json just follows the serialization contract that serde defined. In that case, we'd be able to serialize/deserialize in the binary format without the user being exposed to that and without a JSON intermediary. Serde is the intermediary in that case, and it's the Serde data model that we're working with rather than the JSON data model. 

It's more like the DataContract serialization from WCF, where the shape of the serialized object is separate from the specific JSON/XML/etc encoding.

There are similar things that can be done in other languages but it is certainly trickier to do in a language like .NET without involving a JSON library somewhere.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
FabianMeiswinkel: +1 - Databricks IMO also has a reaosnable trade-off - even Spring - we will need separate design documents for several of these topics - I just wanted to get the areas listed and see when we need to close on them - and ideally have some assignments on who will drive each topic.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
FabianMeiswinkel: Correct - this is one of the gray areas of semantic versioning - because whether new major versions of dependencies justify always new major versions of the outer component or not is not always clear. In the driver model we cannot ever leak API of the driver package in any SDK-publci surface area - this way we are only bound by bheavior changes to decide whether a driver change requires a new major SDK version - and this is something we would carefully avoid of course in vast majority of cases.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
FabianMeiswinkel: In the driver we cannot avoid tranbscoding. IN SDKs we could - in Java for example the bianry encoding implementation i staretd would have simply been a Jackson JsonReader+JsonWriter implementation. .Net when not using StreamAsync APIs uses a very similar model. But it also really depends on how SDKs expose serialization customization options - any Stream/byte[] based APIs like in .Net the StreamAsync APIs of course reuqire transcoding to Json somewhere.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
FabianMeiswinkel: Honestly I thik we should expect to be panic-free. Any panic should result in crash of the entire app - which is very impactful and should be an absolute outlier. Anything else would only be a bandaid - and none that you can really support with cross language support IMO.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
FabianMeiswinkel: I think our ideas go in a similar direction - my point is there will be some configs to decide when to collect full diagnostics - there will probably be no sceanrios never creating diagnostics and only playgroudn workloads always colelcting them - designing the guard rails to control reaosnable perf tarde-offs but still alllowing us to support customers is something we need to invest on early (even before releasing any pre-view to customers)

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
FabianMeiswinkel: I highly doubt that the built-in OTel tools help - at least in Java and .Net all Otel does is allow sampling after the fact - so, the collector can decide what sample-percentage to send to some monotiroing system like AppInsights. Some of the thresholds mentioned above (and realized in Java) are ways to help use guard-rails to make decisions whether to veen produce diagnostics.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
FabianMeiswinkel: Yes - identified that I missed that topic under the shower this morning and added that section :-)

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
FabianMeiswinkel: I think that really depends on each SDK. In Java for example there are some Cdentral team ideas that would force us to change publci API - for example get rid of Reactor and stick to native Future etc.

FabianMeiswinkel: But we should have good reasons for any breakign change - knowing that each will increase the adoption barrier. Although with gen AI today ithe API level breakign changes are honestly the leats of our problems - the breaking behavior changes are much harder to build trust around.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
FabianMeiswinkel: Agreed - should be in azure_data_cosmos_native

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
analogrelay: Yep, the main reason I'm thinking about panic recovery is as a backstop for unexpected panics, since even with Rust we need some engineering discipline to stay panic-free. Rust gives us a lot of tools to stay pretty much entirely panic-free (aside from truly random failures like hardware/OS faults, which are beyond our scope)

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
kirankumarkolli: Browser scenarios we can scope out for now.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,240 @@
...
+
+
+
+## Rust driver vs. SDK layering
+
+Rust driver (azure_data_cosmos_native) and Rust SDK (azure_data_cosmos) will have different versioning and supportability stories. 
+
+### Supportability
+
+I think we have to be very clear in documents (and if necessary get legal approval) that we will not support the Rust driver (azure_data_cosmos_native) directly - we will support the SDKs using the driver internally (Rust SDK azure_data_cosmos as well as other languages) - but if a customer or application is using the driver directly this should technically be possible, we should document the API surface area etc. - but we should be very clear that we won't support it. The author of the component using the azure_data_cosmos_native driver should own the supportability story there.
```
kirankumarkolli, line 19: One scenario is new language integration, is possible that 3P might take lead and we might need to help/support them.

Public message is to NOT support it but we might selective have exceptions.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,240 @@
...
...a_cosmos) and the new major versions of SDKs in other languages we should seriously evaluate whether we can change the support lifecycle. Today the Azure SDK supports each major version for at least 3+ years after announcing deprecation. This is a very long time - but on the other hand there is no guarantee for providing hotfixes for any specific minor version - customers are expected to always upgrade to the very latest version if they want to consume any fix. This has caused major support concerns by enterprise customers and at least in Java we have moved to announcing certain minor versions for which we would provide hotfixes if needed (security hotfixes or very critical bug fixes) allowing customers to minimize changes they have to take (no need to update external dependencies etc.). It would probably be good to start a discussion with CELA or whoever needed to see whether we can formalize such a model - basically declaring certain minor versions as LTS (Long Term support) vs. Short-term-support and then be explicitly about the support lifecycle for each . The idea of expecting customers to always jump to the latest version just to consume a hotfix is naive and not justifiable.
```
kirankumarkolli, line 23: Lets explore this further. 

At-least the initial versions IMO lets not burden ourselves with LTS and backporting. Once we reach a mature state we can explore it.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
```diff
@@ -0,0 +1,240 @@
...
+
+### Versioning
+
+Both - Rust driver (azure_data_cosmos_native) and SDK (azure_data_cosmos) should stick to semantic versioning - but the SDK (azure_data_cosmos) will have to be much more careful if/when to accept breaking changes (new major versions), because each major version will have to be supported for a long time. Since the driver as mentioned above won't ever be supported directly we can in theory publish new major versions with breaking changes as often as we want - we just have to realize that we still have to be able to maintain supported SDKs using the driver internally without exposing breaking changes.
```
kirankumarkolli, line 27: Its very very important aspect. 

It's unrealistic to expect that Driver will move forward with any breaking change and its language responsibility.  What are the scenarios where Driver major versions are needed with-out language major version?

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
kirankumarkolli: We can start with this assumption for now. 
I suppose some these principles can be tuned/revisited on-demand in future.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
analogrelay: I call some of this out in my related document in #3502, but some of this still needs to be expanded on.

The driver has some unique tools it can use that we haven't been able to use in the past:

1. We know of and own every single supported consumer of the Driver.
2. A particular SDK version is always coupled to a specific `major.minor` Driver (perhaps `major.minor.patch` too, but that's TBD).
3. The SDK is the only thing that exposes a (supported) customer-facing API.

That allows us to update SDKs to take on new major versions of the Driver within major, minor, or even patch releases, IF the SDK can do so in a way that has no upstream customer impact.

Consider the most trivial kind of breaking change: An API rename. We accidentally shipped the API `cosmos_container_cretate` in Driver v1.0 and didn't notice it until we released a supported Go SDK v2.0 using it (to implement `ContainerClient.Create`). Renaming this to `cosmos_container_create` in the Driver is a breaking change[^1], and from a SemVer perspective, necessitates creating Driver v2.0. As long as we update the **Go SDK** so that it calls the correct driver API (`cosmos_container_create`), this change is non-breaking to SDK customers, and could be shipped in v2.0.1 of the Go SDK (or 2.1.0, or 3.0.0, whichever we prefer).

[^1]: Unless we add a shim with the old name but let's say we don't because that's more illustrative here

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
FabianMeiswinkel: Agreed - but 3P should support their SDK - and we might need second-level support to them in this case?

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
FabianMeiswinkel: Agreed - my main point is that we need to start the conversation how we can even apply an end-of-support lifecycle for individual minor versions - it doesn't exist today

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
FabianMeiswinkel: Agreed - exposed APIs in driver will have breaking changes - since the only supported users are our SDKs that is fine - as long as thier public surface area does not change. Similar for behaviro changes - although at that point we are back to opt-in only for any possibly breaking behavior changes. So, it is still super important to go over these possibly breaking changes we think we need to make soon.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
analogrelay: Yep, absolutely. I actually _do_ see this as an explicit benefit of the Driver approach. We do have _something_ we can point users of `rand(Language)` at and say "look, we won't give you formal support, but this is something that works and could be adapted into your language". We need to be clear about that, and we can even provide _as available_ "community" support for that. It's not like we have to immediately close and lock every issue asking about the driver, but we do need to make very clear what our official support policy is.

sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md
analogrelay: Rust's tracing infrastructure primarily uses "head-based sampling", where sampling decisions are made before any traces would be emitted in a given scope. I completely agree we should do some research here and only use the tools that work for our environment, but I also know that quite a bit of effort has been put into high-efficiency tracing in Rust and I wouldn't want us to disregard that.

PR #3502, opened 2026-01-09
cosmos: initial design doc structure and versioning/support doc

analogrelay:
This PR adds a few documents to describe the Cosmos DB Native Driver approach and lay out the versioning/support policy_between the SDK and the Driver only_ (it does not cover the support policy between the SDK and users)


sdk/cosmos/docs/driver/VersioningAndSupport.md
```diff
@@ -0,0 +1,89 @@
...
+
+The Native Driver follows Semantic Versioning (SemVer) principles, but as an internal component, it has more flexibility to make breaking changes, as long as those can be adopted by wrapping SDKs without breaking their own versioning guarantees.
+
+Thus, the versioning contract for the Native Driver is between the Native Driver and the wrapping SDKs, not between the Native Driver and end-users. Wrapping SDKs take a dependency on a specific major version of the Native Driver, and can choose when to upgrade to a new major version, as long as the SDK team is willing to take on the support burden of doing so.
+
+Since the SDK is often deployed _with_ a copy of the Native Driver (e.g., via static linking), breaking changes in the Native Driver do not necessarily impact end-users, as long as the wrapping SDK can continue to function correctly. However, there may be dynamic linking scenarios where a user will want to upgrade the Native Driver independently of the SDK (such as for a security patch). Thus we should strive to follow a standard SemVer approach as much as possible.
```
Pilchie, line 24: Hmm - I hadn't expected we would want to support end-users upgrading the Native Driver independently of the SDK.

analogrelay: Yeah, I'm quite hesitant for that but I wanted to point out the places it could be done.

I was thinking of security patches when I wrote that but honestly, I think we'd service each SDK (updating its embedded version of the driver) when we hit that kind of issue.

sdk/cosmos/docs/README.md
```diff
@@ -0,0 +1,26 @@
...
+## Azure Cosmos DB Native Driver
+
+Cosmos DB clients are complex. The client is responsible for managing connections to backend replicas, seeking quorum, handling cross-regional failover and retry, hedging requests, and much more. Implementing all of this logic in every language SDK already leads to significant duplication of effort, and makes it difficult to ensure consistency across SDKs. The Native Driver seeks to solve this problem by providing a single, shared implementation of the core Cosmos DB client logic that can be used across multiple language SDKs.
+
+The Cosmos DB Native Driver is implemented in Rust, and exposes two FFI-compatible interfaces:
+
+1. A low-level C-compatible interface, which can be used directly from C and C++ applications, or via FFI bindings in other languages.
+2. A Rust-native interface, designed to be used by the higher-level Azure SDK for Rust Cosmos DB client.
+
+The Native Driver can be found in the `azure_data_cosmos_driver` crate.
```
analogrelay, line 14: I gave the Native Driver a different name (from the current azure_data_cosmos_native) because I think we might actually want to consider building two separate pure-Rust crates here, one containing the driver and one containing the SDK (which depends on the driver). We can discuss that more later though

sdk/cosmos/docs/driver/VersioningAndSupport.md
kirankumarkolli: Multi version library in same app (not fully sure on real-scenario) is a possible scenario.

sdk/cosmos/docs/driver/VersioningAndSupport.md
```diff
@@ -0,0 +1,89 @@
...
+
+So, for any given SDK version (`X.Y`), the native driver major version it depends on (`A`) is fixed. The SDK can choose to upgrade to a new major version of the Native Driver (`B`) in a future SDK release (`X.Z`), but this is not automatic.
+
+### Versioning Rules
+
+1. **Patch releases** of the Native Driver (e.g., `A.B.C` to `A.B.D`) can be adopted by any new version (major, minor, patch) of the wrapping SDK without breaking versioning guarantees. This allows for bug fixes and security patches to be rolled out without requiring SDK upgrades. In **addition**, patch releases of the Native Driver can be installed directly by end-users in dynamic linking scenarios, even if the wrapping SDK is not upgraded. The SDK guarantees that it will continue to function correctly with any patch version of the Native Driver within the same major.minor version.
```
kirankumarkolli, line 30: >This allows for bug fixes and security patches to be rolled out without requiring SDK upgrades

Can you please expand more?

kirankumarkolli: Its clarified later as a possibility and necessarily might not be end-user scenario.

analogrelay: I should clarify it a bit. The intent is to clarify that we can general adopt a patch release in any category of SDK update.

sdk/cosmos/docs/driver/VersioningAndSupport.md
FabianMeiswinkel: Following standard semVer is still a good idea.

FabianMeiswinkel: My 2 cents - while I agree with the goal above (and in theory being able to update native driver as logn as we stick to semver) I would call it out explicitly as a non-goal form supportability persepctive. If we support it, we have to test it - and this would blow up the test matrix unnecessarily. So, keep the semver reference for our own sanity - but we should only support combinations we have thoroughly tested.

PR #3504, opened 2026-01-09
[Storage] `StorageError` error type

vincenttran-msft:
This PR introduces the `StorageError` error model, which allows for a cleaner programmatic access and `Display` experience for error parsing.

### DIFF
--- a/sdk/storage/azure_storage_blob/assets.json
+++ b/sdk/storage/azure_storage_blob/assets.json
@@ -1,6 +1,6 @@
 {
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "rust",
-  "Tag": "rust/azure_storage_blob_3d3cb80960",
+  "Tag": "rust/azure_storage_blob_1a51a711de",
   "TagPrefix": "rust/azure_storage_blob"
 }


sdk/storage/azure_storage_blob/src/models/storage_error.rs
```diff
@@ -0,0 +1,248 @@
...
+    /// The HTTP status code.
+    pub status_code: azure_core::http::StatusCode,
+    /// The Storage error code, if available.
+    pub error_code: Option<StorageErrorCode>,
+    /// The error message, if available.
+    pub message: Option<String>,
+    /// The request ID from the x-ms-request-id header, if available.
+    pub request_id: Option<String>,
+    /// The raw HTTP response.
+    pub raw_response: RawResponse,
```
vincenttran-msft: Remove it 😄

sdk/storage/azure_storage_blob/src/models/storage_error.rs
```diff
@@ -0,0 +1,248 @@
...
+                StorageError::from_xml(*status, raw_response.as_ref().clone())
+            }
+            // TODO: We may have to handle other ErrorKind variants, but catch-all for now.
+            _ => Err(azure_core::Error::with_message(
+                azure_core::error::ErrorKind::DataConversion,
+                "ErrorKind was not HttpResponse and could not be parsed.",
+            )),
+        }
+    }
+}
```
vincenttran-msft: Let's make it `Option<StorageError>` so we can return `None` here.

sdk/storage/azure_storage_blob/src/models/storage_error.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// A general reason for the error, if available.
+    reason: Option<String>,
+    /// Additional authentication error details, if available.
+    authentication_error_detail: Option<String>,
+    /// Details about signature mismatch errors, if available.
+    signature_mismatch_error_detail: Option<String>,
+    /// The name of the query parameter that caused the error, if available.
+    query_parameter_name: Option<String>,
+    /// The value of the query parameter that caused the error, if available.
+    query_parameter_value: Option<String>,
```
vincenttran-msft: This is what I landed on from our original list. Maybe I would also add:

- ExtendedErrorDetail
- OAuthAuthorizationErrorDetail
- ACLAuthorizationErrorDetail
?

sdk/storage/azure_storage_blob/src/models/storage_error.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// The HTTP status code from the copy source, if available.
+    copy_source_status_code: Option<String>,
```
jalauzon-msft: Maybe this should be of type `StatusCode`?

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -808,3 +810,130 @@ async fn test_immutability_policy(ctx: TestContext) -> Result<(), Box<dyn Error>
...
+    );
+    assert!(
+        storage_error.request_id().is_some(),
+        "Expected request_id to be populated from headers."
+    );
+    assert!(
+        storage_error.additional_error_info().is_empty(),
+        "Expected no additional_error_info for bodiless response."
+    );
+
```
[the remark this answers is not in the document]
vincenttran-msft: This forces an error due to the scenario being the specified blob doesn't exist (and in this case, as well as the container).

sdk/storage/azure_storage_blob/src/models/error.rs
```diff
@@ -0,0 +1,266 @@
...
+        status_code: azure_core::http::StatusCode,
+        error_code: Option<StorageErrorCode>,
+        request_id: Option<String>,
+        raw_response: RawResponse,
+    ) -> Result<Self, azure_core::Error> {
+        #[derive(Deserialize)]
+        #[serde(rename = "Error")]
+        struct StorageErrorXml {
+            #[serde(rename = "Code")]
+            code: Option<String>,
```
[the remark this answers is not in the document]
vincenttran-msft: We need it as part of the XML so that when we parse the XML fields that it gets captured and mapped to `code`. Otherwise it would fall into `additional_fields` which is undesirable.

sdk/storage/azure_storage_blob/src/models/mod.rs
```diff
@@ -3,6 +3,7 @@
 
 pub(crate) mod content_range;
 mod extensions;
+mod storage_error;
```
heaths: Call this module `error` and export it publicly i.e.,
```suggestion
pub mod error;
```

You can re-export `error::StorageError` from the crate root.

See the [guidelines](https://azure.github.io/azure-sdk/rust_introduction.html#rust-errors-crate-definitions).

sdk/storage/azure_storage_blob/src/models/error.rs
```diff
@@ -0,0 +1,266 @@
...
+    copy_source_status_code: Option<azure_core::http::StatusCode>,
+    /// The error code from the copy source, if available.
+    copy_source_error_code: Option<String>,
+    /// The error message from the copy source, if available.
+    copy_source_error_message: Option<String>,
+    /// Additional fields from the error response that weren't explicitly mapped.
+    additional_error_info: HashMap<String, String>,
+}
+
+impl StorageError {
```
heaths, line 38: Accessors aren't idiomatic to rust. Why are you exposing these? If the dev wants to change a field for some reason, why prevent them? It can't hurt anything.

sdk/storage/azure_storage_blob/src/models/storage_error.rs
```diff
@@ -0,0 +1,266 @@
...
+impl std::error::Error for StorageError {}
+
+impl TryFrom<azure_core::Error> for StorageError {
+    type Error = azure_core::Error;
+
+    fn try_from(error: azure_core::Error) -> Result<Self, Self::Error> {
+        match error.kind() {
+            ErrorKind::HttpResponse {
+                status,
+                error_code: _,
```
heaths: Tip: to just match any other fields, add `..` at the end e.g.,

```rust
ErrorKind::HttpResponse {
    status,
    raw_response,
    ..
}
```

sdk/storage/azure_storage_blob/src/models/storage_error.rs
```diff
@@ -0,0 +1,266 @@
...
+                raw_response,
+            } => {
+                let raw_response = raw_response.as_ref().ok_or_else(|| {
+                    azure_core::Error::with_message(
+                        azure_core::error::ErrorKind::DataConversion,
+                        "Cannot convert to StorageError: raw_response is missing.",
+                    )
+                })?;
```
heaths: While I appreciate this probably wouldn't happen, it's weird to err when converting to an err. Could you either return a `StorageError` with less information (ideal), or you could match e.g.,

```rust
ErrorKind::HttpResponse {
    status,
    raw_response: Some(raw_response),
    ...
}
```

In which case, it'll only match if `raw_response` is `Some(RawResponse)` and then the `raw_response` you get is what was in `Some(RawResponse)`.

sdk/storage/azure_storage_blob/examples/storage_error.rs
```diff
@@ -0,0 +1,140 @@
...
+//! # Prerequisites
+//!
+//! - Set the `AZURE_STORAGE_ACCOUNT_NAME` environment variable to your storage account name
+//! - Authenticate using Azure CLI: `az login`
+//!
+//! # Usage
+//!
+//! ```bash
+//! az login
+//! $env:AZURE_STORAGE_ACCOUNT_NAME="<your-storage-account>"
```
heaths: This is not how you set env vars in `bash` (which is what you said the region is, which is what we generally use in Rust) and will fail.
```bash
AZURE_STORAGE_ACCOUNT_NAME="<your-storage-account>"
```

sdk/storage/azure_storage_blob/examples/storage_error.rs
```diff
@@ -0,0 +1,140 @@
...
+                if !additional_info.is_empty() {
+                    println!("\n=== Additional Error Information ===");
+                    for (key, value) in additional_info {
+                        println!("  {}: {}", key, value);
+                    }
+                }
+
+                // StorageError implements Display for easy logging
+                println!("\n=== Full Error Display ===");
+                println!("{}", storage_error);
```
heaths: IMO: you did all this to have a great `Display` impl. Why not show that off first rather than show the customer how to pick apart details when most probably won't need to?

heaths (changes requested):
Not exported per guidelines, but I also have some thoughts on how it's used.

sdk/storage/azure_storage_blob/src/models/storage_error.rs
vincenttran-msft: Yeah we did think this was abit odd when reviewing as well. I just went with returning a skeleton `StorageError`

sdk/storage/azure_storage_blob/src/models/error.rs
```diff
@@ -257,6 +213,26 @@ impl TryFrom<azure_core::Error> for StorageError {
...
+                    status_code: *status,
+                    error_code: None,
+                    message,
+                    request_id: None,
+                    reason: None,
+                    authentication_error_detail: None,
+                    copy_source_status_code: None,
+                    copy_source_error_code: None,
+                    copy_source_error_message: None,
+                    additional_error_info: HashMap::new(),
```
heaths, line 233: Tip: you could just use:

```rust
status_code: *status,
message,
additional_error_info: HashMap::new(),
..Default::default()
```

vincenttran-msft: Ah, I believe I did try that but the issue being this:
>the trait bound `azure_core::http::StatusCode: std::default::Default` is not satisfied
the trait `std::default::Default` is not implemented for `azure_core::http::StatusCode`

Issue #2758, opened 2025-07-03
Support rehydrating the Poller


heaths:
Splitting up #2458 into different tasks in priority order. Using the serialization format in <https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/RehydrationToken.cs> - which the architects agreed to sometime back but only .NET seems to support currently - we should support rehydrating the Poller based on [general guidelines](https://azure.github.io/azure-sdk/general_design.html#general-lro-continuation).

PR #3506, opened 2026-01-10
Make Poller public API consistent with Pager

### LINKED ISSUE #3503: Make Poller API consistent with Pager API
Splitting up #3423 for the API and implementation. Prior to GA, at least, we should have a consistent API between `Pager` and `Poller` for ease of use and maintenance. We can focus on any implementation changes later and optionally, since the current implementation has been working well. As we prepare for release, I don't want to risk the stability unnecessarily but we need to nail down the public API surface even for use only intended for other Azure SDK for Rust crates.

heaths:
Resolves #3503

### DIFF
--- a/sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
+++ b/sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
@@ -166,8 +166,8 @@ impl CertificateClient {
         let parameters: Body = parameters.into();
 
         Ok(Poller::new(
-            move |next_link: PollerState<Url>, poller_options| {
-                let (mut request, next_link) = match next_link {
+            move |poller_state: PollerState<Url>, poller_options| {
+                let (mut request, next_link) = match poller_state {
                     PollerState::More(next_link) => {
                         // Make sure the `api-version` is set appropriately.
                         let qp = next_link


sdk/core/azure_core/CHANGELOG.md
```diff
@@ -16,12 +16,18 @@
...
 - Changed `Pager::from_callback` to `Pager::new` which now requires the caller to return a `Pin<Box<dyn Future>>`.
+- Changed `Poller::from_callback` to `Poller::new` which now requires the caller to return a `Pin<Box<dyn Future>>`.
+- Changed `PollerResult::next` field to `PollerResult::continuation_token`.
+- Changed `PollerState` generic parameter from `N` to `C`.
 - Moved `BearerTokenAuthorizationPolicy` into `azure_core::http::policies::auth`.
 - Removed `ItemIterator::with_continuation_token()`. Pass a continuation token to `PagerOptions::continuation_token` instead.
 - Removed `PageIterator::with_continuation_token()`. Pass a continuation token to `PagerOptions::continuation_token` instead.
 - Removed `Pager::from_stream`.
+- Removed `Poller::from_stream`.
+- Renamed `BoxedFuture` to `PagerResultFuture` in `pager` module.
```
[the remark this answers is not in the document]
heaths: That was net new. No need to mention. Doesn't really rise to "Added Features" either as it's a pretty small implementation detail that shouldn't even be noticed since you don't refer to it directly.

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
```diff
@@ -218,7 +218,7 @@ impl CertificateClient {
                         PollerStatus::InProgress => PollerResult::InProgress {
                             response: rsp,
                             retry_after,
-                            next: next_link,
+                            continuation_token: next_link,
```
RickWinter, line 221: Why the name change? Will it always be a token. Is the "next_link" a legacy name that should have been something else?

heaths: It's either a next link or continuation token, as the docs have always said. I changed it to be consistent. Keep in mind that with #2758 - matching Go's implementation that I'm told was the architect-agreed serialization format - the DX APIs will sport a continuation token. While this is probably more often a next link, I felt keeping it consistent across `Pager` and `Poller` would be apt. I can change it back if you prefer.

I also felt for both `Pager` and `Poller`, "continuation token" was less specific than "next link". That's always a link. That said, the "token" part of "continuation token" also seems a bit specific. We could call both just "continuation" if that sounds better.

/cc @LarryOsterman @analogrelay

LarryOsterman (approved):

sdk/keyvault/azure_security_keyvault_certificates/src/clients.rs
heaths: Several of us chatted offline and I looked at some other SDK languages' source - which is all over the place - so in lieu of anything better, I'll stick with `continuation_token`. It seems more generic than `next_link` - a sentiment shared previous a while back with a couple others - and even though `_token` makes it seem a little implementation-specific, just `continuation` doesn't make sense - feels "lacking something".

Most Azure services do seem to use a "next link", FWIW, but hopefully with docs "continuation token" as a more generic concept will suffice.

PR #3515, opened 2026-01-12
Cosmos: Remove query engine and provide our own query result pager

### LINKED ISSUE #3413: QueryExecutor needs to implement its own Stream
With the changes introduced in #3372 we no longer store a `Stream` in pagers, so there's no good way to implement the `FeedPager<T>` as a "specialization" of `ItemIterator<P, C>`.

After discussing offline, I've commented out the body of `QueryExecutor::into_stream` and terminated with `unimlemented!()`. From the sounds of it, seems like just using your own stream implementation would be more flexible anyway.

analogrelay:
Fixes #3413 

When @heaths rewrote the pager, it broke how we were doing cross-partition query pagers. At the time, I agreed that it would be reasonable here for Cosmos to have its own pagination logic, since it needs to handle these cross-partition scenarios, and we already use our own `FeedPage<T>` type instead of `Response<T>`. At the time, Heath just commented out the old code and marked it `unimplemented!()` (at my suggestion). Now, I'm coming back to restore it, and bringing a few refactorings along with it:

* I moved Throughput Offer APIs to a crate-local `OffersClient`, which matches what we do in some other SDKs. This isn't visible to users, it's just used _by_ `ContainerClient`/`DatabaseClient` to implement throughput-related operations. This removes the throughput stuff from `CosmosPipeline`, which was clunky anyway
* I brought the "gateway-based" query execution logic in to `QueryExecutor`, which is now an enum with two different executor implementations: Gateway and QueryEngine.
* I moved the `send_query_request` logic out of `CosmosPipeline` and now the places that need to issue these queries construct a `QueryExecutor::gateway` and execute it. The result is the same code savings, but better architecture IMO.

I also restored the query engine tests, which are all passing for me.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
+++ b/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Bugs Fixed
 
 ### Other Changes
+- Use new `FeedItemIterator` for query operations. ([#3515])(https://github.com/Azure/azure-sdk-for-rust/pull/3515)
 
 ## 0.30.0 (2026-01-21)
 


sdk/cosmos/azure_data_cosmos/src/feed.rs
```diff
@@ -107,17 +107,173 @@ impl<T: DeserializeOwned> FeedPage<T> {
...
-
 /// Represents a stream of pages from a Cosmos DB feed.
 ///
 /// See [`FeedPage`] for more details on Cosmos DB feeds.
-pub type FeedPager<T> = ItemIterator<FeedPage<T>, String>;
+#[pin_project::pin_project]
+pub struct FeedItemIterator<T: ConditionalSend> {
+    #[pin]
+    #[cfg(not(target_arch = "wasm32"))]
+    pages: BoxStream<'static, azure_core::Result<FeedPage<T>>>,
```
heaths, line 117: Nit: FWIW, might be easier to reason about and maintain if you just use `Pin<Box<T>>` in both places (or is this not `Pin<T>`? I forget). If anything, the `wasm32-unknown-unknown` equivalent is `LocalBoxStream` if you wanted to "abbreviate" both.

Pilchie (approved):

sdk/cosmos/azure_data_cosmos/src/feed.rs
```diff
@@ -107,17 +107,173 @@ impl<T: DeserializeOwned> FeedPage<T> {
...
+    #[pin]
+    #[cfg(target_arch = "wasm32")]
```
FabianMeiswinkel, line 119: Why do we even need the wasm32 stuff? Would we ever test/support the RUST SDK in wasm32? I thought we were in agreement that Web-Assemblies are not really a target audience (at least for now)?

FabianMeiswinkel (approved):
LGTM - Thanks

heaths (approved):

sdk/cosmos/azure_data_cosmos/src/query/executor.rs
```diff
@@ -1,246 +1,117 @@
...
-    FeedPage, FeedPager, Query, QueryOptions,
+    conditional_send::ConditionalSend, constants, cosmos_request::CosmosRequest,
+    cosmos_request::CosmosRequestBuilder, operation_context::OperationType,
+    pipeline::CosmosPipeline, resource_context::ResourceLink, FeedPage, Query,
 };
 
-pub struct QueryExecutor<T: DeserializeOwned> {
+/// A query executor that sends queries directly to the gateway endpoint.
+///
+/// This executor does not support cross-partition queries and requires a partition key to be specified.
```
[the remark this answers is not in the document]
analogrelay: Yes, but that is not the long-term plan. We'll be reviewing and addressing cross-partition query again later.

PR #3548, opened 2026-01-15
[Storage] Add default Storage-level logging (exemption list & examples)

vincenttran-msft:
As title states.
Touches Azure Core code to re-export `LoggingOptions` to avoid having to dig into `typespec_client_core` directly.


vincenttran-msft:
Closed in favor of https://github.com/Azure/azure-sdk-for-rust/pull/3575 I think it was the weird Cargo.lock interaction. Not sure how that artifact appeared

PR #3551, opened 2026-01-15
Define our own BlobName for Storage

heaths:


heaths:
There's a bug in the emitter currently that is not using `crate::models::BlobName` with `@alternateType` and instead using our generic `Value`. This has to be fixed before we can merge my Azure/azure-rest-api-specs#39654 PR and regenerate for this one (should be no changes since I manually changed the generated source).

sdk/storage/azure_storage_blob/src/models/mod.rs
```diff
@@ -77,3 +81,317 @@ pub use crate::generated::models::{
...
+
+    #[test]
+    fn as_str_invalid_utf8_ffff() {
+        let blob_name = BlobName {
+            content: Some("%FF%FF".to_string()),
+            encoded: Some(true),
+        };
+        let result = blob_name.as_str();
+        assert!(result.is_err());
+    }
```
vincenttran-msft, line 266: As mentioned offline, these percent-encodings in your test cases `%FF%FF` and `%FF%FE` may be wrong, here is some snippets from Fiddler of the actual requests and the relevant `ListBlobs` excerpt:
>PUT /containerytasgvnx/testablob%EF%BF%BEfffe HTTP/1.1
PUT /containerytasgvnx/testablob%EF%BF%BFffff HTTP/1.1

Image

heaths:
We like #3556 better, though we need some changes in the emitter and maybe even TypeSpec that @jhendrixMSFT (mostly) and I are discussing with the TypeSpec team.

Issue #3522, opened 2026-01-12
Cosmos: Build a Partition Key Range Cache


simorenoh:
Create a Pk Range cache for partition-scoped operations and session token handling in the Rust Cosmos DB SDK

kundadebdatta:
Yes - just re-assigned this item on my name.

PR #3553, opened 2026-01-15
Cosmos: Build `PartitionKeyRangeCache` to Support Caching of PK-Ranges

kundadebdatta:
## Description:
 
This PR is introducing `PartitionKeyRangeCache` and `CollectionRoutingMap` to support the caching of the Partition Key Ranges in the cosmos rust SDK to optimize query routing and reduce metadata requests to Cosmos DB.

### Components Added

| File | Description |
|------|-------------|
| partition_key_range_cache.rs | Main cache for `CollectionRoutingMap` per collection |
| async_cache.rs | Generic async cache with TTL support and conditional refresh |
| collection_routing_map.rs | Stores ordered PK ranges with efficient lookup operations |
| `partition_key_range.rs` | PK range data model |
| `range.rs` | Generic range type with overlap detection |

### Key Features

- **TTL-based caching**: 5-minute expiration for routing maps
- **Incremental updates**: Uses change feed (`A-IM: Incremental Feed`) with ETag continuation
- **Force refresh support**: ETag comparison detects when server data hasn't changed
- **Split range tracking**: Tracks "gone" ranges via parent references for partition splits
- **Efficient lookups**: Binary search for overlapping range queries, HashMap for ID lookups

### Public APIs

```rust
// Find PK ranges overlapping with an effective partition key range
pub async fn resolve_overlapping_ranges(
    &self,
    collection_rid: &str,
    range: Range<String>,
    force_refresh: bool,
) -> Result<Option<Vec<PartitionKeyRange>>, Error>

// Look up a specific PK range by ID
pub async fn resolve_partition_key_range_by_id(
    &self,
    collection_resource_id: &str,
    partition_key_range_id: &str,
    force_refresh: bool,
) -> Option<PartitionKeyRange>
```

### Architecture

```
PartitionKeyRangeCache
    └── AsyncCache<String, CollectionRoutingMap>
            └── CollectionRoutingMap
                    ├── range_by_id: HashMap<String, PartitionKeyRange>
                    ├── ordered_partition_key_ranges: Vec<PartitionKeyRange>
                    └── ordered_ranges: Vec<Range<String>> (for binary search)
```

### Testing
Unit tests cover cache refresh logic, routing map creation, overlapping range queries, and ETag-based refresh decisions.

### Sequence Diagram

Below sequence diagram explains the interaction in more details:

```mermaid
sequenceDiagram
    participant Client as Caller
    participant PKRangeCache as PartitionKeyRangeCache
    participant AsyncCache as AsyncCache<String, CollectionRoutingMap>
    participant Pipeline as GatewayPipeline
    participant GEM as GlobalEndpointManager
    participant CosmosDB as Cosmos DB Service

    %% Resolve Overlapping Ranges Flow
    rect rgb(240, 248, 255)
        Note over Client, CosmosDB: resolve_overlapping_ranges(collection_rid, range, force_refresh)
        Client->>PKRangeCache: resolve_overlapping_ranges()
        PKRangeCache->>PKRangeCache: try_lookup(collection_rid, None)
        PKRangeCache->>AsyncCache: get(key, should_refresh, compute_fn)
        
        alt Cache Hit (valid entry exists)
            AsyncCache-->>PKRangeCache: Ok(CollectionRoutingMap)
        else Cache Miss or Expired
            AsyncCache->>PKRangeCache: invoke compute_fn
            PKRangeCache->>PKRangeCache: get_routing_map_for_collection()
            loop Fetch all PK Ranges (Change Feed)
                PKRangeCache->>PKRangeCache: execute_partition_key_range_read_change_feed()
                PKRangeCache->>GEM: resolve_service_endpoint()
                GEM-->>PKRangeCache: endpoint URL
                PKRangeCache->>Pipeline: send(cosmos_request, context)
                Pipeline->>CosmosDB: GET /pkranges (A-IM: Incremental Feed)
                CosmosDB-->>Pipeline: PkRangesResponse + ETag
                Pipeline-->>PKRangeCache: Response<PkRangesResponse>
                Note right of PKRangeCache: Extract ETag for continuation
            end
            PKRangeCache->>PKRangeCache: Create CollectionRoutingMap
            PKRangeCache-->>AsyncCache: Ok(CollectionRoutingMap)
            AsyncCache->>AsyncCache: Store with TTL (5 min)
            AsyncCache-->>PKRangeCache: Ok(CollectionRoutingMap)
        end
        
        alt force_refresh = true AND previous exists
            PKRangeCache->>PKRangeCache: try_lookup(collection_rid, Some(previous))
            PKRangeCache->>AsyncCache: get(key, should_force_refresh, compute_fn)
            Note right of AsyncCache: should_force_refresh checks<br/>if ETags match (no server changes)
            alt ETags Match (force refresh needed)
                AsyncCache->>AsyncCache: Remove cached entry
                AsyncCache->>PKRangeCache: invoke compute_fn
                Note right of PKRangeCache: Fetch fresh data from service
            else ETags Differ (cache is current)
                AsyncCache-->>PKRangeCache: Return cached value
            end
        end
        
        PKRangeCache->>PKRangeCache: routing_map.get_overlapping_ranges(range)
        PKRangeCache-->>Client: Ok(Vec<PartitionKeyRange>)
    end

    %% Resolve by Partition Key Range ID Flow
    rect rgb(255, 248, 240)
        Note over Client, CosmosDB: resolve_partition_key_range_by_id(collection_rid, pk_range_id, force_refresh)
        Client->>PKRangeCache: resolve_partition_key_range_by_id()
        PKRangeCache->>PKRangeCache: try_lookup(collection_rid, None)
        PKRangeCache->>AsyncCache: get(key, should_refresh, compute_fn)
        AsyncCache-->>PKRangeCache: Ok(CollectionRoutingMap) or compute
        
        alt force_refresh = true
            PKRangeCache->>PKRangeCache: try_lookup with previous_value
        end
        
        PKRangeCache->>PKRangeCache: routing_map.try_get_range_by_partition_key_range_id()
        PKRangeCache-->>Client: Option<PartitionKeyRange>
    end

    %% CollectionRoutingMap Creation/Combination
    rect rgb(240, 255, 240)
        Note over PKRangeCache, CosmosDB: Routing Map Construction
        
        alt No Previous Routing Map
            PKRangeCache->>PKRangeCache: Filter out gone ranges (from parents)
            PKRangeCache->>PKRangeCache: CollectionRoutingMap::try_create_complete_routing_map()
        else Has Previous Routing Map (Incremental Update)
            PKRangeCache->>PKRangeCache: previous_map.try_combine(new_ranges, etag)
            Note right of PKRangeCache: Merges existing + new ranges<br/>Tracks gone/split ranges
        end
    end
```

## Closing issues

Fixes Azure/azure-sdk-for-rust#3522

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
@@ -12,7 +12,7 @@ use crate::{
 use std::sync::Arc;
 
 use crate::cosmos_request::CosmosRequest;
-use crate::handler::cosmos_connection::CosmosConnection;
+use crate::handler::container_connection::ContainerConnection;
 use crate::operation_context::OperationType;
 use crate::routing::container_cache::ContainerCache;
 use crate::routing::global_endpoint_manager::GlobalEndpointManager;
@@ -28,7 +28,7 @@ pub struct ContainerClient {
     link: ResourceLink,
     items_link: ResourceLink,
     pipeline: Arc<GatewayPipeline>,
-    cosmos_connection: Arc<CosmosConnection>,
+    container_connection: Arc<ContainerConnection>,
     container_id: String,
 }
 
@@ -44,7 +44,7 @@ impl ContainerClient {
             .item(container_id);
         let items_link = link.feed(ResourceType::Documents);
 
-        let collection_cache = ContainerCache::new(
+        let container_cache = ContainerCache::new(
             pipeline.clone(),
             link.clone(),
             global_endpoint_manager.clone(),
@@ -52,12 +52,12 @@ impl ContainerClient {
         let partition_key_range_cache = PartitionKeyRangeCache::new(
             pipeline.clone(),
             database_link.clone(),
-            Arc::from(collection_cache.clone()),
+            Arc::from(container_cache.clone()),
             Arc::from(global_endpoint_manager.clone()),
         );
-        let cosmos_connection = Arc::from(CosmosConnection::new(
+        let container_connection = Arc::from(ContainerConnection::new(
             pipeline.clone(),
-            Arc::from(collection_cache),
+            Arc::from(container_cache),
             Arc::from(partition_key_range_cache),
         ));
 
@@ -65,7 +65,7 @@ impl ContainerClient {
             link,
             items_link,
             pipeline,
-            cosmos_connection,
+            container_connection,
             container_id: container_id.to_string(),
         }
     }
@@ -94,7 +94,7 @@ impl ContainerClient {
         let options = options.unwrap_or_default();
         let cosmos_request =
             CosmosRequest::builder(OperationType::Read, self.link.clone()).build()?;
-        self.cosmos_connection
+        self.container_connection
             .send(cosmos_request, options.method_options.context)
             .await
     }
@@ -139,7 +139,7 @@ impl ContainerClient {
         let cosmos_request = CosmosRequest::builder(OperationType::Replace, self.link.clone())
             .json(&properties)
             .build()?;
-        self.cosmos_connection
+        self.container_connection
             .send(cosmos_request, options.method_options.context)
             .await
     }
@@ -208,7 +208,7 @@ impl ContainerClient {
         let options = options.unwrap_or_default();
         let cosmos_request =
             CosmosRequest::builder(OperationType::Delete, self.link.clone()).build()?;
-        self.cosmos_connection
+        self.container_connection
             .send(cosmos_request, options.method_options.context)
             .await
     }
@@ -292,7 +292,7 @@ impl ContainerClient {
             .partition_key(partition_key.into())
             .build()?;
 
-        self.cosmos_connection
+        self.container_connection
             .send(cosmos_request, options.method_options.context)
             .await
     }
@@ -378,7 +378,7 @@ impl ContainerClient {
             .partition_key(partition_key.into())
             .build()?;
 
-        self.cosmos_connection
+        self.container_connection
             .send(cosmos_request, options.method_options.context)
             .await
     }
@@ -465,7 +465,7 @@ impl ContainerClient {
             .partition_key(partition_key.into())
             .build()?;
 
-        self.cosmos_connection
+        self.container_connection
             .send(cosmos_request, options.method_options.context)
             .await
     }
@@ -518,7 +518,7 @@ impl ContainerClient {
             .request_headers(&options)
             .build()?;
 
-        self.cosmos_connection
+        self.container_connection
             .send(cosmos_request, options.method_options.context)
             .await
     }
@@ -557,7 +557,7 @@ impl ContainerClient {
             .request_headers(&options)
             .build()?;
 
-        self.cosmos_connection
+        self.container_connection
             .send(cosmos_request, options.method_options.context)
             .await
     }
@@ -633,7 +633,7 @@ impl ContainerClient {
             .json(&patch)
             .build()?;
 
-        self.cosmos_connection
+        self.container_connection
             .send(cosmos_request, options.method_options.context)
             .await
     }
--- a/sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
@@ -11,14 +11,14 @@ use std::sync::Arc;
 
 /// Handler for managing transport-level operations with Cosmos DB.
 #[derive(Debug, Clone)]
-pub struct CosmosConnection {
+pub struct ContainerConnection {
     pipeline: Arc<GatewayPipeline>,
     container_cache: Arc<ContainerCache>,
     pk_range_cache: Arc<PartitionKeyRangeCache>,
 }
 
-impl CosmosConnection {
-    /// Creates a new [`CosmosConnection`] with the specified dependencies.
+impl ContainerConnection {
+    /// Creates a new [`ContainerConnection`] with the specified dependencies.
     ///
     /// # Arguments
     ///
@@ -136,64 +136,6 @@ mod tests {
     }
 
     #[test]
-    fn new_cosmos_connection() {
-        let endpoint_manager = create_endpoint_manager();
-        let pipeline = create_gateway_pipeline(&endpoint_manager);
-        let container_cache = create_container_cache(pipeline.clone(), endpoint_manager.clone());
-        let pk_range_cache = create_pk_range_cache(
-            pipeline.clone(),
-            container_cache.clone(),
-            Arc::new(endpoint_manager),
-        );
-
-        let connection = CosmosConnection::new(pipeline, container_cache, pk_range_cache);
-
-        // Verify the connection was created successfully
-        assert!(std::mem::size_of_val(&connection) > 0);
-    }
-
-    #[test]
-    fn cosmos_connection_clone() {
-        let endpoint_manager = create_endpoint_manager();
-        let pipeline = create_gateway_pipeline(&endpoint_manager);
-        let container_cache = create_container_cache(pipeline.clone(), endpoint_manager.clone());
-        let pk_range_cache = create_pk_range_cache(
-            pipeline.clone(),
-            container_cache.clone(),
-            Arc::new(endpoint_manager),
-        );
-
-        let connection = CosmosConnection::new(pipeline, container_cache, pk_range_cache);
-        let cloned_connection = connection.clone();
-
-        // Both should be valid instances
-        assert!(std::mem::size_of_val(&connection) > 0);
-        assert!(std::mem::size_of_val(&cloned_connection) > 0);
-    }
-
-    #[test]
-    fn cosmos_connection_has_required_fields() {
-        let endpoint_manager = create_endpoint_manager();
-        let pipeline = create_gateway_pipeline(&endpoint_manager);
-        let container_cache = create_container_cache(pipeline.clone(), endpoint_manager.clone());
-        let pk_range_cache = create_pk_range_cache(
-            pipeline.clone(),
-            container_cache.clone(),
-            Arc::new(endpoint_manager),
-        );
-
-        let connection = CosmosConnection::new(pipeline, container_cache, pk_range_cache);
-
-        // Test that connection has pipeline, container_cache, and pk_range_cache
-        // by checking that it implements Debug (which requires all fields to implement Debug)
-        let debug_output = format!("{:?}", connection);
-        assert!(debug_output.contains("CosmosConnection"));
-        assert!(debug_output.contains("pipeline"));
-        assert!(debug_output.contains("container_cache"));
-        assert!(debug_output.contains("pk_range_cache"));
-    }
-
-    #[test]
     fn cosmos_request_builder_creates_valid_request() {
         let request = create_cosmos_request();
         assert_eq!(request.operation_type, OperationType::Read);
@@ -201,7 +143,7 @@ mod tests {
     }
 
     #[test]
-    fn cosmos_connection_with_preferred_locations() {
+    fn container_connection_with_preferred_locations() {
         let pipeline = azure_core::http::Pipeline::new(
             option_env!("CARGO_PKG_NAME"),
             option_env!("CARGO_PKG_VERSION"),
@@ -232,14 +174,15 @@ mod tests {
             Arc::new(endpoint_manager),
         );
 
-        let connection = CosmosConnection::new(gateway_pipeline, container_cache, pk_range_cache);
+        let connection =
+            ContainerConnection::new(gateway_pipeline, container_cache, pk_range_cache);
 
         // Verify the connection was created successfully with preferred locations
         assert!(std::mem::size_of_val(&connection) > 0);
     }
 
     #[test]
-    fn multiple_cosmos_connections_share_caches() {
+    fn multiple_container_connections_share_caches() {
         let endpoint_manager = create_endpoint_manager();
         let pipeline = create_gateway_pipeline(&endpoint_manager);
         let container_cache = create_container_cache(pipeline.clone(), endpoint_manager.clone());
@@ -250,17 +193,17 @@ mod tests {
         );
 
         // Create multiple connections sharing the same caches
-        let connection1 = CosmosConnection::new(
+        let connection1 = ContainerConnection::new(
             pipeline.clone(),
             container_cache.clone(),
             pk_range_cache.clone(),
         );
-        let connection2 = CosmosConnection::new(
+        let connection2 = ContainerConnection::new(
             pipeline.clone(),
             container_cache.clone(),
             pk_range_cache.clone(),
         );
-        let connection3 = CosmosConnection::new(pipeline, container_cache, pk_range_cache);
+        let connection3 = ContainerConnection::new(pipeline, container_cache, pk_range_cache);
 
         // All connections should be valid
         assert!(std::mem::size_of_val(&connection1) > 0);
@@ -322,7 +265,7 @@ mod tests {
     }
 
     #[test]
-    fn cosmos_connection_debug_implementation() {
+    fn container_connection_debug_implementation() {
         let endpoint_manager = create_endpoint_manager();
         let pipeline = create_gateway_pipeline(&endpoint_manager);
         let container_cache = create_container_cache(pipeline.clone(), endpoint_manager.clone());
@@ -332,7 +275,7 @@ mod tests {
             Arc::new(endpoint_manager),
         );
 
-        let connection = CosmosConnection::new(pipeline, container_cache, pk_range_cache);
+        let connection = ContainerConnection::new(pipeline, container_cache, pk_range_cache);
 
         // Verify Debug trait is properly implemented
         let debug_str = format!("{:?}", connection);
--- a/sdk/cosmos/azure_data_cosmos/src/routing/partition_key_range.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/routing/partition_key_range.rs
@@ -144,7 +144,8 @@ impl PartialEq for PartitionKeyRange {
 
 impl Eq for PartitionKeyRange {}
 
-// Implement Hash for PartitionKeyRange
+// Implement a Manual Hash for PartitionKeyRange, because only the ID, RID,
+// and min/max should be considered for equality/hashing.
 impl Hash for PartitionKeyRange {
     fn hash<H: Hasher>(&self, state: &mut H) {
         self.id.hash(state);
--- a/sdk/cosmos/azure_data_cosmos/src/handler/mod.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/handler/mod.rs
@@ -2,5 +2,5 @@
 // Licensed under the MIT License.
 
 //! Handler types for request processing and retry logic.
-pub(crate) mod cosmos_connection;
+pub(crate) mod container_connection;
 pub(crate) mod retry_handler;


sdk/cosmos/azure_data_cosmos/src/routing/partition_key_range_cache.rs
```diff
@@ -0,0 +1,690 @@
...
+                .await?;
+
+            last_status_code = response.status();
+            change_feed_next_if_none_match = response
+                .headers()
+                .get_optional_string(&HeaderName::from_static("etag"));
+
+            // If status is 304 (NotModified), the body is empty, so skip parsing
+            if last_status_code == StatusCode::NotModified {
+                break;
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
```diff
@@ -0,0 +1,53 @@
...
+        let container_prop = self
+            .container_cache
+            .resolve_by_id("sdk_rust_container".parse()?, None, false)
+            .await?;
+        let _pk_range = self
+            .pk_range_cache
+            .resolve_partition_key_range_by_id(&container_prop.id, "0".as_ref(), false)
+            .await;
+        self.pipeline.send(cosmos_request, context).await
+    }
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed. Added unit tests to close the test gaps.

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
```diff
@@ -0,0 +1,53 @@
...
+    container_cache: Arc<ContainerCache>,
+    pk_range_cache: Arc<PartitionKeyRangeCache>,
+}
+
+impl CosmosConnection {
+    /// Creates a new `TransportHandler` with the specified pipeline.
+    ///
+    /// # Arguments
+    ///
+    /// * `pipeline` - The Cosmos pipeline to use for sending requests.
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/routing/partition_key_range_cache.rs
```diff
@@ -0,0 +1,690 @@
...
+                .collect();
+
+            let filtered_tuples: Vec<_> = tuples
+                .into_iter()
+                .filter(|(range, _)| !gone_ranges.contains(&range.id))
+                .collect();
+
+            CollectionRoutingMap::try_create_complete_routing_map(
+                filtered_tuples,
+                String::new(),
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/routing/collection_routing_map.rs
```diff
@@ -0,0 +1,487 @@
...
+        &self,
+        effective_partition_key_value: &str,
+    ) -> Result<&PartitionKeyRange, Error> {
+        if effective_partition_key_value >= MAXIMUM_EXCLUSIVE_EFFECTIVE_PARTITION_KEY {
+            return Err(Error::with_message(
+                azure_core::error::ErrorKind::Other,
+                "effectivePartitionKeyValue out of range",
+            ));
+        }
+
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/routing/collection_routing_map.rs
```diff
@@ -0,0 +1,487 @@
...
+                        Some(INVALID_PK_RANGE_ID)
+                    }
+                }
+                Err(_) => {
+                    tracing::error!(
+                        "Could not parse partition key range Id as int {} for collectionRid {}",
+                        range.id,
+                        collection_unique_id
+                    );
+                    None
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/routing/service_identity.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    /// Gets the federation ID
+    pub fn federation_id(&self) -> &str {
+        &self.federation_id
+    }
+
+    /// Gets the service name (URI)
+    pub fn service_name(&self) -> &String {
+        &self.service_name
+    }
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/routing/collection_routing_map.rs
```diff
@@ -0,0 +1,487 @@
...
+                .binary_search_by(|probe| Self::compare_range_min(probe, provided_range))
+                .unwrap_or_else(|idx| if idx > 0 { idx - 1 } else { 0 });
+
+            let max_index = match self
+                .ordered_ranges
+                .binary_search_by(|probe| Self::compare_range_max(probe, provided_range))
+            {
+                Ok(idx) => idx,
+                Err(idx) => std::cmp::min(self.ordered_partition_key_ranges.len() - 1, idx),
+            };
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
```diff
@@ -0,0 +1,53 @@
...
+        context: Context<'_>,
+    ) -> azure_core::Result<Response<T>> {
+        let container_prop = self
+            .container_cache
+            .resolve_by_id("sdk_rust_container".parse()?, None, false)
+            .await?;
+        let _pk_range = self
+            .pk_range_cache
+            .resolve_partition_key_range_by_id(&container_prop.id, "0".as_ref(), false)
+            .await;
```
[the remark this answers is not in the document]
kundadebdatta: Removed the unnecessary cache warm ups.

sdk/cosmos/azure_data_cosmos/src/routing/partition_key_range_cache.rs
```diff
@@ -0,0 +1,690 @@
...
+            }
+
+            let body_string = response.into_body().into_string()?;
+
+            // Deserialize the response body to extract Vec<PartitionKeyRange>
+            let pk_ranges_response: PkRangesResponse = serde_json::from_str(&body_string)
+                .map_err(|e| Error::new(azure_core::error::ErrorKind::DataConversion, e))?;
+
+            ranges.extend(pk_ranges_response.partition_key_ranges);
+        }
```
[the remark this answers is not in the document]
kundadebdatta: Thanks! This is now addressed.

sdk/cosmos/.dict.txt
```diff
@@ -29,4 +29,7 @@ msvc
 winget
 makefiles
 ctest
-cmake
\ No newline at end of file
+cmake
+
+# Common rust syntaxes
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/routing/partition_key_range_cache.rs
```diff
@@ -0,0 +1,690 @@
...
+#[async_trait]
+pub trait CollectionRoutingMapCache: Send + Sync {
+    async fn try_lookup(
+        &self,
+        collection_rid: &str,
+        previous_value: Option<Arc<CollectionRoutingMap>>,
+        request: Option<Arc<CosmosRequest>>,
+    ) -> Response<Option<Arc<CollectionRoutingMap>>>;
+}
+
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/routing/partition_key_range.rs
```diff
@@ -0,0 +1,225 @@
...
+            resource_id: None,
+            self_link: None,
+            etag: None,
+            timestamp: None,
+            min_inclusive,
+            max_exclusive,
+            rid_prefix: None,
+            throughput_fraction: 0.0,
+            target_throughput: None,
+            status: PartitionKeyRangeStatus::default(),
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/routing/partition_key_range_cache.rs
```diff
@@ -0,0 +1,690 @@
...
+        current_value: Option<CollectionRoutingMap>,
+    ) -> bool {
+        match (previous_value, current_value) {
+            (Some(prev), Some(curr)) => {
+                prev.change_feed_next_if_none_match == curr.change_feed_next_if_none_match
+            }
+            _ => false,
+        }
+    }
+
```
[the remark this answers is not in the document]
kundadebdatta: Yes - this was a pending refactor. Needed to refactor the async_cache implementation to fit in the force refresh computation, using a callback.

sdk/cosmos/azure_data_cosmos/src/routing/partition_key_range.rs
```diff
@@ -0,0 +1,225 @@
...
+impl Eq for PartitionKeyRange {}
+
+// Implement Hash for PartitionKeyRange
+impl Hash for PartitionKeyRange {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.id.hash(state);
+        self.resource_id.hash(state);
+        self.min_inclusive.hash(state);
+        self.max_exclusive.hash(state);
+    }
```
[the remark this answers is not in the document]
kundadebdatta: This is now taken care. Updated the `PartialEq` implementation.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -23,25 +27,45 @@ use serde::{de::DeserializeOwned, Serialize};
...
+            pipeline.clone(),
+            database_link.clone(),
+            Arc::from(collection_cache.clone()),
+            Arc::from(global_endpoint_manager.clone()),
+        );
+        let cosmos_connection = Arc::from(CosmosConnection::new(
+            pipeline.clone(),
+            Arc::from(collection_cache),
+            Arc::from(partition_key_range_cache),
+        ));
```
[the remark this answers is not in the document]
analogrelay: Copilot is kinda right, it would be good to share this cache across other containers, but that's not the primary purpose. I'd say this is OK for now. I do think there's some naming confusion with `CosmosConnection` though, but I'll cover that below.

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
```diff
@@ -23,11 +24,16 @@ pub struct DatabaseClient {
...
+    pipeline: Arc<GatewayPipeline>,
+    global_endpoint_manager: GlobalEndpointManager,
 }
 
 impl DatabaseClient {
-    pub(crate) fn new(pipeline: Arc<CosmosPipeline>, database_id: &str) -> Self {
+    pub(crate) fn new(
+        pipeline: Arc<GatewayPipeline>,
+        database_id: &str,
+        global_endpoint_manager: GlobalEndpointManager,
```
analogrelay, line 35: Perhaps we should be storing the GEM in an Arc 🤔 . It depends on how heavy it is, and if it stores any shared state that isn't already in Arcs.

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
```diff
@@ -0,0 +1,341 @@
...
+use crate::cosmos_request::CosmosRequest;
+use crate::pipeline::GatewayPipeline;
+use crate::routing::container_cache::ContainerCache;
+use crate::routing::partition_key_range_cache::PartitionKeyRangeCache;
+use azure_core::http::{Context, Response};
+use std::sync::Arc;
+
+/// Handler for managing transport-level operations with Cosmos DB.
+#[derive(Debug, Clone)]
+pub struct CosmosConnection {
```
analogrelay: Since this is a per-container type, perhaps `ContainerConnection` to be clearer?

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
```diff
@@ -0,0 +1,341 @@
...
+            pipeline.clone(),
+            container_cache.clone(),
+            Arc::new(endpoint_manager),
+        );
+
+        let connection = CosmosConnection::new(pipeline, container_cache, pk_range_cache);
+
+        // Verify the connection was created successfully
+        assert!(std::mem::size_of_val(&connection) > 0);
+    }
```
analogrelay: Feels like an unnecessary test to me

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
```diff
@@ -0,0 +1,341 @@
...
+            Arc::new(endpoint_manager),
+        );
+
+        let connection = CosmosConnection::new(pipeline, container_cache, pk_range_cache);
+        let cloned_connection = connection.clone();
+
+        // Both should be valid instances
+        assert!(std::mem::size_of_val(&connection) > 0);
+        assert!(std::mem::size_of_val(&cloned_connection) > 0);
+    }
```
analogrelay: Another unnecessary test

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
```diff
@@ -0,0 +1,341 @@
...
+        let connection = CosmosConnection::new(pipeline, container_cache, pk_range_cache);
+
+        // Test that connection has pipeline, container_cache, and pk_range_cache
+        // by checking that it implements Debug (which requires all fields to implement Debug)
+        let debug_output = format!("{:?}", connection);
+        assert!(debug_output.contains("CosmosConnection"));
+        assert!(debug_output.contains("pipeline"));
+        assert!(debug_output.contains("container_cache"));
+        assert!(debug_output.contains("pk_range_cache"));
+    }
```
analogrelay: And another.

Did you use Copilot to generate these tests? That's a great idea, but I do find it needs some guidance to avoid these simple "test spam" kind of tests. We can look at getting some instructions into an AGENTS.md for that so we don't have to remember it for every prompt.

sdk/cosmos/azure_data_cosmos/src/routing/container_cache.rs
```diff
@@ -40,17 +41,20 @@ impl ContainerCache {
...
     /// # Returns
     /// A new `ContainerCache` instance ready for caching container metadata
-    pub(crate) fn new(pipeline: Pipeline, global_endpoint_manager: GlobalEndpointManager) -> Self {
+    pub(crate) fn new(
+        pipeline: Arc<GatewayPipeline>,
+        container_link: ResourceLink,
+        global_endpoint_manager: GlobalEndpointManager,
+    ) -> Self {
         let container_properties_cache = AsyncCache::new(
             Duration::from_secs(300), // Default 5 minutes TTL
```
FabianMeiswinkel: why? We do not use TTL based caches for a reason - it would not preserve you from having to react to error conditions that could be mitigated by cache refresh and just creates noise if there are no changes?

kundadebdatta: Good point Fabian. The `GlobalEndpointManager` and `LocationCache` also uses the `async_cache` and Initially we didn't wanted to have a background task to refresh the cache in every 10 mins (like we do for Java and .NET SDKs) for better rust performance, and therefore, we decided to use a TTL based model to force expire the entry so that the cache would end up refreshing itself. But this was specifically done for the GEM and Location cache.
 
I agree that the `container_cahce` and `pk_range_cache` does not require to set the TTL and solely just depend on the `force_refresh` model. Hence, I refactored the PR to support an optional ttl based initialization of the async_cache.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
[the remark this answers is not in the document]
kundadebdatta: The data plane operations are done on a specific container and hence, the cache should be tied to that particular container itself. Sharing of the container_cache is not needed for the imidiate term.

Resolving this comment.

sdk/cosmos/azure_data_cosmos/src/routing/container_cache.rs
FabianMeiswinkel: NIT - please also update the PR description

sdk/cosmos/azure_data_cosmos/src/routing/range.rs
```diff
@@ -0,0 +1,350 @@
...
+#![allow(dead_code)]
+
+use serde::{Deserialize, Serialize};
+use std::cmp::Ordering;
+use std::fmt;
+use std::hash::{Hash, Hasher};
+
+/// Represents a range with generic bounds that implement comparison
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Range<T>
```
FabianMeiswinkel, line 12: Should not be public

sdk/cosmos/azure_data_cosmos/src/routing/partition_key_range.rs
```diff
@@ -0,0 +1,226 @@
...
+// Licensed under the MIT License.
+#![allow(dead_code)]
+
+use crate::routing::range::Range;
+use serde::{Deserialize, Serialize};
+use std::collections::HashSet;
+use std::hash::{Hash, Hasher};
+/// Represents a partition key range in the Azure Cosmos DB service.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct PartitionKeyRange {
```
FabianMeiswinkel, line 11: Should not be public

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
FabianMeiswinkel, line 27: Should not be public

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -23,25 +27,45 @@ use serde::{de::DeserializeOwned, Serialize};
...
 
+        let collection_cache = ContainerCache::new(
+            pipeline.clone(),
+            link.clone(),
+            global_endpoint_manager.clone(),
+        );
+        let partition_key_range_cache = PartitionKeyRangeCache::new(
+            pipeline.clone(),
+            database_link.clone(),
+            Arc::from(collection_cache.clone()),
```
FabianMeiswinkel: Should not clone? The whole purpose of these caches is to not have multiple instances (which would result in unnecessary hcache misses (on TTL etc) on each instance) - same for global endpoint manager

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -72,16 +73,17 @@ impl CosmosClient {
         let global_endpoint_manager =
             GlobalEndpointManager::new(endpoint.clone(), preferred_regions, pipeline_core.clone());
 
-        let pipeline = Arc::new(CosmosPipeline::new(
+        let pipeline = Arc::new(GatewayPipeline::new(
             endpoint,
             pipeline_core,
-            global_endpoint_manager,
+            global_endpoint_manager.clone(),
```
FabianMeiswinkel, line 79: no clone (see above)

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -126,16 +128,17 @@ impl CosmosClient {
...
```
FabianMeiswinkel, line 134: no clone (see above)

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -174,7 +177,11 @@ impl CosmosClient {
     /// # Arguments
     /// * `id` - The ID of the database.
     pub fn database_client(&self, id: &str) -> DatabaseClient {
-        DatabaseClient::new(self.pipeline.clone(), id)
+        DatabaseClient::new(
+            self.pipeline.clone(),
+            id,
+            self.global_endpoint_manager.clone(),
```
FabianMeiswinkel, line 183: no clone (see above)

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
```diff
@@ -45,7 +52,12 @@ impl DatabaseClient {
     /// # Arguments
     /// * `name` - The name of the container.
     pub fn container_client(&self, name: &str) -> ContainerClient {
-        ContainerClient::new(self.pipeline.clone(), &self.link, name)
+        ContainerClient::new(
+            self.pipeline.clone(),
+            &self.link,
+            name,
+            self.global_endpoint_manager.clone(),
```
FabianMeiswinkel, line 59: no clone (see above)

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -0,0 +1,341 @@
...
+            ClientOptions::default(),
+            Vec::new(),
+            Vec::new(),
+            None,
+        );
+        let endpoint = Url::parse("https://test.documents.azure.com").unwrap();
+        Arc::new(GatewayPipeline::new(
+            endpoint,
+            pipeline_core,
+            endpoint_manager.clone(),
```
FabianMeiswinkel, line 89: no clone (see above)

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    #[test]
+    fn new_cosmos_connection() {
+        let endpoint_manager = create_endpoint_manager();
+        let pipeline = create_gateway_pipeline(&endpoint_manager);
+        let container_cache = create_container_cache(pipeline.clone(), endpoint_manager.clone());
+        let pk_range_cache = create_pk_range_cache(
+            pipeline.clone(),
+            container_cache.clone(),
```
FabianMeiswinkel: no clone (see above)

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    #[test]
+    fn cosmos_connection_clone() {
+        let endpoint_manager = create_endpoint_manager();
+        let pipeline = create_gateway_pipeline(&endpoint_manager);
+        let container_cache = create_container_cache(pipeline.clone(), endpoint_manager.clone());
+        let pk_range_cache = create_pk_range_cache(
+            pipeline.clone(),
+            container_cache.clone(),
```
FabianMeiswinkel: no clone (see above)

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    #[test]
+    fn cosmos_connection_has_required_fields() {
+        let endpoint_manager = create_endpoint_manager();
+        let pipeline = create_gateway_pipeline(&endpoint_manager);
+        let container_cache = create_container_cache(pipeline.clone(), endpoint_manager.clone());
+        let pk_range_cache = create_pk_range_cache(
+            pipeline.clone(),
+            container_cache.clone(),
```
FabianMeiswinkel: no clone (see above)

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -0,0 +1,341 @@
...
+        let endpoint_manager = GlobalEndpointManager::new(
+            endpoint.clone(),
+            vec![Cow::Borrowed("East US"), Cow::Borrowed("West US")],
+            pipeline.clone(),
+        );
+
+        let gateway_pipeline = Arc::new(GatewayPipeline::new(
+            endpoint,
+            pipeline,
+            endpoint_manager.clone(),
```
FabianMeiswinkel, line 165: no clone (see above)

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -0,0 +1,341 @@
...
+            CosmosClientOptions::default(),
+        ));
+
+        let container_cache =
+            create_container_cache(gateway_pipeline.clone(), endpoint_manager.clone());
+        let pk_range_cache = create_pk_range_cache(
+            gateway_pipeline.clone(),
+            container_cache.clone(),
```
FabianMeiswinkel, line 173: no clone (see above)

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    #[test]
+    fn multiple_cosmos_connections_share_caches() {
+        let endpoint_manager = create_endpoint_manager();
+        let pipeline = create_gateway_pipeline(&endpoint_manager);
+        let container_cache = create_container_cache(pipeline.clone(), endpoint_manager.clone());
+        let pk_range_cache = create_pk_range_cache(
+            pipeline.clone(),
+            container_cache.clone(),
```
FabianMeiswinkel, line 191: no clone (see above)

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -0,0 +1,341 @@
...
+            Arc::new(endpoint_manager),
+        );
+
+        // Create multiple connections sharing the same caches
+        let connection1 = CosmosConnection::new(
+            pipeline.clone(),
+            container_cache.clone(),
```
FabianMeiswinkel, line 198: no clone (see above)

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -0,0 +1,341 @@
...
+            pk_range_cache.clone(),
```
FabianMeiswinkel, line 199: no clone (see above)

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -0,0 +1,341 @@
...
+        );
+        let connection2 = CosmosConnection::new(
+            pipeline.clone(),
+            container_cache.clone(),
```
FabianMeiswinkel, line 203: no clone (see above)

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -0,0 +1,341 @@
...
+            pk_range_cache.clone(),
+        );
+        let connection3 = CosmosConnection::new(pipeline, container_cache, pk_range_cache);
+
+        // All connections should be valid
```
FabianMeiswinkel, line 208: If I don't have a misunderstanding the assertions do not reflect the goal - // Create multiple connections sharing the same caches - because some connections clone caches - so, end-up using different cache instances?

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
kundadebdatta: When you look at the `AsyncCache` structure:

```rust
pub struct AsyncCache<K, V> {
    store: Arc<RwLock<HashMap<K, CacheEntry<V>>>>,  // <-- Arc-wrapped!
    ttl: Option<Duration>,
}
```

And `ContainerCache`:

```rust
pub struct ContainerCache {
    pipeline: Arc<GatewayPipeline>,
    container_link: ResourceLink,
    global_endpoint_manager: GlobalEndpointManager,
    container_properties_cache: AsyncCache<String, ContainerProperties>,  // <-- Contains Arc internally
}
```
It's safe to invoke `clone()` on the caches.

**Why It's Safe**

When we call `collection_cache.clone()`:

- `AsyncCache` uses `Arc<RwLock<HashMap<...>>>` internally for its store field.
- Cloning an `Arc` only increments the reference count - it does NOT create a new `HashMap`
Both and `PartitionKeyRangeCache` will share the same underlying cache data.

collection_cache.clone()
       │
       ▼
┌──────────────────────┐     ┌──────────────────────┐
│ ContainerCache #1    │     │ ContainerCache #2    │
│ (in CosmosConnection)│     │ (in PKRangeCache)    │
├──────────────────────┤     ├──────────────────────┤
│ store: Arc ──────────┼─────┼──► Arc ─────────────┐│
└──────────────────────┘     └─────────────────────┼┘
                                                   │
                             ┌─────────────────────▼─┐
                             │  RwLock<HashMap<...>> │
                             │  (SHARED - single     │
                             │   instance)           │
                             └───────────────────────┘

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
kundadebdatta: It should be okay to use `clone()`, explained [in this comment.](https://github.com/Azure/azure-sdk-for-rust/pull/3553#discussion_r2713589401)

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
kundadebdatta: It should be okay to use `clone()`, explained [in this comment.](https://github.com/Azure/azure-sdk-for-rust/pull/3553#discussion_r2713589401)

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
kundadebdatta: Discussed offline. It should be okay to use `clone()`, explained [in this comment.](https://github.com/Azure/azure-sdk-for-rust/pull/3553#discussion_r2713589401)

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
kundadebdatta: Discussed offline. It should be okay to use `clone()`, explained [in this comment.](https://github.com/Azure/azure-sdk-for-rust/pull/3553#discussion_r2713589401)

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
kundadebdatta: Discussed offline. It should be okay to use `clone()`, explained [in this comment.](https://github.com/Azure/azure-sdk-for-rust/pull/3553#discussion_r2713589401)

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
kundadebdatta: Discussed offline. It should be okay to use `clone()`, explained [in this comment.](https://github.com/Azure/azure-sdk-for-rust/pull/3553#discussion_r2713589401)

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
kundadebdatta: Discussed offline. It should be okay to use `clone()`, explained [in this comment.](https://github.com/Azure/azure-sdk-for-rust/pull/3553#discussion_r2713589401)

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
kundadebdatta: Discussed offline. It should be okay to use `clone()`, explained [in this comment.](https://github.com/Azure/azure-sdk-for-rust/pull/3553#discussion_r2713589401)

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
kundadebdatta: Discussed offline. It should be okay to use `clone()`, explained [in this comment.](https://github.com/Azure/azure-sdk-for-rust/pull/3553#discussion_r2713589401)

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
kundadebdatta: Discussed offline. It should be okay to use `clone()`, explained [in this comment.](https://github.com/Azure/azure-sdk-for-rust/pull/3553#discussion_r2713589401)

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
kundadebdatta: Discussed offline. It should be okay to use `clone()`, explained [in this comment.](https://github.com/Azure/azure-sdk-for-rust/pull/3553#discussion_r2713589401)

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
kundadebdatta: Discussed offline. It should be okay to use `clone()`, explained [in this comment.](https://github.com/Azure/azure-sdk-for-rust/pull/3553#discussion_r2713589401)

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
kundadebdatta: Discussed offline. It should be okay to use `clone()`, explained [in this comment.](https://github.com/Azure/azure-sdk-for-rust/pull/3553#discussion_r2713589401)

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
analogrelay: Sometimes I wonder if there should be a trait that's basically the same as clone but called something like `AddRef`, which Arc implements and could make this clearer.

@kundadebdatta is absolutely right here though, it's safe to clone because the GEM contains small or `Arc`-wrapped types.

Given the difficulty in illustrating this, we should consider reworking the type a bit though. Instead of the GEM holding Arcs, it could directly own its data and then consumers of it would hold `Arc<GlobalEndpointManager>`. Pushing shared ownership as high up the stack as possible is a good design pattern to follow since it makes ownership clearer. Plus, direct ownership is always the clearest and cleanest option over shared ownership.

FabianMeiswinkel: Technically this is true - my comment is around how we can ensure the code is maintainable and easy to understand. In my mind - the consumer of collection_cache (for example) should not have to rely on whether AsyncCache stores its values in a Hashmap (which would result inAsyncCache.clone doing deepclone or Arc<HashMap> resuting in effectively getting the same hashmap from clone. What you want here semantically is shared onwership of a single collection_cache (and others) instance - so, better to model it that way and make it independent of how AsyncCache is implemented internally - meaning using an Arc<ColelctionCache> to indicate the shared ownershiop?
What is the problem with clearly marking the shared ownership of a singleton instance?

sdk/cosmos/azure_data_cosmos/src/routing/mod.rs
```diff
@@ -2,6 +2,11 @@
...
 
 pub(crate) mod async_cache;
+pub(crate) mod collection_routing_map;
 pub(crate) mod container_cache;
 pub(crate) mod global_endpoint_manager;
 pub(crate) mod location_cache;
+pub(crate) mod partition_key_range;
+pub(crate) mod partition_key_range_cache;
+pub(crate) mod range;
+pub(crate) mod service_identity;
```
analogrelay, line 12: We should change this so that the `routing` module isn't `pub` up in `lib.rs` rather than making all these `pub(crate)`. The layout should be something like this:

```
mod routing {
    pub mod global_endpoint_manager {
        pub struct GlobalEndpointManager {
```

Because `routing` is **private** at the crate route, **everything** within it (regardless of whether an individual item is `pub`) is hidden from consumers outside the crate, but still visible to _every other module in the crate_ (thus effectively `pub(crate)`.

Think of it this way: We make visibility decisions at the module level. Is `GlobalEndpointManager` a type that is needed outside of `global_endpoint_manager.rs` and it's child modules? If so (it certainly is), then it's defined as `pub`. Is `mod global_endpoint_module` needed outside of `routing`? If so, then it's `pub`. Is `mod routing` something needed by public consumers of the crate? **No**, so it's not `pub`.

`pub(crate)` is rarely needed, it's mostly for adding internal APIs to modules that already need to be `pub` for some reason, and it doesn't make sense to put that internal type in a different module.

sdk/cosmos/azure_data_cosmos/src/routing/partition_key_range.rs
analogrelay: As I noted in https://github.com/Azure/azure-sdk-for-rust/pull/3553/changes#r2713700476 , we need to rethink our default approach to visibility in Rust. In C#, visibility is always at the Assembly level, but in Rust it's always at the individual module level (more like JS with it's import/export syntax than C#).

`pub(crate)` and other `pub(*)` variants are escape hatches that are rarely needed.

sdk/cosmos/azure_data_cosmos/src/routing/partition_key_range.rs
```diff
@@ -0,0 +1,226 @@
...
+        self.id == other.id
+            && self.resource_id == other.resource_id
+            && self.min_inclusive == other.min_inclusive
+            && self.max_exclusive == other.max_exclusive
+    }
+}
+
+impl Eq for PartitionKeyRange {}
+
+// Implement Hash for PartitionKeyRange
```
analogrelay: A comment as to why we need to manually impl `Hash` would be appropriate here:


```suggestion
// Implement Hash for PartitionKeyRange manually because only the ID, RID, and min/max should be considered for equality/hashing.
```

sdk/cosmos/.dict.txt
```diff
@@ -29,4 +29,7 @@ msvc
 winget
 makefiles
 ctest
-cmake
\ No newline at end of file
+cmake
+
+# Common Rust syntax
+rfind
```
analogrelay, line 35: This could be moved up to the top-level Rust dict in the repo, but it's fine to keep it here for now and avoid blocking the PR on the Core Team's approval.

sdk/cosmos/azure_data_cosmos/src/routing/partition_key_range.rs
kundadebdatta: This is now updated.

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
kundadebdatta: Removed.

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
kundadebdatta: Removed.

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
kundadebdatta: Removed.

sdk/cosmos/azure_data_cosmos/src/handler/cosmos_connection.rs
kundadebdatta: This is now updated.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
kundadebdatta: It should be okay to use `clone()`, explained [in this comment.](https://github.com/Azure/azure-sdk-for-rust/pull/3553#discussion_r2713589401)

FabianMeiswinkel (approved):
As discussed offline - I am fine addressing the open discussing of clone vs. using Arc<xxxCache> for a follow-up PR

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
kundadebdatta: Created [this issue ](https://github.com/Azure/azure-sdk-for-rust/issues/3585)to address these tech-depth/ better design approaches.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
kundadebdatta: Created [this issue ](https://github.com/Azure/azure-sdk-for-rust/issues/3585)to address these tech-depth/ better design approaches.

sdk/cosmos/azure_data_cosmos/src/routing/mod.rs
kundadebdatta: Created [this issue ](https://github.com/Azure/azure-sdk-for-rust/issues/3585)to address these tech-depth/ better design approaches.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
kundadebdatta: Created [this issue ](https://github.com/Azure/azure-sdk-for-rust/issues/3585)to address these tech-depth/ better design approaches.

sdk/cosmos/azure_data_cosmos/src/routing/partition_key_range.rs
kundadebdatta: Created [this issue ](https://github.com/Azure/azure-sdk-for-rust/issues/3585)to address these tech-depth/ better design approaches.

sdk/cosmos/azure_data_cosmos/src/routing/range.rs
kundadebdatta: Created [this issue ](https://github.com/Azure/azure-sdk-for-rust/issues/3585)to address these tech-depth/ better design approaches.

PR #3556, opened 2026-01-15
[Storage] `BlobName` custom deserialization approach

vincenttran-msft:


sdk/storage/azure_storage_blob/tests/blob_client_options.rs
```diff
@@ -272,7 +267,7 @@ async fn test_list_blobs_with_versions(ctx: TestContext) -> Result<(), Box<dyn E
     let mut current_versions = 0;
 
     for blob_item in &blob_items {
-        let name = blob_item.name.as_ref().unwrap().content.as_ref().unwrap();
+        let name = blob_item.name.as_ref().unwrap();
```
vincenttran-msft, line 270: Here you can see no longer having to do `.content.as_ref().unwrap()` and instead `blob_item.name` is already `Option<String>`

sdk/storage/azure_storage_blob/src/generated/models/pub_models.rs
```diff
@@ -223,8 +223,13 @@ pub struct BlobItemInternal {
     pub metadata: Option<BlobMetadata>,
 
     /// The name of the blob.
-    #[serde(rename = "Name", skip_serializing_if = "Option::is_none")]
-    pub name: Option<BlobName>,
+    #[serde(
+        rename = "Name",
+        skip_serializing_if = "Option::is_none",
+        deserialize_with = "crate::models::deserialize_blob_name",
```
vincenttran-msft, line 229: So really, I think the ask is whether TypeSpec supports custom `deserialize_with` where the implementation is from the crate.

vincenttran-msft (commented):
proof of concept

sdk/storage/azure_storage_blob/src/generated/models/pub_models.rs
heaths: @jhendrixMSFT do we have any way to do this? This is cleaner certainly.

jhendrixMSFT: This will be enabled via https://github.com/Azure/typespec-rust/pull/841.
```ts
@@clientOption(Model.field, "deserialize_with", "path::to::your::deserializer", "rust");
```

jhendrixMSFT: Also note that at present this decorator requires a suppression in the tsp.
```ts
model Model {
    #suppress "@azure-tools/typespec-client-generator-core/client-option" "<reason for suppression>"
    field string;
}
```

vincenttran-msft:
We have the longterm solution in place.

PR #3559, opened 2026-01-16
Update emitter to 0.33.0 and release azure_core@0.31.0

heaths:
Also release dependencies of and crates related to azure_core e.g., typespec_client_core.

### DIFF
--- a/eng/emitter-package.json
+++ b/eng/emitter-package.json
@@ -6,7 +6,7 @@
   "devDependencies": {
     "@azure-tools/typespec-azure-core": "~0.64.0",
     "@azure-tools/typespec-azure-rulesets": "~0.64.0",
-    "@azure-tools/typespec-client-generator-core": "~0.64.0",
+    "@azure-tools/typespec-client-generator-core": "~0.64.1",
     "@typespec/compiler": "^1.8.0",
     "@typespec/http": "^1.8.0",
     "@typespec/openapi": "^1.8.0",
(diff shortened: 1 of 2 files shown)


eng/emitter-package.json
```diff
@@ -1,17 +1,17 @@
...
-    "@azure-tools/typespec-client-generator-core": "~0.63.0",
-    "@typespec/compiler": "^1.7.0",
-    "@typespec/http": "^1.7.0",
-    "@typespec/openapi": "^1.7.0",
-    "@typespec/rest": "~0.77.0",
-    "@typespec/versioning": "~0.77.0",
-    "@typespec/xml": "~0.77.0"
+    "@azure-tools/typespec-azure-core": "~0.64.0",
+    "@azure-tools/typespec-azure-rulesets": "~0.64.0",
+    "@azure-tools/typespec-client-generator-core": "~0.64.0",
```
jhendrixMSFT: Hmm this should be `~0.64.1` which is what we have in the emitter's `package.json` (strange that it didn't pick this up).

heaths: I authored it and must've missed that. Fixing now.

heaths: Is the `tsp-client generate-lock-file` supposed to update it? Seems it'd only update the lock file, not the authored file. It's an annoying task that maybe the Azure SDK MCP could do @maririos ?

heaths: That said, the version in the lockfile was correct since the package actually required it, but I'll update it anyway.

.vscode/mcp.json
```diff
@@ -0,0 +1,12 @@
+{
+  "servers": {
+    "azure-sdk-mcp": {
+      "type": "stdio",
+      "command": "pwsh",
+      "args": [
+        "${workspaceFolder}/eng/common/mcp/azure-sdk-mcp.ps1",
+        "-Run"
+      ]
+    },
```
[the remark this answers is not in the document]
heaths: Maybe next time.

PR #3560, opened 2026-01-16
Test with --all-features

### LINKED ISSUE #1746: Pass `--all-features` to any commands that use it, typically, e.g., linting.

heaths:
Resolves #1746

### DIFF
--- a/eng/scripts/Test-Packages.ps1
+++ b/eng/scripts/Test-Packages.ps1
@@ -20,7 +20,7 @@ function Invoke-CargoTestWithJsonOutput (
   # and uploaded to DevOps for display in the Tests tab
   # (requires -Z unstable-options)
   $result = Invoke-LoggedCommand `
-    "cargo +nightly test $TestParams --package $PackageName --no-fail-fast -- --format json -Z unstable-options" `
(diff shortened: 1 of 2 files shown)


eng/scripts/Test-Packages.ps1
```diff
@@ -91,6 +91,10 @@ foreach ($package in $packagesToTest) {
       $featuresArg = "--features " + ($features -join ",")
     }
   }
+  else {
+    # Allow existence of empty test-features.txt to denote only default features.
```
analogrelay: If `--all-features` tests are now working, I think we can remove the `test-features.txt` file altogether. It was really just a hack to allow Cosmos to run "all features" while other crates were failing.

heaths: Good point. Let's see if this passes first just in case we need to keep it for whatever reason (doubtful). If it does, I'll remove the support entirely along with your file.

heaths: Shoot, I forgot to exclude some other work. Well, since I have to upgrade this anyway, I'll just make those changes now.

heaths:
@analogrelay does it look like it ran the Cosmos tests you expected?

PR #3575, opened 2026-01-21
[Storage] Default logging options for Blobs package

vincenttran-msft:
As title states.
Also re-exports `LoggingOptions` from `azure_core` since I believe the intention is that they shouldn't be from `typespec_client_core` directly if we expect a customer to use it?

### DIFF
--- a/sdk/core/azure_core_macros/src/tracing_new.rs
+++ b/sdk/core/azure_core_macros/src/tracing_new.rs
@@ -713,4 +713,57 @@ mod tests {
             "Parsed tokens do not match expected tokens"
         );
     }
+
+    #[test]
+    fn parse_new_with_arbitrary_function_call() {
+        // Test that arbitrary function calls (not Ok/Arc::new) are passed through unchanged,
+        // preserving their semicolons.
+        setup_tracing();
+        let attr = quote!("Az.TestNamespace");
+        let new_function = quote! {
+            pub fn from_url(
+                url: Url,
+                options: Option<ClientOptions>,
+            ) -> Result<Self> {
+                let mut options = options.unwrap_or_default();
+                apply_defaults(&mut options.client_options);
+                Ok(Self {
+                    url,
+                })
+            }
+        };
+        let actual =
+            parse_new(attr, new_function).expect("Failed to parse new function declaration");
+
+        println!("Parsed tokens: {actual}");
+
+        let expected = quote! {
+            pub fn from_url(
+                url: Url,
+                options: Option<ClientOptions>
+            ) -> Result<Self> {
+                let mut options = options.unwrap_or_default();
+                apply_defaults(&mut options.client_options);
+                Ok(Self {
+                    tracer: options
+                        .client_options
+                        .instrumentation
+                        .tracer_provider
+                        .as_ref()
+                        .map(|tracer_provider| {
+                            tracer_provider.get_tracer(
+                                Some("Az.TestNamespace"),
+                                option_env!("CARGO_PKG_NAME").unwrap_or("UNKNOWN"),
+                                option_env!("CARGO_PKG_VERSION"),
+                            )
+                        }),
+                    url,
+                })
+            }
+        };
+        assert!(
+            crate::tracing::tests::compare_token_stream(actual, expected),
+            "Parsed tokens do not match expected tokens"
+        );
+    }
 }


sdk/storage/azure_storage_blob/src/logging.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// Applies the default Azure Storage Blob logging configuration.
+    ///
+    /// This method prepends the storage-specific allowed headers and query parameters
+    /// to the user's existing logging options. User-specified options are preserved and
+    /// take effect in addition to the storage defaults.
+    fn apply_storage_logging_defaults(&mut self);
+}
+
+impl StorageLoggingExt for ClientOptions {
+    fn apply_storage_logging_defaults(&mut self) {
```
heaths: Suggestion: why not just have this as a bare function that you just pass a `&ClientOptions` into? You need to import the trait or fn either way, so keep it simple. It's private either way, so ergonomics is less important.

sdk/storage/azure_storage_blob/examples/storage_logging.rs
```diff
@@ -0,0 +1,118 @@
...
+    // Create OAuth credentials using Azure CLI
+    println!("Authenticating with Azure CLI...");
+    let credential = AzureCliCredential::new(None)?;
+
+    // Create BlobContainerClient
+    let endpoint = format!("https://{}.blob.core.windows.net", account);
+
+    // Azure Storage headers (x-ms-version, x-ms-request-id, etc.) are logged by default.
+    // No additional logging configuration is required for standard tracing output.
+    let client_options = BlobContainerClientOptions::default();
```
heaths: The client does this, so a minimal case would just be that you pass `None` to `BlobContainerClient::new()`. The result will be the same.

sdk/storage/azure_storage_blob/examples/storage_logging.rs
```diff
@@ -0,0 +1,118 @@
...
+        Some(credential),
+        Some(client_options),
+    )?;
+
+    // Create BlobClient
+    let blob_client = container_client.blob_client(blob_name);
+
+    // Create container if does not exist
+    println!("Creating container '{}'...", container_name);
+    match container_client.create_container(None).await {
```
jalauzon-msft: The error handling in this call is probably not a great example to have in our samples as it uses string matching and not the new error stuff we are working on. I would leave this out by either just not caring if it exists or using our `exists` method.

sdk/storage/azure_storage_blob/src/logging.rs
```diff
@@ -0,0 +1,191 @@
...
+    "x-ms-has-immutability-policy",
+    "x-ms-has-legal-hold",
+    // Range headers
+    "x-ms-range",
+    "x-ms-source-range",
+    // Encryption
+    "x-ms-request-server-encrypted",
+    "x-ms-server-encrypted",
+    "x-ms-encryption-algorithm",
+    "x-ms-encryption-key-sha256",
```
jalauzon-msft: Was this in the .NET list? Feels like something we should redact by default...

vincenttran-msft: [Yep:](https://github.com/Azure/azure-sdk-for-net/blob/9b4720f76a24d85a13dacc9b652b5bfb4bf6fe50/sdk/storage/Azure.Storage.Blobs/src/BlobClientOptions.cs#L287)
>L287 Diagnostics.LoggedHeaderNames.Add("x-ms-request-server-encrypted");
   L288 Diagnostics.LoggedHeaderNames.Add("x-ms-server-encrypted");
   L316 Diagnostics.LoggedHeaderNames.Add("x-ms-encryption-algorithm");
   L338 Diagnostics.LoggedHeaderNames.Add("x-ms-encryption-key-sha256");

Let me know if you have strong feelings towards redacting these in Rust.

sdk/core/azure_core_macros/src/tracing_new.rs
```diff
@@ -80,6 +80,16 @@ fn is_arc_new_call(func: &syn::Expr) -> bool {
...
+/// Checks if the call expression is `Ok(...)` or `Arc::new(...)`.
+fn is_ok_or_arc_new_call(call: &syn::ExprCall) -> bool {
+    if let syn::Expr::Path(path) = call.func.as_ref() {
+        let last_ident = &path.path.segments.last().unwrap().ident;
+        last_ident == "Ok" || is_arc_new_call(call.func.as_ref())
+    } else {
+        false
+    }
+}
+
```
vincenttran-msft: This was written heavily with the help with Copilot, as originally when switching to the bare function call, there was tons of "missing semicolons" compiler errors all over the place. Here is the summary of the changes after authoring this change:
>Fixed a bug where arbitrary function calls (like `apply_defaults(...))` in client new functions were incorrectly processed by the macro, potentially dropping semicolons. Now only `Ok(...)` and `Arc::new(...)` calls are transformed; other calls pass through unchanged.

sdk/storage/azure_storage_blob/src/logging.rs
jalauzon-msft: Specifically, `x-ms-encryption-key-sha256` should probably be left out. It is the shae256 of the actual encryption key.

sdk/core/azure_core_macros/src/tracing_new.rs
```diff
@@ -173,14 +183,14 @@ pub fn parse_new(attr: TokenStream, item: TokenStream) -> Result<TokenStream> {
...
-                        if c.args.len() != 1 {
-                            trace!("Call expression does not have exactly one argument, emitting statement: {stmt:?}");
-                            // If the call does not have exactly one argument, just return it as is.
-                             stmt.to_token_stream()
-                        }
-                        else {
+                        // Only process single-argument calls that are Ok(...) or Arc::new(...).
+                        // Other function calls should be passed through unchanged to preserve
+                        // the original statement (including semicolon).
+                        if c.args.len() == 1 && is_ok_or_arc_new_call(c) {
```
heaths: I don't understand why this is needed. If the original expression is a final return when a semicolon shouldn't be used unless the whole expression is preceeded by a `return`, why change it at all? If the original expression is wrong, the macro shouldn't change it. That's not the macros intent to fix lint.

Can you show me an example of where/how it was failing? When was something returning `Arc::new(..)` from a traced expression vs. `Ok(..)`?

/cc @LarryOsterman

sdk/storage/azure_storage_blob/src/logging.rs
```diff
@@ -0,0 +1,169 @@
...
+];
+
+/// Applies the default Azure Storage Blob logging configuration to client options.
+///
+/// This function prepends the storage-specific allowed headers and query parameters
+/// to the user's existing logging options. User-specified options are preserved and
+/// take effect in addition to the storage defaults.
+pub(crate) fn apply_storage_logging_defaults(options: &mut ClientOptions) {
+    // Prepend storage-specific headers to any user-specified headers
+    let user_headers = std::mem::take(&mut options.logging.additional_allowed_header_names);
```
heaths: Since `options` is already `&mut`, why do you need `take()` here?

```rust
let headers = &mut options.logging.additional_allowed_header_names);
headers.extend(STORAGE_ALLOWED_HEADERS); // or maybe by reference
```

sdk/core/azure_core_macros/src/tracing_new.rs
vincenttran-msft: Certainly. So reverting `tracing_new.rs` back to main, yields this:
image

Error (I only pasted it for `BlobClient` but there are 6x instances of this, one for each client):
```
error: expected `;`, found keyword `let`
  --> sdk\storage\azure_storage_blob\src\clients\blob_client.rs:64:68
   |
64 |         apply_storage_logging_defaults(&mut options.client_options);
   |                                                                    ^ help: add `;` here
65 |
66 |         let storage_headers_policy = Arc::new(StorageHeadersPolicy);
   |         --- unexpected token
```

As mentioned above, the solution that is currently proposed was written with Copilot, with it citing this as the reasoning for having to implement its "fix":
>Fixed a bug where arbitrary function calls (like apply_defaults(...)) in client new functions were incorrectly processed by the macro, potentially dropping semicolons. Now only Ok(...) and Arc::new(...) calls are transformed; other calls pass through unchanged.

heaths: Okay, I think I kinda get what's going on here. But this still feels more like a workaround than a proper fix. I'd rather @LarryOsterman decide whether to go with what you have here for now and fix it properly in another PR, or if the real fix is simple enough just inline it.

sdk/core/azure_core_macros/src/tracing_new.rs
LarryOsterman: Why were arbitrary function calls decorated with `#[tracing::new]`? That should only be applied to public constructor functions.

sdk/core/azure_core_macros/src/tracing_new.rs
LarryOsterman: Ok - I've figured out the root cause. In this case, copilot correctly identified the bug - so your fix does work but it's not sufficient (and more importantly, it will introduce unnecessary fragility that will probably cause problems down the line).

The problem is that `foo(abc)` is an *expression* and `foo(abc);` is a *statement*. And the `parse_call_expr()` returns a token stream containing the *expression* if the function being called doesn't match the expected pattern for `new` operations. That's why the trailing semicolon disappeared, triggering the bug.

The correct fix is to have `parse_call_expr` return an `Option<TokenStream>` and if it returns None, call `stmt.to_token_stream()` (as is currently done). And obviously if it returns `Some(s)`, it should return the `s`.

This also means changing the `parse_call_expr` function to change all the places where it calls `call.to_token_stream()` to instead return `None`.

The reason I prefer this solution is that if we ever run into a future case where we need to add special handling of call expressions (in other words if we find a call that we need to modify that isn't `Ok(xxx)` or `Arc::new(xxx)`, then my proposed solution only requires modifications to parse_call_expr, but the copilot generated solution requires modifying (and renaming) the `is_ok_or_arc_new()` function.

And honestly, it's the way this should have been written in the first place.

Does that make sense?

vincenttran-msft: Gotcha. I did my best attempt at it in the next iteration. If it doesn't seem sufficient, please feel free to just commit directly to this branch (especially since this _is_ Core code anyways 😛)

sdk/core/azure_core_macros/src/tracing_new.rs
```diff
@@ -718,57 +713,4 @@ mod tests {
             "Parsed tokens do not match expected tokens"
         );
     }
-
-    #[test]
```
LarryOsterman: FWIW, I'd keep this test. This was a clear missed case in the original implementation, and it would be good to keep the test to make sure it doesn't get regressed.

heaths (approved):

jalauzon-msft (approved):

PR #3577, opened 2026-01-21
Initial azure_resourcemanager_keyvault client library

heaths:


heaths:
I'm going to split out Copilot instructions and probably the markdownlint changes into separate PRs just to keep everything focused, but wanted to get a preview of what I have working so far.

heaths:
Will rebase on #3587 once merged.

sdk/keyvault/azure_resourcemanager_keyvault/src/clients.rs
```diff
@@ -0,0 +1,167 @@
...
+                )
+            })
+        }
+        _ => Err(Error::with_message(
+            ErrorKind::Other,
+            "cloud configuration is not supported",
+        )),
+    }?;
+    let mut scope = Url::parse(audience)?;
+    scope.set_path("/.default");
```
chlowell: I believe the audience for ARM is always a URL, however it's best to treat it as an opaque string and simply append `/.default` to it because
- we should trust callers to know what they're doing, and let Azure decide whether the audience is correct
- parsing could mutate the audience e.g. by trimming a trailing `/`. We don't want that because `https://foo//.default` and `https://foo/.default` are different scopes

sdk/keyvault/azure_resourcemanager_keyvault/src/clients.rs
```diff
@@ -0,0 +1,167 @@
...
+    cloud::{CloudConfiguration, CustomConfiguration},
+    error::ErrorKind,
+    http::Url,
+    Error, Result,
+};
+pub use key_vault_clients::KeyVaultClientOptions;
+
+/// Audience for Azure Key Vault Resource Manager requests.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+pub struct Audience;
```
chlowell, line 19: Yep, this is the pattern I imagined. Future design question: could/should we have a single type for ARM?

sdk/keyvault/azure_resourcemanager_keyvault/src/clients.rs
heaths: > We don't want that because https://foo//.default and https://foo/.default are different scopes

🤨

Given that, should I trim the trailing slashes from the implementation here? I'm assuming we just want `foo/.default` and not `foo//.default`, right?

sdk/keyvault/azure_resourcemanager_keyvault/src/clients.rs
heaths: Wouldn't each crate potentially have a different scope? Or maybe I misunderstood you the other day.

sdk/keyvault/azure_resourcemanager_keyvault/src/clients.rs
chlowell: No; the SDK shouldn't have an opinion about a `CustomConfiguration` audience. It may be that only one of `https://foo//.default` and `https://foo/.default` is valid. If we prefer one over the other, we may choose wrongly for--and block--some scenario.

sdk/keyvault/azure_resourcemanager_keyvault/src/clients.rs
chlowell: All ARM services in a given cloud have the same audience(s). It's therefore reasonable to have a single `Audience` type  that applies to all ARM clients, but off the top of my head I don't know how much convenience for developers we could create that way.

sdk/keyvault/azure_resourcemanager_keyvault/src/clients.rs
heaths: My point is that I wrote:

```rust
       CloudConfiguration::AzurePublic => Ok("https://management.core.windows.net/"),
```

So I added that trailing slash, which will result in `https://management.core.windows.net//.default`. But I could just as easily define these as:

```rust
       CloudConfiguration::AzurePublic => Ok("https://management.core.windows.net"),
```

Without the trailing slash if the `//.default` is wrong in this case. If you're saying that I should keep the trailing slash and simply append `/.default`, I will. I just want to make sure we're on the same page. The <https://learn.microsoft.com/entra/identity-platform/scopes-oidc#the-default-scope> link explained the difference between `/` and `//` but didn't really answer which is appropriate in *this* case.

I mean, all the endpoints lack a trailing slash, though I know there's not a strong correlation there.

heaths:
Superseded by #4289

PR #3593, opened 2026-01-22
Fix cargo-semver-checks against new doc version

heaths:
Rust 1.93 bumped the doc version to v57 which cargo-semver-checks 0.45.0 (still latest) does not yet support.

### DIFF
--- a/eng/pipelines/templates/jobs/pack.yml
+++ b/eng/pipelines/templates/jobs/pack.yml
@@ -49,7 +49,7 @@ jobs:
 
   - template: /eng/pipelines/templates/steps/use-rust.yml@self
     parameters:
-      # 19.3 just came out and bumped the doc version to v57, which cargo-semver-checks does not yet support.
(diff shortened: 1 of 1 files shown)


heaths:
To note, I tried to use the MSRV - 1.85 - locally with semver-checks 0.45.0 but it's too old. semver-checks said it requires 1.89 or newer. So I figured I'd go with the last known good build of 1.92.

We'll probably want to do something along these lines in perpetuity, however. Anytime the latest stable release bumps the doc version there will be a problem. Can't say I've seen this before here, but we also haven't had semver-checks integrated that long. Could be a regular thing, and I imagine semver-checks kicks out older versions. Wonder if we should have a little script to use Rust version N-1. Then again, we currently lock to semver-checks 0.45.0 so we'd need to revisit that as well. I suppose as long as we lock to it (we should probably define those versions in a better place - maybe even parameters at the top here), at least we're in a *known* stable.

weshaggard (approved):

PR #3599, opened 2026-01-23
Fault Injection

tvaron3:
### Summary

Adds a fault injection framework behind the `fault_injection` feature flag that allows simulating HTTP-level failures (503, 500, 429, 408, 410, etc.) against the Cosmos DB client. This enables testing retry policies, regional failover, and error handling without requiring actual service failures.

### Architecture

```text
┌──────────────────────────────────────────────────────┐
│  CosmosClient                                        │
│  ┌────────────────────────────────────────────────┐  │
│  │  GatewayPipeline                               │  │
│  │  ┌──────────────┐   ┌───────────────────────┐  │  │
│  │  │ Retry Policy │──▶│ FaultClient (HttpClient)││  │
│  │  └──────────────┘   │  ┌─────────────────┐   │ │  │
│  │                     │  │ Rule matching    │   ││  │
│  │                     │  │ (operation type, │   ││  │
│  │                     │  │  region, container)│  │   │
│  │                     │  └────────┬────────┘   │  │   │
│  │                     │       match? │         │  │   │
│  │                     │     ┌───────┴───────┐  │  │   │
│  │                     │  ┌──▼──┐      ┌─────▼┐ │  │   │
│  │                     │  │Fault│      │ Real  │ │  │   │
│  │                     │  │Error│      │Request│ │  │   │
│  │                     │  └─────┘      └──────┘ │  │   │
│  │                     └───────────────────────┘  │   │
│  └────────────────────────────────────────────────┘   │
└──────────────────────────────────────────────────────┘
```

The FaultClient wraps the real HttpClient at the transport layer, intercepting requests below the retry policy. This means injected faults trigger the same retry/failover behavior as real service errors.

### Usage

```rust
use azure_data_cosmos::fault_injection::{
    FaultInjectionClientBuilder, FaultInjectionConditionBuilder,
    FaultInjectionErrorType, FaultInjectionResultBuilder,
    FaultInjectionRuleBuilder, FaultOperationType,
};
use azure_data_cosmos::CosmosClientOptions;
use std::sync::Arc;
use std::time::{Duration, Instant};

// 1. Define what error to inject
let error = FaultInjectionResultBuilder::new()
    .with_error(FaultInjectionErrorType::ServiceUnavailable)
    .with_delay(Duration::from_millis(100))
    .with_probability(1.0)
    .build();

// 2. Define when to inject it
let condition = FaultInjectionConditionBuilder::new()
    .with_operation_type(FaultOperationType::ReadItem)
    .with_region("West US".into())
    .build();

// 3. Create a rule (Arc enables runtime enable/disable)
let rule = Arc::new(FaultInjectionRuleBuilder::new("region-failover-test", error)
    .with_condition(condition)
    .with_hit_limit(5)
    .with_end_time(Instant::now() + Duration::from_secs(30))
    .build());

// 4. Inject into client options
let options = FaultInjectionClientBuilder::new()
    .with_rule(rule.clone())
    .inject(CosmosClientOptions::default());

let client = CosmosClient::new(&endpoint, credential, Some(options))?;
// Reads targeting "West US" will now fail with 503

// Rules can be toggled at runtime
rule.disable(); // temporarily stop injecting faults
rule.enable();  // resume fault injection
```

### Rule configuration

```text
┌────────────┬──────────────────────────────────────────────────────────┐
│ Parameter  │ Description                                              │
├────────────┼──────────────────────────────────────────────────────────┤
│ error_type │ HTTP error to inject (503, 500, 429, 408, 410, 403, 404) │
├────────────┼──────────────────────────────────────────────────────────┤
│ delay      │ Artificial latency applied after the request             │
├────────────┼──────────────────────────────────────────────────────────┤
│ probability│ 0.0–1.0 chance of fault injection per request            │
├────────────┼──────────────────────────────────────────────────────────┤
│ hit_limit  │ Max number of times the rule fires                       │
├────────────┼──────────────────────────────────────────────────────────┤
│ start_time │ Absolute time at which the rule becomes active            │
├────────────┼──────────────────────────────────────────────────────────┤
│ end_time   │ Absolute time at which the rule expires                   │
├────────────┼──────────────────────────────────────────────────────────┤
│ enabled    │ Runtime toggle via enable()/disable() on shared Arc rule  │
├────────────┼──────────────────────────────────────────────────────────┤
│ condition  │ Filter by operation type, region, or container            │
└────────────┴──────────────────────────────────────────────────────────┘
```

### Changes by area

**New: `src/fault_injection/` module (1,394 lines)**

- `mod.rs` — module docs, `FaultOperationType` and `FaultInjectionErrorType` enums
- `http_client.rs` — `FaultClient` HttpClient wrapper that intercepts requests and injects faults
- `result.rs` — `FaultInjectionResult` / `FaultInjectionResultBuilder` for error type, delay, probability
- `condition.rs` — `FaultInjectionCondition` / `FaultInjectionConditionBuilder` for operation type, region, container filters
- `rule.rs` — `FaultInjectionRule` / `FaultInjectionRuleBuilder` with timing, hit limits, and runtime enable/disable
- `client_builder.rs` — `FaultInjectionClientBuilder` to inject fault client into `CosmosClientOptions`

**Modified: pipeline integration**

- `cosmos_request.rs` — adds `add_fault_injection_headers()` to tag requests with operation type
- `pipeline/mod.rs` — calls fault injection header injection when feature enabled
- `constants.rs` — adds internal `FAULT_INJECTION_OPERATION` header
- `options/mod.rs` — adds `fault_injection_enabled` field (feature-gated)
- `clients/cosmos_client.rs` — threads `fault_injection_enabled` through to pipeline
- `Cargo.toml` — adds `fault_injection` feature flag

**Modified: documentation fix**

- `routing/location_cache.rs` — fixes stale doc comment on `get_applicable_endpoints()`

**New tests (37 total):**

- `emulator_tests/cosmos_fault_injection.rs` — 12 integration tests (probability, hit limits, timing, container targeting, rule priority, enable/disable, metadata isolation)
- `multi_write_tests/cosmos_multi_write_fault_injection.rs` — 9 integration tests (multi-region write failover with fault injection)
- `src/fault_injection/http_client.rs` — 11 unit tests (rule matching, error injection, substatus headers, delays)
- `src/fault_injection/{rule,result,condition}.rs` — 5 unit tests (builders, enable/disable, probability clamping)
- `framework/test_client.rs` — test framework support for fault injection clients

### Design decisions

- **Rules behind `Arc`**: `FaultInjectionClientBuilder::with_rule()` accepts `Arc<FaultInjectionRule>` so callers can retain a handle for runtime `enable()`/`disable()` toggling via `AtomicBool`.
- **Absolute timing**: Rules use `start_time: Instant` and `end_time: Option<Instant>` instead of relative delays/durations for clarity and auditability. Defaults: starts immediately, never expires.
- **Consuming builder**: Builder methods take `self` (not `&mut self`) to prevent reuse of partially-configured builders.
- **Private fields + getters**: All struct fields are private with public getter methods for immutability.
- **Transport-layer injection**: Faults are injected below the retry policy so they trigger the same retry/failover behavior as real service errors.

### Follow ups

- Add connection errors and retry policy for relevant errors https://github.com/Azure/azure-sdk-for-python/issues/45052

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/fault_injection/condition.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/fault_injection/condition.rs
@@ -10,28 +10,11 @@ use crate::regions::RegionName;
 #[derive(Clone, Default, Debug)]
 pub struct FaultInjectionCondition {
     /// The type of operation to which the fault injection applies.
-    operation_type: Option<FaultOperationType>,
+    pub operation_type: Option<FaultOperationType>,
     /// The region to which the fault injection applies.
-    region: Option<RegionName>,
+    pub region: Option<RegionName>,
     /// The container ID to which the fault injection applies.
-    container_id: Option<String>,
-}
-
-impl FaultInjectionCondition {
-    /// Returns the operation type this condition matches, if set.
-    pub fn operation_type(&self) -> Option<FaultOperationType> {
-        self.operation_type
-    }
-
-    /// Returns the region this condition matches, if set.
-    pub fn region(&self) -> Option<&RegionName> {
-        self.region.as_ref()
-    }
-
-    /// Returns the container ID this condition matches, if set.
-    pub fn container_id(&self) -> Option<&str> {
-        self.container_id.as_deref()
-    }
+    pub container_id: Option<String>,
 }
 
 /// Builder for creating a FaultInjectionCondition.
@@ -93,8 +76,8 @@ mod tests {
     fn builder_default() {
         let builder = FaultInjectionConditionBuilder::default();
         let condition = builder.build();
-        assert!(condition.operation_type().is_none());
-        assert!(condition.region().is_none());
-        assert!(condition.container_id().is_none());
+        assert!(condition.operation_type.is_none());
+        assert!(condition.region.is_none());
+        assert!(condition.container_id.is_none());
     }
 }
--- a/sdk/cosmos/azure_data_cosmos/src/fault_injection/mod.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/fault_injection/mod.rs
@@ -218,8 +218,11 @@ impl fmt::Display for FaultOperationType {
 }
 
 impl FromStr for FaultOperationType {
-    type Err = String;
+    type Err = ();
 
+    /// Parses a string into a `FaultOperationType`.
+    ///
+    /// Returns `Err(())` if the string is not a recognized operation type.
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
             "ReadItem" => Ok(FaultOperationType::ReadItem),
@@ -235,7 +238,7 @@ impl FromStr for FaultOperationType {
             "MetadataReadDatabaseAccount" => Ok(FaultOperationType::MetadataReadDatabaseAccount),
             "MetadataQueryPlan" => Ok(FaultOperationType::MetadataQueryPlan),
             "MetadataPartitionKeyRanges" => Ok(FaultOperationType::MetadataPartitionKeyRanges),
-            _ => Err(format!("unknown FaultOperationType: {}", s)),
+            _ => Err(()),
         }
     }
 }
--- a/sdk/cosmos/azure_data_cosmos/src/fault_injection/result.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/fault_injection/result.rs
@@ -11,24 +11,14 @@ use super::FaultInjectionErrorType;
 #[derive(Clone, Debug)]
 pub struct FaultInjectionResult {
     /// The type of server error to inject.
-    error_type: Option<FaultInjectionErrorType>,
+    pub error_type: Option<FaultInjectionErrorType>,
     /// Delay before injecting the error.
-    delay: Duration,
+    pub delay: Duration,
     /// Probability of injecting the error (0.0 to 1.0).
     probability: f32,
 }
 
 impl FaultInjectionResult {
-    /// Returns the error type to inject, if set.
-    pub fn error_type(&self) -> Option<FaultInjectionErrorType> {
-        self.error_type
-    }
-
-    /// Returns the delay applied after the fault response.
-    pub fn delay(&self) -> Duration {
-        self.delay
-    }
-
     /// Returns the probability of injecting the fault (0.0 to 1.0).
     pub fn probability(&self) -> f32 {
         self.probability
@@ -99,11 +89,8 @@ mod tests {
             .with_error(FaultInjectionErrorType::Timeout)
             .build();
 
-        assert_eq!(
-            error.error_type().unwrap(),
-            FaultInjectionErrorType::Timeout
-        );
-        assert_eq!(error.delay(), Duration::ZERO);
+        assert_eq!(error.error_type.unwrap(), FaultInjectionErrorType::Timeout);
+        assert_eq!(error.delay, Duration::ZERO);
         assert!((error.probability() - 1.0).abs() < f32::EPSILON);
     }
 
--- a/sdk/cosmos/azure_data_cosmos/src/fault_injection/rule.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/fault_injection/rule.rs
@@ -13,17 +13,17 @@ use super::result::FaultInjectionResult;
 #[derive(Debug)]
 pub struct FaultInjectionRule {
     /// The condition under which to inject the fault.
-    condition: FaultInjectionCondition,
+    pub condition: FaultInjectionCondition,
     /// The result to inject when the condition is met.
-    result: FaultInjectionResult,
+    pub result: FaultInjectionResult,
     /// The absolute time at which the rule becomes active.
-    start_time: Instant,
+    pub start_time: Instant,
     /// The absolute time at which the rule expires, if set.
-    end_time: Option<Instant>,
+    pub end_time: Option<Instant>,
     /// The total hit limit of the rule.
-    hit_limit: Option<u32>,
+    pub hit_limit: Option<u32>,
     /// Unique identifier for the fault injection scenario.
-    id: String,
+    pub id: String,
     /// Whether the rule is currently enabled.
     enabled: AtomicBool,
 }
@@ -43,36 +43,6 @@ impl Clone for FaultInjectionRule {
 }
 
 impl FaultInjectionRule {
-    /// Returns the condition under which the fault is injected.
-    pub fn condition(&self) -> &FaultInjectionCondition {
-        &self.condition
-    }
-
-    /// Returns the result to inject when the condition is met.
-    pub fn result(&self) -> &FaultInjectionResult {
-        &self.result
-    }
-
-    /// Returns the absolute time at which the rule becomes active.
-    pub fn start_time(&self) -> Instant {
-        self.start_time
-    }
-
-    /// Returns the absolute time at which the rule expires, if set.
-    pub fn end_time(&self) -> Option<Instant> {
-        self.end_time
-    }
-
-    /// Returns the total hit limit of the rule, if set.
-    pub fn hit_limit(&self) -> Option<u32> {
-        self.hit_limit
-    }
-
-    /// Returns the unique identifier for this rule.
-    pub fn id(&self) -> &str {
-        &self.id
-    }
-
     /// Returns whether the rule is currently enabled.
     pub fn is_enabled(&self) -> bool {
         self.enabled.load(std::sync::atomic::Ordering::SeqCst)
@@ -183,12 +153,12 @@ mod tests {
         let before = Instant::now();
         let rule = FaultInjectionRuleBuilder::new("test-rule", create_test_error()).build();
 
-        assert_eq!(rule.id(), "test-rule");
-        assert!(rule.start_time() >= before);
-        assert!(rule.start_time() <= Instant::now());
-        assert!(rule.end_time().is_none());
-        assert!(rule.hit_limit().is_none());
-        assert!(rule.condition().operation_type().is_none());
+        assert_eq!(rule.id, "test-rule");
+        assert!(rule.start_time >= before);
+        assert!(rule.start_time <= Instant::now());
+        assert!(rule.end_time.is_none());
+        assert!(rule.hit_limit.is_none());
+        assert!(rule.condition.operation_type.is_none());
         assert!(rule.is_enabled());
     }
 }
--- a/sdk/cosmos/azure_data_cosmos/src/fault_injection/http_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/fault_injection/http_client.rs
@@ -61,19 +61,19 @@ impl FaultClient {
         }
 
         // Check if the rule has started
-        if now < rule.start_time() {
+        if now < rule.start_time {
             return false;
         }
 
         // Check if the rule has expired
-        if let Some(end_time) = rule.end_time() {
+        if let Some(end_time) = rule.end_time {
             if now >= end_time {
                 return false;
             }
         }
 
         // Check if we've exceeded the hit limit on the rule
-        if let Some(hit_limit) = rule.hit_limit() {
+        if let Some(hit_limit) = rule.hit_limit {
             if rule_state.hit_count.load(Ordering::SeqCst) >= hit_limit {
                 return false;
             }
@@ -84,11 +84,11 @@ impl FaultClient {
 
     /// Checks if the request matches the rule's condition.
     fn matches_condition(&self, request: &Request, rule: &FaultInjectionRule) -> bool {
-        let condition = rule.condition();
+        let condition = &rule.condition;
         let mut matches = true;
 
         // Check operation type if specified
-        if let Some(expected_op) = condition.operation_type() {
+        if let Some(expected_op) = condition.operation_type {
             let request_op = request
                 .headers()
                 .get_optional_str(&constants::FAULT_INJECTION_OPERATION)
@@ -106,14 +106,14 @@ impl FaultClient {
         }
 
         // Check region if specified
-        if let Some(region) = condition.region() {
+        if let Some(region) = &condition.region {
             if !request.url().as_str().contains(region.as_str()) {
                 matches = false;
             }
         }
 
         // Check container ID if specified
-        if let Some(container_id) = condition.container_id() {
+        if let Some(container_id) = &condition.container_id {
             if !request.url().as_str().contains(container_id) {
                 matches = false;
             }
@@ -144,7 +144,7 @@ impl FaultClient {
         }
 
         // Generate the appropriate error based on error type
-        let error_type = match server_error.error_type() {
+        let error_type = match server_error.error_type {
             Some(et) => et,
             None => return None, // No error type set, pass through
         };
@@ -234,7 +234,7 @@ impl HttpClient for FaultClient {
                 // Increment hit count
                 rule_state.hit_count.fetch_add(1, Ordering::SeqCst);
                 // Clone and return the result
-                Some(rule_state.rule.result().clone())
+                Some(rule_state.rule.result.clone())
             } else {
                 None
             }
@@ -262,8 +262,8 @@ impl HttpClient for FaultClient {
 
         // Apply delay after the request is sent
         if let Some(result) = fault_result {
-            if result.delay() > Duration::ZERO {
-                let delay = azure_core::time::Duration::try_from(result.delay())
+            if result.delay > Duration::ZERO {
+                let delay = azure_core::time::Duration::try_from(result.delay)
                     .unwrap_or(azure_core::time::Duration::ZERO);
                 azure_core::async_runtime::get_async_runtime()
                     .sleep(delay)
--- a/sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_fault_injection.rs
+++ b/sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_fault_injection.rs
@@ -926,7 +926,7 @@ pub async fn fault_injection_enable_disable_rule() -> Result<(), Box<dyn Error>>
             .build(),
     );
 
-    assert_eq!(rule.id(), "enable-disable-test");
+    assert_eq!(rule.id, "enable-disable-test");
     assert!(rule.is_enabled());
 
     let rule_handle = Arc::clone(&rule);


sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_http_client.rs
```diff
@@ -0,0 +1,540 @@
...
+                "Partition Is Gone - Injected fault",
+            ),
+            FaultInjectionErrorType::WriteForbidden => (
+                StatusCode::Forbidden,
+                Some(SubStatusCode::WRITE_FORBIDDEN),
+                "Write Forbidden - Injected fault",
+            ),
+            FaultInjectionErrorType::DatabaseAccountNotFound => (
+                StatusCode::Forbidden,
+                Some(SubStatusCode::DATABASE_ACCOUNT_NOT_FOUND),
```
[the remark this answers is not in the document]
tvaron3: Will add tests for these

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_http_client.rs
```diff
@@ -0,0 +1,540 @@
...
+                .duration_since(std::time::UNIX_EPOCH)
+                .unwrap_or_default()
+                .subsec_nanos();
+            let random = (nanos % 1000) as f32 / 1000.0;
+            if !server_error.probability.is_finite()
+                || server_error.probability <= 0.0
+                || random >= server_error.probability
+            {
+                return None; // Don't inject fault this time
+            }
```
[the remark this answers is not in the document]
tvaron3: There is a comment saying that it is a // Use a simple time-based pseudo-random check

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_http_client.rs
```diff
@@ -0,0 +1,540 @@
...
+    fn is_rule_applicable(&self, rule_state: &mut RuleState) -> bool {
+        let elapsed = self.created_at.elapsed();
+        let rule = &rule_state.rule;
+
+        // Check if we've passed the start delay
+        if elapsed < rule.start_delay {
+            return false;
+        }
+
+        // Activate the rule if not already activated
```
ananth7592: What if the testing is explicitly de-activating the rule and activating it again all in the scope of a single client instance?

tvaron3: There is currently no way to activate a deactivated rule. To achieve that behavior you could do something like the following: create a rule that deactivates after 1 min and a duplicate rule with a start delay of 3 mins. This would simulate deactivating a rule and activating it in one single client instance.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_client_builder.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// Creates a new FaultInjectionClientBuilder.
+    pub fn new() -> Self {
+        Self { rules: Vec::new() }
+    }
+
+    /// Injects the fault injection client into the CosmosClientOptions.
+    /// Called after building the fault conditions.
+    ///
+    /// This wraps the existing transport (or creates a default one) with the fault injection client.
+    pub fn inject(&self, mut options: CosmosClientOptions) -> CosmosClientOptions {
```
analogrelay: IMO, this should consume `self`. It's intended to be the final method called on this type.

```suggestion
    pub fn inject(self, mut options: CosmosClientOptions) -> CosmosClientOptions {
```

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_client_builder.rs
```diff
@@ -0,0 +1,70 @@
...
+        // Create a default http client
+        let inner_client: Arc<dyn azure_core::http::HttpClient> =
+            azure_core::http::new_http_client();
+
+        let fault_client = FaultClient::new(inner_client, self.rules.clone());
```
analogrelay: If this consumes `self`, then this clone is unnecessary:


```suggestion
        let fault_client = FaultClient::new(inner_client, self.rules);
```

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_client_builder.rs
analogrelay: If a user really wants to inject into two `CosmosClientOptions` with a single `FaultInjectionClientBuilder`, we can make `FaultInjectionClientBuilder` cloneable, putting the decision in the callers hands.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_condition.rs
```diff
@@ -0,0 +1,177 @@
...
+            "BatchItem" => Some(FaultOperationType::BatchItem),
+            "ChangeFeedItem" => Some(FaultOperationType::ChangeFeedItem),
+            "MetadataReadContainer" => Some(FaultOperationType::MetadataReadContainer),
+            "MetadataReadDatabaseAccount" => Some(FaultOperationType::MetadataReadDatabaseAccount),
+            "MetadataQueryPlan" => Some(FaultOperationType::MetadataQueryPlan),
+            "MetadataPartitionKeyRanges" => Some(FaultOperationType::MetadataPartitionKeyRanges),
+            _ => None,
+        }
+    }
+}
```
analogrelay: The [`FromStr`](https://doc.rust-lang.org/stable/std/str/trait.FromStr.html) trait is the standard way to do this. Then we get both `FaultOperationType::from_str("foo")` **and** `let o: FaultOperationType = "foo".parse()` for free.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_condition.rs
```diff
@@ -0,0 +1,177 @@
...
+            FaultOperationType::PatchItem => "PatchItem",
+            FaultOperationType::BatchItem => "BatchItem",
+            FaultOperationType::ChangeFeedItem => "ChangeFeedItem",
+            FaultOperationType::MetadataReadContainer => "MetadataReadContainer",
+            FaultOperationType::MetadataReadDatabaseAccount => "MetadataReadDatabaseAccount",
+            FaultOperationType::MetadataQueryPlan => "MetadataQueryPlan",
+            FaultOperationType::MetadataPartitionKeyRanges => "MetadataPartitionKeyRanges",
+        }
+    }
+}
```
analogrelay: Implement this as an inherent `as_str()` function:


```suggestion
impl FaultOperationType {
    pub fn as_str(&self) -> &'static str {
        match op {
            FaultOperationType::ReadItem => "ReadItem",
            FaultOperationType::QueryItem => "QueryItem",
            FaultOperationType::CreateItem => "CreateItem",
            FaultOperationType::UpsertItem => "UpsertItem",
            FaultOperationType::ReplaceItem => "ReplaceItem",
            FaultOperationType::DeleteItem => "DeleteItem",
            FaultOperationType::PatchItem => "PatchItem",
            FaultOperationType::BatchItem => "BatchItem",
            FaultOperationType::ChangeFeedItem => "ChangeFeedItem",
            FaultOperationType::MetadataReadContainer => "MetadataReadContainer",
            FaultOperationType::MetadataReadDatabaseAccount => "MetadataReadDatabaseAccount",
            FaultOperationType::MetadataQueryPlan => "MetadataQueryPlan",
            FaultOperationType::MetadataPartitionKeyRanges => "MetadataPartitionKeyRanges",
        }
    }
}
```

It's not functionally different from what you have, but more idiomatic. Implementing a `From<T> for &'static str` is generally avoided, though honestly that's more convention/tradition than because of a specific technical reason.

Also, impl `std::fmt::Display` and have it just use `as_str` to write, that also gets us a free impl of `ToString` and a `.to_string()` function.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_condition.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// The container ID to which the fault injection applies.
+    /// By default, the fault injection applies to all containers.
+    pub(crate) container_id: Option<String>,
+}
+
+/// The type of operation to which the fault injection applies.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Default)]
+pub enum FaultOperationType {
+    /// Read items.
+    #[default]
```
analogrelay: Unless we _need_ to implement `Default` here, I wouldn't. Defaulting to ReadItem feels unclear to me.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_result.rs
```diff
@@ -0,0 +1,160 @@
...
+    DatabaseAccountNotFound,
+}
+
+/// Represents a server error to be injected.
+#[derive(Clone, Debug)]
+pub struct FaultInjectionResult {
+    /// The type of server error to inject.
+    pub(crate) error_type: Option<FaultInjectionErrorType>,
+    /// Number of times to inject the error.
+    /// Default is that it will be injected forever.
```
analogrelay:
```suggestion
    /// If `None`, the fault will be injected in every request.
```

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_client_builder.rs
```diff
@@ -0,0 +1,70 @@
...
+        options.fault_injection_enabled = true;
+
+        options
+    }
+
+    /// Adds a fault injection rule to the builder.
+    pub fn with_rule(&mut self, rule: FaultInjectionRule) -> &mut Self {
+        self.rules.push(rule);
+        self
+    }
```
analogrelay: We need to be consistent with our builder patterns. Later on we use `mut self` and here we're using `&mut self`. It annoys me that Rust requires that we choose one or the other (IMO they could probably make method chaining work with `&mut self` but they don't).

`mut self` (consuming, which lends itself to method chaining more appropriately) is the more common pattern here.


```suggestion
    pub fn with_rule(mut self, rule: FaultInjectionRule) -> &mut Self {
        self.rules.push(rule);
        self
    }
```

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_rule.rs
```diff
@@ -0,0 +1,255 @@
...
+use super::fault_injection_condition::FaultInjectionCondition;
+use super::fault_injection_result::FaultInjectionResult;
+
+/// A fault injection rule that defines when and how to inject faults.
+#[derive(Clone, Debug)]
+pub struct FaultInjectionRule {
+    /// The condition under which to inject the fault.
+    pub condition: FaultInjectionCondition,
+    /// The result to inject when the condition is met.
+    pub result: Box<FaultInjectionResult>,
```
analogrelay: Is there a reason this is `Box`ed? It's not wrong necessarily, but it would be good to include a justification.

sdk/cosmos/azure_data_cosmos/src/fault_injection/rule.rs
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    /// Sets the delay before starting the fault injection.
+    pub fn with_start_delay(mut self, start_delay: Duration) -> Self {
+        self.start_delay = start_delay;
+        self
+    }
+
+    /// Sets the total hit limit of the rule.
+    pub fn with_hit_limit(mut self, hit_limit: u32) -> Self {
```
analogrelay, line 120: This seems to be duplicating the `times` on `FaultInjectionResult`. Do we need both? I prefer it here, on the Rule, rather than on the Result. That way the Result is pure data describing a single result, and the repetition is something inherent to the Rule.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_rule.rs
```diff
@@ -0,0 +1,255 @@
...
+        let rule = FaultInjectionRuleBuilder::new("clone-test", create_test_error())
+            .with_hit_limit(3)
+            .build();
+
+        let cloned = rule.clone();
+
+        assert_eq!(cloned.id, rule.id);
+        assert_eq!(cloned.hit_limit, rule.hit_limit);
+        assert_eq!(cloned.duration, rule.duration);
+    }
```
analogrelay: We don't need tests for things like `clone`. Be aware that Copilot sometimes goes a little overboard on edge cases and test coverage. I like to advise it to "AVOID writing tests for trivial or obvious functionality, like derived trait behavior, constructor behavior, etc."

sdk/cosmos/azure_data_cosmos/src/fault_injection/rule.rs
```diff
@@ -0,0 +1,255 @@
...
+            duration: self.duration,
+            start_delay: self.start_delay,
+            hit_limit: self.hit_limit,
+            id: self.id,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
analogrelay, line 140: Honestly, _most_ of the various builder tests in this PR are unnecessary IMO. Copilot likes to generate tests for everything, and I don't mind extra coverage in general, but there comes a point at which it's just review noise and unnecessary test run delays.

sdk/cosmos/azure_data_cosmos/src/fault_injection/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+    FaultInjectionCondition, FaultInjectionConditionBuilder, FaultOperationType,
+};
+
+#[cfg(feature = "fault_injection")]
+pub use fault_injection_result::{
+    FaultInjectionErrorType, FaultInjectionResult, FaultInjectionResultBuilder,
+};
+
+#[cfg(feature = "fault_injection")]
+pub use fault_injection_rule::{FaultInjectionRule, FaultInjectionRuleBuilder};
```
analogrelay: Make this whole module conditional on `#[cfg(feature = "fault_injection")]`. Then you don't need the repeated `cfg` attributes on each item.

analogrelay: In fact, it already is. So these can just be removed.


```suggestion
mod fault_http_client;
mod fault_injection_client_builder;
mod fault_injection_condition;
mod fault_injection_result;
mod fault_injection_rule;

pub use fault_injection_client_builder::FaultInjectionClientBuilder;
pub use fault_injection_condition::{
    FaultInjectionCondition, FaultInjectionConditionBuilder, FaultOperationType,
};
pub use fault_injection_result::{
    FaultInjectionErrorType, FaultInjectionResult, FaultInjectionResultBuilder,
};
pub use fault_injection_rule::{FaultInjectionRule, FaultInjectionRuleBuilder};
```

sdk/cosmos/azure_data_cosmos/src/fault_injection/README.md
analogrelay, line 1: Love that there are detailed docs, but would prefer these be actual doc comments on the `fault_injection` module and related types. You could probably ask Copilot to remove this file and instead focus on expanding out the doc comments on the module and types within. It can be quite common to have pretty detailed high-level documentation on a module, for example: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/index.html

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -159,6 +161,53 @@ impl CosmosRequest {
...
+                Some(FaultOperationType::MetadataReadDatabaseAccount)
+            }
+            (OperationType::QueryPlan, ResourceType::Documents) => {
+                Some(FaultOperationType::MetadataQueryPlan)
+            }
+            (OperationType::ReadFeed, ResourceType::PartitionKeyRanges) => {
+                Some(FaultOperationType::MetadataPartitionKeyRanges)
+            }
+            _ => None,
+        };
```
analogrelay: Make this a function on `FaultOperationType`, like this:

```rust
impl FaultOperationType {
    pub fn from_operation_and_resource(operation_type: OperationType, resource_type: ResourceType) -> Option<Self> {
        // ...
    }
}
```

It's fine for it to be a separate type, I think that's a cleaner API, but the conversion from OperationType/ResourceType into FaultOperationType really belongs with the FaultOperationType, not on `CosmosRequest`.

sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -40,9 +40,10 @@ workspace = true
 [features]
 default = ["hmac_rust"]
 key_auth = [] # Enables support for key-based authentication (Primary Keys and Resource Tokens)
+preview_query_engine = ["serde_json/raw_value"] # Enables support for the PREVIEW external query engine
```
analogrelay: The `preview_query_engine` feature doesn't need to come back right now.

sdk/cosmos/azure_data_cosmos/tests/multi_write_tests/cosmos_multi_write_fault_injection.rs
```diff
@@ -0,0 +1,531 @@
...
+            "408 Request Timeout",
+            FaultInjectionErrorType::Timeout,
+            StatusCode::RequestTimeout,
+        ),
+        (
+            "410 Partition Is Gone",
+            FaultInjectionErrorType::PartitionIsGone,
+            StatusCode::Gone,
+        ),
+    ];
```
analogrelay: Small test organization preference: Make these separate test functions for each case. You could start simple with a parameterized test "implementation" function (NOT marked `#[tokio::test]`), and then define separate `#[tokio::test]` functions that just call the implementation with appropriate parameters. There's no built-in parameterized testing in Rust, nor have we approved a 3p solution in this repo yet.

sdk/cosmos/azure_data_cosmos/tests/multi_write_tests/cosmos_multi_write_fault_injection.rs
```diff
@@ -0,0 +1,531 @@
...
+            let result = fault_container_client.create_item(&pk, &item, None).await;
+
+            let response = result.unwrap();
+            let request_url = response
+                .request()
+                .clone()
+                .into_raw_request()
+                .url()
+                .to_string();
+            println!("Write succeeded via failover, final URL: {}", request_url);
```
analogrelay: nit: Should this be an assert?

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_rule.rs
FabianMeiswinkel: Agreed - and would be good to capture that in copilot instructions.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_rule.rs
```diff
@@ -0,0 +1,255 @@
...
+/// A fault injection rule that defines when and how to inject faults.
+#[derive(Clone, Debug)]
+pub struct FaultInjectionRule {
+    /// The condition under which to inject the fault.
+    pub condition: FaultInjectionCondition,
+    /// The result to inject when the condition is met.
+    pub result: Box<FaultInjectionResult>,
+    /// Duration for which the fault injection is active.
+    /// default is infinite duration.
+    pub duration: Duration,
```
FabianMeiswinkel: NIT: my preference would be to work with absolute start and end - much easier to audit and debug IMO?

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_rule.rs
```diff
@@ -0,0 +1,255 @@
...
+use std::time::Duration;
+
+use super::fault_injection_condition::FaultInjectionCondition;
+use super::fault_injection_result::FaultInjectionResult;
+
+/// A fault injection rule that defines when and how to inject faults.
+#[derive(Clone, Debug)]
+pub struct FaultInjectionRule {
+    /// The condition under which to inject the fault.
+    pub condition: FaultInjectionCondition,
```
FabianMeiswinkel: all the fields should be immutable/read-only/private IMO?

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_rule.rs
FabianMeiswinkel: In java we also allow an existing rule to be disabled - IMO that is pretty important - so, that would be the only mutable field.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_result.rs
```diff
@@ -0,0 +1,160 @@
...
+    WriteForbidden,
+    /// 403-1008 Forbidden from server.
+    DatabaseAccountNotFound,
+}
+
+/// Represents a server error to be injected.
+#[derive(Clone, Debug)]
+pub struct FaultInjectionResult {
+    /// The type of server error to inject.
+    pub(crate) error_type: Option<FaultInjectionErrorType>,
```
FabianMeiswinkel: why pub crate? I think same here - we would want public getters and fields to be immutable?

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_result.rs
```diff
@@ -0,0 +1,160 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+//! Defines fault injection results including server errors.
+
+use std::time::Duration;
+
+/// Represents different server error types that can be injected for fault testing.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum FaultInjectionErrorType {
```
FabianMeiswinkel: QUESTION: This is really me still struggling with Rust - but why did this enum go into this code file? I find it unnecessarily hard to find code because the decision in which file to put it is completely artificial?

I get the option for tests and even private supproted types - but in general either all in mod.rs or strcts/types with narrow scope into separate files? 

Just an honest question - no need to change immediatley - just trying to udnerstand how you guys decide this in your changes.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_http_client.rs
FabianMeiswinkel: In Java we allow isEnabeld to be mutable (only mutable state of a rule - I think it would be good to keep this.

tvaron3: Added a way to explicitily enable and disable a rule. Added a test as well for this scenario.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_client_builder.rs
tvaron3: fixed

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_condition.rs
tvaron3: Implemented the FromStr trait

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_rule.rs
tvaron3: Change the fields on all the public structs to be private. Exposed getters for everything. The rule enablement is the only thing that can be changed.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_result.rs
tvaron3: Change the fields on all the public structs to be private. Exposed getters for everything. The rule enablement is the only thing that can be changed.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_result.rs
tvaron3: Honestly no good reason from my side. I started by putting everything in mod.rs besides the fault_http_client but then it seemed like a lot for a single file so asked copilot to separate it into different files.

sdk/cosmos/azure_data_cosmos/src/fault_injection/rule.rs
tvaron3: Agree, removed the times

sdk/cosmos/azure_data_cosmos/src/fault_injection/rule.rs
tvaron3: Removed all the tests besides the ones for default values. I think these are good for protecting against regressions. Will be more careful with copilot in the future 😄

sdk/cosmos/azure_data_cosmos/tests/multi_write_tests/cosmos_multi_write_fault_injection.rs
tvaron3: yes fixed added the assert

sdk/cosmos/azure_data_cosmos/Cargo.toml
tvaron3: 😆 removed

sdk/cosmos/azure_data_cosmos/tests/multi_write_tests/cosmos_multi_write_fault_injection.rs
tvaron3: Changed to follow this approach.

sdk/cosmos/azure_data_cosmos/src/fault_injection/README.md
tvaron3: I removed the Readme file and added description in mod.rs. The only info lost was how to run fault injection in our tests but I think we can just use the existing tests as examples.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_condition.rs
tvaron3: Agree removed

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_result.rs
tvaron3: removed times option

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_client_builder.rs
tvaron3: Makes sense changed it to consume self instead of borrowing

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_condition.rs
tvaron3: made these changes

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_rule.rs
tvaron3: Removed most tests only left the ones checkign for default values or disallowed values

sdk/cosmos/azure_data_cosmos/src/fault_injection/mod.rs
tvaron3: fixed

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
tvaron3: Followed the recommended implementation

sdk/cosmos/azure_data_cosmos/src/fault_injection/rule.rs
analogrelay: No worries, it takes time to get used to it ;). Copilot _can_ be really useful for generating tests, you just have to get used to it's nuances.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_result.rs
analogrelay: Given that there's a `fault_injection/mod.rs`, I think it would be reasonable and expected to have these enums there instead of in `fault_injection_result.rs` (which should, IMO, be called `fault_injection/result.rs`, since we already have the context.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_rule.rs
tvaron3: Added the ability to disable an existing rule and changed the timings to be an absolute time and end.

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_result.rs
tvaron3: Moving the enums to mod.rs and will rename the files

sdk/cosmos/azure_data_cosmos/src/fault_injection/rule.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    /// Returns the total hit limit of the rule, if set.
+    pub fn hit_limit(&self) -> Option<u32> {
+        self.hit_limit
+    }
+
+    /// Returns the unique identifier for this rule.
+    pub fn id(&self) -> &str {
+        &self.id
+    }
```
analogrelay: I won't block the PR on this, but simple accessors like this are much less common in Rust. Since types are just data, it's fine to mark fields as `pub` rather than keep them private and define simple `pub fn` accessors. There are always exceptions, but in this case I think all of these could just be pub fields instead of having accessors.

Since it's not a universal rule in Rust, I don't mind if we really want to keep it like this and I certainly don't think you should churn the PR just for that, but something to think about for the future.

sdk/cosmos/azure_data_cosmos/src/fault_injection/result.rs
```diff
@@ -5,47 +5,39 @@
...
+
+    /// Returns the delay applied after the fault response.
+    pub fn delay(&self) -> Duration {
+        self.delay
+    }
+
+    /// Returns the probability of injecting the fault (0.0 to 1.0).
+    pub fn probability(&self) -> f32 {
+        self.probability
+    }
```
analogrelay: Same comment here about accessors, these could just be pub fields IMO.

sdk/cosmos/azure_data_cosmos/src/fault_injection/mod.rs
```diff
@@ -1,36 +1,241 @@
...
-#[cfg(feature = "fault_injection")]
-pub use fault_injection_rule::{FaultInjectionRule, FaultInjectionRuleBuilder};
+impl fmt::Display for FaultOperationType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(self.as_str())
+    }
+}
+
+impl FromStr for FaultOperationType {
+    type Err = String;
```
analogrelay: nit: Since there's really only one possible error case here (it's not a valid operation type), you could also use the unit type to return a simple "An error occurred" marker. The user already has the input string, they don't need it in their error message:

```rust
type Err = ();
```

Then, you'd return `Err(())` and you're effectively just saying "The One Possible Error occurred". That would also need to be documented in doc-comments ("If the provided string isn't a valid FaultOperationType, this returns `Err(())`").

Like with my other suggestions, this is totally non-blocking though.

analogrelay (approved):
Just some non-blocking style things. You can merge without them if you'd like.

sdk/cosmos/azure_data_cosmos/src/fault_injection/mod.rs
tvaron3: followed this approach

sdk/cosmos/azure_data_cosmos/src/fault_injection/rule.rs
tvaron3: Made most of the fields pub. I left the AtomicBool for enabling private and the probability

sdk/cosmos/azure_data_cosmos/src/fault_injection/result.rs
tvaron3: Made most of the fields pub. I left the AtomicBool for enabling private and the probability

sdk/cosmos/azure_data_cosmos/src/fault_injection/fault_injection_rule.rs
tvaron3: No reason fixed

sdk/cosmos/azure_data_cosmos/src/fault_injection/condition.rs
```diff
@@ -0,0 +1,83 @@
...
+            container_id: self.container_id,
+        }
+    }
+}
+
+impl Default for FaultInjectionConditionBuilder {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
analogrelay, line 69: nit: This could just be `derive(Default)`on `FaultInjectionConditionBuilder`, the derived implementation is identical to this one. It's harmless to have this here, just pointing that out.

tvaron3: will make this change in next pr.

Issue #3571, opened 2026-01-21
Cosmos: Effective Preferred Regions


simorenoh:
The SDK does not currently support effective preferred regions, and only routes read requests to the hub region due to not having any preferred regions provided by the user.

Effective preferred regions should be populated when a customer with a multi-regional account setup uses the SDK without specifying a set of preferred regions, making it such that all regions in the account are available to be used.

Issue #3535, opened 2026-01-13
Cosmos: Excluded Regions


simorenoh:
Add support for excluded regions in the Rust Cosmos DB SDK to provide users with customizable fine-grained regional routing.

PR #3602, opened 2026-01-23
Cosmos: Excluded Regions + Effective Preferred Regions

simorenoh:
This pull request introduces support for "excluded regions" in the Azure Cosmos DB Rust SDK, allowing users to specify regions that should be avoided for request routing and retries. The changes propagate the `excluded_regions` option from client and request options through request construction, retry policies, and endpoint management, ensuring that excluded regions are respected at every relevant layer. These changes ensure users have fine-grained control over which Azure regions are used or avoided for Cosmos DB operations and retries.

These changes also add **effective preferred regions** to the SDK, changing the previously existing logic for cross-regional availability. While previously the SDK only had cross-regional logic with provided preferred regions, now the SDK will populate all account regions if no preferred regions are provided. Additionally, the SDK will now use preferred regions only as a preferred ordering list and not as an exhaustive list - if an account with regions A, B and C is used with a client that specifies regions [B, C] as preferred regions, the SDK will populate [B, C, A] as effective preferred regions internally. Providing excluded regions is the only way to exclude regions from being used now.

This PR takes care of all point operations and a follow up issue has been made to ensure that queries are also added into the flow of requests: https://github.com/Azure/azure-sdk-for-rust/issues/3603

**Behavior**
Before we send any request to the service, the logic below is applied to choose the endpoint that a request will be routed to.
```
resolve_service_endpoint(request)
│
├─ IF use_preferred_locations == false
│  OR (write operation AND single-master account)
│  └─► Route directly to account_write_locations[index]
│      (Preferred regions and excluded regions are IGNORED)
│
└─ ELSE
   └─► get_applicable_endpoints()
       ├─ Select read or write endpoints based on operation
       ├─ Apply excluded regions (request-level or client-level)
       ├─ Order by: available preferred → unavailable → default
       └─► Route to endpoints[index]
```
Based on the logic above, excluded locations should be respected for *any non-retry scenario* that is not a single write region account doing write requests, or requests that mark `use_preferred_locations` as False internally. That is, in a stable state of sending only successful requests, all requests should have the logic above applied to them.

For error retry scenarios:
404.1002s on single writer accounts will set `use_preferred_locations` to false in order to retry once in the hub region
403.3s will set `use_preferred_locations` to false for the same, since it is only expected in write region failover scenarios for single writer accounts

The most important changes are:

**Feature: Excluded Regions Support**
- Added an `excluded_regions` field to `CosmosClientOptions`, `ItemOptions`, `CosmosRequest`, and `CosmosRequestBuilder`, allowing users to specify regions to avoid for all or individual requests. [[1]](diffhunk://#diff-83b86c8914ca51292cd83f1d6f28aa81c51fc537c1b7fcd8058777b63cd888b0R39) [[2]](diffhunk://#diff-83b86c8914ca51292cd83f1d6f28aa81c51fc537c1b7fcd8058777b63cd888b0R267-R271) [[3]](diffhunk://#diff-633cae4c33eb85be5dd592b71955f38ade584a4fa091ad59a68201649ceb7b72R64) [[4]](diffhunk://#diff-633cae4c33eb85be5dd592b71955f38ade584a4fa091ad59a68201649ceb7b72R180)
- Implemented `.set_excluded_regions()` in `CosmosRequestBuilder`, propagating the option to the built request object. [[1]](diffhunk://#diff-633cae4c33eb85be5dd592b71955f38ade584a4fa091ad59a68201649ceb7b72R228-R235) [[2]](diffhunk://#diff-633cae4c33eb85be5dd592b71955f38ade584a4fa091ad59a68201649ceb7b72R280)

**Request and Client Integration**
- Updated all relevant methods in `ContainerClient` to call `.set_excluded_regions()` when building requests, ensuring per-request excluded regions are respected. [[1]](diffhunk://#diff-de85c06964b9ac3fa3e745ad5f9027fe82839cb5fbcb643de960694f7fd21b3cR293) [[2]](diffhunk://#diff-de85c06964b9ac3fa3e745ad5f9027fe82839cb5fbcb643de960694f7fd21b3cR380) [[3]](diffhunk://#diff-de85c06964b9ac3fa3e745ad5f9027fe82839cb5fbcb643de960694f7fd21b3cR468) [[4]](diffhunk://#diff-de85c06964b9ac3fa3e745ad5f9027fe82839cb5fbcb643de960694f7fd21b3cR522) [[5]](diffhunk://#diff-de85c06964b9ac3fa3e745ad5f9027fe82839cb5fbcb643de960694f7fd21b3cR562) [[6]](diffhunk://#diff-de85c06964b9ac3fa3e745ad5f9027fe82839cb5fbcb643de960694f7fd21b3cR639)
- Modified `CosmosClient` to pass `excluded_regions` from options into the `GlobalEndpointManager`. [[1]](diffhunk://#diff-9e4e40a45e7a837de9541e5cf90dd9bbc6b3e5aea150e9667c22439d4f30872bR80-R82) [[2]](diffhunk://#diff-9e4e40a45e7a837de9541e5cf90dd9bbc6b3e5aea150e9667c22439d4f30872bR143-R145)

**Retry Policy and Endpoint Management**
- Enhanced `ClientRetryPolicy` and its constructor to accept and store `excluded_regions`, and updated retry logic to use them when selecting endpoints. [[1]](diffhunk://#diff-a04e1092dbe47827ef76086e227f7ab87c6199b411e0aeb026073c1c6d9c83a6R73-R75) [[2]](diffhunk://#diff-a04e1092dbe47827ef76086e227f7ab87c6199b411e0aeb026073c1c6d9c83a6L92-R95) [[3]](diffhunk://#diff-a04e1092dbe47827ef76086e227f7ab87c6199b411e0aeb026073c1c6d9c83a6R107) [[4]](diffhunk://#diff-a04e1092dbe47827ef76086e227f7ab87c6199b411e0aeb026073c1c6d9c83a6R134) [[5]](diffhunk://#diff-a04e1092dbe47827ef76086e227f7ab87c6199b411e0aeb026073c1c6d9c83a6L244-R249)
- Updated `BackOffRetryHandler` to pass `excluded_regions` into the retry policy.
- Updated `GlobalEndpointManager::new` and `LocationCache` to accept and use `excluded_regions`.

It is also worth noting some other small changes were made - for instance, the LocationCache's `location_info.write_endpoints` and `location_info.read_endpoints` are no longer computed from the list of preferred regions, since this list is now effectively made dynamic by users who provide excluded regions on a per-request basis. Other small details to the routing logic within the location_cache's `resolve_service_endpoint` logic was also changed to reflect the same.

**Tests**
- Updated all test constructions of `GlobalEndpointManager` and `ClientRetryPolicy` to include the new `excluded_regions` parameter, ensuring test coverage for the new feature. [[1]](diffhunk://#diff-d4d682dab1613501d02fade1f08748c30e0720db55bb44eee625d76b56f23a2eL72-R72) [[2]](diffhunk://#diff-d4d682dab1613501d02fade1f08748c30e0720db55bb44eee625d76b56f23a2eR159) [[3]](diffhunk://#diff-a04e1092dbe47827ef76086e227f7ab87c6199b411e0aeb026073c1c6d9c83a6R548) [[4]](diffhunk://#diff-a04e1092dbe47827ef76086e227f7ab87c6199b411e0aeb026073c1c6d9c83a6R566) [[5]](diffhunk://#diff-a04e1092dbe47827ef76086e227f7ab87c6199b411e0aeb026073c1c6d9c83a6R588-R605) [[6]](diffhunk://#diff-6ade14c1f09908e0e16a72d070dbcadba0135fb5a4a665f9bf2aae716409a5faR286) [[7]](diffhunk://#diff-6ade14c1f09908e0e16a72d070dbcadba0135fb5a4a665f9bf2aae716409a5faR304) [[8]](diffhunk://#diff-6ade14c1f09908e0e16a72d070dbcadba0135fb5a4a665f9bf2aae716409a5faR326) [[9]](diffhunk://#diff-ac8d03394d53a19ce5582c178fa6aee4fa43ab65ead4f7831b3120e5808f3353L191-R191) [[10]](diffhunk://#diff-ac8d03394d53a19ce5582c178fa6aee4fa43ab65ead4f7831b3120e5808f3353R208)

Addresses https://github.com/Azure/azure-sdk-for-rust/issues/3535 and https://github.com/Azure/azure-sdk-for-rust/issues/3571


sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
tvaron3, line 627: Should query exclude regions support be added in this pr as well? If not might be good to make a gh issue to follow up

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
```diff
@@ -182,10 +189,21 @@ impl LocationCache {
         write_locations: Vec<AccountRegion>,
         read_locations: Vec<AccountRegion>,
     ) -> Result<(), &'static str> {
+        // Populate effective preferred locations if none are set
```
tvaron3: I think this logic should be slightly different. We should always populate all the regions and the preferred regions are only an ordering. Based off of the next gen sdk design principles https://github.com/Azure/azure-sdk-for-rust/blob/4edae73d6cccee78556522af75464729b83ab399/sdk/cosmos/azure_data_cosmos_native/docs/next_generation_sdks_design_principles.md?plain=1#L98

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
```diff
@@ -88,7 +88,10 @@ impl BackOffRetryHandler {
                 self.global_endpoint_manager.clone(),
             ))
         } else {
-            RetryPolicy::Client(ClientRetryPolicy::new(self.global_endpoint_manager.clone()))
+            RetryPolicy::Client(ClientRetryPolicy::new(
```
tvaron3, line 93: We should also hook up excluded regions for metadata requests in the lifecycle of a document operation. Later on it might be harder to hook up.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -241,7 +249,7 @@ impl ClientRetryPolicy {
         if self.can_use_multiple_write_locations {
             let endpoints = self
                 .global_endpoint_manager
-                .applicable_endpoints(self.operation_type.unwrap());
+                .applicable_endpoints(self.operation_type.unwrap(), self.excluded_regions.as_ref());
```
tvaron3, line 240: Not 100% sure on the session retry policy but that is one place where excluded locations might be violated because I believe the expected behavior is that it retries on the hub. Just wanted to bring it up

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
simorenoh: right good catch - regions should only be excluded with excluded regions, preferred regions merely sets the ordering of preference. Made the updates.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
simorenoh: This piece in particular is for MRW flows within the session retry policy, which I believe just goes through available write regions. In Python at least it seems as though excluded regions plays a role in that case, and we also document as much in our doc: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/cosmos/azure-cosmos/docs/ErrorCodesAndRetries.md

Single writer flow is right below and will always retry on the hub region once.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
simorenoh: I'll make a follow-up for this since it doesn't seem like queries are plugged into the retry_context as of now. Follow up issue: https://github.com/Azure/azure-sdk-for-rust/issues/3603

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
simorenoh: You're not wrong - but these internal metadata requests don't yet exist for the different item request flows from what I saw. We don't yet create partition key range wrappers for partition-specific logic, we don't have cross-partition queries that might require fetching partition ranges, we don't fetch query plans either, and we don't yet have a session token cache that might need to query ranges based on the partition key of the operation.

Taking a second look, this logic should already be there for the most part, and will need to be pieced together afterward once we have the internal requests - when we call resolve_service_endpoint within the MetadataRequestRetryPolicy, the request object being passed in will have excluded locations in it if they are populated, and those are the ones that will be used within the location cache regardless. The logic depends on the passing of the request excluded regions to the internal request at this point.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
tvaron3: Sorry, my worry is this part below. If an account has R1, R2, and R3 and R1 is the hub and R1 is excluded, then we should not retry on hub for single write accounts. I think this would still happen here because `retry_request_on_preferred_locations` is false. The mwr scenario should be fine because I think excluded locations will be applied there because `retry_request_on_preferred_locations` is true. 
```
self.retry_context = Some(RetryContext {
                retry_location_index: 0,
                retry_request_on_preferred_locations: false,
                route_to_hub: false,
            });
```

At least this is my understand atm.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
tvaron3: perfect thanks for the explanation. Makes sense.

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
```diff
@@ -871,4 +954,178 @@ mod tests {
...
+        let cache = create_custom_test_location_cache(Some(pref_regions), excl_regions);
+
+        let builder = CosmosRequest::builder(
+            OperationType::Read,
+            ResourceLink::root(ResourceType::Documents),
+        );
+
+        let cosmos_request = builder
+            .clone()
+            .set_excluded_regions(Some(vec!["Location 3".to_string()]))
```
tvaron3: Could we also test this with an empty string so that it effectively removes client exclude regions

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
tvaron3: Ahh i see in the description that this is intentional. I am also not so sure that this should be treated same as 403.3 because 403.3 is a change in the account topology so the client settings could be stale. A 404:1002 is just a consistency issue and a customer might prefer for reads to fail. Don't have strong opinion either way, but I think we should document this in the public api.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -263,6 +264,11 @@ pub struct ItemOptions<'a> {
     ///
     /// Custom headers will not override headers that are already set by the SDK.
     pub custom_headers: HashMap<HeaderName, HeaderValue>,
+    /// Regions to be skipped from regional routing preferences. The regions in this list are specified as the names of the Azure Cosmos locations like, 'West US', 'East US' and so on.
+    /// If all preferred regions were excluded, the primary/hub region will be used to route requests.
```
tvaron3: Nit: Now that all regions are always used I think it should be the following
```suggestion
    /// If all regions were excluded, the primary/hub region will be used to route requests.
```

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
simorenoh: line 1095 below has that test (empty list, discard excluded client regions)

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -290,6 +290,7 @@ impl ContainerClient {
             .request_headers(&options)
             .json(&item)
             .partition_key(partition_key.into())
+            .set_excluded_regions(options.excluded_regions)
```
kundadebdatta: Naming: I think we agreed not to use `set`/ `get` in function names. Let's just call this `excluded_regions`.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -36,6 +36,7 @@ pub struct CosmosClientOptions {
     pub application_name: Option<String>,
     pub application_region: Option<String>,
     pub application_preferred_regions: Vec<Cow<'static, str>>,
+    pub excluded_regions: Vec<String>,
```
kundadebdatta: I am assuming this list is going to hold the set of pre-defined regions. If that's the case, we could use `Vec<Cow<'static, str>>` just like we do for `application_preferred_regions`.

In Rust, `Vec<String>` holds a list of owned, heap-allocated strings, while `Vec<Cow<'static, str>>` holds a list that can contain either owned strings (`Strin`g) or borrowed string literals (`&'static str`). The choice depends on whether the data must always be owned and potentially mutated, or if static, borrowed data can be leveraged for performance. 

[See this](https://doc.rust-lang.org/std/borrow/enum.Cow.html) for more details.

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
```diff
@@ -183,10 +190,22 @@ impl LocationCache {
         write_locations: Vec<AccountRegion>,
         read_locations: Vec<AccountRegion>,
     ) -> Result<(), &'static str> {
+        // Build effective preferred locations: preferred regions first, then remaining account regions
+        let mut effective_preferred_locations = self.locations_info.preferred_locations.clone();
+        for location in &read_locations {
+            let loc_cow = Cow::Owned(location.name.clone());
+            if !effective_preferred_locations
+                .iter()
+                .any(|p| p.eq_ignore_ascii_case(&location.name))
```
Pilchie: For accounts with a large number of regions this might be slow (n^2), right?  Are we worried about that?  

Would something like this work (off the top of my head, not type-checked) instead?
``` rust
let mut effective_preferred_locations = self.locations_info.preferred_locations.clone();
effective_preferred_locations
    .extend(read_locations.map(|loc| Cow::Owned(loc.name.clone())
    .into_iter()
    .distinct(|loc1, loc2| loc1.eq_ignore_ascii_case(loc2))
    .collect();
```

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
```diff
@@ -486,11 +504,24 @@ impl LocationCache {
         endpoints_by_location: &HashMap<String, Url>,
         request: RequestOperation,
         default_endpoint: &Url,
+        request_excluded_regions: Option<&Vec<String>>,
     ) -> Vec<Url> {
         let mut endpoints = Vec::new();
         let mut unavailable_endpoints = Vec::new();
+        let mut effective_preferred_locations = self.locations_info.preferred_locations.clone();
+        // Remove excluded regions from effective preferred locations
```
Pilchie, line 515: I thought our decision was to keep them, but put them last in the list? @FabianMeiswinkel ?

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
simorenoh: Above does not work unfortunately - there's two borrowings happening within the same context here (using `.extend` and using `.filter`) but we can just use a HashSet to increase speed to O(n) with the tradeoff being holding a couple extra strings for a little bit. Went ahead and did that.

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
simorenoh: Out of curiosity, what function would excluded regions serve in that case? This is the usual behavior you would get from only specifying preferred regions.

Currently (with these changes) we updated the preferred region logic so that regions not in the list are appended to be used as well (just less preferred). Effective preferred regions is now always applied with or without preferred regions.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -263,6 +264,11 @@ pub struct ItemOptions<'a> {
     ///
     /// Custom headers will not override headers that are already set by the SDK.
     pub custom_headers: HashMap<HeaderName, HeaderValue>,
+    /// Regions to be skipped from regional routing preferences. The regions in this list are specified as the names of the Azure Cosmos locations like, 'West US', 'East US' and so on.
+    /// If all regions were excluded, the primary/hub region will be used to route requests.
+    /// If None is provided, client-level excluded regions will be used.
+    /// If an empty vector is provided, no regions will be excluded for this request.
+    pub excluded_regions: Option<Vec<String>>,
```
Pilchie:
```suggestion
    pub excluded_regions: Option<Vec<Cow<'static, str>>,
```

This should also be `Cow<'static, str>`, right?

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -67,6 +67,9 @@ pub struct ClientRetryPolicy {
     /// Context information for routing the next retry attempt to a specific location
     retry_context: Option<RetryContext>,
 
+    /// Regions excluded from routing for the current request
+    excluded_regions: Option<Vec<String>>,
```
Pilchie:
```suggestion
    excluded_regions: Option<Vec<Cow<'static, str>>,
```

And flow through here too?

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -86,7 +89,10 @@ impl ClientRetryPolicy {
     ///
     /// # Returns
     /// A new `ClientRetryPolicy` instance configured with default retry limits and throttling behavior
-    pub fn new(global_endpoint_manager: Arc<GlobalEndpointManager>) -> Self {
+    pub fn new(
+        global_endpoint_manager: Arc<GlobalEndpointManager>,
+        excluded_regions: Option<Vec<String>>,
```
Pilchie:
```suggestion
        excluded_regions: Option<Vec<Cow<'static, str>>,
```

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -61,6 +61,7 @@ pub struct CosmosRequest {
     pub query_string: Option<String>,
     pub continuation: Option<String>,
     pub entity_id: Option<String>,
+    pub excluded_regions: Option<Vec<String>>,
```
Pilchie:
```suggestion
    pub excluded_regions: Option<Vec<Cow<'static, str>>,
```

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -175,6 +177,7 @@ pub struct CosmosRequestBuilder {
     authorization_token_type: AuthorizationTokenType,
     continuation: Option<String>,
     entity_id: Option<String>,
+    excluded_regions: Option<Vec<String>>,
```
Pilchie:
```suggestion
    excluded_regions: Option<Vec<Cow<'static, str>>,
```

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -221,6 +225,14 @@ impl CosmosRequestBuilder {
         self
     }
 
+    pub fn excluded_regions(mut self, excluded_regions: Option<Vec<String>>) -> Self {
```
Pilchie:
```suggestion
    pub fn excluded_regions(mut self, excluded_regions: Option<Vec<Cow<'static, str>>) -> Self {
```

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
Pilchie: Two things:
1. I believe there are 3 states for a region: preferred, neutral, excluded, which would be in that order, so it's not *quite* the same as just having preferred at the front the list.
2. Adding something that is in preferred_locations to excluded_locations would move it from the front of the list to the back. So, still possible to try if necessary for availability, but as a last ditch effort. Especially since we don't have preferred_locations on a per-request basis, right?

Anyway, I'm totally happy to be overruled by @FabianMeiswinkel here if just removing them entirely is what we should do, I just thought I remembered a conversation where we didn't want to cause an availability issue because all other regions failed.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
FabianMeiswinkel: Let us please discuss in one of the syncups. My 2 cents - we should change to also even prefer rerty in hub region in Multi-Writer scenarios. More and more customers will attempt to write into hub-region (cheaper RU-wise) - so, reads when doing x-region retry on 404/1002 very well could prefer hub region (because chances are higher there)

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
FabianMeiswinkel: I would even go so far to ask why we do not use a newtype for region in public surface area?

FabianMeiswinkel: Blocking mostly on getting this cleared up - other changes look fine to me for this PR (the 404/1002 question discussed can be addressed in separate PR)

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
simorenoh: Post discussion, keeping it as-is with special consideration for hub region in different scenarios.
Verify cross-SDK behavior.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -36,6 +37,7 @@ pub struct CosmosClientOptions {
     pub application_name: Option<String>,
     pub application_region: Option<RegionName>,
     pub application_preferred_regions: Vec<RegionName>,
+    pub excluded_regions: Vec<RegionName>,
```
Pilchie, line 39: It seems odd that this is `Vec<RegionName>`, while `ItemOptions` uses `Vec<Cow<'static, str>>`.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
```diff
@@ -197,11 +201,15 @@ impl GlobalEndpointManager {
     ///
     /// # Returns
     /// A vector of applicable endpoint URLs
-    pub fn applicable_endpoints(&self, operation_type: OperationType) -> Vec<Url> {
+    pub fn applicable_endpoints(
+        &self,
+        operation_type: OperationType,
+        excluded_regions: Option<&Vec<Cow<'static, str>>>,
```
Pilchie: `::new` takes `Vec<RegionName>`, but this takes `Vec<Cow<'static, str>>`.  Should we be consistent here?

Pilchie (approved):
A couple of nits on consistency around use of `RegionName` vs `Cow<'static, str>`, but otherwise LGTM.

Issue #3570, opened 2026-01-20
Cosmos: Expand `impl Into<PartitionKey>` Examples and Doc Comments


analogrelay:
This issue tracks adding more concrete usage examples and additional tests to help users understand all the ways to specify partition keys with the `impl Into<PartitionKey>` API pattern.

- Add new tests to `partition_key.rs` demonstrating how to specify partition keys using various types (e.g., &str, String, () for empty, Vec types, etc) and validate these code paths work as intended.
- Update relevant doc comments in both the SDK API surface and models to include these usage patterns for developers as reference.
- Ensure documentation links to or references the new exhaustive tests so users can find working code examples easily.

Goal: Bring confidence that users can specify partition keys in all supported ways and improve discoverability with robust examples and test coverage.

analogrelay:
We're good with the examples and tests now.

PR #3606, opened 2026-01-24
docs(cosmos): add exhaustive partition key examples and tests (#3570)

shripad-jyothinath:


Pilchie:
@analogrelay can you take a look at this when you get back?

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
```diff
@@ -22,6 +22,14 @@ use crate::constants;
 ///     "SELECT * FROM c",
 ///     "a single string partition key",
 ///     None).unwrap();
+/// container_client.query_items::<serde_json::Value> (
+///     "SELECT * FROM c",
+///     String::from("An owned partition key"),
+///     None).unwrap();  
+/// container_client.query_items::<serde_json::Value> (
+///     "SELECT * FROM c",
+///     (), // Using () to signal Cross-Partition query
```
FabianMeiswinkel: Cross-partition query currently is not supported yet. Samples should also include Hierarchical partition keys (including partial ones) once cross-partition queries are enabled.

FabianMeiswinkel (changes requested):
Worth changing the comments/asmples - but let's wait until HPK and cross partition query support is there.

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
analogrelay: > Cross-partition query currently is not supported yet

The SDK has previously supported the limited Gateway-mode cross-partition queries (SELECT and WHERE only) and we did use `()` to signal that. However, you're right that we've temporarily removed that as we build proper cross-partition query support.

shripad-jyothinath: Thanks for the clarification! I wasn't aware cross-partition support was temporarily disabled.

I have pushed a commit that removes the () / cross-partition examples from the documentation.

I kept the examples for String, &str, and numbers (and the corresponding ownership test), as those address the ownership confusion highlighted in the original issue and are currently supported.

Ready for re-review!

analogrelay:
@shripad-jyothinath apologies for delays here, but we're ready to merge this. I can't retarget your PR to `release/azure_data_cosmos-previews` though, which is the branch we're currently shipping out of. Could you rebase it and change the base branch on the PR itself? Once that's done I think we can merge it! Please note I made a commit to resolve a conflict, so you'll want to pull the latest to your local branch.

PR #3608, opened 2026-01-26
Add custom agent to analyze source

heaths:
Initial custom agent to analyze source. For now it's only spelling and I tested it against #3577 to add a bunch of words from the ARM TypeSpec to the existing Key Vault custom dictionary.

### DIFF
--- a/.github/agents/Analyze.agent.md
+++ b/.github/agents/Analyze.agent.md
@@ -29,7 +29,7 @@ Spelling errors should be fixed in public APIs and documentation.
 - Run `npx cspell lint -c .vscode/cspell.json sdk/<service>/**` to check spelling.
 - Show a summary of unique, case-insensitive words.
 - Prompt the user which words should be changed and to what they should be changed.
-  Those changes should be made to all files under `sdk/<service>/` and in #changes.
+  Those changes should be made to all files under `sdk/<service>/` and in the #changes context.
 - Remaining words should be added to `sdk/<service>/.dict.txt` as a sorted list of lowercase unique words.
 - If `sdk/<service>/.dict.txt` does not exist, create it as a plain text file and add an entry to `.vscode/cspell.json` like for `sdk/keyvault` e.g.,
 
@@ -39,7 +39,7 @@ Spelling errors should be fixed in public APIs and documentation.
         "name": "keyvault",
         "path": "../sdk/keyvault/.dict.txt",
         "noSuggest": true
-      },
+      }
     ],
     "overrides": [
       {


.github/agents/Analyze.agent.md
```diff
@@ -0,0 +1,56 @@
...
+Do not modify code.
+
+The following instructions are relative to the repository root unless stated otherwise.
+Understand the directory structure from `AGENTS.md` in the root.
+
+## Spelling
+
+Spelling errors should be fixed in public APIs and documentation.
+
+- Run `npx cspell lint -c .vscode/cspell.json sdk/<service>/**` to check spelling.
```
benbp, line 29: We are also implementing these types of commands with additional logic into the azsdk mcp/cli. See [spelling](https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/PackageCheckTool.cs#L76) for example. This way our agent instructions can be simplified to `call 'azsdk package validate cspell' and implement recommended fixes` and we can bake some of the extra business logic into code. The mcp server then gets auto-installed into the gh copilot agent context (see https://github.com/Azure/azure-sdk-for-net/blob/main/.github/workflows/copilot-setup-steps.yml#L40-L43).

heaths: Why would we want to bake repo-specific code into a common location as oppose to having azsdk use well-known agent names in the repos to handle language-specific tasks? When I originally set up cspell for devs and PRs in .NET (later abstracted for all languages like we have now), I didn't foresee just how often a shared file gets updated for other languages so here, for Rust, I expanded on that. If we made all the other languages consistent (or even improve on what I have here), I could see centralized code; however, right now, testing that locally in my repo, getting the tool updated, and getting that into partners' hands is a much longer dev loop how you describe than what I described.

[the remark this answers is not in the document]
heaths: @weshaggard @danieljurek do we lock `cspell` down in other repos? Can't say I've seen that, nor am I sure I'd want a `package.json` in the root repo. That can have downstream effects for other tools and workspace detection e.g., in VSCode to activate extensions. I could instead lock it on the command line if we'd prefer.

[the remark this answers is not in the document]
danieljurek: We do lock spelling in `eng/common/spelling` and invoke it using [Invoke-Cspell.ps1](https://github.com/Azure/azure-sdk-for-rust/blob/main/eng/common/spelling/Invoke-Cspell.ps1) 

In the case of finding spelling errors, a simple command in azsdk-cli that finds changed files and pipes the list of paths into Invoke-Cspell.ps1 is probably enough.

_Semi-related:_

I had some success yesterday setting a baseline of spelling errors in a REST API specs repo by instructing an agent to the effect of: 

> Run spell check according to `spelling-all.yaml`, filter for "Unknown word" and then set exceptions in the service's cspell.yaml according to instructions in https://aka.ms/ci-fix#spell-check.

You might have use something like that and say "if there are spelling errors in change parts of files in this PR, just correct them in source code and re-run spell check and build/analyze."

danieljurek: For repos that sync `eng/common` the code and calling interface are consistent for spell checking (modulo specifying the cspell config file)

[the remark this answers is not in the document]
heaths: We lay out service crate dictionaries differently, though. And I see no reason to code that if a skill could do it. See what I did in my own repo that works well: https://github.com/heaths/azsdk-samples-mcp/pull/23

I'm fine with relying on the script to *show* misspellings, but dealing with them might be a job better left to such a skill. We could have a skill like the one I wrote invoke the script instead of running `npx cspell` directly. How does that sound?

In general, it would be interesting to see how many processes we could leverage skills/instructions/prompts where we don't need 100% reproducibility or perfection, like misspellings (as long as they get fixed/exempted).

heaths: Spell checking, yes. Misspelling mitigation, no. See my comment above for an idea.

heaths:
I've been testing some ideas in https://github.com/heaths/azsdk-samples-mcp and need to update this PR e.g., use `SKILLS.md` instead.

Issue #3617, opened 2026-01-27
Bump versions before GA


heaths:
We should consider bumping dependencies' versions before GA. @chlowell is already working on `reqwest` but running into issues because of changes to their features. @jhendrixMSFT ran into an issue in the emitter repo because the `Cargo.lock` file wasn't checked in and `time` got updated to require an MSRV of 1.88, which was released on June 26th, 2025, so well within our N-6mo window equivalent to Go's N-1 schedule. That would also allow us to finally use let-chains e.g.,

```rust
if let Some(s) = m.s && s == "foo" {
    todo!();
}
```

PR #3620, opened 2026-01-27
Upgrade to reqwest 0.13.1

chlowell:
This entailed recording tests again because the new reqwest doesn't put a space between comma-separated header values. Part of #3617

### DIFF
--- a/sdk/core/azure_core_opentelemetry/assets.json
+++ b/sdk/core/azure_core_opentelemetry/assets.json
@@ -1,6 +1,6 @@
 {
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "rust",
-  "Tag": "rust/azure_core_opentelemetry_341a93c45d",
+  "Tag": "rust/azure_core_opentelemetry_0ad28807f5",
   "TagPrefix": "rust/azure_core_opentelemetry"
 }
--- a/sdk/eventhubs/assets.json
+++ b/sdk/eventhubs/assets.json
@@ -2,5 +2,5 @@
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "rust",
   "TagPrefix": "rust/eventhubs",
-  "Tag": "rust/eventhubs_e100d23e41"
+  "Tag": "rust/eventhubs_ac0003d6c8"
 }
--- a/sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/assets.json
+++ b/sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/assets.json
@@ -1,0 +1,6 @@
+{
+  "AssetsRepo": "Azure/azure-sdk-assets",
+  "AssetsRepoPrefixPath": "rust",
+  "TagPrefix": "rust/eventhubs/azure_messaging_eventhubs_checkpointstore_blob",
+  "Tag": "rust/eventhubs/azure_messaging_eventhubs_checkpointstore_blob_9fb394f73c"
+}
--- a/sdk/keyvault/assets.json
+++ b/sdk/keyvault/assets.json
@@ -2,5 +2,5 @@
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "rust",
   "TagPrefix": "rust/keyvault",
-  "Tag": "rust/keyvault_11d39772f0"
+  "Tag": "rust/keyvault_b99e0fab7e"
 }
--- a/sdk/storage/azure_storage_blob/assets.json
+++ b/sdk/storage/azure_storage_blob/assets.json
@@ -1,6 +1,6 @@
 {
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "rust",
-  "Tag": "rust/azure_storage_blob_a7d18a7ad9",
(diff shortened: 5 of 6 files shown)


Cargo.lock
```diff
@@ -2438,6 +2478,20 @@ dependencies = [
  "web-sys",
 ]
 
+[[package]]
+name = "ring"
```
chlowell, line 2486: This is from the optional `ring` feature of `rustls-webpki`, which we don't select

heaths: It comes in from `hyper-rustls`:

```text
ring <- rustls-webpki <- rustls
	<- tokio-rustls <- hyper-rustls
	<- hyper-rustls <- reqwest (added dependency)
```

If we can elide that, that would be preferred. Otherwise we need to re-file an exemption every 3 months attesting we don't actually use it. I'd prefer not to go back to that. /cc @RickWinter

heaths: Looking at <https://docs.rs/crate/reqwest/0.13.1/source/Cargo.toml> and comparing with our `Cargo.toml` files, I'm not entirely clear why it was pulled in. `http2` support that Cosmos adds only pulls in `hyper-rustls` if `rustls` was enabled anyway.

Looking at this diff, only `reqwest` and `tower-http` were updated but it doesn't seem `tower-http` is the culprit from manually reversing the tree. Even `cargo tree -i ring --all-features` shows no results, but the code scanning tool only looks at the text.

heaths: `hyper-rustls` is an optional feature and from our feature selections across all `Cargo.toml` files, we don't appear to enable it. But it's added to `reqwest`'s dependencies. That's what starts the chain. It's an optional dependency, and `stream` nor any of the other features we enable like `http2` support in Cosmos should necessitate it.

heaths: Diffing 0.13.0 and 0.13.1 `Cargo.toml` files:

```diff
59c59
< __rustls-aws-lc-rs = ["hyper-rustls?/aws-lc-rs", "tokio-rustls?/aws-lc-rs", "rustls?/aws-lc-rs", "quinn?/aws-lc-rs"]
---
> __rustls-aws-lc-rs = ["hyper-rustls?/aws-lc-rs", "tokio-rustls?/aws-lc-rs", "rustls?/aws-lc-rs", "quinn?/rustls-aws-lc-rs"]
71c71
< http2 = ["h2", "hyper/http2", "hyper-util/http2", "hyper-rustls?/http2"]
---
> http2 = ["dep:h2", "hyper/http2", "hyper-util/http2", "hyper-rustls?/http2"]
115c115
< version = "0.13.0"
---
> version = "0.13.1"
207c207
< features = ["rustls", "runtime-tokio"]
---
> features = ["runtime-tokio"]
```

I remember you originally had 0.13.0 when you were asking me about this, which had `rustls` and was removed in 0.13.1. I wonder if the `Cargo.lock` is just stale. I'll upgrade directly to 0.13.1 locally and see if that makes a difference.

sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -28,7 +28,7 @@ uuid.workspace = true
 azure_identity.workspace = true
 azure_core_test = { workspace = true, features = ["tracing"] }
 clap.workspace = true
-reqwest.workspace = true
+reqwest = { workspace = true, features = ["http2"] }
```
heaths, line 31: This is the cause. I did a clean upgrade to 0.13.1 and didn't get `ring`. I even tried 0.13.0 just to see if that was the issue since the diff between 0.13.0..0.13.1 removed `rustls` as a feature dependency. Neither time was `ring` pulled in.

Then I remember you also made this change and when I added that `ring` was pulled in.

So, now for *how* it gets pulled in.

1. `ring` is only a dependency of `rustls-webpki` version 0.103.9 in `Cargo.lock`.
2. `rustls-webpki` has `ring` as an optional non-default feature: <https://docs.rs/crate/rustls-webpki/0.103.9/source/Cargo.toml#49>
3. `rustls-webpki` is only a dependency of `rustls` version 0.23.36 in `Cargo.lock`.
4. `rustls` has `rustls-webpki` as an optional feature with only `alloc` turned on by default: <https://docs.rs/crate/rustls/0.23.36/source/Cargo.toml#54-58>
5. `rustls` enables `ring` as an optional feature: <https://docs.rs/crate/rustls/0.23.36/source/Cargo.toml#124>

Any path I take has ring as an optional weak dependency, so I looked and found this: https://github.com/rust-lang/cargo/issues/10801

If Cosmos really needs `http2` support right now, we're going to have to escalate the code analysis check to do a more thorough analysis if it finds a simple string hit e.g., check if `cargo tree -i` returns 0 (used) or not (unused).

heaths: @analogrelay @FabianMeiswinkel do you need http2 support currently, or can we punt this down the road and escalate in parallel?

heaths: If nothing else, I would pull this out into a separate issue/PR since it's new. It wasn't enabled before, though I know Cosmos was interested.

heaths: Cosmos said they would be adding HTTP2 support soon enough it really doesn't make much sense to not just address this now. I'll work with the code scanning team internally on a more permanent solution. *We* in this repo have the right stuff in place via `cargo deny` to make sure we never actually use `ring`.

PR #3622, opened 2026-01-28
Cosmos Response

tvaron3:
## Description
Wrapped all the operations in a cosmos response except queries because those use paging. The cosmos response exposes public api for getting the following information. Main motivation for exposing the cosmos response is to get the endpoint where a request went to be able to validate retry policies with fault injection. 
- the last request used by the operation
   - later this will change with the cosmos diagnostics
- etag
- session token 
- any response header
- status code 
- ru charge

## Testing 
Validated that responses include all the relevant information for document operations. Added helper methods for getting the global endpoint and the hub region endpoint from the ci pipeline.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/tests/multi_write_tests/cosmos_multi_write.rs
+++ b/sdk/cosmos/azure_data_cosmos/tests/multi_write_tests/cosmos_multi_write.rs
@@ -9,14 +9,14 @@ use std::borrow::Cow;
 use std::error::Error;
 use std::sync::{Arc, Mutex};
 
+use azure_data_cosmos::regions::RegionName;
 use azure_data_cosmos::{
     clients::DatabaseClient,
     models::{ContainerProperties, ThroughputProperties},
     CosmosClientOptions, CreateContainerOptions,
 };
-use tracing_subscriber::layer::SubscriberExt;
-use azure_data_cosmos::regions::RegionName;
 use framework::{TestClient, HUB_REGION, SATELLITE_REGION};
+use tracing_subscriber::layer::SubscriberExt;
 
 /// A simple layer that captures log messages into a shared buffer
 struct CaptureLayer {
@@ -192,7 +192,8 @@ pub async fn multi_write_preferred_locations() -> Result<(), Box<dyn Error>> {
     // write to satellite region
     TestClient::run_with_unique_db(
         async |_, db_client| {
-            create_container_and_write_item(db_client, CONTAINER_ID, SATELLITE_REGION.as_str()).await
+            create_container_and_write_item(db_client, CONTAINER_ID, SATELLITE_REGION.as_str())
+                .await
         },
         Some(options_with_preferred_locations(vec![
             SATELLITE_REGION.into(),


sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// Returns `Some(&str)` if the header exists,
+    /// or `None` if the header doesn't exist.
+    pub fn get_optional_header_str(&self, name: &HeaderName) -> Option<&str> {
+        self.response.headers().get_optional_str(name)
+    }
+
+    /// Returns the final endpoint used to fulfill the operation.
+    pub fn endpoint(&self) -> Url {
+        self.request.clone().into_raw_request().url().clone()
+    }
```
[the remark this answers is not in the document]
tvaron3: fixed

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
```diff
@@ -0,0 +1,99 @@
...
+        self.response.headers().get_str(name)
+    }
+
+    /// Gets an optional header value as a string by name.
+    ///
+    /// Returns `Some(&str)` if the header exists,
+    /// or `None` if the header doesn't exist.
+    pub fn get_optional_header_str(&self, name: &HeaderName) -> Option<&str> {
+        self.response.headers().get_optional_str(name)
+    }
```
simorenoh: Wondering why we need both of these - I'd think just the optional would be enough? Do let me know if im misunderstanding the use of having both. Seems like we only use the optional one below as well.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -90,13 +90,14 @@ impl ContainerClient {
...
         options: Option<ReadContainerOptions<'_>>,
-    ) -> azure_core::Result<Response<ContainerProperties>> {
+    ) -> azure_core::Result<CosmosResponse<ContainerProperties>> {
         let options = options.unwrap_or_default();
         let cosmos_request =
             CosmosRequest::builder(OperationType::Read, self.link.clone()).build()?;
         self.container_connection
             .send(cosmos_request, options.method_options.context)
             .await
+            .map(|(response, endpoint)| CosmosResponse::new(response, endpoint))
```
simorenoh: nit: it says you're passing the endpoint based on the variable naming but you're actually passing the request right?

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
tvaron3: I just copied the azure core response return methods, but since its not used I think I can just remove now. If we need it in future we can always add it.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
tvaron3: yes changed it. good catch

sdk/cosmos/azure_data_cosmos/tests/framework/test_client.rs
```diff
@@ -102,6 +105,52 @@ fn get_shared_database_id() -> &'static str {
...
+    // Parse the connection string to get the account endpoint
+    let connection_string: ConnectionString = connection_string_env
+        .parse()
+        .expect("Failed to parse connection string");
+
+    let account_endpoint = connection_string.account_endpoint.trim_end_matches('/');
+
+    // Parse the URL to extract the host and insert the hub region
+    // Expected format: https://accountname.documents.azure.com:443
+    // Target format: accountname-region.documents.azure.com (host only, no scheme/port)
```
Pilchie: This doesn't add the region though, right?

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -90,13 +90,14 @@ impl ContainerClient {
...
         options: Option<ReadContainerOptions<'_>>,
-    ) -> azure_core::Result<Response<ContainerProperties>> {
+    ) -> azure_core::Result<CosmosResponse<ContainerProperties>> {
         let options = options.unwrap_or_default();
         let cosmos_request =
             CosmosRequest::builder(OperationType::Read, self.link.clone()).build()?;
         self.container_connection
             .send(cosmos_request, options.method_options.context)
             .await
+            .map(|(response, request)| CosmosResponse::new(response, request))
```
Pilchie: Why not change `GatewayPipeline::send` and `ContainerConnection::send` to return `CosmosResponse` so that you don't have to `.map` everywhere?

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
```diff
@@ -0,0 +1,91 @@
...
+/// A response from a Cosmos DB operation.
+///
+/// This wraps the underlying Azure Core typed response and provides convenient access
+/// to headers, status code, the original request, and Cosmos-specific response metadata.
+#[derive(Debug)]
+pub struct CosmosResponse<T> {
+    /// The underlying typed HTTP response.
+    response: Response<T>,
+    /// The final endpoint used to fulfill the operation.
+    endpoint: Url,
```
Pilchie: If you only keep the `endpoint`, why pass the whole request in? Do we expect to need more in the future?

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
```diff
@@ -0,0 +1,91 @@
...
+    /// Gets an optional header value as a string by name.
+    ///
+    /// Returns `Some(&str)` if the header exists,
+    /// or `None` if the header doesn't exist.
+    pub fn get_optional_header_str(&self, name: &HeaderName) -> Option<&str> {
+        self.response.headers().get_optional_str(name)
+    }
+
+    /// Returns the final endpoint used to fulfill the operation.
+    pub fn endpoint(&self) -> Url {
```
Pilchie:
```suggestion
    pub fn endpoint(&self) -> &Url {
```

Maybe let the caller decide if they need to clone or not

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
```diff
@@ -0,0 +1,91 @@
...
+use url::Url;
+
+/// A response from a Cosmos DB operation.
+///
+/// This wraps the underlying Azure Core typed response and provides convenient access
+/// to headers, status code, the original request, and Cosmos-specific response metadata.
+#[derive(Debug)]
+pub struct CosmosResponse<T> {
+    /// The underlying typed HTTP response.
+    response: Response<T>,
```
FabianMeiswinkel, line 22: Azure Core's Response<T> should not leak into our public surface area - it forces JSON - while we might want to handle serialization differently etc.

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
FabianMeiswinkel: Just remove this - at least form public surface area - we will eventually have diagnostics and that would be a superset.

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
FabianMeiswinkel: Never mind - it is module-internal already

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
FabianMeiswinkel: Whether you keep it for now or not internally I don't care.

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
FabianMeiswinkel: Why do we need this in public surface area? Would be part of diagnostics later?

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /// Returns the request charge (RU consumption) for this operation, if available.
+    pub fn request_charge(&self) -> Option<f64> {
+        self.get_optional_header_str(&crate::constants::REQUEST_CHARGE)
+            .and_then(|s| s.parse::<f64>().ok())
+    }
+
+    /// Returns the activity ID for this request, if available.
+    pub fn activity_id(&self) -> Option<&str> {
```
FabianMeiswinkel: Remove from public surface area? Should be part of diagnostics?

sdk/cosmos/azure_data_cosmos/tests/framework/test_client.rs
```diff
@@ -102,6 +105,52 @@ fn get_shared_database_id() -> &'static str {
...
+        let account_name = &host[..pos];
+        format!("{}-{}.documents.azure.com", account_name, region_suffix)
+    } else {
+        // Fallback: just return the host as-is if it doesn't match expected format
+        host.to_string()
+    }
+}
+
+pub fn get_global_endpoint() -> String {
+    let connection_string_env = std::env::var(CONNECTION_STRING_ENV_VAR)
```
FabianMeiswinkel: Just curious - why even use connection string - that has surpassed its use with push to AAD?

FabianMeiswinkel (approved):
LGTM - Thanks

sdk/cosmos/azure_data_cosmos/tests/framework/test_client.rs
tvaron3: Yes your right. Stale comment. Updated it

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
tvaron3: changed to this approach.

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
tvaron3: It's still public surface area but behind the fault injection feature flag. Left a note that it is subject to change. I also changed to return the entire request as would need to validate request headers and other request properties as well later on.

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
tvaron3: I need this to validate cross region retries for fault injection. Yes later it will be part of diagnostics but need something I can use now for integration tests.

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
tvaron3: removed

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
tvaron3: changed to return the request now

sdk/cosmos/azure_data_cosmos/tests/framework/test_client.rs
tvaron3: True added the account host env variable and build it from there.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -101,8 +102,15 @@ impl GatewayPipeline {
...
-        // Convert RawResponse into typed Response<T> and return with the final request
-        res.map(|r| (r.into(), cosmos_request))
+        // Extract the final endpoint URL from the request context
+        // let endpoint = cosmos_request
+        //     .request_context
+        //     .location_endpoint_to_route
+        //     .clone()
+        //     .unwrap_or_else(|| self.endpoint.clone());
+
+        // Convert RawResponse into CosmosResponse with the final endpoint
```
Pilchie: Now this comment is stale.

tvaron3: fixed

Pilchie (approved):

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// Returns the final request used to fulfill the operation.
+    /// This api is subject to change without a major version bump.
+    ///
+    #[cfg(feature = "fault_injection")]
+    pub fn request(&self) -> &CosmosRequest {
+        &self.request
+    }
+
+    /// Consumes the response and returns the response body.
+    pub fn into_body(self) -> azure_core::http::response::ResponseBody {
```
kushagraThapar, line 61: this exposes core's ResponseBody, intentional?

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// Returns the ETag from this response, if available.
+    pub fn etag(&self) -> Option<&str> {
+        self.get_optional_header_str(&azure_core::http::headers::ETAG)
+    }
+}
+
+impl<T: DeserializeOwned> CosmosResponse<T> {
+    /// Deserializes the response body into a model type.
+    pub fn into_model(self) -> azure_core::Result<T> {
+        self.response.into_body().json()
```
kushagraThapar, line 86: curious how this handles cases when the json is not a valid type? like sometimes when gateway throws an error and its sort of a webpage and not a valid json?

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
kushagraThapar: If we decide to overload core's ResponseBody in future, and have something like CosmosResponseBody, does this allow us to do so in RUST?

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -101,8 +102,8 @@ impl GatewayPipeline {
         // Delegate to the retry handler, providing the sender callback
         let res = self.retry_handler.send(&mut cosmos_request, sender).await;
 
-        // Convert RawResponse into typed Response<T>
-        res.map(Into::into)
+        // Convert RawResponse into CosmosResponse with the final request
+        res.map(|r| CosmosResponse::new(r.into(), cosmos_request))
```
kushagraThapar, line 106: curious about the error handling here? Do we return some sort of Cosmos specific error or just general RUST error in the cases of invalid json handling or serialization issues (like duplicate fields, or something of the sorts which we have recently faced in Java SDK on cfp)?

kushagraThapar: And in cases of error, do we return cosmos headers, etc. or any other useful information like diagnostics?

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
tvaron3: There is no overloading in rust. we would have to add another method or do a breaking change. I agree we should change it in the future. Right now kept behavior the same as before. Goal of this pr is to be able to expose necessary information for validating fault injection and retry policies.

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
tvaron3: So the result returned is an azure core error or a successful response. Customer will need to have error handling like the following example where we keep retrying on 404 https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/cosmos/azure_data_cosmos/tests/framework/test_client.rs#L376

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
tvaron3: This will return a azure core error. The error is bubbled up to the customer. It will not include diagnostics. Not sure about headers but that will all change when diagnostics are implemented.

tvaron3: Any retries happen in the retry handler above

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
tvaron3: Added some tests for invalid json. It bubbles up the error like this to the customer. `Error: Error { context: Custom(Custom { kind: DataConversion, error: Error("EOF while parsing a value", line: 1, column: 0) }) }`

PR #3624, opened 2026-01-28
Cosmos: Resolve PR 3515 - Remove query engine and provide our own query result pager

### LINKED ISSUE #3413: QueryExecutor needs to implement its own Stream

simorenoh:
Resolving the merge conflicts on https://github.com/Azure/azure-sdk-for-rust/pull/3515 that were not easily resolvable in the same branch. I don't have write permissions to Ashley's fork, so had to make a separate PR from mine. As a side-note, this PR also links the `query_items` request to our retry mechanisms.

Copying the description for easier reviewing:

Fixes #3413 

When @heaths rewrote the pager, it broke how we were doing cross-partition query pagers. At the time, I agreed that it would be reasonable here for Cosmos to have its own pagination logic, since it needs to handle these cross-partition scenarios, and we already use our own `FeedPage<T>` type instead of `Response<T>`. At the time, Heath just commented out the old code and marked it `unimplemented!()` (at my suggestion). Now, I'm coming back to restore it, and bringing a few refactorings along with it:

* I moved Throughput Offer APIs to a crate-local `OffersClient`, which matches what we do in some other SDKs. This isn't visible to users, it's just used _by_ `ContainerClient`/`DatabaseClient` to implement throughput-related operations. This removes the throughput stuff from `CosmosPipeline`, which was clunky anyway
* I brought the "gateway-based" query execution logic in to `QueryExecutor`, which is now an enum with two different executor implementations: Gateway and QueryEngine.
* I moved the `send_query_request` logic out of `CosmosPipeline` and now the places that need to issue these queries construct a `QueryExecutor::gateway` and execute it. The result is the same code savings, but better architecture IMO.

I also restored the query engine tests, which are all passing for me.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
@@ -16,6 +16,18 @@ use azure_core::http::{response::Response, Context, PipelineSendOptions, RawResp
 use std::sync::Arc;
 use url::Url;
 
+/// Success codes: 200-299 range plus 304 (Not Modified)
+const SUCCESS_CODES: [u16; 101] = {
+    let mut codes = [0u16; 101];
+    let mut i = 0;
+    while i < 100 {
+        codes[i] = 200 + i as u16;
+        i += 1;
+    }
+    codes[100] = 304;
+    codes
+};
+
 /// Newtype that wraps an Azure Core pipeline to provide a Cosmos-specific pipeline which configures our authorization policy and enforces that a [`ResourceType`] is set on the context.
 #[derive(Debug, Clone)]
 pub struct GatewayPipeline {
@@ -60,17 +72,6 @@ impl GatewayPipeline {
         let sender = |req: &mut CosmosRequest| {
             let pipeline = self.pipeline.clone();
             let ctx = context.clone();
-            // Success codes: 200-299 range plus 304 (Not Modified)
-            const SUCCESS_CODES: [u16; 101] = {
-                let mut codes = [0u16; 101];
-                let mut i = 0;
-                while i < 100 {
-                    codes[i] = 200 + i as u16;
-                    i += 1;
-                }
-                codes[100] = 304;
-                codes
-            };
             let success_options = CheckSuccessOptions {
                 success_codes: &SUCCESS_CODES,
             };


sdk/cosmos/azure_data_cosmos/src/query/executor.rs
```diff
@@ -1,246 +1,117 @@
...
     http_pipeline: Arc<GatewayPipeline>,
-    container_link: ResourceLink,
     items_link: ResourceLink,
     context: Context<'static>,
-    query_engine: QueryEngineRef,
-    base_request: Option<Request>,
     query: Query,
-    pipeline: Option<OwnedQueryPipeline>,
-
-    // Why is our phantom type a function? Because that represents how we _use_ the type T.
```
tvaron3, line 28: Nit: Isn't this comment still relevant?

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -8,10 +8,11 @@
 
 ### Breaking Changes
 
+- Changed return type of query methods from `FeedPager<T>` (an alias for `ItemIterator<FeedPage<T>, String>`) to `FeedItemIterator<T>`, which implements `Stream<Item = Result<T>>` and provides `into_pages()` for page-level access. ([#3515](https://github.com/Azure/azure-sdk-for-rust/pull/3515))
+
 ### Bugs Fixed
 
 ### Other Changes
-
```
tvaron3, line 15: nit: revert

sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_query.rs
```diff
@@ -106,7 +106,7 @@ pub async fn single_partition_query_with_projection() -> Result<(), Box<dyn Erro
 pub async fn cross_partition_query_with_projection_and_filter() -> Result<(), Box<dyn Error>> {
     TestClient::run_with_unique_db(
         async |run_context, db_client| {
-            let items = test_data::generate_mock_items(10, 10);
+            let items = test_data::generate_mock_items(10, 2);
```
tvaron3, line 109: just curious, why did these change ?

sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_query.rs
```diff
@@ -133,43 +133,3 @@ pub async fn cross_partition_query_with_projection_and_filter() -> Result<(), Bo
     )
     .await
 }
-
-#[tokio::test]
-pub async fn cross_partition_query_with_order_by_fails_without_query_engine(
```
Pilchie, line 138: Shouldn't we keep this until we add the cross-partition code in?

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -62,171 +53,42 @@ impl GatewayPipeline {
...
+    ) -> azure_core::Result<RawResponse> {
         cosmos_request.client_headers(&self.options);
         // Prepare a callback delegate to invoke the http request.
-        let sender = move |req: &mut CosmosRequest| {
+        let sender = |req: &mut CosmosRequest| {
+            let pipeline = self.pipeline.clone();
             let ctx = context.clone();
-            let url = req.resource_link.clone();
+            let success_options = CheckSuccessOptions {
+                success_codes: &[200, 201, 202, 204, 304],
```
FabianMeiswinkel: We need to reconsider how we check success - just whitlisting specific status codes will break sooner or later - backend/service relies on the fact that we honor HTTP semantic - like 200 - 299 success etc. - this is diffferent than in any other SDK and while having stricter validations is understandable - we have to thing this through e2e.

sdk/cosmos/azure_data_cosmos/src/operation_context.rs
```diff
@@ -23,6 +25,27 @@ pub enum OperationType {
 }
 
 impl OperationType {
+    /// Returns the content type of the body for this operation type, if any.
+    ///
+    /// If `None` is returned, no body content type header should be set and the request is expected to have no body.
```
FabianMeiswinkel, line 30: For query plan that is not true? The query plan API needs to send the query plan in requets payload - and uses POST?

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
```diff
@@ -84,7 +84,7 @@ impl PartitionKey {
     /// An empty list of partition key values, which is used to signal a cross-partition query, when querying a container.
     pub const EMPTY: PartitionKey = PartitionKey(Vec::new());
 
-    #[cfg_attr(not(feature = "preview_query_engine"), allow(dead_code))]
+    #[allow(dead_code)]
```
FabianMeiswinkel, line 87: Just curious - would this be needed even if there would be a test executing it?

IMO for public surface area we should have at least some sanity unit tests?

sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_query.rs
FabianMeiswinkel: +1

sdk/cosmos/azure_data_cosmos/src/operation_context.rs
simorenoh: moved query plan into APPLICATION_JSON to match with its behavior

sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_query.rs
simorenoh: I don't see any reason why - changed it back

sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_query.rs
simorenoh: re-added this

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
simorenoh: Will track this as part of a greater query testing effort.

tvaron3 (approved):

PR #3628, opened 2026-01-28
Add azsdk mcp to copilot setup

benbp:
CC @heaths @RickWinter

### DIFF
--- a/.github/workflows/copilot-setup-steps.yml
+++ b/.github/workflows/copilot-setup-steps.yml
@@ -52,6 +52,11 @@ jobs:
           echo "CARGO_TERM_COLOR=always" >> $GITHUB_ENV
           echo "RUSTFLAGS=-D warnings" >> $GITHUB_ENV
 
+      - name: Install azsdk mcp server
+        shell: pwsh
+        run: |
+          ./eng/common/mcp/azure-sdk-mcp.ps1 -InstallDirectory $HOME/bin
+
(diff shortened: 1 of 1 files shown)


benbp:
@RickWinter I'm not sure I understand the failure case here. The underyling install script validates the download and copies it to the specified file location. Github actions images have powershell installed otherwise that yaml shell setting wouldn't be supported. If the subsequent coding agent mcp startup fails it does so silently, which is a big pain, but I don't think any checks here have any bearing on that unless we want to try to start the server during then send a sigterm to it.

.github/workflows/copilot-setup-steps.yml
```diff
@@ -76,3 +76,9 @@ jobs:
           echo "=== Workspace Structure ==="
           find . -name "Cargo.toml" | head -10
           echo "=== Environment Ready for Copilot Coding Agent ==="
+
+      - name: Install azsdk mcp server
```
heaths: Just a nit, but not blocking: shouldn't this go before "Verify workspace setup" and "Environment summary". Right now those are cargo-specific, but I could imagine someone/some LLM thinking those are good places to dump other output that might be affected by your script here. Seems this should be closer to the beginning.

benbp: I wasn't sure what the best place was, I'll move it before those thanks.

PR #3642, opened 2026-01-30
Allow localhost authority for mocking

jlaundry:
Since `azure_identity` 0.28, only HTTPS authority endpoints have been allowed. This creates a problem when unit testing, for example, when we want to test our own SpecificAzureCredential implementation.

This PR simply allows HTTP authority endpoints when the host address is localhost.

### DIFF
--- a/sdk/identity/azure_identity/src/lib.rs
+++ b/sdk/identity/azure_identity/src/lib.rs
@@ -154,10 +154,13 @@ fn get_authority_host(env: Option<Env>, cloud: Option<&CloudConfiguration>) -> R
 
     let url = Url::parse(&authority_host)?;
     if url.scheme() != "https" {
-        return Err(Error::with_message(
-            ErrorKind::Other,
-            format!("authority host doesn't use HTTPS scheme: {authority_host}"),
-        ));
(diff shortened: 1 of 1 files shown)


heaths:
@chlowell can you take a look?

chlowell (changes requested):
Allowing `http://localhost` as the base URL for Entra ID is weird and scary because that's definitely not the base URL for Entra ID. I recommend instead following the SDK's approach for testing credentials against a fake Entra: implement  `azure_core::http::HttpClient` for some type to return programmatically defined responses instead of sending HTTP requests i.e., move your mock in-proc. See for example [ClientSecretCredential tests](https://github.com/Azure/azure-sdk-for-rust/blob/9806c5618da8bc2678569ad523a9141826257760/sdk/identity/azure_identity/src/client_secret_credential.rs#L250) and implementation of [MockSts](https://github.com/Azure/azure-sdk-for-rust/blob/9806c5618da8bc2678569ad523a9141826257760/sdk/identity/azure_identity/src/lib.rs#L384). If you prefer the external server approach, you could use a mock transport to send your credential's outgoing requests to that server, or a pipeline policy to change those requests' URLs.

PR #3644, opened 2026-01-31
refactor(core): optimize UserAgentPolicy and add validation

raushan728:
Refactored `UserAgentPolicy` to improve code style and runtime safety.

- Consolidated imports to follow repository standards.
- Removed redundant string allocation in `send` method.
- Added CRLF validation and a unit test to prevent malformed headers.


sdk/core/azure_core/src/http/policies/user_agent.rs
```diff
@@ -1,14 +1,15 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
```
LarryOsterman, line 1: Silly question: Why did you remove the copyright and license notice?

sdk/core/azure_core/src/http/policies/user_agent.rs
```diff
@@ -79,7 +80,14 @@ impl Policy for UserAgentPolicy {
         request: &mut Request,
         next: &[Arc<dyn Policy>],
     ) -> PolicyResult {
-        request.insert_header(USER_AGENT, HeaderValue::from(self.header.to_string()));
+        if self.header.contains(['\r', '\n']) {
```
LarryOsterman: Silly question: Would it make sense to URL encode the header rather than just looking for CR/LF characters? The only element of the user-agent which is user configurable is the application name, and by URL encoding that field, it allows an application name to include arbitrary characters in the header.

Also, CR and LF are not the only illegal characters in a header - if you're going to reject the headers, you should reject all non-header octets.

sdk/core/azure_core/src/http/policies/user_agent.rs
```diff
@@ -79,7 +80,14 @@ impl Policy for UserAgentPolicy {
...
+            return Err(crate::error::Error::new(
+                crate::error::ErrorKind::DataConversion,
+                "User agent header contains invalid characters",
+            ));
+        }
+
+        request.insert_header(USER_AGENT, HeaderValue::from(self.header.clone()));
```
heaths: What @LarryOsterman said, but FWIW, in reference to your PR comment, this still allocates a new string necessarily:

> Removed redundant string allocation in send method.

That said, it's certainly less complex since it's not going through the formatter so still better. It does make me wonder, however, whether the `header` field should instead be a `Cow<'static, str>` instead.

heaths: Never mind. As @LarryOsterman reminded me, the header *value* is an owned `String` so it wouldn't matter anyway. It'd be too rare that a `Cow` would actually make sense for the value, so please disregard.

sdk/core/azure_core/src/http/policies/user_agent.rs
LarryOsterman: Also, validation should be done in the constructor, that way the failure doesn't happen during pipeline execution (there will be many pipeline `Send` calls happening and the value of the `header` is immutable after construction.

sdk/core/azure_core/src/http/policies/user_agent.rs
raushan728: Sorry about that i accidentally removed it thinking it was unnecessary. I've restored it

sdk/core/azure_core/src/http/policies/user_agent.rs
raushan728: Done. I've moved the validation to the constructor and implemented percent-encoding for the application ID as suggested. Thanks for the guidance

sdk/core/azure_core/src/http/policies/user_agent.rs
```diff
@@ -54,19 +60,49 @@ impl<'a> UserAgentPolicy {
...
+                }) {
+                    panic!("application_id contains invalid characters. Only RFC 9110 tokens are allowed.");
+                }
+
                 format!("{application_id} azsdk-rust-{crate_name}/{crate_version} {platform_info}")
             }
             None => format!("azsdk-rust-{crate_name}/{crate_version} {platform_info}"),
         };
 
+        let header = HeaderValue::from(header_str);
```
LarryOsterman: This is a *complete* nit, but the `header` variable is unnecessary (and will likely be elided by the compiler)

raushan728: Nit addressed. Inlined the conversion

PR #3645, opened 2026-02-02
feat: Add SAS token support to azure_storage_queue

lsiddiquee:
# Adds support for Shared Access Signature (SAS) token authentication to the Azure Queue Storage client library, aligning with the authentication patterns used in azure_storage_blob.

### Features Added
- Added `from_url()` constructors for `QueueClient` and `QueueServiceClient` to enable direct URL-based client construction
- Added support for SAS token authentication through URL query parameters
- Made `credential` parameter optional (`Option<Arc<dyn TokenCredential>>`) on both `new()` and `from_url()` methods to support multiple authentication scenarios:
  - Entra ID authentication (when credential is provided)
  - SAS token authentication (when credential is None and URL contains SAS token)
  - Public access (when credential is None and no authentication in URL)

### Breaking Changes
- Changed `credential` parameter to `Option<Arc<dyn TokenCredential>>` on client constructors
- Changed `queue_name` parameter from owned `String` to `&str` reference on `QueueClient::new()`
- Removed `queue_name()` accessor from `QueueClient`
- Removed public `endpoint` struct field (now accessed via method)
- Changed `QueueServiceClient::queue_client()` to return `Result<QueueClient>` instead of `QueueClient`

### Implementation Details
- Created extension methods on generated clients (`GeneratedQueueClient` and `GeneratedQueueServiceClient`) to implement custom construction logic
- Added conditional pipeline building with `BearerTokenAuthorizationPolicy` only when credentials are provided
- Fixed test recording issue where child clients created by `QueueServiceClient` weren't inheriting test proxy instrumentation
- Stored `QueueServiceClientOptions` in `QueueServiceClient` to pass client options to child `QueueClient` instances

### Examples and Tests
- Updated examples to demonstrate both Entra ID and SAS token authentication
- Added comprehensive test coverage for SAS token scenarios
- Refactored examples to use shared test logic for both authentication methods
- All tests pass in both record and playback modes

### Documentation
- Updated README with corrected credential parameter usage
- Added example README documenting required environment variables for SAS token testing
- Updated CHANGELOG following azure_storage_blob conventions

### DIFF
--- a/sdk/storage/azure_storage_queue/CHANGELOG.md
+++ b/sdk/storage/azure_storage_queue/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release History
 
-## 0.4.0 (Unreleased)
+## 0.3.0 (Unreleased)
 
 ### Features Added
 
@@ -8,6 +8,7 @@
   - `QueueClient::from_url()`
   - `QueueServiceClient::from_url()`
 - Added support for SAS (shared access signature) URLs via the new `from_url()` methods.
+- Added `continuation_token` to `PagerOptions` for methods that return a `Pager`.
 
 ### Breaking Changes
 
@@ -16,19 +17,6 @@
 - Changed the `queue_name` parameter from owned `String` to `&str` reference on `QueueClient::new()`.
 - The `credential` parameter is now `Option<Arc<dyn TokenCredential>>` on `new()` and `from_url()` client constructors, allowing for construction of public access clients and clients using SAS tokens.
 - Changed `QueueServiceClient::queue_client()` to return `Result<QueueClient>` instead of `QueueClient`.
-
-### Bugs Fixed
-
-### Other Changes
-
-## 0.3.0 (Unreleased)
-
-### Features Added
-
-- Added `continuation_token` to `PagerOptions` for methods that return a `Pager`.
-
-### Breaking Changes
-
 - Removed `Pager::with_continuation_token()` for methods that return a `Pager`.
 
 ### Bugs Fixed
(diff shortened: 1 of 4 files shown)


sdk/storage/azure_storage_queue/CHANGELOG.md
```diff
@@ -1,5 +1,26 @@
 # Release History
 
+## 0.4.0 (Unreleased)
```
vincenttran-msft: Small nit: Since we are on `0.3.0` unreleased still, all these changelog entries should probably move down 😄 and then `0.4.0` section deleted

lsiddiquee: Thanks for the catch. I confused unreleased with beta release.

sdk/storage/azure_storage_queue/tests/queue_service_client.rs
```diff
@@ -246,3 +246,111 @@ fn assert_successful_response<T, F>(response: &Response<T, F>) {
         response.status()
     );
 }
+
+/// Test creating, listing, and deleting a queue using SAS token authentication with from_url
+#[recorded::test]
+async fn test_sas_token_from_url(ctx: TestContext) -> Result<()> {
```
jalauzon-msft, line 252: Unfortunately, we can't really check-in SAS tests for now as we don't have the ability to generate a SAS from the SDK. Also be careful about accidentally recording a real SAS URL into the assets and that would leak your account info (and flag us). I took a look at your assets for this PR, and it doesn't look like you leaked anything (the recordings for these tests were empty for some reason) but best to just remove these checked-in tests for now.

When removing tests, it's best to delete the recordings from the assets too.

PR #3648, opened 2026-02-02
Normalize CODEOWNERS file

danieljurek:
This PR normalizes the CODEOWNERS file by sorting owners, labels, and entries.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -55,29 +55,29 @@
 ####################
 
 # PRLabel: %Cosmos
-/sdk/cosmos/             @Azure/azure-cosmos-rust-sdk
+/sdk/cosmos/    @Azure/azure-cosmos-rust-sdk
 
+# AzureSdkOwners: @heaths
 # ServiceLabel: %Cosmos
-# AzureSDKOwners: @heaths
 # ServiceOwners: @Pilchie
 
+# PRLabel: %Event Hubs
+/sdk/eventhubs/    @j7nw4r
+
+# AzureSdkOwners: @j7nw4r
+# ServiceLabel: %Event Hubs
+
 # PRLabel: %KeyVault
-/sdk/keyvault/           @Azure/azure-sdk-write-keyvault @heaths
+/sdk/keyvault/    @Azure/azure-sdk-write-keyvault @heaths
 
 # ServiceLabel: %KeyVault
 # ServiceOwners: @Azure/azure-sdk-write-keyvault
 
 # PRLabel: %Storage
-/sdk/storage/            @heaths @RickWinter @LarryOsterman @vincenttran-msft @jalauzon-msft
+/sdk/storage/    @heaths @jalauzon-msft @LarryOsterman @RickWinter @vincenttran-msft
 
+# AzureSdkOwners: @jalauzon-msft @vincenttran-msft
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -55,29 +55,29 @@
 ####################
 
 # PRLabel: %Cosmos
-/sdk/cosmos/             @Azure/azure-cosmos-rust-sdk
+/sdk/cosmos/                                                       @Azure/azure-cosmos-rust-sdk
 
+# AzureSdkOwners:                                                  @heaths
```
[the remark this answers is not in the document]
danieljurek: This is as intended.

.github/CODEOWNERS
```diff
@@ -55,29 +55,29 @@
 ####################
 
 # PRLabel: %Cosmos
-/sdk/cosmos/             @Azure/azure-cosmos-rust-sdk
+/sdk/cosmos/                                                       @Azure/azure-cosmos-rust-sdk
```
heaths: Nit: I hate changes like this. If the longest line "necessitating" this spacing is ever changed, then the whole file changes. It's a massive diff with no real value. Can we not abuse spacing like this please? /cc @weshaggard

.github/CODEOWNERS
```diff
@@ -55,29 +55,29 @@
...
-/sdk/keyvault/           @Azure/azure-sdk-write-keyvault @heaths
+/sdk/keyvault/                                                     @Azure/azure-sdk-write-keyvault @heaths
 
 # ServiceLabel: %KeyVault
-# ServiceOwners: @Azure/azure-sdk-write-keyvault
+# ServiceOwners:                                                   @Azure/azure-sdk-write-keyvault
 
 # PRLabel: %Storage
-/sdk/storage/            @heaths @RickWinter @LarryOsterman @vincenttran-msft @jalauzon-msft
+/sdk/storage/                                                      @heaths @jalauzon-msft @LarryOsterman @RickWinter @vincenttran-msft
```
heaths: Can you also pick up <https://github.com/Azure/azure-sdk-for-rust/pull/3661/changes#diff-3d36a1bf06148bc6ba1ce2ed3d19de32ea708d955fed212c0d27c536f0bd4da7>? I worry about the order of all the "meta labels". That line is more about Storage than Event Hubs, so should it go above the Storage service labels?

.github/CODEOWNERS
danieljurek: I'm going to move to using a constant amount of padding. It's less code, less stuff to maintain for a file that's going to be read/written primarily by automation. 

That being said, spacing was calculated per-entry so an expansion of one CODEOWNERS entry wouldn't affect others. So if you had a really long path it'd just push out the owners on that line, not others.

.github/CODEOWNERS
heaths: Based on our offline conversation, it sounds like that's not the case - that those other labels' position in the file are irrelevant so we should be good here. Still, might be good to pick up anyway so Storage has less of a major diff to deal with if you go in first. /cc @vincenttran-msft

danieljurek: If that branch merges soon I'll re-run normalization on it and get things sorted.

heaths (approved):

Issue #3518, opened 2026-01-12
Cosmos: Per Partition Automatic Failover


simorenoh:
Support PPAF in the Rust Cosmos DB SDK.

Issue #3171, opened 2025-10-09
Cosmos: Add Support for Partition Scoped Automatic Failover


kundadebdatta:
### Acceptance Criteria:

- Add support for per-partition automatic failover.
- Add support for per partition circuit breaker.

Issue #3519, opened 2026-01-12
Cosmos: Per Partition Circuit Breaker


simorenoh:
Support PPCB in the Rust Cosmos DB SDK

PR #3656, opened 2026-02-03
Cosmos: Build `GlobalPartitionEndpointManager` to Support Per Partition Automatic Failover

kundadebdatta:
## Summary

This PR introduces **partition-level failover** for the Azure Cosmos DB Rust SDK, enabling the client to detect per-partition failures and re-route individual partition key ranges to alternate regional endpoints — without affecting healthy partitions. This brings the Rust SDK in line with the .NET and Java SDKs' PPAF/PPCB capabilities.

Two independent but complementary mechanisms are introduced:

| Mechanism | Scope | Trigger |
|---|---|---|
| **Per-Partition Automatic Failover (PPAF)** | Write requests on **single-master** accounts | `403/3` (WriteForbidden) — the service signals a regional write outage |
| **Per-Partition Circuit Breaker (PPCB)** | Read requests (any account) + write requests on **multi-master** accounts | Consecutive failure count exceeds configurable threshold (default: reads > 2, writes > 5) |

### New Modules

| File | Purpose |
|---|---|
| `src/routing/global_partition_endpoint_manager.rs` (~2,400 lines) | Core PPAF/PPCB engine — failover maps, circuit breaker logic, background failback loop, 71 unit tests |
| `src/hash.rs` (~478 lines) | Effective Partition Key (EPK) hash computation (V1 + V2) |
| `src/murmur_hash.rs` (~283 lines) | MurmurHash3 implementation (32-bit + 128-bit) for EPK generation |

### Modified Modules

| File | Change |
|---|---|
| `src/retry_policies/client_retry_policy.rs` | Integrates GPEM into the retry pipeline — intercepts `403/3`, `503`, `429/3092` to trigger partition failover |
| `src/handler/container_connection.rs` | Resolves `PartitionKeyRange` before sending, populates `resolved_partition_key_range` and `resolved_collection_rid` on `RequestContext` |
| `src/clients/cosmos_client.rs` | Constructs `GlobalPartitionEndpointManager` and threads it through the pipeline |
| `src/clients/database_client.rs`, `src/clients/container_client.rs` | Pass `GlobalPartitionEndpointManager` down to `ContainerConnection` |
| `src/pipeline/mod.rs`, `src/handler/retry_handler.rs` | Wire GPEM into `GatewayPipeline` → `BackOffRetryHandler` → `ClientRetryPolicy` |
| `src/partition_key.rs` | Added `get_hashed_partition_key_string()` for EPK computation; added `Bool`, `Infinity`, `Undefined` variants |
| `src/request_context.rs` | Replaced placeholder `PartitionKeyRange` with real type from `routing::partition_key_range`; added `resolved_collection_rid` |
| `src/resource_context.rs` | Added `database_id()` and `container_id()` extraction from `ResourceLink` paths |
| `src/cosmos_request.rs` | Added `database_id()` / `container_id()` convenience methods |
| `src/constants.rs` | Added `SYSTEM_RESOURCE_NOT_AVAILABLE` (3092) sub-status code |
| `src/routing/location_cache.rs` | Bug fix: clear endpoints before repopulating on refresh; restore availability-ordered endpoint rebuild |

## Architecture

```mermaid
sequenceDiagram
    participant App as Application
    participant CRP as ClientRetryPolicy
    participant GPEM as GlobalPartitionEndpointManager
    participant GEM as GlobalEndpointManager
    participant Service as Cosmos DB Service

    Note over App, Service: Normal Request Flow (PPAF/PPCB enabled)

    App->>CRP: send(request)
    CRP->>CRP: before_send_request(request)
    CRP->>GPEM: try_add_partition_level_location_override(request)
    GPEM->>GPEM: Lookup PartitionKeyRange in failover map
    alt Override exists & threshold met
        GPEM-->>CRP: true (request.location_endpoint_to_route = override)
    else No override
        GPEM-->>CRP: false (use default endpoint)
    end
    CRP->>Service: HTTP request to resolved endpoint

    alt Success (2xx)
        Service-->>CRP: 200 OK
        CRP-->>App: CosmosResponse
    else Write Forbidden (403/3) — Single Master
        Service-->>CRP: 403 / WriteForbidden
        Note over CRP: PPAF Path
        CRP->>GPEM: try_mark_endpoint_unavailable_for_partition_key_range(req)
        GPEM->>GPEM: Add override → next account read region
        GPEM-->>CRP: true
        CRP-->>CRP: Retry with Duration::ZERO
        CRP->>GPEM: try_add_partition_level_location_override(request)
        GPEM-->>CRP: Route to failover region
        CRP->>Service: Retry to failover region
    else Service Unavailable (503) or 429/3092
        Service-->>CRP: 503 / ServiceUnavailable
        Note over CRP: PPCB Path
        CRP->>GPEM: increment_failure_counter_and_check_if_can_failover(req)
        GPEM->>GPEM: Increment consecutive failure count
        alt Threshold exceeded
            GPEM-->>CRP: true
            CRP->>GPEM: try_mark_endpoint_unavailable(req)
            GPEM->>GPEM: Move to next preferred read region
        else Below threshold
            GPEM-->>CRP: false
        end
        CRP->>CRP: should_retry_on_unavailable_endpoint_status_codes()
    end

    Note over GPEM: Background Failback Loop (every 5 min)
    loop Every background_connection_init_interval_secs
        GPEM->>GPEM: Scan failover maps for stale entries
        GPEM->>GPEM: partition_unavailability_duration exceeded?
        alt Eligible for failback
            GPEM->>GPEM: mark_endpoints_to_healthy()
            GPEM->>GPEM: Remove override → traffic returns to original region
        end
    end
```

## Key Design Decisions

- **Two failover maps** — partition_key_range_to_location_for_write handles PPAF (single-master writes → read regions); partition_key_range_to_location_for_read_and_write handles PPCB (reads + multi-master writes → preferred read regions). They are independent to avoid cross-contamination.

- **Sequential region exhaustion** — When a partition fails over, it tries each available region in order. If all regions are exhausted, the override is removed and the partition falls back to default routing.

- **Configurable thresholds via env vars** — All critical parameters are overridable:

  - AZURE_COSMOS_CIRCUIT_BREAKER_CONSECUTIVE_FAILURE_COUNT_FOR_READS (default: 2)
  - AZURE_COSMOS_CIRCUIT_BREAKER_CONSECUTIVE_FAILURE_COUNT_FOR_WRITES (default: 5)
  - AZURE_COSMOS_CIRCUIT_BREAKER_TIMEOUT_COUNTER_RESET_WINDOW_IN_MINUTES (default: 5)
  - AZURE_COSMOS_ALLOWED_PARTITION_UNAVAILABILITY_DURATION_IN_SECONDS (default: 5)
  - AZURE_COSMOS_PPCB_STALE_PARTITION_UNAVAILABILITY_REFRESH_INTERVAL_IN_SECONDS (default: 300)
  
- **EPK hash computation** — Added MurmurHash3 and V1/V2 effective partition key hash generation so that ContainerConnection can resolve PartitionKeyRange from a PartitionKey before sending, enabling per-partition routing.

- **Background failback** — A fire-and-forget async loop periodically marks stale failed partitions back to healthy, restoring traffic to the original region after the unavailability window expires.

## Closing issues

Fixes Azure/azure-sdk-for-rust#3171 #3518 #3519

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/partition_key.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/partition_key.rs
@@ -184,7 +184,6 @@ impl AsHeaders for PartitionKey {
                     json.push_str(&num.to_string());
                 }
                 InnerPartitionKeyValue::Infinity => json.push_str("\"Infinity\""),
-                InnerPartitionKeyValue::Undefined => json.push_str("undefined"),
             }
 
             json.push(',');


sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2386 @@
...
+        let interval = Duration::seconds(self.background_connection_init_interval_secs);
+
+        loop {
+            // Use the runtime-agnostic sleep from azure_core
+            get_async_runtime().sleep(interval).await;
+
+            info!("GlobalPartitionEndpointManager: initiate_circuit_breaker_failback_loop() un-deterministically marking the failed partitions back to healthy.");
+
+            if let Err(e) = self.initiate_failback_to_unhealthy_endpoints().await {
+                tracing::error!("GlobalPartitionEndpointManager: initiate_circuit_breaker_failback_loop() - filed to mark the failed partitions back to healthy. Exception: {}", e);
```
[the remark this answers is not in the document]
kundadebdatta: Thanks, this is now addressed.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2386 @@
...
+        } else {
+            "Circuit Breaker"
+        };
+
+        // Get the resolved collection RID from the request context
+        let collection_rid = request
+            .request_context
+            .resolved_collection_rid
+            .clone()
+            .unwrap();
```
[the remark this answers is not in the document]
kundadebdatta: Thanks, this is now addressed.

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -30,19 +32,48 @@ impl ContainerConnection {
...
+                .await?;
+
+            let pk_def = container_prop.partition_key;
+            let routing_map = self
+                .pk_range_cache
+                .try_lookup(&container_prop.id, None)
+                .await?
+                .unwrap();
+
+            let partition_key = cosmos_request.clone().partition_key.unwrap();
```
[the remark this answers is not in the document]
kundadebdatta: Thanks, this is now addressed.

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -30,19 +32,48 @@ impl ContainerConnection {
...
+            let epk = partition_key
+                .get_hashed_partition_key_string(pk_def.kind, pk_def.version.unwrap() as u8);
```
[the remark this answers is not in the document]
kundadebdatta: Thanks, this is now addressed.

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
```diff
@@ -161,16 +177,16 @@ impl AsHeaders for PartitionKey {
...
-    Number(serde_json::Number), // serde_json::Number has special integer handling, so we'll use that.
-}
+//
+// // We don't want to expose the implementation details of PartitionKeyValue (specifically the use of serde_json::Number), so we use this inner private enum to store the data.
+// #[derive(Debug, Clone, PartialEq, Eq)]
+// enum InnerPartitionKeyValue {
+//     Null,
+//     String(Cow<'static, str>),
+//     Number(serde_json::Number), // serde_json::Number has special integer handling, so we'll use that.
+// }
```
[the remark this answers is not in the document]
kundadebdatta: Thanks, this is now addressed.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2386 @@
...
+                partition_key_range_failover.current,
+                partition_key_range.id
+            );
+
+            request.request_context.route_to_location_endpoint(
+                partition_key_range_failover
+                    .current
+                    .clone()
+                    .parse()
+                    .unwrap(),
```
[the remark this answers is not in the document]
kundadebdatta: Thanks, this is now addressed.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2386 @@
...
+            self.is_request_eligible_for_partition_failover(request, true)
+        else {
+            return false;
+        };
+
+        let collection_rid = request
+            .request_context
+            .resolved_collection_rid
+            .clone()
+            .unwrap();
```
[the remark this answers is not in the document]
kundadebdatta: Thanks, this is now addressed.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -422,6 +451,20 @@ impl ClientRetryPolicy {
         if status_code == StatusCode::Forbidden
             && sub_status_code == Some(SubStatusCode::WRITE_FORBIDDEN)
         {
+            if self
+                .partition_key_range_location_cache
+                .partition_level_failover_enabled()
+                && self
+                    .partition_key_range_location_cache
+                    .try_mark_endpoint_unavailable_for_partition_key_range(
+                        &self.request.clone().unwrap(),
```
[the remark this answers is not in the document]
kundadebdatta: Thanks, this is now addressed.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -91,15 +99,20 @@ impl ClientRetryPolicy {
     /// A new `ClientRetryPolicy` instance configured with default retry limits and throttling behavior
     pub fn new(
         global_endpoint_manager: Arc<GlobalEndpointManager>,
+        partition_key_range_location_cache: Arc<GlobalPartitionEndpointManager>,
         excluded_regions: Option<Vec<RegionName>>,
     ) -> Self {
+        // let partition_key_range_location_cache =
+        //     GlobalPartitionEndpointManager::new(global_endpoint_manager.clone(), true, true);
```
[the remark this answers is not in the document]
kundadebdatta: Thanks, this is now addressed.

sdk/cosmos/azure_data_cosmos/src/hash.rs
```diff
@@ -0,0 +1,537 @@
...
+//         let mut ms: Vec<u8> = Vec::new();
+//         write_for_hashing_v2(comp, &mut ms);
+//         let hash_128 = murmurhash3_128(&ms, 0);
+//         let mut hash_bytes = hash_128.to_le_bytes();
+//         hash_bytes.reverse();
+//         hash_bytes[0] &= 0x3F;
+//         pieces.push(bytes_to_hex_upper(&hash_bytes));
+//     }
+//     pieces.join("").to_uppercase()
+// }
```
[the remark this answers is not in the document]
kundadebdatta: Thanks, this is now addressed.

sdk/cosmos/azure_data_cosmos/src/hash.rs
```diff
@@ -0,0 +1,537 @@
...
+
+    //     for (components, expected) in cases {
+    //         let actual = get_hashed_partition_key_string(
+    //             &components,
+    //             Some(PartitionKeyKind::MultiHash),
+    //             Some(2),
+    //         );
+    //         assert_eq!(actual, expected, "Mismatch for multi-hash composite key");
+    //     }
+    // }
```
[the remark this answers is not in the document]
kundadebdatta: Thanks, this is now addressed.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2386 @@
...
+            PartitionKeyRange,
+            (String, String, PartitionHealthStatus),
+        >,
+    ) {
+        for (pk_range, mapping) in pk_range_uri_mappings.iter_mut() {
+            info!(
+                "Un-deterministically marking the original failed endpoint: {}, for the PkRange: {}, collectionRid: {} back to healthy.",
+                mapping.0,
+                pk_range.id,
+                mapping.1
```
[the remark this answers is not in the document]
kundadebdatta: Thanks, this is now addressed.

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
```diff
@@ -138,10 +154,10 @@ impl AsHeaders for PartitionKey {
                     json.push('"');
                 }
                 InnerPartitionKeyValue::Number(ref num) => {
-                    json.push_str(
-                        serde_json::to_string(&serde_json::Value::Number(num.clone()))?.as_str(),
-                    );
+                    json.push_str(&num.to_string());
                 }
+                InnerPartitionKeyValue::Infinity => json.push_str("\"Infinity\""),
+                InnerPartitionKeyValue::Undefined => json.push_str("undefined"),
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2443 @@
...
+
+            request.request_context.route_to_location_endpoint(
+                partition_key_range_failover
+                    .current
+                    .clone()
+                    .parse()
+                    .unwrap(),
+            );
+
+            return true;
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -30,19 +32,97 @@ impl ContainerConnection {
...
+                                .try_lookup(&container_prop.id, Some(routing_map))
+                                .await?;
+
+                            if let Some(refreshed_routing_map) = refreshed_routing_map {
+                                let pk_range = refreshed_routing_map
+                                    .get_range_by_effective_partition_key(&epk)?;
+                                cosmos_request.request_context.resolved_partition_key_range =
+                                    Some(pk_range.clone());
+                            }
+                        }
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -30,19 +32,97 @@ impl ContainerConnection {
...
+                            let refreshed_routing_map = self
+                                .pk_range_cache
+                                .try_lookup(&container_prop.id, Some(routing_map))
+                                .await?;
+
+                            if let Some(refreshed_routing_map) = refreshed_routing_map {
+                                let pk_range = refreshed_routing_map
+                                    .get_range_by_effective_partition_key(&epk)?;
+                                cosmos_request.request_context.resolved_partition_key_range =
+                                    Some(pk_range.clone());
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/hash.rs
```diff
@@ -0,0 +1,478 @@
...
+    }
+
+    /// Truncate strings to 100 characters for V1 hashing (other types unchanged).
+    fn truncate_for_v1_hashing(&self) -> InnerPartitionKeyValue {
+        match self {
+            InnerPartitionKeyValue::String(s) => {
+                if s.len() > 100 {
+                    InnerPartitionKeyValue::String(s[..100].to_string())
+                } else {
+                    InnerPartitionKeyValue::String(s.clone())
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -30,19 +32,97 @@ impl ContainerConnection {
...
+                if let Some(pk_range) = cosmos_request.partition_key_range_identity.as_ref() {
+                    let pk_range = self
+                        .pk_range_cache
+                        .resolve_partition_key_range_by_id(
+                            &pk_range.collection_rid,
+                            &pk_range.partition_key_range_id,
+                            false,
+                        )
+                        .await
+                        .unwrap();
```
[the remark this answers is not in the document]
kundadebdatta: This is now addressed.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2444 @@
...
+
+        let weak_self = Arc::downgrade(self);
+        let shutdown_token = self.background_task_manager.cancellation_token();
+        // Spawn via BackgroundTaskManager so the task is tracked and will be
+        // signaled to stop when the manager (and thus the client) is dropped.
+        // We capture a Weak<Self> (not Arc<Self>) to avoid a reference cycle
+        // that would prevent the GlobalPartitionEndpointManager from ever
+        // being dropped.
+        self.background_task_manager.spawn(Box::pin(async move {
+            Self::initiate_circuit_breaker_failback_loop(weak_self, shutdown_token).await;
```
kundadebdatta: ### Background Task Lifecycle: Why `initialize_and_start_circuit_breaker_failback_background_refresh` Does Not Leak

The spawned circuit-breaker failback task (`self.background_task_manager.spawn(...)`) is designed to **terminate gracefully** when the `CosmosClient` (and its owning `GlobalPartitionEndpointManager`) is dropped. Two independent mechanisms guarantee this:

#### 1. Cancellation Token (cooperative shutdown)

The `BackgroundTaskManager` owns a `CancellationToken` (`Arc<AtomicBool>`). When the manager is dropped — which happens when the client is disposed — it sets the token to `true`. The loop checks this flag in **two** places:

- **Before** sleeping — so it exits immediately if shutdown happened between iterations.
- **After** waking up — so it exits if shutdown happened *during* the sleep.

This ensures the task reacts within at most one sleep interval.

#### 2. `Weak<Self>` (prevents reference cycle)

The task captures a `Weak<GlobalPartitionEndpointManager>`, **not** an `Arc`. This is critical because:

- If the task held an `Arc<Self>`, dropping the client would **not** drop the manager (the task's `Arc` would keep it alive forever → reference cycle → memory leak).
- With `Weak<Self>`, once all strong references (held by the client pipeline) are gone, `weak_self.upgrade()` returns `None` and the loop exits.

#### Why both mechanisms?

| Mechanism | Guards against |
|---|---|
| `shutdown_token` | Fast exit even while the `Weak` is still upgradeable (e.g., another thread briefly holds a strong ref) |
| `Weak<Self>` | Exit even if the `BackgroundTaskManager` drop runs out of order or the token check races with the sleep |

Together they ensure the background task **cannot outlive the client** and **cannot prevent the manager from being deallocated**.

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -30,19 +32,103 @@ impl ContainerConnection {
...
+                        // we release the borrow on routing_map before possibly moving
+                        // it into try_lookup for a refresh.
+                        match routing_map.get_range_by_effective_partition_key(&epk) {
+                            Ok(pk_range) => {
+                                cosmos_request.request_context.resolved_partition_key_range =
+                                    Some(pk_range.clone());
+                            }
+                            Err(_) => {
+                                // Refresh the routing map and retry.
+                                let refreshed_routing_map = self
```
tvaron3, line 113: What if it fails again? Shouldn't this be handled by metadata retry policies?

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
```diff
@@ -79,16 +81,21 @@ impl BackOffRetryHandler {
                 self.global_endpoint_manager.clone(),
             ))
         } else {
-            RetryPolicy::Client(ClientRetryPolicy::new(
+            RetryPolicy::Client(Box::from(ClientRetryPolicy::new(
```
tvaron3, line 85: Just curious why is this a box now ?

sdk/cosmos/azure_data_cosmos/src/models/partition_key_definition.rs
```diff
@@ -66,22 +67,10 @@ impl<S1: Into<String>, S2: Into<String>, S3: Into<String>> From<(S1, S2, S3)>
     }
 }
 
-/// Represents the kind of a partition key.
```
tvaron3, line 70: I think it's better to keep this one instead of the one in hash because this one is rightfully public api while the one in hash is private. I get confused with rust visibility modifiers so might be wrong.

sdk/cosmos/azure_data_cosmos/src/resource_context.rs
```diff
@@ -90,6 +100,19 @@ impl ResourceType {
                 | ResourceType::PartitionKeyRanges
         )
     }
+
+    /// Returns `true` if the resource type is partitioned.
+    pub fn is_partitioned(&self) -> bool {
+        matches!(
+            self,
+            ResourceType::Documents
```
tvaron3: just curious, do we need all these ? I don't know operations that use ResourceType::Attachment, ResourceType::PartitionedSystemDocument, ResourceType::RetriableWriteCachedResponse

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -436,9 +478,30 @@ impl ClientRetryPolicy {
             return Some(self.should_retry_on_session_not_available());
```
tvaron3, line 478: Shouldn't the retries for 404:1002 be modified as well to retry on the new write region for that partition? Will this be part of subsequent pr.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2446 @@
...
+use crate::cosmos_request::CosmosRequest;
+use crate::operation_context::OperationType;
+use crate::resource_context::ResourceType;
+use crate::routing::global_endpoint_manager::GlobalEndpointManager;
+use crate::routing::partition_key_range::PartitionKeyRange;
+use tracing::info;
+use url::Url;
+
+/// Represents the health status of a transport address.
+/// The numeric values indicate priority for replica selection (lower = healthier).
```
tvaron3: nit: 
```suggestion
/// The numeric values indicate priority for partition selection (lower = healthier).
```

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2446 @@
...
+    /// An `Arc<Self>` that can be shared across threads and async tasks.
+    pub fn new(
+        global_endpoint_manager: Arc<GlobalEndpointManager>,
+        partition_level_failover_enabled: bool,
+        partition_level_circuit_breaker_enabled: bool,
+    ) -> Arc<Self> {
+        let instance = Arc::new(Self {
+            global_endpoint_manager,
+            partition_unavailability_duration_secs:
+                Self::allowed_partition_unavailability_duration_secs(5),
```
tvaron3: Should the defaults be constants for easier discoverability? Also since there is only one default, we could just use it in the method instead of it being an arg

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2446 @@
...
+
+    /// Returns the stale partition unavailability refresh interval in seconds.
+    ///
+    /// This determines how frequently the background failback loop runs to check
+    /// whether previously failed partitions can be restored to healthy status.
+    /// Reads from the `AZURE_COSMOS_PPCB_STALE_PARTITION_UNAVAILABILITY_REFRESH_INTERVAL_IN_SECONDS`
+    /// environment variable, falling back to `default` if the variable is unset or not parseable.
+    fn stale_partition_unavailability_refresh_interval_secs(default: i64) -> i64 {
+        std::env::var(
+            "AZURE_COSMOS_PPCB_STALE_PARTITION_UNAVAILABILITY_REFRESH_INTERVAL_IN_SECONDS",
```
tvaron3, line 160: Since the background task will always shift back traffic every 5 minutes to the partition regardless of health, I think this should exponentially increase with some long max.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2446 @@
...
+    /// original (previously failed) endpoints.
+    ///
+    /// On each iteration the loop sleeps for [`background_connection_init_interval_secs`]
+    /// seconds and then calls [`initiate_failback_to_unhealthy_endpoints`]. Any errors
+    /// during the failback attempt are logged but do not terminate the loop.
+    ///
+    /// The loop exits when either:
+    /// - `shutdown_token` is set to `true` (the owning [`BackgroundTaskManager`] was dropped), or
+    /// - `weak_self.upgrade()` returns `None` (all strong `Arc` references are gone).
+    #[allow(dead_code)]
```
tvaron3: Should this method be dead code ?

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2446 @@
...
+
+    /// The endpoint URI that this partition key range is currently routed to.
+    pub current: String,
+
+    /// The endpoint URI that originally failed, triggering the failover chain.
+    /// Used by the failback loop to know which endpoint to restore.
+    pub first_failed_location: String,
+
+    /// The collection resource ID (RID) associated with this partition key range.
+    pub collection_rid: String,
```
tvaron3, line 859: There could be same partition key range for different collections. Shouldn't the mapping be (partition key range, collection rid) -> failoverinfo.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2446 @@
...
+    pub fn new(collection_rid: String, current_location: String) -> Self {
+        Self {
+            collection_rid,
+            current: current_location.clone(),
+            first_failed_location: current_location,
+            failed_locations: Mutex::new(HashMap::new()),
+            consecutive_read_request_failure_count: AtomicI32::new(0),
+            consecutive_write_request_failure_count: AtomicI32::new(0),
+            read_request_failure_counter_threshold:
+                Self::circuit_breaker_consecutive_failure_count_for_reads(10),
```
tvaron3: Same question as above on the defaults. Also in the pr description it says that for reads the default is 2 jfyi

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2446 @@
...
+    }
+
+    /// Returns the timeout counter reset window in minutes for the circuit breaker.
+    ///
+    /// If the elapsed time between two consecutive failures exceeds this window,
+    /// the read and write failure counters are reset to zero. Reads from the
+    /// `AZURE_COSMOS_CIRCUIT_BREAKER_TIMEOUT_COUNTER_RESET_WINDOW_IN_MINUTES`
+    /// environment variable, falling back to `default` if unset or not parseable.
+    fn circuit_breaker_timeout_counter_reset_window_mins(default: i64) -> i64 {
+        std::env::var("AZURE_COSMOS_CIRCUIT_BREAKER_TIMEOUT_COUNTER_RESET_WINDOW_IN_MINUTES")
```
tvaron3, line 932: Why reset the counter if it is for consecutive errors? Wouldn't this prevent circuit breaker from helping sparse workloads

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -436,9 +478,30 @@ impl ClientRetryPolicy {
...
+            sub_status_code,
+        ) {
+            error!(
+                "Operation will NOT be retried on local region. \
+                     Treating SystemResourceUnavailable (429/3092) as ServiceUnavailable (503). \
+                     Status code: 429, sub status code: 3092"
+            );
+
+            return Some(
+                self.try_mark_endpoint_unavailable_for_pk_range_and_retry_on_service_unavailable(
```
tvaron3, line 492: Sorry for my learning why is this 429 handled differently for ppaf ? If there is a design document, please share it offline 😄

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2446 @@
...
+    ///
+    /// This value controls how long a partition must remain marked unavailable before
+    /// the background failback loop considers it eligible for health re-evaluation.
+    /// Reads from the `AZURE_COSMOS_ALLOWED_PARTITION_UNAVAILABILITY_DURATION_IN_SECONDS`
+    /// environment variable, falling back to `default` if the variable is unset or not parseable.
+    fn allowed_partition_unavailability_duration_secs(default: i64) -> i64 {
+        std::env::var("AZURE_COSMOS_ALLOWED_PARTITION_UNAVAILABILITY_DURATION_IN_SECONDS")
+            .ok()
+            .and_then(|v| v.parse().ok())
+            .unwrap_or(default)
```
analogrelay, line 149: Is there a reason these are env vars and not options passed in through `CosmosClientOptions`? I think we should avoid magic env vars unless absolutely necessary. I get that these are probably settings we don't want users to mess with but if that's the case we can put them in a sub-options type and document clearly that the defaults are best.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2446 @@
...
+    partition_level_automatic_failover_enabled: AtomicBool,
+
+    /// Flag to determine if partition level circuit breaker is enabled.
+    partition_level_circuit_breaker_enabled: AtomicBool,
+
+    /// Manages background tasks and signals them to stop when dropped.
+    background_task_manager: BackgroundTaskManager,
+}
+
+impl std::fmt::Debug for GlobalPartitionEndpointManager {
```
analogrelay: This is pretty much just the same implementation generated by `#[derive(Debug)]`. I don't think we need it explicitly here.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
kundadebdatta: The Box on retry_handler heap-allocates `ClientRetryPolicy` because it's significantly larger than the other two enum variants (`MetadataRequestRetryPolicy` and `ResourceThrottleRetryPolicy`).

Without Box, the `RetryPolicy` enum would be sized to fit its largest variant. `ClientRetryPolicy` holds an `Arc<GlobalEndpointManager>`, an `Arc<GlobalPartitionEndpointManager>`, multiple retry counters, optional `RetryContext`, and an optional `CosmosRequest` clone — making it much heavier than the other two variants. Every instance of `RetryPolicy`, regardless of which variant, would pay the memory cost of the largest one.

By boxing `ClientRetryPolicy`:

- The enum stays small — the Client variant is just a pointer (8 bytes on 64-bit), keeping RetryPolicy's stack size close to that of Metadata and ResourceThrottle.
- Metadata/control-plane requests don't waste stack space — they'd otherwise carry padding the size of ClientRetryPolicy they never use.
- Move operations are cheap — moving a Box is a pointer copy, whereas moving a large ClientRetryPolicy inline would copy all its fields.

In short: it's a standard Rust pattern to Box a disproportionately large enum variant so the other variants (and the enum as a whole) don't bloat in size.

sdk/cosmos/azure_data_cosmos/src/background_task_manager.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// of every loop iteration and return early when it is `true`.
+    pub fn cancellation_token(&self) -> Arc<AtomicBool> {
+        Arc::clone(&self.shutdown)
+    }
+
+    /// Returns `true` if the shutdown signal has been set.
+    #[allow(dead_code)]
+    pub fn is_shutdown(&self) -> bool {
+        self.shutdown.load(Ordering::SeqCst)
+    }
```
analogrelay: No dead code please, just delete it.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
kundadebdatta: The mapping is crrently `Arc<RwLock<HashMap<PartitionKeyRange, PartitionKeyRangeFailoverInfo>>>`, where the `PartitionKeyRange` is an object rather than a number or string. Therefore, it will be unique for a specific collection and a partition key range combination. So, we should be covered.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
kundadebdatta: Well, we deviate from the Java implementation here a bit. Think of it in this way, we want to fialover a partition, if there are `X` number of consecutive failures within `Y` sliding time-window. If that time period is over, then we would like to reset the counter to make sure we do not failover a partition for any transient connectivity issues.

Let me know if that make sense.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
kundadebdatta: Currently, we do not support partition level failovers for 404/1002 status-sub status code combination in the .NET SDK, since this is not a trigger for partition level failover. Do you know if python does it ?

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
kundadebdatta: For now, we do not have a way to identify is a partition is healthy from the GW. I will add a follow-up item to discuss how we can optimize it, and if we can use a exponentially  increasing value for this.

sdk/cosmos/azure_data_cosmos/src/handler/retry_handler.rs
tvaron3: Thank you for the explanation makes sense!

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
tvaron3: Yes we do here is the code for reference https://github.com/Azure/azure-sdk-for-python/blob/67b9d0f3812222ea234825dc87ad368437666912/sdk/cosmos/azure-cosmos/azure/cosmos/_session_retry_policy.py#L104. @simorenoh would have more context

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
tvaron3: Makes sense

sdk/cosmos/azure_data_cosmos/src/background_task_manager.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    /// Spawns a fire-and-forget background task on the async runtime.
+    ///
+    /// The returned [`SpawnedTask`] future is intentionally dropped — the task
+    /// runs independently until it completes or observes the shutdown signal.
+    pub fn spawn(&self, future: TaskFuture) {
+        drop(get_async_runtime().spawn(future));
```
analogrelay: Dropping the future returned by `.spawn` will cancel the task immediately. We _need_ to store it to keep it alive. We can then drop all the futures during shutdown.

sdk/cosmos/azure_data_cosmos/src/background_task_manager.rs
```diff
@@ -0,0 +1,128 @@
...
+
+/// Manages the lifecycle of background tasks spawned by a client.
+///
+/// Tasks spawned through this manager receive a shared cancellation token
+/// ([`AtomicBool`]) that they should check periodically. When the manager is
+/// dropped, the token is set to `true`, signaling all tasks to exit gracefully.
+#[derive(Debug)]
+pub(crate) struct BackgroundTaskManager {
+    /// Shared shutdown flag. `true` means "please stop".
+    shutdown: Arc<AtomicBool>,
```
analogrelay: I don't think we need this. If we hold on to all the futures in a `Vec`, dropping `BackgroundTaskManager` will drop the futures, and dropping the futures will cancel the tasks. `drop(future)` is _how_ Rust communicates cancellation to futures.

sdk/cosmos/azure_data_cosmos/src/hash.rs
```diff
@@ -0,0 +1,478 @@
...
+                panic!("Hash partitioning only supports version 1 or 2");
+            }
+        },
+        // hpk only supports V2
+        PartitionKeyKind::MultiHash => {
+            panic!("MultiHash currently not supported. Pending additional testing.");
+            // if version != 2 {
+            //     panic!("MultiHash partitioning only supports version 2");
+            // }
+            // get_effective_partition_key_for_multi_hash_partitioning_v2(pk_value)
```
analogrelay: Just remove the commented-out code.

sdk/cosmos/azure_data_cosmos/src/hash.rs
```diff
@@ -0,0 +1,478 @@
...
+        }
```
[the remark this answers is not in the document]
analogrelay: I don't think we need to remove the panic, but please open an issue tracking the work to remove it. Panics are very dangerous since they crash the entire process.

sdk/cosmos/azure_data_cosmos/src/hash.rs
```diff
@@ -0,0 +1,478 @@
...
+                }
+            }
+            _ => self.clone(),
+        }
+    }
+}
+
+/// Returns a hex string representation of a partition key value.
+pub fn get_hashed_partition_key_string(
+    pk_value: &[InnerPartitionKeyValue],
```
analogrelay: This can take a `&[&InnerPartitionKeyValue]` so that cloning isn't required. We don't need to own or modify the values, just read them to generate the hash.

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
```diff
@@ -88,6 +89,20 @@ impl PartitionKey {
...
+    ///
+    /// # Arguments
+    /// * `kind` - The partition key kind (Hash or MultiHash)
+    /// * `version` - The hash version (1 or 2)
+    ///
+    /// # Returns
+    /// A hex-encoded string representing the hashed partition key
+    pub fn get_hashed_partition_key_string(&self, kind: PartitionKeyKind, version: u8) -> String {
+        let inner_values: Vec<InnerPartitionKeyValue> =
+            self.0.iter().map(|v| v.0.clone()).collect();
```
analogrelay: When `get_hashed_partition_key_string` accepts `&[&InnerPartitionKeyValue]`, we can just make this into a `Vec<&InnerPartitionKeyValue>`

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
```diff
@@ -88,6 +89,20 @@ impl PartitionKey {
...
+
+    /// Returns a hex string representation of the partition key hash.
+    ///
+    /// # Arguments
+    /// * `kind` - The partition key kind (Hash or MultiHash)
+    /// * `version` - The hash version (1 or 2)
+    ///
+    /// # Returns
+    /// A hex-encoded string representing the hashed partition key
+    pub fn get_hashed_partition_key_string(&self, kind: PartitionKeyKind, version: u8) -> String {
```
analogrelay: Create an `EffectivePartitionKey` newtype to hold the string. We shouldn't be moving EPKs around as strings.

sdk/cosmos/azure_data_cosmos/src/resource_context.rs
```diff
@@ -51,14 +51,19 @@ impl LinkSegment {
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[allow(dead_code)] // For the variants. Can be removed when we have them all implemented.
```
analogrelay: Let's get rid of this and delete dead code. We can add new ResourceTypes _when_ we need them.

sdk/cosmos/azure_data_cosmos/src/models/partition_key_definition.rs
analogrelay: Yeah, this is the one exposed through the `PartitionKeyDefinition` on `ContainerProperties`, so it makes sense to live here IMO

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
analogrelay: Let's get rid of as much of the `#[allow(dead_code)]` attributes as possible. Dead code adds confusion and noise to the code base. These are almost all internal components, we don't need to pre-emptively bring in APIs and code

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
kundadebdatta: Discussed oflfine. We are good with the current retry mechanism.

sdk/cosmos/azure_data_cosmos/src/resource_context.rs
kundadebdatta: Removed the `dead_code` annotations.

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
kundadebdatta: Thanks! I have added a new type `EffectivePartitionKey` to capture the EPK.

sdk/cosmos/azure_data_cosmos/src/hash.rs
kundadebdatta: Refactored this.

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
kundadebdatta: Refactored.

sdk/cosmos/azure_data_cosmos/src/hash.rs
kundadebdatta: Removed.

sdk/cosmos/azure_data_cosmos/src/background_task_manager.rs
kundadebdatta: Refacotored with the latest commit. Please take a look and let me know your thoughts.

sdk/cosmos/azure_data_cosmos/src/background_task_manager.rs
kundadebdatta: Agreed. Refactored with the latest commit. Please take a look and let me know your thoughts.

sdk/cosmos/azure_data_cosmos/src/background_task_manager.rs
kundadebdatta: Removed.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
kundadebdatta: Refactoed. Thanks! I believe the change was motivated from a clippy warning.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
kundadebdatta: Discussed offline. This is a special condition where a System Resource Not Available sub-status (3092) is returned with a 429. We would need to failover the partition in such conditions.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
kundadebdatta: Moved the defaults to a file constant for better readability.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
kundadebdatta: Removed.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
kundadebdatta: Moved the defaults to a file constant for better readability.

sdk/cosmos/azure_data_cosmos/src/resource_context.rs
kundadebdatta: I have removed the ones which are not used at the moment. We can always come back in future and add as many as necessary.

sdk/cosmos/azure_data_cosmos/src/models/partition_key_definition.rs
kundadebdatta: Refactored this with the latest commit.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
kundadebdatta: Yes - This is more of a config than a public contract. And historically we have exposed the configs as environment variables so that we can change/ update them in the future.

Once we put these in client options, these becomes our contracts, which will be difficult to change in the future, if any.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -26,6 +27,7 @@ pub struct CosmosClient {
     databases_link: ResourceLink,
     pipeline: Arc<GatewayPipeline>,
     global_endpoint_manager: Arc<GlobalEndpointManager>,
+    global_partition_endpoint_manager: Arc<GlobalPartitionEndpointManager>,
```
FabianMeiswinkel, line 30: I don't particularly like adding GEM and GPEM in parallel - to me the logic for routing to a specific region should be encapsualted in one component in a new SDK - the reason why in some other SDKs they might exist in parallel is because we had to introduce PPAF carefully - avoiding possible regressions. This is not the case in Rust. That said - don't block the PR on this - like the pipeline, this can be refactored in the driver.

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -30,19 +32,103 @@ impl ContainerConnection {
...
             container_cache,
             pk_range_cache,
+            global_partition_endpoint_manager,
         }
     }
 
     pub async fn send<T>(
         &self,
-        cosmos_request: CosmosRequest,
+        mut cosmos_request: CosmosRequest,
```
FabianMeiswinkel, line 48: Just my 2 cents - I really think the state management in the pipeline needs to be cleaned-up. The mut you add here means the send method takes ownership of the CosmosRequest - since sending (at least logically the same) CosmosRequest to multiple ContainerConnection instances in parallel (hedging, retries) is possible it requires some cloning upstream. There should be a clear contract of what parts of a CosmosRequest are immutable (all that form the intend) vs. some internal routing state - and for that the scope needs to be clear - is it relevant/mutable only for the lifecycle of an actual hhtp request - or for the entire operation etc. Again, no need to block teh PR on this IMO - but we have to refactor/fix this fast - otherwise we will end-up in the same mess like in Java/-Net quickly.

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -30,19 +32,103 @@ impl ContainerConnection {
...
+                } else if let Some(partition_key) = cosmos_request.partition_key.as_ref() {
+                    let routing_map = self
+                        .pk_range_cache
+                        .try_lookup(&container_prop.id, None)
+                        .await?;
+
+                    if let Some(routing_map) = routing_map {
+                        // Use a safe default version (2) when the service omits the version field,
+                        // since get_hashed_partition_key_string only supports version 1 or 2.
+                        let pk_version = pk_def.version.unwrap_or(2) as u8;
```
FabianMeiswinkel, line 98: Instead of u8 - why not use newtype for pkdefversion

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -30,19 +32,103 @@ impl ContainerConnection {
...
+                            &pk_range.collection_rid,
+                            &pk_range.partition_key_range_id,
+                            false,
+                        )
+                        .await
+                    {
+                        cosmos_request.request_context.resolved_partition_key_range =
+                            Some(resolved.clone());
+                    }
+                } else if let Some(partition_key) = cosmos_request.partition_key.as_ref() {
```
FabianMeiswinkel, line 89: i think there is a case missing where cosmos_request only contains feedrange? Possibly this is not available in Rust? If so, let me know and I will file an issue to track.

sdk/cosmos/azure_data_cosmos/src/models/partition_key_definition.rs
```diff
@@ -4,6 +4,19 @@
 use azure_core::fmt::SafeDebug;
 use serde::{Deserialize, Serialize};
 
+/// Represents the partition key kind.
+#[derive(PartialEq, Eq, Clone, Default, SafeDebug, Deserialize, Serialize)]
+pub enum PartitionKeyKind {
```
FabianMeiswinkel: EVERY enum in public surface area should use #[non_exhaustive]

sdk/cosmos/azure_data_cosmos/src/models/partition_key_definition.rs
```diff
@@ -4,6 +4,19 @@
...
+    /// Standard single-path hashing.
+    #[default]
+    Hash,
+    /// Multi-path (hierarchical) partition keys.
+    MultiHash,
+    /// A partition key kind not recognized by this SDK version.
+    #[serde(other)]
+    Other,
```
FabianMeiswinkel: Why? We shoudl rather fail hard? Nothing the SDK could do in that case?

FabianMeiswinkel: If the goal is to unify not understood values when deserializing service response to be able to react to it with a proper error vs. failing on deserialization i get the intend - but this shoud never leak into the public surface area.

FabianMeiswinkel: So, if serder-other is needed - we need to different contarcts - just one pub(create)/internal for the serialization contract and one external to have a proper public API

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2422 @@
...
+use crate::operation_context::OperationType;
+use crate::resource_context::ResourceType;
+use crate::routing::global_endpoint_manager::GlobalEndpointManager;
+use crate::routing::partition_key_range::PartitionKeyRange;
+use tracing::info;
+use url::Url;
+
+/// Default duration (in seconds) a partition must remain marked unavailable before
+/// the background failback loop considers it eligible for health re-evaluation.
+const DEFAULT_ALLOWED_PARTITION_UNAVAILABILITY_DURATION_SECS: i64 = 5;
```
FabianMeiswinkel, line 22: env vars with dedicated units is fine - in code we should always normalize on duration - even in consts

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2422 @@
...
+
+/// Default window (in minutes) after which the circuit breaker resets its failure
+/// counters if no new failure has been recorded.
+const DEFAULT_CIRCUIT_BREAKER_TIMEOUT_COUNTER_RESET_WINDOW_MINS: i64 = 5;
+
+/// Represents the health status of a transport address.
+/// The numeric values indicate priority for partition selection (lower = healthier).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[repr(i32)]
+pub enum PartitionHealthStatus {
```
FabianMeiswinkel, line 42: pub(crate)?

If really pub (would surprise me) add #[non_exhaustive]

FabianMeiswinkel: Never mind - routing module is not part of public surface area. I prefer being specific with visibility on types to not rely on module's visibility (to avoid this confusion in code reviews for example) - but can be ignored for this PR - I have explained that in the cosmso-design-struct skill PR.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
FabianMeiswinkel: My driver PR addresses this - alls ettings can be overriden via API in Driver - only defauklt values can be specified in env variables. I would not spend too much time on this in this PR - will change anyway when moving to driver.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2422 @@
...
+    ///
+    /// This is a non-deterministic health restoration: once the unavailability window
+    /// has elapsed, the endpoint is optimistically assumed to be healthy again.
+    /// The actual verification happens when subsequent requests are routed back to
+    /// the original endpoint.
+    ///
+    /// # Arguments
+    /// * `pk_range_uri_mappings` - Mutable reference to the map of partition key ranges
+    ///   to their `(collection_rid, original_failed_location, health_status)` tuples.
+    fn mark_endpoints_to_healthy(
```
FabianMeiswinkel, line 322: why? didn't we say this should never be done?

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2422 @@
...
+    ///
+    /// This path handles failover of reads (any account type) and writes on multi-master
+    /// accounts to alternate preferred read regions.
+    ///
+    /// # Arguments
+    /// * `request` - The Cosmos request to evaluate.
+    ///
+    /// # Returns
+    /// `true` if the request is eligible for partition-level circuit breaker, `false` otherwise.
+    pub fn is_request_eligible_for_partition_level_circuit_breaker(
```
FabianMeiswinkel, line 381: shouldn't this filter to resourceType == Document (and StoredProc + Execute) only?

Actually master partition could be used for circuite breaker as well - but would probably need special handling?

Same for tail partition - anyway - would like to see the handling of metadata requests be spelled out here more granularly.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2422 @@
...
+                        )))
+    }
+
+    /// Validates whether the given request is eligible for any form of partition-level failover.
+    ///
+    /// Performs the following checks:
+    /// 1. At least one partition-level failover mode (automatic failover or circuit breaker)
+    ///    must be enabled.
+    /// 2. The request must target a resource type that supports partition-level failover
+    ///    (see [`can_use_partition_level_failover_locations`]).
```
FabianMeiswinkel, line 406: dito - metadata and tail partition / master partition handling?

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
```diff
@@ -0,0 +1,2422 @@
...
+                    .can_circuit_breaker_trigger_partition_failover(request.is_read_only_request())
+            {
+                return false;
+            }
+
+            let triggered_by = if self
+                .partition_level_automatic_failover_enabled
+                .load(Ordering::SeqCst)
+            {
+                "Automatic Failover"
```
FabianMeiswinkel, line 510:
```suggestion
                // TODO @fabianm - Move this to newtype and capture in DiagnosticsContext when porting to driver
                "Automatic Failover"
```

sdk/cosmos/azure_data_cosmos/src/cosmos_request.rs
```diff
@@ -105,6 +101,16 @@ impl CosmosRequest {
         self.operation_type.is_read_only()
     }
 
+    /// Returns the database ID extracted from the request's resource link, if present.
```
FabianMeiswinkel: again - not blocking on this PR - but i shudder when reading the "if present" suffix. This has caused so many subtle and hard to debug bugs in .Net and Java - my goal is to eliminate them by not making RID optional - anyway - topic for the driver PR.

sdk/cosmos/azure_data_cosmos/src/murmur_hash.rs
```diff
@@ -0,0 +1,283 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+//
+// Implementation of the MurmurHash3 128-bit hash function.
```
FabianMeiswinkel: No - at best a variation of MurmurHash - comments should spell this out explicity to avoid confusion

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
```diff
@@ -88,6 +90,23 @@ impl PartitionKey {
...
+    /// * `kind` - The partition key kind (Hash or MultiHash)
+    /// * `version` - The hash version (1 or 2)
+    ///
+    /// # Returns
+    /// An `EffectivePartitionKey` representing the hashed partition key
+    pub fn get_hashed_partition_key_string(
+        &self,
+        kind: PartitionKeyKind,
+        version: u8,
+    ) -> EffectivePartitionKey {
```
FabianMeiswinkel, line 128: Either here or for EffectivePartitionKey - I woudl like to see a few dozen test cases of EPK values for PK-values across hash v1 and v2, HPK (including partial pks with the new normailzation on lenght .net is doing etc.) - run the pkvalue to EPK in .net and collect the input/output and run test cases here in CI pipeline to ensure we are consistent?

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
```diff
@@ -161,17 +181,9 @@ impl AsHeaders for PartitionKey {
 /// Represents a value for a single partition key.
 ///
 /// You shouldn't need to construct this type directly. The various implementations of [`Into<PartitionKey>`] will handle it for you.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq)]
```
FabianMeiswinkel: Why remoe Eq? I assume because of f64 (handling for NaN and +0 and -0 are different? In dirver I created custom handling to still allow using Eq - maybe worht taking a look at.

sdk/cosmos/azure_data_cosmos/src/request_context.rs
```diff
@@ -2,26 +2,20 @@
...
-pub struct PartitionKeyRange;
-
 /// Carries per-request routing, partition resolution, retry and regional state.
 ///
 /// `RequestContext` is mutated during pipeline execution to track which
 /// endpoints have failed, how the request should be routed (by region index
 /// or explicit endpoint), resolved partition ranges, session tokens, and
 /// various internal flags influencing retries and cache refresh behavior.
 #[allow(dead_code)]
-#[derive(Clone, Debug, PartialEq, Eq, Default)]
```
FabianMeiswinkel, line 24: why remove Eq

sdk/cosmos/azure_data_cosmos/src/resource_context.rs
```diff
@@ -234,6 +238,34 @@ impl ResourceLink {
             .join(&self.path())
             .expect("ResourceLink should always be url-safe")
     }
+
+    /// Extracts the database ID from the resource link path.
+    ///
+    /// The resource link path follows the pattern `dbs/{db_id}/...`.
+    /// Returns `None` if the path does not contain a database segment.
+    pub fn database_id(&self) -> Option<String> {
```
FabianMeiswinkel: Not worth spending much effor on in this PR - but take a look at the driver PR where I try to normalize all the ResourceReference handling and make it more predictable.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -95,6 +95,8 @@ impl CosmosClient {
                 options.enable_partition_level_circuit_breaker,
             );
 
+        #[allow(clippy::arc_with_non_send_sync)]
+        // On wasm32 SpawnedTask is !Send; Arc is still correct.
```
analogrelay: You can specify a `reason` in the `allow` itself:

```suggestion
        #[allow(clippy::arc_with_non_send_sync, reason = "Wasm32 doesn't include Send, but it's also single-threaded so it's fine")]
```

Also, slight suggestion to reword the message

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -181,6 +183,8 @@ impl CosmosClient {
             options.enable_partition_level_circuit_breaker,
         );
 
+        #[allow(clippy::arc_with_non_send_sync)]
+        // On wasm32 SpawnedTask is !Send; Arc is still correct.
```
analogrelay:
```suggestion
        #[allow(clippy::arc_with_non_send_sync, reason = "Wasm32 doesn't include Send, but it's also single-threaded so it's fine")]
```

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -182,6 +182,7 @@ mod tests {
         let partition_manager =
             GlobalPartitionEndpointManager::new(endpoint_manager.clone(), false, false);
         (
+            #[allow(clippy::arc_with_non_send_sync)]
```
analogrelay:
```suggestion
        #[allow(clippy::arc_with_non_send_sync, reason = "Wasm32 doesn't include Send, but it's also single-threaded so it's fine")]
```

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -265,6 +266,7 @@ mod tests {
         let partition_manager =
             GlobalPartitionEndpointManager::new(endpoint_manager.clone(), false, false);
 
+        #[allow(clippy::arc_with_non_send_sync)]
```
analogrelay:
```suggestion
        #[allow(clippy::arc_with_non_send_sync, reason = "Wasm32 doesn't include Send, but it's also single-threaded so it's fine")]
```

sdk/cosmos/azure_data_cosmos/src/routing/container_cache.rs
```diff
@@ -174,6 +174,7 @@ mod tests {
         let endpoint = Url::parse("https://test.documents.azure.com").unwrap();
         let partition_manager =
             GlobalPartitionEndpointManager::new(endpoint_manager.clone(), false, false);
+        #[allow(clippy::arc_with_non_send_sync)]
```
analogrelay:
```suggestion
        #[allow(clippy::arc_with_non_send_sync, reason = "Wasm32 doesn't include Send, but it's also single-threaded so it's fine")]
```

sdk/cosmos/azure_data_cosmos/src/models/partition_key_definition.rs
analogrelay: Another option here is to use a string newtype instead of an enum and only define the known values as constants. Internally, we can inspect the string and if it's not one of the known values we can produce an error.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
analogrelay: > Once we put these in client options, these becomes our contracts, which will be difficult to change in the future, if any.

I would just say that once we have magic env vars we may give out to customers in certain places, we have just as much of a contract as a public field on an options type. Maybe there's history for revoking these in the Cosmos SDK, but I've never seen them be treated as anything other than public APIs in things like .NET.

Another option is `#[doc(hidden)]` fields, which hide them from docs and intellisense but the compiler will allow them. In general, I'd prefer we have magic _fields_ to tweak things if we really think we need to expose an API to tweak. We have a lot more control over fields.

A third option is some internal "quirks" flags concept that is gated behind a feature flag (`CosmosClientBuilder::new(...).with_quirk(CosmosQuirks::ALLOWED_PARTITION_UNAVAILABILITY, Duration::from_secs(...))`), where both `with_quirk` and `CosmosQuirks` are hidden behind a `quirks` feature flag.

sdk/cosmos/azure_data_cosmos/src/hash.rs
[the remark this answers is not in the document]
analogrelay: This isn't correct anymore, but I'll address that in the sync this morning.

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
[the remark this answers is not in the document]
analogrelay: I think an "Undefined" partition key is usually serialized as `{}` in the partition key header

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
kundadebdatta: Make sense. We can get this refactored in the Driver.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
kundadebdatta: The long term goal is to have an API at the GW that can hint us if a partition is healthy. For now, TBH there is no definitive way of identifying if a partition become healthy, hence trying to mark the endpoints healthy (only for PPCB usecase though) is the failback mechanism for now.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/routing/container_cache.rs
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos/src/models/partition_key_definition.rs
kundadebdatta: Refactored the code to use String type.

analogrelay (approved):

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
kundadebdatta: I will add more tests with the upcoming follow-up PR on PPAF enablement. Hence, resolving this comment for now.

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
kundadebdatta: Thanks @FabianMeiswinkel! I will mark this as a follow up for the driver implementation.

sdk/cosmos/azure_data_cosmos/src/murmur_hash.rs
kundadebdatta: Updated the verbiage.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
kundadebdatta: Thanks! Yes, I agree the check should filter `resourceType == Document (and StoredProc + Execute)` as of now.

Let's keep the master partition handling as a follow up task.

sdk/cosmos/azure_data_cosmos/src/routing/global_partition_endpoint_manager.rs
kundadebdatta: Let's keep the master partition handling as a follow up task. I will create a separate item to track this.

sdk/cosmos/azure_data_cosmos/src/request_context.rs
kundadebdatta: I implemented the `Eq` manually to support `f16`. This is now reverted.

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
kundadebdatta: I implemented the `Eq` manually to support `f16`.

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
kundadebdatta: Will track this as my next follow up PR.

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
kundadebdatta: Thanks! Let me follow up on this and get back.

PR #3661, opened 2026-02-03
[Storage] Removing thick convenience layer for `azure_storage_blob` crate

vincenttran-msft:
This PR aims to strip away the thick convenience layer on our `azure_storage_blob` crate. This includes:

- Removing handwritten implementations that were just direct pass-throughs
- Renaming at the TypeSpec definition level, clients that had handwritten implementations just as a means of renaming
- Removes the handwritten client abstraction concept entirely -- all handwritten functionality is now directly `impl` onto the underlying generated client

.`tsp:` https://github.com/Azure/azure-rest-api-specs/pull/40134

Also temporarily expanding codeowners for EventHubs to avoid any accidental blockages (due to their hard dependency on us, we sometimes have to fix-up EventHubs code)

### DIFF
--- a/eng/tools/generate_api_report/Cargo.toml
+++ b/eng/tools/generate_api_report/Cargo.toml
@@ -11,7 +11,7 @@ serde = { version = "1.0", features = ["derive"] }
 toml = "0.8.20"
 
 [dependencies]
-rustdoc-types = "0.57.0"
+rustdoc-types = "0.33.0"
 serde_json = "1.0"
 toml = "0.8.20"
 
--- a/eng/tools/generate_api_report/rust-toolchain.toml
+++ b/eng/tools/generate_api_report/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "nightly-2026-01-15"
+channel = "nightly-2025-01-12"
 components = [
   "cargo",
   "rust-docs",


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -149,66 +126,43 @@ impl BlobClient {
...
+            version: self.version.clone(),
+            tracer: self.tracer.clone(),
         }
     }
 
+    // TODO: Can we just rename endpoint() on generated to this? Applies widely.
     /// Gets the URL of the resource this client is configured for.
     pub fn url(&self) -> &Url {
-        &self.client.endpoint
+        &self.endpoint
```
vincenttran-msft, line 216: @jhendrixMSFT Joel, do you know if there is any way we can control the name of the `endpoint()` function that is currently generated:
```rust
/// Returns the Url associated with this client.
    pub fn endpoint(&self) -> &Url {
        &self.endpoint
    }
```

sdk/storage/azure_storage_blob/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 - Renamed `BlobItemInternal` to `BlobItem`.
 - Renamed `BlobPropertiesInternal` to `BlobProperties`.
+- Changed `BlobClient`'s `set_metadata` parameter `metadata` type from `HashMap<String, String>` to `&HashMap<String, String>`.
```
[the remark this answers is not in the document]
vincenttran-msft: Top portion of the comment is valid.
For the internal renames:

- For the first two, that is true, but we will be removing these. This is temporary in order to open up the method name for our handwritten implementation.
- As for swapping implementations, that is incorrect. Both this and VS Code Copilot both thought that was the case, but what actually happened was `BlockBlobClient::upload()` was renamed to `BlockBlobClient::upload_internal()` (which is part of the effort in the above bullet point)

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
jhendrixMSFT: There isn't.  Note that the name is part of the [design guidelines](https://azure.github.io/azure-sdk/rust_introduction.html).

We could likely use the upcoming `@clientOption` decorator if necessary.  However, that's @heaths call if we want to support it or not.

jalauzon-msft: The name `endpoint` does not really make sense for Storage as it contains more than just the Storage endpoint, it contains the full URL to the particular resource and can even include the SAS query params as well. I think the Storage setup may be a bit unique here, but I think calling this `endpoint` would confuse customers.

heaths: We had talked about this before and I'm not against renaming it, but I also question if and where we actually define what an "endpoint" is. In the REST API sense, an "endpoint" can include at least the path, so it's not just the host. It's already a pretty vague term such that I think we need a compelling reason to rename it. But let's address that in a separate PR regardless and not hold this one up on it.

.github/CODEOWNERS
```diff
@@ -74,7 +74,7 @@
 # AzureSDKOwners: @vincenttran-msft @jalauzon-msft
 
 # PRLabel: %Event Hubs
-/sdk/eventhubs/          @j7nw4r
+/sdk/eventhubs/          @j7nw4r @heaths @RickWinter @LarryOsterman @vincenttran-msft @jalauzon-msft @ronniegeraghty
```
heaths: I don't think 1) we all need to be on this line, and 2) some of us need to be on storage-related things anyway. I'd actually recommend a more specific line like so:


```suggestion
/sdk/eventhubs/          @j7nw4r

% PRLabel: %Event hubs %Storage
/sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/          @j7nw4r @vincenttran-msft @jalauzon-msft
```

How this pairs with `ServiceLabel` and `AzureSDKOwners` below or above it, I'm not sure where it should go. /cc @weshaggard @hallipr @danieljurek

sdk/storage/azure_storage_blob/src/models/mod.rs
```diff
@@ -6,72 +6,16 @@ pub mod error;
 mod extensions;
 pub mod method_options;
 
+pub use crate::generated::models::*;
```
heaths, line 11: Yay! That said, since *this* module is now the module people see in docs, add a module (outer) comment after the copyright. Just copying the ones from the `mod.rs` you're importing, something like:

```rust
//! Model types for this crate
```

Or whatever we put in the generated code. Same for `clients`.

sdk/storage/azure_storage_blob/src/models/mod.rs
```diff
@@ -6,72 +6,16 @@ pub mod error;
...
-    StorageErrorCode, StorageServiceStats,
+    BlobClientDownloadInternalOptions as BlobClientDownloadOptions,
+    BlobClientDownloadInternalResult as BlobClientDownloadResult,
+    BlobClientDownloadInternalResultHeaders as BlobClientDownloadResultHeaders,
+    BlockBlobClientUploadInternalOptions as BlockBlobClientUploadOptions,
+    BlockBlobClientUploadInternalResult as BlockBlobClientUploadResult,
+    BlockBlobClientUploadInternalResultHeaders as BlockBlobClientUploadResultHeaders,
 };
 
 pub use error::{Result, StorageError};
```
heaths: Why `pub use`? The `error` module should be public from the root and you can re-export `Result` and `StorageError`. You can `use` it anywhere, but it shouldn't be a public re-export.

sdk/storage/azure_storage_blob/CHANGELOG.md
```diff
@@ -10,6 +10,14 @@
 
 - Renamed `BlobItemInternal` to `BlobItem`.
 - Renamed `BlobPropertiesInternal` to `BlobProperties`.
+- Renamed `PageBlobClient::upload_page()` to `upload_pages()`.
```
heaths, line 17: A *lot* of methods also changed, but I appreciate listing them all is 1) daunting to write, and 2) daunting to read such that no one probably would. But you might want to mention at least that a lot of method names were "simplified" or "shortened" with an example like `BlobContainerClient::create_container` to just `BlobContainerClient::create` or something like that.

heaths (approved):
Overall, nice! Just a couple highly suggested changes.

.github/CODEOWNERS
danieljurek: That format is correct in this context. Less specific paths to more specific paths... (The Service Label stuff is a matter of aesthetics and "findability"... that stuff is going to change soon)

.github/CODEOWNERS
```diff
@@ -74,7 +74,10 @@
 # AzureSDKOwners: @vincenttran-msft @jalauzon-msft
 
 # PRLabel: %Event Hubs
-/sdk/eventhubs/          @j7nw4r @heaths @RickWinter @LarryOsterman @vincenttran-msft @jalauzon-msft @ronniegeraghty
+/sdk/eventhubs/          @j7nw4r
+
+% PRLabel: %Event hubs %Storage
```
heaths: Oops, that was supposed to be `# PRLabel: %Eent hubs %Storage`. Sorry.

sdk/storage/azure_storage_blob/CHANGELOG.md
heaths: I didn't see a CHANGELOG.md update, which I assumed I should given your thumbs up.

vincenttran-msft: Good catch. I decided to just exhaustively list (since we were only 2 APIs off). Daunting to read? Agreed! But I want to be explicit especially since we aren't still at the final cut.

heaths: Oh, nice! I saw one or two instances and just figured that "class" of change was pervasive. Yeah, this is simpler and more searchable.

sdk/storage/azure_storage_blob/src/clients/append_blob_client.rs
jalauzon-msft: nit: `new` should probably be at the top of the struct.

sdk/storage/azure_storage_blob/src/clients/blob_service_client.rs
jalauzon-msft: This is not part of this PR but maybe we could fix it: We probably don't need a `from_url` on the service client.

jalauzon-msft, line 100: Consider for another PR: Maybe we should offer `blob_client` here to go straight to a Blob.

vincenttran-msft: Addressed in this PR!

vincenttran-msft: Done here!

jalauzon-msft (approved):

.github/CODEOWNERS
```diff
@@ -76,6 +76,9 @@
 # PRLabel: %Event Hubs
 /sdk/eventhubs/          @j7nw4r
 
+# PRLabel: %Event Hubs %Storage
+/sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/          @j7nw4r @vincenttran-msft @jalauzon-msft
```
danieljurek: @vincenttran-msft -- I created a merge conflict for you with my recent change to `main`. To get your file consistent with main just checkout the current state of `main` for `.github/CODEOWNERS` and then put this subtly different version of your change in around line 66: 

```codeowners

# PRLabel: %Event Hubs %Storage
/sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/    @j7nw4r @jalauzon-msft @vincenttran-msft
```

My changes to CODEOWNERS altered the sorting of labels, aliases, and adjusted spacing a little bit. 

Your sorting of this entry directly under `sdk/eventhubs/` is correct.

vincenttran-msft: Will do!

PR #3662, opened 2026-02-03
Add --all-features to cargo test guidance

Pilchie:
Update the Test Generation section in copilot-instructions.md to recommend using `--doc` and `--all-features` flags when running `cargo test` to ensure no tests are missed.

This ensures both doc tests and feature-gated tests are included when validating code changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -121,9 +121,11 @@ cargo build --workspace
 
 ### Testing
 
+When running `cargo test`, use `--all-features` to ensure no tests are missed.
+
 ```bash
 # Run tests for a specific crate
-cargo test -p <crate-name>
+cargo test -p <crate-name> --all-features
 
 # Run integration tests with recordings
 cargo test -p <crate-name> --test <test-name>


.github/copilot-instructions.md
```diff
@@ -43,6 +43,8 @@ Use these instructions for test generation as well.
 
 ## Test Generation
 
+-   When running `cargo test`, make sure to use the `--all-features` flag to ensure no tests are missed. (e.g., `cargo test -p <crate-name> --all-features`).
+-   When finished, be sure to check documentation works by running `cargo test -p <crate-name> --doc`.
```
heaths: These should go in `AGENTS.md`. The guidance given to us was that the project structure and other technical notes e.g., commands should be in `AGENTS.md` while `.github/copilot-instructions.md` is more about style.

This might change at some point given evidence to the contrary we are experiencing, but we'll do this across repos.

If you look at `AGENTS.md` or even have Copilot add these to test invocations, you'll see the appropriate commands to update.

Pilchie: Moved.

PR #3688, opened 2026-02-06
Tell dependabot to ignore npm packages in eng/common

danieljurek:
Example noisy PRs that came up after enabling dependabot in the repo: 

* https://github.com/Azure/azure-sdk-for-rust/pull/3684
* https://github.com/Azure/azure-sdk-for-rust/pull/3685

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,17 @@
 version: 2
 updates:
-- package-ecosystem: "npm"
-  directory: "/"
-  schedule:
-    interval: weekly
-  exclude-paths:
-    - "eng/common"
+  - package-ecosystem: npm
+    directory: "/"
+    schedule:
+      interval: weekly
+    exclude-paths:
+      - "eng/common"
+
+  - package-ecosystem: cargo
+    directory: "/"
+    # Check for updates every week on Monday at 4:00 AM Pacific time
+    schedule:
+      interval: weekly
+      day: monday
+      time: "04:00"
+      timezone: "US/Pacific"


.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+- package-ecosystem: "npm"
```
heaths: This seems wrong. We're not an `npm` package system and have no `package.json` in our root. But if you're supplying explicit configuration, this needs to be correctly configured for Rust / cargo.

danieljurek: This disables NPM scanning in `eng/common` which generates noise PRs that will be handled by the central workflow. Since we discussed a specific timing, I can add in cargo with that timing.

heaths: Maybe I'm misremembering, but I thought as soon as you introduce one package ecosystem in `dependabot.yml`, you're basically saying: "this is all I have; don't check other stuff". If not, great. That was my fear.

Regardless, we should probably add the cargo ecosystem in there just to be safe, even if we're just specifying default. Perhaps I'm just not as trusting against possible regressions.

heaths (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+- package-ecosystem: "npm"
+  directory: "/"
+  schedule:
+    interval: weekly
+  exclude-paths:
+    - "eng/common"
```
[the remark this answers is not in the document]
danieljurek: This is what I get for using github.dev in the browser instead of going through all the ceremony locally. Good catch if true, Copilot.

.github/dependabot.yml
```diff
@@ -1,8 +1,17 @@
 version: 2
 updates:
-- package-ecosystem: "npm"
-  directory: "/"
-  schedule:
-    interval: weekly
-  exclude-paths:
-    - "eng/common"
+  - package-ecosystem: npm
+    directory: "/"
```
heaths, line 4: And you're sure this won't be an issue? I mean, it's not going to expect `package.json`s anywhere else?

heaths (approved):
Assume my concern isn't valid, LGTM.

.github/dependabot.yml
danieljurek: This setting is intended to narrowly exclude the known sources of noise. 

I think dependabot should open PRs if it finds relevant package.json files elsewhere in the repository. The purpose of this is to specifically exclude `eng/common`. Such package.json files would be worth discussing in such a circumstance.

.github/dependabot.yml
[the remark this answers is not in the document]
danieljurek: Weird... the comment I thought I was responding to was one about formatting...

PR #3689, opened 2026-02-06
Downgrade reqwest

heaths:
Related to #3867 and effectively reverts #3620, but I at least upgraded to the current 0.12.x version.
Ran all tests. We'll have to revisit once a new version of `reqwest` is released with https://github.com/seanmonstar/reqwest/pull/2927 merged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -166,12 +166,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atomic-waker"
-version = "1.1.2"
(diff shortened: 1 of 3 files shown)


heaths:
@seanmonstar just shipped a new version of `reqwest` so I'm trying that out as well, but I first have to bump our MSRV to 1.88 because of the `time` crate update (seriously: who bumps their MSRV in a patch release and thinks that's a good idea?!) and that it didn't trigger `azure_core` tests at least. @danieljurek is looking into that.

Issue #3692, opened 2026-02-06
APIView support broken after 1.88 MSRV update


heaths:
Since we had to update the MSRV to 1.88 in #3690 the `generate_api_report` tool is broken because it can't run on MSRV 1.88 crates since it's effectively pinned to 1.86-nightly itself:

```text
Running: cargo +nightly-2025-01-12 rustdoc -Z unstable-options --output-format json --package azure_messaging_eventhubs --all-features
Failed to generate JSON file: error: rustc 1.86.0-nightly is not supported by the following packages:
  azure_core@0.32.0 requires rustc 1.88
  azure_core@0.32.0 requires rustc 1.88
  azure_core@0.32.0 requires rustc 1.88
  azure_core_macros@0.6.0 requires rustc 1.88
  time@0.3.47 requires rustc 1.88.0
  time-core@0.1.8 requires rustc 1.88.0
  time-core@0.1.8 requires rustc 1.88.0
  time-macros@0.2.27 requires rustc 1.88.0
  typespec@0.12.0 requires rustc 1.88
```

We'll need to update this, but I seem to recall the previous dev saying he had to pin the version because the JSON format we get backed changed so frequently, so that will be a concern we'll have to check for and fix as needed.

heaths:
Tracked down that `nightly-2025-05-09` was the last 1.88 nightly build (based on branch date from their `main` branch), but the tool fails with:

> Running: cargo +nightly-2025-05-09 rustdoc -Z unstable-options --output-format json --package typespec --all-features
> Error: Error("missing field `name`", line: 1, column: 774)
> Processing rustdoc output for package: typespec

Unfortunately, `--output-format` is still unstable: https://github.com/rust-lang/cargo/issues/13283. I was hoping if we're going to have to fix the tool, we could do it once.

PR #3693, opened 2026-02-06
Disable the APIView generation tool

heaths:
Temporarily works around #3692 until we can fix the tool.


RickWinter (approved):
Open a release blocking issues to put this back in.

danieljurek (approved):
Approving in the interest of unblocking. I'm not certain of the downstream affects to the release submission of API View. For best success, I recommend the release be run with `Skip.CreateApiReview` set to `true` ([TSG](https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/429/Troubleshooting-guide-EngSys-team?anchor=can-we-override-or-disable-release-check)) but it might still have failures after the crates have published. 

I think this change will otherwise unblock release.

PR #3698, opened 2026-02-07
[Storage] Align `azure_storage_queue` with Blobs + SAS support

vincenttran-msft:
`.tsp`: https://github.com/Azure/azure-rest-api-specs/pull/40244

### DIFF
--- a/sdk/storage/azure_storage_queue/src/clients/queue_service_client.rs
+++ b/sdk/storage/azure_storage_queue/src/clients/queue_service_client.rs
@@ -4,7 +4,10 @@
 use crate::{
     clients::QueueClient,
     generated::{
-        clients::{QueueServiceClient as GeneratedQueueServiceClient, QueueServiceClientOptions},
+        clients::{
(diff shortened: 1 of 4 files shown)


sdk/storage/azure_storage_queue/src/clients/queue_service_client.rs
```diff
@@ -4,39 +4,115 @@
...
+impl GeneratedQueueServiceClient {
+    /// Creates a new GeneratedQueueServiceClient from a service URL.
+    ///
+    /// # Arguments
+    ///
+    /// * `service_url` - The full URL of the Azure storage account, for example `https://myaccount.queue.core.windows.net/`
+    /// * `credential` - An optional implementation of [`TokenCredential`] that can provide an Entra ID token for authentication. If None, the URL must contain authentication information (e.g., SAS token).
+    /// * `options` - Optional configuration for the client.
+    #[tracing::new("Storage.Queues.Service")]
+    pub fn from_url(
```
jalauzon-msft, line 38: Like we decided for Blob, service client probably doesn't need a `from_url`

PR #3699, opened 2026-02-09
Ranged download

jaschrep-msft:
`managed_download()` now supports getting specific blob ranges.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -266,6 +266,7 @@ dependencies = [
  "azure_identity",
  "opentelemetry",
  "opentelemetry_sdk",
+ "reqwest",
  "tokio",
  "tracing",
  "tracing-subscriber",


sdk/storage/azure_storage_blob/src/models/method_options.rs
```diff
@@ -53,6 +53,10 @@ pub struct BlobClientManagedDownloadOptions<'a> {
     /// A default value will be chosen if none is provided.
     pub partition_size: Option<NonZero<usize>>,
 
+    /// Optional. Range of the blob to download.
+    /// None will result in the entire blob being downloaded.
```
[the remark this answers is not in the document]
vincenttran-msft: I think this is a good comment (the point about being clear with exact semantics)

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -37,17 +41,41 @@ where
...
-    let total_ranges = content_range.total_length().div_ceil(partition_size);
-
-    let mut ranges = (1..total_ranges).map(move |i| {
-        i * partition_size
-            ..min(
-                i * partition_size + partition_size,
-                content_range.total_length(),
-            )
-    });
+    let range = range.unwrap_or(0..u64::MAX);
```
vincenttran-msft: So I see that we now make it `Option<>` so that we can handle `None` case and map that to download-all. My question is- would it be better in that case to just omit range entirely? Or is it better to specify a needlessly large number and still do a "ranged download".

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -982,31 +980,49 @@ async fn test_managed_download_beneath_partition_size(
...
-        downloaded_data.put(bytes);
+        let mut downloaded_data = BytesMut::new();
+        while let Some(bytes) = download_stream.try_next().await? {
+            downloaded_data.put(bytes);
+        }
+        let downloaded_data = downloaded_data.freeze();
+        assert_eq!(
+            &downloaded_data,
+            match download_range {
+                Some(r) => &data[r.0 as usize..min(r.1 as usize, DATA_LEN)],
```
heaths: This is a bad DX. Sizes generally should be `usize` anyway, so making people convert to use them *always* is not going to be a good idea.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -337,9 +338,9 @@ impl<'a> BlobClientDownloadBehavior<'a> {
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 impl PartitionedDownloadBehavior for BlobClientDownloadBehavior<'_> {
-    async fn transfer_range(&self, range: Range<u64>) -> Result<AsyncRawResponse> {
+    async fn transfer_range(&self, range: Option<Range<u64>>) -> Result<AsyncRawResponse> {
         let mut opt = self.options.clone();
-        opt.range = Some(format!("bytes={}-{}", range.start, range.end - 1));
+        opt.range = range.map(|r| format!("bytes={}-{}", r.start, r.end - 1));
```
heaths: `Range<T>::end` already is exclusive, so why `- 1`?

From <https://doc.rust-lang.org/stable/std/ops/struct.Range.html>:

> A (half-open) range bounded inclusively below and exclusively above (`start..end`).
>
> The range `start..end` contains all values with `start <= x < end`. It is empty if `start >= end`.

sdk/storage/azure_storage_blob/src/models/content_range.rs
```diff
@@ -27,6 +27,7 @@ impl ContentRange {
         self.start
     }
 
+    /// Index of the final byte of the range (inclusive).
```
heaths: This is contrary to `std::ops::Range`. I think that's going to cause problems. Also, why not document the rest of the struct?

sdk/storage/azure_storage_blob/src/models/method_options.rs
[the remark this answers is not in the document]
heaths: Think of it like in .NET: first paragraph is the summary. And because it's markdown, there's no new paragraph until a blank line or other section. The subsequent paragraphs are the description until the next heading or code.

I would also write this more like a sentence: "Optional range of the blob to download".

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
heaths: Idiomatic rust would have you treat it as an `Option<T>` throughout instead of synthesizing values like this. But if it makes subsequent easier, I don't see any reason why you couldn't do this.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -37,17 +41,41 @@ where
...
+            _ => Err(err)?,
+        },
+    };
+
+    let mut ranges: VecDeque<_> = match initial_response
+        .headers()
+        .get_as::<ContentRange, _>(&"content-range".into())
+    {
+        Ok(content_range) => {
+            let remainder_start = content_range.end() + 1;
```
heaths: Along with my comment on `ContentRange`, just make this exclusive like ranges are in Rust so it's idiomatic. Start is inclusive, end is exclusive. Then it's consistent and devs don't end up missing a byte at the end because they used it wrong. You document it, sure, but we know not everyone reads that so it should feel natural.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -37,17 +41,41 @@ where
...
+            (Some(StatusCode::RequestedRangeNotSatisfiable), 0) => {
+                client.transfer_range(None).await?
+            }
+            _ => Err(err)?,
+        },
+    };
+
+    let mut ranges: VecDeque<_> = match initial_response
+        .headers()
+        .get_as::<ContentRange, _>(&"content-range".into())
```
heaths: You might consider just defining a header that implements `Header`. Simplifies this code quite a bit. We have examples in `typespec_client_core` and `azure_core` if you need.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -135,115 +173,186 @@ mod tests {
...
+                            raw_response: None,
+                        }
+                        .into_error());
+                    }
+                    let range = range.start..min(range.end, data_len as u64);
+                    headers.insert(
+                        "content-range",
+                        ContentRange::new(range.start, range.end - 1, self.data.len() as u64)
+                            .to_string(),
+                    );
```
heaths: And with  `ContentRange` implemented idiomatically and implementing `Header` (or have a separate type) you can just implement `From<>` and this code cleans up nicely.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -86,23 +114,28 @@ impl<T: Future + ConditionalSend> DownloadRangeFuture for T {}
...
     use tokio::{
         sync::Mutex,
         time::{sleep, Duration},
     };
 
     use super::*;
 
+    pub const KB: usize = 1024;
+    pub const MB: usize = KB * 1024;
+    pub const GB: usize = MB * 1024;
```
heaths, line 145: It's a little hard to completely see why you convert with `as usize` in lots of places above - which is a terrible DX to require if indeed that is the case - but if this is the only reason why, just define these as u64s. Same for some of your declarations below. Showing examples - even in tests - to devs of having to cast to `usize` all the time is part of that bad DX and will teach LLMs bad practices.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
jaschrep-msft: @vincenttran-msft it's not clear what you're asking. We never actually submit an unranged download to the network unless we hit the invalid range case.

This range gets clamped down again pretty fast. It's converted out of `Option<>` to avoid cluttering the range clamping code with a bunch of `unwrap_or_else` calls.

I can probably rename variables to clarify.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
jaschrep-msft: The HTTP Range header is inclusive on both ends.

sdk/storage/azure_storage_blob/src/models/content_range.rs
jaschrep-msft: This is a representation of the HTTP Content-Range header. It uses inclusive values. I can write a converter method between this struct and rust ranges that handles this, but I don't think I should misrepresent the model of the header value.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
jaschrep-msft: If i define these as u64s, then i just have to convert back to usizes in different places. Rust memory structures use usizes and azure core http-body structures use u64. This is code that moves http body content into memory. These conversions need to happen somewhere.

I spent a decent amount of time thinking about which one to use in each case, with plenty of revision back and forth to try and simplify the code. I even considered dual-defining some values in tests, though that got ugly fast.

I am interested in a broader solution that eliminates this problem, but I don't have one onhand.

sdk/storage/azure_storage_blob/src/models/content_range.rs
jaschrep-msft: hmm. actually on further thought I align with your suggestion.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
heaths: I get that, but that's not idiomatic to Rust. We can and probably should do the translation for them.

Imagine this: devs are used to this:

```rust
let arr = [0, 1, 2, 3, 4];
assert_eq!(arr[1..4], [1, 2, 3]);
```

Then with your API, they'd instead get back `[1, 2, 3, 4]`. Don't you think that will come as a bit of a surprise? Be idiomatic. That's the first guideline for Azure SDKs.

jaschrep-msft: This is us doing the translation for them. The provided range is exclusive. When translating it to the serialized header value, we are making it inclusive. This code does what you want.

I believe our options struct should generate as an `Option<Range<usize>>` and the generated code should do this translation for us, but that is going to require a PR to our typespec.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
jaschrep-msft: discussed offline. converting everything to `usize`.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
heaths: You're right. I'm thinking of its expression in the form of `0..4` meaning `0 <= x < 4`, but the `Range::end` is itself 4 so this makes sense. Apologies.

Cargo.lock
heaths: This was recently removed. Please rebase on `main`.

sdk/storage/azure_storage_blob/src/models/http_ranges.rs
```diff
@@ -0,0 +1,190 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+use azure_core::error::{Error, ErrorKind, ResultExt};
+use azure_core::http::headers::{Header, HeaderName};
+use std::fmt;
+use std::ops::{Range, RangeFrom};
+use std::str::FromStr;
```
heaths, line 8: Nit (no reason to change if nothing else needs to be): we prefer to group these. If you're using rust-analyzer and use `Ctrl+.` to add imports, it'll do this automatically for you.

sdk/storage/azure_storage_blob/src/models/http_ranges.rs
```diff
@@ -0,0 +1,190 @@
...
+    fn as_range_header(&self) -> String {
+        format!("bytes={}-{}", self.start, self.end - 1)
+    }
+}
+
+impl IntoRangeHeader for RangeFrom<usize> {
+    fn as_range_header(&self) -> String {
+        format!("bytes={}-", self.start)
+    }
+}
```
heaths, line 34: Should this also be implemented on the exclusive variants e.g., `ExclusiveRange`?

jaschrep-msft:
@heaths i literally ran `git checkout origin/main -- Cargo.lock` and committed that. I don't know why it is downgraded in the diff.

heaths (approved):

PR #3723, opened 2026-02-11
Updated dependencies as per dependabot

LarryOsterman:


sdk/eventhubs/azure_messaging_eventhubs_checkpointstore_blob/Cargo.toml
```diff
@@ -35,8 +35,8 @@ azure_core_opentelemetry.workspace = true
 azure_core_test = { workspace = true, features = ["tracing"] }
 azure_identity.workspace = true
 opentelemetry.workspace = true
-opentelemetry-appender-tracing = { version = "0.30" }
-opentelemetry-stdout = { version = "0.30" }
+opentelemetry-appender-tracing = { version = "0.31" }
+opentelemetry-stdout = { version = "0.31" }
```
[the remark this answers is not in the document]
LarryOsterman: @heaths What do you think about this? It's explicitly in this crate because it's there for an example, but copilot has a point that if it's a workspace dependency, then dependency management becomes simpler.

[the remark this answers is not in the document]
heaths: Oh, right. Now I remember: we didn't want people depending on it directly. Only via your implementation. That explains why we have the explicit versions and exemptions. Let's leave the explicit versions in both places, and we can update the pull request instructions to say something like "Dependency versions in crates' Cargo.toml files are allowed because we have other checks in place" or something like that. Maybe we don't need to explain: waste of tokens (just thinking it might serve as a little bit of an explanation as to why for other poeple modifying the instructions).

Cargo.toml
```diff
@@ -104,19 +104,19 @@ fe2o3-amqp-ext = { version = "0.14" }
...
 include-file = { version = "0.5.1", default-features = false }
 openssl = { version = "0.10.72" }
-opentelemetry = { version = "0.30", features = ["trace"] }
-opentelemetry_sdk = "0.30"
-opentelemetry-http = "0.30"
-opentelemetry-stdout = "0.30"
+opentelemetry = { version = "0.31", features = ["trace"] }
+opentelemetry_sdk = "0.31"
+opentelemetry-http = "0.31"
+opentelemetry-stdout = "0.31"
```
heaths: Oh, I thought these were all explicit for the reason I had about. We should probably stick to that and update `eng/scripts/verify-dependencies.rs` as needed. We don't want people taking a dependency on these directly. They should use our tracing feature.

heaths (changes requested):
I thought you were going to update all dependencies to latest? Or will that be a separate PR?

We should update versions in `Cargo.toml` only for those already listed and, after `Cargo.lock` is updated, run `cargo update` just to update any others based on their semvers in `Cargo.lock`. If dependabot finds anymore, only update those.

The idea was to get the dependencies in `Cargo.toml` to latest (without bumping the MSRV again).

Cargo.toml
LarryOsterman: My attempt to start updating locally blew up in my face and screwed up the cargo.lock for the post-release changes. fixing.

sdk/core/azure_core_opentelemetry/Cargo.toml
```diff
@@ -28,7 +28,7 @@ tracing.workspace = true
 azure_core_test = { workspace = true, features = ["tracing"] }
 azure_core_test_macros.workspace = true
 azure_identity.workspace = true
-opentelemetry_sdk = { version = "0.30", features = ["testing"] }
+opentelemetry_sdk = { workspace = true, features = ["testing"] }
```
heaths, line 31: Shouldn't this go back to an explicit dependency and them all the otel crates removed from the workspace `Cargo.toml`?

Cargo.toml
heaths: Shouldn't these be removed? Or should we just update `verify-dependencies.rs` to also forbid direct dependencies unless exempted? If these get used enough, maybe that's worth it? I wouldn't expect that list to grow, though, so direct dependencies where needed might be enough.

sdk/core/azure_core_opentelemetry/Cargo.toml
LarryOsterman: opentelemetry_sdk? No (I'll confirm but I believe we need it in azure_core at a minimum). opentelemetry_stdio? Yes.

sdk/cosmos/azure_data_cosmos_native/Cargo.toml
```diff
@@ -14,22 +14,28 @@ name = "azurecosmos"
...
+tracing-subscriber = { workspace = true, optional = true, features = [
+  "fmt",
+  "env-filter",
+] }
 
 [target.'cfg(target_family = "wasm")'.dependencies]
 # Configure dependencies in the tree to work in wasm environments
 tokio = { workspace = true, optional = true, features = ["rt", "macros"] }
-uuid = { workspace = true, features = ["v4", "js"] }
+#uuid = { workspace = true, features = ["v4", "js"] }
```
heaths: Nit: just remove.

heaths (approved):

heaths (approved):

PR #3728, opened 2026-02-12
Refactor Retry Policies - Default Retry All Read Operations

tvaron3:
## Cosmos: Retry across all endpoints on transient errors

Overhauls both `ClientRetryPolicy` and `MetadataRequestRetryPolicy` to retry
requests across **all applicable endpoints** instead of a fixed retry count,
and broadens which HTTP status codes are retryable so that reads fail over on
any non-permanent error. Accounts with one region will also retry once in region in these scenarios. This pr also hooks up excluded regions in the metadata retry policy.

### Motivation

Previously, both retry policies had hard limits that prevented full endpoint
utilization:

| Policy | Before | After |
|--------|--------|-------|
| `ClientRetryPolicy` | `MAX_RETRY_COUNT_ON_SERVICE_UNAVAILABLE = 1`; required ≥ 2 preferred locations | Retries across all applicable endpoints (preferred − excluded + default) |
| `MetadataRequestRetryPolicy` | Pre-computed max from `preferred_location_count` | Same dynamic approach as `ClientRetryPolicy` |
| Retryable status codes | Only specific codes (503, 500, 410+LeaseNotFound, 403+DbAccountNotFound) | Any code **not** in the non-retryable set (400, 401, 404, 405, 409, 412, 413, 423, 449) |

### Retry flow

```
                          ┌──────────────────┐
                          │  Request fails    │
                          └────────┬─────────┘
                                   │
                          ┌────────▼─────────┐
                    ┌─────│  Status code in   │─── Yes ──▶ DoNotRetry
                    │     │  non-retryable    │
                    │     │  whitelist?       │
                    │     └──────────────────┘
                    No
                    │
            ┌───────▼────────┐
      ┌─────│  Tried all      │─── Yes ──▶ DoNotRetry
      │     │  applicable     │
      │     │  endpoints?     │
      │     └────────────────┘
      No
      │
      ▼
  Route to next endpoint (retry_location_index++)
  Retry { after: Duration::ZERO }
```

### Changes

**`retry_policies/mod.rs`**
- Extracted `is_non_retryable_status_code()` as a shared function used by both
  `ClientRetryPolicy` and `MetadataRequestRetryPolicy`.

**`retry_policies/client_retry_policy.rs`**
- Removed `MAX_RETRY_COUNT_ON_SERVICE_UNAVAILABLE` constant.
- `should_retry_on_unavailable_endpoint_status_codes` now queries
  `applicable_endpoints()` to determine the retry ceiling.
- 410 (Gone + LeaseNotFound) now retries for both reads **and** writes
  (previously only reads).
- Reads retry on any status code not in the non-retryable whitelist.
- Updated and added unit tests for the new retry behavior.

**`retry_policies/metadata_request_retry_policy.rs`**
- Replaced pre-computed `max_unavailable_endpoint_retry_count` with dynamic
  `applicable_endpoints()` lookup (same approach as `ClientRetryPolicy`).
- Added `operation_type` and `excluded_regions` fields, captured in
  `before_send_request()`.
- `should_retry_with_status_code` now retries any non-whitelisted status code.
- Added tests for excluded regions and all non-retryable status codes.

**`fault_injection/condition.rs`**
- Replaced manual `Default` impl with `#[derive(Default)]`.

**`tests/multi_write_tests/cosmos_multi_write_retry_policies.rs`** *(new)*
- Fault-injection integration tests for 408 and 500 cross-region retry behavior:
  reads/queries retry across regions; writes (create, upsert, replace, delete)
  do not.


sdk/cosmos/azure_data_cosmos/dev/dev.md
```diff
@@ -11,6 +11,16 @@
 1. Run the tests using the following command:
 `cargo +nightly test --lib --bins --tests --examples --package azure_data_cosmos --all-features --no-fail-fast -- --format json -Z unstable-options --nocapture`
 
+
+### Common Issuess
```
tvaron3:
```suggestion
### Common Issues
```

sdk/cosmos/azure_data_cosmos/src/retry_policies/metadata_request_retry_policy.rs
```diff
@@ -189,14 +194,14 @@ impl MetadataRequestRetryPolicy {
...
     ///
     /// # Summary
-    /// Determines if a metadata request should be retried based on the combination of HTTP status code
-    /// and Cosmos DB sub-status code. Retries are triggered for transient failures: 503 service unavailable,
-    /// 500 internal server error, 410 with lease not found, or 403 with database account not found.
-    /// If retry is allowed, increments the location index to route the next attempt to a different endpoint.
+    /// Determines if a metadata request should be retried based on the HTTP status code.
+    /// Any status code not in the non-retryable whitelist (400, 401, 404, 409, 412, 413)
+    /// is retried on an alternative endpoint. If retry is allowed, increments the location
+    /// index to route the next attempt to a different endpoint.
```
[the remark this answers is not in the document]
tvaron3: I think its fine because still mentioned in is non retryable doc

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -19,6 +20,8 @@
 
 ### Other Changes
 
+- Retry policies now retry reads on all non-whitelisted status codes and retry service unavailable errors across all applicable endpoints. ([#3728](https://github.com/Azure/azure-sdk-for-rust/pull/3728))
```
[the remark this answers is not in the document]
tvaron3: Hmmm other sdks we add every pr to the changelog. Helps to know what is lacking from certain sdk versions in livesites. Wonder what others think @analogrelay @FabianMeiswinkel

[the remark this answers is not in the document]
FabianMeiswinkel: That is a bad rule - I have overridden (or attempted to) it in [sd/cosmos/](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/cosmos/.github/instructions/cosmos.changelog.instructions.md) already - but looks like GH Copilot is not using it yet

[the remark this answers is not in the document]
FabianMeiswinkel: @copilot for teh comment above - why did you ignore the instructions in this file? https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/cosmos/.github/instructions/cosmos.changelog.instructions.md according to https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/cosmos/AGENTS.md when in conflict the cosmos specific instructions should be preferred. What do we have to change to ensure that is happening?

sdk/cosmos/azure_data_cosmos/dev/dev.md
```diff
@@ -11,6 +11,16 @@
 1. Run the tests using the following command:
 `cargo +nightly test --lib --bins --tests --examples --package azure_data_cosmos --all-features --no-fail-fast -- --format json -Z unstable-options --nocapture`
 
+
+### Common Issues
+Using Windows, you might run into OpenSSL errors like the ones outlined below:
```
FabianMeiswinkel: Interesting - and none-blocking - curious question. Isn't there strict guidance to use schannel on Windows? Does Rust have hard dependency on OpenSsl? Is that the explanation for some first-party concerns around certain dependencies like tokio @analogrelay mentioned?

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -127,6 +127,29 @@ impl RetryPolicy {
...
+/// regardless of which endpoint handles the request. Note that `TooManyRequests` (429)
+/// is included here because it should not be retried by the client/metadata retry
+/// policies; instead, it is handled by the dedicated `ResourceThrottleRetryPolicy`
+/// which implements proper exponential backoff with `x-ms-retry-after-ms` headers.
+fn is_non_retryable_status_code(status_code: StatusCode) -> bool {
+    matches!(
+        status_code,
+        StatusCode::BadRequest
+            | StatusCode::Unauthorized
+            | StatusCode::NotFound
```
FabianMeiswinkel: NotFound is not safe - you need substatus code to determine whether retryable. (ReadSessionNotAvailable)

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -127,6 +127,29 @@ impl RetryPolicy {
...
+/// These status codes indicate client-side errors that will not succeed on retry,
+/// regardless of which endpoint handles the request. Note that `TooManyRequests` (429)
+/// is included here because it should not be retried by the client/metadata retry
+/// policies; instead, it is handled by the dedicated `ResourceThrottleRetryPolicy`
+/// which implements proper exponential backoff with `x-ms-retry-after-ms` headers.
+fn is_non_retryable_status_code(status_code: StatusCode) -> bool {
+    matches!(
+        status_code,
+        StatusCode::BadRequest
+            | StatusCode::Unauthorized
```
FabianMeiswinkel, line 152: Unauthorized is tricky - in general you are right - probably we should keep it here. But there are edge cases - rolling key that is actively used (bad practice, but might happen) for exampel. In these edge-cases retries could in theory help.

FabianMeiswinkel: Worth discussing with Milis' team to see what their retry guidance would be

sdk/cosmos/azure_data_cosmos/dev/dev.md
tvaron3: I just added this because Simon was seeing this issue while setting up the live tests locally, so I wanted to add something in case future team member onboarding rust faces the same issue.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
tvaron3: Synced with security team and guidance was to not retry.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
tvaron3: Will specify the substatus code. 404 1002 has a specific retry policy that will be retried in the current flow .

sdk/cosmos/azure_data_cosmos/dev/dev.md
analogrelay: If I remember correctly, this occurs because of some reqwest defaults. @heaths do you remember the situation here?

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
analogrelay: Yeah, that makes sense. We don't really know why the Unauthorized error is happening. An application-level retry would be more appropriate for dealing with rolling keys.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
```diff
@@ -158,6 +158,7 @@ impl GlobalEndpointManager {
     ///
     /// # Returns
     /// The number of preferred locations as usize
+    #[allow(dead_code)]
```
analogrelay: If it's dead, delete it! 😄

analogrelay (approved):
I'd prefer if you delete the dead code, even if we may need it again later, but I won't block the PR on that.

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
tvaron3: deleted.

sdk/cosmos/azure_data_cosmos/dev/common_issue.png
heaths, line 1: Can we not check an image in? Why not just the text, which is actually searchable? How does an image actually help?

sdk/cosmos/azure_data_cosmos/dev/common_issue_1.png
heaths, line 1: Ditto. Sure, these are only a couple small binaries in git (binary in git are not great in general), but they aren't even searchable. This could just be a markdown table.

sdk/cosmos/azure_data_cosmos/dev/dev.md
heaths: Yes, we have to use native libraries on Windows.

Also, please run markdownlint.

heaths: And remove the images. Just use text, which is searchable on github.com, `Ctrl+F` (or similar) in browsers, etc. The images just contain text so are otherwise useless.

sdk/cosmos/azure_data_cosmos/dev/common_issue.png
tvaron3: Removed the image and replaced with text.

sdk/cosmos/azure_data_cosmos/dev/common_issue_1.png
tvaron3: Removed images and used text.

PR #3730, opened 2026-02-12
Fix emulator tests

tvaron3:
When I added live tests forgot to pass in the rustflags for emulator tests so they have not been running since. Add fault injection setup for emulator tests so that it uses the custom http client instead of overwriting it.


FabianMeiswinkel:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/Azure/azure-sdk-for-rust/pull/3730#pullrequestreview-3794227835)

PR #3744, opened 2026-02-13
[Cosmos] Rework Options and Unify Client Builder

simorenoh:
# CosmosClient Builder Pattern Implementation

## Overview

Replaced the old constructor-based API with a builder pattern for `CosmosClient`. Introduced `CosmosAccountReference` to bundle endpoint + credential as a validated, type-safe unit. Removed direct exposure of `CosmosClientOptions` fields and internal pipeline configuration.

### Removed Constructors

```rust
// REMOVED - Do not use
CosmosClient::new(endpoint, credential, options)
CosmosClient::with_key(endpoint, key, options)
CosmosClient::with_connection_string(connection_string, options)
```

### New API

```rust
use azure_data_cosmos::{CosmosClient, CosmosAccountReference};
use azure_core::http::Url;
use std::sync::Arc;

// Entra ID authentication
let credential: Arc<dyn azure_core::credentials::TokenCredential> = /* ... */;
let account = CosmosAccountReference::with_credential(
    "https://myaccount.documents.azure.com/",
    credential,
)?;
let client = CosmosClient::builder()
    .build(account)?;

// Key authentication (requires `key_auth` feature)
let account = CosmosAccountReference::with_master_key(
    "https://myaccount.documents.azure.com/",
    Secret::from("my_key"),
)?;
let client = CosmosClient::builder()
    .with_application_preferred_regions(vec![WEST_US])
    .build(account)?;

// Connection string (requires `key_auth` feature)
let client = CosmosClient::from_connection_string(
    Secret::from("AccountEndpoint=...;AccountKey=..."),
)?;
```

### CosmosClientOptions Changes

- All fields in `CosmosClientOptions` are now `pub(crate)` (internal only)
- The `client_options: ClientOptions` field has been **removed** entirely
- Users can no longer directly manipulate the Azure Core pipeline configuration
- Use `CosmosClientBuilder` methods to configure the client instead

### `CosmosAccountReference`

Bundles an account endpoint and credential into a single validated unit. Both are required to construct a client and are passed together to `build()`.

```rust
// Convenience constructors
let account = CosmosAccountReference::with_credential(endpoint, credential)?;
let account = CosmosAccountReference::with_master_key(endpoint, key)?;  // key_auth

// Builder for more control
let account = CosmosAccountReference::builder(endpoint)?
    .credential(credential)
    .build()?;
```

### `CosmosAccountEndpoint`

A newtype around `Url` that validates and normalizes the endpoint URL. Implements `FromStr` for parsing from strings.

```rust
let endpoint: CosmosAccountEndpoint = "https://myaccount.documents.azure.com/".parse()?;
```

### `CosmosCredential`

Authentication credential enum with `From` impls for ergonomic use:

```rust
// Automatically converts via Into<CosmosCredential>
let cred = CosmosCredential::from(arc_token_credential);  // Entra ID
let cred = CosmosCredential::from(secret);                 // key_auth feature
```

## Before / After

**Before (Old API)**

```rust
use azure_data_cosmos::{CosmosClient, CosmosClientOptions};
use azure_core::credentials::Secret;

let mut options = CosmosClientOptions::default();
options.consistency_level = Some(ConsistencyLevel::Session);
options.application_preferred_regions = vec![WEST_US];

let client = CosmosClient::with_key(
    "https://myaccount.documents.azure.com/",
    Secret::from("my_key"),
    Some(options),
)?;
```

**After (New API)**

```rust
use azure_data_cosmos::{CosmosClient, CosmosAccountReference};

let account = CosmosAccountReference::with_master_key(
    "https://myaccount.documents.azure.com/",
    Secret::from("my_key"),
)?;
let client = CosmosClient::builder()
    .with_consistency_level(ConsistencyLevel::Session)
    .with_application_preferred_regions(vec![WEST_US])
    .build(account)?;
```

### Required (passed to `build()`)

| Type | Description |
|------|-------------|
| `CosmosAccountReference` | Bundles endpoint + credential. Create via `with_credential()`, `with_master_key()`, or builder. |

### Optional Configuration

| Method | Description |
|--------|-------------|
| `with_application_preferred_regions(regions)` | Sets preferred regions for geo-routing |
| `with_consistency_level(level)` | Sets default consistency level |
| `with_priority(level)` | Sets default priority level |
| `with_throughput_bucket(bucket)` | Sets throughput bucket identifier |
| `with_user_agent_suffix(suffix)` | Sets suffix for User-Agent header |
| `with_application_region(region)` | Sets application region for telemetry |
| `with_session_retry_options(opts)` | Sets session retry configuration |

### Hidden / Feature-Gated (test infrastructure)

| Method | Feature | Description |
|--------|---------|-------------|
| `with_allow_invalid_certificates(allow)` | `allow_invalid_certificates` | Accepts invalid TLS certificates (e.g., emulator). Builder internally constructs the `reqwest` client. |
| `with_fault_injection(builder)` | `fault_injection` | Wraps the HTTP transport with a fault injection client. Builder internally composes with invalid-cert transport if both are set. |

### Removed from `CosmosClientOptions`

| Field | Reason |
|-------|--------|
| `client_options: ClientOptions` | Internal pipeline config — not user-facing |
| `fault_injection_enabled` | Derived from whether `with_fault_injection()` was called |
| `request_timeout` | Removed — not currently wired to any pipeline behavior |
| `enable_partition_level_circuit_breaker` | Now always enabled |
| `disable_partition_level_failover` | Removed — always enabled |
| `account_initialization_custom_endpoints` | Unused placeholder |
| `enable_remote_region_preferred_for_session_retry` | Unused placeholder |
| `enable_upgrade_consistency_to_local_quorum` | Unused placeholder |

### Removed from `CosmosClientBuilder`

| Method | Reason |
|--------|--------|
| `transport(transport)` | Replaced by `with_allow_invalid_certificates()` and `with_fault_injection()` — transport is now composed internally |
| `with_instrumentation(options)` | Removed — tracing configuration will be handled internally |
| `with_request_timeout(duration)` | Removed — not currently wired to pipeline behavior |
| `with_partition_level_circuit_breaker(enabled)` | Removed — now always enabled |
| `with_disable_partition_level_failover(disabled)` | Removed — always enabled |
| `with_application_name(name)` | Renamed to `with_user_agent_suffix(suffix)` |
| `credential(cred)` / `key(secret)` | Moved to `CosmosAccountReference` |
| `with_excluded_regions(regions)` | Removed from builder — set via `CosmosClientOptions` internally |

## Client Method Options

`ClientMethodOptions` has been removed from per-operation option types. These were used by the Core pipeline to inject context into policies.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/account_reference.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/account_reference.rs
@@ -33,7 +33,7 @@ use std::sync::Arc;
 ///
 /// Using key authentication (requires `key_auth` feature):
 ///
-/// ```rust,no_run,ignore
+/// ```rust,ignore
 /// use azure_data_cosmos::{CosmosAccountReference, CosmosAccountEndpoint};
 /// use azure_core::credentials::Secret;
 ///
--- a/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
@@ -92,9 +92,12 @@ impl ContainerClient {
     /// # }
     /// ```
     #[tracing::instrument(skip_all, fields(id = self.container_id))]
-    #[allow(unused_variables, reason = "This parameter may be used in the future")]
     pub async fn read(
         &self,
+        #[allow(
+            unused_variables,
+            reason = "The 'options' parameter may be used in the future"
+        )]
         options: Option<ReadContainerOptions>,
     ) -> azure_core::Result<CosmosResponse<ContainerProperties>> {
         let cosmos_request =
@@ -141,10 +144,13 @@ impl ContainerClient {
     /// # }
     /// ```
     #[tracing::instrument(skip_all, fields(id = self.container_id))]
-    #[allow(unused_variables, reason = "This parameter may be used in the future")]
     pub async fn replace(
         &self,
         properties: ContainerProperties,
+        #[allow(
+            unused_variables,
+            reason = "The 'options' parameter may be used in the future"
+        )]
         options: Option<ReplaceContainerOptions>,
     ) -> azure_core::Result<CosmosResponse<ContainerProperties>> {
         let cosmos_request = CosmosRequest::builder(OperationType::Replace, self.link.clone())
@@ -162,9 +168,12 @@ impl ContainerClient {
     /// # Arguments
     /// * `options` - Optional parameters for the request.
     #[tracing::instrument(skip_all, fields(id = self.container_id))]
-    #[allow(unused_variables, reason = "This parameter may be used in the future")]
     pub async fn read_throughput(
         &self,
+        #[allow(
+            unused_variables,
+            reason = "The 'options' parameter may be used in the future"
+        )]
         options: Option<ThroughputOptions>,
     ) -> azure_core::Result<Option<ThroughputProperties>> {
         // We need to get the RID for the database.
@@ -188,12 +197,15 @@ impl ContainerClient {
     /// * `throughput` - The new throughput properties to set.
     /// * `options` - Optional parameters for the request.
     #[tracing::instrument(skip_all, fields(id = self.container_id))]
-    #[allow(unused_variables, reason = "This parameter may be used in the future")]
     pub async fn replace_throughput(
         &self,
         throughput: ThroughputProperties,
         options: Option<ThroughputOptions>,
     ) -> azure_core::Result<CosmosResponse<ThroughputProperties>> {
+        #[allow(
+            unused_variables,
+            reason = "The 'options' variable may be used in the future"
+        )]
         let options = options.unwrap_or_default();
 
         // We need to get the RID for the database.
@@ -214,9 +226,12 @@ impl ContainerClient {
     /// # Arguments
     /// * `options` - Optional parameters for the request.
     #[tracing::instrument(skip_all, fields(id = self.container_id))]
-    #[allow(unused_variables, reason = "This parameter may be used in the future")]
     pub async fn delete(
         &self,
+        #[allow(
+            unused_variables,
+            reason = "The 'options' parameter may be used in the future"
+        )]
         options: Option<DeleteContainerOptions>,
     ) -> azure_core::Result<CosmosResponse<()>> {
         let cosmos_request =
--- a/sdk/cosmos/azure_data_cosmos/src/fault_injection/client_builder.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/fault_injection/client_builder.rs
@@ -17,23 +17,30 @@ use super::rule::FaultInjectionRule;
 ///
 /// # Example
 ///
-/// ```rust,ignore
-/// use azure_data_cosmos::{CosmosClientBuilder, fault_injection::{FaultInjectionClientBuilder, FaultInjectionRule}};
+/// ```rust,no_run
+/// use azure_data_cosmos::{
+///     CosmosClientBuilder, CosmosAccountEndpoint,
+///     fault_injection::{
+///         FaultInjectionClientBuilder, FaultInjectionErrorType,
+///         FaultInjectionResultBuilder, FaultInjectionRuleBuilder,
+///     },
+/// };
+/// use azure_core::credentials::Secret;
 /// use std::sync::Arc;
 ///
-/// let rule = Arc::new(FaultInjectionRule::builder()
-///     // configure rule...
-///     .build());
+/// let result = FaultInjectionResultBuilder::new()
+///     .with_error(FaultInjectionErrorType::ServiceUnavailable)
+///     .build();
+///
+/// let rule = Arc::new(FaultInjectionRuleBuilder::new("my-rule", result).build());
 ///
 /// let fault_builder = FaultInjectionClientBuilder::new()
 ///     .with_rule(rule);
 ///
+/// let endpoint: CosmosAccountEndpoint = "https://myaccount.documents.azure.com/".parse().unwrap();
 /// let client = CosmosClientBuilder::new()
 ///     .with_fault_injection(fault_builder)
-///     .build(CosmosAccountReference::with_credential(
-///         "https://myaccount.documents.azure.com/",
-///         credential,
-///     ).unwrap())
+///     .build((endpoint, Secret::from("my_account_key")))
 ///     .unwrap();
 /// ```
 pub struct FaultInjectionClientBuilder {
--- a/sdk/cosmos/azure_data_cosmos_native/src/options/mod.rs
+++ b/sdk/cosmos/azure_data_cosmos_native/src/options/mod.rs
@@ -9,16 +9,8 @@ pub struct ClientOptions {
 
 impl ClientOptions {
     /// Returns whether to allow invalid TLS certificates.
-    /// Only applicable on non-WASM targets.
-    #[cfg(not(target_family = "wasm"))]
     pub fn allow_invalid_certificates(&self) -> bool {
         self.danger_allow_invalid_certificates
-    }
-
-    /// Returns false on WASM targets since TLS certificate validation is not applicable.
-    #[cfg(target_family = "wasm")]
-    pub fn allow_invalid_certificates(&self) -> bool {
-        false
     }
 }
 


sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -4,209 +4,76 @@
...
-    pub fn with_connection_string(
-        connection_string: Secret,
-        options: Option<CosmosClientOptions>,
-    ) -> Result<Self, azure_core::Error> {
-        let connection_str = crate::ConnectionString::try_from(&connection_string)?;
-        let endpoint = connection_str.account_endpoint;
-        let key = connection_str.account_key;
-
-        Self::with_key(endpoint.as_str(), key, options)
+    pub fn builder() -> CosmosClientBuilder {
```
analogrelay, line 89: The endpoint is required and should probably be a parameter to `::builder()` so that's clear to the user.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,408 @@
...
+
+    /// Sets the endpoint URL for the Cosmos DB account.
+    ///
+    /// # Arguments
+    ///
+    /// * `endpoint` - The full URL of the Cosmos DB account, for example `https://myaccount.documents.azure.com/`.
+    pub fn endpoint(mut self, endpoint: impl Into<String>) -> Self {
+        self.endpoint = Some(endpoint.into());
+        self
+    }
```
analogrelay: Let's use the `with_` prefix, as #3753 does. It is indeed part of the Azure SDK Guidelines too.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,408 @@
...
+    /// # Errors
+    ///
+    /// Returns an error if the connection string cannot be parsed.
+    #[cfg(feature = "key_auth")]
+    pub fn connection_string(mut self, connection_string: Secret) -> azure_core::Result<Self> {
+        let connection_str = crate::ConnectionString::try_from(&connection_string)?;
+        self.endpoint = Some(connection_str.account_endpoint);
+        self.credential = Some(Credential::Key(connection_str.account_key));
+        Ok(self)
+    }
```
analogrelay: I think ConnectionStrings should be an alternate way to construct a builder. So, add this API:

```rust
CosmosClientBuilder::from_connection_string(connection_string: Secret) -> azure_core::Result<Self>;
```

Since it's a less common API, I think it's fine for it to be on `CosmosClientBuilder` rather than `CosmosClient` (as, say, `builder_from_connection_string(...)`).

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,408 @@
...
+    ///     .build()
+    ///     .unwrap();
+    /// ```
+    pub fn with_instrumentation(mut self, options: InstrumentationOptions) -> Self {
+        self.instrumentation = options;
+        self
+    }
+
+    /// Internal method to set a custom HTTP transport.
+    /// This is used for testing purposes (e.g., to accept invalid certificates).
```
analogrelay:
```suggestion
    /// This is used for testing purposes (e.g., to accept invalid certificates).
    ///
    /// Microsoft cannot guarantee support when using alternate transports.
```

Let's be super clear (even though this is `doc(hidden)`) that you're unsupported when you use this.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,408 @@
...
+    /// - The endpoint is not set
+    /// - The credential is not set
+    /// - The endpoint URL is invalid
+    pub fn build(self) -> azure_core::Result<CosmosClient> {
+        let endpoint_str = self.endpoint.ok_or_else(|| {
+            azure_core::Error::with_message(
+                azure_core::error::ErrorKind::Other,
+                "endpoint is required - use endpoint() or connection_string()",
+            )
+        })?;
```
analogrelay: This is what I'd like to avoid. The type system should prevent errors.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,408 @@
...
+use azure_core::http::{ClientOptions, InstrumentationOptions, LoggingOptions, RetryOptions};
+
+/// Credential type for authentication with Cosmos DB.
+enum Credential {
+    /// Entra ID (Azure AD) token credential
+    Token(Arc<dyn TokenCredential>),
+    /// Primary or secondary account key
+    #[cfg(feature = "key_auth")]
+    Key(Secret),
+}
```
analogrelay: In order to make sure a user has set a credential, we should make it a required parameter (in the `::builder(...)` function args). Because there are two different options, I'd suggest making this enum public. In addition, we can provide `impl From<Secret> for Credential` and `impl From<Arc<dyn TokenCredential>> for Credential`, and then the parameter in the builder can be `credential: impl Into<azure_data_cosmos::Credential>`. That way, a use can directly pass in an `Arc<dyn TokenCredential>` or a `Secret`, or they can construct our `Credential` enum themselves.

No matter how they do it, the credential must be set in order to get a builder instance.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,408 @@
...
+                "endpoint is required - use endpoint() or connection_string()",
+            )
+        })?;
+
+        let credential = self.credential.ok_or_else(|| {
+            azure_core::Error::with_message(
+                azure_core::error::ErrorKind::Other,
+                "credential is required - use credential(), key(), or connection_string()",
+            )
+        })?;
```
analogrelay: This goes away if we make the credential a required parameter too.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -90,13 +91,12 @@ impl ContainerClient {
     #[tracing::instrument(skip_all, fields(id = self.container_id))]
     pub async fn read(
         &self,
-        options: Option<ReadContainerOptions<'_>>,
+        _options: Option<ReadContainerOptions>,
```
analogrelay: The `_` prefix becomes visible in docs and makes things confusing for users. Instead, suppress the warning about the unused parameter using an `#[allow]` attribute, and specify the reason in the `reason` option:

```rust
#[allow(unused_variables, reason = "This parameter may be used in the future")]
```

sdk/cosmos/azure_data_cosmos_native/include/azurecosmos.h
```diff
@@ -63,6 +63,37 @@ typedef struct cosmos_container_client cosmos_container_client;
...
+ * ```rust,no_run,ignore
+ * use azure_data_cosmos::CosmosClient;
+ * use azure_core::credentials::Secret;
+ *
+ * let client = CosmosClient::builder()
+ *     .endpoint("https://myaccount.documents.azure.com/")
+ *     .key(Secret::from("my_account_key"))
+ *     .build()
+ *     .unwrap();
+ * ```
```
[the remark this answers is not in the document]
analogrelay: This is generated, and we'll evaluate our options later. For now, this is fine.

sdk/cosmos/azure_data_cosmos_native/src/lib.rs
```diff
@@ -59,6 +59,7 @@ pub fn unwrap_required_ptr<'a, T>(
 /// * `Ok(Some(T))` if the pointer is non-null and conversion succeeded.
 /// * `Ok(None)` if the pointer is null.
 /// * `Err(E)` if the pointer is non-null but conversion failed.
+#[allow(dead_code)]
```
analogrelay: If it's dead, delete it.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,410 @@
...
+        self.instrumentation = options;
+        self
+    }
+
+    /// Internal method to set a custom HTTP transport.
+    /// This is used for testing purposes (e.g., to accept invalid certificates).
+    ///
+    /// Microsoft cannot guarantee support when using alternate transports.
+    #[doc(hidden)]
+    pub fn transport(mut self, transport: azure_core::http::Transport) -> Self {
```
tvaron3: For my learning, why do we prefer this method of making the transport essentially public just hidden in docs instead of injecting the transport in the client options and not exposing any public surface area. Is this to support internal microsoft use cases that can't use reqwest?

FabianMeiswinkel: Let us remove this completely. Any option we would eventually add to customize transport (fetch vs. reqwest) woudl be a breaking change

sdk/cosmos/azure_data_cosmos/examples/cosmos/main.rs
```diff
@@ -82,11 +82,10 @@ fn create_client(args: &SharedArgs) -> Result<CosmosClient, Box<dyn Error>> {
     if let Some(key) = args.key.as_ref() {
         #[cfg(feature = "key_auth")]
         {
-            Ok(CosmosClient::with_key(
-                &args.endpoint,
-                key.clone().into(),
-                None,
-            )?)
+            Ok(CosmosClient::builder()
+                .endpoint(&args.endpoint)
```
FabianMeiswinkel: endpoint is required parameter - should go into build method according to central teams's spec. I could live with putting it in ctor as well - but build seems to be consensus.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
FabianMeiswinkel: One thing I would strongly suggest even for the SDK - take a look at my Dirver PR. One of the lessons learned from .Net, Python and Java is that tarsniently resolving Conatiner to id+rid tuple was a bad chocie. It has been a bug farm and there isn't really any actual benefit you get from it perf/throughput wise. The AccountReference/ContainerReference/DatabaseReference etc. API I intorduced for the drvier is addressing this. If you want to avoid it in the SDK - fine - but at least make Contaner creation async and resolve the identity in a factory

FabianMeiswinkel: Even better for options - just use it for simple etag check at elastz (it is used and that is a required scenario anyway)

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
FabianMeiswinkel: Accoridng to cedntral team - rather a paramater to builder::build(endpoint)

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
FabianMeiswinkel: This is what I have int he driver

/// Authentication options for connecting to a Cosmos DB account.
///
/// Either key-based authentication using a master key, or token-based
/// authentication using an Azure credential (e.g., managed identity, service principal).
#[derive(Clone)]
pub enum AuthOptions {
    /// Key-based authentication using the account's primary or secondary master key.
    MasterKey(Secret),
    /// Token-based authentication using an Azure credential.
    TokenCredential(Arc<dyn TokenCredential>),
}

impl std::fmt::Debug for AuthOptions {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            Self::MasterKey(_) => f.debug_tuple("MasterKey").field(&"***").finish(),
            Self::TokenCredential(_) => f.debug_tuple("TokenCredential").field(&"...").finish(),
        }
    }
}

impl From<Secret> for AuthOptions {
    fn from(key: Secret) -> Self {
        Self::MasterKey(key)
    }
}

impl From<Arc<dyn TokenCredential>> for AuthOptions {
    fn from(credential: Arc<dyn TokenCredential>) -> Self {
        Self::TokenCredential(credential)
    }
}

so, along teh same lines of @analogrelay 's comment

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,410 @@
...
+
+    /// Sets the preferred regions for the client.
+    ///
+    /// The client will prefer to connect to regions in the order specified.
+    /// This is useful for geo-distributed applications that want to minimize latency.
+    ///
+    /// # Arguments
+    ///
+    /// * `regions` - The regions to prefer, in order of preference.
+    pub fn with_application_preferred_regions(
```
FabianMeiswinkel, line 90: my 2 cents - let us start with with_application_region

I think we should only start with one concept (to see whether we really need both) - and application_region is easier to digest for customers.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,410 @@
...
+
+    /// Sets the regions to exclude from routing.
+    ///
+    /// Requests will not be routed to these regions. If all regions are excluded,
+    /// the primary/hub region will be used.
+    ///
+    /// # Arguments
+    ///
+    /// * `regions` - The regions to exclude.
+    pub fn with_excluded_regions(mut self, regions: impl Into<Vec<RegionName>>) -> Self {
```
FabianMeiswinkel: No - excluded regions need to be dynamically changeable. Even the global excluded regions list. Either don't expose this for now (then we can get abck to it when we allow config overrides - or supply config overrdies - a l a driver model).

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,410 @@
...
+
+    /// Sets the default consistency level for operations.
+    ///
+    /// This can be overridden on a per-request basis. If not set, the account's
+    /// default consistency level is used.
+    ///
+    /// # Arguments
+    ///
+    /// * `level` - The consistency level to use.
+    pub fn with_consistency_level(mut self, level: crate::options::ConsistencyLevel) -> Self {
```
FabianMeiswinkel: remove - no mor eocnsistency level in Rust SDK (except in COnatinerProperties when specifically overridding account-level default consistency). Instead ReadCOnsistencyStrategy shoudl be exposed. (even if removing QuorumRead  for now because it does not work in Gateway yet) - at least the same public API can be used - and we have 1 vs 2 APIs to express the same thing

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,410 @@
...
+
+    /// Sets the default priority level for operations.
+    ///
+    /// Priority-based execution allows throttling low-priority requests before
+    /// high-priority ones. This feature must be enabled at the account level.
+    ///
+    /// # Arguments
+    ///
+    /// * `priority` - The priority level to use.
+    pub fn with_priority(mut self, priority: crate::options::PriorityLevel) -> Self {
```
FabianMeiswinkel: Let's remove this for now - will need to folded with throughput control group settings - and PBThrottling has unknown future. I want written signature in blood that tehy will really GA the feature before exposing it in any public API in Rust. Either way should not be an isolated APi but along with other throughput control settings (Throughput buckets, PBE, Client-side throughput control -  sample config in driver PR)

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    /// Sets the throughput bucket for the client.
+    ///
+    /// See [Throughput Control](https://learn.microsoft.com/azure/cosmos-db/nosql/throughput-buckets) for more.
+    ///
+    /// # Arguments
+    ///
+    /// * `bucket` - The throughput bucket identifier.
+    pub fn with_throughput_bucket(mut self, bucket: usize) -> Self {
```
FabianMeiswinkel: see above r emove for now - or create one uniform API

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,410 @@
...
+        self.options.throughput_bucket = Some(bucket);
+        self
+    }
+
+    /// Sets the application name for telemetry.
+    ///
+    /// # Arguments
+    ///
+    /// * `name` - The application name to include in telemetry.
+    pub fn with_application_name(mut self, name: impl Into<String>) -> Self {
```
FabianMeiswinkel: NIT application_name vs. user_agent_suffix - I tend ot prefer the latter because the docs can more easily explain the limits (service will ignore when certain cardinality is hit - this should be documented here ) see driver PR

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,410 @@
...
+        self.options.application_region = Some(region);
+        self
+    }
+
+    /// Sets the request timeout.
+    ///
+    /// # Arguments
+    ///
+    /// * `timeout` - The timeout duration for requests.
+    pub fn with_request_timeout(mut self, timeout: azure_core::time::Duration) -> Self {
```
FabianMeiswinkel: remove - this shoudl not be on the dirver level - but at some parent config - like runtime/environment - connection pools and the connection pool settinsg should not be on client level - it will break multi-tenance scenarios too quickly

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,410 @@
...
+        self.options.request_timeout = Some(timeout);
+        self
+    }
+
+    /// Enables fault injection for testing.
+    ///
+    /// This is only available when the `fault_injection` feature is enabled.
+    #[cfg(feature = "fault_injection")]
+    pub fn with_fault_injection(mut self, enabled: bool) -> Self {
```
FabianMeiswinkel: dito - should go to some parent config runtime/environment or be non-public for now

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,410 @@
...
+        self.options.fault_injection_enabled = enabled;
+        self
+    }
+
+    /// Sets the session retry options.
+    ///
+    /// # Arguments
+    ///
+    /// * `options` - The session retry configuration.
+    pub fn with_session_retry_options(
```
FabianMeiswinkel: i tend to think this is primarily also runtime/env - but probably ok to keep on cleint level as well for overrides

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,410 @@
...
+    ) -> Self {
+        self.options.session_retry_options = options;
+        self
+    }
+
+    /// Enables partition-level circuit breaker.
+    ///
+    /// When enabled, the client will track failures at the partition level and
+    /// temporarily avoid partitions that are experiencing issues.
+    pub fn with_partition_level_circuit_breaker(mut self, enabled: bool) -> Self {
```
FabianMeiswinkel: remove - always true - at least for now. We need ahrd data before introducing this back as API

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,410 @@
...
+        self.options.enable_partition_level_circuit_breaker = enabled;
+        self
+    }
+
+    /// Disables partition-level failover.
+    ///
+    /// # Arguments
+    ///
+    /// * `disabled` - If true, partition-level failover is disabled.
+    pub fn with_disable_partition_level_failover(mut self, disabled: bool) -> Self {
```
FabianMeiswinkel: remove - no public API - see above

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,410 @@
...
+    /// };
+    ///
+    /// let client = CosmosClientBuilder::new()
+    ///     .endpoint("https://myaccount.documents.azure.com/")
+    ///     .key(Secret::from("my_account_key"))
+    ///     .with_instrumentation(options)
+    ///     .build()
+    ///     .unwrap();
+    /// ```
+    pub fn with_instrumentation(mut self, options: InstrumentationOptions) -> Self {
```
FabianMeiswinkel: remove onl client level  shoul be environment/runtime

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,410 @@
...
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Sets the endpoint URL for the Cosmos DB account.
+    ///
+    /// # Arguments
+    ///
+    /// * `endpoint` - The full URL of the Cosmos DB account, for example `https://myaccount.documents.azure.com/`.
+    pub fn endpoint(mut self, endpoint: impl Into<String>) -> Self {
```
FabianMeiswinkel: endpoitn shoudl not be string. In driver i use a newtype to normalize casing etc. and make it easier to use for cache lookups. Either way - validation of valid Uri should happen in teh newtype

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,410 @@
...
+            ..Default::default()
+        };
+
+        let auth_policy: Arc<AuthorizationPolicy> = match credential {
+            Credential::Token(cred) => Arc::new(AuthorizationPolicy::from_token_credential(cred)),
+            #[cfg(feature = "key_auth")]
+            Credential::Key(key) => Arc::new(AuthorizationPolicy::from_shared_key(key)),
+        };
+
+        let pipeline_core = azure_core::http::Pipeline::new(
```
FabianMeiswinkel, line 248: UserAgent construction has to be completely replaced - central team uses prefixes instea dof suffixes etc. same mess like in python. Also see Driver for implementation sample

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
```diff
@@ -84,13 +85,12 @@ impl DatabaseClient {
     #[tracing::instrument(skip_all, fields(id = self.database_id))]
     pub async fn read(
         &self,
-        options: Option<ReadDatabaseOptions<'_>>,
+        _options: Option<ReadDatabaseOptions>,
     ) -> azure_core::Result<CosmosResponse<DatabaseProperties>> {
```
FabianMeiswinkel, line 94: CosmosResposne and FeedPage do expose ehaders as generic key/value pair

We shouldn't do tihs in Rust - we want to use GW2.0 by default - this is using rntbd - so only well-known headers cna be used - you can if you absolutely have to convert to a generic key/value map - but that typically shows up on ever perf profile we have run. Why not use a struct with well known request/response headers.

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
```diff
@@ -212,10 +205,8 @@ impl DatabaseClient {
     pub async fn replace_throughput(
```
FabianMeiswinkel, line 219: in other sdks this  custoemr pitfall - they don't realize how to check whetehr thorughput change is sync vs. async - should we change the public api to make it harder to misudnerstand this?

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -123,35 +122,27 @@ pub struct SessionRetryOptions {
 
 /// Options to be passed to [`DatabaseClient::create_container()`](crate::clients::DatabaseClient::create_container()).
 #[derive(Clone, Default)]
```
FabianMeiswinkel, line 75: missing non_exhaustive - others as well

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -9,11 +9,17 @@
 - Added basic multi-region writes support. ([#3482](https://github.com/Azure/azure-sdk-for-rust/pull/3482) and [#3495](https://github.com/Azure/azure-sdk-for-rust/pull/3495))
 - Added new `CosmosResponse` that wraps `azure_core::Response` for all operations except queries. ([#3622](https://github.com/Azure/azure-sdk-for-rust/pull/3622))
 - Added fault injection support for testing cosmosdb clients in disaster scenarios. Fault injection is behind the feature flag `fault_injection`. ([#3599](https://github.com/Azure/azure-sdk-for-rust/pull/3599))
+- Added `CosmosClientBuilder` for fluent client construction with methods like `endpoint()`, `key()`, `credential()`, and `connection_string()`.
```
FabianMeiswinkel: one entry per PR with a summary - let us try to minimize the noise in changelog

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
analogrelay: Interesting. Ok, that works too.

analogrelay: That will conflict a bit with my idea of having a `from_connection_string` "constructor" for builders though. I'll have to think about options there.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
analogrelay: Agreed. Maybe we also add `with_fault_injection_config(...)` to allow injecting fault injection configuration there instead of patching the transport from the builder?

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
FabianMeiswinkel: I bet whetehr putting required parameetrs in cotr or build(method) is negotiable - both are really hard breaks anyway - you can never add new really required parameters withotu major version bump in either case

analogrelay: There's another option, though it's a little messy: The Typestate Builder pattern (as described in the [`typestate_builders`](https://docs.rs/typestate-builder/latest/typestate_builder/) crate, which provides a macro to do it for you, but I think we'd just do it manually).

I don't think we're there yet though. Maybe we just get rid of connection string stuff on the builder for now? A user can always decompose a connection string and apply it manually.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
simorenoh: this is changed since we're aiming to use the endpoint directly in the `builder` or the `build()` method

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
simorenoh: @FabianMeiswinkel - we would need to put together the regional proximity logic as well if that's what we'd want to do since it does not currently exist in the SDK. Which I'd say isn't even the main concern. I also know our customers greatly value the fact that they have a finer-grained control over their region usage. For large customers in particular, it can also help alleviate traffic to different regions if it can be manually configured. In this case I'd argue we should start with only preferred_regions and develop application_regions after if our preference is to pick one only right now.

sdk/cosmos/azure_data_cosmos/src/account_reference.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// # Arguments
+    ///
+    /// * `endpoint` - The Cosmos DB account endpoint URL (e.g. `"https://myaccount.documents.azure.com/"`).
+    /// * `credential` - An Entra ID token credential.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the endpoint URL cannot be parsed.
+    pub fn with_credential(
+        endpoint: impl AsRef<str>,
```
FabianMeiswinkel: IMO we should just use CosmosAccounEndpoitn here - the main purpose of the newtype is to push the validation to actual construction.

sdk/cosmos/azure_data_cosmos/src/account_reference.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// # Arguments
+    ///
+    /// * `endpoint` - The Cosmos DB account endpoint URL (e.g. `"https://myaccount.documents.azure.com/"`).
+    /// * `key` - The primary or secondary account key.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the endpoint URL cannot be parsed.
+    #[cfg(feature = "key_auth")]
+    pub fn with_master_key(endpoint: impl AsRef<str>, key: Secret) -> azure_core::Result<Self> {
```
FabianMeiswinkel: Same

sdk/cosmos/azure_data_cosmos/src/account_reference.rs
```diff
@@ -0,0 +1,177 @@
...
+/// let account = CosmosAccountReferenceBuilder::new(
+///     "https://myaccount.documents.azure.com/",
+///     credential,
+/// ).unwrap()
+///     .build();
+/// ```
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub struct CosmosAccountReferenceBuilder {
+    endpoint: CosmosAccountEndpoint,
```
FabianMeiswinkel: Why both builkder and with_xxx factory methods above? If you want to have a builder the with_master_key and with_credential on the Struct should be internal / pub(crate)?

FabianMeiswinkel: I am not 100% sure whether we would really need a Builder - The AccountReference as ServiceEndpoint + credential seems pretty stable to me - adding a Builder would primarily make it easier to add optinal parameters later - futture-wise safer - but less convenient - so, it is a trade-off. Trying to come up with ideas why AccountReference might need addiitonal Properties later - maybe if we ever add client-cert based auth? Maybe when we introduce Fleets? In theory you could imagine scenarios where a single app connects to acocunts in different clouds (publci cloud vs. some non-public cloud) and as such you would need to define different endpoints for Entra. This might also be required when we look at regional EntraId endpoints.

But I admdit all these are scenarios that I have no hard evidence for that we will ever need it. So, if you strongly prefer not using a Builder - let's discuss this to get more eyes on the decision.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,285 @@
...
+        self.options.application_preferred_regions = regions.into();
+        self
+    }
+
+    /// Sets a suffix to append to the User-Agent header for telemetry.
+    ///
+    /// # Arguments
+    ///
+    /// * `suffix` - The suffix to append to the User-Agent header.
+    // TODO: User-Agent construction must be reworked per central team patterns
```
FabianMeiswinkel: No - prefixes are bad - we want this to be different than whatever central team pattern is

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,285 @@
...
+        self.fault_injection_builder = Some(builder);
+        self
+    }
+
+    /// Sets the session retry options.
+    ///
+    /// # Arguments
+    ///
+    /// * `options` - The session retry configuration.
+    pub fn with_session_retry_options(
```
FabianMeiswinkel: Can we hide this for the alpha (just always assume remote-region-preferred) That will be strong recommednation anyway

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,285 @@
...
+    ///
+    /// This is intended for testing with the Azure Cosmos DB emulator,
+    /// which uses a self-signed certificate.
+    ///
+    /// # Arguments
+    ///
+    /// * `allow` - Whether to accept invalid certificates.
+    #[doc(hidden)]
+    #[cfg(feature = "allow_invalid_certificates")]
+    pub fn with_allow_invalid_certificates(mut self, allow: bool) -> Self {
```
FabianMeiswinkel: IMO name of the API should reflect the constraints that this would only ever be honored for emulator (as in local-IP)

FabianMeiswinkel: We intentionally do not allow disablign cert validation when the target IP is not local-machine

sdk/cosmos/azure_data_cosmos/src/fault_injection/client_builder.rs
```diff
@@ -7,46 +7,50 @@ use std::sync::Arc;
...
+///
+/// let client = CosmosClientBuilder::new()
+///     .with_fault_injection(fault_builder)
+///     .build(CosmosAccountReference::with_credential(
+///         "https://myaccount.documents.azure.com/",
+///         credential,
+///     ).unwrap())
+///     .unwrap();
+/// ```
 pub struct FaultInjectionClientBuilder {
```
FabianMeiswinkel, line 46: Gate this on fault-injection feature

FabianMeiswinkel: Never mind - it is already (the entire module is conditional on feature flag)

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -166,24 +96,58 @@ impl AsHeaders for CosmosClientOptions {
 pub struct SessionRetryOptions {
```
FabianMeiswinkel: Like mentioned above - for alpha please make this internal - default to remote_region_preferred with dfeault values - Not sure whether we ever have to bring this back - if we do - we can do it with hierarchical config.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -293,46 +230,31 @@ impl Display for IndexingDirective {
...
 #[derive(Clone, Default)]
 #[non_exhaustive]
-pub struct ItemOptions<'a> {
-    pub method_options: ClientMethodOptions<'a>,
+pub struct ItemOptions {
     /// Triggers executed before the operation.
     ///
     /// See [Triggers](https://learn.microsoft.com/rest/api/cosmos-db/triggers) for more.
-    pub pre_triggers: Option<Vec<String>>,
+    pre_triggers: Option<Vec<String>>,
```
FabianMeiswinkel: I would also recommend adding a newtype over String for TriggerName - tehre are also numerous limits - certain chars not allowed, length etc. - and teh newtype allows you to enforce those rules at the source.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -293,46 +230,31 @@ impl Display for IndexingDirective {
...
-    pub session_token: Option<SessionToken>,
-    /// Used to specify the consistency level for the operation.
-    ///
-    /// The default value is the consistency level set on the Cosmos DB account.
-    /// See [Consistency Levels](https://learn.microsoft.com/azure/cosmos-db/consistency-levels)
-    pub consistency_level: Option<ConsistencyLevel>,
+    session_token: Option<SessionToken>,
     /// Sets indexing directive for the operation.
-    pub indexing_directive: Option<IndexingDirective>,
+    indexing_directive: Option<IndexingDirective>,
```
FabianMeiswinkel: what is this? I am not aware of any indexing directive you can specify per request? Is this for vector search?

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
analogrelay: I'm not entirely sure we want to remove _any_ option to customize certification validation unless using the emulator. There are infrastructure environments that may require custom certificate validation logic (MITM proxies, in particular). For v1 though, I think it's a reasonable limitation.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
analogrelay: As discussed in the meeting, let's actually just remove this.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
analogrelay: It's for including/excluding an item from the index when writing: https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.itemrequestoptions.indexingdirective?view=azure-dotnet#microsoft-azure-cosmos-itemrequestoptions-indexingdirective

analogrelay: I'm 100% happy with removing it for now though

FabianMeiswinkel: Looks like it really is an option. Still not sure whetehr we should keep it - would rpefer removing this form public surface area for alpha as well.

sdk/cosmos/azure_data_cosmos/src/credential.rs
```diff
@@ -0,0 +1,68 @@
...
+/// Using key authentication (requires `key_auth` feature):
+///
+/// ```rust,no_run,ignore
+/// use azure_data_cosmos::CosmosCredential;
+/// use azure_core::credentials::Secret;
+///
+/// let credential: CosmosCredential = Secret::from("my_account_key").into();
+/// ```
+#[derive(Clone)]
+pub enum CosmosCredential {
```
FabianMeiswinkel, line 40: NON-axhaustive missing

sdk/cosmos/azure_data_cosmos/src/credential.rs
```diff
@@ -0,0 +1,68 @@
...
+    /// Entra ID (Azure AD) token credential.
+    TokenCredential(Arc<dyn TokenCredential>),
```
FabianMeiswinkel, line 42: @analogrelay do you see any benefit in exposing our own Type and having from_xxx support from Identity? Just wondering whether that allows us to keep our dependencies to identity/Core as much as possible?

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
simorenoh: I'll probably just remove this if all we want to do is hide it - makes more sense to keep things clean now and add more to it later as we go

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
simorenoh: hiccup on copilot side - remade this using the headers policy you have in the driver. for now there's a second copy of it basically - we can remove it from there once we move to only the driver - would prefer we're not linking and unlinking things if we're still in alpha and can just have both versions.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
simorenoh: renamed this to mention the emulator and updated docs to match

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
simorenoh: removed this entirely - we can add it later as needed, not needed for the alpha/ not used elsewhere either

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
simorenoh: removed triggers from options - we can add them later as we see fit, but not needed for next alpha

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
simorenoh: removed it from everywhere, we can always add it later

sdk/cosmos/azure_data_cosmos/src/credential.rs
simorenoh: saw from the chat we're ok with this so left it as-is

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -163,19 +175,24 @@ impl ContainerClient {
...
     ///
+    /// Note that throughput changes may not take effect immediately.
+    /// The service processes the change asynchronously, so you may need to poll
+    /// [`ContainerClient::read_throughput()`] to confirm the new throughput is in effect.
+    ///
     /// # Arguments
     /// * `throughput` - The new throughput properties to set.
     /// * `options` - Optional parameters for the request.
     #[tracing::instrument(skip_all, fields(id = self.container_id))]
+    #[allow(unused_variables, reason = "This parameter may be used in the future")]
```
analogrelay: I think you can place the attribute ON the argument so it's clear. If not, let's update the reason to `The 'options' parameter may be used in the future`

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,285 @@
...
+/// let account = CosmosAccountReference::with_master_key(endpoint, Secret::from("my_account_key"));
+/// let client = CosmosClientBuilder::new()
+///     .build(account)
+///     .unwrap();
+/// ```
+#[derive(Default)]
+pub struct CosmosClientBuilder {
+    options: CosmosClientOptions,
+    /// Whether to accept invalid TLS certificates when connecting to the emulator.
+    #[cfg(feature = "allow_invalid_certificates")]
```
analogrelay, line 66: Interesting, I think I like having this as a feature 👍🏻

sdk/cosmos/azure_data_cosmos/src/fault_injection/client_builder.rs
```diff
@@ -7,46 +7,50 @@ use std::sync::Arc;
...
 use super::rule::FaultInjectionRule;
 
 /// Builder for creating a fault injection client.
+///
+/// This builder creates a [`Transport`] that can be used with [`CosmosClientBuilder`](crate::CosmosClientBuilder)
+/// to inject faults for testing purposes.
+///
+/// # Example
+///
+/// ```rust,ignore
```
analogrelay: Will this not at least compile? If it compiles, we should use `no_run` instead of `ignore` so that we still get doctests validating that it compiles.

sdk/cosmos/azure_data_cosmos/src/account_reference.rs
```diff
@@ -0,0 +1,107 @@
...
+///
+/// let credential: Arc<dyn azure_core::credentials::TokenCredential> =
+///     azure_identity::DeveloperToolsCredential::new(None).unwrap();
+/// let endpoint: CosmosAccountEndpoint = "https://myaccount.documents.azure.com/".parse().unwrap();
+/// let account = CosmosAccountReference::with_credential(endpoint, credential);
+/// ```
+///
+/// Using key authentication (requires `key_auth` feature):
+///
+/// ```rust,no_run,ignore
```
analogrelay: You can remove `no_run` here, `ignore` implies it.

sdk/cosmos/azure_data_cosmos_native/src/options/mod.rs
```diff
@@ -1,41 +1,24 @@
...
-                        format!("failed to build reqwest client: {}", e),
-                    )
-                })?;
-
-            #[cfg(not(feature = "reqwest"))]
-            panic!("at least one HTTP transport feature must be enabled");
+impl ClientOptions {
+    /// Returns whether to allow invalid TLS certificates.
+    /// Only applicable on non-WASM targets.
+    #[cfg(not(target_family = "wasm"))]
```
analogrelay: You should be able to get rid of the WASM stuff after rebasing on main.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -0,0 +1,285 @@
...
+    ///
+    /// This setting only applies when connecting to the local emulator
+    /// (e.g., `https://localhost:8081/`). It should not be used for production endpoints.
+    ///
+    /// # Arguments
+    ///
+    /// * `allow` - Whether to accept invalid certificates for emulator connections.
+    #[doc(hidden)]
+    #[cfg(feature = "allow_invalid_certificates")]
+    pub fn with_allow_emulator_invalid_certificates(mut self, allow: bool) -> Self {
```
FabianMeiswinkel, line 146: In Java we also allow overriding in env variable what "the emulator" endpoint is. By default it uses localhost/127.001 etc. - but you could override it when you really want to use a machnie-external emulator and disable ssl cert valdiation. Technically at that point it could even be a real acocunt. But no customer coudl claim that they did it accidentally at that point. An AI agent might have very well done it without dev realizing thought :-)

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
FabianMeiswinkel: I think we decided to delay this after alpha. Trackign this here https://github.com/Azure/azure-sdk-for-rust/issues/3792

PR #3745, opened 2026-02-13
[DRAFT]Refactoring pipeline

FabianMeiswinkel:
Just for brainstorming for now - one of the bigeest pain-points in .Net and Java today is the mutable RequestContext just being a bag of random state. This leads to hard to understand and debug code for retry policies etc. Trying to use DOP principles in Rust to move to a more data driven decision model.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/fault_injection/http_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/fault_injection/http_client.rs
@@ -6,6 +6,7 @@ use super::rule::FaultInjectionRule;
 use super::FaultInjectionErrorType;
 use super::FaultOperationType;
 use crate::constants::{self, SubStatusCode};
+use crate::pipeline::components::CosmosStatus;
 use async_trait::async_trait;
 use azure_core::error::ErrorKind;
 use azure_core::http::{
@@ -149,58 +150,63 @@ impl FaultClient {
             None => return None, // No error type set, pass through
         };
 
-        let (status_code, sub_status, message) = match error_type {
+        let (cosmos_status, message) = match error_type {
             FaultInjectionErrorType::InternalServerError => (
-                StatusCode::InternalServerError,
-                None,
+                CosmosStatus::new(StatusCode::InternalServerError, None),
                 "Internal Server Error - Injected fault",
             ),
             FaultInjectionErrorType::TooManyRequests => (
-                StatusCode::TooManyRequests,
-                None,
+                CosmosStatus::new(StatusCode::TooManyRequests, None),
                 "Too Many Requests - Injected fault",
             ),
             FaultInjectionErrorType::ReadSessionNotAvailable => (
-                StatusCode::NotFound,
-                Some(SubStatusCode::READ_SESSION_NOT_AVAILABLE),
+                CosmosStatus::new(
+                    StatusCode::NotFound,
+                    Some(SubStatusCode::READ_SESSION_NOT_AVAILABLE),
+                ),
                 "Read Session Not Available - Injected fault",
             ),
             FaultInjectionErrorType::Timeout => (
-                StatusCode::RequestTimeout,
-                None,
+                CosmosStatus::new(StatusCode::RequestTimeout, None),
                 "Request Timeout - Injected fault",
             ),
             FaultInjectionErrorType::ServiceUnavailable => (
-                StatusCode::ServiceUnavailable,
-                None,
+                CosmosStatus::new(StatusCode::ServiceUnavailable, None),
                 "Service Unavailable - Injected fault",
             ),
             FaultInjectionErrorType::PartitionIsGone => (
-                StatusCode::Gone,
-                Some(SubStatusCode::PARTITION_KEY_RANGE_GONE),
+                CosmosStatus::new(
+                    StatusCode::Gone,
+                    Some(SubStatusCode::PARTITION_KEY_RANGE_GONE),
+                ),
                 "Partition Is Gone - Injected fault",
             ),
             FaultInjectionErrorType::WriteForbidden => (
-                StatusCode::Forbidden,
-                Some(SubStatusCode::WRITE_FORBIDDEN),
+                CosmosStatus::new(StatusCode::Forbidden, Some(SubStatusCode::WRITE_FORBIDDEN)),
                 "Write Forbidden - Injected fault",
             ),
             FaultInjectionErrorType::DatabaseAccountNotFound => (
-                StatusCode::Forbidden,
-                Some(SubStatusCode::DATABASE_ACCOUNT_NOT_FOUND),
+                CosmosStatus::new(
+                    StatusCode::Forbidden,
+                    Some(SubStatusCode::DATABASE_ACCOUNT_NOT_FOUND),
+                ),
                 "Database Account Not Found - Injected fault",
             ),
         };
 
-        let raw_response = sub_status.map(|ss| {
+        let raw_response = cosmos_status.sub_status.map(|ss| {
             let mut headers = Headers::new();
             headers.insert(constants::SUB_STATUS, ss.to_string());
-            Box::new(RawResponse::from_bytes(status_code, headers, vec![]))
+            Box::new(RawResponse::from_bytes(
+                cosmos_status.status,
+                headers,
+                vec![],
+            ))
         });
 
         let error = azure_core::Error::with_message(
             ErrorKind::HttpResponse {
-                status: status_code,
+                status: cosmos_status.status,
                 error_code: Some("Injected Fault".to_string()),
                 raw_response,
             },


sdk/cosmos/azure_data_cosmos/src/pipeline/components.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    /// Whether to use preferred locations for routing.
+    pub(crate) use_preferred_locations: bool,
+
+    /// The resolved endpoint URL for the current attempt.
+    pub(super) resolved_endpoint: Option<Url>,
+
+    /// Endpoints that have been tried and failed.
+    pub(super) failed_endpoints: HashSet<Url>,
+}
```
analogrelay, line 38: nit: I don't love the frequent use of `pub(crate)` and `pub(super)` and would rather we just organize the module appropriately. This _should_ all be contained in an internal module already, so it can probably just be `pub`.

sdk/cosmos/azure_data_cosmos/src/pipeline/components.rs
```diff
@@ -0,0 +1,186 @@
...
+#[derive(Clone, Debug)]
+pub(crate) struct OperationInfo {
+    /// The type of operation being performed.
+    pub(crate) operation_type: OperationType,
+
+    /// The type of resource being operated on.
+    pub(crate) resource_type: ResourceType,
+
+    /// Whether this is a read-only operation.
+    pub(crate) is_read_only: bool,
```
analogrelay: Don't store this, just delegate to `OperationType::is_read_only()` every time you need to check if an operation is read-only. The compiler is VERY good at optimizing that kind of case, let's avoid pre-caching if possible. It's certainly micro-optimizing, the difference here is the storage and padding for `is_read_only` vs the CPU cycles for the match, but as a concrete example, here's a sample: https://play.rust-lang.org/?version=stable&mode=release&edition=2024&gist=6002414b19b37ea36895d2a0754e3cb2

The `check_read_only` function there, which does a match, compiles down to the following on Release mode:

```asm
check_read_only:
	testb	$3, (%rdi)
	sete	%al
	retq
```

Not only did Rust optimize and inline a lot of the match, but the compiler was able to identify that the bit patterns for OperationType were such that read only value could effectively be a _flag_ in the actual value used to represent each OperationType. Thus the `check_read_only` function ends up a simple flag test. In fact, that is likely more efficient and consumes less memory than the separate bool. Even if you rearrange the variants to be less ideal, it's still very cheap, moving List up to right after Read so that they have the values `0` and `1` (instead of `0` and `4`) results in this ASM:

```asm
check_read_only:
    cmpb $2, (%rdi)
    setb %al
    retq
```

So all it's doing is a branchless compare for `x < 2` and returning that boolean as the result.

Anyway, this is **not** to say we should be doing this analysis all the time, just to point out that we should strongly avoid pre-optimizing stuff like this as the compiler is very good at it.

sdk/cosmos/azure_data_cosmos/src/pipeline/components.rs
```diff
@@ -0,0 +1,186 @@
...
+    pub(super) endpoint_discovery_enabled: bool,
+
+    /// Number of preferred locations available.
+    pub(super) preferred_location_count: usize,
+
+    /// Regions excluded from routing.
+    pub(crate) excluded_regions: Option<Vec<RegionName>>,
+
+    /// Whether this is a metadata (control-plane) operation.
+    pub(super) is_metadata: bool,
```
analogrelay: Same comment here. Don't store things like this that can be trivially computed from the OperationType

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -28,81 +38,291 @@ const SUCCESS_CODES: [u16; 101] = {
...
+            let endpoint = self
+                .global_endpoint_manager
+                .resolve_endpoint_from_components(&routing, &op);
+
+            routing.resolved_endpoint = Some(endpoint.clone());
+
+            // Stamp the resolved endpoint on the request for wire conversion
+            cosmos_request
+                .request_context
+                .route_to_location_endpoint(endpoint.clone());
```
analogrelay: Couldn't the `RoutingState` just be a field on `CosmosRequest` instead of a separate local? That way it's available here AND when actually sending the request.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -28,81 +38,291 @@ const SUCCESS_CODES: [u16; 101] = {
...
+                let resource_link = cosmos_request.resource_link.clone();
+                let mut raw_req = cosmos_request.clone().into_raw_request();
+                let success_options = CheckSuccessOptions {
+                    success_codes: &SUCCESS_CODES,
+                };
+                let pipeline_send_options = PipelineSendOptions {
+                    skip_checks: false,
+                    check_success: success_options,
+                };
                 let ctx_owned = ctx.with_value(resource_link).into_owned();
```
analogrelay, line 171: I think we can probably remove this, if we're controlling the pipeline. It exists so that the authorization policy can be a standard `azure_core::http::Policy` (and the only way to carry data through the pipeline to that policy is through the context). But we could just have our own Authorization step in our pipeline that sets the header directly on the Request generated by the final `into_raw_request`

analogrelay: But that can be a separate refactoring, to keep this simple.

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -28,81 +38,291 @@ const SUCCESS_CODES: [u16; 101] = {
...
+    } else {
+        gem.mark_endpoint_unavailable_for_write(endpoint);
     }
+    _ = gem.refresh_location(force_refresh).await;
+}
+
+/// Extracts HTTP status and Cosmos sub-status from a result.
+fn extract_status_info(
+    result: &azure_core::Result<RawResponse>,
+) -> (Option<azure_core::http::StatusCode>, Option<SubStatusCode>) {
```
analogrelay: I think we should build a `CosmosStatus` newtype that pairs a `StatusCode` with a `SubStatusCode`. That also allows us to have proper string representations for sub-status codes that are overloaded between different status-codes.

sdk/cosmos/azure_data_cosmos/src/pipeline/retry_decisions.rs
```diff
@@ -0,0 +1,631 @@
...
+
+use crate::constants::SubStatusCode;
+use azure_core::http::StatusCode;
+use azure_core::time::Duration;
+use url::Url;
+
+use super::components::{OperationInfo, RetryState, RoutingState, ThrottleState};
+
+/// Default retry interval between attempts (milliseconds).
+const RETRY_INTERVAL_MS: i64 = 1000;
```
analogrelay: nit: Use `Duration` for all times, it's the same storage, just a friendlier newtype.

sdk/cosmos/azure_data_cosmos/src/pipeline/retry_decisions.rs
```diff
@@ -0,0 +1,631 @@
...
+            retry.failover_count += 1;
+        }
+        RetryDecision::RetryOnWriteEndpoint { .. } => {
+            retry.session_retry_count += 1;
+        }
+        RetryDecision::RetrySameEndpoint { .. } | RetryDecision::Done | RetryDecision::Abort => {}
+    }
+    retry
+}
+
```
analogrelay: I could see these `apply` functions being defined ON `RetryState`. They're clear transforms that all take the current state as the first parameter. That's where a `mut self` parameter really makes sense, and it's still in keeping with DOP principles, because you're really just doing it for syntax organization and aesthetics

sdk/cosmos/azure_data_cosmos/src/pipeline/retry_decisions.rs
```diff
@@ -0,0 +1,631 @@
...
+    /// Non-retryable — return the result to the caller.
+    Abort,
+}
+
+/// Evaluates a data-plane request outcome and returns a retry decision.
+///
+/// This is a pure function: it reads the current state without mutating it.
+/// The caller is responsible for applying state updates based on the returned
+/// decision via [`apply_data_plane_decision`].
+pub(super) fn decide_data_plane_retry(
```
analogrelay: I love this pattern of `decide_` and `apply_`. For a little more organization, you could make it `RetryDecision::for_data_plane(...)` and then `RetryState::apply_...` (as noted below). But the principles here are absolutely sound, that's just organization and aesthetics. Starting off with bare functions is good DOP design, but since all the `decide_` functions return a `RetryDecision`, it makes sense to group them, and similarly for the `apply_` functions that transform a `...State` object.

sdk/cosmos/azure_data_cosmos/src/pipeline/retry_decisions.rs
```diff
@@ -0,0 +1,631 @@
...
+        let retry = RetryState::default();
+        let op = write_op(false, 2);
+        let decision = decide_data_plane_retry(
+            StatusCode::Forbidden,
+            Some(SubStatusCode::WRITE_FORBIDDEN),
+            &retry,
+            &op,
+        );
+        assert!(matches!(decision, RetryDecision::RetryNextRegion { .. }));
+    }
```
analogrelay, line 391: These tests are such a good illustration of how DOP makes all this so much clearer. We can provide inputs and assert the decision without having to set up a whole pipeline.

sdk/cosmos/azure_data_cosmos/src/pipeline/retry_decisions.rs
```diff
@@ -0,0 +1,631 @@
...
+        if retry.failover_count > 0 {
+            Duration::milliseconds(RETRY_INTERVAL_MS)
+        } else {
+            Duration::ZERO
+        }
+    } else {
+        Duration::milliseconds(RETRY_INTERVAL_MS)
+    };
+
+    RetryDecision::RetryNextRegion { delay }
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/pipeline/components.rs
```diff
@@ -0,0 +1,186 @@
...
+            max_attempts,
+            cumulative_wait: Duration::ZERO,
+            max_wait: Duration::seconds(max_wait_seconds),
+        }
+    }
+}
+
+impl Default for ThrottleState {
+    fn default() -> Self {
+        Self::new(5, 10)
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/pipeline/retry_decisions.rs
```diff
@@ -0,0 +1,631 @@
...
+    let backoff_ms = retry_after_ms.unwrap_or(DEFAULT_THROTTLE_BACKOFF_MS);
+    let delay = Duration::milliseconds(backoff_ms);
+
+    let new_cumulative = throttle.cumulative_wait + delay;
+    if new_cumulative > throttle.max_wait {
+        return RetryDecision::Abort;
+    }
+
+    RetryDecision::RetrySameEndpoint { delay }
+}
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -28,81 +38,291 @@ const SUCCESS_CODES: [u16; 101] = {
...
-        // but the transport isn't injected with the fault injection client
-        if self.fault_injection_enabled {
-            cosmos_request.add_fault_injection_headers();
+            // Update routing state on the request_context for the next attempt
+            cosmos_request.request_context.clear_route_to_location();
+            cosmos_request
+                .request_context
+                .route_to_location_index(routing.location_index, routing.use_preferred_locations);
         }
+    }
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
FabianMeiswinkel: Yes - will do this when moving to driver

sdk/cosmos/azure_data_cosmos/src/pipeline/retry_decisions.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/pipeline/retry_decisions.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/pipeline/retry_decisions.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/pipeline/components.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/pipeline/components.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/pipeline/components.rs
FabianMeiswinkel: Fixed - also see cosmos-design-struct.md for details

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
```diff
@@ -28,81 +38,291 @@ const SUCCESS_CODES: [u16; 101] = {
...
+                    } else {
+                        retry = apply_data_plane_decision(retry, &decision);
+                    }
+
+                    routing = mark_endpoint_failed(routing, &endpoint);
+                    routing = apply_routing_for_next_region(
+                        routing,
+                        &retry,
+                        status != Some(azure_core::http::StatusCode::Forbidden),
+                    );
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/pipeline/mod.rs
FabianMeiswinkel: I like the idea - won't fully make this change on this branch thought - but will do so, when folding this into the driver.

PR #3752, opened 2026-02-14
Connection Errors Retry Policy

tvaron3:
# Connection error retry policies for Cosmos DB client

   Adds transport-level error classification and retry logic to the Cosmos DB client retry policy, enabling safe retries for connection failures (reads **and** writes) and read-only retries for response timeouts. Also
  configures default HTTP client timeouts for `reqwest`.

   ## Motivation

   Previously, the retry policy only handled HTTP-level errors (status codes like 429, 503, etc.). Transport-level failures — connection refused, DNS errors, response timeouts — were not classified or retried, causing
  unnecessary request failures especially in multi-region deployments.

   ## Architecture

   ### Request Sent Status Classification

   ```
   ┌──────────────────────────────────────────────────────┐
   │              azure_core::Error received               │
   └──────────────────────┬───────────────────────────────┘
                          │
                          ▼
              ┌───────────────────────┐
              │  Walk source() chain  │
              │  for reqwest::Error   │
              └───────────┬───────────┘
                          │
               ┌──────────┼──────────┐
               ▼          ▼          ▼
        is_connect()  is_timeout() is_decode()/
            │             │        is_status()
            ▼             ▼          │
        ┌────────┐  ┌─────────┐     ▼
        │NotSent │  │ Unknown │  ┌──────┐
        └────────┘  └─────────┘  │ Sent │
                                 └──────┘
   ```

   - **`NotSent`** — request never reached the server (connection refused, DNS failure). Safe to retry both reads and writes.
   - **`Sent`** — response was received (decode error, redirect, status). Write may have been applied.
   - **`Unknown`** — ambiguous (response timeout, body error). Only safe to retry reads.

   ### Connection Failure Retry Flow

   ```
   ┌─────────────────────┐
   │  Connection Error    │
   │  (NotSent)           │
   └─────────┬───────────┘
             │
             ▼
   ┌─────────────────────┐     ≤ 3 retries
   │  Retry same endpoint │◄──────────────┐
   │  (1s delay)          │               │
   └─────────┬───────────┘               │
             │ > 3 retries                │
             ▼                            │
   ┌─────────────────────┐     transient? │
   │  Mark endpoint       │───── yes ─────┘
   │  unavailable (R+W)   │
   └─────────┬───────────┘
             │
             ▼
   ┌─────────────────────┐
   │  Refresh locations   │
   │  Reset retry counter │
   └─────────┬───────────┘
             │
             ▼
   ┌─────────────────────┐
   │  Failover to next    │
   │  available endpoint  │
   └─────────────────────┘
   ```

   ### Response Timeout Retry Flow

   ```
   ┌─────────────────────────┐
   │   Response Timeout       │
   │   (Unknown sent status)  │
   └──────────┬──────────────┘
              │
        ┌─────┴─────┐
        │            │
     Read?        Write?
        │            │
        ▼            ▼
     Retry       DoNotRetry
     (service     (write may
     unavailable   have been
     path)         applied)
   ```

   ## Changes

   ### Retry Policy (`retry_policies/`)
   - **`RequestSentStatus` enum** and **`RequestSentExt` trait** — classifies errors as `NotSent`, `Sent`, or `Unknown` by walking the error source chain and inspecting `reqwest::Error` flags
   - **`should_retry_on_connection_failure()`** — retries up to 3 times on the same endpoint, then marks it unavailable and fails over
   - **`should_retry_error()`** — checks `RequestSentStatus` before HTTP status codes, enabling transport-level retry decisions
   - WASM fallback via string analysis when `reqwest::Error` downcast is unavailable

   ### Fault Injection (`fault_injection/`)
   - New error types: `ConnectionError` and `ResponseTimeout` in `FaultInjectionErrorType`
   - Generates **real `reqwest::Error` instances** via actual network I/O:
     - Connection errors → TCP connect to `127.0.0.1:1` (refused)
     - Response timeouts → TCP connect to a local black-hole listener (accepts but never responds)

   ### HTTP Client Defaults (`constants.rs`, `cosmos_client.rs`)
   - Default timeouts when no custom transport is provided:
     - **Connection timeout**: 1s 
     - **Request timeout**: 65s
   - `reqwest` promoted from dev-dependency to full dependency with `azure_core` `reqwest` feature enabled


sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -177,3 +178,164 @@ fn get_substatus_code_from_response(response: &RawResponse) -> Option<SubStatusC
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn wrap_reqwest_error(err: reqwest::Error) -> azure_core::Error {
+        azure_core::Error::with_error(ErrorKind::Io, err, "failed to execute `reqwest` request")
+    }
+
+    #[tokio::test]
+    async fn reqwest_connect_error_is_not_sent() {
```
tvaron3: Remove all the tests except this one

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -177,3 +178,164 @@ fn get_substatus_code_from_response(response: &RawResponse) -> Option<SubStatusC
...
+            _ => RequestSentStatus::Unknown,
+        }
+    }
+}
+
+/// Determines [`RequestSentStatus`] from a [`reqwest::Error`].
+///
+/// Maps reqwest error inspection methods in priority order:
+/// - `is_timeout()`: `Unknown` — could be connect timeout (not sent) or read timeout (sent)
+/// - `is_connect()` / `is_request()`: `NotSent` — pre-send failures
```
tvaron3: Fix this stale doc

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -19,6 +19,8 @@
 
 ### Other Changes
 
+- Added default HTTP client timeouts and added retries for connection errors.
```
tvaron3: The pr number is #3752 add the pr link follow format in this file

Cargo.lock
tvaron3, line 1065: revert the changes in this file

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -29,6 +31,44 @@ pub struct CosmosClient {
...
+                .map(|h| std::borrow::Cow::Borrowed(h.as_str()))
+                .collect(),
+            additional_allowed_query_params: vec![],
+        };
+
+        if client_options.transport.is_none() {
+            // There is also a read timeout but this is addressed by the total timeout
+            let http_client = reqwest::ClientBuilder::new()
+                .connect_timeout(DEFAULT_CONNECTION_TIMEOUT)
+                .timeout(DEFAULT_REQUEST_TIMEOUT)
```
tvaron3, line 62: Remove the pool idle timeout and the tcp keep alive

sdk/cosmos/azure_data_cosmos/src/fault_injection/http_client.rs
```diff
@@ -149,6 +182,17 @@ impl FaultClient {
             None => return None, // No error type set, pass through
         };
 
+        // Connection-level faults produce real reqwest errors via actual failing requests.
+        match error_type {
```
tvaron3: Refactor this to return in each case and make a function for creating the error to reuse in all the cases

sdk/cosmos/azure_data_cosmos/src/constants.rs
```diff
@@ -203,6 +203,31 @@ pub(crate) const PREFER_MINIMAL: HeaderValue = HeaderValue::from_static("return=
...
+pub(crate) const DEFAULT_CONNECTION_TIMEOUT: std::time::Duration =
+    std::time::Duration::from_secs(1);
+
+/// Default overall request timeout (65s).
+///
+pub(crate) const DEFAULT_REQUEST_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(65);
+
+/// Default connection pool idle timeout (90s).
+///
+/// Balances avoiding stale connections with minimizing unnecessary reconnects.
```
tvaron3: Remove pool default and tcp keep alive default

sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_fault_injection.rs
```diff
@@ -973,7 +994,268 @@ pub async fn fault_injection_enable_disable_rule() -> Result<(), Box<dyn Error>>
...
         },
-        Some(TestOptions::new().with_fault_injection_builder(fault_builder)),
+        Some(TestOptions::new().with_fault_client_options(fault_options)),
+    )
+    .await
+}
+
+/// Test that ConnectionError faults on reads are retried and succeed after hit_limit.
+#[tokio::test]
+pub async fn fault_injection_connection_error_read_retries() -> Result<(), Box<dyn Error>> {
```
tvaron3: Remove all the tests in this file

sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -14,32 +14,34 @@ keywords = ["sdk", "azure", "rest", "cloud", "cosmos"]
 categories = ["api-bindings"]
 
 [dependencies]
-async-lock.workspace = true
 async-trait.workspace = true
-azure_core.workspace = true
+azure_core = { workspace = true, features = ["reqwest"] }
```
tvaron3: Are the changes in this file necessary?

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
tvaron3: Done. Removed all tests except `reqwest_connect_error_is_not_sent` — the remaining tests are covered by integration tests and the `client_retry_policy` unit tests.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
tvaron3: Done. Updated the doc comment to accurately reflect the actual priority order: `is_connect()` first, then `is_timeout()`, then `is_decode()/is_redirect()/is_status()`.

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
tvaron3: Done. Added the PR link: `([#3752](https://github.com/Azure/azure-sdk-for-rust/pull/3752))`

Cargo.lock
tvaron3: Done. Reverted `Cargo.lock` to main via `git checkout main -- Cargo.lock`. The build regenerates it correctly from the workspace deps.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
tvaron3: Done. Removed `pool_idle_timeout` and `tcp_keepalive` from the reqwest builder. Only `connect_timeout` and `timeout` remain.

sdk/cosmos/azure_data_cosmos/src/constants.rs
tvaron3: Done. Removed `DEFAULT_POOL_IDLE_TIMEOUT` and `DEFAULT_TCP_KEEPALIVE` constants.

sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_fault_injection.rs
tvaron3: Done. Removed all new emulator tests (connection error read/write retries, response timeout read/write). The multi-write fault injection tests still cover these scenarios.

sdk/cosmos/azure_data_cosmos/Cargo.toml
tvaron3: Yes, these changes are necessary:
1. `reqwest.workspace = true` — `cosmos_client.rs` directly uses `reqwest::ClientBuilder` to configure default timeouts, and `retry_policies/mod.rs` downcasts to `reqwest::Error` for sent-status classification. Both are production code paths, not behind a feature flag.
2. `azure_core = { features = ["reqwest"] }` — required so `azure_core` exposes the reqwest-based `HttpClient` impl that wraps our configured `reqwest::Client` as `Arc<dyn HttpClient>`.

sdk/cosmos/azure_data_cosmos/src/fault_injection/http_client.rs
tvaron3: Done. Merged both match blocks into a single match on error_type. ConnectionError and ResponseTimeout return early via generate_connection_error/generate_timeout_error; the remaining HTTP-level variants fall through to build the synthetic error response. No more unreachable arms.

sdk/cosmos/azure_data_cosmos/src/fault_injection/http_client.rs
```diff
@@ -209,6 +250,57 @@ impl FaultClient {
...
+    }
+
+    /// Generates a real `reqwest::Error` with `is_timeout() == true` by connecting
+    /// to a local TCP listener that accepts but never responds.
+    async fn generate_timeout_error(&self) -> azure_core::Error {
+        let addr = self.timeout_listener_addr;
+
+        // Ensure the listener is ready and a TCP connection can be established
+        // before we issue the timed request. This prevents the timeout from firing
+        // during the connect phase (which would produce is_request(), not is_timeout()).
```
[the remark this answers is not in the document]
tvaron3: Done. Fixed the comment to say `is_connect()` instead of `is_request()`.

sdk/cosmos/azure_data_cosmos/src/fault_injection/http_client.rs
```diff
@@ -44,9 +52,34 @@ impl FaultClient {
...
+        // The thread runs as a daemon — it will be cleaned up when the process exits.
+        std::thread::spawn(move || {
+            for stream in listener.incoming() {
+                match stream {
+                    Ok(_stream) => {
+                        // Hold the connection open: park the stream in a thread that
+                        // sleeps forever so the TCP connection stays established.
+                        std::thread::spawn(move || {
+                            std::thread::sleep(Duration::from_secs(u64::MAX));
+                        });
```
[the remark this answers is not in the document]
tvaron3: Done. Replaced the thread-per-connection pattern with a `Vec<TcpStream>` that retains accepted connections without spawning additional threads.

sdk/cosmos/azure_data_cosmos/src/constants.rs
```diff
@@ -203,6 +203,22 @@ pub(crate) const PREFER_MINIMAL: HeaderValue = HeaderValue::from_static("return=
 
 pub const ACCOUNT_PROPERTIES_KEY: &str = "account_properties_key";
 
+// Default HTTP client timeouts.
+// See `next_generation_sdks_design_principles.md` for design rationale.
+
+/// Default TCP connection timeout (500ms).
+/// 500 ms it times out locally.
```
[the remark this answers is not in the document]
tvaron3: Done. Fixed the doc comment to say "1s" and "After 1 second" to match the actual `from_secs(1)` value.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -177,3 +178,126 @@ fn get_substatus_code_from_response(response: &RawResponse) -> Option<SubStatusC
...
+}
+
+/// WASM fallback: reqwest doesn't expose error type inspection methods on WASM,
+/// so we fall back to string analysis which is less reliable.
+#[cfg(target_arch = "wasm32")]
+fn wasm_request_sent_status(error: &azure_core::Error) -> RequestSentStatus {
+    let msg = error.to_string().to_lowercase();
+
+    // Connection-related errors (before sending)
+    if msg.contains("dns") || msg.contains("connection refused") || msg.contains("connect") {
```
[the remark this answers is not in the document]
tvaron3: Done. Replaced the broad `"connect"` pattern with more specific matches: `"connection refused"`, `"connection error"`, and `"connection timed out"` to avoid false positives.

sdk/cosmos/azure_data_cosmos/src/constants.rs
```diff
@@ -203,6 +203,22 @@ pub(crate) const PREFER_MINIMAL: HeaderValue = HeaderValue::from_static("return=
...
+///
+/// Aggressive default per design doc: fast failure on downed nodes improves P9x latency.
+#[cfg(not(target_arch = "wasm32"))]
+pub(crate) const DEFAULT_CONNECTION_TIMEOUT: std::time::Duration =
+    std::time::Duration::from_secs(1);
+
+/// Default overall request timeout (65s).
+///
```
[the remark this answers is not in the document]
tvaron3: Done. Added rationale explaining the 65s timeout is chosen to remain above typical 60s service timeouts while allowing retry attempts, with a reference to the design doc.

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -19,6 +19,8 @@
 
 ### Other Changes
 
+- Added default HTTP client timeouts and added retries for connection errors. ([#3752](https://github.com/Azure/azure-sdk-for-rust/pull/3752))
```
heaths, line 23: You've also made a breaking change by taking a strict dependency on `reqwest`. I think, at the very least, you need to add your own `reqwest` feature to mimc what we have in `azure_core` and make that conditional. For first-party support, they cannot and will not use `reqwest`.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
```diff
@@ -177,3 +178,142 @@ fn get_substatus_code_from_response(response: &RawResponse) -> Option<SubStatusC
...
+        }
+
+        // WASM: reqwest errors don't support downcast, so fall back to string analysis.
+        #[cfg(target_arch = "wasm32")]
+        {
+            let status = wasm_request_sent_status(self);
+            if status != RequestSentStatus::Unknown {
+                return status;
+            }
+        }
```
heaths: As stated offline, I'm open to doing this behind our own `reqwest` feature in core to define a separate, say, `ErrorKind::Connection`. That should simplify the dependencies here and could be actionable for any dev using any service. Cosmos could still have a custom retry policy or whatever you feel necessary then based on whether it's an `ErrorKind::Connection` or `ErrorKind::HttpResponse` then. How about that?

simorenoh (approved):

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
tvaron3: I put reqwest behind a feature flag and made it on by default.

sdk/cosmos/azure_data_cosmos/src/retry_policies/mod.rs
tvaron3: Added the logic for ErrorKind:ConnectionAborted and ErrorKind timeout and I think those are enough to dictate whether to retry or not

sdk/cosmos/azure_data_cosmos/src/fault_injection/http_client.rs
```diff
@@ -44,9 +52,33 @@ impl FaultClient {
...
+            let mut _connections: Vec<std::net::TcpStream> = Vec::new();
+            for stream in listener.incoming() {
+                match stream {
+                    Ok(stream) => {
+                        _connections.push(stream);
+                    }
+                    Err(_) => continue,
+                }
+            }
+        });
```
analogrelay: This feels a little messy. Why do the connections need to out-live the `FaultClient`?

sdk/core/typespec/src/error/mod.rs
```diff
@@ -26,6 +26,14 @@ pub enum ErrorKind {
...
         raw_response: Option<Box<RawResponse>>,
     },
+    /// A connection to the server could not be established.
+    ///
+    /// The request was never sent, so it is safe to retry both reads and writes.
+    ConnectionAborted,
+    /// The operation timed out before completing.
+    ///
+    /// It is unknown whether the server received the request.
+    Timeout,
```
heaths: Too specific. We just want variants that are actionable. Details can be in any sort of error message if necessary, but, practically, they rarely are. When we reduced the number of variants, we asked ourselves: what kinds are actionable, and what can you do? If the connection fails - whether it was closed by the remote or timed out - you can retry later or inform the user and let them decide. That's it. So I think a single `Connection` error kind is all that's required here.

heaths: I'll add, though: does Cosmos need to know the difference between these two?

sdk/cosmos/azure_data_cosmos/src/fault_injection/http_client.rs
tvaron3: Changed it to just return a fabricated error instead of trying to create a real error and returning it. The reason I originally took this approach was because the reqwest error was completely private but now that it's abstracted by core I think I can just return an error with the errorkinds that core will throw.

sdk/core/typespec/src/error/mod.rs
tvaron3: Agree no reason to differentiate. I removed the timeout error kind and left just Connection

sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -44,7 +44,7 @@ key_auth = [
 ] # Enables support for key-based authentication (Primary Keys and Resource Tokens)
 hmac_rust = ["azure_core/hmac_rust"]
 hmac_openssl = ["azure_core/hmac_openssl"]
-fault_injection = ["reqwest"] # Enables support for fault injection testing
+fault_injection = [] # Enables support for fault injection testing
```
heaths, line 47: Is this intended for devs using your crates as well? If so, it should be in the `packages.metadata.docs.rs` section below as well.

heaths: Huh. Seems that's grown stale anyway. You should also update that - maybe in another PR if fault injection is only for your team - to include `reqwest` and, I think, `key_auth` as well, right @analogrelay ?

tvaron3: fault injection is expected to be used by our customers as well. It is not just for us. I can do a follow up pr to update packages.metadata.docs.rs depending on what @analogrelay thinks

analogrelay: Yep, this, and `key_auth` should go in `packages.metadata.docs.rs`

PR #3753, opened 2026-02-15
Skill to enforce struct design decisions

FabianMeiswinkel:
# Add Cosmos SDK struct design validation skill

## Summary

Adds a new [Copilot Agent Skill](https://code.visualstudio.com/docs/copilot/customization/agent-skills) that enforces consistent struct design conventions across `sdk/cosmos` crates. The skill validates visibility modifiers, field privacy, `#[non_exhaustive]` usage, required-field constructors (`new()`), conditional `Default` correctness, and builder pattern (`with_*` setters) on public structs. It can auto-fix violations or report them as errors — giving contributors a single command to audit and align struct design before committing or during code review.

## What's new

New file: `sdk/cosmos/.github/skills/cosmos-design-struct.md`

A declarative skill definition with three arguments:

| Argument | Type | Default | Description |
|---|---|---|---|
| `scope` | string | `all` | Crate to validate (e.g., `azure_data_cosmos`, `azure_data_cosmos_native`) or `all` for everything under `sdk/cosmos` |
| `auto-fix` | boolean | `false` | When `true`, automatically fix violations (adjust visibility, add `#[non_exhaustive]`, generate getters/setters, add/remove `Default`, generate `new()` constructors, update call sites). When `false`, only report violations with proposed changes. |
| `changed-only` | boolean | `true` | When `true`, only scan `.rs` files that differ between the current local branch and `main` (via `git diff --name-only main`). This limits work when the skill is triggered automatically. When `false`, scan all `.rs` files under the target path. |

The skill follows the [Agent Skills open standard](https://agentskills.io/) and sets `disable-model-invocation: false` since it requires model reasoning to classify structs and generate accessor methods.

## How it works

When triggered, the skill instructs the agent to:

1. **Resolve the target path** from the `scope` argument (specific crate or all of `sdk/cosmos`)
2. **Determine file scope** — either all `.rs` files under the target path (`changed-only=false`) or only files that differ from `main` (`changed-only=true`, the default). Always skips `generated/` subdirectories.
3. **Scan all struct declarations** in the resolved files
4. **Classify each struct** into one of three categories:
   - **Truly public** — `pub` and all ancestor modules are also `pub` (reachable from outside the crate)
   - **Effectively scoped** — `pub` but inside a non-public module (not externally reachable)
   - **Internal** — used only within its declaring module or submodule
5. **Apply visibility rules** — ensure struct and field visibility matches intended scope
6. **Apply truly-public struct rules** — enforce `#[non_exhaustive]`, private fields with getters, `with_*` builder setters, required-field constructors, and conditional `Default`
7. **Auto-fix or report** violations depending on the `auto-fix` argument
8. **Produce a summary** highlighting public surface area changes and breaking change warnings

### For **truly public** structs

- **`#[non_exhaustive]`** is required (forward compatibility when fields are added)
  - Technically, `#[non_exhaustive]` would not be needed if any field is non-public. The reason for still adding `#[non_exhaustive]` consistently is motivated by wanting to use the clear semver signal (and protect against situations where accidentally the last non-public field is removed from a struct).
- **Fields must never be `pub`** — use private fields with getter methods
  - If crate-internal code outside the defining module needs direct field access (mutation, nested traversal), use `pub(crate)` instead of fully public
- **Getter methods** return `&T` for non-`Copy` types, `T` for `Copy` types
- **`with_*` setter methods** use the consuming builder pattern (`with_<field_name>(mut self, value) -> Self`) for **optional** fields only

### Required-field constructors (Step 6f)

Structs with required fields (e.g., `ContainerProperties`, `DatabaseProperties`) must:

1. Provide a `pub fn new(...)` constructor taking all required fields
2. **Not** derive or implement `Default` — a zero value would be semantically invalid
3. **Not** provide `with_*` setters for required fields — they are set exclusively via `new()`
4. Still provide `with_*` setters for all optional fields, and getters for all fields

**Determining required fields:**
- For new structs: the developer specifies them
- For existing structs: inferred from doc comments ("required"), non-`Option` fields with meaningless defaults (e.g., `id: ""`), server rejection of defaults, or universal call-site patterns

**Example:**
```rust
// All-optional struct — uses Default + with_*
let options = ItemOptions::default()
    .with_session_token(token);

// Struct with required fields — uses new() + with_* for optional
let props = ContainerProperties::new("myContainer", "/partitionKey")
    .with_default_ttl(Duration::from_secs(3600));
```

### Conditional `Default` (Step 6e)

`Default` is required **only** when every field is optional. If a struct has any required field, `Default` is removed — Rust convention is that `Default::default()` must produce a semantically valid value.

### For **effectively scoped** and **internal** structs

- Use the most restrictive visibility that compiles
  - If a struct is `pub` inside a non-pub module, the skill enforces the effective visibility modifier (e.g., `pub(crate)`) on the struct itself. This makes visibility obvious at the declaration site and prevents accidental exposure if the module later becomes public.
- Omit `#[non_exhaustive]` (unnecessary when external code cannot reach the struct)
- Fields can use `pub` without further restriction (struct-level visibility already constrains access)

### Exemptions

- **Newtype structs** are exempt from `#[non_exhaustive]`, private fields, getters/setters, and `Default` rules
- **Options structs** (e.g., `ItemOptions`, `QueryOptions`) are exempt from the required-field constructor rule — all their fields are optional by design

## Usage examples

In VS Code Copilot Chat (agent mode), the skill is automatically picked up when the task matches its description. You can also invoke it explicitly:

Validate all Cosmos structs (report only, scanning changed files):
```
/cosmos-design-struct
```

Validate a specific crate, all files:
```
/cosmos-design-struct scope=azure_data_cosmos changed-only=false
```

Auto-fix violations across all crates, changed files only:
```
/cosmos-design-struct auto-fix=true
```

Full audit with auto-fix across all files:
```
/cosmos-design-struct scope=all auto-fix=true changed-only=false
```

## Why a skill instead of a prompt file?

- **On-demand activation**: Skills are loaded automatically when the agent determines the task matches the skill description, so contributors don't need to know about or remember a slash command.
- **Cross-tool compatibility**: Skills follow an open standard and work across VS Code, GitHub Copilot CLI, and the GitHub Copilot coding agent.
- **Scoped to Cosmos**: The skill lives under `sdk/cosmos/.github/skills/`, keeping it discoverable and co-located with the code it validates.

## Pre-Requisites

- Enable skills in VS Code: Settings → `chat.useAgentSkills`
- Add cosmos skills: Settings → `chat.agentSkillsLocations` → add the `sdk/cosmos/.github/skills` folder

## Testing

- Ran the skill against all of `sdk/cosmos` with `auto-fix=false` to validate violation reporting
- Ran the skill against all of `sdk/cosmos` with `auto-fix=true changed-only=false` to validate automated fixes across all structs, including:
  - Adding `new()` constructors to 10 structs with required fields
  - Removing `Default` from structs with required fields
  - Removing `with_*` setters for required fields
  - Updating ~48 call sites across source and test files
- Ran `cosmos-pre-commit-validation changed-only=false` — all 7 checks pass (fmt, build, clippy, docs, unit tests, emulator cfg check, multi-write cfg check)

## Related

- [VS Code Agent Skills docs](https://code.visualstudio.com/docs/copilot/customization/agent-skills)
- [Agent Skills open standard](https://agentskills.io/)
- Cosmos-specific instructions: `sdk/cosmos/AGENTS.md`, `.github/instructions/cosmos.instructions.md`
- Companion skill: `sdk/cosmos/.github/skills/cosmos-pre-commit-validation.md`

### DIFF
--- a/sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
+++ b/sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
@@ -121,7 +121,22 @@ For truly public structs, require `#[non_exhaustive]` **when all named fields ar
 
 If one or more fields are non-public, `#[non_exhaustive]` is optional and typically redundant for construction control.
 
-#### b) Field visibility on truly public structs — choose based on validation/invariant needs
+#### b) Prefer type-system enforcement for validation and invariants
+
+Before making a field non-public solely to validate it in a `with_*` setter or constructor, evaluate whether the **type system** can enforce the invariant instead. Newtypes, enums, and other constrained wrapper types make invalid states unrepresentable at compile time, which is stronger and more ergonomic than runtime checks.
+
+**Decision order** (prefer earlier options):
+
+1. **Enum** — When the domain is a closed set of named values (e.g., `ConsistencyLevel`, `IndexingMode`), use an enum. Invalid variants simply cannot be expressed.
+2. **Newtype with construction-time validation** — When values must satisfy a constraint (range, format, normalization), introduce a newtype that validates or normalizes in `new()` / `From` and guarantees the invariant internally (e.g., `RegionName` normalizes casing on construction; `SubStatusCode` wraps a raw number with parse validation).
+3. **Newtype for semantic clarity** — Even without a hard constraint, a newtype can prevent accidental misuse of stringly-typed or primitive fields (e.g., wrapping `String` as `DatabaseId` to avoid mixing up identifiers).
+4. **Runtime validation in `with_*`/constructor** — Use this as a **last resort**, only when type-level enforcement is impractical (e.g., cross-field invariants that span multiple values, constraints that depend on external state, or cases where introducing a new type would create excessive API friction for negligible safety gain).
+
+When a newtype is introduced, the field holding it can often remain `pub` because the invariant is encoded in the type itself — external code cannot construct an invalid value regardless of field visibility. See subsection (g) for newtype struct conventions.
+
+> **Rule of thumb**: If a `with_*` setter contains a `.clamp()`, range check, format validation, or normalization, that logic almost certainly belongs in a newtype's constructor instead.
+
+#### c) Field visibility on truly public structs — choose based on validation/invariant needs
 
 Fields on truly public structs may be `pub` **or** non-public. Choose visibility by checking (1) whether validation/invariant enforcement is needed and (2) whether crate-internal code in other modules needs direct non-getter access:
 
@@ -137,17 +152,18 @@ Fields on truly public structs may be `pub` **or** non-public. Choose visibility
 **How to determine the correct visibility**:
 
 1. Decide first whether the field requires invariant/validation enforcement.
-2. If no, `pub` is acceptable.
-3. If yes, keep it non-public and evaluate crate-internal access needs to choose private vs `pub(crate)`/`pub(super)`.
+2. If yes, check whether a newtype or enum can encode the invariant in the type (see subsection (b) above). If the type itself enforces the invariant, `pub` is acceptable.
+3. If the invariant cannot be encoded in a type, keep the field non-public and evaluate crate-internal access needs to choose private vs `pub(crate)`/`pub(super)`.
+4. If no invariant is needed, `pub` is acceptable.
 
-#### c) Getter methods for readable fields
+#### d) Getter methods for readable fields
 
 Every field that external consumers need to **read** must have a getter method:
 
 - Named after the field (e.g., `fn session_token(&self) -> &str`)
 - Returns `&T` for non-`Copy` types, or `T` for `Copy` types (e.g., `bool`, `u32`, `f64`)
 
-#### d) Construction APIs: required fluent pattern
+#### e) Construction APIs: required fluent pattern
 
 For truly public structs, construction APIs must be ergonomic without over-engineering. The baseline constructor and optional setters must follow this contract:
 
@@ -178,7 +194,7 @@ For `with_*` setters in **either** style (on the target type or on a builder), u
 
 Separate builder types are still optional. Fluent `with_*` support is required only when optional fields are present and the struct has more than one field.
 
-#### e) Required fields and setter placement
+#### f) Required fields and setter placement
 
 A **required field** is one that must be set for the struct to be semantically valid.
 
@@ -190,7 +206,7 @@ A **required field** is one that must be set for the struct to be semantically v
 
 When inferring required fields on existing structs, use docs, service behavior, and call-site usage patterns.
 
-#### f) Exemptions
+#### g) Exemptions
 
 **Newtype structs** are exempt from the named-field construction rules. Since a newtype wraps a single value, the full named-field struct rules do not apply. Instead, newtypes should:
 
@@ -203,7 +219,7 @@ When inferring required fields on existing structs, use docs, service behavior, 
 
 Serde derive macros (`Serialize`, `Deserialize`) work on private fields — no `pub(crate)` is needed for serde compatibility.
 
-#### g) Optional builder guidance (when used)
+#### h) Optional builder guidance (when used)
 
 If a separate builder type is used, follow these conventions:
 
@@ -212,6 +228,7 @@ If a separate builder type is used, follow these conventions:
 3. Provide `with_*` setters for optional fields.
 4. Provide terminal `build(self, ...) -> <Type>` (or `azure_core::Result<Type>` when fallible).
 5. Keep required fields on `build(...)`, not as optional builder state.
+6. Add `<Type>::builder(... required args ...) -> <Type>Builder` to initialize the builder type.
 
 These conventions apply only when a builder exists; they are not a requirement to introduce one.
 
@@ -222,12 +239,12 @@ These conventions apply only when a builder exists; they are not a requirement t
 1. Adjust visibility modifiers on structs and fields according to Steps 4–6.
 2. Add `#[non_exhaustive]` to truly public structs where all named fields are public and the attribute is missing.
 3. Remove `#[non_exhaustive]` from non-public structs that have it unnecessarily.
-4. For truly public structs, apply field visibility decisions per Step 6b rather than forcing all `pub` fields to non-public:
+4. For truly public structs, apply field visibility decisions per Step 6c (and type-system preferences per Step 6b) rather than forcing all `pub` fields to non-public:
   - If a field has no validation/invariant requirements, `pub` is allowed.
   - If invariants/validation are required, make the field non-public and provide/update constructor or `with_*` APIs as needed.
   - When non-public fields need ownership extraction without cloning, prefer `From`/`Into` trait-based conversion; add targeted `into_*` methods only when a trait-based API is not a good fit.
   - Generate getter methods for fields that external consumers need to read and that are non-public.
-5. Ensure each truly public struct has the required construction API (Step 6d), prioritizing rule compliance even if changes are semver-breaking:
+5. Ensure each truly public struct has the required construction API (Step 6e), prioritizing rule compliance even if changes are semver-breaking:
   a. Enforce mixed required+optional rules strictly (`new(required...)` or `build(required...)`) and remove/avoid `Default` in that case.
   b. If no ergonomic construction API exists, add the simplest valid option:
         - all-optional/simple types: `Default` + `with_*` on the target type.
@@ -304,6 +321,7 @@ Breaking changes include:
 - For ownership extraction from non-public fields, prefer standard `From`/`Into` traits first. Use targeted `into_*` methods as an exception when extracting a specific owned field without cloning is clearer than trait conversion.
 - Reference `sdk/cosmos/AGENTS.md` for canonical model, options, and builder patterns.
 - Do not skip required fixes to avoid semver-breaking outcomes; apply the rules and report all breaking changes clearly in the summary.
-- When generating **new** structs, apply these rules from the start — decide field visibility from invariants and API ergonomics up front, then keep construction APIs consistent with the field mix (all-optional: `Default` + `with_*`; mixed required+optional: `new` + `with_*`; required-only: simplest API, optional builder).
+- When generating **new** structs, apply these rules from the start — decide field visibility from invariants and API ergonomics up front, then keep construction APIs consistent with the field mix (all-optional: `Default` + `with_*`; mixed required+optional: `new` + `with_*`; required-only: simplest API, optional builder). Also evaluate whether any constrained fields warrant a newtype or enum rather than a bare primitive type (see Step 6b).
 - For **new** structs, explicitly ask the developer which fields are required if not obvious from the context.
 - For **existing** structs, infer required fields from: (1) doc comments mentioning "required", (2) server rejection of default values, (3) every call site always setting the field, (4) non-`Option` type with no semantically valid zero value.
+- When a field requires invariant enforcement, prefer encoding the invariant in the type system (newtypes, enums, constrained wrappers) over runtime validation in `with_*` setters or constructors. Setter-level validation (clamping, range checks, format normalization) is a last resort for invariants that cannot be practically expressed in the type system (see Step 6b).


sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
```diff
@@ -90,50 +83,222 @@ pub struct ContainerProperties {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(deserialize_with = "deserialize_ttl")]
     #[serde(serialize_with = "serialize_ttl")]
-    pub analytical_storage_ttl: Option<Duration>,
+    analytical_storage_ttl: Option<Duration>,
 
     /// A [`SystemProperties`] object containing common system properties for the container.
     #[serde(flatten)]
-    pub system_properties: SystemProperties,
+    pub(crate) system_properties: SystemProperties,
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/models/account_properties.rs
```diff
@@ -12,111 +12,244 @@ use serde_json::Value;
...
 #[derive(SafeDebug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
 #[safe(true)]
 #[serde(rename_all = "camelCase")]
 pub struct AccountRegion {
-    pub name: RegionName,
+    pub(crate) name: RegionName,
 
     #[serde(with = "crate::serde::url")]
-    pub database_account_endpoint: Url,
+    pub(crate) database_account_endpoint: Url,
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/models/account_properties.rs
```diff
@@ -12,111 +12,244 @@ use serde_json::Value;
...
-    pub dbs: String,
+    dbs: String,
 
     /// Regions currently accepting writes for the account (multi-master may yield >1).
-    pub writable_locations: Vec<AccountRegion>,
+    pub(crate) writable_locations: Vec<AccountRegion>,
 
     /// Regions from which the account can be read (includes writable regions plus any read regions).
-    pub readable_locations: Vec<AccountRegion>,
+    pub(crate) readable_locations: Vec<AccountRegion>,
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -30,38 +31,39 @@ impl Display for SessionToken {
...
+    enable_partition_level_circuit_breaker: bool,
+    disable_partition_level_failover: bool,
+    enable_upgrade_consistency_to_local_quorum: bool,
     /// The desired throughput bucket for the client
     ///
     /// See [Throughput Control in Azure Cosmos DB](https://learn.microsoft.com/azure/cosmos-db/nosql/throughput-buckets) for more.
-    pub throughput_bucket: Option<usize>,
-    pub session_retry_options: SessionRetryOptions,
+    throughput_bucket: Option<usize>,
+    pub(crate) session_retry_options: SessionRetryOptions,
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -218,61 +542,194 @@ impl Display for IndexingDirective {
...
+    /// When this value is true, write operations will respond with the new value of the resource being written.
+    pub(crate) enable_content_response_on_write: bool,
+    /// The desired throughput bucket for this request.
+    pub(crate) throughput_bucket: Option<usize>,
+    /// Priority based execution allows users to set a priority for each request.
+    pub(crate) priority: Option<PriorityLevel>,
+    /// Additional headers to be included in the request.
+    pub(crate) custom_headers: HashMap<HeaderName, HeaderValue>,
+    /// Regions to be skipped from regional routing preferences.
+    pub(crate) excluded_regions: Option<Vec<RegionName>>,
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/models/mod.rs
```diff
@@ -81,11 +103,29 @@ pub struct SystemProperties {
 #[safe(true)]
 pub struct DatabaseProperties {
     /// The ID of the database.
-    pub id: String,
+    id: String,
 
     /// A [`SystemProperties`] object containing common system properties for the database.
     #[serde(flatten)]
-    pub system_properties: SystemProperties,
+    pub(crate) system_properties: SystemProperties,
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/models/mod.rs
```diff
@@ -50,27 +50,49 @@ pub struct SystemProperties {
...
-    pub resource_id: Option<String>,
+    pub(crate) resource_id: Option<String>,
 
     /// A [`OffsetDateTime`] representing the last modified time of the resource.
     #[serde(default)]
     #[serde(rename = "_ts")]
     #[serde(skip_serializing)]
     #[serde(deserialize_with = "deserialize_cosmos_timestamp")]
-    pub last_modified: Option<OffsetDateTime>,
+    last_modified: Option<OffsetDateTime>,
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
```diff
@@ -0,0 +1,364 @@
...
+
+### Step 3 — Scan struct declarations
+
+- Find all `struct` declarations in the `.rs` files identified in Step 2.
+
+### Step 4 — Classify each struct
+
+Classify every struct into exactly one of these categories:
+
+1. **Truly public** — The struct is `pub` and **all** ancestor modules up to the crate root are also `pub` (the struct is reachable from outside the crate). These structs get the full set of rules: `#[non_exhaustive]`, private fields, getters, `with_*` setters, and either `Default` or `new()` depending on whether the struct has required fields.
```
simorenoh: I thought we *didn't* want to have the `with_` prefix on the setters used in our builders - asking because I see this guidance propagated to all the other files in the examples

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
```diff
@@ -0,0 +1,364 @@
+---
+name: cosmos-design-struct
+description: >
+  Enforce consistent struct design conventions across sdk/cosmos crates. Validates visibility modifiers,
+  field privacy, #[non_exhaustive] usage, builder pattern (with_* setters), required-field constructors (new()),
```
simorenoh: Same comment as below on the `with_` prefix on setters.

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
analogrelay: We need to make a call, but the general pattern I see in most places, and [in the Azure SDK Guidelines](https://azure.github.io/azure-sdk/rust_introduction.html#rust-builders), is to have the `with_` prefix. We technically don't need it if the builder is a separate type (since `ItemOptionsBuilder::session_token` is a setter and `ItemOptions::session_token` is a getter, and it's mostly clear which is which).

Given that, I think we should standardize on the `with_` prefix in builders.

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
```diff
@@ -0,0 +1,364 @@
...
+- Find all `struct` declarations in the `.rs` files identified in Step 2.
+
+### Step 4 — Classify each struct
+
+Classify every struct into exactly one of these categories:
+
+1. **Truly public** — The struct is `pub` and **all** ancestor modules up to the crate root are also `pub` (the struct is reachable from outside the crate). These structs get the full set of rules: `#[non_exhaustive]`, private fields, getters, `with_*` setters, and either `Default` or `new()` depending on whether the struct has required fields.
+
+2. **Effectively scoped** — The struct has a `pub` visibility modifier but lives inside a module that restricts visibility (e.g., `pub(crate) mod`, `pub(super) mod`). The struct is **not** reachable from outside the crate. These structs should:
+   - Have the **effective visibility** annotated explicitly on the struct (e.g., `pub(crate) struct Foo` not `pub struct Foo` inside a `pub(crate) mod`).
```
analogrelay, line 90: I need to do a little thinking on this, but I'm OK with it for now (especially since it really only impacts _us_ as engineers on the SDK team). I don't see this approach often in Rust, instead I tend to see more granular organization of modules, but I know that can throw up other roadblocks to understanding the code. For now, let's stick with this approach, but I'm going to do a little more research and see if we want to refine this approach later.

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
```diff
@@ -0,0 +1,364 @@
...
+pub fn with_session_token(mut self, value: impl Into<String>) -> Self {
+    self.session_token = Some(value.into());
+    self
+}
+```
+
+- Prefix: `with_<field_name>`
+- Takes `mut self`, returns `Self`
+- Use `impl Into<T>` for string-like and convertible parameters where appropriate
+- For structs with required fields (Step 6f), do **not** generate `with_*` setters for required fields — they are set exclusively via `new()`
```
analogrelay: The SDK guidelines pattern is to use a separate builder type. We should align with that unless we want to open that guideline up for change (which we absolutely _can_ do).

The main reason for a separate builder is that it keeps options types immutable. It does mean that if a user wants to use one set of options for a call and then make one tweak for a second call, they have to keep the Builder around instead of the Options type, but the impact of that is pretty minimal.

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
FabianMeiswinkel: Discussed offline - SDK guidelines require with_XXX prefix - https://azure.github.io/azure-sdk/rust_introduction.html#rust-builders

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
FabianMeiswinkel: Discussed offline - SDK guidelines require with_XXX prefix - https://azure.github.io/azure-sdk/rust_introduction.html#rust-builders

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
FabianMeiswinkel: This should never cause breakign changes - so, we should be able to iterate on it later

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
FabianMeiswinkel: ACK - I will move the builder to its own type. I assume worst case for convenience we could create a builder factory with an options instance in cases where keeping around the builder is too painful for customers?

FabianMeiswinkel: As discussed offline - we agreed that the overhead of having to use a separate Builder type is optional.

sdk/cosmos/azure_data_cosmos/src/fault_injection/result.rs
```diff
@@ -8,24 +8,46 @@ use std::time::Duration;
 use super::FaultInjectionErrorType;
 
 /// Represents a server error to be injected.
+#[non_exhaustive]
 #[derive(Clone, Debug)]
 pub struct FaultInjectionResult {
     /// The type of server error to inject.
-    pub error_type: Option<FaultInjectionErrorType>,
+    pub(crate) error_type: Option<FaultInjectionErrorType>,
```
FabianMeiswinkel: fields should be private here

sdk/cosmos/azure_data_cosmos/src/models/account_properties.rs
```diff
@@ -12,111 +12,244 @@ use serde_json::Value;
 use url::Url;
 
 /// Represents a single regional endpoint for the Cosmos DB account (readable or writable).
+#[non_exhaustive]
 #[derive(SafeDebug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
 #[safe(true)]
 #[serde(rename_all = "camelCase")]
 pub struct AccountRegion {
-    pub name: RegionName,
+    pub(crate) name: RegionName,
```
FabianMeiswinkel: fields should be private here

FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/fault_injection/result.rs
FabianMeiswinkel: Fixed

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
```diff
@@ -0,0 +1,278 @@
...
+Additional rules:
+
+- If a struct is marked `pub` but lives inside a non-public module (e.g., `pub struct Foo` inside `pub(crate) mod internal`), change the struct to use the **effective** visibility: `pub(crate) struct Foo`. This makes the actual visibility obvious at the struct declaration site without requiring the reader to trace module ancestry.
+- Fields on effectively-scoped or internal structs **can** use `pub` — the struct-level visibility already constrains access. This is an intentional choice to reduce repetitive `pub(crate)` or `pub(super)` annotations on fields while still making the effective visibility clear and easy to review from the struct declaration alone.
+
+### Step 6 — Apply truly-public struct rules
+
+These rules apply **only** to structs classified as **truly public** in Step 3:
+
+#### a) `#[non_exhaustive]` required
```
analogrelay: We should add an exception: If a type has private fields, `#[non_exhaustive]` is just noise and unnecessary. A private field already makes a struct unconstructable from outside the module.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -21,6 +21,7 @@ use azure_core::credentials::Secret;
 use azure_core::http::{LoggingOptions, RetryOptions};
 
 /// Client for Azure Cosmos DB.
+#[non_exhaustive]
```
analogrelay: This is a great example of a place where `#[non_exhaustive]` is unnecessary noise.

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
```diff
@@ -219,11 +220,11 @@ impl DatabaseClient {
...
         let db = self.read(None).await?.into_model()?;
         let resource_id = db
-            .system_properties
-            .resource_id
+            .system_properties()
+            .resource_id()
             .expect("service should always return a '_rid' for a database");
 
-        let offers_client = OffersClient::new(self.pipeline.clone(), resource_id);
+        let offers_client = OffersClient::new(self.pipeline.clone(), resource_id.to_owned());
```
analogrelay: This is an unfortunate side-effect of private fields. It's no longer possible to move out of our model types. Instead, we have to take the `&str` we get back from `resource_id()` and force it to be copied to a new `String`.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -70,7 +72,201 @@ pub struct CosmosClientOptions {
...
+    /// Returns `true` if partition level failover is disabled.
+    pub fn disable_partition_level_failover(&self) -> bool {
+        self.disable_partition_level_failover
+    }
+
+    /// Sets whether partition level failover is disabled.
+    pub fn with_disable_partition_level_failover(mut self, disabled: bool) -> Self {
+        self.disable_partition_level_failover = disabled;
+        self
+    }
```
analogrelay: I think the naming convention is a little clunky for enable/disable flags. `disable_partition_level_failover` feels like a _verb_ that modifies the value. I don't have a great answer for what to do differently though.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
FabianMeiswinkel: I requested in the skill to keep it for the explicit semver signal - as well as preventing accidentally exposing public struct without it when someone removes a private field. But given we have a skill now that will probably be used by agents for code review I can live with removing that - and it reduces indeed noise.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
FabianMeiswinkel: the name of the filed should have been partition_level_failover_disabled in the first place :-)

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -30,38 +31,39 @@ impl Display for SessionToken {
...
     // when the cosmos client options is changed to builder pattern, this shouldn't be exposed directly to customers
     // right now it is exposed behind a feature flag
     #[cfg(feature = "fault_injection")]
-    pub fault_injection_enabled: bool,
-    pub application_preferred_regions: Vec<RegionName>,
-    pub excluded_regions: Vec<RegionName>,
-    pub account_initialization_custom_endpoints: Option<HashSet<String>>,
+    pub(crate) fault_injection_enabled: bool,
+    pub(crate) application_preferred_regions: Vec<RegionName>,
+    pub(crate) excluded_regions: Vec<RegionName>,
```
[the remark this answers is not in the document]
FabianMeiswinkel: Addressed in commit aa98446c5fd04b7667d0687e52c7af9279033cc8. Made application_preferred_regions and excluded_regions private in CosmosClientOptions, and updated CosmosClient to use getter-based reads (application_preferred_regions().to_vec() / excluded_regions().to_vec()). Commit: https://github.com/Azure/azure-sdk-for-rust/commit/aa98446c5fd04b7667d0687e52c7af9279033cc8

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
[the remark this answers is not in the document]
FabianMeiswinkel: Addressed in commit aa98446c5fd04b7667d0687e52c7af9279033cc8. Made session_retry_options private in CosmosClientOptions; this field is now accessor-only via session_retry_options() with no crate-internal direct field reads. Commit: https://github.com/Azure/azure-sdk-for-rust/commit/aa98446c5fd04b7667d0687e52c7af9279033cc8

sdk/cosmos/azure_data_cosmos/src/models/account_properties.rs
```diff
@@ -12,111 +12,252 @@ use serde_json::Value;
...
+    /// Gets the database account endpoint for this region.
+    pub fn database_account_endpoint(&self) -> &Url {
+        &self.database_account_endpoint
+    }
 }
 
 /// Describes replica set sizing characteristics for user/system replication policies.
-#[derive(SafeDebug, Clone, Serialize, Deserialize)]
+#[non_exhaustive]
+#[derive(SafeDebug, Clone, Default, Serialize, Deserialize)]
```
Pilchie: These are `Default` because they are wire formats where not everything might be specified?

sdk/cosmos/azure_data_cosmos/src/models/mod.rs
```diff
@@ -50,42 +50,96 @@ pub struct SystemProperties {
...
 pub struct DatabaseProperties {
     /// The ID of the database.
-    pub id: String,
+    id: String,
 
     /// A [`SystemProperties`] object containing common system properties for the database.
     #[serde(flatten)]
-    pub system_properties: SystemProperties,
+    #[serde(default)]
+    system_properties: SystemProperties,
```
Pilchie: There's no way to set this to anything other than `::default()`?

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
FabianMeiswinkel: This can be addressed by appropriate deconstruction when needed frequently, correct? Added this for the sake of having an example here. Not perfect - but IMO a reasonable trade-off.

FabianMeiswinkel: Actually, I take that back - it woudl be tedious to identify when this is needed frequently.

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
```diff
@@ -0,0 +1,278 @@
...
+| Usage scope | Struct visibility |
+|---|---|
+| Only within the declaring module | No visibility modifier (private) |
+| Within the parent module | `pub(super)` |
+| Within the crate | `pub(crate)` |
+| Outside the crate | `pub` |
+
+Additional rules:
+
+- If a struct is marked `pub` but lives inside a non-public module (e.g., `pub struct Foo` inside `pub(crate) mod internal`), change the struct to use the **effective** visibility: `pub(crate) struct Foo`. This makes the actual visibility obvious at the struct declaration site without requiring the reader to trace module ancestry.
```
Pilchie, line 111: I really like this - I've been struggling to understand what the visibility of our types was, and this is useful.

Pilchie (approved):

sdk/cosmos/azure_data_cosmos/src/models/mod.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/models/account_properties.rs
FabianMeiswinkel: Should not have implemented Default - reverted.

sdk/cosmos/azure_data_cosmos/src/clients/database_client.rs
FabianMeiswinkel: I changed the Skill to not force private fields

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
FabianMeiswinkel: Fixed

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
FabianMeiswinkel: rename into content_response_on_write_enabled also add with_content_response_on_write_enabled

sdk/cosmos/azure_data_cosmos/examples/cosmos/create.rs
```diff
@@ -85,10 +85,8 @@ impl CreateCommand {
                 let pk = PartitionKey::from(&partition_key);
                 let item: serde_json::Value = serde_json::from_str(&json)?;
 
-                let options = ItemOptions {
-                    enable_content_response_on_write: show_updated,
-                    ..Default::default()
-                };
+                let mut options = ItemOptions::default();
+                options.enable_content_response_on_write = show_updated;
```
FabianMeiswinkel: use fluent with_content_response_on_write_enabled(show_updated) instead.

sdk/cosmos/azure_data_cosmos/examples/cosmos/replace.rs
```diff
@@ -74,10 +74,8 @@ impl ReplaceCommand {
                 let pk = PartitionKey::from(&partition_key);
                 let item: serde_json::Value = serde_json::from_str(&json)?;
 
-                let options = ItemOptions {
-                    enable_content_response_on_write: show_updated,
-                    ..Default::default()
-                };
+                let mut options = ItemOptions::default();
+                options.enable_content_response_on_write = show_updated;
```
FabianMeiswinkel: use fluent with_content_response_on_write_enabled(show_updated) instead.

sdk/cosmos/azure_data_cosmos/examples/cosmos/upsert.rs
```diff
@@ -33,10 +33,8 @@ impl UpsertCommand {
         let pk = PartitionKey::from(&self.partition_key);
         let item: serde_json::Value = serde_json::from_str(&self.json)?;
 
-        let options = ItemOptions {
-            enable_content_response_on_write: self.show_updated,
-            ..Default::default()
-        };
+        let mut options = ItemOptions::default();
+        options.enable_content_response_on_write = self.show_updated;
```
FabianMeiswinkel: use fluent with_content_response_on_write_enabled(show_updated) instead.

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
```diff
@@ -0,0 +1,309 @@
...
+If one or more fields are non-public, `#[non_exhaustive]` is optional and typically redundant for construction control.
+
+#### b) Field visibility on truly public structs — choose based on validation/invariant needs
+
+Fields on truly public structs may be `pub` **or** non-public. Choose visibility by checking (1) whether validation/invariant enforcement is needed and (2) whether crate-internal code in other modules needs direct non-getter access:
+
+| Scenario | Field visibility |
+|---|---|
+| Field has no validation/invariant constraints and direct access is acceptable for external consumers | **`pub`** is allowed. `with_*` setters and direct field modification may coexist. |
+| Field requires validation/invariant enforcement (e.g., value constraints, normalization, coupled-field invariants) | **Non-public** (private by default; `pub(crate)` only when justified by crate-internal usage). Route external mutation through `with_*`/constructor APIs. |
```
analogrelay, line 146: As a follow-up (not in this PR): We should add a note about introducing newtypes or other type system changes when validation/invariants could be enforced that way. Basically that the type system is the ideal place for this validation and validation in `with_` setters is a last resort.

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
```diff
@@ -0,0 +1,309 @@
...
+
+#### g) Optional builder guidance (when used)
+
+If a separate builder type is used, follow these conventions:
+
+1. Name it `<Type>Builder`.
+2. Keep builder fields private.
+3. Provide `with_*` setters for optional fields.
+4. Provide terminal `build(self, ...) -> <Type>` (or `azure_core::Result<Type>` when fallible).
+5. Keep required fields on `build(...)`, not as optional builder state.
```
analogrelay, line 230:
```suggestion
1. Name it `<Type>Builder`.
2. Keep builder fields private.
3. Provide `with_*` setters for optional fields.
4. Provide terminal `build(self, ...) -> <Type>` (or `azure_core::Result<Type>` when fallible).
5. Keep required fields on `build(...)`, not as optional builder state.
6. Add `<Type>::builder(... required args ...) -> <Type>Builder` to initialize the builder type.
```

No need to re-run the skill with this change, and I won't block the PR on it, but adding this would be useful.

sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_containers.rs
```diff
@@ -154,18 +140,16 @@ pub async fn container_crud_hierarchical_pk() -> Result<(), Box<dyn Error>> {
...
+            let properties = ContainerProperties::new(
+                "TheContainer",
+                ("/parent", "/child", "/grandchild").into(),
+            )
+            .with_indexing_policy(
+                IndexingPolicy::default()
+                    .with_included_path("/*")
+                    .with_excluded_path(r#"/"_etag"/?"#)
+                    .with_indexing_mode(IndexingMode::Consistent),
+            );
```
analogrelay, line 152: This is honestly quite nice looking.

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_containers.rs
FabianMeiswinkel: Agreed

sdk/cosmos/.github/skills/cosmos-design-struct/SKILL.md
FabianMeiswinkel: Fixed

PR #3759, opened 2026-02-17
fix: normalize whitespace azd errors

### LINKED ISSUE #3704: Normalize whitespace in AzureDeveloperCliCredential errors
Extreme nit - in fact, not even suggesting you fix it now, but maybe think about it for later: why is there a blank space after the message? I mean, I assume this is a result of removing `trim()` which would return a new `String` - justified - but isn't that our message. Why is there a trailing space?

_Originally posted by @heaths in https://github.com/Azure/azure-sdk-for-rust/pull/3674#discussion_r2789970332_

kx0101:
- normalize whitespace in `AzureDeveloperCliCredential` error messages by trimming the message and adding a leading newline
- update tests to reflect the new error format
- `cargo test -p azure_identity azure_developer_cli_credential` - all 9 tests pass

Fixes #3704

### DIFF
--- a/sdk/identity/azure_identity/src/lib.rs
+++ b/sdk/identity/azure_identity/src/lib.rs
@@ -178,7 +178,10 @@ fn authentication_error(credential_name: &str, err: Error) -> Error {
         stringify!(WorkloadIdentityCredential) => "#workload",
         _ => "",
     };
+    const WHITESPACE: &[char; 3] = &['\t', '\x0c', ' '];
+
     let err_str = err.to_string();
+    let err_str = err_str.trim_matches(WHITESPACE);
     let separator = if err_str.starts_with('\n') { "" } else { " " };
     let mut message = format!("{credential_name} authentication failed.{separator}{err_str}");
     if !link_fragment.is_empty() {


danieljurek:
@kx0101 -- If you're contributing as a Microsoft employee, follow instructions in https://aka.ms/opensource/link so future PRs are properly tagged.

sdk/identity/azure_identity/src/azure_developer_cli_credential.rs
```diff
@@ -54,9 +54,9 @@ impl OutputProcessor for AzdTokenResponse {
     fn get_error_message(stderr: &str) -> Option<String> {
         // Try to parse azd's JSON error output and extract the message
         if let Ok(error_output) = from_json::<_, AzdErrorOutput>(stderr) {
-            let message = &error_output.data.message;
+            let message = error_output.data.message.trim();
```
heaths, line 57: @chlowell did you just get rid of this call for some reason?

sdk/identity/azure_identity/src/azure_developer_cli_credential.rs
```diff
@@ -231,7 +231,7 @@ mod tests {
         let err = run_test(1, "", stderr, None).await.expect_err("error");
         assert!(matches!(err.kind(), ErrorKind::Credential));
         assert_eq!(
-            "AzureDeveloperCliCredential authentication failed. \nERROR: not logged in, run `azd auth login` to login\n\nTo troubleshoot, visit https://aka.ms/azsdk/rust/identity/troubleshoot#azd",
```
heaths, line 234: If we're going to be trimming whitespace and allocating new `String`s anyway, we may as well trim whitespace from the returned error message e.g.:

```suggestion
            "AzureDeveloperCliCredential authentication failed.\nERROR: not logged in, run `azd auth login` to login\nTo troubleshoot, visit https://aka.ms/azsdk/rust/identity/troubleshoot#azd",
```

Also, the double `\n\n` was to add separate separation for readability. Not sure we want to remove that. @chlowell?

heaths (changes requested):

sdk/identity/azure_identity/src/azure_developer_cli_credential.rs
kx0101: @heaths Thanks for the feedback! I've addressed your concerns:

  1. regarding the trailing space - good catch! I've updated `authentication_error` to not add a space when the error message starts with a newline

  Now the output is `"...failed.\nERROR..."` instead of `"...failed. \nERROR..."`

  2. the` \n\n` separator - I think the double newline existed because the original message from azd had a trailing `\n` (e.g., `"\nERROR: ...\n"`). And since we now `.trim()` the message and add back just a leading `\n`, I guess we no longer get the double newline

chlowell: We add `\n` when linking to the troubleshooting guide i.e., we add it to all these errors. Every azd error message I've seen has at least one leading and trailing newline, so the double newline was an artifact of not trimming azd's message. Personally I'm happy with a newline and not a full blank line before the TSG link and that's consistent with error messages from other credential types, however I'm open to changing this formatting if you strongly believe a blank line improves legibility.

sdk/identity/azure_identity/src/azure_developer_cli_credential.rs
chlowell: I believe I liked the simplicity of accepting azd's string and not having an opinion about whether it should have leading or trailing whitespace. In practice this would produce the format you like because every azd error I've seen is surrounded by newlines. But I suppose it's better--more durable over time--for us to control the whitespace here.

heaths: We could just remove space characters. There are other ways to trim. Not a big deal, though, if we're primarily just thinking about terminal rendering.

sdk/identity/azure_identity/src/lib.rs
```diff
@@ -178,7 +178,9 @@ fn authentication_error(credential_name: &str, err: Error) -> Error {
         stringify!(WorkloadIdentityCredential) => "#workload",
         _ => "",
     };
-    let mut message = format!("{credential_name} authentication failed. {err}");
+    let err_str = err.to_string();
+    let separator = if err_str.starts_with('\n') { "" } else { " " };
```
heaths, line 185: Try to avoid `to_string()`. It relies on formatting as a blanket impl of `ToString` for every `Display` impl. `&str::to_owned()` is usually best.

That said, why not just trim spaces?

```rust
const WHITESPACE: &[char; 3] = &['\t', '\x0c', ' '];
assert_eq!(" Just trim spaces.\n".trim_matches(WHITESPACE), "Just trim spaces.\n");
```

Cleaner, IMO.

kx0101: err though is an `azure_core::Error` not a `&str`, which means `.to_owned()` isn't applicable here

` .to_string()` is the only way to get a `String` from it and then use it on `trim_matches(WHITESPACE)`

how about:
`azure_identity/src/lib.rs`
```rust
const WHITESPACE: &[char; 3] = &['\t', '\x0c', ' '];

let err_str = err.to_string();
let err_str = err_str.trim_matches(WHITESPACE);
let separator = if err_str.starts_with('\n') { "" } else { " " };
let mut message = format!("{credential_name} authentication failed.{separator}{err_str}");
```

I also kept the separator logic so errors still read as `"failed. something went wrong"` rather `"failed.something went wrong"`

heaths: What I was hoping to avoid was that conditional. Do we need it anymore if we're not stripping newlines?

kx0101: if i'm not mistaken, if we drop the conditional, we'd need to always use `\n` as separator, which changes plain-text errors from a single line (`"failed. something went wrong"`) to two lines (`"failed.\nsomething went wrong"`). Is that what you had in mind?

heaths: I believe this is what @chlowell wanted. The newlines are to make it more readable for humans. These's aren't errors intended to be retried generally. You can either auth or not, and humans (or an agent, whatever) will need to parse that message. The more friendly we can make it for humans, the better. Newlines are one such way.

I would hope that no one is generally using an error message itself to make programmatic decisions. Rarely are error messages deemed stable in any language/framework.

chlowell: There's a good guideline if we don't already have it: error messages are not contractual and may change at any time.

For these errors specifically, I imagine messages like
```
AzureDeveloperCliCredential authentication failed.
{verbatim azd error message}
To troubleshoot, ...
```
I don't have a strong opinion on the number of newlines in the string and think it's reasonable to split it up this way.

Just to be clear, these are the requirements I have in mind for error messages (most aren't relevant to this PR):
1. identify the error as an authN error programmatically (`ErrorKind::Credential`) and textually ("authentication failed")
1. name the credential type
1. when we have an external message e.g. from another process or a server, include its text verbatim
    - remove surrounding whitespace as needed because it's meaningless and interferes with presentation (so far as I know, only azd messages have leading and trailing newlines)
1. link to the troubleshooting guide

kx0101: funny how much you can learn from such a small PR 😄 

really appreciate the detailed feedback

PR #3592, opened 2026-01-22
Driver-03: DiagnosticsContext and AsyncCache

FabianMeiswinkel:
# Summary
This PR introduces the initial skeleton structure for azure_data_cosmos_driver, a new core implementation layer for Azure Cosmos DB that provides transport, routing, and protocol handling for cross-language SDK reuse.

# Details
see sdk/cosmos/ARCHITECTURE.md 

# Open issues
- Intentionally not using azure_data_cosmos_driver yet from azure_data_cosmos
- Bare minimum http pipeline in driver - next step will be to move logic from azure_data_cosmos and refactor it to fit in the driver
- Bare minimum test coverage - eventually azure_data_cosmos will use azure_data_cosmos_driver and that will provide us more code coverage

# Open questions
- for now I am leaning towards non-exhaustive struct/enum to avoid extension to become breaking even in the driver. Technically that would not strictly be necessary because we can easily deal with compile-time breakign changes between sdk and driver. So, let me know when you feel strongly that the non-exhaustive should not be the default.

PR #3769, opened 2026-02-19
[DRAFT]Driver-04: Add AsyncCache

FabianMeiswinkel:


Cargo.toml
```diff
@@ -151,14 +152,21 @@ url = "2.5"
 uuid = { version = "1.20", features = ["v4"] }
 wasm-bindgen-futures = "0.4"
 wasm-bindgen-test = "0.3"
+windows-sys = { version = "0.61", features = [
+  "Win32_System_Threading",
+  "Win32_System_SystemInformation",
+] }
```
heaths: 1. These aren't being used.
2. Even if they were, I don't think we should be pulling in specific features that only Cosmos needs. I would pass only `default-features = false` and then in your workspace import add the features you need. This is a big dependency and I want to be careful about the impact to devs.
3. Even if they were, you're workspace import should be predicated only on windows like we have in some other `Cargo.toml`s e.g. `target.'cfg(windows)'.dependencies` or something like that.

FabianMeiswinkel: @heaths  - no worries - I am just breakign up a large change into separate PRs - I will reach out for the last one (where the windows-sys dependency is added). And we can check how to best approach it - bottom-line all we need is basic functionality to measure CPU usage and memory consumption. But you can ignore this for now.

FabianMeiswinkel: @heaths - I have reverted the repo-wide cargo.toml changes in this PR (for Driver-01 to Driver-04 they are not needed yet). For the last iteration Driver-05 -https://github.com/Azure/azure-sdk-for-rust/pull/3592 where the System-monitoring is added, I added windows-sys but slightly different to minimize impact

# repo cargo.toml
```TOML
[workspace.dependencies]
windows-sys = { version = "0.61", default-features = false }
```

# azure_data_cosmos_driver cargo.toml
```TOML
[target.'cfg(windows)'.dependencies]
windows-sys = { workspace = true, features = [
  "Win32_System_Threading",
  "Win32_System_SystemInformation",
] }
```

Is that a reasonable approach?

PR #3771, opened 2026-02-19
Driver-02: Bare minimum transport and e2e tests

FabianMeiswinkel:
This PR just adds a bare minimum transport and some emulator tests. I will completely rewrite the transport part next week - adding this here only to be able ot run something e2e before adding the AsyncCache and DiagnosticsContext in subsequent PRs. So, please take this review with a grain of salt. Really just a tactical addition to get to the reviews of AsyncCache and DiagnosticsContext


sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/tracked_transport.rs
```diff
@@ -0,0 +1,99 @@
...
+        match self.kind() {
+            ErrorKind::Io => {
+                let msg = self.to_string().to_lowercase();
+                if msg.contains("dns")
+                    || msg.contains("resolve")
+                    || msg.contains("connection refused")
+                    || msg.contains("no route to host")
+                    || msg.contains("network unreachable")
+                    || msg.contains("connection reset")
+                        && (msg.contains("before") || msg.contains("establish"))
```
analogrelay: We should file an issue or something flagging this as brittle and to-be-replaced before GA.

Pilchie, line 48: How confident are we in this heuristic? Would it be better to have an allow-list of full mesages?

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -36,6 +39,15 @@ pub struct CosmosDriver {
 }
 
 impl CosmosDriver {
+    fn should_retry_transport_failure(
+        attempt: usize,
+        max_transport_retries: usize,
+        is_idempotent: bool,
+        request_sent: RequestSentStatus,
+    ) -> bool {
+        attempt < max_transport_retries && is_idempotent && request_sent.definitely_not_sent()
```
Pilchie: Why does it have to idempotent if we know it wasn't sent?

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -197,49 +212,104 @@ impl CosmosDriver {
...
+        // Step 9: Select and create appropriate pipeline
+        let transport = self.runtime.transport();
+        let is_dataplane = uses_dataplane_pipeline(resource_type, operation_type);
+        let pipeline = if is_dataplane {
+            transport.create_dataplane_pipeline(&endpoint, auth)
+        } else {
+            transport.create_metadata_pipeline(&endpoint, auth)
+        };
+
+        // Step 12+: Execute with a slim transport retry wrapper.
```
Pilchie: We went from Step 8 to Step 12.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -197,49 +212,104 @@ impl CosmosDriver {
...
-        let response_headers = CosmosResponseHeaders::from_headers(&Headers::new());
-        let _dummy_result = crate::models::CosmosResponse::new(
-            Vec::new(),
-            response_headers,
-            CosmosStatus::from_parts(azure_core::http::StatusCode::Ok, None),
-        );
+            if let Some(pk) = operation.partition_key() {
+                let _partition_key_definition = operation
+                    .container()
+                    .map(|container| container.partition_key_definition());
```
Pilchie, line 261: Do we expect to need this later?

sdk/cosmos/azure_data_cosmos_driver/tests/framework/test_client.rs
```diff
@@ -0,0 +1,372 @@
...
+        };
+
+        let conn_str: ConnectionString = connection_string.parse()?;
+        let endpoint = conn_str.account_endpoint().parse()?;
+        let key = conn_str.account_key().secret().to_string();
+        let account = AccountReference::with_master_key(endpoint, key);
+
+        // Build runtime with emulator certificate handling
+        let connection_pool = ConnectionPoolOptions::builder()
+            .with_emulator_server_cert_validation(EmulatorServerCertValidation::DANGEROUS_DISABLED)
```
Pilchie: Shouldn't we do this only if the key is the emulator key?

sdk/cosmos/azure_data_cosmos_driver/tests/framework/test_client.rs
```diff
@@ -0,0 +1,372 @@
...
+        partition_key_path: &str,
+    ) -> Result<ContainerReference, Box<dyn Error>> {
+        let driver = self
+            .client
+            .runtime
+            .get_or_create_driver(self.client.account.clone(), None)
+            .await?;
+
+        let body = format!(
+            r#"{{"id": "{}", "partitionKey": {{"paths": ["{}"], "kind": "Hash"}}}}"#,
```
Pilchie: Should we default to Hashv2 here?

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
```diff
@@ -0,0 +1,448 @@
...
+        } else {
+            self.dataplane_transport.clone()
+        }
+    }
+
+    /// Creates an authenticated pipeline with headers and authorization policies.
+    fn create_authenticated_pipeline(
+        &self,
+        transport: Transport,
+        auth: &Credential,
```
Pilchie: nit: rename parameter to `credential`

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
FabianMeiswinkel: For sure - added this as dummy to avoid dead-code because this API WILL be used.

sdk/cosmos/azure_data_cosmos_driver/tests/framework/test_client.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/tests/framework/test_client.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/tracked_transport.rs
FabianMeiswinkel: Will change next week - Heath, Ashley, Tomas and I had conversations on fidning a reasonable way to identify this for reqwest (default) and fetch (the 1st party http client). Full Message or contains - both ar ehacks that can't be maintained. So, leaving this in for now - but will definitely change it in the PR next week. Added a comment to capture that as well.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
FabianMeiswinkel: Fixed

PR #3772, opened 2026-02-19
Driver-01: Minimalistic Driver skeleton

FabianMeiswinkel:
# Summary
This PR introduces the initial skeleton structure for azure_data_cosmos_driver, a new core implementation layer for Azure Cosmos DB that provides transport, routing, and protocol handling for cross-language SDK reuse.

# Details
see sdk/cosmos/ARCHITECTURE.md 

# Open issues
- Intentionally not using azure_data_cosmos_driver yet from azure_data_cosmos
- No actual work done in execute_operation - not even bare minimum transport
- No Diagnostics yet
- Bare minimum test coverage - eventually azure_data_cosmos will use azure_data_cosmos_driver and that will provide us more code coverage

# Following PRs add additional features:
https://github.com/Azure/azure-sdk-for-rust/pull/3771
https://github.com/Azure/azure-sdk-for-rust/pull/3770
https://github.com/Azure/azure-sdk-for-rust/pull/3769
https://github.com/Azure/azure-sdk-for-rust/pull/3592

### DIFF
--- a/sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
+++ b/sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
@@ -241,9 +241,6 @@ impl CosmosDriver {
             "execute_operation is not implemented in this transport-free driver build",
         ))
     }
-}
-
-impl CosmosDriver {
     /// Resolves a container by database and container name.
     ///
     /// Reads the database and container from the service to obtain their
@@ -289,10 +286,7 @@ impl CosmosDriver {
         db_name: &str,
         container_name: &str,
     ) -> azure_core::Result<ContainerReference> {
-        let account = self.account().clone();
-        let db_name = db_name.to_owned();
-        let container_name = container_name.to_owned();
-        let db_ref = DatabaseReference::from_name(account.clone(), db_name.clone());
+        let db_ref = DatabaseReference::from_name(self.account().clone(), db_name.to_owned());
         let options = OperationOptions::new();
 
         let db_result = self
@@ -312,7 +306,7 @@ impl CosmosDriver {
 
         let container_result = self
             .execute_operation(
-                CosmosOperation::read_container_by_name(db_ref, container_name.clone()),
+                CosmosOperation::read_container_by_name(db_ref.clone(), container_name.to_owned()),
                 options,
             )
             .await?;
@@ -330,10 +324,10 @@ impl CosmosDriver {
             })?;
 
         Ok(ContainerReference::new(
-            account,
-            db_name,
+            db_ref.into_account(),
+            db_props.id.into_owned(),
             db_rid,
-            container_name,
+            container_props.id.clone().into_owned(),
             container_rid,
             &container_props,
         ))
--- a/sdk/cosmos/azure_data_cosmos_driver/src/models/resource_reference.rs
+++ b/sdk/cosmos/azure_data_cosmos_driver/src/models/resource_reference.rs
@@ -46,6 +46,11 @@ impl DatabaseReference {
         &self.account
     }
 
+    /// Consumes this database reference and returns the parent account.
+    pub fn into_account(self) -> AccountReference {
+        self.account
+    }
+
     /// Returns the database name, if this is a name-based reference.
     pub fn name(&self) -> Option<&str> {
         self.id.name()


sdk/cosmos/azure_data_cosmos_driver/ARCHITECTURE.md
```diff
@@ -0,0 +1,919 @@
+# Azure Cosmos DB Driver - API Overview
+
+This document provides a comprehensive overview of the `azure_data_cosmos_driver` public API for team discussion and onboarding.
+
+## Three-Layer Architecture
+
+The Azure Cosmos DB Rust ecosystem consists of three distinct layers:
+
+```text
```
analogrelay, line 9: Non-Blocking: This diagram is a bit confusing since the Rust SDK doesn't use `azure_data_cosmos_native`

sdk/cosmos/azure_data_cosmos_driver/ARCHITECTURE.md
```diff
@@ -0,0 +1,919 @@
...
+pub async fn read_container(
+    &self,
+    container_ref: ContainerReference,
+) -> Result<ContainerProperties> { /* ... */ }
+```
+
+**Rationale**:
+- Metadata schemas are defined by the Cosmos DB service, not applications
+- Typed APIs prevent misuse (e.g., setting invalid partition key definitions)
+- Applications should never parse raw JSON from metadata operations
```
analogrelay: Are we OK with new properties added to metadata responses being completely inaccessible to customers on previous versions of the Driver? I actually do wonder if we should provide some mechanism to read raw metadata from the driver.

Also, does this apply to query APIs like `query_databases` and `query_containers` where the user can project the response properties?

This is just for later discussion though, non-blocking for this PR.

sdk/cosmos/azure_data_cosmos_driver/ARCHITECTURE.md
```diff
@@ -0,0 +1,919 @@
...
+- Cross-language SDKs all use identical metadata models (service contract)
+
+### Current Implementation
+
+The driver currently:
+- ✅ Uses typed metadata models internally for cache resolution (`DatabaseProperties`, `ContainerProperties`)
+- ✅ Returns raw bytes for all data plane operations
+- ⚠️ Does not yet expose high-level typed metadata APIs publicly (operations return `CosmosResult` with raw bytes)
```
analogrelay: I think this may actually be the right approach in the long-run, allowing the SDK to ser/de metadata types itself, but still giving the Driver enough info to understand it. It means re-parsing the SDK-provided JSON into data types, but for metadata operations my sense is that cost is negligible.

sdk/cosmos/azure_data_cosmos_driver/ARCHITECTURE.md
```diff
@@ -0,0 +1,919 @@
...
+```
+
+---
+
+## Configuration Hierarchy
+
+Configuration cascades from most general to most specific:
+
+```text
+Environment Variables (COSMOS_*)
```
analogrelay:
```suggestion
Environment Variables (AZURE_COSMOS_*)
```

The general rule of thumb I heard from the core team was that any env vars be prefixed `AZURE_[SERVICE}_...`

Can be done in a follow-up to reduce churn.

sdk/cosmos/azure_data_cosmos_driver/ARCHITECTURE.md
```diff
@@ -0,0 +1,919 @@
...
+                                            │       ├── TransportComplete
+                                            │       └── TransportFailed
+                                            ├── timestamp: Instant
+                                            ├── duration_ms: Option<u64>
+                                            └── details: Option<String>
+```
+
+### Pipeline Events & Reqwest Limitations
+
+> ⚠️ **Team Discussion Point**: The driver uses `reqwest` as the HTTP transport. Unlike Reactor Netty (used in the Java SDK), reqwest does **not** expose fine-grained connection lifecycle callbacks.
```
analogrelay: As long as the diagnostics format _supports_ tracking this data, we can add it later though. I believe `reqwest` gives us access to the underlying [`tower`] primitives that we could use to inject our own tracing here, if we control the pipeline.

sdk/cosmos/azure_data_cosmos_driver/src/models/mod.rs
```diff
@@ -0,0 +1,574 @@
...
+            system_properties: SystemProperties::default(),
+        }
+    }
+}
+
+/// Immutable container properties that never change after creation.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct ImmutableContainerProperties {
+    partition_key: PartitionKeyDefinition,
+}
```
analogrelay: Maybe embed this in `ContainerProperties` (with `#[serde(flatten)]`)? That would be kinda clunky for external customers, but it would be OK for SDKs using the driver and the SDK can always adapt it to something cleaner.

sdk/cosmos/azure_data_cosmos_driver/src/models/mod.rs
```diff
@@ -0,0 +1,574 @@
...
+
+impl DatabaseProperties {}
+
+/// Properties of a Cosmos DB container.
+///
+/// Returned by container read/query operations and used when creating/updating containers.
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[non_exhaustive]
+pub(crate) struct ContainerProperties {
```
analogrelay, line 74: Maybe our metadata types **in the driver** should have an `extra: HashMap<String, serde_json::Value>` field to capture unexpected extra properties from the backend? (Using the pattern here: https://serde.rs/attr-flatten.html)

sdk/cosmos/azure_data_cosmos_driver/src/models/mod.rs
```diff
@@ -0,0 +1,574 @@
...
+    }
+}
+
+impl std::fmt::Display for SessionToken {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(&self.0)
+    }
+}
+
+// ── TriggerInvocation (moved from triggers.rs) ──────────────────────────────
```
analogrelay: We should have an agent make a pass to clean up some of these comments. It bothers me that agents so frequently put "progress" notes in comments. But as with all my feedback so far, this is non-blocking.

sdk/cosmos/azure_data_cosmos_driver/src/models/mod.rs
```diff
@@ -0,0 +1,574 @@
...
+}
+
+// ── ThroughputControlGroupName (moved from throughput_control.rs) ───────────
+
+/// Unique name identifying a throughput control group.
+///
+/// This name is serialized into request headers when referencing a control group.
+/// The group configuration is defined separately via [`ThroughputControlGroupOptions`](crate::options::ThroughputControlGroupOptions).
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct ThroughputControlGroupName(pub Cow<'static, str>);
```
analogrelay, line 461: We could probably have Copilot whip up a macro to simplify defining these newtypes for us and remove code repetition.

sdk/cosmos/azure_data_cosmos_driver/src/models/etag.rs
```diff
@@ -0,0 +1,156 @@
...
+///
+/// // Update only if the resource hasn't changed (optimistic concurrency)
+/// let condition = ETagCondition::if_match("\"abc123\"");
+///
+/// // Create only if the resource doesn't exist
+/// let condition = ETagCondition::if_none_match("*");
+/// ```
+#[non_exhaustive]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ETagCondition {
```
analogrelay: Non-Blocking: I'd prefer a name like `RequestCondition`, since it's really defining a condition for accepting the request.

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_status.rs
```diff
@@ -0,0 +1,1426 @@
...
+            status_code,
+            sub_status: Some(SubStatusCode::new(sub_status_code)),
+        }
+    }
+
+    /// Sets the sub-status code on this `CosmosStatus`, returning the modified value.
+    pub fn with_sub_status(mut self, sub_status_code: u32) -> Self {
+        self.sub_status = Some(SubStatusCode::new(sub_status_code));
+        self
+    }
```
analogrelay: I don't think we need both of these. `CosmosStatus::new(status_code).with_sub_status(sub_status_code)` isn't much more verbose and works the same way.

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_result.rs
```diff
@@ -0,0 +1,167 @@
...
+/// println!("Status: {}", status);
+/// println!("RU Charge: {}", result.headers().request_charge().unwrap_or(0.0));
+/// if status.is_success() {
+///     let body = result.into_body();
+///     // Deserialize body...
+/// }
+/// ```
+#[derive(Debug)]
+#[non_exhaustive]
+pub struct CosmosResult {
```
analogrelay: Non-Blocking: This should be `CosmosResponse`. To me, "Result" implies that it's related to `std::result::Result<T, E>`, which this isn't.

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_headers.rs
```diff
@@ -0,0 +1,341 @@
...
+    etag: Option<ETag>,
+
+    /// Continuation token for pagination (`x-ms-continuation`).
+    continuation: Option<String>,
+
+    /// Item count in response (`x-ms-item-count`).
+    item_count: Option<u32>,
+
+    /// Cosmos substatus code (`x-ms-substatus`).
+    substatus: Option<SubStatusCode>,
```
analogrelay: These could probably just be `pub` fields.

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_headers.rs
```diff
@@ -0,0 +1,341 @@
...
+///
+/// Only whitelisted request headers are supported.
+#[derive(Clone, Debug, Default)]
+#[non_exhaustive]
+pub struct CosmosRequestHeaders {
+    /// Activity ID for request correlation (`x-ms-activity-id`).
+    activity_id: Option<ActivityId>,
+
+    /// Session token for session consistency (`x-ms-session-token`).
+    session_token: Option<SessionToken>,
```
analogrelay: These could be `pub` fields

sdk/cosmos/azure_data_cosmos_driver/src/models/user_agent.rs
```diff
@@ -0,0 +1,220 @@
...
+/// User agent string for HTTP requests.
+///
+/// The user agent is automatically computed with a static prefix containing:
+/// - Azure SDK identifier (`azsdk-rust-`)
+/// - SDK name and version
+/// - OS name and architecture
+/// - Rust version (compile time)
+///
+/// An optional suffix can be appended (typically from [`UserAgentSuffix`],
+/// [`WorkloadId`], or [`CorrelationId`]).
```
analogrelay, line 31: I assume the wrapper SDK would also inject a suffix here so we can track that as well.

sdk/cosmos/azure_data_cosmos_driver/src/models/user_agent.rs
```diff
@@ -0,0 +1,220 @@
...
+        // after truncation and ASCII stripping, so that `suffix()` matches
+        // what was actually appended.
+        let effective_suffix = match &suffix {
+            Some(_) => full_user_agent
+                .strip_prefix(base_stripped.as_str())
+                .and_then(|rest| rest.strip_prefix(' '))
+                .filter(|s| !s.is_empty())
+                .map(String::from),
+            _ => None,
+        };
```
analogrelay: Non-Blocking: This allocates more than it needs to, I think. It could all be done with length checks on the strings rather than actually formatting. If we strip non ascii characters, it's safe to use `str::len` to check string length.

sdk/cosmos/azure_data_cosmos_driver/src/options/env_parsing.rs
analogrelay, line 1: Would love some tests in this file, but it can be a follow up.

sdk/cosmos/azure_data_cosmos_driver/src/options/policies.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub const ENABLED: Self = Self(true);
+    /// Content response is disabled (no response body).
+    pub const DISABLED: Self = Self(false);
+}
+
+impl From<bool> for ContentResponseOnWrite {
+    fn from(value: bool) -> Self {
+        Self(value)
+    }
+}
```
analogrelay, line 36: Any particular reason this isn't an `enum`?

sdk/cosmos/azure_data_cosmos_driver/src/options/policies.rs
```diff
@@ -0,0 +1,169 @@
...
+
+/// Controls whether quota information is included in responses.
+///
+/// When enabled, container quota stats are returned in the response headers.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash)]
+pub struct QuotaInfoEnabled(pub bool);
```
analogrelay: Enum?

sdk/cosmos/azure_data_cosmos_driver/src/options/policies.rs
```diff
@@ -0,0 +1,169 @@
...
+/// use azure_data_cosmos_driver::options::EmulatorServerCertValidation;
+///
+/// // Safe default: validation is enabled.
+/// let validation = EmulatorServerCertValidation::ENABLED;
+///
+/// // Dangerous: disables certificate validation for emulator use.
+/// let validation = EmulatorServerCertValidation::DANGEROUS_DISABLED;
+/// ```
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash)]
+pub struct EmulatorServerCertValidation(bool);
```
analogrelay: Enum?

sdk/cosmos/azure_data_cosmos_driver/src/options/policies.rs
```diff
@@ -0,0 +1,169 @@
...
+    fn from_iter<I: IntoIterator<Item = T>>(iter: I) -> Self {
+        Self(iter.into_iter().map(Into::into).collect())
+    }
+}
+
+/// Controls whether JavaScript stored procedure logging is enabled.
+///
+/// When enabled, script logs from stored procedures are included in the response.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash)]
+pub struct ScriptLoggingEnabled(pub bool);
```
analogrelay: Enum?

sdk/cosmos/azure_data_cosmos_driver/src/models/account_reference.rs
```diff
@@ -0,0 +1,381 @@
...
+    }
+}
+
+impl Eq for AccountEndpoint {}
+
+impl Hash for AccountEndpoint {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.0.hash(state);
+    }
+}
```
analogrelay, line 60: These are just the same implementations that `derive` generates, so we could just derive them.

sdk/cosmos/azure_data_cosmos_driver/src/models/account_reference.rs
```diff
@@ -0,0 +1,381 @@
...
+
+impl std::fmt::Debug for AuthOptions {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::MasterKey(_) => f.debug_tuple("MasterKey").field(&"***").finish(),
+            Self::TokenCredential(_) => f.debug_tuple("TokenCredential").field(&"...").finish(),
+        }
+    }
+}
```
analogrelay, line 101: We can use `derive(SafeDebug)` from `azure_core` for this, and mark the secret values `#[safe(false)]`

sdk/cosmos/azure_data_cosmos_driver/src/models/account_reference.rs
```diff
@@ -0,0 +1,381 @@
...
+        account.endpoint.clone()
+    }
+}
+
+/// Authentication options for connecting to a Cosmos DB account.
+///
+/// Either key-based authentication using a master key, or token-based
+/// authentication using an Azure credential (e.g., managed identity, service principal).
+#[derive(Clone)]
+pub enum AuthOptions {
```
analogrelay: I'd prefer the name `Credential`, since AuthOptions to me implies that there will be additional options here, but as an enum it's one or the other.

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_operation.rs
```diff
@@ -0,0 +1,680 @@
...
+            resource_type,
+            resource_reference,
+            partition_key: None,
+            request_headers: CosmosRequestHeaders::new(),
+            body: None,
+        }
+    }
+
+    /// Creates a new operation with the specified type, resource reference, and partition key.
+    fn new_with_partition_key(
```
analogrelay: This should just be `with_partition_key`. The general pattern for secondary constructors is `with_x` (which does indeed collide with transforming setters, so we may need to break that pattern sometimes.

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_resource_reference.rs
```diff
@@ -0,0 +1,642 @@
...
+    /// Reference to the parent account (always required).
+    account: AccountReference,
+    /// Reference to the parent database (optional, depends on resource type).
+    database: Option<DatabaseReference>,
+    /// Reference to the parent container (optional, depends on resource type).
+    container: Option<ContainerReference>,
+    /// The resource name (mutually exclusive with RID for identification).
+    name: Option<Cow<'static, str>>,
+    /// The resource identifier (RID) (mutually exclusive with name for identification).
+    rid: Option<Cow<'static, str>>,
```
analogrelay: If these are mutually exclusive, they should be an `enum`.

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_resource_reference.rs
```diff
@@ -0,0 +1,642 @@
...
+            ResourceType::StoredProcedure
+            | ResourceType::Trigger
+            | ResourceType::UserDefinedFunction
+            | ResourceType::PartitionKeyRange => self.name.is_none() && self.rid.is_none(),
+            ResourceType::Offer => self.rid.is_none(),
+        }
+    }
+
+    /// Returns the parent's path for signing feed operations.
+    fn parent_signing_link(&self) -> String {
```
analogrelay: While the empty `String` is cheap to create (because it's just an empty `Vec`), I'd prefer this was `Option<String>`, for semantic accuracy.

sdk/cosmos/azure_data_cosmos_driver/src/models/request_charge.rs
```diff
@@ -0,0 +1,272 @@
...
+/// ```
+#[derive(Clone, Copy, Debug, Default, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct RequestCharge(f64);
+
+impl RequestCharge {
+    /// Creates a new `RequestCharge` from a raw `f64` value.
+    ///
+    /// NaN is normalized to `0.0` and negative zero is normalized to positive
+    /// zero so that `RequestCharge` can implement [`Eq`] and [`Hash`].
```
analogrelay, line 44: Does that mean this could use the `FiniteF64` type you created to store the value?

sdk/cosmos/azure_data_cosmos_driver/src/models/resource_id.rs
```diff
@@ -0,0 +1,583 @@
...
+    }
+}
+
+/// A resource identifier (RID) - internal Cosmos DB identifier.
+///
+/// RIDs are base64-encoded internal identifiers assigned by Cosmos DB.
+/// They encode the resource hierarchy (account → database → container → document).
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+pub(crate) struct ResourceRid(Cow<'static, str>);
```
analogrelay: nit: I'd just call this `ResourceId` since that's basically what `RID` stands for.

sdk/cosmos/azure_data_cosmos_driver/src/models/resource_id.rs
```diff
@@ -0,0 +1,583 @@
...
+// Internal ID Enums (pub(crate))
+// =============================================================================
+// These enums enforce either all-names or all-RIDs addressing at compile time.
+
+/// Generic resource identifier: either by name or by RID.
+///
+/// This is reused across resource reference types (including databases) to avoid
+/// duplicating identical `ByName`/`ByRid` enums per resource.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub(crate) enum ResourceIdentifierType {
```
analogrelay: nit: Just `ResourceIdentifier`?

sdk/cosmos/azure_data_cosmos_driver/src/models/resource_id.rs
```diff
@@ -0,0 +1,583 @@
...
+        match self {
+            Self::Empty => write!(f, "RID string is empty"),
+            Self::InvalidLength => write!(f, "RID has invalid byte length"),
+            Self::InvalidBase64 => write!(f, "RID contains invalid Base64"),
+        }
+    }
+}
+
+#[cfg(test)]
+impl std::error::Error for RidParseError {}
```
analogrelay: These should all just go _inside_ `mod tests`

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -0,0 +1,623 @@
...
+    /// For each property in `RuntimeOptions`, the first defined value is used.
+    pub fn effective_runtime_options(
+        &self,
+        operation_options: &OperationOptions,
+    ) -> RuntimeOptions {
+        // Start with operation-level options (highest priority)
+        let operation_runtime = operation_options.runtime();
+
+        // Get driver-level options
+        let driver_runtime = self.options.runtime_options().snapshot();
```
analogrelay, line 77: Where is the `.snapshot()` method coming from? It can just be `.clone()`, right?

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -0,0 +1,623 @@
...
+        );
+
+        Err(azure_core::Error::with_message(
+            azure_core::error::ErrorKind::Other,
+            "execute_operation is not implemented in this transport-free driver build",
+        ))
+    }
+}
+
+impl CosmosDriver {
```
analogrelay: Doesn't really need to be a separate `impl` block

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -0,0 +1,623 @@
...
+    /// ```
+    pub async fn resolve_container(
+        &self,
+        db_name: &str,
+        container_name: &str,
+    ) -> azure_core::Result<ContainerReference> {
+        let account = self.account().clone();
+        let db_name = db_name.to_owned();
+        let container_name = container_name.to_owned();
+        let db_ref = DatabaseReference::from_name(account.clone(), db_name.clone());
```
analogrelay: You probably don't need to `.clone()` these, if you could move them back out of the `db_ref` when we're constructing the `ContainerReference`.

sdk/cosmos/azure_data_cosmos_driver/ARCHITECTURE.md
FabianMeiswinkel: Sorry - this is left-over - i moved back-and-forth in my mind behind exposing typed APIs ofr metadata or not - latest is that I think we should let SDKs deal with serialization. Mostly because I think it would unnecessarily complicate the FFI story. And serializing metadata shouldn't be a big challenge

sdk/cosmos/azure_data_cosmos_driver/ARCHITECTURE.md
FabianMeiswinkel: agreed - and that is my final conclusion as well - just missed to update this doc - will do.

sdk/cosmos/azure_data_cosmos_driver/ARCHITECTURE.md
FabianMeiswinkel: My exploration o fthe reqwest extension points did not show a way how to do so - only Otel tracing etc. but without exposing the primitives. Anyway - agreed - not blocking

sdk/cosmos/azure_data_cosmos_driver/src/models/account_reference.rs
FabianMeiswinkel: @analogrelay - makes sense - would you be ok with CosmosCredential - my gut feeling is that Credential could be overly generic (even more so than one of its values TokenCredential)?

sdk/cosmos/azure_data_cosmos_driver/ARCHITECTURE.md
tvaron3: Feels weird for fabric customers which are not on azure but using this sdk

sdk/cosmos/azure_data_cosmos_driver/ARCHITECTURE.md
analogrelay: There's [`ClientBuilder::connector_layer`](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.connector_layer) which allows us to apply a [`tower::Layer`](https://docs.rs/tower/latest/tower/trait.Layer.html) to the Tower service that handles connections. It's quite low-level but I believe it would allow us to track connection establishment and teardown.

sdk/cosmos/azure_data_cosmos_driver/src/models/account_reference.rs
analogrelay: Honestly, I feel like we're being overly cautious with the `Cosmos` prefix. People are way more comfortable using module prefixing in Rust (`azure_data_cosmos_driver::Credential`, etc.) and don't tend to worry about overly generic types when some portion of the module path can always be used to disambiguate. But I'm fine with `CosmosCredential` if you prefer that approach.

sdk/cosmos/azure_data_cosmos_driver/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+description = "Core implementation layer for Azure Cosmos DB - provides transport, routing, and protocol handling for cross-language SDK reuse"
+readme = "README.md"
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+homepage = "https://github.com/azure/azure-sdk-for-rust"
+documentation = "https://docs.rs/azure_data_cosmos_driver"
+keywords = ["sdk", "azure", "cosmos", "driver", "internal"]
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Release History
+
+## 0.1.0 (Unreleased)
+
+### Features Added
+
+- Initial release of `azure_data_cosmos_driver` (core Cosmos DB protocol implementation for cross-language SDK reuse). ([#3592](https://github.com/Azure/azure-sdk-for-rust/pull/3592))
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/options/region.rs
```diff
@@ -0,0 +1,469 @@
...
+    ("southafricawest", "South Africa West"),
+    ("southafricanorth", "South Africa North"),
+    ("uaecentral", "UAE Central"),
+    ("uaenorth", "UAE North"),
+    ("centraluseuap", "Central US EUAP"),
+    ("eastus2euap", "East US 2 EUAP"),
+    ("northeurope2", "North Europe 2"),
+    ("easteurope", "easteurope"),
+    ("apacsoutheast2", "APAC Southeast 2"),
+    ("uksouth2", "UK South 2"),
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/options/runtime_options.rs
```diff
@@ -0,0 +1,351 @@
...
+    pub diagnostics_thresholds: Option<DiagnosticsThresholds>,
+    /// End-to-end latency policy for timeout management.
+    pub end_to_end_latency_policy: Option<EndToEndOperationLatencyPolicy>,
+    /// Regions to exclude from routing.
+    pub excluded_regions: Option<ExcludedRegions>,
+    /// Read consistency strategy for read operations.
+    pub read_consistency_strategy: Option<ReadConsistencyStrategy>,
+    /// Content response on write setting.
+    pub content_response_on_write: Option<ContentResponseOnWrite>,
+}
```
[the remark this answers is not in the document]
FabianMeiswinkel: Builder pattern is not required but optional

sdk/cosmos/azure_data_cosmos_driver/src/options/triggers.rs
```diff
@@ -0,0 +1,35 @@
...
+
+/// Collection of triggers to include in a request.
+#[non_exhaustive]
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct TriggerOptions {
+    /// Triggers to execute before the operation.
+    pub pre_triggers: Vec<TriggerInvocation>,
+    /// Triggers to execute after the operation.
+    pub post_triggers: Vec<TriggerInvocation>,
+}
```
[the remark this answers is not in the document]
FabianMeiswinkel: Builder pattern is not require dbut optional

sdk/cosmos/azure_data_cosmos_driver/ARCHITECTURE.md
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/ARCHITECTURE.md
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/ARCHITECTURE.md
FabianMeiswinkel: Motivation like @tvaron3 mentioned was the inten to distinguish Azure Cosmos DB and Fabric Cosmos - so, tired to remove the tension around Azure. But I absolutely don't care - so, as long as we are consistent i am fine with AZURE_COSMOS as well.

sdk/cosmos/azure_data_cosmos_driver/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+#[tokio::main]
+async fn main() -> azure_core::Result<()> {
+    // Use logged-in developer credentials (Azure CLI, azd, etc.)
+    let credential = DeveloperToolsCredential::new(None)?;
+
+    let account = AccountReference::with_credential(
+        Url::parse("https://myaccount.documents.azure.com:443/").unwrap(),
+        credential,
+    );
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/ARCHITECTURE.md
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/ARCHITECTURE.md
FabianMeiswinkel: Noted - will look in to this a bit when working on the transport next week.

sdk/cosmos/azure_data_cosmos_driver/src/models/resource_id.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/models/resource_id.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/models/mod.rs
FabianMeiswinkel: I think i rather get rid of ImmutableContainerProperties for now - the idea was to forntload COnatiner resolution but keey especiall PK in the Identifier/Reference ContainerReference that is passed around between app and driver for every operation - PK we need in many places, itis immutable - so much better than bleeding the need to possibly get container properties asynchronously (nut >99% form cache) everywhere. ImmutableContainerProperties could in theory contain other immutable ContainerProperties like UniqueKey as well - but I don't really see the need for it yet - and if this comes back I can always introduce an immutable containerproperties bag then.

sdk/cosmos/azure_data_cosmos_driver/src/models/mod.rs
FabianMeiswinkel: Fixed the moved form comments in this file - did not find obviously similar comments in other files.

sdk/cosmos/azure_data_cosmos_driver/src/models/mod.rs
FabianMeiswinkel: IMO defining a newtype is easy enough and going forward will probably happen less frequently - so, not spending time on it now at least.

sdk/cosmos/azure_data_cosmos_driver/src/models/etag.rs
FabianMeiswinkel: Renamed to Precondition

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_status.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_result.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_headers.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_headers.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/models/user_agent.rs
FabianMeiswinkel: Yes - in Java + Spark/Kafka the wrapper would have used a UserAgentSuffix of ItsOwnSuffix + custoemr-provided-suffix - I think that worked reasonably well. If there is a need we can always add something dedciated for the wrapper's suffix later

sdk/cosmos/azure_data_cosmos_driver/src/models/user_agent.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/options/env_parsing.rs
FabianMeiswinkel: Added

sdk/cosmos/azure_data_cosmos_driver/src/options/policies.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/options/policies.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/options/policies.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/options/policies.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/models/mod.rs
FabianMeiswinkel: Don't think that is needed - because ContainerProperties and AccountProperties are used internally only  and i moved away form the idea of having typed metadata apis (even metadata apis should just go throgh execute_operation and payload serialization/deserialization should happen upstream in teh SDK IMO).

sdk/cosmos/azure_data_cosmos_driver/src/models/account_reference.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/models/account_reference.rs
FabianMeiswinkel: I'll proceed with Credential - the CosmosPrefix discussion just came up so often with the central team that it is engraved :-) But if Rust community is different API review might also have different outcome there.

sdk/cosmos/azure_data_cosmos_driver/src/models/account_reference.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_operation.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_resource_reference.rs
FabianMeiswinkel: Good catch - I created the ResourceIdentifier for it and used it in most places already - but forgot this one.

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_resource_reference.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/models/request_charge.rs
FabianMeiswinkel: Yes - good catch. Also helpful when reintorducing teh CPU usage

sdk/cosmos/azure_data_cosmos_driver/src/models/resource_id.rs
FabianMeiswinkel: Done

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
FabianMeiswinkel: Added the into_account there - but think into_xxx is probably missing in a few more cases - but can be added when needed later.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
FabianMeiswinkel: returns a Runtime cloned instance under  a reader lock on the Arc<RwLock<RuntimeOptions>> (which is used for Environment/Client settings. Is the question mainly about the name?

analogrelay: Ok, that makes sense. The name is fine if it's not actually cloning.

PR #3775, opened 2026-02-19
[Storage] Improve `BlobName` deserialization, regenerate after `v0.35` emitter release

vincenttran-msft:
`.tsp`: https://github.com/Azure/azure-rest-api-specs/pull/40558
As title states.

Changes

- Regenerates against `v0.35` emitter version
- Adds custom deserializer logic for `BlobName` so that accessing it now yields an `Option<String>`
- Client options bags are now fully generated, removes them from the handwritten code

### DIFF
--- a/sdk/cosmos/azure_data_cosmos_native/include/azurecosmos.h
+++ b/sdk/cosmos/azure_data_cosmos_native/include/azurecosmos.h
@@ -220,8 +220,6 @@ typedef struct cosmos_runtime_options {
 
 
 
-
-
 #ifdef __cplusplus
 extern "C" {
 #endif // __cplusplus
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -266,6 +266,7 @@ dependencies = [
  "azure_identity",
  "opentelemetry",
  "opentelemetry_sdk",
+ "reqwest",
  "tokio",
  "tracing",
  "tracing-subscriber",
--- a/sdk/storage/azure_storage_blob/tsp-location.yaml
+++ b/sdk/storage/azure_storage_blob/tsp-location.yaml
@@ -1,4 +1,4 @@
 directory: specification/storage/Microsoft.BlobStorage
-commit: 2b2ac518533ada45666a11b30c55fe90904c52d6
+commit: fe6a867f358d3feafd203f53882d60bbe4ef62b0
 repo: Azure/azure-rest-api-specs
 additionalDirectories: 


vincenttran-msft:
@heaths @LarryOsterman Any ideas on the breaks in EventHubs? I can't even get the tests to run locally, but they are working fine in Blobs. I am hitting that same missing `x-ms-recording` error:

```
failures:

---- list_checkpoints stdout ----
Error: Error { context: SimpleMessage(DataConversion, "header not found x-recording-id") }
```

sdk/cosmos/azure_data_cosmos_native/include/azurecosmos.h
```diff
@@ -220,6 +220,8 @@ typedef struct cosmos_runtime_options {
 
 
 
+
+
```
vincenttran-msft: Any ideas why this file keeps randomly showing up? It hasn't been in previous PRs since I consciously unstage and revert, but it seems like some formatter really wants these new lines here?

jalauzon-msft (approved):

Cargo.toml
```diff
@@ -115,6 +115,7 @@ opentelemetry = { version = "0.31", features = ["trace"] }
 opentelemetry-appender-tracing = { version = "0.31" }
 opentelemetry-stdout = { version = "0.31" }
 opentelemetry_sdk = "0.31"
+percent-encoding = "2.3"
```
heaths, line 118: Popular, active, and dual MIT/Apache license. Approved.

sdk/storage/azure_storage_blob/src/models/mod.rs
```diff
@@ -10,6 +10,55 @@ pub mod method_options;
...
+/// # Example
+///
+/// ```ignore
+/// #[derive(Deserialize)]
+/// struct MyStruct {
+///     #[serde(deserialize_with = "deserialize_blob_name")]
+///     name: Option<String>,
+/// }
+/// ```
+pub fn deserialize_blob_name<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
```
heaths, line 37: I won't block this PR, but I would recommend following the same pattern we use in `typespec_client_core` for optional vs. non-optional deserialization. We typically have the non-optional at the top level, then a public `option` module (or something like that) with the optional form.

That said, we do that in the core stack where either may be needed. Not as much of a concern here, but do keep in mind that this signature can never change after GA so if we end up needing a non-optional variant later, we'll need a new pattern. That's really all I'm thinking of here: following a consistent pattern.

PR #3778, opened 2026-02-19
Cosmos: initial design doc for hierarchical config

analogrelay:
This PR creates an initial design doc for Hierarchical Configuration in the Cosmos DB SDK for Rust.

The idea is fairly straightforward, we group our options into Option Groups, structs that each participate in one or more layers (Env Vars, Runtime, Account/Driver Instance, Operation). Not all Option Groups participate in all layers (some aren't relevant to individual operations, for example). The structs we accept when creating a Client/Driver, calling an operation method, etc. are all aggregates of the groups that are relevant at that layer. When we're actually _reading_ options, we use macro-generated "View" types which read through the layers to get the correct value for the context in which the option is needed.

Because there's a fair bit of code, the design doc includes the development of a proc-macro `#[derive(CosmosOptions)]` to automatically generate most of these types. The automatically generated types should be _internal_, since they are the View types and the env-var reading logic.

This is based off our current SDK options patterns, so there will be some modifications as we migrate over to the Driver, but if you read past the specifics a bit, I think the general approach here is pretty clear. Once we have some more of the Driver code landed, I can update the design to be more specific.

### DIFF
--- a/.github/workflows/post-apiview.yml
+++ b/.github/workflows/post-apiview.yml
@@ -11,8 +11,9 @@ permissions:
 jobs:
   post-apiview:
     name: After APIView
-    runs-on: ubuntu-24.04
+    runs-on: ["self-hosted", "1ES.Pool=azsdk-pool-github"]
     if: |
+      toJson(github.event.check_run.pull_requests) != '[]' &&
       github.event.check_run.check_suite.app.name == 'Azure Pipelines' && (
       contains(github.event.check_run.name, 'Build Build') ||
       contains(github.event.check_run.name, 'Build Analyze') )
--- a/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
+++ b/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Removed `ContainerClient::patch_item`, `PatchDocument`, and `PatchOperation` temporarily to redesign the PATCH API for safe idempotency. Use a Read/Modify/Replace model with ETag-based optimistic concurrency instead.
 - Changed return type of query methods from `FeedPager<T>` (an alias for `ItemIterator<FeedPage<T>, String>`) to `FeedItemIterator<T>`, which implements `Stream<Item = Result<T>>` and provides `into_pages()` for page-level access. ([#3515](https://github.com/Azure/azure-sdk-for-rust/pull/3515))
 - Introduced `CosmosClientBuilder` for constructing `CosmosClient` instances, replacing constructor-based API. Removed `consistency_level`, `priority`, `throughput_bucket`, `excluded_regions`, `SessionRetryOptions`, triggers, and `IndexingDirective` from options. Simplified `CosmosAccountReference` to take `CosmosAccountEndpoint` directly. Made option struct fields private with getters and `with_*` setters. ([#3744](https://github.com/Azure/azure-sdk-for-rust/pull/3744))
+- Made `CosmosClientBuilder::build()` and `DatabaseClient::container_client()` async to prepare for future cache population (account, collection, partition key range caches).
 - Support for `wasm32-unknown-unknown` has been removed ([#3377](https://github.com/Azure/azure-sdk-for-rust/issues/3377))
 
 ### Bugs Fixed
--- a/sdk/cosmos/azure_data_cosmos/README.md
+++ b/sdk/cosmos/azure_data_cosmos/README.md
@@ -67,7 +67,7 @@ async fn example() -> Result<(), Box<dyn std::error::Error>> {
         .parse()?;
     let account = CosmosAccountReference::with_credential(endpoint, credential);
     let cosmos_client = CosmosClient::builder()
-        .build(account)?;
+        .build(account).await?;
     Ok(())
 }
 ```
@@ -102,7 +102,7 @@ async fn example(cosmos_client: CosmosClient) -> Result<(), Box<dyn std::error::
         value: "2".into(),
     };
 
-    let container = cosmos_client.database_client("myDatabase").container_client("myContainer");
+    let container = cosmos_client.database_client("myDatabase").container_client("myContainer").await;
 
     // Create an item
     container.create_item("partition1", item, None).await?;
--- a/sdk/cosmos/azure_data_cosmos/examples/cosmos/create.rs
+++ b/sdk/cosmos/azure_data_cosmos/examples/cosmos/create.rs
@@ -80,7 +80,7 @@ impl CreateCommand {
                 show_updated,
             } => {
                 let db_client = client.database_client(&database);
-                let container_client = db_client.container_client(&container);
+                let container_client = db_client.container_client(&container).await;
 
                 let pk = PartitionKey::from(&partition_key);
                 let item: serde_json::Value = serde_json::from_str(&json)?;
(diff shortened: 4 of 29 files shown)


sdk/cosmos/azure_data_cosmos/docs/HierarchicalConfigModel.md
```diff
@@ -0,0 +1,657 @@
...
+## Design
+
+### Layers
+
+The configuration model defines four layers, listed from lowest to highest priority:
+
+| Layer | Scope | Lifetime | Example |
+|-------|-------|----------|---------|
+| **Environment** | Process-wide | Static | `AZURE_COSMOS_CONSISTENCY_LEVEL=Session` |
+| **Runtime** | Application-global | App lifetime | Shared defaults across all clients |
```
FabianMeiswinkel, line 51: I think we need Shared mutable defaults - meaning over the lifetime of an app (or cleint - same model) the defaults can change

sdk/cosmos/azure_data_cosmos/docs/HierarchicalConfigModel.md
```diff
@@ -0,0 +1,657 @@
...
+| **Account** | Per `CosmosClient` | Client lifetime | Options passed at client construction |
+| **Operation** | Per request | Single call | Options passed to `create_item()`, etc. |
+
+When a value is needed, the system walks from the highest-priority layer that applies down
```
FabianMeiswinkel, line 55: walking the layers only when needed instead of snapshoting the settings at a  point in time make it harder to debug edge cases - Ideally we woudl snapshot the effective configs at being of an oepration - this allows us to also emit it in diagnostics etc. without the values flip-flopping (assuming account/client-wide values are changing) within the scope of an operation.

sdk/cosmos/azure_data_cosmos/docs/HierarchicalConfigModel.md
```diff
@@ -0,0 +1,657 @@
...
+
+Operation types compose the operation-layer option groups relevant to that operation.
+
+#### Data-plane operations
+
+```rust
+/// Options for item CRUD operations (create, read, replace, upsert, delete, patch).
+#[derive(Clone, Default)]
+#[non_exhaustive]
+pub struct ItemOptions<'a> {
```
FabianMeiswinkel: While I like the elegance of generating this - as a custoemr I would rather prefer havign the options easy to explore in a flat structure - there are rarely so many that it would need groups.

analogrelay: My concern with a flat list is that it makes the layering much messier and less clear. I don't necessarily agree that a flat list is better either, though I could definitely see an argument that putting the operation-specific options a user is highly likely to use (like session tokens, read consistency, etc) directly on the config type. I'll give a look at simplifying some of this without messing up the layering too much

sdk/cosmos/azure_data_cosmos/docs/HierarchicalConfigModel.md
analogrelay: Yep, if these are mutable (and I agree some definitely need to be) I think a better approach will be to snapshot at the start of an operation. I do think the way we can achieve this is by taking Arcs pointing to the current version of each option group though. The more we can avoid interior mutation the safer we'll be when it comes to concurrent mutation.

Let me work up a snapshotting model now that I better understand what you were doing in the driver PR

sdk/cosmos/azure_data_cosmos/docs/HierarchicalConfigModel.md
Pilchie: Does that apply to env vars as well? Should the be read at app startup, or on each operation?

analogrelay: I wondered about that. Theoretically, you CAN change environment variables while a process is executing, but it's not standardized behaviour available across all OSes and involves really weird hacks. I don't think we can expect to support that. Most systems treat env vars as immutable after startup and I think we should too. If we want to provide a built-in way for the options to be changed while the process is executing, we would need to read from (and reload on change) a config file. Or, the customer can do so with whatever mechanism they use to configure the Runtime layer.

FabianMeiswinkel: Env vars are immutable in Java for example anyway (System properties are technically mutable) - and we could map the concepts to both in theory. But I agree - IMO the env variables (which are just one non-code way to provide overridden defaults) shoudl be read-only. We will have first-class API to override settings via API - and that can be mapped to any mutable mechanism like system properties in Java if we would really want to.

sdk/cosmos/azure_data_cosmos/docs/HierarchicalConfigModel.md
```diff
@@ -0,0 +1,798 @@
...
+   they would move to the top.
+
+3. **Thread-safe option group replacement** — Replacing an `Arc<T>` field on a layer struct
+   requires `&mut` access, meaning the caller needs an exclusive handle to the component
+   that owns the layer options. This is fine for single-threaded or task-local usage, but
+   concurrent replacement from multiple threads would require the user to wrap the client
+   (or the layer struct) in an `RwLock` or similar. Should the SDK provide a built-in
+   thread-safe mutation mechanism (e.g., using `ArcSwap` or an internal `RwLock` around
+   each `Arc` field), or leave this as a user responsibility? The current design favors
+   simplicity — users who need concurrent replacement can add their own synchronization.
```
analogrelay, line 798: This is something we do need to solve. Otherwise, you can't actually mutate options once you share the client between threads (unless you provide your own lock around it). An alternate design is something like this:

```rust
pub struct DriverOptions {
    retry: RwLock<Arc<RetryOptions>>
}
```

That would allow mutating the `DriverOptions::retry` field with a `&`, and thus sharing the options across threads (we may also want to wrap the `DriverOptions` itself in an `Arc`, but that would just be for cross-thread sharing rather than for snapshotting.

However, that does create a fair bit of CPU contention because it requires acquiring a read-lock every time we want to snapshot the options. That's fairly cheap but certainly not free. There are also write-time costs, but I care less about those.

There's an external crate called [`arc-swap`](http://docs.rs/arc-swap) that provides an `ArcSwap` type that uses a  more complex strategy to avoid contention on a single read lock by using some atomic tricks. It's fairly well used and stable. In my benchmarking, it's possible to achieve an order of magnitude of speed improvement using this.

My inclination is we could start with RwLock and move to ArcSwap if we wanted to later. As part of using RwLock, we'd keep the actual field private, which means we could change to ArcSwap as a non-breaking change in the future.

FabianMeiswinkel: I looekd at arc-swap as well for the Dirver but decided to use Arc<RwLock<RuntimeOptions>> there - see https://github.com/Azure/azure-sdk-for-rust/blob/33bcffc54e73698c2eeeeb58f2ca538c157ba02c/sdk/cosmos/azure_data_cosmos_driver/src/options/runtime_options.rs#L188 - IMO RwLock seems sufficient (at least until we have hard evidence that is not). We read this on the hotpath - actual changes at higher levels will be rare. And even while RwLock will have overhead for reads - we have to snapshot options anyway - that will be more expensive - and on our "hot path" we do actual I/O - so, if we are at a point where the read under RwLock is the bottleneck we are talking about micro-optimization for sure :-)

FabianMeiswinkel (approved):
LGTM now

PR #3781, opened 2026-02-20
Remove WASM support

### LINKED ISSUE #3377: Eliminate Send conditions on wasm32-unknown-unknown

heaths:
Resolves #3377

### DIFF
--- a/sdk/core/typespec_client_core/CHANGELOG.md
+++ b/sdk/core/typespec_client_core/CHANGELOG.md
@@ -9,7 +9,6 @@
 
 ### Breaking Changes
 
-- Removed explicit build support for WASM.
 - Support for `wasm32-unknown-unknown` has been removed ([#3377](https://github.com/Azure/azure-sdk-for-rust/issues/3377))
 
 ### Bugs Fixed


FabianMeiswinkel (approved):
sdk/cosmos LGTM

analogrelay (approved):
Cosmos still LGTM

LarryOsterman (approved):

chlowell (approved):

PR #3796, opened 2026-02-24
[Cosmos] Application regions

simorenoh:
When a user sets `application_region` on the CosmosClientBuilder, the SDK now automatically generates an internal preferred regions list sorted by geographic proximity (round-trip time) from that region.

`with_application_region` is now the only public API for influencing regional routing preference. The previous `with_application_preferred_regions` API has been removed — customers can no longer specify an explicit ordering of preferred regions.

  Changes

- New module region_proximity.rs — Contains a static RTT lookup table covering 96 Azure regions. Each source region maps to a pre-sorted list of all other regions by ascending round-trip time. The generate_preferred_region_list function returns the sorted list for a given source region, or None for unrecognized regions.
- cosmos_client_builder.rs — build() now validates application_region against known region constants (returning an error for unknown regions) and calls generate_preferred_region_list to populate the internal preferred regions passed to GlobalEndpointManager.
- Removed `with_application_preferred_regions`.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
+++ b/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
@@ -4,10 +4,11 @@
 
 ### Features Added
 
-- Added `excluded_regions` to `ItemOptions` for additional regional routing options. ([#3602](https://github.com/Azure/azure-sdk-for-rust/pull/3602))
+- Added `with_excluded_regions` to `ItemOptions` for additional regional routing options. ([#3602](https://github.com/Azure/azure-sdk-for-rust/pull/3602))
 - Added `effective_preferred_regions` to the client, ensuring multi-region accounts use all regions for cross-regional availability without supplying regional preferences to their client. ([#3602](https://github.com/Azure/azure-sdk-for-rust/pull/3602))
 - Added basic multi-region writes support. ([#3482](https://github.com/Azure/azure-sdk-for-rust/pull/3482) and [#3495](https://github.com/Azure/azure-sdk-for-rust/pull/3495))
 - Added new `CosmosResponse` that wraps `azure_core::Response` for all operations except queries. ([#3622](https://github.com/Azure/azure-sdk-for-rust/pull/3622))
+- Added transactional batch support for executing multiple operations atomically within the same partition key. ([#3664](https://github.com/Azure/azure-sdk-for-rust/pull/3664))
 - Added fault injection support for testing cosmosdb clients in disaster scenarios. Fault injection is behind the feature flag `fault_injection`. ([#3599](https://github.com/Azure/azure-sdk-for-rust/pull/3599))
 
 ### Breaking Changes
--- a/.github/workflows/post-apiview.yml
+++ b/.github/workflows/post-apiview.yml
@@ -11,7 +11,7 @@ permissions:
 jobs:
   post-apiview:
     name: After APIView
-    runs-on: ["self-hosted", "1ES.Pool=azsdk-pool-github"]
+    runs-on: ["self-hosted", "1ES.Pool=azsdk-pool-github-runners"]
     if: |
       toJson(github.event.check_run.pull_requests) != '[]' &&
       github.event.check_run.check_suite.app.name == 'Azure Pipelines' && (
--- a/eng/common/TestResources/New-TestResources.ps1
+++ b/eng/common/TestResources/New-TestResources.ps1
@@ -179,7 +179,7 @@ try {
         }
         Write-Verbose "Overriding test resources search directory to '$root'"
     }
-    
+
     $templateFiles = @()
 
     "$ResourceType-resources.json", "$ResourceType-resources.bicep" | ForEach-Object {
@@ -203,7 +203,7 @@ try {
 
     # returns empty string if $ServiceDirectory is not set
     $serviceName = GetServiceLeafDirectoryName $ServiceDirectory
-    
+
     # in ci, random names are used
     # in non-ci, without BaseName, ResourceGroupName or ServiceDirectory, all invocations will
     # generate the same resource group name and base name for a given user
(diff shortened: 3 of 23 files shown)


sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -254,7 +248,16 @@ impl CosmosClientBuilder {
             None,
         );
 
-        let preferred_regions = self.options.application_preferred_regions.clone();
+        let preferred_regions = if let Some(ref region) = self.options.application_region {
+            crate::region_proximity::generate_preferred_region_list(region).ok_or_else(|| {
+                azure_core::Error::new(
+                    azure_core::error::ErrorKind::Other,
+                    format!("unknown application region: {region}"),
+                )
```
analogrelay: Could we just fall back to using the regions returned in the Database Account Properties in the order specified there? I'm not sure we should _fail to build_ if a user specifies a region unknown to the SDK. I know Proximity API will help with that in the future but do we know that will land by GA? Do we know it will work in all cases?

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -283,3 +286,43 @@ impl CosmosClientBuilder {
         })
     }
 }
+
+#[cfg(all(test, feature = "key_auth"))]
```
analogrelay: Does this need to be limited to the key_auth feature? We're not actually connecting, we could use `with_credential`

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -17,6 +16,7 @@
...
 - Changed return type of query methods from `FeedPager<T>` (an alias for `ItemIterator<FeedPage<T>, String>`) to `FeedItemIterator<T>`, which implements `Stream<Item = Result<T>>` and provides `into_pages()` for page-level access. ([#3515](https://github.com/Azure/azure-sdk-for-rust/pull/3515))
 - Introduced `CosmosClientBuilder` for constructing `CosmosClient` instances, replacing constructor-based API. Removed `consistency_level`, `priority`, `throughput_bucket`, `excluded_regions`, `SessionRetryOptions`, triggers, and `IndexingDirective` from options. Simplified `CosmosAccountReference` to take `CosmosAccountEndpoint` directly. Made option struct fields private with getters and `with_*` setters. ([#3744](https://github.com/Azure/azure-sdk-for-rust/pull/3744))
+- Removed `with_application_preferred_regions` API. Use `with_application_region` to set a single region; the SDK generates preferred regions by geographic proximity. ([#3796](https://github.com/Azure/azure-sdk-for-rust/pull/3796))
```
analogrelay:
```suggestion
- Removed `with_application_preferred_regions` API. Use `with_application_region` to set the Azure region the app is executing in (or the closest region to the actual location you're running in); the SDK generates preferred regions by geographic proximity. ([#3796](https://github.com/Azure/azure-sdk-for-rust/pull/3796))
```

sdk/cosmos/azure_data_cosmos/src/region_proximity.rs
```diff
@@ -0,0 +1,9780 @@
...
+
+use crate::regions::RegionName;
+
+/// Generates a list of preferred regions sorted by geographic proximity (RTT)
+/// from the given source region.
+///
+/// Returns `None` if the source region is not recognized.
+/// The returned list is sorted by ascending round-trip time (RTT) in milliseconds,
+/// with the source region itself always first (RTT = 0).
+pub(crate) fn generate_preferred_region_list(source: &RegionName) -> Option<Vec<RegionName>> {
```
analogrelay: We probably don't need to allocate for this, especially if we adjust how we generate this. I don't think we actually need the RTT at runtime, just the ordered list of preferred regions for a given region. If we use that, we can return `Option<&'static [RegionName]>` from this and just return a pointer to the static array

sdk/cosmos/azure_data_cosmos/src/region_proximity.rs
```diff
@@ -0,0 +1,9780 @@
...
+        "westeurope" => Some(&WESTEUROPE_RTT),
+        "westindia" => Some(&WESTINDIA_RTT),
+        "westus" => Some(&WESTUS_RTT),
+        "westus2" => Some(&WESTUS2_RTT),
+        "westus3" => Some(&WESTUS3_RTT),
+        _ => None,
+    }
+}
+
+static AUSTRALIACENTRAL_RTT: [(&str, u64); 96] = [
```
analogrelay: We should try to use the existing constants in `RegionName`.

sdk/cosmos/azure_data_cosmos/src/routing/location_cache.rs
```diff
@@ -1479,4 +1481,313 @@ mod tests {
...
+        assert_eq!(
+            read_endpoints[1].as_str(),
+            "https://test-westus.documents.azure.com/",
+            "West US should be second"
+        );
+        assert_eq!(
+            read_endpoints[2].as_str(),
+            "https://test-westeurope.documents.azure.com/",
+            "West Europe should be third (farthest)"
+        );
```
analogrelay: nit: This could just be done by using a `.map(...).collect()` to get the list of endpoints into a simple `Vec<&str>` and then `assert_eq!` over the whole vector. But that's non-blocking style guidance.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
simorenoh: I followed the .NET SDK implementation for this decision - which I believe was made since otherwise the regional behavior may not be as transparent to users otherwise (we might choose to do provide an ordering/ preference that they don't necessarily expect or want). What you're saying makes sense from a non-blocking perspective of letting the application get started anyway though, maybe something we could document. @FabianMeiswinkel - thoughts on whether this should be changed for Rust?

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -249,12 +255,14 @@ impl CosmosClientBuilder {
...
-                )
-            })?
+            crate::region_proximity::generate_preferred_region_list(region)
+                .ok_or_else(|| {
+                    azure_core::Error::new(
+                        azure_core::error::ErrorKind::Other,
+                        format!("unknown application region: {region}"),
+                    )
+                })?
+                .to_vec()
```
analogrelay: We'll do this in a follow-up, but calling `.to_vec()` here sorta defeats the memory savings from returning `Option<&'static [RegionName]>`. We can find a more efficient way to store this in `GlobalEndpointManager` so that it can accept a `&'static [RegionName]` OR a `Vec<RegionName>` in the future.

But let's not churn that here unless we see this as an issue in profiles.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
simorenoh: went ahead with the suggestion, we won't fail

PR #3797, opened 2026-02-24
[Storage] Align deserializer w/ `typespec_client_core` offerings + increased test coverage, fix logging name conflict

vincenttran-msft:
`.tsp:` https://github.com/Azure/azure-rest-api-specs/pull/40707

- Adds optional and non-optional variants for our deserialization offerings to align with `typespec_client_core`
- Adds a checked in E2E test with the unicode characters that cause the encoded field to be set to `True`, to test the codepath for our auto-encoding deserializer
- Fixes the following issue:
>warning: output filename collision at /home/heaths/src/azure-sdk-for-rust2/target/debug/examples/storage_logging
  |
  = note: the example target `storage_logging` in package `azure_storage_queue v0.4.0 (/home/heaths/src/azure-sdk-for-rust2/sdk/storage/azure_storage_queue)` has the same output filename as the example target `storage_logging` in package `azure_storage_blob v0.10.0 (/home/heaths/src/azure-sdk-for-rust2/sdk/storage/azure_storage_blob)`
  = note: this may become a hard error in the future; see <[https://github.com/rust-lang/cargo/issues/6313>](https://github.com/rust-lang/cargo/issues/6313%3E)
  = help: consider changing their names to be unique or compiling them separately
warning: output filename collision at /home/heaths/src/azure-sdk-for-rust2/target/debug/examples/storage_logging.dwp
  |
  = note: the example target `storage_logging` in package `azure_storage_queue v0.4.0 (/home/heaths/src/azure-sdk-for-rust2/sdk/storage/azure_storage_queue)` has the same output filename as the example target `storage_logging` in package `azure_storage_blob v0.10.0 (/home/heaths/src/azure-sdk-for-rust2/sdk/storage/azure_storage_blob)`
  = note: this may become a hard error in the future; see <[https://github.com/rust-lang/cargo/issues/6313>](https://github.com/rust-lang/cargo/issues/6313%3E)
  = help: consider changing their names to be unique or compiling them separately


sdk/storage/azure_storage_blob/src/models/mod.rs
```diff
@@ -10,52 +10,104 @@ pub mod method_options;
...
+            };
+
+            if blob_name.encoded.unwrap_or_default() {
+                let decoded = percent_decode_str(&content)
+                    .decode_utf8()
+                    .map_err(D::Error::custom)?;
+                Ok(Some(decoded.into_owned()))
+            } else {
+                Ok(Some(content))
+            }
```
heaths, line 109: You can actually avoid the duplication here if you just return early if it's `None` (or whatever is an appropriate action) then delegate to your other function to do the rest with the `Some(blob_name)` captured `blob_name`.

heaths (approved):

PR #3800, opened 2026-02-25
Making AccountProperties pub(crate)

FabianMeiswinkel:


sdk/cosmos/azure_data_cosmos/src/models/account_properties.rs
```diff
@@ -2,132 +2,37 @@
...
-    pub enable_multiple_write_locations: bool,
-
-    /// Indicates if continuous backup (point-in-time restore) is enabled for the account.
-    pub continuous_backup_enabled: bool,
-
-    /// Enables synchronous commit across N regions for stricter durability guarantees.
-    pub enable_n_region_synchronous_commit: bool,
-
-    /// Allows failover at a per-partition granularity instead of full-region only.
-    pub enable_per_partition_failover_behavior: bool,
```
kundadebdatta, line 106: `enable_per_partition_failover_behavior` would be required for PPAF dynamic enablement.

FabianMeiswinkel: As discussed offline - removed dead_code - knowing that we will have to bring back a few of the properties. But that should be done when we atrt using them.

simorenoh (approved):

PR #3803, opened 2026-02-25
Cosmos: Spec for new Rust SDK options

analogrelay:
I used Copilot to conduct an analysis of options across the .NET, Java, Python, and Go SDKs (I didn't include JS but I don't see a major difference in options there). I've uploaded [the raw output of that analysis to a gist](https://gist.github.com/analogrelay/8d209ee7aecb6796dea725c29f0f6ca9).

Using that analysis, I built this proposal to migrate all our options towards the hierarchical config group. This document lists out the option groups we want, the layers they apply to, and the types/env vars they use. I covered most of what we've talked about before in terms of making breaking changes (like removing Consistency Level), but this is the place to make sure those are covered before we move forward with applying this change.

You may find it easier to review this [rendered version without diff markers](https://github.com/analogrelay/azure-sdk-for-rust/blob/13ab241b9db59e5b9a5cedc42040fa96e051dfa8/sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md), since it's a brand new file.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
+++ b/sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
@@ -1,6 +1,6 @@
 # Azure Cosmos DB Rust SDK — Configuration Options Specification
 
-This document specifies the configuration option types for the Rust SDK (`azure_data_cosmos`), aligned with the [Hierarchical Configuration Model](HierarchicalConfigModel.md).
+This document specifies the configuration option types for the Rust SDK (`azure_data_cosmos`), aligned with the Hierarchical Configuration Model.
 
 ## Table of Contents
 


sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
```diff
@@ -0,0 +1,570 @@
...
+    /// Session consistency — monotonic reads/writes within a session.
+    Session,
+}
+```
+
+Must implement `FromStr` for environment variable parsing.
+
+> **Note:** `ConsistencyLevel` (with variants `Strong`, `BoundedStaleness`, `Session`, `ConsistentPrefix`, `Eventual`) is retained as a model type for account-level properties returned by the service. It is **not** used in configuration options.
+
+#### `PriorityLevel`
```
FabianMeiswinkel, line 104: IMO we need a group
Server-side throughput control options
- Priority Level and ThroughputBucket
- We will liekly also need to supprot client-side throughput control
- The mdoel in Java is we expose ThroughutControl groups that can be referenced by identifier - and are eitehr client-or server side - so, throughput control group is the data that is referenced in OperationOptions and possibly defaults in client --> runtime

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
```diff
@@ -0,0 +1,570 @@
...
+}
+```
+
+Must implement `FromStr` for environment variable parsing.
+
+#### `IndexingDirective`
+
+**Removed.** See [§6.4](#64-indexing-directive-indexing_directive).
+
+#### `EtagCondition`
```
FabianMeiswinkel: My preference would be Precondition instead of EtagCondition

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
```diff
@@ -0,0 +1,570 @@
...
+
+#### `SessionToken`
+
+Retained as-is. Wraps `String`. Implements `From<String>`, `Display`, `Clone`.
+
+---
+
+## 3. Option Groups
+
+### 3.1 `RequestOptions`
```
FabianMeiswinkel: Let us please not use any RequestXXX terms - all our options (with public surface area) have at least the scope of an operation (e. g. for all requests within the scope of teh operation - like hedging etc.)

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
```diff
@@ -0,0 +1,570 @@
...
+```rust
+#[derive(CosmosOptions)]
+#[options(layers(runtime, account, operation))]
+pub struct RequestOptions { /* fields below */ }
+```
+
+| Option | Type | Env Var | Notes |
+|---|---|---|---|
+| `read_consistency_strategy` | `Option<ReadConsistencyStrategy>` | `AZURE_COSMOS_READ_CONSISTENCY_STRATEGY` | Read consistency for the operation. Replaces the legacy `consistency_level` field. The SDK enforces weakening-only semantics relative to the account default. |
+| `priority` | `Option<PriorityLevel>` | `AZURE_COSMOS_PRIORITY` | Priority-based execution level (`High` or `Low`). |
```
FabianMeiswinkel: For thorughput - you have the otpion of adding three options
priority, throughputbucket and maybe clientthroughputgroup or such

But instead using its own OptionGroup makes more sense - and also allows us to merge the competing concepts under one API (throughputControlGroup)

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
```diff
@@ -0,0 +1,570 @@
...
+| `throughput_bucket` | `Option<usize>` | `AZURE_COSMOS_THROUGHPUT_BUCKET` | Throughput control bucket for the request. |
+| `excluded_regions` | `Option<Vec<RegionName>>` | `AZURE_COSMOS_EXCLUDED_REGIONS` | Regions to exclude from routing. `None` inherits from a lower layer; `Some(vec![])` explicitly clears exclusions. Env var is comma-separated (e.g. `"West US,East US"`). |
+| `content_response_on_write` | `Option<bool>` | `AZURE_COSMOS_CONTENT_RESPONSE_ON_WRITE` | Whether write operations return the resource body in the response. Only applicable to write operations; ignored by reads and queries. Cascades from runtime → account → operation, matching .NET/Java/Go behavior. |
```
FabianMeiswinkel, line 159: do we even need this? What are the scenarios where we expect customers to return the entire document? Patch maybe? I guess for patch it woudl make sense.

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
```diff
@@ -0,0 +1,570 @@
...
+
+### 3.2 `ConnectionOptions`
+
+**Layers:** Runtime, Account
+
+Options controlling network connection behavior. Not available at the operation layer because connections are managed at the client level.
```
FabianMeiswinkel, line 165: Should not be available at accoutn level either - connection pools should be owned by the runtime (shared tenant apps isolating in 1 avvount per tenant)

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
```diff
@@ -0,0 +1,570 @@
...
+
+```rust
+#[derive(CosmosOptions)]
+#[options(layers(runtime, account))]
+pub struct RetryOptions { /* fields below */ }
+```
+
+| Option | Type | Env Var | Notes |
+|---|---|---|---|
+| `session_retry` | `Option<SessionRetryOptions>` | — | Nested group for session-consistency retry behavior on 404/1002 errors. Marked `#[option(nested)]`. |
```
FabianMeiswinkel, line 225: I would argue we don't need this - we just have options in older SDKs because we could not change the default behavior - now remote_region_preferred seems to eb the obvious choice? Especially because Gateway kind of takes this decision out of our hands anyway (at least Gateway 2.0 - it will always use siilar algorithm like .Net and Java in Remote_region_Preferred)

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
```diff
@@ -0,0 +1,570 @@
...
+| `enable_partition_level_circuit_breaker` | `Option<bool>` | `AZURE_COSMOS_ENABLE_PARTITION_LEVEL_CIRCUIT_BREAKER` | Enable partition-level circuit breaker for transient failure isolation. |
+| `disable_partition_level_failover` | `Option<bool>` | `AZURE_COSMOS_DISABLE_PARTITION_LEVEL_FAILOVER` | Disable automatic partition-level failover to other replicas. |
```
FabianMeiswinkel: This should not even exist - disabling PPAF on an acocunt where PPAF is enabled is danerous and would degarde availability.

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
```diff
@@ -0,0 +1,570 @@
...
+```rust
+#[derive(CosmosOptions)]
+#[options(layers(runtime, account))]
+pub struct RetryOptions { /* fields below */ }
+```
+
+| Option | Type | Env Var | Notes |
+|---|---|---|---|
+| `session_retry` | `Option<SessionRetryOptions>` | — | Nested group for session-consistency retry behavior on 404/1002 errors. Marked `#[option(nested)]`. |
+| `enable_partition_level_circuit_breaker` | `Option<bool>` | `AZURE_COSMOS_ENABLE_PARTITION_LEVEL_CIRCUIT_BREAKER` | Enable partition-level circuit breaker for transient failure isolation. |
```
FabianMeiswinkel: defautl should be true - so maybe rename to disabled_xxx

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
```diff
@@ -0,0 +1,570 @@
...
+| `disable_partition_level_failover` | `Option<bool>` | `AZURE_COSMOS_DISABLE_PARTITION_LEVEL_FAILOVER` | Disable automatic partition-level failover to other replicas. |
+
+### 3.6 `SessionRetryOptions`
```
FabianMeiswinkel, line 227: Remove altogether

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
```diff
@@ -0,0 +1,570 @@
...
+```rust
+#[derive(CosmosOptions)]
+#[options(layers(runtime, account))]
+pub struct CosmosAccountOptions { /* fields below */ }
+```
+
+| Option | Type | Env Var | Notes |
+|---|---|---|---|
+| `user_agent_suffix` | `Option<String>` | `AZURE_COSMOS_USER_AGENT_SUFFIX` | Application identifier appended to the User-Agent header for telemetry. |
+| `account_initialization_custom_endpoints` | `Option<HashSet<String>>` | `AZURE_COSMOS_CUSTOM_ENDPOINTS` | Custom endpoints for initial account discovery (private endpoints, etc.). Env var is comma-separated. |
```
FabianMeiswinkel: Why String not Uri?

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
FabianMeiswinkel: Or make it quirk?

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
FabianMeiswinkel: Or make it quirk

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
```diff
@@ -0,0 +1,570 @@
...
+
+```rust
+#[derive(CosmosOptions)]
+#[options(layers(runtime, account))]
+pub struct QuirkOptions { /* fields below */ }
+```
+
+| Option | Type | Env Var | Notes |
+|---|---|---|---|
+| `upgrade_consistency_to_local_quorum` | `Option<bool>` | `AZURE_COSMOS_UPGRADE_CONSISTENCY_TO_LOCAL_QUORUM` | Upgrade bounded-staleness / strong reads to local-quorum when possible. Only modify when advised by Azure Support. |
```
FabianMeiswinkel: Remove - this is available via ReadConsistencyStrategy

FabianMeiswinkel: LatestCommitted

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
```diff
@@ -0,0 +1,570 @@
...
+### 5.3 `QueryOptions`
+
+Options for query operations (`query_items`, `query_items_single_partition`).
+
+```rust
+#[derive(Clone, Default)]
+#[non_exhaustive]
+pub struct QueryOptions {
+    // Layered option group
+    pub request: RequestOptions,
```
FabianMeiswinkel: query metrics, query advisor, PK?

FabianMeiswinkel (changes requested):
I love the overall concept - few changes IMO are required to merge.

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
analogrelay: I think removing it, when it's been present in SDKs for a while, would be unnecessarily disruptive

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
analogrelay: It should absolutely be a URL, I just missed that on review. Good catch!

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
analogrelay: Yep, though I think PK should be an argument to query_items so the user is explicit about if they want single-partition (and which partition key) or cross-partition

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
analogrelay: I've actually decided to defer throughput control to a separate spec, since it's got enough complication to it and it looks like we have a pretty in-depth pattern in Java that we might want to bring over here.

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
analogrelay: See my other comment, I'm going to remove it for now and set up a new Throughput Control spec.

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
FabianMeiswinkel: Makes sense to me.

sdk/cosmos/azure_data_cosmos/docs/ConfigurationOptions.md
FabianMeiswinkel: Makes sense

PR #3809, opened 2026-02-25
Removed URL dependency from opentelemetry

LarryOsterman:
Bonus: Use workspace dependencies for identity deployed_live_test.

### DIFF
--- a/sdk/identity/azure_identity/tests/tools/deployed_live_test/Cargo.toml
+++ b/sdk/identity/azure_identity/tests/tools/deployed_live_test/Cargo.toml
@@ -15,7 +15,7 @@ axum = { version = "0.8", default-features = false, features = [
 azure_core = { path = "../../../../../core/azure_core" }
 azure_identity = { path = "../../../" }
 azure_storage_blob = { path = "../../../../../storage/azure_storage_blob" }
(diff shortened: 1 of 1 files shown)


sdk/identity/azure_identity/tests/tools/deployed_live_test/Cargo.toml
```diff
@@ -7,11 +7,15 @@ publish = false
...
-azure_core = { path = "../../../../../core/azure_core" }
-openssl = { version = "0.10", features = ["vendored"]}
-tokio = { version = "1.0", features = ["full"] }
-url = "2.5"
-axum = { version = "0.8", default-features = false, features = ["http1", "tokio", "query"] }
-serde = { version = "1.0", features = ["derive"] }
+openssl = { workspace = true, features = ["vendored"] }
+serde = { workspace = true, features = ["derive"] }
+tokio = { workspace = true, features = ["full"] }
+url.workspace = true
```
heaths: This crate isn't in the workspace, though, nor do I think we should add it. I'm already worried when we start adding a lot more crates in the future about build perf.

heaths (changes requested):

PR #3813, opened 2026-02-26
cosmos: reference azure_core 0.32 for mid-band release

analogrelay:
We need to do a release ASAP for a customer need, and we're depending on unreleased code in azure_core. This PR reverts that temporarily and moves our dependency from path-based to depending on the released versions of core libraries (0.32).

I'll open a follow-up PR after this is released to go back to referencing the live code, so that we can release again after the next Core release wave.

### DIFF
--- a/eng/common/scripts/Delete-RemoteBranches.ps1
+++ b/eng/common/scripts/Delete-RemoteBranches.ps1
@@ -8,6 +8,12 @@ param(
   # For sync-eng/common work, we use regex as "^sync-eng/common.*-(?<PrNumber>\d+).*$".
   # For sync-.github/workflows work, we use regex as "^sync-.github/workflows.*-(?<PrNumber>\d+).*$".
   $BranchRegex,
+  # When set, directly delete this exact branch name without querying all branches.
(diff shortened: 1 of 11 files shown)


eng/scripts/verify-dependencies.rs
```diff
@@ -24,6 +24,10 @@ static EXEMPTIONS: &[(&str, &str)] = &[
     ("azure_core_test", "dotenvy"),
     ("azure_canary", "serde"),
     ("azure_data_cosmos_native", "cbindgen"),
+    // Temporary: Allow azure_data_cosmos to release depending on the latest release of azure_core and azure_identity
+    ("azure_data_cosmos_native", "azure_core"),
+    ("azure_data_cosmos", "azure_core"),
+    ("azure_data_cosmos", "azure_identity"),
```
heaths, line 30: Per offline discussion, this is fine temporarily; however, if you're going to revert this and then open a [release branch](https://azure.github.io/azure-sdk/policies_repobranching.html#release-branches) in case you have to ship more releases anyway, why not do it now? Then you don't need to revert this later after you release this next version.

heaths (approved):

PR #3815, opened 2026-02-26
Document breaking model change mitigation

heaths:


RickWinter (approved):
Change looks good. The note is more for service authors than SDK consumers. We should look for a way to distinguish consumer vs author notes. No need to address in this PR though. Just something for us to clean up across all docs.

Issue #3783, opened 2026-02-20
Cosmos: Add PPAF Dynamic Enablement


kundadebdatta:
### Acceptance Criteria:

- Add support for PPAF and PPCB dynamic enablement. PPAF and PPCB should be enabled if partition level failover is enabled at the cosmos account level. If not - they must be disabled.
- The PPCB enablement should honor the client options circuit breaker config.

PR #3831, opened 2026-02-27
Cosmos: Adds PPAF Dynamic Enablement Support

kundadebdatta:
### Summary

This PR adds dynamic enablement support for per-partition automatic failover (PPAF) and per-partition circuit breaker in the Cosmos DB Rust SDK. Previously, these flags were set only at client construction time and could not be updated in response to service-side configuration changes. With this PR, the flags are dynamically refreshed whenever account properties are fetched from the Cosmos DB service.

### Motivation

The Cosmos DB service can signal per-partition failover behavior via account properties (`enable_per_partition_failover_behavior`). Without dynamic enablement, SDK-side partition-level failover and circuit breaker settings remain static for the client's lifetime, even if the service-side configuration changes. This PR closes that gap by wiring account property refreshes to partition-level failover configuration updates.

#### `GlobalEndpointManager` — Background Refresh & Callback Support

- **`new()` now returns `Arc<Self>`** instead of plain `Self`. The manager is always wrapped in `Arc` since it needs to spawn a background task holding a weak self-reference.
- **Account refresh callback**: Added `OnAccountRefreshCallback` type alias and `set_on_account_refresh_callback()` method. A callback can be registered to be invoked whenever `refresh_location()` fetches *new* account properties from the service (not from cache).
- **Background account refresh loop**: Added `initialize_and_start_background_account_refresh()` and `initiate_background_account_refresh_loop()` to periodically refresh account properties on a configurable interval (default: 10 seconds). The loop uses `Weak<Self>` to avoid preventing the manager from being dropped.
- **`BackgroundTaskManager` integration**: Background tasks are spawned via `BackgroundTaskManager` so they are cancelled when the client is dropped.
- **Custom `Debug` impl**: Replaced `#[derive(Debug)]` with a manual `Debug` implementation to handle the non-Debug callback field.
- **`refresh_location()` now invokes callback**: Tracks whether an HTTP call was made and invokes the registered callback with the fetched `AccountProperties` when applicable.
- **`Duration` migration**: Switched from `std::time::Duration` to `azure_core::time::Duration` for consistency with the rest of the SDK.

#### `GlobalPartitionEndpointManager` — Dynamic Configuration Methods

- **`configure_partition_level_automatic_failover()`**: New public method to dynamically set whether per-partition automatic failover is enabled.
- **`configure_per_partition_circuit_breaker()`**: New public method to dynamically set whether per-partition circuit breaker is enabled.

#### `CosmosClient` — Callback Wiring

- Both `new()` and `with_connection_string()` now register an `on_account_refresh` callback on `GlobalEndpointManager` that dynamically updates the `GlobalPartitionEndpointManager` flags:
  - `partition_level_automatic_failover_enabled` ← from `account_props.enable_per_partition_failover_behavior`
  - `partition_level_circuit_breaker_enabled` ← from `account_props.enable_per_partition_failover_behavior || options.enable_partition_level_circuit_breaker`
- Removed the redundant duplicate import of `GlobalPartitionEndpointManager`.

#### `AsyncCache`

- Migrated from `std::time::Duration` to `azure_core::time::Duration` for consistency across all usages.

#### Test Updates

- Updated all call sites across `container_connection.rs`, `client_retry_policy.rs`, `metadata_request_retry_policy.rs`, `container_cache.rs`, and `global_partition_endpoint_manager.rs` to accommodate `GlobalEndpointManager::new()` returning `Arc<Self>` (removed redundant `Arc::new()` wrappers).

### Files Changed

| File | Change |
|------|--------|
| `src/clients/cosmos_client.rs` | Callback registration, `Arc` return type adoption |
| `src/handler/container_connection.rs` | Test helper updates for `Arc` return |
| `src/retry_policies/client_retry_policy.rs` | Test helper updates for `Arc` return |
| `src/retry_policies/metadata_request_retry_policy.rs` | Test helper updates for `Arc` return |
| `src/routing/async_cache.rs` | `Duration` type migration |
| `src/routing/container_cache.rs` | Test helper updates for `Arc` return |
| `src/routing/global_endpoint_manager.rs` | Background refresh loop, callback support, `Arc<Self>` return, documentation |
| `src/routing/global_partition_endpoint_manager.rs` | Dynamic configuration methods, test updates |

### Testing

- All existing unit tests updated to work with the new `Arc<Self>` return type.
- Existing tests for `GlobalPartitionEndpointManager` verify flag state after construction.
- The dynamic enablement methods (`configure_partition_level_automatic_failover`, `configure_per_partition_circuit_breaker`) use atomic stores and are safe for concurrent access.

### Related

- Issue: #3783

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
@@ -15,7 +15,7 @@ use azure_core::Error;
 use std::collections::HashMap;
 use std::fmt::Debug;
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{Arc, Mutex, Weak};
+use std::sync::{Arc, Mutex, OnceLock, Weak};
 use tracing::info;
 use url::Url;
 
@@ -44,8 +44,11 @@ pub(crate) struct GlobalEndpointManager {
     /// Cache for account properties with 600 second TTL to reduce redundant service calls
     account_properties_cache: AsyncCache<&'static str, AccountProperties>,
 
-    /// Optional callback invoked when account properties are refreshed via HTTP call
-    on_account_refresh: Mutex<Option<OnAccountRefreshCallback>>,
+    /// Optional callback invoked when account properties are refreshed via HTTP call.
+    /// Uses `OnceLock` because the callback is registered exactly once during client
+    /// construction and never changes afterward. This avoids the need for a `Mutex`
+    /// (which is error-prone in async code) and is completely lock-free on reads.
+    on_account_refresh: OnceLock<OnAccountRefreshCallback>,
 
     /// Flag indicating if the background connection initialization task is active.
     background_account_refresh_active: AtomicBool,
@@ -108,7 +111,7 @@ impl GlobalEndpointManager {
             location_cache,
             pipeline,
             account_properties_cache,
-            on_account_refresh: Mutex::new(None),
+            on_account_refresh: OnceLock::new(),
             background_account_refresh_active: AtomicBool::new(false),
             background_task_manager: BackgroundTaskManager::new(),
             background_account_refresh_interval: Duration::seconds(
@@ -126,11 +129,15 @@ impl GlobalEndpointManager {
     /// fetches new account properties from the service (not when serving from cache). This is useful
     /// for updating partition-level failover configurations when account properties change.
     ///
+    /// This method must be called at most once (during client construction). Subsequent calls
+    /// are silently ignored because the callback slot is backed by a [`OnceLock`].
+    ///
     /// # Arguments
     /// * `callback` - The callback function to invoke with the refreshed account properties
     pub fn set_on_account_refresh_callback(&self, callback: OnAccountRefreshCallback) {
-        let mut guard = self.on_account_refresh.lock().unwrap();
-        *guard = Some(callback);
+        // OnceLock::set returns Err if already initialized; we intentionally ignore that
+        // because the callback is expected to be set exactly once.
+        let _ = self.on_account_refresh.set(callback);
     }
 
     /// Returns the default hub endpoint URL for the Cosmos DB account.
@@ -349,17 +356,11 @@ impl GlobalEndpointManager {
             )
             .await?;
 
-        // Invoke the registered callback if an HTTP call was made
+        // Invoke the registered callback if an HTTP call was made.
+        // `OnceLock::get` is lock-free, so this is safe to call from async code.
         let was_http_call_made = http_call_made.load(Ordering::SeqCst);
         if was_http_call_made {
-            // Clone the callback out of the mutex, then drop the lock before invoking it
-            // to avoid holding the mutex during arbitrary user code (prevents potential deadlocks).
-            let callback = {
-                let guard = self.on_account_refresh.lock().unwrap();
-                guard.as_ref().map(Arc::clone)
-            };
-
-            if let Some(callback) = callback {
+            if let Some(callback) = self.on_account_refresh.get() {
                 callback(&account_properties);
             }
         }


sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
```diff
@@ -1,8 +1,9 @@
 use async_lock::RwLock;
+use azure_core::time::Duration;
```
FabianMeiswinkel, line 2: @analogrelay would we only allow the azure_core type in public surface area? is there an From<std::time::Duration conversion already on the azure::core Duration?

FabianMeiswinkel: To me this also seems like a candidate for teh bare minimum core types crate - like Azure cores's Error ?

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
```diff
@@ -60,23 +92,45 @@ impl GlobalEndpointManager {
...
         );
 
-        Self {
+        let instance = Arc::new(Self {
             default_endpoint,
             location_cache,
             pipeline,
             account_properties_cache,
-        }
+            on_account_refresh: Mutex::new(None),
```
FabianMeiswinkel: the on_account_refresh Mutex seems weird - you take it and then do an async operation under the lock that is awaited. Usually Mutex requires the lock to be released at the same thread it was taken - and await often does not gauarntee continuation on the same thread. To me this looks like the worng synchronization conept - why is it not sufficient use something like an atomic reference and you only continue when your increment return 0 - so, only one thread would kick-off async task?

FabianMeiswinkel: or use an async semaphore (tokio/asyn std)?

FabianMeiswinkel (changes requested):
One question around synchronization choice which I think will need some more work/discussion

sdk/cosmos/azure_data_cosmos/src/routing/global_endpoint_manager.rs
kundadebdatta: I agree - `std::sync::Mutex` in async code is a recognized anti-pattern because it's fragile — a future refactor could accidentally hold it across `.await`. Since the callback is set exactly once during initialization and never changes, we can replace the `Mutex` with `std::sync::OnceLock`, which is completely lock-free for reads and makes the single-set intent explicit.

sdk/cosmos/azure_data_cosmos/src/routing/async_cache.rs
analogrelay: FYI, `azure_core::time::Duration` is just a re-export of `time::Duration`, from the third-party [`time`](https://docs.rs/time) crate, so that's the actual underlying type being used. Rust considers `::azure_core::time::Duration` and `::time::Duration` as two names for the same type.

I'm fine with having `azure_core::time::Duration` on the public surface area, and if we're doing that we should be consistent about it.

> is there an From<std::time::Duration conversion already on the azure::core Duration?

There are indeed [conversions to/from `std::time::Duration` defined on `time::Duration`](https://docs.rs/time/latest/time/struct.Duration.html#impl-Sum%3C%26Duration%3E-for-Duration), though they are fallible because `time::Duration` is allowed to be negative.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -268,15 +269,44 @@ impl CosmosClientBuilder {
...
+            GlobalPartitionEndpointManager::new(
+                global_endpoint_manager.clone(),
+                false,
+                enable_partition_level_circuit_breaker,
+            );
+
+        // Register the callback for account refresh to update partition-level failover config
+        let partition_manager_clone = Arc::clone(&global_partition_endpoint_manager);
+
+        global_endpoint_manager.set_on_account_refresh_callback(Arc::new(
```
analogrelay, line 298: To me, this indicates a high amount of coupling between these two types. It makes sense that there's some separation here in other SDKs given the history, but I think when we move to the driver we should be either decoupling these components OR just uniting them into one component.

analogrelay (approved):
I have concerns about the coupling of GEM and GPEM, but let's address those as we move things to the driver. Otherwise, nothing blocking.

PR #3832, opened 2026-02-27
[Storage] `set_metadata` properly generated, Queues repacking fully generated

vincenttran-msft:
This PR contains the work required to get the following two issues in the title addressed. 
`set_metadata` typespec changes: https://github.com/Azure/azure-rest-api-specs/pull/40868
queues typespec changes: https://github.com/Azure/azure-rest-api-specs/pull/40871

### DIFF
--- a/eng/emitter-package-lock.json
+++ b/eng/emitter-package-lock.json
@@ -11,7 +11,7 @@
         "@azure-tools/typespec-azure-core": "~0.65.0",
         "@azure-tools/typespec-azure-resource-manager": "~0.65.0",
         "@azure-tools/typespec-azure-rulesets": "~0.65.0",
-        "@azure-tools/typespec-client-generator-core": "~0.65.1",
+        "@azure-tools/typespec-client-generator-core": "~0.65.3",
         "@typespec/compiler": "^1.9.0",
         "@typespec/events": "~0.79.0",
         "@typespec/http": "^1.9.0",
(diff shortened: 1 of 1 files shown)


eng/emitter-package.json
```diff
@@ -18,4 +18,4 @@
     "@typespec/versioning": "~0.79.0",
     "@typespec/xml": "~0.79.0"
   }
```
heaths, line 20: These need to be in sync as well. I'll start a separate PR to update everything, as we normally do (pre-GA only).

sdk/storage/azure_storage_queue/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 - Changed `QueueClient::set_metadata()` `metadata` parameter from owned `HashMap<String, String>` to `&HashMap<String, String>`.
 - Support for `wasm32-unknown-unknown` has been removed ([#3377](https://github.com/Azure/azure-sdk-for-rust/issues/3377))
+- `SentMessage` ownership semantics changed, and code that previously moved fields like `message_id`/`pop_receipt` may require `clone()` (or borrowing) updates.
```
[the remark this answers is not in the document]
vincenttran-msft: We never released with those -- those were a temporary shoe-in.

jalauzon-msft, line 14: Why? This doesn't seem right, does it?

eng/emitter-package.json
heaths: The update PR is in if you want to merge and/or rebase on that before merging this. Probably would be fine even just merging this as-is as well.

sdk/storage/azure_storage_queue/CHANGELOG.md
vincenttran-msft: More context in the issue here:
https://github.com/Azure/typespec-rust/issues/868

`SentMessage` is now a newtype wrapping the internal generated struct to it prevents callers from depending on `SentMessageInternal` directly (the internal gen struct) and enables the `Deserialize` impl to do the `QueueMessagesList` -> `SentMessage` massage.

The side effect is that fields can no longer be *moved* out (ie. owned) since access is via `Deref` (a shared ref). Any code doing `let id = msg.message_id;` will need `msg.message_id.clone()` instead.

PR #3835, opened 2026-02-28
Add emitter update script and update to 0.36.0

heaths:
Adds a script to update the emitter (along with a Copilot skill) and optional regenerate clients.
Includes breaking changes to Key Vault keys, which was expected.


RickWinter (approved):

eng/emitter-package.json
```diff
@@ -1,13 +1,13 @@
...
   "dependencies": {
-    "@azure-tools/typespec-rust": "0.35.0"
+    "@azure-tools/typespec-rust": "0.36.0"
   },
   "devDependencies": {
     "@azure-tools/typespec-azure-core": "~0.65.0",
     "@azure-tools/typespec-azure-resource-manager": "~0.65.0",
     "@azure-tools/typespec-azure-rulesets": "~0.65.0",
-    "@azure-tools/typespec-client-generator-core": "~0.65.1",
+    "@azure-tools/typespec-client-generator-core": "~0.65.3",
```
jhendrixMSFT: Can we bump this to `0.65.4`, it's the latest version.

eng/emitter-package.json
```diff
@@ -1,13 +1,13 @@
...
     "@typespec/compiler": "^1.9.0",
     "@typespec/events": "~0.79.0",
     "@typespec/http": "^1.9.0",
```
jhendrixMSFT: Let's also bump this one to `1.9.1` to pick up the missing `content-type` header fix for storage.

sdk/keyvault/azure_security_keyvault_keys/src/generated/models/method_options.rs
```diff
@@ -25,9 +25,6 @@ pub struct KeyClientCreateKeyOptions<'a> {
 /// Options to be passed to [`KeyClient::decrypt()`](crate::generated::clients::KeyClient::decrypt())
 #[derive(Clone, Default, SafeDebug)]
 pub struct KeyClientDecryptOptions<'a> {
-    /// The version of the key.
-    pub key_version: Option<String>,
```
jhendrixMSFT, line 29: @heaths these are all the expected key version params to be promoted to required, right?

.github/skills/update-emitter/SKILL.md
```diff
@@ -0,0 +1,28 @@
+---
+name: update-emitter
```
LarryOsterman, line 2: This may need to move to `\skills` at some point.

sdk/keyvault/azure_security_keyvault_keys/src/generated/models/method_options.rs
heaths: For crypto operations, yes. @chlowell, @vcolin7, and I decided it was best. In all other languages, they are semi-optional parameters and while not idiomatic to pass `""` for a semi-optional parameter in Rust, the idea is that you *really, REALLY* should *NOT* use versionless key IDs for crypto operations or you could be locked out of your data.

.github/skills/update-emitter/SKILL.md
heaths: The thought here was that I'll support Copilot as a first-party offering. When they support `/skills` (which is probably after CC does), I'm happy to move it. I talked with several PMs and we agreed that, even while Copilot supports CC paths, having those paths and not Copilot paths in our Microsoft repos doesn't look great.

LarryOsterman: Makes sense, but I thought I saw that your other PR put the skills in `/skills`, which is why I mentioned it.

PR #3836, opened 2026-02-28
Consolidate agent instructions and improve skills

heaths:
Merge `.github/copilot-instructions.md` into `AGENTS.md` as the single source of truth for coding conventions, persona, prerequisites, and test generation guidelines. Restructure `AGENTS.md` to follow: project structure, agent capabilities, coding conventions, build, test, lint.

Migrate `.markdownlintrc` to `.markdownlint-cli2.yaml` with friendly rule names and schema validation. Add `lint-markdown` skill `package.json` for pinned markdownlint-cli2. Update `check-spelling` skill to use `Invoke-Cspell.ps1` directly. Create new `create-crate` skill for TypeSpec-based crate setup.

Update `CONTRIBUTING.md` and `cosmos.instructions.md` references. Add `node_modules/` to root `.gitignore`. Remove `compatibility` frontmatter from skill files.

### DIFF
--- a/.github/skills/create-crate/SKILL.md
+++ b/.github/skills/create-crate/SKILL.md
@@ -10,21 +10,22 @@ All new service crates must be generated from TypeSpec specifications in [Azure/
 1. **Install tsp-client**
    - Follow `eng/common/tsp-client/README.md` to install dependencies
    - Run `npm ci` from `eng/common/tsp-client/`
-   - Use `npx --prefix eng/common/tsp-client tsp-client <command>` to invoke it from the repository root
+   - Use `npm exec --prefix eng/common/tsp-client -- tsp-client <command>` to invoke it from the repository root
 
 2. **Find the TypeSpec spec**
    - Look for a `tspconfig.yaml` under `specification/<service>/` in [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs)
    - Check the `tspconfig.yaml` for `@azure-tools/typespec-rust` emitter configuration
+   - If no emitter configuration exists, stop and report an error indicating that the TypeSpec specification in [Azure/azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs) must be updated before generating the client
 
 3. **Initialize the crate**
-   - Run `tsp-client init --tsp-config <url>` from the repository root
+   - Run `npm exec --prefix eng/common/tsp-client -- tsp-client init --tsp-config <url>` from the repository root
    - `<url>` is the GitHub URL to the `tspconfig.yaml`
    - Always use a specific commit SHA in the URL — never a branch or tag, which can move
    - Example: `https://github.com/Azure/azure-rest-api-specs/blob/<commit-sha>/specification/<service>/<rp>/tspconfig.yaml`
 
 4. **Generate the client**
    - Use the `azsdk_package_generate_code` MCP tool, or
-   - Run `tsp-client update` from the crate directory
+   - Run `npm exec --prefix eng/common/tsp-client -- tsp-client update` from the crate directory
 
 5. **Add hand-written wrappers**
    - Create `clients.rs` (or similar) for custom client constructors, authentication setup, and convenience methods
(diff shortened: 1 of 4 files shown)


AGENTS.md
```diff
@@ -4,30 +4,18 @@ This document provides guidance for AI agents (e.g., GitHub Copilot, MCP servers
...
-
-Rust (MSRV: 1.85)
-
-### Key Technologies
-
-- Rust toolchain with Cargo
-- TypeSpec for API specification and code generation
-- OpenTelemetry for distributed tracing
-- Test Proxy for recorded integration tests
+- **Primary Language**: Rust (MSRV: 1.88)
```
LarryOsterman: You might want to remove the MSRV reference here. Copilot tends to treat the agents.md file as the canonical truth, ignoring artifacts like cargo.toml.

AGENTS.md
```diff
@@ -107,9 +98,66 @@ AI agents **should not**:
...
+
+### Dependencies
+
+- Dependencies should be defined in the root workspace's `Cargo.toml` file.
+- Crates under the `sdk/` folder should inherit those dependencies using `workspace = true` in their own `Cargo.toml` files.
+
+### General
+
+- Prioritize safety, efficiency, and correctness.
+- Respect Rust's ownership and borrowing rules.
```
LarryOsterman, line 159: Add something telling it to write idiomatic rust whenever possible - I've caught opus trying to create "from_value" and "to_value" methods instead of implementing the "try_from" or "from" trait.

.github/skills/create-crate/SKILL.md
```diff
@@ -0,0 +1,38 @@
...
+All new service crates must be generated from TypeSpec specifications in [Azure/azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs). Do not hand-write client libraries from scratch.
+
+1. **Install tsp-client**
+   - Follow `eng/common/tsp-client/README.md` to install dependencies
+   - Run `npm ci` from `eng/common/tsp-client/`
+   - Use `npx --prefix eng/common/tsp-client tsp-client <command>` to invoke it from the repository root
+
+2. **Find the TypeSpec spec**
+   - Look for a `tspconfig.yaml` under `specification/<service>/` in [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs)
+   - Check the `tspconfig.yaml` for `@azure-tools/typespec-rust` emitter configuration
```
LarryOsterman, line 17:
```suggestion
   - Check the `tspconfig.yaml` for `@azure-tools/typespec-rust` emitter configuration
   - If no emitter configuration exists, terminate the process with an error indicating that the customer needs to update the `Azure/azure-rest-api-specs` repository before starting the client generation
```

If you don't do that, copilot can get extremely confused and depending on the LLM used may go off the rails - better to short circuit the process early.

AGENTS.md
heaths: I wavered on this too. Yeah, let's get rid of the MSRV reference. Instead, maybe:

```suggestion
- **Primary Language**: Rust
- **Minimum supported Rust version (MSRV)**: found in the root [Cargo.toml] file
```

If it needs it for any reason, it knows how to find it.

AGENTS.md
heaths: We can add something like,

```markdown
- Write idiomatic Rust and well-documented Rust code that follows conventions in `std` like `from()` or `try_from()`.
- Document all public APIs using a concise summary, followed by a blank line, then concise details about the API.
- Use the `?` operator for calls that return a `Result` in a function that returns a `Result`, or for calls that return an `Option` in a function that returns `Option`.
```

These were a few things that just stood out when comparing with the original CP instructions. What you said made me think we had something like what you suggested before, but not quite so direct.

LarryOsterman: Makes sense. 

I was playing with similar skills on Friday while messing with the monitor log ingestion and attestation SDKs and have some stuff I'd like to try with your skills (opus came up with some specialized instructions w.r.t. testing for example, but I figure i can add that to your PR once it's in.

PR #3848, opened 2026-03-03
Driver-Transport-Spec - add crossbeam dependency

FabianMeiswinkel:


sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md
```diff
@@ -2472,6 +2594,218 @@ at the Runtime and Account layers only (not Operation), the effective walk is
...
+            Owned::new(new_state),
+            Ordering::AcqRel,
+            &guard,
+        );
+        unsafe { guard.defer_destroy(old); }
+    }
+}
+```
+
+Alternatively, `arc_swap::ArcSwap<T>` (which uses crossbeam epoch internally) provides a
```
analogrelay: > which uses crossbeam epoch internally

It doesn't... at least not currently. Probably hallucinated, though not really a problem for our situation since we're not using it.

FabianMeiswinkel: Fixed

PR #3860, opened 2026-03-05
Driver logging spec

FabianMeiswinkel:


sdk/cosmos/azure_data_cosmos_driver/docs/LOGGING_SPEC.md
```diff
@@ -0,0 +1,999 @@
...
+| ---------------------------- | ------------------------------------------------------------------------- | --- |
+| **Log Sink**                 | Callback interface for consumers to receive driver-internal logs          | 1   |
+| **Internal Instrumentation** | Structured log points inside the driver using `driver_log!`               | 2   |
+| **Sampling Diagnostics**     | Rate-limited operation-diagnostic entries (errors + threshold violations) | 3   |
+
+### Design Principles
+
+- **Lazy formatting**: The `driver_log!` macro checks the enabled level *before* calling `format!()`. No formatting cost when the level is disabled.
+- **Pre-formatted at sink boundary**: `LogSink::emit` receives `&[LogEntry]` with pre-formatted `String` messages. This simplifies FFI consumers that must copy data across language boundaries.
+- **Batched delivery**: A `LogBuffer` accumulates entries and flushes them in batches (configurable size, configurable flush interval) to reduce per-entry overhead. Auto-flush on drop is enabled by default.
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/docs/LOGGING_SPEC.md
analogrelay, line 1: My initial sense is that a lot of what's here is duplicating what `tracing` already provides us. I would have expected that we could achieve most of our logging needs using `tracing` and use a custom subscriber to do things like FFI callbacks and/or batching. We can still build our own sampling logic to control when we emit diagnostics to `tracing`. We can also use something like [`with_default`](https://docs.rs/tracing/latest/tracing/subscriber/fn.with_default.html) to override the default tracing subscriber with our own custom subscriber configuration if we need to perform custom actions to buffer/sample events.

FabianMeiswinkel: I considered that - but IMO that opens too much surface area - tracing is not just/primarily used for logging but to emit spans etc. I absolutely don't think we shoudl even try to forward spans to FFI - Open Telemetry spans shoudl be emitted in the SDK level - idiomatic to whatever the language of teh SDK uses. What I think we will need is some way of allowing to get additional info out of the driver when Diagnostics (our form of detailed Spans) is not sufficient. You coudl probably achieve this with instructions to disallow event! and span! macros in the driver and then build a subscriber that is only dealing with trace! - but I am not sure that this would really be simler - and certainly vaguer contract?

FabianMeiswinkel: Discussed offline - will close this PR and rework the spec to stick using `tracing` natievly in the driver - using a custom subscriber in azure_data_cosmos_native is a better approach.

PR #3861, opened 2026-03-05
Cosmos: Fixes Circuit Breaker Failover Logic for Multi-Master Writes on 403/3

kundadebdatta:
### Problem

On **multi-master** accounts, when a write request receives a `403/3` (WriteForbidden) response, the Per-Partition Circuit Breaker (PPCB) was not properly tracking failures or applying partition-level failover. This caused the partition override to never take effect, leaving the request stuck retrying via the generic account-level failover path without partition-level awareness.

### Root Cause

In `ClientRetryPolicy::should_retry_on_http_status()`, the 403/3 handler used a broad `partition_level_failover_enabled()` guard and then called `try_mark_endpoint_unavailable_for_partition_key_range()` directly — **without** going through the `ClientRetryPolicy`-level PPAF/PPCB eligibility checks:

```rust
if self
    .partition_key_range_location_cache
    .partition_level_failover_enabled()         // ← broad flag check only
    && self.request.is_some()
    && self
        .partition_key_range_location_cache
        .try_mark_endpoint_unavailable_for_partition_key_range(...)  // ← called directly
{
    return Some(RetryResult::Retry { after: Duration::ZERO });
}
// automatic failover support needed to be plugged in here.
return Some(self.should_retry_on_endpoint_failure(...).await);
```

This had two issues:

1. **PPCB failure counters were never incremented on 403/3.** The `try_mark_endpoint_unavailable_for_partition_key_range()` method moves the partition to the next region but does **not** call `increment_request_failure_counter_and_check_if_partition_can_failover()`. The counter is only incremented when `ClientRetryPolicy::is_request_eligible_for_partition_level_circuit_breaker()` is called — and it was bypassed on this code path.

2. **The PPCB threshold was bypassed on mark, but enforced on override application.** Even though `try_mark_endpoint_unavailable_for_partition_key_range()` would succeed (creating a failover map entry and moving the partition to the next region), the subsequent `try_add_partition_level_location_override()` call in `before_send_request()` checks `can_circuit_breaker_trigger_partition_failover()`, which verifies the failure count exceeds the threshold. Since the counter was never incremented, this check always returned `false`, so the **override was never applied** — the retry went back to the original (failed) endpoint.

**Net effect:** On multi-master accounts, 403/3 responses created phantom failover entries that were never applied, causing requests to repeatedly hit the failed endpoint.

### Fix

Replaced the broad `partition_level_failover_enabled()` check with the specific `ClientRetryPolicy`-level PPAF and PPCB eligibility checks:

```rust
if self.request.is_some()
    && (self.is_request_eligible_for_per_partition_automatic_failover()
        || self.is_request_eligible_for_partition_level_circuit_breaker())
    && self
        .partition_key_range_location_cache
        .try_mark_endpoint_unavailable_for_partition_key_range(...)
{
    return Some(RetryResult::Retry { after: Duration::ZERO });
}
return Some(self.should_retry_on_endpoint_failure(...).await);
```

This ensures:

- **`is_request_eligible_for_partition_level_circuit_breaker()`** (on `ClientRetryPolicy`) calls `increment_request_failure_counter_and_check_if_partition_can_failover()`, which **increments the PPCB failure counter** on every 403/3 and returns `true` only when the threshold is exceeded (default: 5 for writes).
- **Before threshold is reached:** PPCB returns `false`, `try_mark_endpoint_unavailable` is short-circuited via `&&`, and the request falls through to `should_retry_on_endpoint_failure()` (account-level retry) — correct behavior while the circuit breaker is counting.
- **After threshold is exceeded:** PPCB returns `true`, `try_mark_endpoint_unavailable` creates the failover entry and moves the partition to the next region. The override in `before_send_request()` is now also applied because `can_circuit_breaker_trigger_partition_failover()` sees a counter that exceeds the threshold.

Also removed the stale comment `// automatic failover support needed to be plugged in here.` since PPAF/PPCB eligibility checks are now properly integrated.

### Changes

- **`src/retry_policies/client_retry_policy.rs`** (+3 / -5)
  - `should_retry_on_http_status`: Replaced the broad `partition_level_failover_enabled()` guard with specific `is_request_eligible_for_per_partition_automatic_failover()` and `is_request_eligible_for_partition_level_circuit_breaker()` checks. Removed stale TODO comment.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
@@ -529,7 +529,7 @@ impl ClientRetryPolicy {
         {
             if self.request.is_some()
                 && (self.is_request_eligible_for_per_partition_automatic_failover()
-                    || self.is_request_eligible_for_partition_level_circuit_breaker())
+                    || self.increment_failure_counter_and_check_circuit_breaker_eligibility())
                 && self
                     .partition_key_range_location_cache
                     .try_mark_endpoint_unavailable_for_partition_key_range(
@@ -613,7 +613,7 @@ impl ClientRetryPolicy {
         if let Some(request) = self.request.as_ref() {
             if is_system_resource_unavailable_for_write
                 || self.is_request_eligible_for_per_partition_automatic_failover()
-                || self.is_request_eligible_for_partition_level_circuit_breaker()
+                || self.increment_failure_counter_and_check_circuit_breaker_eligibility()
             {
                 return self
                     .partition_key_range_location_cache
@@ -644,8 +644,8 @@ impl ClientRetryPolicy {
         false
     }
 
-    /// Checks if request is eligible for partition-level circuit breaker.
-    fn is_request_eligible_for_partition_level_circuit_breaker(&self) -> bool {
+    /// Increments failure counter and checks if request is eligible for partition-level circuit breaker.
+    fn increment_failure_counter_and_check_circuit_breaker_eligibility(&self) -> bool {
         if let Some(request) = self.request.as_ref() {
             return self
                 .partition_key_range_location_cache
@@ -739,12 +739,14 @@ impl ClientRetryPolicy {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::models::AccountRegion;
     use crate::operation_context::OperationType;
     use crate::partition_key::PartitionKey;
     use crate::regions;
     use crate::regions::RegionName;
     use crate::resource_context::{ResourceLink, ResourceType};
     use crate::routing::global_endpoint_manager::GlobalEndpointManager;
+    use crate::routing::partition_key_range::PartitionKeyRange;
     use azure_core::http::headers::Headers;
     use azure_core::http::ClientOptions;
     use azure_core::Bytes;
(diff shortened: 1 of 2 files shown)


sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -527,10 +527,9 @@ impl ClientRetryPolicy {
         if status_code == StatusCode::Forbidden
             && sub_status_code == Some(SubStatusCode::WRITE_FORBIDDEN)
         {
-            if self
```
tvaron3, line 530: Changelog entry as this is bugfix, but lgtm otherwise

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -527,10 +527,9 @@ impl ClientRetryPolicy {
...
-                .partition_key_range_location_cache
-                .partition_level_failover_enabled()
-                && self.request.is_some()
+            if self.request.is_some()
```
tvaron3, line 530: We should also have some unit tests or integration tests to prevent a future regression.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
```diff
@@ -527,10 +527,9 @@ impl ClientRetryPolicy {
...
+                && (self.is_request_eligible_for_per_partition_automatic_failover()
```
FabianMeiswinkel, line 531: NIT - I think part of why this was not foudn earlier is that this method has side-effects - IMO any method named with isXXXX shoudl never ever have side-effects - it should be idempotent. The fact that this has side effects (incrementing atomic counters is something the name should reveal).

kundadebdatta: Agreed - Renamed the method to `increment_failure_counter_and_check_circuit_breaker_eligibility` for clarity.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
kundadebdatta: Tests are added as a part of the latest commit.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
kundadebdatta: Updated change log.

Issue #3793, opened 2026-02-24
Cosmos Driver: Prime metadata caches at client creation time


analogrelay:
The Cosmos DB SDK really depends on having full metadata for a container in order to perform operations. Right now, that metadata is lazy-loaded, which means a lot of operations can take a longer amount of time because the cache needs to be populated on the first request. We normally refresh the cache in the background, so it's only this cache priming operation that takes longer.

In order to simplify the clients a bit, and make that behavior clearer to users, let's make `DatabaseClient::container_client` and `CosmosClient::database_client` async, and have them perform the cache priming IN the actual creation function.

We're already making these functions async, but this work item track doing the cache priming in those now-async functions.

FabianMeiswinkel:
That is what i had initially - but removed due to dead code :-)

I don't think we need anything put PK-def - but sure -we can extend for any of the immutabel properties (UniqueKey is also immutable for example)

The colelction delete+recreate enforcement is something I will take care of in the browser - we need to always capture the ecpected CollectionRid in a request header to achieve that. To eventually support renames we would need to add a similar rntbd header for expected id when using rid for addressing

PR #3864, opened 2026-03-05
[Cosmos] Driver: prime metadata caches

simorenoh:
Based on #3793, these changes prime the account metadata cache and the container cache when initializing a client instance or container instance - these are the only two caches we have currently in the driver. Changes are limited to the driver, where these caches will live moving forward.

Follow-up PR would be to create the partition key range cache for the driver and prime the partition key range cache for a container on the initial container call as well. With PPCB enabled by default, this ensures we're not making additional calls during data plane requests that will regardless require partition-scoped information per-request.


sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/account_metadata_cache.rs
```diff
@@ -205,19 +206,33 @@ impl AccountProperties {
...
+
 /// Cache for Cosmos DB account metadata.
 ///
 /// Stores account properties keyed by account endpoint.
 #[derive(Debug)]
 pub(crate) struct AccountMetadataCache {
     cache: AsyncCache<AccountEndpoint, AccountProperties>,
+
+    /// Tracks the last time each endpoint's metadata was refreshed.
+    last_refresh: async_lock::RwLock<std::collections::HashMap<AccountEndpoint, Instant>>,
```
analogrelay, line 226: Is there a reason this isn't just stored in the AsyncCache entry itself?

sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/account_metadata_cache.rs
```diff
@@ -243,10 +258,94 @@ impl AccountMetadataCache {
         let properties = fetch_fn().await?;
 
         // Cache the successfully fetched properties.
-        Ok(self
+        let result = self
+            .cache
+            .get_or_insert_with(endpoint.clone(), || async { properties })
+            .await;
+
+        // Record the fetch time after caching, so a concurrent loser
```
analogrelay: 😆 Maybe we should say "concurrent thread that loses the race" instead of "concurrent loser". Though it **is** funny

sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/account_metadata_cache.rs
simorenoh: @analogrelay I went with a separate map to avoid changing the AsyncCache's current interface since it didn't have any timestamped details for values - we could try something like this for the entry if not I guess? and then make sure we take out the account properties from the TimestampedAccountProperties for the rest of the logic within the SDK?
```
  struct TimestampedAccountProperties {
       properties: AccountProperties,
       fetched_at: Instant,
   }
```
Just wasn't sure how far we wanted to go to accommodate this behavior

analogrelay: Let's keep it separate for now, but maybe we should do a little rework of AsyncCache to support things like detecting stale and/or expired entries.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/account_metadata_cache.rs
```diff
@@ -243,10 +258,95 @@ impl AccountMetadataCache {
...
+                None => true, // Never fetched
+            }
+        };
+
+        if !is_stale {
+            // Not stale — return the current cached value.
+            return Ok(cached);
+        }
+
+        // Fetch outside the cache primitive so we can handle errors.
```
xinlian12: **🟡 Recommendation** · Design: Single-Pending-I/O Violation

**Concurrent `refresh_if_stale` calls will cause duplicate network fetches**

The method calls `fetch_fn()` *outside* `get_or_refresh_with()`, then passes the already-fetched result as the factory closure (`|| async { properties }`). Two concurrent `refresh_if_stale` calls will both hit the network before either reaches the cache primitive.

`get_or_refresh_with` was specifically designed so that concurrent requests for the same key share one I/O operation (via `Arc::ptr_eq` detection + single factory execution). By pre-fetching, this method bypasses that guarantee.

**Trace**: Thread A checks staleness → stale. Thread B checks staleness → stale. Both call `fetch_fn().await` (two network calls). Thread A wins `get_or_refresh_with`, Thread B's fetch result is discarded. Safe but wasteful.

**Why it matters**: Under load (e.g., many threads detecting staleness during a regional failover), this could produce N redundant metadata fetches instead of 1.

**Suggestion**: Move the fetch inside `get_or_refresh_with`'s factory closure. The error-handling concern (returning stale data on failure) could be addressed by wrapping the factory result or by enhancing `AsyncCache` to support `Result`-returning factories. Since this is dead code, this can be addressed before the consumer is added.

---

sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/account_metadata_cache.rs
```diff
@@ -457,4 +557,60 @@ mod tests {
...
+            .unwrap();
+
+        // Should return the cached value without calling the factory
+        assert!(result.is_some());
+        assert_eq!(result.unwrap().write_region().unwrap().as_str(), "westus");
+        assert_eq!(counter.load(Ordering::SeqCst), 0);
+    }
+
+    #[tokio::test]
+    async fn refresh_if_stale_refreshes_when_threshold_exceeded() {
```
xinlian12, line 618: **🟡 Recommendation** · Testing: Missing Error Path Coverage

**No tests for the graceful degradation path (fetch failure → return stale cached value)**

The two existing tests cover: (a) fresh entry returns cached, (b) stale entry refreshes. Missing:

1. **Stale + fetch failure + cached value exists** → should return the existing cached value (the "prefer stale data over no data" guarantee)
2. **Stale + fetch failure + no cached value** → should propagate the error

These are the most critical behavioral guarantees of `refresh_if_stale` — the error handling is the whole reason the method fetches outside the cache primitive.

**Suggestion**: Add two tests:
```rust
#[tokio::test]
async fn refresh_if_stale_returns_cached_on_fetch_error() {
    let cache = AccountMetadataCache {
        cache: AsyncCache::new(),
        last_refresh: async_lock::RwLock::new(std::collections::HashMap::new()),
        staleness_threshold: Duration::from_secs(0),
    };
    let endpoint = test_endpoint("myaccount");
    // Populate cache
    cache.get_or_fetch(endpoint.clone(), || async { Ok(test_properties("westus")) }).await.unwrap();
    // Stale refresh fails → should return cached "westus"
    let result = cache.refresh_if_stale(endpoint, || async {
        Err(azure_core::Error::message(azure_core::error::ErrorKind::Other, "network failure"))
    }).await.unwrap();
    assert_eq!(result.unwrap().write_region().unwrap().as_str(), "westus");
}
```

---

sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/account_metadata_cache.rs
```diff
@@ -243,10 +258,95 @@ impl AccountMetadataCache {
...
+
+        let endpoint_for_timestamp = endpoint.clone();
+
+        let result = self
             .cache
-            .get_or_insert_with(endpoint, || async { properties })
-            .await)
+            .get_or_refresh_with(
+                endpoint,
+                |_existing| true, // We already determined staleness above
```
xinlian12, line 345: **🟡 Recommendation** · Design: TOCTOU in Staleness Check

**Staleness is pre-computed outside the lock, then `|_existing| true` unconditionally forces refresh**

`get_or_refresh_with` accepts a `should_force_refresh` predicate designed to decide *inside the locked section* whether to refresh. Passing `|_existing| true` bypasses this — staleness is determined entirely outside, before acquiring any lock.

This creates a TOCTOU window: between the staleness check and the `get_or_refresh_with` call, another thread may have already refreshed. Combined with finding #1 (fetch outside the primitive), this means two concurrent threads both detect staleness, both fetch, and only one wins.

**Why it matters**: Not a correctness bug (`Arc::ptr_eq` in `get_or_refresh_with` handles the concurrent update correctly), but it means two levels of staleness checking that don't compose cleanly. If the staleness logic moved *into* the predicate (checking `last_refresh` there), both the TOCTOU and the duplicate-fetch issues would be resolved together.

**Suggestion**: When `refresh_if_stale` is wired up, consider passing the staleness logic into the predicate:
```rust
|_existing| {
    let timestamps = self.last_refresh.blocking_read(); // or restructure
    // ... check staleness here
}
```
This would close the TOCTOU window and allow the fetch to happen inside the factory, restoring single-pending-I/O.

---

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -315,6 +315,41 @@ impl CosmosDriver {
...
 
+    /// Eagerly primes the account metadata cache.
+    ///
+    /// Fetches account properties from the service and caches them so that
+    /// regional endpoint information is available before the first
+    /// [`execute_operation`](Self::execute_operation) call. This avoids
+    /// cold-start latency on the first data-plane operation.
+    ///
+    /// Returns an error if the account is unreachable.
+    pub async fn initialize(&self) -> azure_core::Result<()> {
```
xinlian12, line 313: **🟢 Suggestion** · Testing: Missing Public API Tests

**`initialize()` and `prime_container()` are public API but have no tests**

Per the changelog, these are the user-facing feature of this PR. Neither has unit or integration tests. While `initialize()` is a thin wrapper around `get_or_fetch` (which is tested), and `prime_container()` delegates to `resolve_container_by_name()`, the public API contract should have its own tests.

This is especially important because `initialize()` is auto-called in `get_or_create_driver()` — testing the integration between creation and priming would catch regressions if the auto-prime is accidentally removed.

**Suggestion**: Add tests verifying: (a) `initialize()` populates the account metadata cache, (b) `prime_container()` populates the container cache, (c) calling `initialize()` twice is idempotent. These may require test infrastructure for mocking the transport layer.

---

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -315,6 +315,41 @@ impl CosmosDriver {
         &self.options
     }
 
+    /// Eagerly primes the account metadata cache.
```
xinlian12, line 300: **🟢 Suggestion** · Documentation: Missing Auto-Call Note

**Doc comment doesn't mention `initialize()` is automatically called during driver creation**

An external consumer seeing `pub async fn initialize()` might call it manually, not knowing `get_or_create_driver()` already invokes it. While calling it twice is safe (idempotent via `get_or_insert_with`), the doc should set expectations.

**Suggestion**: Add a note to the doc comment:
```rust
/// # Note
///
/// This method is automatically called during driver creation via
/// [`CosmosDriverRuntime::get_or_create_driver()`]. Manual calls are safe
/// (idempotent) but typically unnecessary unless the automatic initialization
/// failed — check logs for a warning about cache priming failure.
```

---

sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/account_metadata_cache.rs
```diff
@@ -205,19 +206,33 @@ impl AccountProperties {
     }
 }
 
+/// Default minimum interval between metadata refreshes (10 minutes).
+///
+/// Matches the SDK's `GlobalEndpointManager` TTL and background refresh interval.
+const DEFAULT_STALENESS_THRESHOLD: Duration = Duration::from_secs(600);
```
xinlian12, line 216: **💬 Observation** · Design: Hardcoded Staleness Threshold

The 10-minute threshold is a `const`, not configurable. The doc comment claims it "matches the SDK's `GlobalEndpointManager` TTL and background refresh interval." For reference, the Java SDK uses 5 minutes for background refresh and a separate force-refresh interval. The .NET SDK (`GlobalEndpointManager`) also uses 5 minutes as the default `backgroundRefreshLocationTimeIntervalInMS`.

Not blocking — this is dead code and the value can be adjusted when wired up. Worth verifying the 10-minute figure against the specific cross-SDK behavior being targeted when the consumer PR lands.

---

simorenoh: This is the whole point of the default, it's just a static variable - here we are matching the value of the Rust SDK's existing GEM, not the other SDKs

sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/account_metadata_cache.rs
simorenoh: added these

sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/account_metadata_cache.rs
simorenoh: addressed with a double checked lock

sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/account_metadata_cache.rs
simorenoh: addressed with double checked lock as well

Issue #3846, opened 2026-03-03
Health checks for gateway endpoints


tvaron3:
## Summary

Add health checking logic for gateway endpoints. The solution should be easy to switch the API for an upcoming gateway API that gives health information. For now it can use the get database account success to gauge gateway availability.

## Context

Originally tracked in: https://github.com/Azure/azure-sdk-for-python/issues/45341

The Python SDK has an implementation for this pattern.

PR #3866, opened 2026-03-05
[Cosmos] Health checks spec for gateway endpoints

tvaron3:
## Summary

Spec for proactive gateway endpoint health checking in `azure_data_cosmos_driver`. Addresses #3846.

## Design

The health check system probes regional gateway endpoints in the background to detect unavailability and recovery, feeding results into the endpoint routing system so requests avoid unhealthy regions.

### Key components

- **`HealthProbe` trait** — swappable probe abstraction. Current implementation uses `GetDatabaseAccount` (lightweight metadata read). When the dedicated gateway health API ships, only the probe implementation changes — orchestration, scheduling, state management, and config remain untouched.
- **`GetDatabaseAccountProbe`** — reuses the existing `CosmosTransport` metadata pipeline for authenticated `GET /` calls with per-probe timeout.
- **`EndpointHealthMonitor`** — background `tokio` task that sweeps all regional endpoints concurrently every 5 minutes (configurable). Each probe has exponential backoff retry (3 attempts, 500ms initial delay).
- **Endpoint selection** — probes all read endpoints + first write endpoint (deduplicated), matching the Python SDK pattern.
- **State integration** — writes to the same `AccountEndpointStateStore` from the Transport Pipeline Spec (§4.4). Healthy probes remove endpoints from the unavailable set; failed probes add them. No changes needed in `execute_operation` — routing automatically respects the updated state.
- **Startup probing** — optional blocking sweep during driver init to verify endpoints before accepting requests.
- **`HealthCheckOptions`** — public configuration with env var overrides (`AZURE_COSMOS_HEALTH_CHECK_*`).

### Python SDK alignment

Defaults match the Python SDK: 5-minute refresh interval, 3 probe retries with 500ms initial backoff, concurrent probing of read + write endpoints. The partition-level circuit breaker is out of scope (covered separately in Transport Pipeline Spec §4.5).

### Open questions

5 questions flagged for discussion in the spec (startup blocking vs background, emulator behavior, metadata refresh piggybacking, gateway API contract, options visibility).

## Files changed

- `sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md` — full spec with component design, integration points, configuration, testing strategy, and 10-step migration plan

### DIFF
--- a/sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
+++ b/sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
@@ -67,29 +67,29 @@ request-triggered refresh would leave long gaps without health information.
 ## 2. Architecture Overview
 
 ```text
-┌─────────────────────────────────────────────────────────────┐
-│                    CosmosDriverRuntime                       │
-│                                                             │
-│  ┌─────────────────┐  ┌──────────────────────────────────┐  │
-│  │ AccountMetadata  │  │ EndpointHealthMonitor            │  │
-│  │ Cache            │  │                                  │  │
-│  │ (account props)  │  │  • owns background task          │  │
-│  └─────────────────┘  │  • schedules probes              │  │
-│                        │  • emits LocationEffects          │  │
-│                        │                                  │  │
-│                        │  ┌────────────────────────────┐  │  │
-│                        │  │ HealthProbe (enum)          │  │  │
-│                        │  │  • GetAccountMetadata       │  │  │
-│                        │  │  • (future: HealthEndpoint) │  │  │
-│                        │  └────────────────────────────┘  │  │
-│                        └──────────────────────────────────┘  │
-│                                     │                        │
-│                                     ▼                        │
-│                        ┌──────────────────────────────────┐  │
-│                        │ LocationStateStore                │  │
-│                        │   apply(LocationEffect)           │  │
-│                        └──────────────────────────────────┘  │
-└─────────────────────────────────────────────────────────────┘
++---------------------------------------------------------------+
+|                      CosmosDriverRuntime                      |
+|                                                               |
+|  +-------------------+ +--------------------------------+     |
+|  | AccountMetadata   | | EndpointHealthMonitor          |     |
+|  | Cache             | |                                |     |
+|  | (account props)   | |  * owns background task        |     |
+|  +-------------------+ |  * schedules probes            |     |
+|                        |  * emits LocationEffects       |     |
+|                        |                                |     |
+|                        |  +--------------------------+  |     |
+|                        |  | HealthProbe (enum)       |  |     |
+|                        |  |  * GetAccountMetadata    |  |     |
+|                        |  |  * (future variant)      |  |     |
+|                        |  +--------------------------+  |     |
+|                        +--------------------------------+     |
+|                                         |                     |
+|                                         v                     |
+|                        +--------------------------------+     |
+|                        | LocationStateStore             |     |
+|                        |   apply(LocationEffect)        |     |
+|                        +--------------------------------+     |
++---------------------------------------------------------------+
 ```
 
 ### Relationship to Transport Pipeline Spec


sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,861 @@
...
+pub(crate) enum ProbeResult {
+    /// The endpoint responded successfully.
+    Healthy {
+        /// How long the probe took (for diagnostics/logging).
+        latency: Duration,
+    },
+    /// The endpoint did not respond or returned an error.
+    Unhealthy {
+        /// Diagnostic reason for the failure.
+        reason: String,
```
tvaron3: Will this hook up to the diagnostic context described in the other specs with status codes etc.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,861 @@
...
+        );
+        let mut context = Context::default();
+        context.insert(AuthorizationContext::new(
+            azure_core::http::Method::Get,
+            ResourceType::DatabaseAccount,
+            "",
+        ));
+
+        // Apply probe-specific timeout.
+        let result = tokio::time::timeout(
```
tvaron3: These should change the timeouts on the http layer dynamically

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,861 @@
...
+    /// Default: 180s (3 minutes).
+    pub max_retry_delay: Duration,
+
+    /// Timeout for individual probe requests.
+    /// Default: 5s.
+    pub probe_timeout: Duration,
+
+    /// Whether to run health checks on startup.
+    /// Default: true.
+    pub probe_on_startup: bool,
```
tvaron3: This should always happen and does not need a variable

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,861 @@
...
+/// Returns the final `ProbeResult` after exhausting retries or
+/// getting a healthy response.
+async fn probe_with_retry(
+    probe: &dyn HealthProbe,
+    endpoint: &CosmosEndpoint,
+    config: &HealthMonitorConfig,
+) -> ProbeResult {
+    let mut delay = config.initial_retry_delay;
+
+    for attempt in 0..=config.max_probe_retries {
```
tvaron3: This should be part of the new client retry policy described in the transport.spec

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,861 @@
...
+
+This means:
+- Reactive failures are detected instantly (no waiting for next sweep).
+- Proactive sweeps detect recovery (no waiting for expiration timer).
+- Both write atomically via `swap()` on the same store.
+
+### 4.4 Diagnostics Integration
+
+Health probe activity is recorded in the `DiagnosticsContext` system. The existing
+`ExecutionContext::CircuitBreakerProbe` variant (already defined in the Transport Pipeline
```
tvaron3: It should have its own variant and as ppcb is a separate feature

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,861 @@
...
+Following the Python SDK pattern, key thresholds can be overridden via environment
+variables for debugging and tuning without code changes:
+
+| Env Var | Field | Default |
+|---------|-------|---------|
+| `AZURE_COSMOS_HEALTH_CHECK_INTERVAL_SECS` | `refresh_interval` | 300 |
+| `AZURE_COSMOS_HEALTH_CHECK_MAX_RETRIES` | `max_probe_retries` | 3 |
+| `AZURE_COSMOS_HEALTH_CHECK_RETRY_AFTER_MS` | `initial_retry_delay` | 500 |
+| `AZURE_COSMOS_HEALTH_CHECK_PROBE_TIMEOUT_SECS` | `probe_timeout` | 5 |
+| `AZURE_COSMOS_HEALTH_CHECK_ENABLED` | `enabled` | true |
```
tvaron3: Remove the health check enabled.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,861 @@
...
+```
+
+### 5.3 Environment Variable Overrides
+
+Following the Python SDK pattern, key thresholds can be overridden via environment
+variables for debugging and tuning without code changes:
+
+| Env Var | Field | Default |
+|---------|-------|---------|
+| `AZURE_COSMOS_HEALTH_CHECK_INTERVAL_SECS` | `refresh_interval` | 300 |
```
tvaron3: Get rid actually of all the env variables

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,861 @@
...
+  `AccountEndpointStateStore`, `CosmosEndpoint`, `UnavailableReason`).
+- **Enables**: Faster failover and recovery in Steps 3+ (circuit breaker, hedging).
+
+---
+
+## 11. Open Questions
+
+| # | Question | Status |
+|---|----------|--------|
+| 1 | Should the startup probe block driver creation or run in the background? The Python SDK runs it in a background daemon. Blocking gives better first-request guarantees but adds latency to client construction. | **Proposed: blocking** (matches user expectation that client is ready after construction) |
```
tvaron3: It should not block client creation

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,861 @@
...
+| 2 | Should health checks be disabled for emulator mode? The emulator is single-region and always localhost. | **Proposed: yes** (skip when connection string is `"emulator"`) |
```
tvaron3: No should be fine to keep with emulator but true that does not give much value.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,861 @@
...
+| 3 | Should the health monitor also trigger an `AccountMetadataCache` refresh on sweep? The Python SDK re-fetches account properties alongside health checks. | **Proposed: yes** (piggyback on the `GetDatabaseAccount` response to refresh topology) |
```
tvaron3: I think better to decouple this but leave this open to discussion.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+        );
+        let transport_request = TransportRequest {
+            method: Method::Get,
+            url: endpoint.url().clone(),
+            region: None,
+            headers: Headers::new(),
+            body: None,
+            auth_context,
+            execution_context: ExecutionContext::HealthCheckProbe,
+            deadline: None, // Per-request timeout deferred; uses client-level timeout.
```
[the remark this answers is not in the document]
tvaron3: This will be part of these changes https://github.com/Azure/azure-sdk-for-rust/pull/3871

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+  to fail before marking an endpoint unavailable.
+- **Stale availability state**: Once an endpoint is marked unavailable, the driver has no
+  way to detect when it recovers except via expiration timers.
+
+Health checks address all three by probing regional endpoints in the background and updating
+the `AccountEndpointState` (§4.4 of the Transport Pipeline Spec).
+
+### Design Constraint: Swappable Probe API
+
+The Cosmos DB gateway team is building a dedicated health endpoint. Until that ships, we use
```
analogrelay: Is it possible for us to detect when this endpoint is available? Ideally, we'd like to just start using it when it becomes available instead of requiring users to update their SDK

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+make it trivial to swap the underlying probe API without changing the health check
+orchestration, scheduling, or state management.
+
+### Non-Goals
+
+- **Partition-level health / circuit breaker**: Covered separately in Transport Pipeline
+  Spec §4.5. Health checks here are **endpoint-level** (regional gateway availability).
+- **Connection-pool shard health**: Covered in Transport Pipeline Spec §6.5–6.6 (per-shard
+  eviction). That is a transport-layer concern; this spec is about service-level reachability.
+- **User-facing health API**: Health checks are internal. No public API is exposed.
```
analogrelay, line 51: Some telemetry/logging of health check results and outcome (proactively marking endpoints up/down) would be good. We shouldn't be spamming logs with "Health check OK" messages though :).

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+/// Trait for probing a single Cosmos DB gateway endpoint.
+///
+/// Implementations must be `Send + Sync` (shared across the background task
+/// and potentially multiple concurrent probes).
+///
+/// The trait is intentionally minimal: one method, one endpoint, one result.
+/// The orchestration layer handles scheduling, concurrency, retries, and
+/// state updates.
+#[async_trait]
+pub(crate) trait HealthProbe: Send + Sync {
```
analogrelay: I wouldn't use a trait for this. Rather I'd just use an enum: `HealthProbe::GetAccountMetadata` and `HealthProbe::HealthEndpoint(...)` (where the latter stores the details about the endpoint to use. This isn't something a user can provide an implementation for. See [the guidance on dyn trait in the DOP principles](https://analogrelay.github.io/dop-in-rust/dyn_trait/01-dyn-trait.html#when-not-to-use-dyn-trait-continued)

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+                        message: format!("HTTP {}", status.status_code()),
+                    },
+                    latency,
+                }
+            }
+            TransportOutcome::TransportError { error, .. } => {
+                ProbeResult::Unhealthy {
+                    failure: ProbeFailure {
+                        status: None,
+                        message: format!("transport error: {error}"),
```
analogrelay: I'd prefer we carry the error itself in the failure, if we can, and defer formatting anything into a string until we need it as a string.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+/// Spawned by `CosmosDriver` during initialization. Runs a periodic
+/// sweep that probes all relevant endpoints and updates the
+/// `AccountEndpointStateStore`.
+pub(crate) struct EndpointHealthMonitor {
+    config: HealthMonitorConfig,
+    probe: Arc<dyn HealthProbe>,
+    endpoint_state_store: Arc<AccountEndpointStateStore>,
+    account_metadata_cache: Arc<AccountMetadataCache>,
+    /// Handle to the background sweep task. Dropped on shutdown.
+    _sweep_handle: JoinHandle<()>,
```
analogrelay: Dropping a JoinHandle doesn't abort the task. We should probably bring over some of the background task manager we built in the SDK.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+3. **Nothing else changes.** The monitor, sweep loop, state management,
+   configuration, and diagnostics all remain identical.
+
+This is the key design benefit of the `HealthProbe` trait abstraction.
+
+---
+
+## 8. Shutdown & Resource Cleanup
+
+The `EndpointHealthMonitor` uses a `CancellationToken` to signal the background task to
```
analogrelay: I don't think we need a cancellation token. Rust has its own cancellation model which signals cancellation by dropping the future. If we hold the future in the health monitor itself, dropping the monitor will drop the future, and thus cancel it.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+            &*probe,
+            &endpoint_state_store,
+            &account_metadata_cache,
+        )
+        .await;
+
+        loop {
+            tokio::select! {
+                _ = tokio::time::sleep(config.refresh_interval) => {},
+                _ = shutdown.cancelled() => break,
```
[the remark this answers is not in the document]
analogrelay: We've got some other use of explicit tokio pieces in the driver, we'll need to assess all of that in a single pass, I think.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+                let ep = ep.clone();
+                let probe = probe; // borrow, not move
+                async move {
+                    let result = probe.probe(&ep).await;
+                    (ep, result)
+                }
+            })
+            .collect();
+
+        let results = futures::future::join_all(probe_futures).await;
```
analogrelay: Using a `JoinSet` and then looping with `while let Some(...) = probes.join_next()` would allow us to immediately start processing a given probe's result as it arrives rather than having to wait for all of them.

analogrelay: Though, that couples us to tokio again. But I think we need to take a step back and think about async runtime abstraction more broadly, since we need to do things like this, and using `select!` macros, etc. that aren't standardized in Rust and depend directly on the runtime.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+                        status_code = ?failure.status.as_ref().map(|s| s.status_code()),
+                        sub_status_code = ?failure.status.as_ref().and_then(|s| s.sub_status()),
+                        "health probe failed, marking endpoint unavailable"
+                    );
+                }
+            }
+        }
+
+        endpoint_state_store.swap(new_state);
+    }
```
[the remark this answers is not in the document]
analogrelay: Yep, I think we need to collect a set of `LocationEffect`s and apply them atomically instead.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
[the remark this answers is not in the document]
tvaron3: Removed `probe_timeout` from config and options — per-request timeout deferred to PR #3871.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+// Probe retry is a simple loop in run_sweep():
+for attempt in 0..=config.max_probe_retries {
+    let result = probe.probe(&ep).await;
+    match &result {
+        ProbeResult::Healthy { .. } => return result,
+        ProbeResult::Unhealthy { .. } if attempt < config.max_probe_retries => {
+            tokio::time::sleep(backoff_delay(attempt)).await;
+        }
+        _ => return result,
+    }
```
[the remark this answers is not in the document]
tvaron3: Renamed to `max_probe_attempts` with exclusive range (`0..max_probe_attempts`) so 3 means 3 total attempts.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+    // ... existing fields ...
+
+    /// Options for background endpoint health checking.
+    pub health_check: HealthCheckOptions,
+}
+```
+
+Configuration is programmatic-only via `HealthCheckOptions`. There are no environment
+variable overrides — all tuning is done in code.
+
```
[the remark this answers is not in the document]
tvaron3: Updated PR description to remove env var references — spec is the source of truth.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+}
+```
+
+### 3.7 Startup Probing
+
+On `CosmosDriver` initialization, after the initial `GetDatabaseAccount` succeeds and
+populates the `AccountMetadataCache`, the health monitor spawns the background sweep loop
+which immediately runs the first sweep. The driver does **not** block on the initial sweep
+— the client is usable immediately after construction. The reactive retry/failover path
+covers the window before the first sweep completes.
```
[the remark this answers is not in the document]
tvaron3: Updated PR description to say non-blocking startup, matching the spec.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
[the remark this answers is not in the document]
tvaron3: Replaced tokio APIs with `azure_core::task::spawn/sleep` and `futures::future::select` for runtime-agnostic design.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+            shutdown.clone(),
+        ));
+
+        Self {
+            config,
+            probe,
+            endpoint_state_store,
+            account_metadata_cache,
+            _sweep_handle: sweep_handle,
+        }
```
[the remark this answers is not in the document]
tvaron3: Added `shutdown: CancellationToken` field to struct and implemented `Drop` to cancel it.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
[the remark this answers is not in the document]
tvaron3: Replaced snapshot+swap with per-endpoint `mark_available`/`mark_unavailable` calls to avoid lost concurrent updates.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+
+```rust
+// driver/routing/account_endpoint_state.rs (new in Step 2)
+
+/// Produce a new state with an endpoint marked as available (removed from
+/// the unavailable set).
+///
+/// If the endpoint is not in the unavailable set, the state is returned
+/// unchanged.
+fn mark_endpoint_available(
```
analogrelay: We probably want to structure this as a `LocationEffect` and apply it using those APIs

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+Each probe result is recorded as a diagnostic event containing:
+- **Endpoint URI**: The regional gateway endpoint that was probed.
+- **HTTP status code**: When the probe received an HTTP response.
+- **Sub-status code**: When present in response headers (`x-ms-substatus`).
+- **Latency**: How long the probe took.
+- **Outcome**: Healthy or Unhealthy with the `ProbeFailure` details.
+
+```rust
+// When logging probe attempts:
+tracing::debug!(
```
analogrelay: I'd argue that we should log successful probes as `trace!` or even not log them at all. That's a lot of unnecessary noise. Maybe, if we can track it, the first successful probe after failure could be logged, but an "everything is OK" alarm gets ignored easily :)

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+    health.rs                         # HealthCheckOptions (public)
+```
+
+---
+
+## 7. Swapping to Gateway Health API
+
+When the dedicated gateway health endpoint ships, the migration is:
+
+1. Create `GatewayHealthProbe` implementing `HealthProbe`:
```
analogrelay: It would be ideal if the gateway health probe endpoints could be returned by Get Account Properties or something similar. That way we can select an appropriate strategy during driver initialization

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,853 @@
...
+
+### 9.2 Integration Tests
+
+| Test | What it validates |
+|------|-------------------|
+| `startup_probe_populates_state` | After driver init, unavailable endpoints reflect actual reachability |
+| `background_sweep_detects_recovery` | Mark endpoint unavailable → probe succeeds → endpoint marked available |
+| `health_check_with_fault_injection` | Inject transport error → probe fails → endpoint unavailable |
+
+### 9.3 Mock Probe for Testing
```
analogrelay: I'd rather we test this using FaultInjection, if possible.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+        // Probe all endpoints concurrently.
+        // 429 throttle retries are handled by execute_transport_pipeline().
+        // Probe-level retries are handled by the sweep (see §3.5).
+        let probe_futures: Vec<_> = endpoints
+            .iter()
+            .map(|ep| {
+                let ep = ep.clone();
+                let probe = probe; // borrow, not move
+                async move {
+                    let result = probe.probe(&ep).await;
```
FabianMeiswinkel: **🔴 Blocking — Retry logic (§3.5) not integrated into the concurrent probing code (§3.6)**

§3.5 defines the 3-attempt probe-level retry loop with exponential backoff, but the concurrent probing code in §3.6 calls `probe.probe(&ep)` **once** per endpoint — the retry loop from §3.5 is absent:

```rust
async move {
    let result = probe.probe(&ep).await;  // single call, no retry
    (ep, result)
}
```

An implementer reading §3.6 would build a zero-retry sweep. The retry must either be inlined into each concurrent future or factored into a `probe_with_retry` helper called within the `map`.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+            shutdown,
+            _sweep_handle: sweep_handle,
+        }
+    }
+}
+```
+
+### 3.8 `mark_endpoint_available` System Function
+
+New system function for endpoint recovery, part of the Step 2 endpoint state system
```
FabianMeiswinkel: **🟡 Recommendation — Cross-Spec Gap: `mark_endpoint_available` is not defined in the Transport Pipeline Spec**

This introduces `mark_endpoint_available` as a "new system function for endpoint recovery, part of the Step 2 endpoint state system." But the Transport Pipeline Spec §4.4 only defines three system functions: `build_account_endpoint_state`, `mark_endpoint_unavailable`, and `expire_unavailable_endpoints`. The Step 2 migration plan (§8, sub-step 2.1) lists "system functions" but doesn't include `mark_endpoint_available`.

This spec should either:
- a) Explicitly call out that it **adds** a new system function to the Step 2 scope and cross-reference an update to the Transport Pipeline Spec, or
- b) Be accompanied by a matching Transport Pipeline Spec amendment.

Without this, an implementer working from the Transport Pipeline Spec alone will miss this function.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+/// The trait is intentionally minimal: one method, one endpoint, one result.
+/// The orchestration layer handles scheduling, concurrency, retries, and
+/// state updates.
+#[async_trait]
+pub(crate) trait HealthProbe: Send + Sync {
+    /// Probe the given endpoint and return whether it is healthy.
+    ///
+    /// Implementations should use a short timeout (e.g. 5s) to avoid blocking
+    /// the health sweep for too long.
+    async fn probe(&self, endpoint: &AccountEndpoint) -> ProbeResult;
```
FabianMeiswinkel: **🟡 Recommendation — Type Naming Mismatch: `AccountEndpoint` vs `CosmosEndpoint`**

This spec uses `AccountEndpoint` throughout (e.g., `HealthProbe::probe(&self, endpoint: &AccountEndpoint)`, `endpoints_to_probe(state: &AccountEndpointState) -> Vec<AccountEndpoint>`). But the Transport Pipeline Spec §4.4 defines the canonical endpoint type as `CosmosEndpoint`:

```rust
pub(crate) struct CosmosEndpoint {
    region: RegionName,
    url: Url,
    kind: EndpointKind,
}
```

And `AccountEndpointState.unavailable_endpoints` is `HashMap<CosmosEndpoint, ...>`. Using `AccountEndpoint` here will confuse implementers about whether these are the same type or a new wrapper. Should be aligned to `CosmosEndpoint`.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+   suitable for background probes.
+
+```rust
+// Probe retry is a simple loop in run_sweep():
+for attempt in 0..config.max_probe_attempts {
+    let result = probe.probe(&ep).await;
+    match &result {
+        ProbeResult::Healthy { .. } => return result,
+        ProbeResult::Unhealthy { .. } if attempt + 1 < config.max_probe_attempts => {
+            tokio::time::sleep(backoff_delay(attempt)).await;
```
FabianMeiswinkel: **🟡 Recommendation — `backoff_delay(attempt)` is called but never defined**

This retry snippet calls `backoff_delay(attempt)` but the function body, initial delay, maximum delay, and backoff factor are never specified anywhere in the spec.

The PR description says "500ms initial backoff" and the Python SDK uses `retry_after_ms + 2^retry_count` with an initial delay of 500ms and a cap of 3 minutes (180,000ms). But the spec itself doesn't document these values. Without a defined algorithm, implementations could diverge significantly.

Suggest adding to this section:
- Initial delay: 500ms (matching Python)
- Backoff factor: 2x exponential
- Maximum delay cap: 3 minutes (matching Python's `max_retry_after_in_milliseconds = 180000`)

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+            .collect();
+
+        let results = futures::future::join_all(probe_futures).await;
+
+        // Apply results per-endpoint against the latest state to avoid
+        // overwriting concurrent reactive updates (no snapshot+swap).
+        for (endpoint, result) in &results {
+            match result {
+                ProbeResult::Healthy { .. } => {
+                    endpoint_state_store.mark_available(endpoint);
```
FabianMeiswinkel: **🟡 Recommendation — Health monitor bypasses `LocationStateStore.apply()` single-writer model**

The sweep calls `mark_available`/`mark_unavailable` directly per-endpoint, but the Transport Pipeline Spec §4.6 describes `LocationStateStore.apply(effects)` as "the single mutation entry point used by STAGE 6 of the operation loop."

This introduces a second mutation path, bypassing `LocationEffect` routing:
1. **Observability**: `LocationEffect` provides a single observability funnel — health monitor mutations won't be visible in effect logs.
2. **Consistency**: `mark_endpoint_available` doesn't exist as a `LocationEffect` variant — there's no `LocationEffect::MarkEndpointAvailable`.

Suggest either adding `LocationEffect::MarkEndpointAvailable` and routing health results through `apply()` for consistency, or explicitly documenting why the health monitor is an exception to the single-writer model.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+```rust
+// driver/health/monitor.rs
+
+impl EndpointHealthMonitor {
+    /// The main background loop. Runs an initial sweep immediately,
+    /// then periodically until the monitor is dropped.
+    ///
+    /// Uses `azure_core` async primitives (not tokio directly) to remain
+    /// async-runtime agnostic.
+    async fn sweep_loop(
```
FabianMeiswinkel, line 376: **🟡 Recommendation — Python SDK parity divergence: independent sweep vs. piggybacked health checks**

This spec describes an independent `EndpointHealthMonitor` with its own `azure_core::task::spawn` background loop. However, the Python SDK does **not** have an independent background health sweep. Instead:
- `_refresh_endpoint_list_private()` is triggered every 5 minutes (on request, not independently)
- On startup: `_start_background_refresh(self._endpoints_health_check, ...)` — one-shot after init
- On refresh: `_start_background_refresh(self._refresh_database_account_and_health, ...)` — health checks piggybacked on metadata refresh

The independent sweep loop is a design improvement (more reliable probing regardless of request patterns), but it's a **significant architectural divergence** from the Python SDK. Open question #3 hints at this but doesn't acknowledge the inverse relationship.

Suggest adding a "Differences from Python SDK" section so reviewers and future maintainers understand this architectural choice is intentional.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+    /// Steps:
+    /// 1. Snapshot current endpoint state.
+    /// 2. Determine which endpoints to probe.
+    /// 3. Probe all endpoints concurrently.
+    /// 4. Apply results: mark healthy endpoints available, unhealthy unavailable.
+    async fn run_sweep(
+        config: &HealthMonitorConfig,
+        probe: &dyn HealthProbe,
+        endpoint_state_store: &AccountEndpointStateStore,
+        account_metadata_cache: &AccountMetadataCache,
```
FabianMeiswinkel: **🟡 Recommendation — Unused `account_metadata_cache` parameter**

`run_sweep` accepts `account_metadata_cache: &AccountMetadataCache` but only calls `endpoint_state_store.snapshot()` for endpoint selection — the metadata cache is never referenced in the sweep body. This appears to be a forward-looking parameter for open question #3 (piggybacking metadata refresh on sweeps), but currently it's dead code that will confuse implementers.

Either:
- Remove the parameter until open question #3 is resolved, or
- Add a comment explaining it's reserved for future metadata refresh piggybacking.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+The monitor is the central coordinator. It owns the background task, determines which
+endpoints to probe, executes probes concurrently, and applies results to the endpoint
+state store.
+
+```rust
+// driver/health/monitor.rs
+
+/// Configuration for the endpoint health monitor.
+#[derive(Clone, Debug)]
+pub(crate) struct HealthMonitorConfig {
```
FabianMeiswinkel: **🟡 Recommendation — `HealthMonitorConfig` and `HealthCheckOptions` mapping unclear**

§3.3 defines `HealthMonitorConfig` (internal) and §5.1 defines `HealthCheckOptions` (public), both with identical fields (`refresh_interval`, `max_probe_attempts`). The spec doesn't explain the mapping between them:
- Is `HealthMonitorConfig` constructed from `HealthCheckOptions` via `From`/`Into`?
- Does `HealthMonitorConfig` add any internal-only fields later?
- Or should they be collapsed into one type?

Suggest adding a brief note in §5.2 showing how `DriverOptions.health_check: HealthCheckOptions` maps to the internal `HealthMonitorConfig` used by `EndpointHealthMonitor::start()`.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+    /// 2. Determine which endpoints to probe.
+    /// 3. Probe all endpoints concurrently.
+    /// 4. Apply results: mark healthy endpoints available, unhealthy unavailable.
+    async fn run_sweep(
+        config: &HealthMonitorConfig,
+        probe: &dyn HealthProbe,
+        endpoint_state_store: &AccountEndpointStateStore,
+        account_metadata_cache: &AccountMetadataCache,
+    ) {
+        let state = endpoint_state_store.snapshot();
```
FabianMeiswinkel: **🟢 Suggestion — Sweep applies results against stale endpoint list if metadata refreshes mid-sweep**

The sweep calls `endpoint_state_store.snapshot()` at the start, determines endpoints to probe, then applies results with `mark_available`/`mark_unavailable`. If the `AccountMetadataCache` refreshes during the sweep (rebuilding the endpoint state with new regions), the results are applied against a potentially stale endpoint list.

For example, if region X was removed by refresh, the sweep might call `mark_available(regionX)`. The CAS loop will succeed (it applies against the *current* state), but `mark_endpoint_available` only removes from the unavailable set — it wouldn't re-add a region removed from the preferred lists. So the consequence is limited, but worth a brief note here confirming this edge case is safe.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+            &*probe,
+            &endpoint_state_store,
+            &account_metadata_cache,
+        )
+        .await;
+
+        loop {
+            // Wait for either the refresh interval to elapse or shutdown,
+            // using azure_core async primitives (runtime-agnostic).
+            let sleep = azure_core::task::sleep(config.refresh_interval);
```
FabianMeiswinkel: **🟢 Suggestion — No jitter on sweep interval; thundering herd risk**

The sweep loop sleeps for exactly `config.refresh_interval` between sweeps. If multiple driver instances are created simultaneously (e.g., horizontally scaled service), they'll synchronize their health probes, creating a thundering herd against gateway endpoints every 5 minutes.

The Python SDK avoids this because health checks are triggered by metadata refresh gated by `last_refresh_time` checked on request paths (naturally jittered). The independent sweep loop doesn't have this natural jitter.

Suggest adding a small random jitter (e.g., ±10% of `refresh_interval`) to the sweep sleep, or noting this as a future optimization.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+
+### Relationship to Transport Pipeline Spec
+
+This spec extends the Transport Pipeline Spec (§4.4) by adding:
+1. **Proactive probing** — a background task that calls a health probe on each endpoint
+2. **Availability recovery** — marking endpoints available again after successful probes
+3. **Startup probing** — verifying endpoint reachability in the background during client initialization
+
+The `AccountEndpointState`, `AccountEndpointStateStore`, and
+`expire_unavailable_endpoints` systems defined in §4.4 are consumed as-is (once
```
FabianMeiswinkel: **🟢 Suggestion — `expire_unavailable_endpoints` interaction not clarified**

The Transport Pipeline Spec §4.4 defines `expire_unavailable_endpoints(state, now, expiry_duration)` which removes stale unavailable entries based on time. This spec's `mark_available` serves a similar purpose (removing endpoints from the unavailable set). The interaction should be clarified:
- Does `expire_unavailable_endpoints` still run independently of health checks?
- Do health probes make expiration redundant, or do they complement each other?

This matters because a health probe can mark an endpoint available immediately on recovery, while expiration only removes entries after a fixed duration. Both mechanisms write to the same `unavailable_endpoints` map.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+2. **Probe-level (transient failures)**: The health monitor retries failed probes
+   up to `max_probe_retries` times with exponential backoff. This is a simple retry
+   loop in the sweep, not a shared policy object — the pipeline's operation-level
+   retry (`OperationRetryState`) is designed for user-facing operations and not
+   suitable for background probes.
+
+```rust
+// Probe retry is a simple loop in run_sweep():
+for attempt in 0..config.max_probe_attempts {
+    let result = probe.probe(&ep).await;
```
FabianMeiswinkel: **💬 Observation — Python SDK scales per-probe timeout on retry; Rust spec doesn't**

The Python `HealthCheckRetryPolicy.ShouldRetry()` adjusts `request.read_timeout_override` exponentially on each retry (`timeout ** 2`, capped at `ConnectionPolicy.ReadTimeout`). This means early retries use a short timeout (5s) and later retries wait longer, giving transient issues more time to resolve.

The Rust spec's probe retry (§3.5) doesn't mention timeout scaling. The `deadline: None` comment says per-request timeout is deferred to PR #3871. When #3871 lands, consider whether probe retries should also scale the timeout.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+/// Trait for probing a single Cosmos DB gateway endpoint.
+///
+/// Implementations must be `Send + Sync` (shared across the background task
+/// and potentially multiple concurrent probes).
+///
+/// The trait is intentionally minimal: one method, one endpoint, one result.
+/// The orchestration layer handles scheduling, concurrency, retries, and
+/// state updates.
+#[async_trait]
+pub(crate) trait HealthProbe: Send + Sync {
```
FabianMeiswinkel: **💬 Positive — `HealthProbe` trait abstraction is clean and well-designed**

The separation between probe logic and orchestration means the future `GatewayHealthProbe` migration (§7) is genuinely trivial — just a new struct implementing the same trait. The structured `ProbeFailure` type is also a good choice over strings for diagnostics integration. Nice work here.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
FabianMeiswinkel: The discussed approach was for the service team to create the endpoint (which might initially just redirect to getaccount internally) - but even that might take a few weeks until it is deployed

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+  eviction). That is a transport-layer concern; this spec is about service-level reachability.
+- **User-facing health API**: Health checks are internal. No public API is exposed.
+
+---
+
+## 2. Architecture Overview
+
+```text
+┌─────────────────────────────────────────────────────────────┐
+│                    CosmosDriverRuntime                       │
```
FabianMeiswinkel: NIT: fix the whitespaces

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+│                         │   mark_endpoint_unavailable()    │ │
+│                         └─────────────────────────────────┘ │
+└─────────────────────────────────────────────────────────────┘
+```
+
+### Relationship to Transport Pipeline Spec
+
+This spec extends the Transport Pipeline Spec (§4.4) by adding:
+1. **Proactive probing** — a background task that calls a health probe on each endpoint
+2. **Availability recovery** — marking endpoints available again after successful probes
```
FabianMeiswinkel, line 100: IMO worth calling out as an explicit goal that when in doubt we treat endpoints as healthy. We HAVE to AVOID similar issue like in Python where transient endpoint unavailable ends-up in removingn the endpoint for the lifetime of the process.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+/// Result of a single endpoint health probe.
+#[derive(Clone, Debug)]
+pub(crate) enum ProbeResult {
+    /// The endpoint responded successfully.
+    Healthy {
+        /// How long the probe took (for diagnostics/logging).
+        latency: Duration,
+    },
+    /// The endpoint did not respond or returned an error.
+    Unhealthy {
```
FabianMeiswinkel, line 170: Worth also to capture how many consecutive probes failed?

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
FabianMeiswinkel: Except that Ashley makes a good point of enum vs. trait :-)

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+            credential,
+            user_agent,
+        }
+    }
+}
+
+#[async_trait]
+impl HealthProbe for GetDatabaseAccountProbe {
+    async fn probe(&self, endpoint: &AccountEndpoint) -> ProbeResult {
+        let start = Instant::now();
```
FabianMeiswinkel, line 208: Any reason to not just use execute_operation with all other regions excluded?

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
analogrelay: But we expect to have the endpoint at the expected URL by GA? If so, then we don't need the trait/enum at all. We can just change the endpoint we use when it's fully deployed.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+}
+```
+
+### 3.4 Endpoint Selection — Which Endpoints to Probe
+
+Following the Python SDK pattern, the monitor probes:
+
+1. All **read endpoints** in the preferred locations list.
+2. The **first write endpoint** (unless it's already in the read set).
+3. Deduplicates so no endpoint is probed twice per sweep.
```
FabianMeiswinkel, line 276: What about the global endpont - would we ever use the global endpoint in rust for anything but getting account topology? IMO we should not - but worth deciding this explicitly?

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
analogrelay: We should use the existing Jitter helpers we have in the driver to do this :)

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
analogrelay: With agents creating these diagrams quite often, I wonder if we could build a linter that checks for alignment of boxes like this. I've seen agents make this mistake before and if they had a linter they were asked to run that identified the issue, they'd auto fix it.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
```diff
@@ -0,0 +1,855 @@
...
+
+/// Options controlling background endpoint health checking.
+///
+/// Health checks probe regional gateway endpoints to detect unavailability
+/// and recovery. Results feed into the endpoint routing system.
+///
+/// Health checks are always enabled and always run on startup (non-blocking).
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub struct HealthCheckOptions {
```
FabianMeiswinkel: Dow e really need to make this configurable (via public API)? Shouldn't it always align with the get Account topology?

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
FabianMeiswinkel: Yeah - I use a VSCode extension to fix it "Markdown all in one" - but also saw these error regularly in Codex and Claude Opus

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
FabianMeiswinkel: Let us double-check - I don't think there is a clear ETA - and deployment in non-public clouds usually takes 3+ months after publci clouds - so, by GA timeline we probabyl cannot rely yet on the endpoint being there.

analogrelay: Another option is a discovery approach, such as Get Account Properties returning a `healthCheckEndpoints` collection or similar. Since we're going to run Get Account Properties before initializing the driver (as part of async initialization) we could check for that value and use it if present, otherwise fall back to just using Get Account Properties itself.

analogrelay: Basically, I'm just trying to see if we can make this Just Start Working ™️ when the service rolls it out instead of requiring an SDK update.

FabianMeiswinkel: IMO discovery approach via health endpoit is over-engineering and requires also service deployments. My 2 cents - the pragmatic approach is what Tomas is proposing in the PR - if we want something unified by GA to avoid teh eventual SDK update enecssity we need to push service team to accelerate providing the health endpoint. Everything else just over-complicates for some very short-term gain. I am fine with requiring an SDK update as long as it does not change any public API because I expect customers using GA version to upgarde relatively quickly initially anyway.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Updated: failed probes log at `warn!`, first recovery at `info!`, routine successes not logged.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Switched to enum with `GetAccountMetadata` variant per DOP guidance. No trait.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: ProbeFailure now carries `azure_core::Error` directly — string formatting deferred to display time.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Now uses `BackgroundTaskManager` (same as GlobalPartitionEndpointManager). Drop = cancellation via Rust ownership.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Agreed — removed CancellationToken. Using `BackgroundTaskManager` with drop-based cancellation.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Using `FuturesUnordered` for runtime-agnostic concurrent probing with immediate result processing.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Health probe results now routed through `LocationEffect::MarkEndpointAvailable/Unavailable` via `LocationStateStore::apply()`.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Updated: failed probes at `warn!`, first recovery at `info!`, routine successes eliminated.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Replaced MockHealthProbe with FaultInjection-based testing throughout.

tvaron3: Replaced MockHealthProbe with FaultInjection-based testing throughout.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Fixed: retry loop now inlined in `probe_with_retry` helper called within each concurrent future.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Addressed: spec now explicitly requires Transport Spec amendment to add `LocationEffect::MarkEndpointAvailable/Unavailable` variants.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Switched to `CosmosEndpoint` throughout — aligned with Transport Spec canonical type.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Defined: 500ms initial, 2x exponential, 3min cap, ±25% jitter via existing `with_jitter` helper.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Resolved: all results routed through `LocationEffect` via `LocationStateStore::apply()` — single-writer model preserved.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Added Differences from Python SDK section documenting the intentional divergence to independent sweep.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Removed `account_metadata_cache` from struct, `start()`, `sweep_loop()`, and `run_sweep()` entirely.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Collapsed: removed both `HealthMonitorConfig` and `HealthCheckOptions`. All config is hardcoded constants — no public API.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Added note in `run_sweep`: stale endpoint list mid-sweep is safe — `mark_available` on a removed region is a no-op.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Added ±10% jitter on sweep interval using existing `with_jitter` helper.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Expiration-based recovery removed. Health checks are the sole mechanism for endpoint availability/recovery.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Added note in §3.6 to consider timeout scaling on retry when PR #3871 lands.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Fixed ASCII diagram alignment.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Added explicit design principle: endpoints always eligible for routing unless customer-excluded. Transient unavailability deprioritizes but never permanently removes.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Added `consecutive_failures: u32` to `ProbeFailure` and tracking in sweep via `failure_counts` map.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Switched: probes now use `execute_operation` with excluded regions, reusing the full pipeline.

tvaron3: Switched: probes now use `execute_operation` with excluded regions, reusing the full pipeline.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Added explicit decision: global endpoint is for topology discovery only, never probed. Added to §3.5 and Open Question #3.

sdk/cosmos/azure_data_cosmos_driver/docs/HEALTH_CHECKS_SPEC.md
tvaron3: Removed all public configuration. Health check config is fully internal with hardcoded defaults — no `HealthCheckOptions`.

PR #3868, opened 2026-03-05
Cosmos: Layered Options Proc-Macro

analogrelay:
This PR creates the proc-macro defined in the [HierarchicalConfigModel](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/cosmos/azure_data_cosmos/docs/HierarchicalConfigModel.md) spec.

It's just the proc-macro for now, none of the options are added to the driver yet. Proc-macros can be a bit of a pain to review. Keep an eye on the `quote!` macro (I _highly_ recommend [reading the docs briefly](https://docs.rs/quote/latest/quote/) if you aren't familiar with it). The tests give a pretty clear picture of the resulting code too, and there are integration tests that actually validate the layering behavior.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos_macros/src/builder.rs
+++ b/sdk/cosmos/azure_data_cosmos_macros/src/builder.rs
@@ -65,6 +65,7 @@ pub fn generate_builder(input: &OptionsInput) -> Result<TokenStream> {
             #(#setters)*
 
             /// Consumes the builder and returns the constructed option group.
+            #[must_use]
             #vis fn build(self) -> #struct_name #ty_generics {
                 #struct_name {
                     #(#build_fields),*
@@ -123,6 +124,7 @@ mod tests {
                 }
 
                 /// Consumes the builder and returns the constructed option group.
+                #[must_use]
                 pub fn build(self) -> TestOptions {
                     TestOptions {
                         field_a: self.field_a,
--- a/sdk/cosmos/azure_data_cosmos_macros/src/env.rs
+++ b/sdk/cosmos/azure_data_cosmos_macros/src/env.rs
@@ -42,14 +42,37 @@ pub fn generate_from_env(input: &OptionsInput) -> Result<TokenStream> {
                     #field_name: env_var(#env_var)
                         .ok()
                         .map(|v| v.split(',')
-                            .filter_map(|s| s.trim().parse().ok())
+                            .filter_map(|s| {
+                                let trimmed = s.trim();
+                                match trimmed.parse() {
+                                    Ok(parsed) => Some(parsed),
+                                    Err(_) => {
+                                        ::tracing::warn!(
+                                            env_var = #env_var,
+                                            value = trimmed,
+                                            "failed to parse element from environment variable; skipping",
+                                        );
+                                        None
+                                    }
+                                }
+                            })
                             .collect())
                 }
             } else {
                 quote! {
                     #field_name: env_var(#env_var)
                         .ok()
-                        .and_then(|v| v.parse().ok())
+                        .and_then(|v| match v.parse() {
+                            Ok(parsed) => Some(parsed),
+                            Err(_) => {
+                                ::tracing::warn!(
+                                    env_var = #env_var,
+                                    value = %v,
+                                    "failed to parse environment variable; ignoring",
+                                );
+                                None
+                            }
+                        })
                 }
             }
         } else {
@@ -115,7 +138,17 @@ mod tests {
                     Self {
                         field_a: env_var("MY_VAR_A")
                             .ok()
-                            .and_then(|v| v.parse().ok()),
+                            .and_then(|v| match v.parse() {
+                                Ok(parsed) => Some(parsed),
+                                Err(_) => {
+                                    ::tracing::warn!(
+                                        env_var = "MY_VAR_A",
+                                        value = %v,
+                                        "failed to parse environment variable; ignoring",
+                                    );
+                                    None
+                                }
+                            }),
                         field_b: None
                     }
                 }
--- a/sdk/cosmos/azure_data_cosmos_macros/src/parse.rs
+++ b/sdk/cosmos/azure_data_cosmos_macros/src/parse.rs
@@ -60,6 +60,15 @@ impl Layer {
             )),
         }
     }
+
+    /// Returns the canonical ordering index for this layer.
+    fn order(&self) -> usize {
+        match self {
+            Layer::Runtime => 0,
+            Layer::Account => 1,
+            Layer::Operation => 2,
+        }
+    }
 }
 
 /// Parsed representation of a single field in the option struct.
@@ -143,6 +152,25 @@ fn parse_layers_attr(attrs: &[syn::Attribute]) -> Result<Vec<Layer>> {
                 Err(meta.error("expected `layers(...)`"))
             }
         })?;
+
+        // Validate canonical ordering and reject duplicates.
+        for i in 1..layers.len() {
+            if layers[i] == layers[i - 1] {
+                return Err(Error::new(
+                    Span::call_site(),
+                    format!("duplicate layer `{}`", layers[i].ident()),
+                ));
+            }
+            if layers[i].order() <= layers[i - 1].order() {
+                return Err(Error::new(
+                    Span::call_site(),
+                    format!(
+                        "layers must follow canonical ordering: {}",
+                        KNOWN_LAYERS.join(", "),
+                    ),
+                ));
+            }
+        }
 
         return Ok(layers);
     }
@@ -332,6 +360,37 @@ mod tests {
                 "unknown layer `global`, expected one of: runtime, account, operation",
                 e.to_string()
             ),
+            Ok(_) => panic!("expected error"),
+        }
+    }
+
+    #[test]
+    fn out_of_order_layers_errors() {
+        let input: DeriveInput = syn::parse_quote! {
+            #[options(layers(operation, runtime))]
+            struct TestOptions {
+                pub field_a: Option<String>,
+            }
+        };
+        match OptionsInput::from_derive_input(&input) {
+            Err(e) => assert_eq!(
+                "layers must follow canonical ordering: runtime, account, operation",
+                e.to_string()
+            ),
+            Ok(_) => panic!("expected error"),
+        }
+    }
+
+    #[test]
+    fn duplicate_layer_errors() {
+        let input: DeriveInput = syn::parse_quote! {
+            #[options(layers(runtime, runtime))]
+            struct TestOptions {
+                pub field_a: Option<String>,
+            }
+        };
+        match OptionsInput::from_derive_input(&input) {
+            Err(e) => assert_eq!("duplicate layer `runtime`", e.to_string()),
             Ok(_) => panic!("expected error"),
         }
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -395,6 +395,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.116",
+ "tracing",
 ]
 
 [[package]]
--- a/sdk/cosmos/azure_data_cosmos_macros/Cargo.toml
+++ b/sdk/cosmos/azure_data_cosmos_macros/Cargo.toml
@@ -21,5 +21,8 @@ proc-macro2.workspace = true
 quote.workspace = true
 syn.workspace = true
 
+[dev-dependencies]
+tracing.workspace = true
+
 [lints]
 workspace = true


sdk/cosmos/azure_data_cosmos_macros/src/view.rs
```diff
@@ -0,0 +1,563 @@
...
+}
+
+/// Generates a shadow accessor: returns `Option<&T>` walking highest → lowest priority.
+fn generate_shadow_accessor(field: &OptionField, layers: &[Layer]) -> Result<TokenStream> {
+    let field_name = &field.ident;
+    let inner_type = &field.inner_type;
+    let last_layer_idx = layers.len() - 1;
+
+    // Build the chain: operation → account → runtime → env
+    // Layers are stored in order [runtime, account, operation], so we walk in reverse.
```
FabianMeiswinkel, line 97: **🔴 Blocking — Layer ordering determines resolution priority but is not validated or documented**

The `layers(...)` list is positional — the **last** layer listed becomes the highest-priority layer in resolution (gets the special `Option<&'a T>` treatment). But nothing validates or documents that order matters.

Here, layers are iterated in reverse, so the last listed layer is checked first (highest priority). If a user writes `#[options(layers(operation, runtime, account))]`, the macro silently generates a View where `account` gets highest priority and `operation` gets lowest — **backwards** from the spec's `Operation → Account → Runtime → Environment`.

The canonical order (`runtime < account < operation`) is defined in `KNOWN_LAYERS` but never enforced. Consider:
1. Validating that the layer list follows canonical ordering (or at minimum, that `operation` is last when present), or
2. Deriving priority from the enum variant order rather than the position in the attribute list.

sdk/cosmos/azure_data_cosmos_macros/src/view.rs
```diff
@@ -0,0 +1,563 @@
...
+        let field_name = layer.ident();
+        new_body_fields.push(quote! { #field_name });
+    }
+
+    Ok(quote! {
+        /// Snapshot view across all layers for resolution.
+        #[automatically_derived]
+        #vis struct #view_name<'a> {
+            #(#struct_fields),*
+        }
```
FabianMeiswinkel, line 62: **🟡 Design — View lifetime parameter deviates from spec's "self-contained" design**

The [spec explicitly states](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/cosmos/azure_data_cosmos/docs/HierarchicalConfigModel.md):
> "The View is self-contained — it owns or shares all the data it needs. No lifetime parameters, no borrowing issues across await points."

This implementation introduces `'a` by borrowing the bottom (highest-priority) layer as `Option<&'a T>` instead of owning it. When this View is used in async operation handlers, the borrowed data must outlive the View, constraining how it can be passed across `.await` points.

The spec's owned approach (`operation: RequestOptions`) avoids all lifetime complications at the cost of one clone of a small struct. The `Option`-wrapping of all layers is a good pragmatic improvement — but the borrowed-vs-owned choice for the bottom layer is architectural and should be explicitly justified in the spec before this pattern permeates the codebase.

sdk/cosmos/azure_data_cosmos_macros/src/parse.rs
```diff
@@ -0,0 +1,497 @@
...
+}
+
+fn parse_layers_attr(attrs: &[syn::Attribute]) -> Result<Vec<Layer>> {
+    for attr in attrs {
+        if !attr.path().is_ident("options") {
+            continue;
+        }
+
+        let mut layers = Vec::new();
+        attr.parse_nested_meta(|meta| {
```
FabianMeiswinkel, line 139: **🟡 Correctness — No validation of duplicate layers**

`#[options(layers(runtime, runtime))]` would generate a View struct with two `runtime` fields of the same name. The resulting compile error comes from the generated code rather than from the macro itself, making it confusing to debug.

Suggest adding a dedup check here:
```rust
for ident in &layer_idents {
    let layer = Layer::from_str(&ident.to_string(), ident.span())?;
    if layers.contains(&layer) {
        return Err(Error::new(ident.span(), format!("duplicate layer `{}`", ident)));
    }
    layers.push(layer);
}
```

sdk/cosmos/azure_data_cosmos_macros/src/env.rs
```diff
@@ -0,0 +1,140 @@
...
+            if is_vec_type(inner_type) {
+                quote! {
+                    #field_name: env_var(#env_var)
+                        .ok()
+                        .map(|v| v.split(',')
+                            .filter_map(|s| s.trim().parse().ok())
+                            .collect())
+                }
+            } else {
+                quote! {
```
FabianMeiswinkel, line 62: **🟡 Robustness — Silent env var parse failures make configuration bugs invisible**

If a user sets `AZURE_COSMOS_REQUEST_TIMEOUT=abc` for a `u64` field, the field silently becomes `None` with no indication the env var was found but couldn't be parsed. This makes configuration typos extremely hard to debug — the user thinks they've set a value, but the SDK silently ignores it.

Consider at minimum logging a warning (via `tracing::warn!`) when an env var is present but fails to parse:
```rust
#field_name: env_var(#env_var)
    .ok()
    .and_then(|v| match v.parse() {
        Ok(parsed) => Some(parsed),
        Err(_) => {
            // tracing::warn! in the generated code
            None
        }
    })
```

(I know this is consistent with the spec's design, so it's not a deviation — but it's a usability footgun worth addressing.)

sdk/cosmos/azure_data_cosmos_macros/src/view.rs
```diff
@@ -0,0 +1,563 @@
...
+}
+
+/// Generates a merge accessor for `#[option(merge = "extend")]` fields.
+fn generate_merge_accessor(field: &OptionField, layers: &[Layer]) -> Result<TokenStream> {
+    let field_name = &field.ident;
+    let inner_type = &field.inner_type;
+    let last_layer_idx = layers.len() - 1;
+
+    // Merge from lowest to highest priority (env → runtime → account → operation).
+    // Each layer is Optional, so we unwrap before checking the field.
```
FabianMeiswinkel, line 143: **🟡 Performance — Merge accessor clones entire collections unnecessarily**

`v.clone()` clones the **entire** `HashMap`/`Vec` before extending, allocating a temporary that's immediately dropped after iteration.

More efficient patterns that avoid the intermediate collection allocation:

For `HashMap<K, V>`:
```rust
merged.extend(v.iter().map(|(k, v)| (k.clone(), v.clone())));
```

For `Vec<T>`:
```rust
merged.extend(v.iter().cloned());
```

This matters when merge accessors are called on hot paths with large header maps.

sdk/cosmos/azure_data_cosmos_macros/src/view.rs
```diff
@@ -0,0 +1,563 @@
...
+            });
+        } else {
+            // Arc layer — extract nested field and clone into a new Arc
+            new_args.push(quote! {
+                self.#layer_name.as_ref()
+                    .and_then(|l| l.#field_name.as_ref())
+                    .map(|v| ::std::sync::Arc::new(v.clone()))
+            });
+        }
+    }
```
FabianMeiswinkel, line 222: **🟡 Robustness — No compile-time guarantee that nested types have corresponding Views**

If someone writes `#[option(nested)]` on a field whose type doesn't derive `CosmosOptions`, the error comes from generated code (`cannot find type FooView in this scope`) rather than from the macro. This is confusing.

The macro can't verify trait impls at derive time, but consider generating a diagnostic comment or a trait bound that produces a clearer error message pointing back to the user's code.

sdk/cosmos/azure_data_cosmos_macros/src/builder.rs
```diff
@@ -0,0 +1,137 @@
...
+                Self {
+                    #(#default_fields),*
+                }
+            }
+
+            #(#setters)*
+
+            /// Consumes the builder and returns the constructed option group.
+            #vis fn build(self) -> #struct_name #ty_generics {
+                #struct_name {
```
FabianMeiswinkel, line 70: **🟢 Suggestion — `#[must_use]` on `Builder::build()`**

Since `build()` returns an owned struct that should always be captured, adding `#[must_use]` would catch accidental discards:
```rust
RequestOptionsBuilder::new()
    .with_consistency_level("Session".to_string())
    .build(); // Oops, forgot to assign — no warning without #[must_use]
```

Consider generating `#[must_use]` on the `build()` method.

sdk/cosmos/.cspell.json
```diff
@@ -108,6 +109,7 @@
     "pksysdocs",
     "pluggable",
```
FabianMeiswinkel, line 110: **🟢 Nit — `"qself"` is out of alphabetical order**

Case-insensitively, `q > p`, so `"qself"` should appear after `"purgeable"`, not between `"pointee"` and `"PPAF"`.

FabianMeiswinkel (commented):
## Deep Review: Cosmos Layered Options Proc-Macro

Well-designed proc-macro with good test coverage and clean module separation (parse/view/builder/env). The integration tests are thorough — covering shadow resolution, merge semantics, nested groups, and edge cases. Prior Copilot comments appear addressed.

### Summary of findings

**Blocking (1):**
- Layer ordering determines resolution priority but is not validated — writing layers in the wrong order silently produces backwards resolution

**Recommendations (6):**
- View lifetime `'a` deviates from spec's "self-contained" / no-lifetime design — should be explicitly justified
- No duplicate-layer validation — `layers(runtime, runtime)` produces a confusing generated-code error
- Silent env var parse failures — `TIMEOUT=abc` silently becomes `None` with no diagnostic
- No compile-time guarantee nested types have corresponding Views
- Merge accessor clones entire collections unnecessarily (`v.clone()` → `v.iter().cloned()`)
- Missing edge-case tests: duplicate layers, single layer, only-operation layer, wrong order

**Suggestions (2):**
- `#[must_use]` on `Builder::build()` to catch accidentally discarded results
- `"qself"` out of alphabetical order in `.cspell.json`

**Not posted inline but worth noting:**
- Generated types lack `SafeDebug` derive (per AGENTS.md: "Generated code should also use SafeDebug")
- Builder should implement `Default` (since `new()` already produces all-`None`)
- Spec should be updated to document implementation deviations (all-`Option` layers, borrowed bottom layer, always-present env layer, `from_env_vars` test helper)
- No test for env parse failure behavior confirming the "silently None" semantics is intentional

sdk/cosmos/azure_data_cosmos_macros/src/view.rs
FabianMeiswinkel: Won't fix. It is fine for the use case.

sdk/cosmos/azure_data_cosmos_macros/src/view.rs
FabianMeiswinkel: Won't fix - this is a macro - no need to add additional validation.

sdk/cosmos/azure_data_cosmos_macros/src/view.rs
FabianMeiswinkel: Won't fix - perf impact is irrelevant here - it is a macro - only used when generating code.

sdk/cosmos/azure_data_cosmos_macros/src/view.rs
FabianMeiswinkel: Won't fix - again - this is a macro - all good.

sdk/cosmos/azure_data_cosmos_macros/src/parse.rs
FabianMeiswinkel: Won't fix - this is fine - the macro just helps generate code - dev can own avoiding this.

sdk/cosmos/azure_data_cosmos_macros/src/env.rs
FabianMeiswinkel: IMO this is valid feedback. keeping it active

sdk/cosmos/azure_data_cosmos_macros/src/builder.rs
FabianMeiswinkel: NIT but valid

sdk/cosmos/azure_data_cosmos_macros/src/view.rs
analogrelay: True, but it wouldn't be hard at all to add tbh.

sdk/cosmos/.cspell.json
FabianMeiswinkel: Won't fix - adding new terms at the end avoids merge conflicts.

sdk/cosmos/azure_data_cosmos_macros/src/view.rs
analogrelay: Hmm. It might be right here, tbh. The reason for the `'a` lifetime is to make it so that _nested_ views don't have to clone data. But for the top level options we probably still need to use Arcs. The nested views can always be "reborrowed" after an `.await`. If the top-level options are borrowed, that gets messy.

I think we can start here though and revise this if/when it becomes an issue.

sdk/cosmos/azure_data_cosmos_macros/src/view.rs
heaths: One thing to keep in mind is that it's also compiled and executed by the LSP e.g., rust-analyzer. While maybe not bad enough if the count is low to be noticeable, it's worth calling out.

PR #3871, opened 2026-03-05
[Internal] Add dynamic timeout escalation spec

### LINKED ISSUE #3758: Cosmos: Add Support for Gateway Timeouts
### Acceptance Criteria:

Today, the cosmos rust SDK does not have a mechanism to impose timeouts on a GW request. This causes a request to indefinitely wait on the GW for a corresponding response, or the pipeline timeout reaches.

- Add a mechanism to apply `HttpTimeoutPolicy` on all GW metadata and dataplane requests.
- Add a retry mechanism to retry a potential request at-least 3 times, when such timeout occurres, each time with a different timeout value.

tvaron3:
## Dynamic Timeout Spec

Adds `DYNAMIC_TIMEOUTS_SPEC.md` to `sdk/cosmos/azure_data_cosmos_driver/docs/` describing a design for escalating request timeouts on transport retries and adaptive connection timeout tuning.

### Motivation

When a request times out and is retried with the same timeout, it often fails again. Escalating the timeout on each retry gives the operation a better chance of succeeding without making the initial attempt unnecessarily slow.

### Key Design Decisions

- **Data plane request timeouts**: 6s → 10s → 65s (escalating per retry attempt)
- **Metadata request timeouts**: 5s → 10s → 20s (aligned with Java SDK `DatabaseAccount` pattern)
- **Existing retry budget is sufficient**: `max_failover_retries` defaults to 3, which naturally maps to the 3-tier ladder. No retry count change needed.
- **Fixed defaults**: the escalation ladder is not user-configurable; existing `ConnectionPoolOptions` min/max bounds still clamp the effective values
- **E2E deadline**: end-to-end operation deadline still takes precedence over per-attempt ladder values
- **Adaptive connection timeout**: starts at 1s, escalates to 5s on >3 consecutive connection failures per endpoint (one-time persistent transition in `ShardedHttpTransport`)
- **`typespec_client_core` change**: proposes adding `timeout: Option<Duration>` to `Request` for per-request timeout overrides (requires separate cross-team PR)

### Cross-SDK Alignment

The Java SDK implements timeout escalation for gateway/metadata HTTP calls (QueryPlan: 0.5s→5s→10s, DatabaseAccount: 5s→10s→20s). This spec extends the pattern to both data plane and metadata requests in the Rust driver.

### Deferred

- Query plan timeout escalation (not yet implemented)


sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
```diff
@@ -0,0 +1,415 @@
...
+
+### Per-Attempt Connection Timeout
+
+The current architecture creates `HttpClient` instances with a fixed `connect_timeout` set at
+creation time (via `HttpClientConfig`). To support per-attempt connection timeouts, one of the
+following approaches must be used:
+
+#### Option A: Set Timeout on the Request
+
+Extend the per-request context or `Request` to carry both a connection timeout and a request
```
tvaron3: This approach should be taken. Please specify any changes needed in the azure core http request.

sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
```diff
@@ -0,0 +1,415 @@
...
+1. **Per-attempt connection timeout mechanism**: Which approach (Option A/B/C in §8) should be used
+   to apply per-attempt connection timeouts? This depends on the timeline for
+   `ShardedHttpTransport` (see `TRANSPORT_PIPELINE_SPEC.md` §6).
+
+2. **Metadata timeout ladder scope**: Should the metadata ladder (5s → 10s → 20s) apply to all
+   metadata operations uniformly, or should specific metadata operations (e.g., database account
+   reads) have their own ladders? For now, a single metadata ladder is proposed. Query plan and
+   address refresh are deferred.
+
+3. **Interaction with hedging**: When hedging is enabled (see `TRANSPORT_PIPELINE_SPEC.md` §4.2),
```
tvaron3: I think we should take this approach but would like to get others thoughts.

sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
```diff
@@ -0,0 +1,415 @@
...
+   the hedged attempt runs in a secondary region. Should the hedged attempt use the same ladder
+   position as the primary, or always start at tier 0? The hedged attempt is architecturally
+   independent, so starting at tier 0 is likely correct.
+
+4. **Diagnostic reporting**: Should the per-attempt timeout values be recorded in
```
tvaron3: Yes

sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
```diff
@@ -0,0 +1,507 @@
...
+4. **Leverage existing retry budget**: The operation pipeline already supports up to 3 failover
+   retries by default, which naturally maps to the 3 tiers of the escalation ladder.
+
+### Non-Goals
+
+- **User-configurable ladders**: The escalation ladder is a fixed internal default. Users cannot
+  override the step durations. The existing `ConnectionPoolOptions` min/max bounds still act as
+  clamping limits (see [§7](#7-interaction-with-connectionpooloptions-bounds)).
+- **Query plan timeout escalation**: Deferred until query plan execution is implemented.
+- **Address refresh timeout escalation**: Deferred until direct mode is implemented.
```
tvaron3: Direct mode won't be implemented

sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
```diff
@@ -0,0 +1,602 @@
...
+---
+
+## 12. Open Questions
+
+1. **Metadata timeout ladder scope**: Should the metadata ladder (5s → 10s → 20s) apply to all
+   metadata operations uniformly, or should specific metadata operations (e.g., database account
+   reads) have their own ladders? For now, a single metadata ladder is proposed. Query plan and
+   address refresh are deferred.
+
+2. **Connection failure rate threshold**: What failure rate should trigger the connection timeout
```
tvaron3: **🟡 Recommendation** · Completeness: Unresolved Design Parameters

**Failure rate threshold and window are open questions that affect correctness**

The threshold (>50%? >N consecutive?) and window (last 20 attempts? last 30s?) are listed as open questions, but they directly determine whether the adaptive mechanism triggers at all or triggers too aggressively. These aren’t implementation details — they’re design decisions that change the observable behavior.

A threshold that’s too low causes false positives on transient blips (one bad connection in a burst of pool expansion triggers permanent 5s timeout). A threshold that’s too high means the mechanism never activates in practice.

**Suggested fix**: Resolve these before merging the spec. Propose concrete defaults (e.g., “>3 consecutive connection failures within a 30s window”) and document the rationale, even if the values are expected to be tuned later.

---

sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
```diff
@@ -0,0 +1,602 @@
...
+shards per endpoint. It already tracks per-shard health metrics and performs health sweeps.
+
+Connection timeout adaptation fits naturally into this model:
+
+1. **Track connection failure rate**: The `ShardedHttpTransport` monitors connection failures
+   (errors where `is_connect()` returns `true`) across all shards for an endpoint.
+2. **Threshold check**: When the connection failure rate exceeds a configurable threshold (e.g.,
+   >50% of recent connection attempts fail), the transport transitions to the elevated timeout.
+3. **Create new clients**: New `HttpClient` instances are created with `connect_timeout = 5s`
+   via the `HttpClientFactory`. Existing shards with the old timeout continue serving inflight
```
tvaron3: **🟢 Suggestion** · Design: Transition Latency

**“Exactly-once transition” means inflight requests on old shards keep failing at 1s during escalation**

When the transport decides to escalate, new `HttpClient` instances get 5s but existing shards continue with 1s until they drain naturally. If the network issue is affecting all connections, requests on old shards will keep timing out at 1s until those shards are reclaimed by the health sweep.

The transition latency depends entirely on how fast old shards drain — which could be many seconds if they have active connections.

Consider whether existing shards should be actively replaced (e.g., marked unhealthy immediately so the health sweep reclaims them on its next pass) rather than waiting for natural drain.

---

sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
```diff
@@ -0,0 +1,602 @@
...
+   requests and are eventually reclaimed when idle.
+4. **No `azure_core` changes needed**: This is entirely internal to the `ShardedHttpTransport`.
+   The `HttpClientFactory::create()` already accepts `HttpClientConfig` which includes
+   `connect_timeout`.
+
+```rust
+// In ShardedHttpTransport, on health sweep or after connection failure:
+if connection_failure_rate > CONNECT_TIMEOUT_ESCALATION_THRESHOLD {
```
tvaron3: **🟢 Suggestion** · Design: Scope Ambiguity

**Per-endpoint vs global scope for failure rate tracking is unspecified**

The pseudocode shows `connection_failure_rate` as a single value, but doesn’t specify whether it’s tracked per-endpoint or globally across all endpoints.

If one regional endpoint has connectivity issues (e.g., a specific Azure region), should ALL endpoints get the 5s timeout? Per-endpoint tracking would be more precise (only the slow region gets 5s) but adds complexity. Global tracking is simpler but could penalize healthy regions.

The spec should state the scope explicitly. Given that `ShardedHttpTransport` already manages shards per-endpoint, per-endpoint tracking would be a natural fit.

---

tvaron3 (commented):
## PR Deep Review — §10 Adaptive Connection Timeout

Focused review of the adaptive connection timeout section. 3 comments (1 recommendation, 2 suggestions).

sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
```diff
@@ -0,0 +1,598 @@
...
+#### No New Fields on `TransportRequest`
+
+The existing `deadline: Option<Instant>` field already serves as the single source of truth for
+timeout enforcement. The ladder's per-attempt timeout is folded into this field at the point where
+the `TransportRequest` is constructed — no `request_timeout: Duration` field is needed, and there
+is no risk of two competing timeout sources on the same struct.
+
+#### Required Changes
+
+**In `typespec_client_core`** — Add an optional per-request timeout to `Request`:
```
heaths: All client and client-method customization is done via the client options and client method options. Same as in every Azure SDK language. You need to add a field to either your client methods or `azure_core::http::ClientMethodOptions`. Customers should never have to interact with `Request`. And even if this was for client method implementations, that's what `Context` is for. Add a type to it that permiates all through the request pipeline. Don't need to modify `Request` itself.

sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
```diff
@@ -0,0 +1,598 @@
...
+11. [Open Questions](#11-open-questions)
+
+---
+
+## 1. Goals & Motivation
+
+### Why
+
+When a Cosmos DB request times out and is retried, using the same timeout value for the retry is
+often suboptimal. Transient network issues or momentary server load spikes may cause an initial
```
heaths, line 30: This is why the `Retry-After` header exists and is used by every Azure SDK language. Why does that not work here? Do you need a client-driven alternative?

sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
```diff
@@ -0,0 +1,598 @@
...
+    Duration::from_secs(10),
+    Duration::from_secs(20),
+];
+
+// Inside execute_transport_pipeline():
+let ladder = match pipeline_type {
+    PipelineType::DataPlane => DATAPLANE_REQUEST_TIMEOUT_LADDER,
+    PipelineType::Metadata => METADATA_REQUEST_TIMEOUT_LADDER,
+};
+let mut timeout_retry_count = 0_usize;
```
heaths, line 318: This whole thing would better be implemented as a custom `RetryPolicy` Cosmos can add when it creates the `Pipeline` in its client constructor. You can't remove the built-in `RetryPolicy`, but you can pass `RetryMode::none()` instead (this was easier than any default removal options or exclusion options we came up with for Rust).

heaths (changes requested):

sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
tvaron3: retry after is only if we receive a response. This is for scenarios where the request times out do to server doing some work that takes longer than expected.

sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
tvaron3: We use our own pipeline due to the unique constraints for availability and latency.  This is following our current architecture.

sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
heaths: Your pipeline should still be similar to the core one. I appreciate why it was necessary and @analogrelay and I discussed this a while ago. I didn't expect it to go off completely on its own.

You'd have to define a field on `ClientMethodOptions` or your own variant of that. We were trying to avoid deeply nesting, but if you have your own pipeline anyway (which can still look similar to the core pipeline) you can have your own `ClientMethodOptions` as well.

This affects the public API which is an architectural concern. Per [*all languages* guidelines](https://azure.github.io/azure-sdk/general_introduction.html), SDKs are idiomatic but consistent. Using Cosmos should feel similar (within reason) to any other service crate.

sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
analogrelay: Correct, we need to be able to trigger things like hedging where we allow the client to try making a fresh request. Our retry policy goes quite far beyond `Retry-After`.

sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
analogrelay: Between this, region routing, hedging, HTTP/2, custom protocols, etc. I just don't think the `azure_core` pipeline abstractions will help us much here, and trying to rearchitect the abstractions so they can work for us seems counter-productive here.

sdk/cosmos/azure_data_cosmos_driver/docs/DYNAMIC_TIMEOUTS_SPEC.md
analogrelay: > Your pipeline should still be similar to the core one.

I recognize that, but there are some pretty significant differences here:

* Cosmos has its own diagnostics payloads used across all SDKs (we still integrate with OTel and the azure_core/tracing stack, but there is much more beyond that)
* The Cosmos backend needs user-agent reporting in a different way (prefix instead of suffix, IIRC)
* We need to deal with HTTP/2 limitations _on the backend_ (20 stream per connection limit means we have to manage multiple connections)
* We have partition-level failover, regional failover, circuit breakers, background cache refreshes, and more that just doesn't fit cleanly into the pipeline as Policy.
* Cosmos "Operations" often involve _several_ HTTP requests with significant shared context. Cross-partition queries, ReadMany, etc.
* We have practical experience with HTTP proxies degrading the Cosmos DB availability guarantees, and need to control their use
* Custom HTTP headers _don't work the way customers would expect_ (they aren't guaranteed to be honoured by the backend when using the newer binary protocol)

We can aim for some consistency here, for sure. We're trying to follow guidelines as closely as we can, but I really want to avoid giving customers the _false_ impression of consistency. Where there is true consistency, we should strive to align.

I think this is all mostly aligned with what we talked about yesterday, but I just wanted to include it here for posterity.

PR #3875, opened 2026-03-06
driver-transport - step 01

FabianMeiswinkel:
See https://github.com/Azure/azure-sdk-for-rust/blob/release/azure_data_cosmos-previews/sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md - this PR implements step 1 only of that spec

### DIFF
--- a/sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md
+++ b/sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md
@@ -1477,7 +1477,7 @@ principle of separating behavior from state — there's no `CosmosHeadersPolicy`
 Cosmos DB exposes **three gateway flavors** that differ in HTTP version support:
 
 | Gateway            | Usage                | HTTP Version                   | Notes                  |
-|--------------------|----------------------|--------------------------------|------------------------|
+| ------------------ | -------------------- | ------------------------------ | ---------------------- |
 | **RoutingGateway** | Metadata + Dataplane | HTTP/1.1 only                  | Legacy, no ALPN        |
 | **ComputeGateway** | Metadata + Dataplane | HTTP/2 via ALPN                | Preferred for metadata |
 | **Gateway 2.0**    | Dataplane only       | HTTP/2 forced (different port) | No HTTP/1.1 fallback   |
@@ -2341,23 +2341,24 @@ Refactor `execute_operation` to use the new pipeline architecture with the absol
 viable transport. No HTTP/2 sharding, no hedging, no circuit breaker, no session consistency —
 just the structural refactoring with a plain `Arc<dyn HttpClient>`.
 
-| Sub-step | Work Item                                                                                                                                                                                                                                                     | Files                                                          |
-|----------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------|
-| 1.1      | **State components** — Define ECS-style component types (`RoutingDecision`, `OperationRetryState`, `TransportRequest`, `TransportResult`, `ThrottleRetryState`, decision enums)                                                                               | `driver/pipeline/components.rs`                                |
-| 1.2      | **Transport pipeline (slim)** — Implement `execute_transport_pipeline` function with `apply_cosmos_headers`, `sign_request`, and 429 throttle retry only. Uses a plain `Arc<dyn HttpClient>` (no `AdaptiveTransport` yet). Hard-coded retry limits + backoff. | `driver/transport/transport_pipeline.rs`                       |
-| 1.3      | **Operation pipeline (slim)** — Implement `execute_operation_pipeline` core loop with single-region endpoint resolution. Only the happy path + 429 retry propagation. No failover, no hedging. Hard-coded deadline.                                           | `driver/pipeline/operation_pipeline.rs`, `retry_evaluation.rs` |
-| 1.4      | **Wire `execute_operation`** — Connect `CosmosDriver::execute_operation` to the new operation pipeline for all operations. The old pipeline code path remains but is no longer called.                                                                        | `driver/cosmos_driver.rs`                                      |
-| 1.5      | **Unit tests** — Tests for each pure function (`evaluate_transport_result` happy path, `evaluate_transport_retry` for 429, `build_transport_request`).                                                                                                        | `tests/`                                                       |
+| Sub-step | Work Item                                                                                                                                                                                                                                                                                                                                                                                                                            | Files                                                          |
+| -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | -------------------------------------------------------------- |
+| 1.1      | **State components** — Define ECS-style component types (`RoutingDecision`, `OperationRetryState`, `TransportRequest`, `TransportResult`, `ThrottleRetryState`, decision enums)                                                                                                                                                                                                                                                      | `driver/pipeline/components.rs`                                |
+| 1.2      | **Transport pipeline (slim)** — Implement `execute_transport_pipeline` function with `apply_cosmos_headers`, `sign_request`, and 429 throttle retry only. Uses a plain `Arc<dyn HttpClient>` (no `AdaptiveTransport` yet). Hard-coded retry limits + backoff.                                                                                                                                                                        | `driver/transport/transport_pipeline.rs`                       |
+| 1.3      | **Operation pipeline (slim)** — Implement `execute_operation_pipeline` core loop with single-region endpoint resolution. Include happy path plus minimal operation-level transport retries: retry `TransportError` only when request is definitely not sent or operation is idempotent and retry budget remains; propagate non-429 HTTP errors as abort (429 remains transport-level). No failover, no hedging. Hard-coded deadline. | `driver/pipeline/operation_pipeline.rs`, `retry_evaluation.rs` |
+| 1.4      | **Wire `execute_operation`** — Connect `CosmosDriver::execute_operation` to the new operation pipeline for all operations. The old pipeline code path remains but is no longer called.                                                                                                                                                                                                                                               | `driver/cosmos_driver.rs`                                      |
+| 1.5      | **Unit tests** — Tests for each pure function (`evaluate_transport_result` happy path + minimal transport-error retry semantics, `evaluate_transport_retry` for 429, `build_transport_request`).                                                                                                                                                                                                                                     | `tests/`                                                       |
 
 **What works after Step 1**: Operations flow through the new pipeline end-to-end against a
-single region with basic 429 retry. The architecture is in place for incremental additions.
+single region with basic 429 retry and minimal safe transport-error retries (not-sent or
+idempotent-only within budget). The architecture is in place for incremental additions.
 
 ### Step 2: Multi-region failover & endpoint management
 
 Add cross-region failover, `AccountEndpointState` + routing systems, and the `AccountMetadataCache` integration.
 
 | Sub-step | Work Item                                                                                                                                                                                                                                                                                      | Files                                                                                                 |
-|----------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------|
+| -------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------- |
 | 2.1      | **Routing state & systems** — Implement `AccountEndpointState`, the unified `LocationStateStore` (§4.6), and the system functions (`build_account_endpoint_state`, `mark_endpoint_unavailable`, `expire_unavailable_endpoints`). Wire to existing `AccountMetadataCache`.                      | `driver/routing/mod.rs`, `account_endpoint_state.rs`, `location_state_store.rs`, `routing_systems.rs` |
 | 2.2      | **Expand `evaluate_transport_result`** — Add 403.3 (WriteForbidden + cache refresh), 503, 404/1022, 429/3092, 500-for-reads. Return `(OperationAction, Vec<LocationEffect>)` tuples; add `FailoverRetry`, `SessionRetry` action handling and STAGE 6 effect application in the operation loop. | `driver/pipeline/retry_evaluation.rs`, `operation_pipeline.rs`                                        |
 | 2.3      | **Deadline enforcement** — Implement active deadline enforcement in transport pipeline (per-request timeout clamping, stream drop).                                                                                                                                                            | `driver/transport/transport_pipeline.rs`                                                              |
@@ -2370,7 +2371,7 @@ session retry, deadline enforcement. Still HTTP/1.1, no hedging, no circuit brea
 ### Step 3: Session consistency & partition-level circuit breaker
 
 | Sub-step | Work Item                                                                                                                                                                                                                                                                                                                                                                                                   | Files                                                                                                 |
-|----------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------|
+| -------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------- |
 | 3.1      | **Session tracking** — Session token management (resolve, propagate, track LSN).                                                                                                                                                                                                                                                                                                                            | `driver/routing/session_manager.rs`                                                                   |
 | 3.2      | **Partition endpoint state & circuit breaker** — Implement `PartitionEndpointState`, wire it into the existing `LocationStateStore` (§4.6), add `CircuitBreakerConfig` (resolved from `CircuitBreakerOptions` §9.4), and the system functions (`mark_partition_unavailable`, `record_partition_failure/success`, `sweep_partition_health`). Hard-coded default thresholds (read: 2, write: 5, reset: 5min). | `driver/routing/partition_endpoint_state.rs`, `location_state_store.rs`, `circuit_breaker_systems.rs` |
 | 3.3      | **`ReadConsistencyStrategy`** — Wire `effective_read_consistency_strategy` into `SessionState` and endpoint resolution.                                                                                                                                                                                                                                                                                     | `driver/pipeline/components.rs`, `operation_pipeline.rs`                                              |
@@ -2383,7 +2384,7 @@ with failback. Still HTTP/1.1, no hedging.
 ### Step 4: Hedging (speculative execution)
 
 | Sub-step | Work Item                                                                                                                                                                    | Files                        |
-|----------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------|
+| -------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------- |
 | 4.1      | **Hedging implementation** — `execute_hedged` with `tokio::select!` racing. Initial/Hedging `ExecutionContext` tracking. Static threshold first.                             | `driver/pipeline/hedging.rs` |
 | 4.2      | **Dynamic threshold** — P99 latency tracker with safety gates (50–4000 ms).                                                                                                  | `driver/pipeline/hedging.rs` |
 | 4.3      | **Hedging config** — `HedgingThreshold` enum (Dynamic/Static), `HedgingOptions` with `enabled` flag (§9.3). Nested in `RetryOptions`. Overridable at Runtime/Account layers. | `options/availability.rs`    |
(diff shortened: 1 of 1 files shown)


sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/components.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// The HTTP method.
+    pub method: Method,
+    /// The fully resolved URL for this attempt.
+    pub url: Url,
+    /// The resolved region for this attempt, if known.
+    pub region: Option<Region>,
+    /// Headers to send (includes operation-specific and attempt-specific headers).
+    pub headers: Headers,
+    /// Request body bytes (schema-agnostic).
+    pub body: Option<Vec<u8>>,
```
xinlian12: **🟡 Recommendation** · Performance: Memory Allocation

**`body: Option<Vec<u8>>` causes O(n) full-copy on every 429 throttle retry**

The transport pipeline loops up to 9 times for 429 retries. Each iteration clones the body via `request.body.clone()` to build the HTTP request. With `Vec<u8>`, every clone is a full heap allocation + memcpy. For a 256 KB document upsert being throttled, that's 9 × 256 KB ≈ 2.3 MB of unnecessary allocation and copying per operation.

The spec explicitly defines this field as `Option<Bytes>`. `azure_core::Bytes` (which wraps `bytes::Bytes`) is reference-counted — clone is O(1), just an Arc increment.

```suggestion
    pub body: Option<azure_core::Bytes>,
```

Then update the construction site in `build_transport_request` to:
```rust
body: operation.body().map(azure_core::Bytes::from),
```

and the usage in `transport_pipeline.rs` to:
```rust
if let Some(body) = &request.body {
    http_request.set_body(body.clone());
}
```

---

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/components.rs
```diff
@@ -0,0 +1,268 @@
...
+            max_wait_time: DEFAULT_MAX_THROTTLE_WAIT,
+            max_per_retry_delay: DEFAULT_MAX_PER_RETRY_DELAY,
+            fallback_base_delay: DEFAULT_FALLBACK_BASE_DELAY,
+            backoff_factor: DEFAULT_BACKOFF_FACTOR,
+        }
+    }
+
+    /// Computes the fallback delay (used when the service does not send
+    /// `x-ms-retry-after-ms`). Uses exponential backoff from a small base.
+    pub fn fallback_delay(&self) -> Duration {
```
xinlian12: **🟡 Recommendation** · Reliability: Thundering Herd

**No jitter in throttle backoff creates synchronized retry waves**

When many concurrent requests are throttled simultaneously (common during bulk ingestion), they all compute identical retry delays (5ms, 10ms, 20ms…) and retry at exactly the same instants. This creates synchronized retry waves that re-trigger throttling — the classic "thundering herd" problem. The Java Cosmos SDK uses jittered backoff to mitigate this.

Consider adding ±25% jitter. If pulling in `rand` as a dependency is undesirable, a lightweight approach is to derive jitter from the current `Instant` nanos:

```rust
pub fn fallback_delay(&self) -> Duration {
    let multiplier = self.backoff_factor.powi(self.attempt_count as i32);
    let base = self.fallback_base_delay.mul_f64(multiplier);
    // Add ~±25% jitter to avoid synchronized retry waves.
    let nanos = std::time::Instant::now().elapsed().subsec_nanos();
    let jitter_factor = 0.75 + (nanos % 500) as f64 / 1000.0; // range [0.75, 1.25)
    base.mul_f64(jitter_factor)
}
```

---

xinlian12: In java SDK, we actually do use a random salt at least for MetadataThrottlingRetryPolicy

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
```diff
@@ -0,0 +1,440 @@
...
+        return ThrottleAction::Propagate;
+    }
+
+    if throttle_state.attempt_count >= throttle_state.max_attempts {
+        return ThrottleAction::Propagate;
+    }
+
+    // Extract the service-specified retry delay from response headers,
+    // or fall back to exponential backoff.
+    let service_delay = result
```
xinlian12, line 108: **💬 QQ** · Cross-SDK Parity: Fallback Delay

**Fallback delay logic diverges from the Java SDK — intentional?**

The Java SDK's `ResourceThrottleRetryPolicy.checkIfRetryNeeded()` handles the case where the server doesn't provide `x-ms-retry-after-ms` differently:

- **429/3200** (`USER_REQUEST_RATE_TOO_LARGE`): retry **immediately** (0ms) — the server *should have* returned a header; its absence means retry is safe
- **Other 429 sub-statuses** (e.g., 3089): wait a fixed **100ms**
- **No exponential backoff** on the fallback — the `backoffDelayFactor` (default 1) multiplies the *server-provided* delay, not the fallback

The Rust implementation uses **5ms exponential backoff** (5→10→20→40→80→160→320→640→1280ms) for *all* 429s with no sub-status distinction.

Are these intentional design decisions for the Rust SDK, or should they align with the Java behavior?

---

xinlian12 (commented):
Quick question on Java SDK parity for the throttle retry fallback logic.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
FabianMeiswinkel: That is wrong - Java also enforces a 5s max - see https://github.com/Azure/azure-sdk-for-java/blob/8f40a062ad49c93128d62bf0df7c76da57bb4a31/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosException.java#L367

For 429/3200 without x-ms-retry-after having an exponential backoff starting with low durationis intentional - the no delay can result in really bad floods of retries

For other 429 sub-status codes I am not against using differences but would want to know why - we can iterate on this

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/components.rs
FabianMeiswinkel: Fixed by adding a jitter of 25% as well

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
FabianMeiswinkel: I am not plannign to make changes on this at this point - but am leaving the comment open to get more eyes on it to see whether others agree.

sdk/cosmos/azure_data_cosmos_driver/src/driver/mod.rs
```diff
@@ -9,11 +9,58 @@
...
+    // SplitMix64 mixing: quickly reduces correlation in nearby seeds into a well-distributed
+    // 64-bit value suitable for non-cryptographic jitter generation.
+    //
+    // We then normalize to [0.0, 1.0] so callers can scale jitter bands directly.
+    x = x.wrapping_add(0x9E37_79B9_7F4A_7C15);
+    x = (x ^ (x >> 30)).wrapping_mul(0xBF58_476D_1CE4_E5B9);
+    x = (x ^ (x >> 27)).wrapping_mul(0x94D0_49BB_1331_11EB);
+    x ^= x >> 31;
+    (x as f64) / (u64::MAX as f64)
+}
```
analogrelay: I'd go ahead and put this in a `jitter` module

FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
```diff
@@ -0,0 +1,441 @@
...
+
+    let delay = service_delay.unwrap_or_else(|| throttle_state.fallback_delay());
+
+    // Cap individual retry delay to avoid excessive waits.
+    let delay = delay.min(throttle_state.max_per_retry_delay);
+
+    let new_cumulative = throttle_state.cumulative_delay + delay;
+
+    if new_cumulative > throttle_state.max_wait_time {
+        return ThrottleAction::Propagate;
```
FabianMeiswinkel, line 122: Should we instead retry one more time - question is whether immediately or with a backoff (but how large - would be smaller than service-provided). Chekc with Java/.Net on how this is done today.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
```diff
@@ -0,0 +1,240 @@
...
+) -> OperationAction {
+    // Happy path: return the whole TransportResult without destructuring.
+    if matches!(&result.outcome, TransportOutcome::Success { .. }) {
+        return OperationAction::Complete(result);
+    }
+
+    // Error paths: destructure the owned outcome to move error values out
+    // without losing the error source chain.
+    match result.outcome {
+        TransportOutcome::Success { .. } => unreachable!("handled above"),
```
analogrelay: This feels unnecessary to me. The `if matches!` can be collapsed into this arm of the match. It doesn't mean destructuring, you can use the pattern binding syntax `outcome @ TransportOutcome::Success { .. } => return OperationAction::Complete(outcome)`

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
```diff
@@ -0,0 +1,441 @@
...
+
+                throttle_state = new_state;
+                continue;
+            }
+            ThrottleAction::Propagate => return result,
+        }
+    }
+}
+
+fn deadline_exceeded_result(request_sent: RequestSentStatus) -> TransportResult {
```
analogrelay, line 314: nit: I'd usually have these as constructor-style functions on `TransportResult`, e.g. `TransportResult::deadline_exceeded(request_sent: RequestSentStatus)`, but that's not critical.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
```diff
@@ -0,0 +1,441 @@
...
+                azure_core::error::ErrorKind::Other,
+                "end-to-end operation timeout exceeded",
+            ),
+            request_sent,
+        },
+    }
+}
+
+/// Maps an HTTP response to a `TransportResult`.
+async fn map_http_response(
```
analogrelay, line 319: And I'd probably have as `TransportResult::from_http_response`, but again, not critical

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
FabianMeiswinkel: True - Fixed.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
FabianMeiswinkel: instead of failing fast we should do one more retry after math.max(0, e2etimeout-duration-100ms)

FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
```diff
@@ -0,0 +1,473 @@
...
+        let request_handle = diagnostics.start_request(
+            execution_context,
+            pipeline_type,
+            transport_security,
+            region,
+            endpoint_string,
+        );
+
+        // Build HTTP request from TransportRequest
+        let mut http_request = Request::new(request.url.clone(), request.method);
```
FabianMeiswinkel, line 180: Critical sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs:146
[execute_transport_pipeline](vscode-file://vscode-app/c:/Users/fabianm/AppData/Local/Programs/Microsoft%20VS%20Code/0870c2a0c7/resources/app/out/vs/code/electron-browser/workbench/workbench.html) builds and sends [http_request](vscode-file://vscode-app/c:/Users/fabianm/AppData/Local/Programs/Microsoft%20VS%20Code/0870c2a0c7/resources/app/out/vs/code/electron-browser/workbench/workbench.html) without a per-request timeout derived from the operation deadline.
Why this matters: Step 1 spec explicitly calls out clamping request timeout to remaining E2E budget (see sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md:317 and sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md:1388). Current logic only checks deadline before/after attempts (sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs:121 and sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs:233), so a single slow in-flight send/receive can run past deadline.
Recommended fix: add transport-level request timeout plumbing (likely via [azure_core](vscode-file://vscode-app/c:/Users/fabianm/AppData/Local/Programs/Microsoft%20VS%20Code/0870c2a0c7/resources/app/out/vs/code/electron-browser/workbench/workbench.html)/[typespec_client_core](vscode-file://vscode-app/c:/Users/fabianm/AppData/Local/Programs/Microsoft%20VS%20Code/0870c2a0c7/resources/app/out/vs/code/electron-browser/workbench/workbench.html) request options) and apply [max(1ms, deadline - now)](vscode-file://vscode-app/c:/Users/fabianm/AppData/Local/Programs/Microsoft%20VS%20Code/0870c2a0c7/resources/app/out/vs/code/electron-browser/workbench/workbench.html) per attempt before execute_request.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
```diff
@@ -0,0 +1,473 @@
...
+                    effective_delay = deadline_capped_delay(effective_delay, remaining);
+                }
+
+                azure_core::sleep(
+                    azure_core::time::Duration::try_from(effective_delay)
+                        .unwrap_or(azure_core::time::Duration::ZERO),
+                )
+                .await;
+
+                if let Some(deadline) = request.deadline {
```
FabianMeiswinkel, line 276: Major sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs:236
The newly added forced final throttling retry path is not covered by tests. Current tests validate [evaluate_transport_retry](vscode-file://vscode-app/c:/Users/fabianm/AppData/Local/Programs/Microsoft%20VS%20Code/0870c2a0c7/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [deadline_capped_delay](vscode-file://vscode-app/c:/Users/fabianm/AppData/Local/Programs/Microsoft%20VS%20Code/0870c2a0c7/resources/app/out/vs/code/electron-browser/workbench/workbench.html) math only (sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs:371, sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs:457), but not the branch that does one extra retry after [Propagate](vscode-file://vscode-app/c:/Users/fabianm/AppData/Local/Programs/Microsoft%20VS%20Code/0870c2a0c7/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
Why this matters: this is behaviorally significant and easy to regress (especially around near-deadline and [final_delay == 0](vscode-file://vscode-app/c:/Users/fabianm/AppData/Local/Programs/Microsoft%20VS%20Code/0870c2a0c7/resources/app/out/vs/code/electron-browser/workbench/workbench.html) paths).
Recommended fix: add an async unit test with a mock/fake HttpClient that returns throttles and verifies the extra attempt count and deadline behavior.

FabianMeiswinkel: Should the “one final retry” apply only when [request.deadline](vscode-file://vscode-app/c:/Users/fabianm/AppData/Local/Programs/Microsoft%20VS%20Code/0870c2a0c7/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is set, or also when no deadline exists? Current code only does it under [Some(deadline)](vscode-file://vscode-app/c:/Users/fabianm/AppData/Local/Programs/Microsoft%20VS%20Code/0870c2a0c7/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs:242).

FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
FabianMeiswinkel: Fixed

simorenoh (approved):

sdk/cosmos/azure_data_cosmos_driver/src/options/policies.rs
```diff
@@ -47,8 +49,13 @@ pub struct EndToEndOperationLatencyPolicy {
 
 impl EndToEndOperationLatencyPolicy {
```
FabianMeiswinkel, line 50: **Issue 1: Timeout Floor Contract Mismatch (1s vs 1ms)**
File refs:

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs:42
sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs:49
sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs:515
What is happening:

The transport pipeline clamps remaining_request_timeout to a minimum of 1 second.
A test explicitly expects the minimum to be 1 millisecond, and fails accordingly.
The Step 1 spec intent is “never zero/negative timeout, but allow a very small floor,” which aligns much more with 1ms than 1s.
Why this is important:

This is not just a test naming mismatch. It changes runtime behavior significantly.
With 1s floor, a request can run much longer than expected after timeout budget depletion.
In retry-heavy scenarios, this can increase tail latency and violate higher-level operation timeout expectations.
Likely root cause:

Either an intentional “safety guard” was set too high, or spec/test and implementation drifted during refactoring.
Correct direction:

Pick one contract and enforce it across code + tests + spec comments.
Based on current test name and Step 1 language, 1ms is likely the intended behavior.

FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
```diff
@@ -0,0 +1,522 @@
...
+    user_agent: &azure_core::http::headers::HeaderValue,
+    pipeline_type: PipelineType,
+    transport_security: TransportSecurity,
+    diagnostics: &mut DiagnosticsContextBuilder,
+) -> TransportResult {
+    let mut throttle_state = ThrottleRetryState::new();
+    let mut forced_final_throttle_retry = false;
+
+    loop {
+        // Check deadline before each attempt
```
FabianMeiswinkel, line 154: Issue 2: Retry Budget Bypass via Forced Final Throttle Retry
File refs:

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs:140
sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs:280
sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs:281
sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs:292
What is happening:

The retry evaluator can return Propagate when limits are reached (max attempts, max cumulative wait, etc.).
After that, pipeline logic still allows an additional “forced final” throttle retry in some paths.
That means the hard limits are not truly hard.
Why this is important:

Retry budgets are usually part of correctness, not just optimization.
Exceeding them can cause unexpected RU usage, additional pressure on hot partitions, and harder-to-reason operational behavior.
It also makes telemetry misleading because actual attempts can exceed policy limits.
Likely root cause:

A “last chance” reliability pattern was added but not integrated into policy accounting.
Correct direction:

Either remove forced-final behavior entirely.
Or explicitly model it as policy and include it in the counted budget so the cap remains deterministic.

FabianMeiswinkel: Fixed

PR #3879, opened 2026-03-07
Performance enhancements to storage client and added a bunch more storage performance tests to measure different options.

LarryOsterman:
* Added `AsyncResponseBody::collect_into` to allow callers to receive the body data into a buffer they supply.
  * Added test cases for `collect_into`
* Rewrote `AsyncResponseBody::collect` to collect the received `Bytes` into a vector and then copy the received `Bytes` into a single allocated buffer rather than extending an empty buffer.

Also:

* Updated rust performance tests to allow type safe parameters to tests.
* Added a bunch of blob download performance tests.
* Added perf test runners for a number of different variants of collecting data for downloads.

### DIFF
--- a/sdk/core/azure_core_test/src/perf/mod.rs
+++ b/sdk/core/azure_core_test/src/perf/mod.rs
@@ -63,14 +63,14 @@ pub struct PerfTestMetadata {
     pub create_test: CreatePerfTestFn,
 }
 
-/// The expected type of a test option, used in [`PerfRunner.try_get_test_arg`] and [`PerfRunner::try_get_global_arg`]
+/// The expected type of a test option, used in [`PerfRunner::try_get_test_arg`] and [`PerfRunner::try_get_global_arg`]
 ///
 /// This allows a test author to declare the expected numeric type of the option value, which
 /// simplifies the work involved in processing a test option value and reduces the chance of errors in that processing.
 ///
 /// The default option type is `String`,
 #[derive(Debug, Clone, Default)]
-pub enum TestOptionKind {
+pub enum PerfTestOptionKind {
     // Note: We need this type because `clap` requires us to specify the expected type of each argument
     // for proper parsing, and this allows us to leverage that parsing in `try_get_test_arg` and
     // `try_get_global_arg` to get typed arguments without needing to do any additional parsing
@@ -115,7 +115,7 @@ pub struct PerfTestOption {
     pub sensitive: bool,
 
     /// The expected type of the argument value.
-    pub option_type: TestOptionKind,
+    pub option_type: PerfTestOptionKind,
 }
 
 #[derive(Debug, Clone, Default, Serialize)]
@@ -582,17 +582,17 @@ impl PerfRunner {
                     .required(option.mandatory)
                     .global(false);
                 arg = match option.option_type {
-                    TestOptionKind::String => arg.value_parser(clap::value_parser!(String)),
-                    TestOptionKind::Usize => arg.value_parser(clap::value_parser!(usize)),
-                    TestOptionKind::Int8 => arg.value_parser(clap::value_parser!(i8)),
-                    TestOptionKind::Int16 => arg.value_parser(clap::value_parser!(i16)),
-                    TestOptionKind::Int32 => arg.value_parser(clap::value_parser!(i32)),
-                    TestOptionKind::Int64 => arg.value_parser(clap::value_parser!(i64)),
-                    TestOptionKind::Uint8 => arg.value_parser(clap::value_parser!(u8)),
-                    TestOptionKind::Uint16 => arg.value_parser(clap::value_parser!(u16)),
-                    TestOptionKind::Uint32 => arg.value_parser(clap::value_parser!(u32)),
-                    TestOptionKind::Uint64 => arg.value_parser(clap::value_parser!(u64)),
-                    TestOptionKind::Boolean => {
+                    PerfTestOptionKind::String => arg.value_parser(clap::value_parser!(String)),
+                    PerfTestOptionKind::Usize => arg.value_parser(clap::value_parser!(usize)),
+                    PerfTestOptionKind::Int8 => arg.value_parser(clap::value_parser!(i8)),
+                    PerfTestOptionKind::Int16 => arg.value_parser(clap::value_parser!(i16)),
+                    PerfTestOptionKind::Int32 => arg.value_parser(clap::value_parser!(i32)),
+                    PerfTestOptionKind::Int64 => arg.value_parser(clap::value_parser!(i64)),
+                    PerfTestOptionKind::Uint8 => arg.value_parser(clap::value_parser!(u8)),
+                    PerfTestOptionKind::Uint16 => arg.value_parser(clap::value_parser!(u16)),
+                    PerfTestOptionKind::Uint32 => arg.value_parser(clap::value_parser!(u32)),
+                    PerfTestOptionKind::Uint64 => arg.value_parser(clap::value_parser!(u64)),
+                    PerfTestOptionKind::Boolean => {
                         // For boolean options, we can use the presence of the flag to indicate true, and absence to indicate false.
                         // Therefore, we don't need to specify a value parser or expect any arguments for this type.
                         arg.action(clap::ArgAction::SetTrue).num_args(0)
--- a/sdk/storage/azure_storage_blob/perf/download_blob_test.rs
+++ b/sdk/storage/azure_storage_blob/perf/download_blob_test.rs
@@ -10,7 +10,7 @@ use azure_core::{error::ErrorKind, Bytes};
 use azure_core_test::{
     perf::{
         CreatePerfTestReturn, PerfRunner, PerfTest, PerfTestMetadata, PerfTestOption,
-        TestOptionKind,
+        PerfTestOptionKind,
     },
     TestContext,
 };
@@ -81,7 +81,7 @@ impl DownloadBlobTest {
                     short_activator: Some('c'),
                     long_activator: "count",
                     expected_args_len: 1,
-                    option_type: TestOptionKind::Uint32,
+                    option_type: PerfTestOptionKind::Uint32,
                     ..Default::default()
                 },
                 PerfTestOption {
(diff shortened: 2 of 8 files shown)


.waza.yaml
```diff
@@ -0,0 +1,31 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/microsoft/waza/main/schemas/config.schema.json
+
+paths:
+  skills: .github/skills/
+  evals: .github/evals/
+  results: .github/results/
```
heaths: 1. Why is there unrelated stuff in here?
2. Why not put these under the skills under tests so we don't have this cognitive disconnect and have to go fishing?

LarryOsterman: Because this is a draft PR and it has a whole lot of stuff that doesn't belong here. There was a bunch of stuff swept up in this PR that shouldn't have been.

Generally, I treat "draft" PRs as "don't bother reviewing yet".

I can update the root waza.yaml file to put them somewhere else.

heaths: Oh, I know. The title was a dead giveaway too. :smile: But I was curious since we've been talking about some ideas to see what you had so far, and just thought I'd "pre-comment" that these files seem 1) separate from the PR, and 2) I'd like to see organized a little different so they are closer to their "skills under test" (the other SUTs!).

LarryOsterman: Fair - but in this case, I hadn't done my "is this really the changes I want to commit" pass over the PR and honestly didn't even realize they were there. I'm actually pretty close on this one - it's a VERY small set of changes, but it seems to work. And those files were dropped there by waza and were very much a work-in-progress.

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -237,15 +259,17 @@ impl PerfRunner {
...
-            args.try_get_one::<T>(id).with_context(
-                ErrorKind::Other,
-                format!("Failed to get argument '{}' for test.", id),
-            )
+            args.try_get_one::<T>(id)
+                .with_context(
+                    ErrorKind::Other,
+                    format!("Failed to get argument '{}' for test.", id),
+                )
+                .map(|arg| arg.cloned())
```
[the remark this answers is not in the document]
LarryOsterman: The `clap` `try_get_one` method returns an `Option<&T>` the `cloned()` call dereferences the `T` parameter.

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -63,6 +63,22 @@ pub struct PerfTestMetadata {
     pub create_test: CreatePerfTestFn,
 }
 
+#[derive(Debug, Clone, Default)]
+pub enum TestOptionType {
```
heaths: What's this for? Can you add a doc comment? Top-down, I've seen it used in a couple places and still am not clear what it's for. Maybe further below...

Still, a doc comment would help now and in the future when people use it.

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -86,6 +102,9 @@ pub struct PerfTestOption {
 
     /// Argument value is sensitive and should be sanitized.
     pub sensitive: bool,
+
+    /// The type of the argument value.
+    pub option_type: TestOptionType,
```
heaths: Ok, starting to make some sense. But wouldn't it be idiomatic to just make the option (argument) itself an enum? It could implement `FromStr` and just pass that onto each type's `FromStr` impl.

sdk/core/typespec_client_core/src/http/clients/reqwest.rs
```diff
@@ -24,6 +24,8 @@ pub fn new_reqwest_client() -> Arc<dyn HttpClient> {
     // Due to the significant performance impact when disabling connection pooling,
     // it is enabled here by default. See the `azure_core` troubleshooting guide to disable pooling.
     let client = ::reqwest::ClientBuilder::new()
+        .redirect(::reqwest::redirect::Policy::none())
+        .retry(reqwest::retry::never())
```
heaths: Should add a comment - probably just blend with the one above - as to *why* we're doing this. Not only would that help future maintainers, but LLMs as well.

sdk/core/typespec_client_core/src/http/response.rs
```diff
@@ -312,15 +313,51 @@ impl AsyncResponseBody {
...
+            let res = res?;
+            total_length += res.len();
+            bytes.push(res);
+        }
+        let mut final_result = BytesMut::with_capacity(total_length);
+        for b in bytes {
+            final_result.extend(&b);
         }
 
         Ok(final_result.into())
```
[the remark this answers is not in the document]
heaths: Copilot might have a good suggestion here, but I'm also wondering: does Storage send a `content-length` header for chunked data so we get a final size? Might not hurt to check, or maybe that's more of a case where they just use `collect_into()`. Still, might not hurt to check if it's there. We could prealloc a `BytesMut`.

sdk/storage/azure_storage_blob/perf/download_blob_test.rs
```diff
@@ -0,0 +1,273 @@
...
+                    expected_args_len: 1,
+                    ..Default::default()
+                },
+            ],
+            create_test: Self::create_download_blob_test,
+        }
+    }
+
+    /// This method represents the most basic way to download a blob, where we simply stream the contents and do nothing with them. This is useful for testing the performance of streaming downloads without any additional overhead.
+    async fn collect_stream(&self, blob_client: BlobClient) -> azure_core::Result<()> {
```
heaths, line 122: Should this and other tests use the `black_box` to avoid eliding code since you just drop the `result`? Note: it's in `std` stable as of 1.86: `std::hint::black_box`, so no use of `criterion`'s necessary.

sdk/core/azure_core_test/src/perf/mod.rs
LarryOsterman: Done. Basically it allows for strong typing the test arguments.

sdk/core/azure_core_test/src/perf/mod.rs
LarryOsterman: Interesting thought. Let me see what that looks like - I modeled this after `clap`'s option, but your idea has merit.

sdk/storage/azure_storage_blob/perf/download_blob_test.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    /// This method collects the entire blob into memory in 512K chunks using the `collect_into` method on the body
+    async fn collect_into(&self, blob_client: BlobClient) -> azure_core::Result<()> {
+        let response = blob_client.download(None).await?;
+
+        let mut buffer = vec![0u8; 512 * 1024]; // 512 KB buffer
+        response.into_body().collect_into(&mut buffer).await?;
+        black_box(buffer);
+        Ok(())
```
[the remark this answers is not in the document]
LarryOsterman: Actually this will download the first 512KiB of the data and ignore the rest.

sdk/storage/azure_storage_blob/perf/download_blob_test.rs
LarryOsterman: Yes they should - I correctly use `black_box` in the other tests but missed this one.

sdk/storage/azure_storage_blob/perf/download_blob_test.rs
```diff
@@ -0,0 +1,270 @@
...
+                CollectOptions::Simple => {
+                    self.collect_blob_simple(blob_client).await?;
+                }
+                CollectOptions::VecBytes => {
+                    self.collect_blob_bytes_mut(blob_client).await?;
+                }
+                CollectOptions::Into => {
+                    self.collect_into(blob_client).await?;
+                }
+            }
```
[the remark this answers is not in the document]
LarryOsterman: CollectOptions is a simple enum - POD and thus implicitly implements `Copy`, and `Clone`. And more importantly, this compiles just fine.

sdk/core/azure_core_test/src/perf/mod.rs
```diff
@@ -63,6 +63,28 @@ pub struct PerfTestMetadata {
...
 }
 
+/// The expected type of a test option, used in [`PerfRunner.try_get_test_arg`] and [`PerfRunner::try_get_global_arg`]
+///
+/// This allows a test author to declare the expected numeric type of the option value, which
+/// simplifies the work involved in processing a test option value and reduces the chance of errors in that processing.
+///
+/// The default option type is `String`,
+#[derive(Debug, Clone, Default)]
+pub enum TestOptionType {
```
heaths: Small nit, and don't even bother unless 1) you really want to, or 2) there are other changes and you don't hate this suggestion: To associate it with `PerfTestOption` - and, separately, because we tend to use "kind" or "type" across languages including Rust - you might consider `PerfTestOptionKind` instead.

sdk/core/typespec_client_core/src/http/response.rs
```diff
@@ -312,15 +313,51 @@ impl AsyncResponseBody {
 
     /// Collect the stream into a [`Bytes`] collection.
     pub async fn collect(mut self) -> crate::Result<Bytes> {
-        let mut final_result = Vec::new();
+        let mut bytes = Vec::<Bytes>::new();
```
heaths, line 320: Only because we just talked about it offline re: another PR, I wonder if you should add a comment here to the effect that you found this algorithm was cheaper than just managing a single `BytesMut` for future maintainers including LLMs.

sdk/core/typespec_client_core/src/http/response.rs
```diff
@@ -312,15 +313,51 @@ impl AsyncResponseBody {
...
+    ///   if there was an error reading from the stream.
+    ///
+    pub async fn collect_into(mut self, buffer: &mut [u8]) -> crate::Result<usize> {
+        let mut total_copied = 0usize;
+        while let Some(res) = self.next().await {
+            let bytes = res?;
+            // If the remaining space in the buffer won't hold this chunk, fail.
+            if buffer.len() - total_copied < bytes.len() {
+                return Err(crate::Error::with_message(
+                    ErrorKind::Other,
```
heaths: Think maybe this should be `ErrorKind::Io`?

heaths (approved):
Couple thoughts, but nothing blocking; otherwise, LGTM. Nice to see how much this helps!

sdk/core/typespec_client_core/src/http/response.rs
LarryOsterman: Absolutely.

heaths (approved):

PR #3880, opened 2026-03-07
Cosmos: Adds Per Partition Automatic Failover and Circuit Breaker Specs

kundadebdatta:
## Description

Introduces the **Per-Partition Automatic Failover (PPAF) & Per-Partition Circuit Breaker (PPCB)** design spec for `azure_data_cosmos_driver`.

This spec describes partition-level failover mechanisms that complement the existing account-level failover in the driver's 7-stage operation pipeline. Instead of marking an entire region unavailable when a single partition becomes unhealthy, only the affected partition is routed to an alternate region — preserving local latency for healthy partitions.

### What's in the spec

- **Two complementary mechanisms**:
  - **PPAF** — per-partition failover for writes on single-master accounts, triggered by 403/3, 503, 429/3092, 410
  - **PPCB** — per-partition circuit breaker for reads (any account) and writes on multi-master accounts, threshold-gated
- **Component design**: `PartitionEndpointState`, `PartitionFailoverEntry`, `PartitionFailoverConfig` — all managed via the driver's existing lock-free CAS pattern (no `RwLock<HashMap>` like the SDK)
- **Operation pipeline integration**: How partition-level overrides plug into `resolve_endpoint()` (Stage 2), `evaluate_transport_result()` (Stage 5), and `LocationStateStore::apply()` (Stage 6)
- **Background failback loop**: Periodic sweep that expires stale partition overrides, spawned via `BackgroundTaskManager` (#3945)
- **Status code handling matrix**: Complete mapping of HTTP status/sub-status codes to emitted `LocationEffect`s
- **Configuration surface**: All thresholds and intervals configurable via environment variables
- **Test coverage plan**: Pure routing system tests, eligibility tests, circuit breaker counter tests, integration tests, and end-to-end operation loop tests
- **Prerequisites**: Missing pieces that must be implemented (partition key range ID availability, `ResourceType.is_partitioned()`, env var reading, `sync_account_properties` integration)

### Key design decisions

| Decision | Rationale |
|----------|-----------|
| Immutable CAS snapshots (not `RwLock<HashMap>`) | Follows driver's existing lock-free pattern; eliminates reader/writer contention on hot path |
| Two separate maps (PPAF vs PPCB) | Avoids cross-contamination between single-master write failover and multi-master circuit breaker routing strategies |
| Plain counters (not `AtomicI32`) | Entire `PartitionEndpointState` is swapped atomically via CAS — no need for interior atomic counters |
| Failback sweeps both maps | Improvement over SDK which only sweeps PPCB; trivial with immutable-snapshot pattern |
| `BackgroundTaskManager` for failback loop | Provides abort-on-drop, panic safety, and graceful shutdown (#3945) |
| Acceptable CAS counter loss under contention | Delays threshold trigger by at most one failure — better trade-off than introducing locks |

### Dependencies

- #3945 — `BackgroundTaskManager` (must merge first; spec references it for failback loop spawning)

### Files

| File | Action |
|------|--------|
| `azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md` | **New** |

### DIFF
--- a/sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
+++ b/sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
@@ -194,7 +194,7 @@ same lock-free pattern.
 
 > **Configuration resolution**: The PPCB option is read from `DriverOptions` at
 > construction time and stored in `PartitionFailoverConfig`. When the
-> [Hierarchical Configuration Model](../../azure_data_cosmos/docs/HierarchicalConfigModel.md) lands, this will
+> [Hierarchical Configuration Model](https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/cosmos/azure_data_cosmos/docs/HierarchicalConfigModel.md) lands, this will
 > be read from the layered options system (Environment → Runtime → Account →
 > Request). Until then, the `DriverOptions` value is backed by `env::var`.
 


sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+
+The following table maps each status code to its behavior with respect to PPAF/PPCB:
+
+| Status Code | Sub-Status | PPAF Action | PPCB Action | Account-Level Action |
+|---|---|---|---|---|
+| 403 | 3 (WriteForbidden) | Mark partition unavailable → retry | Increment counter → if threshold exceeded, mark unavailable → retry | Fallback: `should_retry_on_endpoint_failure(force_refresh: true)` |
+| 503 | Any | Mark partition unavailable | Mark partition unavailable | `should_retry_on_unavailable_endpoint_status_codes()` |
+| 429 | 3092 (SystemResourceUnavailable) | Mark partition unavailable | Mark partition unavailable | Treated as 503 for write requests on multi-master |
+| 410 | 1022 (LeaseNotFound) | Mark partition unavailable | Mark partition unavailable | `should_retry_on_unavailable_endpoint_status_codes()` |
+| 404 | 1022 (ReadSessionNotAvailable) | N/A | N/A | `should_retry_on_session_not_available()` |
```
FabianMeiswinkel: **🔴 Blocking** · Correctness: Wrong Substatus Code `[Already posted by @Copilot — Unresolved]`

**`404 | 1022` should be `404 | 1002` for ReadSessionNotAvailable**

The matrix row says `404 | 1022 (ReadSessionNotAvailable)`, but in the actual codebase `READ_SESSION_NOT_AVAILABLE` is substatus **1002** (aliased as `PARTITION_KEY_RANGE_GONE` at `constants.rs` line 438). Substatus **1022** is `LEASE_NOT_FOUND`.

Anyone implementing the driver crate from this spec would use the wrong substatus code, causing session-not-available handling to be missed entirely.

Reinforcing @Copilot's existing comment — this is still unresolved.

---

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+## 10. Status Code Handling Matrix
+
+The following table maps each status code to its behavior with respect to PPAF/PPCB:
+
+| Status Code | Sub-Status | PPAF Action | PPCB Action | Account-Level Action |
+|---|---|---|---|---|
+| 403 | 3 (WriteForbidden) | Mark partition unavailable → retry | Increment counter → if threshold exceeded, mark unavailable → retry | Fallback: `should_retry_on_endpoint_failure(force_refresh: true)` |
+| 503 | Any | Mark partition unavailable | Mark partition unavailable | `should_retry_on_unavailable_endpoint_status_codes()` |
+| 429 | 3092 (SystemResourceUnavailable) | Mark partition unavailable | Mark partition unavailable | Treated as 503 for write requests on multi-master |
+| 410 | 1022 (LeaseNotFound) | Mark partition unavailable | Mark partition unavailable | `should_retry_on_unavailable_endpoint_status_codes()` |
```
FabianMeiswinkel: **🔴 Blocking** · Correctness: Spec-Implementation Mismatch

**410/1022 does NOT trigger partition marking in the actual implementation**

This matrix row claims both PPAF and PPCB actions for `410 | 1022 (LeaseNotFound)` are "Mark partition unavailable". However, in the actual implementation at `client_retry_policy.rs` lines 583–588, `410/1022` only calls `should_retry_on_unavailable_endpoint_status_codes()` — there is **no** call to `try_mark_endpoint_unavailable_for_pk_range()`. Only 503 and 429/3092 do partition marking via `try_mark_endpoint_unavailable_for_pk_range_and_retry_on_service_unavailable()`.

**Why it matters:** If the driver crate is implemented from this spec, 410/1022 would trigger partition-level marking that the gateway crate intentionally does not do. This could cause premature partition failovers on transient lease movements.

**Fix:** Update the matrix to show "N/A" for PPAF/PPCB actions on 410/1022, matching the actual code behavior.

---

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+
+```rust
+pub struct PartitionKeyRange {
+    pub id: String,
+    pub min_inclusive: String,
+    pub max_exclusive: String,
+}
+```
+
+Equality and hashing are based on `id` only, which uniquely identifies a partition
```
FabianMeiswinkel: **🔴 Blocking** · Correctness: Equality Semantics Wrong `[Already posted by @Copilot — Unresolved]`

**`PartitionKeyRange` equality uses 4 fields, not just `id`**

The spec states "Equality and hashing are based on `id` only". The actual implementation at `partition_key_range.rs` lines 136–155 uses `id`, `resource_id`, `min_inclusive`, and `max_exclusive` for both `PartialEq` and `Hash`.

**Why it matters:** If someone implements a `PartitionKeyRange` type for the driver with `id`-only hashing, overrides would match on wrong partition keys when `id`s collide across different ranges. This directly affects how entries are looked up in the failover `HashMap`.

**Fix:** Update to specify all four fields used for equality/hashing, and add `resource_id` to the struct definition (see also the struct definition at line 388).

Reinforcing @Copilot's existing comment — this is still unresolved.

---

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+    last_request_failure_time: RwLock<Instant>,
+}
+```
+
+### 5.3 `PartitionKeyRange`
+
+The key type used to identify partitions in the failover maps.
+
+```rust
+pub struct PartitionKeyRange {
```
FabianMeiswinkel: **🟡 Recommendation** · Completeness: Missing Critical Field

**`PartitionKeyRange` struct is missing `resource_id` (and other fields)**

The spec shows `PartitionKeyRange` with only 3 fields (`id`, `min_inclusive`, `max_exclusive`), but the actual implementation has 12 fields, including `resource_id`, `etag`, `timestamp`, `rid_prefix`, `throughput_fraction`, `status`, `lsn`, `parents`, and `owned_archival_pk_range_ids`.

The missing `resource_id` is critical — it’s used in the equality/hashing implementation (see finding above at line 395). Without it, the struct definition contradicts the behavior the spec should preserve.

**Fix:** At minimum, add `resource_id` to the spec struct and note that additional serde fields exist for deserialization.

---

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+
+## 5. Component Design
+
+### 5.1 `GlobalPartitionEndpointManager`
+
+The central coordinator for partition-level failover state. Manages two separate
+failover maps and the background failback loop.
+
+```rust
+pub struct GlobalPartitionEndpointManager {
```
FabianMeiswinkel: **🟡 Recommendation** · Completeness: Missing Concurrency Guard

**`GlobalPartitionEndpointManager` struct is missing `background_connection_init_active: AtomicBool`**

The spec’s struct definition omits the `background_connection_init_active: AtomicBool` field, which is used in the actual code at `global_partition_endpoint_manager.rs` line 83 for the CAS guard that prevents duplicate background task spawning. The `compare_exchange` in `initialize_and_start_circuit_breaker_failback_background_refresh` depends on it.

**Why it matters:** An implementer following this spec would miss the idempotent initialization guard, potentially spawning multiple concurrent failback loops.

**Fix:** Add the field to the struct definition, or add a note in the initialization section describing the CAS guard pattern.

---

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+
+---
+
+## 3. Feature Enablement & Configuration
+
+### Enable/Disable Flags
+
+| Flag | Source | Default | Description |
+|---|---|---|---|
+| `partition_level_circuit_breaker_enabled` | Env var `AZURE_COSMOS_PER_PARTITION_CIRCUIT_BREAKER_ENABLED` | `true` | Master switch for PPCB. When `false`, PPCB is disabled entirely. |
```
FabianMeiswinkel: **🟡 Recommendation** · Consistency: Misleading Label `[Already posted by @Copilot — Unresolved]`

**PPCB "Master switch" label is misleading**

The table describes `partition_level_circuit_breaker_enabled` as a "Master switch for PPCB" with default `true`, suggesting it independently controls PPCB. But in the actual initialization code at `cosmos_client_builder.rs` lines 300–303, the effective PPCB value is `server_flag || env_var`, meaning the env var being `false` doesn’t actually disable PPCB if the server flag is `true`.

The "master switch" label implies full control, but the env var is really just a fallback when the server doesn’t enable it.

Reinforcing @Copilot's existing comment — this is still unresolved.

---

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+
+2. **On failure** (`should_retry`):
+   - Classify the error (403/3, 503, 429/3092, etc.).
+   - If the request is eligible for PPAF or PPCB, record the failure and
+     potentially move the partition to the next available region.
+   - Return `Retry` so the retry loop re-invokes `before_send_request`, which
+     will now apply the updated partition-level override.
+
+3. **Background failback**:
+   - A periodic task scans all failed-over partitions.
```
FabianMeiswinkel: **🟡 Recommendation** · Accuracy: Overstated Scope `[Already posted by @Copilot — Unresolved]`

**Request flow summary says "scans all failed-over partitions" but failback only scans PPCB map**

Bullet 3 says "A periodic task scans all failed-over partitions", but Section 9.4 correctly clarifies the failback loop only scans the PPCB map (not the PPAF map). The summary should be tightened to say "scans all circuit-breaker failed-over partitions" or similar.

Reinforcing @Copilot's existing comment — this is still unresolved.

---

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+    enable_partition_level_circuit_breaker,
+);
+
+// 3. Register account refresh callback for dynamic updates
+global_endpoint_manager.set_on_account_refresh_callback(Arc::new(
+    move |account_props: &AccountProperties| {
+        partition_manager.configure_partition_level_automatic_failover(
+            account_props.enable_per_partition_failover_behavior,
+        );
+        partition_manager.configure_per_partition_circuit_breaker(
```
FabianMeiswinkel: **🟡 Recommendation** · Correctness: Non-Compiling Sample Code

**Initialization code block would not compile — uses moved value**

The spec’s initialization code shows:
```rust
partition_manager.configure_per_partition_circuit_breaker(
    account_props.enable_per_partition_failover_behavior
        || enable_partition_level_circuit_breaker,
);
```

But `partition_manager` is consumed by `GlobalPartitionEndpointManager::new()` on line 206. Using it again inside the closure would fail the borrow checker. The actual code at `cosmos_client_builder.rs` line 294 uses `partition_manager_clone` (a separate `Arc::clone`).

**Fix:** Show the `Arc::clone` pattern:
```rust
let partition_manager_clone = Arc::clone(&partition_manager);
// ... inside closure:
partition_manager_clone.configure_per_partition_circuit_breaker(...);
```

---

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+are incremented on each failure and checked against configurable thresholds.
+
+```
+increment_request_failure_counter_and_check_if_partition_can_failover(request)
+  │
+  ├─ Validate eligibility and extract partition key range + failed location
+  │
+  ├─ Get or insert PartitionKeyRangeFailoverInfo in the appropriate map
+  │
+  ├─ increment_request_failure_counts(is_read_only, current_time):
```
FabianMeiswinkel: **🟡 Recommendation** · Accuracy: Undocumented Race Condition

**TOCTOU race in `increment_request_failure_counts` should be documented**

The spec documents (and the actual code implements) a pattern where `increment_request_failure_counts` reads `last_request_failure_time` under a read lock, then decides whether to reset counters, then increments atomically. Two concurrent threads could both read the same `last_failure_time`, both decide to reset simultaneously, and one thread’s reset could clobber the other’s increment.

The impact is low (off-by-one on a threshold), but a spec should be precise about known concurrency behaviors.

**Fix:** Add a note in Section 13 (Known Issues) acknowledging this benign race condition.

---

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+}
+```
+
+**Critical behavior**: When the partition-level mark succeeds, the retry is
+immediate (`after: 0`). The next `before_send_request` call will find the
+updated override and route to the new region. When the partition-level mark
+fails (e.g., all regions exhausted), the request falls through to account-level
+failover with a location cache refresh.
+
+#### 8.1.3 On 503 / 429/3092 / 410/1022
```
FabianMeiswinkel: **🟡 Recommendation** · Accuracy: Incorrect Grouping

**Section 8.1.3 title incorrectly groups 410/1022 with 503 and 429/3092**

The section title is "On 503 / 429/3092 / 410/1022" and says "For these status codes..." implying all three follow the same `try_mark_endpoint_unavailable_for_pk_range_and_retry_on_service_unavailable()` pattern. But as noted in the status code matrix finding, 410/1022 does NOT call this method — it only does account-level retry via `should_retry_on_unavailable_endpoint_status_codes()`.

**Fix:** Remove 410/1022 from this section’s title. Add a separate brief section for 410/1022 explaining it only does account-level retry.

---

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+read endpoints (`read_endpoints()`) are used for both. Keeping separate maps avoids
+cross-contamination between these two routing strategies.
+
+### 5.2 `PartitionKeyRangeFailoverInfo`
+
+Per-partition failover tracking state. One instance exists per failed-over partition
+key range.
+
+```rust
+pub struct PartitionKeyRangeFailoverInfo {
```
FabianMeiswinkel: **🟡 Recommendation** · Design: Visibility Not Specified

**`PartitionKeyRangeFailoverInfo` struct visibility scope is unclear**

The spec shows `current`, `first_failed_location`, `collection_rid` as `pub` fields, and the struct itself as `pub struct`. In the actual code at `global_partition_endpoint_manager.rs` lines 856–880, the struct is `pub` but lives in a non-public module, making it effectively module-private.

**Why it matters:** If the driver crate needs to access `PartitionKeyRangeFailoverInfo` directly, the module would need to be made public. The spec should clarify whether this type is part of the public API surface or an internal implementation detail.

---

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+
+The failover maps use `RwLock<HashMap<...>>`. The read path
+(`try_route_request_for_partition_level_override`) acquires a read lock, and the
+write paths (`try_mark_endpoint_unavailable_for_partition_key_range`,
+`try_move_next_location`) acquire a write lock. Under high concurrency with many
+distinct partitions failing simultaneously, write-lock contention on the map could
+become a bottleneck.
+
+**Future improvement**: Migration to the unified `LocationStateStore` (see
+[Transport Pipeline Spec §4.6](TRANSPORT_PIPELINE_SPEC.md)) with epoch-guarded
```
FabianMeiswinkel: **🟢 Suggestion** · Documentation: Broken Cross-Reference `[Already posted by @Copilot — Unresolved]`

**Relative link to `TRANSPORT_PIPELINE_SPEC.md` won’t resolve**

The link `[Transport Pipeline Spec §4.6](TRANSPORT_PIPELINE_SPEC.md)` is relative to the current directory (`sdk/cosmos/azure_data_cosmos/docs/`), but `TRANSPORT_PIPELINE_SPEC.md` lives at `sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md`.

**Fix:** `[Transport Pipeline Spec §4.6](../../azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md)`

Reinforcing @Copilot's existing comment — this is still unresolved.

---

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+
+```rust
+pub struct GlobalPartitionEndpointManager {
+    /// Reference to account-level endpoint manager
+    global_endpoint_manager: Arc<GlobalEndpointManager>,
+
+    /// Duration a partition must remain marked unavailable before failback
+    /// Default: 5 seconds
+    /// Env: AZURE_COSMOS_ALLOWED_PARTITION_UNAVAILABILITY_DURATION_IN_SECONDS
+    partition_unavailability_duration_secs: i64,
```
FabianMeiswinkel: **🟢 Suggestion** · Robustness: Negative Duration Edge Case

**`i64` for duration fields allows negative values**

The spec uses `i64` for `partition_unavailability_duration_secs` and `background_connection_init_interval_secs`, matching the current implementation. However, durations are inherently non-negative. If someone sets `AZURE_COSMOS_ALLOWED_PARTITION_UNAVAILABILITY_DURATION_IN_SECONDS=-1`, the env var parsing would succeed and create a duration that every partition immediately exceeds, causing the failback loop to remove all entries every cycle.

**Fix:** Either document that negative values are clamped/rejected, or recommend `u64` in the spec.

---

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+either:
+- The failback loop removes them, or
+- All locations are exhausted and the entry is removed.
+
+This is generally acceptable because region topology changes are rare, but it means
+the override may point to a less-optimal region after a topology change.
+
+---
+
+## 14. Test Coverage
```
FabianMeiswinkel: **🟢 Suggestion** · Completeness: Missing Test Scenarios

**Test coverage section doesn’t mention configuration edge cases**

The test coverage section lists categories but doesn’t mention tests for edge cases in configuration: zero thresholds, negative durations, extremely large values, unparseable env var strings. These are important for robustness.

**Fix:** Add a "Configuration Edge Cases" subsection in the test plan covering:
- Zero failure thresholds
- Negative duration values
- Extremely large values (overflow)
- Unparseable/invalid env var strings
- Empty string env vars

---

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+2. **Faster recovery** — partition-level failovers are tracked independently, allowing
+   the background failback loop to restore each partition as soon as the original
+   region recovers, rather than waiting for a full account-level failover reversal.
+3. **Multi-master write support** — on accounts with multiple write regions, the
+   circuit breaker enables partition-level failover for both reads and writes across
+   preferred regions.
+
+### Two Complementary Mechanisms
+
+The SDK implements two distinct but complementary partition-level failover mechanisms:
```
FabianMeiswinkel: NIT: Be clear that for SM PPAF is not "done" by the SDK - it is purely a service-side decision. Only roel SDK plays is understanding the hints from the service.

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+either the PPAF path or the PPCB path, never both. The decision is based on the
+request's operation type (read vs. write) and whether the account supports multiple
+write locations.
+
+### Design Principles
+
+- **Partition granularity**: Failover state is tracked per `(PartitionKeyRange, Region)` pair.
+- **Threshold-gated**: The circuit breaker does not trip on the first failure. Consecutive
+  failure counters must exceed configurable thresholds before a partition is failed over.
+- **Non-deterministic failback**: After a configurable unavailability window, failed
```
FabianMeiswinkel: This might be acceptable temporarily - not lng-term. And if we use a "test and see if it works now" approach it would need to have very limited scope - like oening the flood gate to checkonly for single request etc. until seeing some success.

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+  partitions are optimistically marked healthy. Actual health is verified by subsequent
+  requests routing back to the original region.
+- **Environment-variable configurable**: All thresholds, windows, and intervals are
+  overridable via environment variables for testing and operational flexibility.
+- **No control-plane dependency**: Failover decisions are made locally by the SDK based
```
FabianMeiswinkel: True for PPCB - certainly not PPAF

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1005 @@
...
+  on observed request failures — no server-side signal is required beyond the HTTP
+  status codes.
+
+---
+
+## 2. Architectural Overview
```
FabianMeiswinkel: In teh driver there is no client retry policy or global endpoint manager (or the partiiton specific) one anymore. Is the plan to reintroduce it? If so we should be clear how - and IMO there should not be two typed of GEP for sure.

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
FabianMeiswinkel: PartitionKeyRangeFailoverInfo  struct should be pub(crate) at most?

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
kundadebdatta: I agree - Ideally the GEM and RetryPolicy should be non-existent in the driver. Since the goal is to still comply with the driver architecture and use the `LocationStateStore`, I have updated the spec accordingly to reflect the changes accurately.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1360 @@
...
+---
+
+## 3. Feature Enablement & Configuration
+
+### Enable/Disable Flags
+
+| Flag | Source | Default | Description |
+|---|---|---|---|
+| `ppcb_enabled` | Env var `AZURE_COSMOS_PER_PARTITION_CIRCUIT_BREAKER_ENABLED` | `true` | Fallback enablement for PPCB when the server flag is not set. The effective PPCB value is `server_flag \|\| env_var`, so PPCB remains enabled if the server flag is `true` regardless of this env var. |
+| `ppaf_enabled` | Server-side `AccountProperties.enable_per_partition_failover_behavior` | `false` | PPAF is enabled when the Cosmos DB account has this flag set. Updated dynamically on each account properties refresh. |
```
analogrelay: Let's try to avoid obscure acronyms in any public APIs. Maybe `partition_circuit_breaker_enabled` and `partition_write_failover_enabled`?

analogrelay: And ppcb_enabled should be set through the new options system in HierarchicalConfigModel.md, not directly from the env var. Though that's not fully landed yet.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1360 @@
...
+    pub ppcb_overrides: HashMap<String, PartitionFailoverEntry>,
+
+    /// PPAF enabled (from AccountProperties.enable_per_partition_failover_behavior).
+    pub ppaf_enabled: bool,
+
+    /// PPCB enabled (from env var + account property).
+    pub ppcb_enabled: bool,
+
+    /// Retained env var value for recomputation on account refresh.
+    pub ppcb_env_enabled: bool,
```
analogrelay: We'll need to read this out of the options instead.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1360 @@
...
+
+```rust
+/// Immutable partition-level endpoint routing state.
+///
+/// Managed via CAS in LocationStateStore alongside AccountEndpointState.
+#[derive(Clone, Debug)]
+pub(crate) struct PartitionEndpointState {
+    /// PPAF map: writes on single-master accounts.
+    /// Key: partition key range ID (String).
+    pub ppaf_overrides: HashMap<String, PartitionFailoverEntry>,
```
analogrelay: PKRangeID should be a newtype rather than just a magic string.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1398 @@
...
+│  Lock-free CAS via crossbeam_epoch::Atomic<T>                               │
+│                                                                             │
+│  ┌─ AccountEndpointState (existing) ────────────────────────────────────┐   │
+│  │  preferred_read_endpoints: Vec<CosmosEndpoint>                       │   │
+│  │  preferred_write_endpoints: Vec<CosmosEndpoint>                      │   │
+│  │  unavailable_endpoints: HashMap<CosmosEndpoint, (Instant, Reason)>   │   │
+│  │  multiple_write_locations_enabled: bool                              │   │
+│  └──────────────────────────────────────────────────────────────────────┘   │
+│                                                                             │
+│  ┌─ PartitionEndpointState (NEW — replaces empty placeholder) ──────────┐  │
```
tvaron3: nit: fix formatting

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1398 @@
...
+
+3. **Effect application** (Stage 6 — `location_state_store.apply`):
+   - `MarkPartitionUnavailable` → CAS loop on `PartitionEndpointState`:
+     insert or update a `PartitionFailoverEntry`, advance to the next available
+     endpoint in the preferred list.
+   - `MarkEndpointUnavailable` → existing CAS on `AccountEndpointState`.
+   - Both effects can be emitted simultaneously for 503/429/410 (partition
+     marking for future requests + endpoint marking for account-level routing).
+
+4. **Background failback**:
```
tvaron3, line 179: We should have some sort of mechanism to fail back slowly insteading of failing back all the requests.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1398 @@
...
+      │   └─ return read_count > read_threshold  (default: 2)
+      └─ else:
+          └─ return write_count > write_threshold  (default: 5)
+```
+
+### 7.2 Threshold Configuration
+
+| Parameter | Default | Environment Variable |
+|---|---|---|
+| Read failure threshold | 2 | `AZURE_COSMOS_CIRCUIT_BREAKER_CONSECUTIVE_FAILURE_COUNT_FOR_READS` |
```
tvaron3: I would change the naming from all variables to exclude consecutive as these aren't consecutive failures. It is misleading.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1398 @@
...
+
+1. `partition_state.ppcb_enabled == true`
+2. The operation targets `ResourceType::Document` or
+   `ResourceType::StoredProcedure` with `OperationType::Execute`
+3. The operation is **either**:
+   - A **read** (`operation.is_read_only()`), **or**
+   - A **write** on a **multi-master** account
+
+```rust
+fn is_eligible_for_ppcb(
```
tvaron3, line 297: **🔴 Blocking** · Correctness: Resource Type Check Too Broad

**PPCB eligibility includes StoredProcedure — diverges from Java SDK**

`is_eligible_for_ppcb()` uses `operation.resource_type().is_partitioned()` which matches `Document | StoredProcedure`. The Java SDK’s `isPerPartitionLevelCircuitBreakingApplicable()` checks **only** `ResourceType.Document`:

```java
// Java SDK — GlobalPartitionEndpointManagerForPerPartitionCircuitBreaker.java:415
if (request.getResourceType() != ResourceType.Document) {
    return false;
}
```

Applying PPCB to StoredProcedure CRUD operations (Create/Replace/Delete/Read) would cause unexpected partition-level failovers that no other SDK does.

**Fix**: PPCB eligibility should check `Document` only — no `StoredProcedure`. Don’t reuse a shared `is_partitioned()` since PPCB (Document-only) differs from PPAF (Document + StoredProcedure/Execute).

---

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1398 @@
...
+A request is eligible when **all** conditions are true:
+
+1. `partition_state.ppaf_enabled == true`
+2. The operation is a **write** (`!operation.is_read_only()`)
+3. The account is **single-master** (`!account_state.multiple_write_locations_enabled`)
+
+```rust
+// Pure eligibility check — evaluated in resolve_endpoint() and in
+// mark_partition_unavailable() routing system
+fn is_eligible_for_ppaf(
```
tvaron3, line 267: **🔴 Blocking** · Correctness: Resource Type Check Missing OperationType Restriction

**PPAF eligibility should restrict StoredProcedure to Execute only**

`is_eligible_for_ppaf()` uses `is_partitioned()` which matches all `StoredProcedure` operations. The Java SDK has a more precise check:

```java
// Java SDK — GlobalPartitionEndpointManagerForPerPartitionAutomaticFailover.java:328
if (resourceType == ResourceType.Document ||
    (resourceType == ResourceType.StoredProcedure && operationType == OperationType.ExecuteJavaScript)) {
```

This means PPAF in Java only applies to StoredProcedure **execution**, not CRUD (Create/Replace/Delete/Read). The Rust spec would incorrectly apply partition-level write failover when creating or deleting stored procedures — operations that aren’t partition-scoped the same way.

**Fix**: Add the OperationType restriction: `Document || (StoredProcedure && Execute)`. Consider separate eligibility helpers rather than a shared `is_partitioned()` since PPCB and PPAF have different resource type rules.

---

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1398 @@
...
+
+## 10. Status Code Handling Matrix
+
+The following table maps each status code to effects emitted by
+`evaluate_transport_result()`:
+
+| Status Code | Sub-Status | LocationEffects Emitted | OperationAction |
+|---|---|---|---|
+| 403 | 3 (WriteForbidden) | `RefreshAccountProperties` + `MarkEndpointUnavailable(WriteForbidden)` + `MarkPartitionUnavailable` | `FailoverRetry` |
+| 503 | Any | `MarkPartitionUnavailable` + `MarkEndpointUnavailable(ServiceUnavailable)` | `FailoverRetry` |
```
tvaron3: **🟡 Recommendation** · Completeness: Missing Status Codes

**Status code matrix is missing 408 and 500 — both trigger PPCB in Java SDK**

The Java SDK also invokes PPCB partition-level failure tracking (`handleLocationExceptionForPartitionKeyRange`) for:

- **408 (RequestTimeout)** — `ClientRetryPolicy.shouldRetryOnRequestTimeout()` and `shouldRetryOnGatewayTimeout()`
- **500 (InternalServerError)** — `shouldRetryOnInternalServerError()`

Without these rows, the Rust driver won’t increment PPCB failure counters for timeout and internal-error scenarios, leaving partitions stuck on unhealthy regions longer than the Java SDK would.

**Fix**: Add rows for 408 and 500 to the matrix specifying they trigger PPCB counter increments.

---

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1398 @@
...
+
+The driver replaces the SDK's `RwLock<HashMap>` with a CAS loop on an immutable
+`PartitionEndpointState` snapshot. Under high concurrency with many partitions
+failing simultaneously, multiple threads may race to CAS the same snapshot, causing
+retry iterations in the CAS loop.
+
+**Mitigation**: Partition failures are infrequent events. Under normal operation, the
+CAS loop completes in a single iteration. Under stress (many concurrent partition
+failures), the CAS may retry a few times, but each retry is a cheap clone +
+functional update — much cheaper than lock contention. If this becomes a measurable
```
tvaron3, line 1219: **🟢 Suggestion** · Scalability: Clone Cost Under Regional Outage

**Document HashMap clone cost at scale**

§13.1 discusses CAS contention but doesn’t mention the **clone cost**. The CAS pattern clones the entire `PartitionEndpointState` (both `ppaf_overrides` and `ppcb_overrides` HashMaps) on every update. During a regional outage affecting hundreds of partitions, each CAS iteration clones hundreds of map entries.

Worth a brief note here acknowledging this trade-off and when it might warrant a redesign (e.g., switching to per-partition atomic entries if clone cost becomes measurable under stress).

---

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
FabianMeiswinkel: Correct - I think the state should be something like healthy, unhealthy, probe_candidate or even multiple stages - initially we should only test for a single request - we have a healthy previous stage (failed-over) - so failing back should only happen when we can reliably do it - and even single request to all seems to wide-scoped - slowly increasing the percent of requests (initially single request - then slowly increasing percentage seems safer)

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
FabianMeiswinkel: +1 in general - but am against inventing new terms ppcb - per_partition_circuit_breaker and ppaf - per_partition_automatic_failover only has few more chars but aligns with terminology used everywhere - consistency is valuable enough to justify 3 or 4 ore chars in then ame

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1398 @@
...
+- **PPAF**: Updated when `sync_account_properties()` processes a new
+  `AccountProperties` response. When the server-side account property
+  `enable_per_partition_failover_behavior` changes, the next CAS swap on
+  `PartitionEndpointState` picks it up.
+
+- **PPCB**: The effective value is:
+  ```
+  enable_per_partition_failover_behavior || env_var_circuit_breaker_enabled
+  ```
+  This means PPCB is enabled if **either** the server flag or the client-side
```
FabianMeiswinkel, line 215: Hint to the service-side excerption discussed with service team.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
FabianMeiswinkel: Actually including StoredProc execution is teh correct thing to do - Java should include it :-)

FabianMeiswinkel: Please follow up with Abhijeet on this - @tvaron3 what is teh status in Python?

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
FabianMeiswinkel: And do we really want to use PKRangeId vs. EPK-Range/FeedRange? What happens when a partition gets split? Would we really want to rely on trying in the original reason and then learning we have to fail-over? I think that is justificable - it is a new partition after all - if so, using PKRanegId is fine - but agreed - should be NewType

FabianMeiswinkel (approved):
LGTM except for few blocking comments called out by @tvaron3  and @analogrelay  - Ia m ok to sign-off assuming these issues get addressed

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
tvaron3: I would actually say all 5xx errors should be tracked considering some customer want to use envoy.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
kundadebdatta: Added the PKRangeID as a new type.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
kundadebdatta: This is updated. Also, inlines with the comment below.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
tvaron3: We actually exclude query plan as well. Same with java. This is the logic in python. Following up with abhijeet. 
```python
if (request.resource_type not in (ResourceType.Document, ResourceType.PartitionKey)
             or request.operation_type == documents._OperationType.QueryPlan):
            return False
```

tvaron3: Synced with Abhijeet. The only concern with sproc is that it is not always targeted at documents. Sproc execution in those cases uses PartitionKey.NONE. PartitionKey.NONE cases can be excluded from PPAF/PPCB perspective.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
kundadebdatta: Updated the spec doc to capture this change.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
kundadebdatta: I will bring this as a discussion point when we finally shoot the PR out to implement PPAF + PPCB.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
kundadebdatta: This is a good point. Updated the spec to use probing based graceful failback.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1360 @@
...
+  └─ Return new_state
+```
+
+**Note**: Unlike the SDK, the driver's failback loop scans **both** maps (PPAF and
+PPCB). This is a deliberate improvement — in the SDK, PPAF entries are only removed
+when all locations are exhausted. The driver's immutable-snapshot pattern makes it
+trivial to sweep both maps in the same CAS operation.
+
+### 9.2 Failback Timing
+
```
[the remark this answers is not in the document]
kundadebdatta: Fixed.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
kundadebdatta: Updated the spec.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
kundadebdatta: This is now updated.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -0,0 +1,1360 @@
...
+### 11.1 Environment Variables
+
+| Variable | Type | Default | Description |
+|---|---|---|---|
+| `AZURE_COSMOS_PER_PARTITION_CIRCUIT_BREAKER_ENABLED` | `bool` | `true` | Master switch for per-partition circuit breaker |
+| `AZURE_COSMOS_ALLOWED_PARTITION_UNAVAILABILITY_DURATION_IN_SECONDS` | `i64` | `5` | Minimum time a partition must be unavailable before failback sweep considers it |
+| `AZURE_COSMOS_PPCB_STALE_PARTITION_UNAVAILABILITY_REFRESH_INTERVAL_IN_SECONDS` | `i64` | `300` | Interval between background failback sweep iterations |
+| `AZURE_COSMOS_CIRCUIT_BREAKER_CONSECUTIVE_FAILURE_COUNT_FOR_READS` | `i32` | `2` | Read failure threshold before circuit trips |
+| `AZURE_COSMOS_CIRCUIT_BREAKER_CONSECUTIVE_FAILURE_COUNT_FOR_WRITES` | `i32` | `5` | Write failure threshold before circuit trips |
```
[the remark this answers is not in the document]
kundadebdatta: These env variables are now renamed.

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
kundadebdatta: Agreed - updated the documentation to reflect this.

sdk/cosmos/azure_data_cosmos/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
kundadebdatta: Agreed - The probing will somewhat fix the non-deterministic failback, however, I will track this as a separate follow up task to come up with a plan.

PR #3887, opened 2026-03-09
Driver Transport - step 02

FabianMeiswinkel:
This PR implements "step 2" of the sdk\cosmos\azure_data_cosmos_driver\docs\TRANSPORT_PIPELINE_SPEC.md specficiation. - multi-region failover and endpoint management

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
+++ b/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Breaking Changes
 
+- Replaced `CosmosClientBuilder::with_application_region()` with a mandatory `RoutingStrategy` parameter on `build()`. Use `RoutingStrategy::ProximityTo(region)` to specify the application region. Also removed `CosmosClientOptions::with_application_region()`. ([#3889](https://github.com/Azure/azure-sdk-for-rust/pull/3889))
+
 ### Bugs Fixed
 
 - Fixes Circuit Breaker Failover Logic for Multi-Master Writes on 403/3. ([#3861](https://github.com/Azure/azure-sdk-for-rust/pull/3861))
--- a/sdk/cosmos/azure_data_cosmos/README.md
+++ b/sdk/cosmos/azure_data_cosmos/README.md
@@ -58,7 +58,9 @@ Instantiate a `DeveloperToolsCredential` to pass to the client. The same instanc
 
 ```rust
 use azure_identity::DeveloperToolsCredential;
-use azure_data_cosmos::{CosmosClient, CosmosAccountReference, CosmosAccountEndpoint};
+use azure_data_cosmos::{
+    CosmosClient, CosmosAccountReference, CosmosAccountEndpoint, RoutingStrategy,
+};
 
 async fn example() -> Result<(), Box<dyn std::error::Error>> {
     let credential: std::sync::Arc<dyn azure_core::credentials::TokenCredential> =
@@ -67,7 +69,8 @@ async fn example() -> Result<(), Box<dyn std::error::Error>> {
         .parse()?;
     let account = CosmosAccountReference::with_credential(endpoint, credential);
     let cosmos_client = CosmosClient::builder()
-        .build(account).await?;
+        .build(account, RoutingStrategy::ProximityTo("East US".into()))
+        .await?;
     Ok(())
 }
 ```
--- a/sdk/cosmos/azure_data_cosmos/examples/cosmos/main.rs
+++ b/sdk/cosmos/azure_data_cosmos/examples/cosmos/main.rs
@@ -1,7 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-use azure_data_cosmos::{CosmosAccountEndpoint, CosmosAccountReference, CosmosClient};
+use azure_data_cosmos::{
+    CosmosAccountEndpoint, CosmosAccountReference, CosmosClient, RoutingStrategy,
+};
 use azure_identity::DeveloperToolsCredential;
 use clap::{Args, CommandFactory, Parser, Subcommand};
 use std::error::Error;
@@ -34,6 +36,10 @@ struct ProgramArgs {
 struct SharedArgs {
     /// The Cosmos DB endpoint to connect to.
     endpoint: String,
+
+    /// The Azure region where the application is running (e.g. "East US").
+    #[arg(long)]
+    region: String,
 
     /// An authentication key to use when connecting to the Cosmos DB account. If omitted, the connection will use Entra ID.
     #[arg(long)]
@@ -81,6 +87,7 @@ pub async fn main() -> Result<(), Box<dyn Error>> {
 
 async fn create_client(args: &SharedArgs) -> Result<CosmosClient, Box<dyn Error>> {
     let endpoint: CosmosAccountEndpoint = args.endpoint.parse()?;
+    let strategy = RoutingStrategy::ProximityTo(args.region.clone().into());
     if let Some(key) = args.key.as_ref() {
         #[cfg(feature = "key_auth")]
         {
@@ -88,17 +95,17 @@ async fn create_client(args: &SharedArgs) -> Result<CosmosClient, Box<dyn Error>
                 endpoint,
                 azure_core::credentials::Secret::from(key.clone()),
             );
-            Ok(CosmosClient::builder().build(account).await?)
+            Ok(CosmosClient::builder().build(account, strategy).await?)
         }
         #[cfg(not(feature = "key_auth"))]
         {
-            let _ = key; // Mark 'key' as used to make the compiler happy.
+            let _ = (key, strategy); // Mark as used to make the compiler happy.
             Err("cannot authenticate with a key unless the 'key_auth' feature is enabled".into())
         }
     } else {
         let cred: Arc<dyn azure_core::credentials::TokenCredential> =
             DeveloperToolsCredential::new(None).unwrap();
         let account = CosmosAccountReference::with_credential(endpoint, cred);
-        Ok(CosmosClient::builder().build(account).await?)
+        Ok(CosmosClient::builder().build(account, strategy).await?)
     }
 }
--- a/sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
@@ -29,7 +29,7 @@ pub use super::cosmos_client_builder::CosmosClientBuilder;
 /// Using Entra ID authentication:
 ///
 /// ```rust,no_run
-/// use azure_data_cosmos::{CosmosClient, CosmosAccountReference, CosmosAccountEndpoint};
+/// use azure_data_cosmos::{CosmosClient, CosmosAccountReference, CosmosAccountEndpoint, RoutingStrategy, regions};
 /// use std::sync::Arc;
 ///
 /// # async fn doc() -> Result<(), Box<dyn std::error::Error>> {
@@ -40,7 +40,7 @@ pub use super::cosmos_client_builder::CosmosClientBuilder;
 ///     .unwrap();
 /// let account = CosmosAccountReference::with_credential(endpoint, credential);
 /// let client = CosmosClient::builder()
-///     .build(account)
+///     .build(account, RoutingStrategy::ProximityTo(regions::EAST_US))
 ///     .await?;
 /// # Ok(())
 /// # }
@@ -49,7 +49,7 @@ pub use super::cosmos_client_builder::CosmosClientBuilder;
 /// Using key authentication (requires `key_auth` feature):
 ///
 /// ```rust,no_run,ignore
-/// use azure_data_cosmos::{CosmosClient, CosmosAccountReference, CosmosAccountEndpoint};
+/// use azure_data_cosmos::{CosmosClient, CosmosAccountReference, CosmosAccountEndpoint, RoutingStrategy, regions};
 /// use azure_core::credentials::Secret;
 ///
 /// # async fn doc() -> Result<(), Box<dyn std::error::Error>> {
@@ -61,7 +61,7 @@ pub use super::cosmos_client_builder::CosmosClientBuilder;
 ///     Secret::from("my_account_key"),
 /// );
 /// let client = CosmosClient::builder()
-///     .build(account)
+///     .build(account, RoutingStrategy::ProximityTo(regions::EAST_US))
 ///     .await?;
 /// # Ok(())
 /// # }
@@ -80,7 +80,7 @@ impl CosmosClient {
     /// # Examples
     ///
     /// ```rust,no_run
-    /// use azure_data_cosmos::{CosmosClient, CosmosAccountReference, CosmosAccountEndpoint};
+    /// use azure_data_cosmos::{CosmosClient, CosmosAccountReference, CosmosAccountEndpoint, RoutingStrategy, regions};
     ///
     /// # async fn doc() -> Result<(), Box<dyn std::error::Error>> {
     /// let credential: std::sync::Arc<dyn azure_core::credentials::TokenCredential> =
@@ -90,7 +90,7 @@ impl CosmosClient {
     ///     .unwrap();
     /// let account = CosmosAccountReference::with_credential(endpoint, credential);
     /// let client = CosmosClient::builder()
-    ///     .build(account)
+    ///     .build(account, RoutingStrategy::ProximityTo(regions::EAST_US))
     ///     .await?;
     /// # Ok(())
     /// # }
--- a/sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
@@ -5,9 +5,8 @@
 
 use crate::{
     pipeline::{AuthorizationPolicy, CosmosHeadersPolicy, GatewayPipeline},
-    regions::RegionName,
     resource_context::{ResourceLink, ResourceType},
-    CosmosAccountReference, CosmosClient, CosmosClientOptions, CosmosCredential,
+    CosmosAccountReference, CosmosClient, CosmosClientOptions, CosmosCredential, RoutingStrategy,
 };
 
 use std::sync::Arc;
@@ -31,12 +30,17 @@ use azure_core::http::{ClientOptions, LoggingOptions, RetryOptions};
 /// [`CosmosAccountReference`] created via convenience constructors, or a tuple of
 /// `(CosmosAccountEndpoint, credential)` or `(Url, credential)`.
 ///
+/// A [`RoutingStrategy`] is also required to specify how the SDK should select regions.
+///
 /// # Examples
 ///
 /// Using Entra ID authentication:
 ///
 /// ```rust,no_run
-/// use azure_data_cosmos::{CosmosClientBuilder, CosmosAccountReference, CosmosAccountEndpoint};
+/// use azure_data_cosmos::{
+///     CosmosClientBuilder, CosmosAccountReference, CosmosAccountEndpoint,
+///     RoutingStrategy, regions,
+/// };
 /// use std::sync::Arc;
 ///
 /// # async fn doc() -> Result<(), Box<dyn std::error::Error>> {
@@ -45,7 +49,7 @@ use azure_core::http::{ClientOptions, LoggingOptions, RetryOptions};
 /// let endpoint: CosmosAccountEndpoint = "https://myaccount.documents.azure.com/".parse().unwrap();
 /// let account = CosmosAccountReference::with_credential(endpoint, credential);
 /// let client = CosmosClientBuilder::new()
-///     .build(account)
+///     .build(account, RoutingStrategy::ProximityTo(regions::EAST_US))
 ///     .await?;
 /// # Ok(())
 /// # }
@@ -54,14 +58,17 @@ use azure_core::http::{ClientOptions, LoggingOptions, RetryOptions};
 /// Using key authentication (requires `key_auth` feature):
 ///
 /// ```rust,no_run,ignore
-/// use azure_data_cosmos::{CosmosClientBuilder, CosmosAccountReference, CosmosAccountEndpoint};
+/// use azure_data_cosmos::{
+///     CosmosClientBuilder, CosmosAccountReference, CosmosAccountEndpoint,
+///     RoutingStrategy, regions,
+/// };
 /// use azure_core::credentials::Secret;
 ///
 /// # async fn doc() -> Result<(), Box<dyn std::error::Error>> {
 /// let endpoint: CosmosAccountEndpoint = "https://myaccount.documents.azure.com/".parse().unwrap();
 /// let account = CosmosAccountReference::with_master_key(endpoint, Secret::from("my_account_key"));
 /// let client = CosmosClientBuilder::new()
-///     .build(account)
+///     .build(account, RoutingStrategy::ProximityTo(regions::EAST_US))
 ///     .await?;
 /// # Ok(())
 /// # }
@@ -93,26 +100,6 @@ impl CosmosClientBuilder {
     /// * `suffix` - The suffix to append to the User-Agent header.
     pub fn with_user_agent_suffix(mut self, suffix: impl Into<String>) -> Self {
         self.options.user_agent_suffix = Some(suffix.into());
-        self
-    }
-
-    /// Sets the application region for routing.
-    ///
-    /// When set, the SDK generates a list of preferred regions sorted by
-    /// geographic proximity (round-trip time) from the given region.
-    /// This allows the client to prefer connecting to regions closest
-    /// to where the application is running.
-    ///
-    /// If not set, the SDK uses the account's configured regions
-    /// in the order returned by the service.
-    ///
-    /// Unknown region names will cause [`build()`](Self::build) to return an error.
-    ///
-    /// # Arguments
-    ///
-    /// * `region` - The region where the application is running.
-    pub fn with_application_region(mut self, region: RegionName) -> Self {
-        self.options.application_region = Some(region);
         self
     }
 
@@ -149,7 +136,7 @@ impl CosmosClientBuilder {
         self
     }
 
-    /// Builds the [`CosmosClient`] with the specified account reference.
+    /// Builds the [`CosmosClient`] with the specified account reference and region selection strategy.
     ///
     /// The account reference bundles an endpoint and credential. You can create one using
     /// [`CosmosAccountReference::with_credential()`] or [`CosmosAccountReference::with_master_key()`].
@@ -160,14 +147,23 @@ impl CosmosClientBuilder {
     /// # Arguments
     ///
     /// * `account` - The account reference containing the endpoint and credential.
+    /// * `routing_strategy` - The strategy for selecting which Azure regions to route requests to.
     ///
     /// # Errors
     ///
     /// Returns an error if the client cannot be constructed.
     pub async fn build(
-        self,
+        mut self,
         account: impl Into<CosmosAccountReference>,
+        routing_strategy: RoutingStrategy,
     ) -> azure_core::Result<CosmosClient> {
+        // Apply the region selection strategy to internal options.
+        match routing_strategy {
+            RoutingStrategy::ProximityTo(region) => {
+                self.options.application_region = Some(region);
+            }
+        }
+
         let (account_endpoint, credential) = account.into().into_parts();
         let endpoint = account_endpoint.into_url();
 
@@ -328,7 +324,7 @@ impl CosmosClientBuilder {
 
 #[cfg(test)]
 mod tests {
-    use crate::{regions, CosmosAccountReference, CosmosClient};
+    use crate::{regions, CosmosAccountReference, CosmosClient, RoutingStrategy};
     use azure_core::credentials::{AccessToken, TokenCredential, TokenRequestOptions};
     use std::sync::Arc;
 
@@ -357,8 +353,10 @@ mod tests {
     #[tokio::test]
     async fn build_with_known_region_succeeds() {
         let result = CosmosClient::builder()
-            .with_application_region(regions::EAST_US)
-            .build(test_account())
+            .build(
+                test_account(),
+                RoutingStrategy::ProximityTo(regions::EAST_US),
+            )
             .await;
         assert!(result.is_ok());
     }
@@ -367,8 +365,7 @@ mod tests {
     async fn build_with_unknown_region_succeeds() {
         let unknown = regions::RegionName::from("unknown");
         let result = CosmosClient::builder()
-            .with_application_region(unknown)
-            .build(test_account())
+            .build(test_account(), RoutingStrategy::ProximityTo(unknown))
             .await;
         assert!(
             result.is_ok(),
@@ -377,15 +374,14 @@ mod tests {
     }
 
     /// When an unknown region is passed, the SDK cannot generate proximity ordering
-    /// so it falls back to account-order (same as no application_region set).
+    /// so it falls back to account-order.
     #[tokio::test]
     async fn build_with_unknown_region_uses_account_order() {
         use crate::models::AccountRegion;
 
         let unknown = regions::RegionName::from("unknown");
         let client = CosmosClient::builder()
-            .with_application_region(unknown)
-            .build(test_account())
+            .build(test_account(), RoutingStrategy::ProximityTo(unknown))
             .await
             .expect("build should succeed");
 
@@ -421,24 +417,20 @@ mod tests {
         );
     }
 
-    #[tokio::test]
-    async fn build_without_application_region_succeeds() {
-        let result = CosmosClient::builder().build(test_account()).await;
-        assert!(result.is_ok());
-    }
-
-    /// Verifies the full flow: builder with application_region → proximity list →
+    /// Verifies the full flow: builder with ProximityTo strategy → proximity list →
     /// GlobalEndpointManager → LocationCache → correctly ordered endpoints.
     ///
     /// Passing in East US should produce a proximity-ordered list of regions with East US 2 first,
     /// then West US, then West Europe.
     #[tokio::test]
-    async fn application_region_produces_proximity_ordered_endpoints() {
+    async fn proximity_strategy_produces_ordered_endpoints() {
         use crate::models::AccountRegion;
 
         let client = CosmosClient::builder()
-            .with_application_region(regions::EAST_US)
-            .build(test_account())
+            .build(
+                test_account(),
+                RoutingStrategy::ProximityTo(regions::EAST_US),
+            )
             .await
             .expect("build should succeed with known region");
 
@@ -487,13 +479,15 @@ mod tests {
     /// Verifies that request-level excluded regions interact correctly with
     /// proximity-ordered preferred regions through the full builder flow.
     #[tokio::test]
-    async fn application_region_with_excluded_regions() {
+    async fn proximity_strategy_with_excluded_regions() {
         use crate::models::AccountRegion;
         use crate::operation_context::OperationType;
 
         let client = CosmosClient::builder()
-            .with_application_region(regions::EAST_US)
-            .build(test_account())
+            .build(
+                test_account(),
+                RoutingStrategy::ProximityTo(regions::EAST_US),
+            )
             .await
             .expect("build should succeed");
 
--- a/sdk/cosmos/azure_data_cosmos/src/fault_injection/client_builder.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/fault_injection/client_builder.rs
@@ -19,7 +19,7 @@ use super::rule::FaultInjectionRule;
 ///
 /// ```rust,no_run
 /// use azure_data_cosmos::{
-///     CosmosClientBuilder, CosmosAccountEndpoint,
+///     CosmosClientBuilder, CosmosAccountEndpoint, RoutingStrategy, regions,
 ///     fault_injection::{
 ///         FaultInjectionClientBuilder, FaultInjectionErrorType,
 ///         FaultInjectionResultBuilder, FaultInjectionRuleBuilder,
@@ -41,7 +41,10 @@ use super::rule::FaultInjectionRule;
 /// let endpoint: CosmosAccountEndpoint = "https://myaccount.documents.azure.com/".parse().unwrap();
 /// let client = CosmosClientBuilder::new()
 ///     .with_fault_injection(fault_builder)
-///     .build((endpoint, Secret::from("my_account_key")))
+///     .build(
+///         (endpoint, Secret::from("my_account_key")),
+///         RoutingStrategy::ProximityTo(regions::EAST_US),
+///     )
 ///     .await
 ///     .unwrap();
 /// # }
--- a/sdk/cosmos/azure_data_cosmos/src/fault_injection/mod.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/fault_injection/mod.rs
@@ -75,10 +75,13 @@
 //! // 5. Create the client with fault injection
 //! let client = CosmosClientBuilder::new()
 //!     .with_fault_injection(fault_builder)
-//!     .build(CosmosAccountReference::with_master_key(
-//!         "https://myaccount.documents.azure.com/".parse().unwrap(),
-//!         Secret::new("my_account_key"),
-//!     ))
+//!     .build(
+//!         CosmosAccountReference::with_master_key(
+//!             "https://myaccount.documents.azure.com/".parse().unwrap(),
+//!             Secret::new("my_account_key"),
+//!         ),
+//!         azure_data_cosmos::RoutingStrategy::ProximityTo("East US".into()),
+//!     )
 //!     .await
 //!     .unwrap();
 //! # }
--- a/sdk/cosmos/azure_data_cosmos/src/lib.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/lib.rs
@@ -34,6 +34,7 @@ pub use models::CosmosResponse;
 pub use options::*;
 pub use partition_key::*;
 pub use query::Query;
+pub use routing_strategy::RoutingStrategy;
 pub use transactional_batch::{
     BatchDeleteOptions, BatchReadOptions, BatchReplaceOptions, BatchUpsertOptions,
     TransactionalBatch, TransactionalBatchOperationResult, TransactionalBatchResponse,
@@ -53,4 +54,5 @@ pub mod regions;
 mod request_context;
 mod retry_policies;
 mod routing;
+mod routing_strategy;
 mod serde;
--- a/sdk/cosmos/azure_data_cosmos/src/options/mod.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/options/mod.rs
@@ -46,11 +46,6 @@ impl CosmosClientOptions {
         self
     }
 
-    pub fn with_application_region(mut self, application_region: impl Into<RegionName>) -> Self {
-        self.application_region = Some(application_region.into());
-        self
-    }
-
     pub fn with_custom_headers(mut self, custom_headers: HashMap<HeaderName, HeaderValue>) -> Self {
         self.custom_headers = custom_headers;
         self
--- a/sdk/cosmos/azure_data_cosmos/src/routing_strategy.rs
+++ b/sdk/cosmos/azure_data_cosmos/src/routing_strategy.rs
@@ -1,0 +1,36 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+//! Region selection strategies for Azure Cosmos DB client routing.
+
+use crate::regions::RegionName;
+
+/// Determines how the SDK selects Azure regions for routing requests.
+///
+/// This is a required parameter when building a [`CosmosClient`](crate::CosmosClient),
+/// ensuring that every client is configured with an explicit region selection strategy.
+///
+/// # Examples
+///
+/// ```rust
+/// use azure_data_cosmos::{RoutingStrategy, regions};
+///
+/// let strategy = RoutingStrategy::ProximityTo(regions::EAST_US);
+/// ```
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub enum RoutingStrategy {
+    /// Select regions by geographic proximity to the given region.
+    ///
+    /// The SDK generates a list of preferred regions sorted by estimated
+    /// proximity to the specified region. Proximity estimates are built
+    /// into the SDK and may differ from actual round-trip times observed
+    /// at runtime.
+    ///
+    /// If the application is not running in an Azure region, specify the
+    /// closest Azure region to the application's actual location.
+    ///
+    /// Specifying an unknown region name results in undefined region
+    /// selection behavior that may change in future versions of the SDK.
+    ProximityTo(RegionName),
+}
--- a/sdk/cosmos/azure_data_cosmos/tests/framework/test_client.rs
+++ b/sdk/cosmos/azure_data_cosmos/tests/framework/test_client.rs
@@ -14,7 +14,7 @@ use azure_data_cosmos::options::ItemOptions;
 use azure_data_cosmos::regions::{RegionName, EAST_US_2, WEST_US_3};
 use azure_data_cosmos::{
     clients::DatabaseClient, ConnectionString, CosmosClient, CreateContainerOptions, PartitionKey,
-    Query,
+    Query, RoutingStrategy,
 };
 use futures::TryStreamExt;
 use std::time::Duration;
@@ -272,10 +272,11 @@ impl TestClient {
         let credential = connection_string.account_key.clone();
         let mut builder = azure_data_cosmos::CosmosClient::builder();
 
-        // Apply application region for the client
-        if let Some(region) = application_region.or(fault_client_application_region) {
-            builder = builder.with_application_region(region);
-        }
+        // Determine the region selection strategy
+        let region = application_region
+            .or(fault_client_application_region)
+            .unwrap_or(HUB_REGION);
+        let strategy = RoutingStrategy::ProximityTo(region);
 
         // Configure invalid certificate acceptance (e.g., for emulator)
         #[cfg(feature = "allow_invalid_certificates")]
@@ -299,9 +300,10 @@ impl TestClient {
         let endpoint: azure_data_cosmos::CosmosAccountEndpoint =
             connection_string.account_endpoint.parse()?;
         let cosmos_client = builder
-            .build(azure_data_cosmos::CosmosAccountReference::with_master_key(
-                endpoint, credential,
-            ))
+            .build(
+                azure_data_cosmos::CosmosAccountReference::with_master_key(endpoint, credential),
+                strategy,
+            )
             .await?;
 
         Ok(TestClient {
@@ -802,11 +804,13 @@ impl TestRunContext {
                 )
             })?;
         CosmosClient::builder()
-            .with_application_region(region)
-            .build(azure_data_cosmos::CosmosAccountReference::with_master_key(
-                endpoint,
-                parsed.account_key.clone(),
-            ))
+            .build(
+                azure_data_cosmos::CosmosAccountReference::with_master_key(
+                    endpoint,
+                    parsed.account_key.clone(),
+                ),
+                RoutingStrategy::ProximityTo(region),
+            )
             .await
     }
 
--- a/sdk/cosmos/azure_data_cosmos_native/c_tests/error_handling.c
+++ b/sdk/cosmos/azure_data_cosmos_native/c_tests/error_handling.c
@@ -14,18 +14,18 @@ int test_null_pointer_handling() {
     cosmos_client *client = NULL;
     cosmos_error_code code;
     
-    code = cosmos_client_create_with_key(NULL, endpoint, key, NULL, &client);
+    code = cosmos_client_create_with_key(NULL, endpoint, key, "East US", NULL, &client);
     ASSERT_ERROR_CODE(code, COSMOS_ERROR_CODE_CALL_CONTEXT_MISSING, "NULL context rejected with CALL_CONTEXT_MISSING");
 
-    code = cosmos_client_create_with_key(&ctx.call_ctx, NULL, key, NULL, &client);
+    code = cosmos_client_create_with_key(&ctx.call_ctx, NULL, key, "East US", NULL, &client);
     ASSERT(code != COSMOS_ERROR_CODE_SUCCESS, "NULL endpoint correctly rejected");
     if (code == COSMOS_ERROR_CODE_SUCCESS && client) cosmos_client_free(client);
 
-    code = cosmos_client_create_with_key(&ctx.call_ctx, endpoint, NULL, NULL, &client);
+    code = cosmos_client_create_with_key(&ctx.call_ctx, endpoint, NULL, "East US", NULL, &client);
     ASSERT(code != COSMOS_ERROR_CODE_SUCCESS, "NULL key correctly rejected");
     if (code == COSMOS_ERROR_CODE_SUCCESS && client) cosmos_client_free(client);
 
-    code = cosmos_client_create_with_key(&ctx.call_ctx, endpoint, key, NULL, NULL);
+    code = cosmos_client_create_with_key(&ctx.call_ctx, endpoint, key, "East US", NULL, NULL);
     ASSERT(code != COSMOS_ERROR_CODE_SUCCESS, "NULL output pointer correctly rejected");
 
     REQUIRE(test_context_create_runtime(&ctx, 0) == TEST_PASS, "Created runtime context for remaining tests");
@@ -60,7 +60,7 @@ int test_invalid_runtime_context() {
     ctx.include_error_details = false;
     
     cosmos_client *client = NULL;
-    cosmos_error_code code = cosmos_client_create_with_connection_string(&ctx, connection_string, NULL, &client);
+    cosmos_error_code code = cosmos_client_create_with_connection_string(&ctx, connection_string, "East US", NULL, &client);
     
     ASSERT(code != COSMOS_ERROR_CODE_SUCCESS, "Invalid/freed runtime context correctly rejected");
     if (client) {
--- a/sdk/cosmos/azure_data_cosmos_native/c_tests/test_common.h
+++ b/sdk/cosmos/azure_data_cosmos_native/c_tests/test_common.h
@@ -75,7 +75,7 @@ static inline int test_context_create_client(test_context *ctx) {
         options.danger_allow_invalid_certificates = true;
     }
     
-    cosmos_error_code code = cosmos_client_create_with_connection_string(&ctx->call_ctx, connection_string, &options, &ctx->client);
+    cosmos_error_code code = cosmos_client_create_with_connection_string(&ctx->call_ctx, connection_string, "East US", &options, &ctx->client);
     if (code != COSMOS_ERROR_CODE_SUCCESS) {
         printf("Failed to create client from connection string\n");
         return TEST_FAIL;
--- a/sdk/cosmos/azure_data_cosmos_native/include/azurecosmos.h
+++ b/sdk/cosmos/azure_data_cosmos_native/include/azurecosmos.h
@@ -70,7 +70,7 @@ typedef struct cosmos_container_client cosmos_container_client;
  * Using Entra ID authentication:
  *
  * ```rust,no_run
- * use azure_data_cosmos::{CosmosClient, CosmosAccountReference, CosmosAccountEndpoint};
+ * use azure_data_cosmos::{CosmosClient, CosmosAccountReference, CosmosAccountEndpoint, RoutingStrategy, regions};
  * use std::sync::Arc;
  *
  * # async fn doc() -> Result<(), Box<dyn std::error::Error>> {
@@ -81,7 +81,7 @@ typedef struct cosmos_container_client cosmos_container_client;
  *     .unwrap();
  * let account = CosmosAccountReference::with_credential(endpoint, credential);
  * let client = CosmosClient::builder()
- *     .build(account)
+ *     .build(account, RoutingStrategy::ProximityTo(regions::EAST_US))
  *     .await?;
  * # Ok(())
  * # }
@@ -90,7 +90,7 @@ typedef struct cosmos_container_client cosmos_container_client;
  * Using key authentication (requires `key_auth` feature):
  *
  * ```rust,no_run,ignore
- * use azure_data_cosmos::{CosmosClient, CosmosAccountReference, CosmosAccountEndpoint};
+ * use azure_data_cosmos::{CosmosClient, CosmosAccountReference, CosmosAccountEndpoint, RoutingStrategy, regions};
  * use azure_core::credentials::Secret;
  *
  * # async fn doc() -> Result<(), Box<dyn std::error::Error>> {
@@ -102,7 +102,7 @@ typedef struct cosmos_container_client cosmos_container_client;
  *     Secret::from("my_account_key"),
  * );
  * let client = CosmosClient::builder()
- *     .build(account)
+ *     .build(account, RoutingStrategy::ProximityTo(regions::EAST_US))
  *     .await?;
  * # Ok(())
  * # }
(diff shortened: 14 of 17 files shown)


sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/components.rs
```diff
@@ -39,34 +41,73 @@ pub(crate) struct RoutingDecision {
...
+
+    /// Whether session retry budget allows another attempt.
+    pub fn can_retry_session(&self) -> bool {
+        self.session_token_retry_count < self.max_session_retries
     }
 
-    /// Advance to the next transport retry attempt.
-    pub fn advance_transport_retry(&self) -> Self {
+    /// Returns a new state advanced for failover retry.
+    pub fn advance_failover(&self) -> Self {
```
analogrelay: This should take `self` by value instead of by reference. If the caller doesn't own this value, it can clone it.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/components.rs
```diff
@@ -39,34 +41,73 @@ pub(crate) struct RoutingDecision {
...
         Self {
-            transport_retry_count: self.transport_retry_count + 1,
-            max_transport_retries: self.max_transport_retries,
+            failover_retry_count: self.failover_retry_count + 1,
+            ..self.clone()
+        }
+    }
+
+    /// Returns a new state advanced for session retry.
+    pub fn advance_session_retry(&self) -> Self {
```
analogrelay: Same here, pass `self` by value

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/components.rs
```diff
@@ -39,34 +41,73 @@ pub(crate) struct RoutingDecision {
...
+        Self {
+            session_token_retry_count: self.session_token_retry_count + 1,
+            ..self.clone()
+        }
+    }
+
+    /// Advances the location index for endpoint list of `list_len`.
+    pub fn advance_location(&self, list_len: usize) -> Self {
```
analogrelay: And here, take `self` by value

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
analogrelay, line 117: Could we pass in the routing decision by-value so that we don't have to do cloning inside `build_transport_request`? If there are things we need from the routing decision later in _this_ function (like the endpoint), we could clone those out of it before calling `build_transport_request`

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -210,8 +283,8 @@ fn build_transport_request(
 
     Ok(TransportRequest {
         method,
+        endpoint: routing.endpoint.clone(),
```
analogrelay, line 347: Lots of clones here. It feels like this should receive an owned `RoutingDecision` and just deconstruct it into the pieces it needs.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -252,3 +325,135 @@ fn build_cosmos_response(
...
+        )
+    }
+
+    fn test_routing() -> RoutingDecision {
+        RoutingDecision {
+            endpoint: Url::parse("https://test.documents.azure.com:443/").unwrap(),
+            region: None,
+        }
+    }
+
```
FabianMeiswinkel, line 448: 🔴 **Blocking — Test compilation errors**

`test_routing()` creates `RoutingDecision { endpoint: Url, region: None }`, but `RoutingDecision` was changed to `RoutingDecision { endpoint: CosmosEndpoint }`. This causes two compiler errors:
- `E0308` — wrong type (`Url` vs `CosmosEndpoint`)
- `E0560` — unknown field `region`

Verified: `cargo test -p azure_data_cosmos_driver --all-features --no-run` fails with these exact errors.

**Fix**: Change `test_routing()` to:
```rust
fn test_routing() -> RoutingDecision {
    RoutingDecision {
        endpoint: CosmosEndpoint::global(
            Url::parse("https://test.documents.azure.com:443/").unwrap(),
        ),
    }
}
```
And add `CosmosEndpoint` to the `use` import from `crate::driver::routing`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -355,8 +407,17 @@ impl CosmosDriver {
...
             .await?;
+
+        // Keep the operation routing snapshot in sync with current account metadata.
+        let previous_generation = Some(self.location_state_store.account_snapshot().generation);
+        let account_state = build_account_endpoint_state(
+            account_properties.as_ref(),
+            self.location_state_store.default_endpoint().clone(),
+            previous_generation,
+        );
+        self.location_state_store.swap_account(account_state);
```
FabianMeiswinkel: 🔴 **Blocking — `swap_account` wipes unavailability marks**

Every `execute_operation` call unconditionally runs `build_account_endpoint_state` (which creates state with `unavailable_endpoints: Default::default()`) and then `swap_account`. This erases endpoint unavailability marks set by concurrent operations' STAGE 6 CAS loop.

**Scenario**:
1. Operation A hits 503 → STAGE 6 applies `MarkEndpointUnavailable` for endpoint X (CAS loop succeeds)
2. Operation B (concurrent) enters `execute_operation`, calls `swap_account` with fresh state → `unavailable_endpoints` is empty
3. Operation A's retry loop snapshots the store → endpoint X is available again → sends to same failing endpoint

This creates a **permanent TOCTOU window** where marks are repeatedly erased, defeating multi-region failover under concurrency.

Additionally, every call bumps the `generation` counter, which causes `LocationIndex.is_current()` to return false for long-running retries, unnecessarily resetting the round-robin index to 0 and potentially trapping retries on the same failing endpoint.

**Fix options**:
1. Remove the `swap_account` from `execute_operation` entirely — let only `refresh_account_properties_if_due` manage store updates
2. Guard with a one-time check (e.g., only swap when `generation == 0`)
3. Have `build_account_endpoint_state` accept the old `unavailable_endpoints` and carry them forward

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -131,8 +173,13 @@ pub(crate) async fn execute_operation_pipeline(
...
-                // → next iteration
+            }
+            OperationAction::SessionRetry { new_state } => {
+                let endpoints_len = location
+                    .account
+                    .preferred_endpoints(operation.is_read_only())
+                    .len();
+                retry_state = new_state.advance_location(endpoints_len);
             }
             OperationAction::Abort { error, status } => {
```
FabianMeiswinkel, line 216: 🔴 **Blocking — `SessionRetry` incorrectly advances location**

`SessionRetry` calls `new_state.advance_location(endpoints_len)`, which moves to the next endpoint. The spec (§4.1) says session retries do **not** change the endpoint — they retry with a different session token on the same or next endpoint via the retry state's own mechanism. Session-not-available is a consistency error resolved by token adjustment, not by failing over to a different region.

Additionally, `SessionRetry` omits the deadline check that `FailoverRetry` performs. This means session retries may wastefully build a request and invoke the transport pipeline just to immediately time out.

**Fix**:
1. Remove the `advance_location()` call for `SessionRetry` — session retries should stay on the same endpoint
2. Add the same deadline check as in the `FailoverRetry` branch

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
```diff
@@ -24,54 +27,162 @@ use super::components::{OperationAction, OperationRetryState, TransportOutcome,
...
+                            endpoint: endpoint.clone(),
+                            reason: UnavailableReason::WriteForbidden,
+                        },
+                    ],
+                );
+            }
+
+            if status.is_read_session_not_available() && retry_state.can_retry_session() {
+                if !retry_state.can_use_multiple_write_locations
+                    && retry_state.session_token_retry_count >= 1
```
FabianMeiswinkel: 🟡 **Recommendation — Session retry count cap deviates from Java SDK behavior**

For single-write accounts, the Java SDK allows **2 session retries** (`retries when sessionTokenRetryCount <= 2`). The Rust implementation allows **1** (`max_session_retries` defaults to 1 and this early-abort condition `session_token_retry_count >= 1` further limits it). For multi-write accounts, Java allows retries up to `endpoints.size()`, Rust caps at 1.

Reference: [ClientRetryPolicy.java - shouldRetryOnSessionNotAvailable](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/ClientRetryPolicy.java#L240-L255)

Under session consistency, fewer retries means higher failure rates for read-after-write scenarios, especially when request routing is asymmetric.

**Fix**: Set `max_session_retries` default to 2 for single-write accounts, and to `endpoints.len()` for multi-write accounts.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -149,13 +196,39 @@ pub(crate) async fn execute_operation_pipeline(
...
-        endpoint: endpoint.url().clone(),
-        region: region.clone(),
+/// Step 2 resolves from `LocationSnapshot` and `AccountEndpointState`.
+fn resolve_endpoint(
+    operation: &CosmosOperation,
+    retry_state: &OperationRetryState,
+    location: &LocationSnapshot,
+) -> RoutingDecision {
+    let account = location.account.as_ref();
+    let account = expire_unavailable_endpoints(account, Instant::now(), Duration::from_secs(60));
```
FabianMeiswinkel: 🟡 **Recommendation — Hardcoded 60s TTL ignores configured value**

`resolve_endpoint` calls `expire_unavailable_endpoints(account, Instant::now(), Duration::from_secs(60))` with a hardcoded TTL, while `LocationStateStore` has a configurable `endpoint_unavailability_ttl` field (initialized from `RuntimeOptions`). These values can diverge.

Configuring a shorter or longer TTL via `RuntimeOptionsBuilder::with_endpoint_unavailability_ttl()` has no effect on the actual routing behavior — a user would set the option thinking it works, but routing ignores it.

**Fix**: Pass the configured TTL through the `LocationStateStore` (add a getter) or pass it as a parameter to `resolve_endpoint`.

sdk/cosmos/azure_data_cosmos_driver/Cargo.toml
```diff
@@ -22,7 +22,7 @@ azure_core = { workspace = true, features = [
   "hmac_rust",
 ] }
 base64.workspace = true
-crossbeam = { workspace = true, features = ["crossbeam-epoch"] }
+crossbeam = { workspace = true, features = ["crossbeam-epoch", "alloc", "std"] }
```
FabianMeiswinkel: 🟡 **Recommendation — Unnecessary crossbeam features bloat dependencies**

The change from `features = ["crossbeam-epoch"]` to `features = ["crossbeam-epoch", "alloc", "std"]` pulls in `crossbeam-channel`, `crossbeam-deque`, and `crossbeam-queue` (visible in `Cargo.lock`). The spec §11.5 explicitly says these should **not** be enabled.

**Fix**: If `std` is actually needed by `crossbeam-epoch` usage, use `crossbeam-epoch = { version = "0.9", default-features = false, features = ["std"] }` directly, or enable only `crossbeam-epoch/std` via the crossbeam umbrella crate. Remove `alloc` and `std` features from the `crossbeam` dependency.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
```diff
@@ -0,0 +1,139 @@
...
+    unavailable.insert(endpoint.clone(), (Instant::now(), reason));
+
+    AccountEndpointState {
+        unavailable_endpoints: unavailable,
+        ..state.clone()
+    }
+}
+
+/// Returns a new state with expired endpoint unavailability removed.
+pub(crate) fn expire_unavailable_endpoints(
```
FabianMeiswinkel, line 75: 🟡 **Recommendation — Expired entries not persisted back to store**

`resolve_endpoint` calls `expire_unavailable_endpoints` to create a local copy with expired entries removed, but this cleaned state is never written back to the `LocationStateStore`. Expired entries accumulate in the store indefinitely and get re-cloned and re-filtered every iteration.

Under sustained failover scenarios, the store's `unavailable_endpoints` map grows unboundedly because entries are only locally filtered, never removed from the actual store. This creates increasing memory and CPU overhead per routing call.

**Fix options**:
1. Run expiry in the `LocationStateStore::apply()` path
2. Add a periodic sweep (e.g., inside `refresh_account_properties_if_due`)
3. Apply expiry as a side-effect during STAGE 6

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+    if preferred_write_endpoints.is_empty() {
+        preferred_write_endpoints.push(default_endpoint.clone());
+    }
+
+    AccountEndpointState {
+        generation,
+        preferred_read_endpoints,
+        preferred_write_endpoints,
+        unavailable_endpoints: Default::default(),
```
FabianMeiswinkel, line 52: 🟢 **Suggestion — `build_account_endpoint_state` creates fresh empty `unavailable_endpoints`**

This `unavailable_endpoints: Default::default()` line means every `build_account_endpoint_state` call produces a state with no unavailability marks. When called from `execute_operation` (which runs for every operation), this wipes marks set by concurrent operations' STAGE 6 CAS loop. See related comment on `cosmos_driver.rs` for the full impact analysis.

If the unconditional `swap_account` in `execute_operation` is retained, this function should accept and preserve the previous `unavailable_endpoints`.

sdk/cosmos/azure_data_cosmos_driver/tests/multi_region_failover.rs
```diff
@@ -0,0 +1,82 @@
...
+        db,
+    );
+
+    let _ = driver
+        .execute_operation(
+            CosmosOperation::read_database(db_ref),
+            OperationOptions::new(),
+        )
+        .await;
+}
```
FabianMeiswinkel, line 96: 🟢 **Suggestion — Integration tests are smoke tests, not behavioral tests**

Both `write_forbidden_triggers_refresh_and_failover` and `session_not_available_retries_across_locations` execute a normal `read_database` with no fault injection and no assertions on retry/failover behavior. They're identical in body.

These tests give false confidence — the test names imply behavioral validation but the tests only confirm "an operation runs without crashing." When fault injection lands (Step 8), these should be converted to real behavioral tests with assertions on retry counts, endpoint selection, and effect application.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
```diff
@@ -24,54 +27,162 @@ use super::components::{OperationAction, OperationRetryState, TransportOutcome,
...
+                }
+
+                return (
+                    OperationAction::FailoverRetry {
+                        new_state: retry_state.advance_failover(),
+                        delay: None,
+                    },
+                    vec![
+                        LocationEffect::MarkPartitionUnavailable(UnavailablePartition {
+                            partition_key_range_id: "unknown".to_owned(),
```
FabianMeiswinkel: 💬 **Observation — `MarkPartitionUnavailable` uses `"unknown"` placeholder**

The `partition_key_range_id: "unknown".to_owned()` is a placeholder for Step 3. The effect is a no-op (store ignores `MarkPartitionUnavailable`), so there's no runtime impact.

Just flagging as a reminder to wire the actual partition key range ID through the `TransportResult` or the `CosmosOperation` when Step 3 is implemented.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -8,18 +8,19 @@
 //! (7-stage loop) that later steps will expand.
 
 use std::sync::Arc;
-use std::time::Instant;
+use std::time::{Duration, Instant};
```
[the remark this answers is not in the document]
analogrelay: I wish we could get Copilot to stop writing these "Step 1", "Step 2" comments altogether. They're not really related to the final code we expect to have here. As it is, I think we'll need to have an agent take a pass to check for them after everything lands.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
analogrelay: Honestly, it would be good to have this as an explicit `// TODO:` in the code. I don't mind having `TODO`s in the code when we know we're replacing something. I'm just not a fan of the constant references to "Step 1", "Step 2", etc. :)

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -131,8 +173,13 @@ pub(crate) async fn execute_operation_pipeline(
                         ));
                     }
                 }
-                retry_state = new_state;
-                // → next iteration
+            }
+            OperationAction::SessionRetry { new_state } => {
```
xinlian12, line 185: **🔴 Blocking** · Correctness: Missing Deadline Enforcement
`[History Rule R-004: RequestSentStatus accuracy / deadline enforcement]`

**`SessionRetry` branch does not check the operation deadline before retrying**

The `FailoverRetry` branch (lines 141–172) checks the end-to-end deadline before proceeding to the next iteration:

```rust
if let Some(d) = deadline {
    if Instant::now() >= d { ... return Err(...) }
}
```

But the `SessionRetry` branch here skips that check entirely — it advances the location and falls through to the next loop iteration without validating remaining time budget.

**Consequence:** A series of session retries (especially if the service keeps returning 404/1002) could run past the configured `end_to_end_latency_policy` timeout, violating the timeout contract. This is the same class of issue that was caught and fixed in Step 1’s `deadline_exceeded_result` (PR #3875).

> 📋 **Rule R-004** (Confidence: High, Severity: Blocking): *"At this point in the code, do we actually know if the request was sent to the server? If we've already received HTTP headers back, the status must be `Sent`, not `Unknown`. If it's a pre-send failure, it must be `NotSent`."* — This rule extends to deadline enforcement: every retry path must validate the remaining time budget before looping. PR #3875 fixed an identical gap where `deadline_exceeded_result()` always returned `NotSent` even after receiving a 429 response.

**Fix:** Add the same deadline check that `FailoverRetry` uses before continuing.

---

sdk/cosmos/azure_data_cosmos_driver/Cargo.toml
xinlian12: **🟡 Recommendation** · Performance: Dependency Bloat
`[History Rule R-013: reqwest/dependency hygiene]`

**Unnecessary crossbeam features pull in unused crates**

The change from `features = ["crossbeam-epoch"]` to `features = ["crossbeam-epoch", "alloc", "std"]` causes `crossbeam` to enable its default sub-crates. The `Cargo.lock` diff confirms this pulls in `crossbeam-channel`, `crossbeam-deque`, and `crossbeam-queue` — none of which are used by the driver.

The transport pipeline spec (§11.5) explicitly states: *"Other crossbeam features (crossbeam-channel, crossbeam-deque, crossbeam-queue) are **not** enabled."*

```suggestion
crossbeam = { workspace = true, features = ["crossbeam-epoch"] }
```

If `alloc`/`std` are actually needed for the `Atomic<T>` API, they’re already implied by `crossbeam-epoch` on non-`no_std` targets.

> 📋 **Rule R-013** (Confidence: High, Severity: Blocking): *"Is `reqwest` (or any specific transport crate) added as an unconditional dependency? First-party support is paramount."* — This rule generalizes to all dependency hygiene: only enable the features you actually use. PR #3848 added crossbeam with minimal features (`crossbeam-epoch` only) per the spec; this PR inadvertently expanded it.

---

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -59,7 +59,35 @@ pub(crate) async fn execute_operation_pipeline(
     diagnostics: DiagnosticsContextBuilder,
 ) -> azure_core::Result<CosmosResponse> {
     let mut diagnostics = diagnostics;
-    let mut retry_state = OperationRetryState::initial();
+    let location_snapshot = location_state_store.snapshot();
+    let max_failover_retries = effective_options
+        .max_failover_retry_count
+        .unwrap_or_else(|| {
+            std::env::var("AZURE_COSMOS_FAILOVER_RETRY_COUNT")
```
xinlian12: **🟡 Recommendation** · Performance: Hot-Path Syscall

**`std::env::var` called on every operation invocation**

`max_failover_retry_count` and `max_session_retry_count` fall back to `std::env::var("AZURE_COSMOS_FAILOVER_RETRY_COUNT")` and `std::env::var("AZURE_COSMOS_SESSION_RETRY_COUNT")` respectively. These are called inside `execute_operation_pipeline`, which runs on **every Cosmos operation**.

`std::env::var` acquires an OS lock (`environ` mutex on Unix, `GetEnvironmentVariable` syscall on Windows) on each call. Under high throughput this becomes measurable contention.

**Suggestion:** Read env vars once during `CosmosDriver::new()` or `LocationStateStore::new()` and store the resolved defaults as fields — the same pattern already used for `endpoint_unavailability_ttl` in `cosmos_driver.rs` lines 230–236.

---

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
xinlian12: **🟢 Suggestion** · Performance: Unnecessary Allocation

**`"unknown".to_owned()` allocates a string on every 503/gone retry for a no-op field**

`partition_key_range_id: "unknown".to_owned()` allocates a heap `String` on every retry path that emits `MarkPartitionUnavailable`. Since the `MarkPartitionUnavailable` effect is a no-op in Step 2 (the store ignores it), this allocation is pure waste.

Options:
- Use `String::new()` (zero-alloc empty string)
- Make `partition_key_range_id` an `Option<String>` since it’s not meaningful until Step 3

---

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Fixed in 9cf3dfbe6 — Updated `test_routing()` to use `CosmosEndpoint::global(...)` and added `CosmosEndpoint` import.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
FabianMeiswinkel: Fixed in 5cb29525b — Replaced unconditional `swap_account` with a new `sync_account_properties` method that uses `apply_account` (CAS loop) and carries forward existing `unavailable_endpoints` from the current state. Both `execute_operation` and `refresh_account_properties_if_due` now use this CAS-based path. Removed the now-unused `swap_account` method. This also resolves Finding #8 (empty `unavailable_endpoints` in `build_account_endpoint_state`).

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Fixed in cfa2d3696 — Added the missing deadline check for `SessionRetry` (same pattern as `FailoverRetry`). The `advance_location` call was kept intentionally: in Gateway mode, in-region replica retries are already handled by Gateway, so the operation-level session retry should advance to the next suitable region (hub/write-region for single-write, round-robin for multi-write). Updated the spec accordingly.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
FabianMeiswinkel: Fixed in 512d2f3a6 — Default `max_session_retries` now matches Java SDK: 2 for single-write accounts, `endpoints.len()` for multi-write accounts. The early-abort guard for single-write changed from `session_token_retry_count >= 1` to `>= 2`. Updated the spec.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Fixed in 20f053d58 — Added `endpoint_unavailability_ttl()` getter to `LocationStateStore` and wired the configured value through to `resolve_endpoint`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
FabianMeiswinkel: Fixed in 8ec342a83 — Added `TODO(Step 3)` comment to the `"unknown"` placeholder.

sdk/cosmos/azure_data_cosmos_driver/tests/multi_region_failover.rs
FabianMeiswinkel: Fixed in 7bf8b3c04 — Added `TODO(Step 8)` comments to both tests documenting what fault injection assertions they should have when Step 8 lands. Module-level doc also notes they're smoke tests pending fault injection.

sdk/cosmos/azure_data_cosmos_driver/Cargo.toml
FabianMeiswinkel: Fixed in 273662f54 — Switched from the `crossbeam` umbrella crate to `crossbeam-epoch` directly with `default-features = false, features = ["std"]`. This eliminates the unnecessary `crossbeam-channel`, `crossbeam-deque`, and `crossbeam-queue` transitive dependencies.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
FabianMeiswinkel: Resolved by 5cb29525b — `sync_account_properties` now carries forward `unavailable_endpoints` from the current state via CAS, so `build_account_endpoint_state`'s `Default::default()` no longer wipes marks.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
FabianMeiswinkel: Fixed in efdc64f26 — `sync_account_properties` now filters expired entries using the configured `endpoint_unavailability_ttl` before carrying `unavailable_endpoints` forward in the CAS loop. Since `sync_account_properties` runs on every operation and on refresh, expired entries are cleaned up promptly.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed in ea5665327 — Updated module and function doc comments to reflect Step 2.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -131,8 +173,13 @@ pub(crate) async fn execute_operation_pipeline(
...
-                retry_state = new_state;
-                // → next iteration
+            }
+            OperationAction::SessionRetry { new_state } => {
+                let endpoints_len = location
+                    .account
+                    .preferred_endpoints(operation.is_read_only())
+                    .len();
+                retry_state = new_state.advance_location(endpoints_len);
             }
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed in cfa2d3696 — deadline check added to SessionRetry. See Finding #3 thread.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -131,8 +173,13 @@ pub(crate) async fn execute_operation_pipeline(
...
                 }
-                retry_state = new_state;
-                // → next iteration
+            }
+            OperationAction::SessionRetry { new_state } => {
+                let endpoints_len = location
+                    .account
+                    .preferred_endpoints(operation.is_read_only())
+                    .len();
+                retry_state = new_state.advance_location(endpoints_len);
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed in cfa2d3696 — see Finding #3 thread. `advance_location` kept intentionally (Gateway mode).

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed in 5cb29525b — see Finding #2 thread.

sdk/cosmos/azure_data_cosmos_driver/tests/multi_region_failover.rs
```diff
@@ -0,0 +1,76 @@
...
+    let db = read_env("AZURE_COSMOS_TEST_DATABASE").expect("missing test database env");
+    let db_ref = azure_data_cosmos_driver::models::DatabaseReference::from_name(
+        driver.account().clone(),
+        db,
+    );
+
+    let _ = driver
+        .execute_operation(CosmosOperation::read_database(db_ref), OperationOptions::new())
+        .await;
+}
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed — see reply on the corresponding finding threads for details on each fix.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed in 20f053d58 — see Finding #5 thread.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
```diff
@@ -14,7 +14,10 @@
 
 use azure_core::http::headers::Headers;
 
-use crate::models::{CosmosOperation, CosmosStatus, SubStatusCode};
+use crate::{
+    driver::routing::{CosmosEndpoint, LocationEffect, UnavailablePartition, UnavailableReason},
+    models::{CosmosOperation, CosmosStatus, SubStatusCode},
+};
 
 use super::components::{OperationAction, OperationRetryState, TransportOutcome, TransportResult};
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed in ea5665327 — Updated module-level doc comments to reflect Step 2.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/components.rs
```diff
@@ -39,34 +41,73 @@ pub(crate) struct RoutingDecision {
 /// Expanded in Step 2 with failover, session, and location index fields.
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed in ea5665327 — Updated all three stale doc comments (components.rs, operation_pipeline.rs, retry_evaluation.rs) to reflect the Step 2 implementation.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
analogrelay: This should get resolved with the new config model from #3778

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/endpoint.rs
```diff
@@ -0,0 +1,102 @@
...
+    Global,
+    /// Regional endpoint (`{account}-{region}.documents.azure.com`).
+    Regional,
+}
+
+/// Type-safe index into preferred endpoint lists.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) struct LocationIndex {
+    index: usize,
+    generation: u64,
```
analogrelay, line 60: `u64` is really big for this. We might actually want to make this a bit smaller. It's internal though, so we could do that if the memory usage becomes an issue

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
```diff
@@ -166,8 +166,8 @@ pub(crate) async fn execute_transport_pipeline(
             ExecutionContext::Retry
         };
 
-        let endpoint_string = request.url.as_str().to_owned();
-        let region = request.region.clone();
+        let endpoint_string = request.endpoint.url().as_str().to_owned();
+        let region = request.endpoint.region().cloned();
```
analogrelay: Another place we're doing quite a bit of cloning

sdk/cosmos/azure_data_cosmos_driver/Cargo.toml
FabianMeiswinkel: Already fixed.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/components.rs
FabianMeiswinkel: Fixed in 94d80d0ca — same commit as above.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/components.rs
FabianMeiswinkel: Fixed in 94d80d0ca — Changed `advance_failover`, `advance_session_retry`, and `advance_location` to take `self` by value. Callers in `retry_evaluation.rs` (where `retry_state` is `&OperationRetryState`) now explicitly `.clone()` before calling.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/components.rs
FabianMeiswinkel: Fixed in 94d80d0ca — same commit as above.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Fixed in 74f11576a — changed `start_request` and `RequestDiagnostics::new` to accept `&CosmosEndpoint` directly instead of separate region/endpoint parameters. The transport pipeline caller no longer pre-clones; decomposition happens inside the diagnostics layer.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
FabianMeiswinkel: Fixed in 74f11576a — `start_request` now accepts `&CosmosEndpoint` directly, so the transport pipeline just passes `&request.endpoint` without any pre-cloning. Cloning/decomposition now happens inside `RequestDiagnostics::new`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Won't fix — passing `RoutingDecision` by value would require cloning the endpoint out beforehand for `evaluate_transport_result` (which needs it after `execute_transport_pipeline` consumes the `TransportRequest`). Net clone count stays the same; the current borrow-based approach is already optimal.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Already fixed in cfa2d3696 — deadline check added to `SessionRetry` branch matching the `FailoverRetry` pattern.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
FabianMeiswinkel: Fixed in c23da1ef2 — replaced with `String::new()` (zero-alloc empty string).

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/endpoint.rs
FabianMeiswinkel: Won't fix for now — `LocationIndex` is `Copy` (`usize + u64` = 16 bytes on 64-bit). Changing to `u32` would nominally save 4 bytes but alignment padding with the `usize` field keeps the struct at 16 bytes anyway. Negligible practical impact; can revisit if memory profiling shows otherwise.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/components.rs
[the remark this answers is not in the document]
FabianMeiswinkel: Already fixed in ea5665327 — updated all stale doc comments referencing "Step 1" / "Step 2" across components.rs, operation_pipeline.rs, and retry_evaluation.rs.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Fixed in 1dc9758b0 — `AZURE_COSMOS_FAILOVER_RETRY_COUNT` and `AZURE_COSMOS_SESSION_RETRY_COUNT` are now read once at `CosmosDriver::new()` and stored as resolved defaults. The pipeline no longer calls `std::env::var` per operation.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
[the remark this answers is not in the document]
FabianMeiswinkel: Already fixed in 5cb29525b — replaced unconditional `swap_account` with CAS-based `sync_account_properties` that preserves `unavailable_endpoints`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
FabianMeiswinkel: Fixed in c23da1ef2 — replaced `"unknown".to_owned()` with `String::new()` (zero-alloc). The effect is still a no-op until Step 3.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
```diff
@@ -0,0 +1,361 @@
...
+    account: Atomic<AccountEndpointState>,
+    partitions: Atomic<PartitionEndpointState>,
+    account_metadata_cache: Arc<AccountMetadataCache>,
+    account_endpoint: AccountEndpoint,
+    account_refresh_fn: AccountRefreshFn,
+    default_endpoint: CosmosEndpoint,
+    endpoint_unavailability_ttl: Duration,
+    refresh_interval: Duration,
+    last_refresh_epoch_ms: AtomicU64,
+}
```
FabianMeiswinkel, line 72: 🔴 **Blocking — Resource leak: `LocationStateStore` missing `Drop` implementation**

`crossbeam_epoch::Atomic<T>` does **not** drop its pointee when the `Atomic` itself is dropped — this is by design since other threads may hold epoch-pinned references. Without an explicit `Drop` impl, both the `AccountEndpointState` and `PartitionEndpointState` heap allocations leak when a `LocationStateStore` is dropped.

For a singleton driver that lives for the app lifetime this is only a leak at process exit (benign). But if drivers are created and destroyed dynamically (e.g. in tests, hot-reconfiguration), the leak accumulates.

Suggested fix:
```rust
impl Drop for LocationStateStore {
    fn drop(&mut self) {
        // SAFETY: We have `&mut self`, so no concurrent access is possible.
        // `epoch::unprotected()` is safe here because we are the sole owner.
        unsafe {
            let guard = epoch::unprotected();
            let acct = self.account.load(Ordering::Relaxed, guard);
            if !acct.is_null() {
                drop(acct.into_owned());
            }
            let part = self.partitions.load(Ordering::Relaxed, guard);
            if !part.is_null() {
                drop(part.into_owned());
            }
        }
    }
}
```

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -68,13 +92,18 @@ pub(crate) async fn execute_operation_pipeline(
...
+        let routing = resolve_endpoint(
+            operation,
+            &retry_state,
+            &location,
+            location_state_store.endpoint_unavailability_ttl(),
+        );
 
         // ── STAGE 3: Build transport request ───────────────────────────
-        let execution_context = if retry_state.transport_retry_count == 0 {
+        let execution_context = if retry_state.failover_retry_count == 0 {
```
FabianMeiswinkel: 🟡 **Recommendation — Diagnostics: failover retries use `ExecutionContext::Retry` instead of `RegionFailover`**

The `ExecutionContext` enum has a `RegionFailover` variant (used in diagnostics tests), but the operation loop only emits `Initial` or `Retry`:

```rust
let execution_context = if retry_state.failover_retry_count == 0 {
    ExecutionContext::Initial
} else {
    ExecutionContext::Retry  // ← Should distinguish failover from transport retry
};
```

This loses diagnostic signal — operators tracing a failing operation can't tell whether a retry was a same-region transport retry or a cross-region failover, which is critical information for debugging regional outages.

Consider: use `ExecutionContext::RegionFailover` when `failover_retry_count > 0`, and leave `Retry` for future same-region transport retries (or remove it if no longer applicable).

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -149,12 +221,52 @@ pub(crate) async fn execute_operation_pipeline(
...
+/// Step 2 resolves from `LocationSnapshot` and `AccountEndpointState`.
+fn resolve_endpoint(
+    operation: &CosmosOperation,
+    retry_state: &OperationRetryState,
+    location: &LocationSnapshot,
+    endpoint_unavailability_ttl: Duration,
+) -> RoutingDecision {
+    let account = location.account.as_ref();
+    let account =
+        expire_unavailable_endpoints(account, Instant::now(), endpoint_unavailability_ttl);
```
FabianMeiswinkel: 🟡 **Recommendation — Performance: hot-path allocation from `expire_unavailable_endpoints` clone**

`resolve_endpoint` calls `expire_unavailable_endpoints(account, ...)` on every operation-loop iteration, which clones the entire `AccountEndpointState` including all endpoint vectors and the `unavailable_endpoints` HashMap.

However, `sync_account_properties` (called earlier in the pipeline) already expires entries via `unavailable.retain(...)`. So this inline expiry is largely redundant — entries are already cleaned at sync time.

Consider either:
1. Remove this call and use the snapshot as-is (since sync already expired entries), or
2. Check expiry inline during the candidate loop (`marked_at + ttl > now`) without cloning the entire state.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
```diff
@@ -24,54 +29,197 @@ use super::components::{OperationAction, OperationRetryState, TransportOutcome,
...
             headers,
             body,
             request_sent,
         } => {
-            // In Step 1, all HTTP errors (non-429) are propagated as abort.
-            // 429 is already handled by the transport pipeline's throttle retry.
-            // Step 2 will add handling for 403/3, 503, 404/1002, etc.
-            if retry_state.can_retry_transport()
-                && (request_sent.definitely_not_sent() || operation.is_idempotent())
+            let request_definitely_not_sent = request_sent.definitely_not_sent();
```
FabianMeiswinkel, line 50: 🟡 **Recommendation — Spec divergence: `request_definitely_not_sent` branch is dead code for `HttpError`**

For `TransportOutcome::HttpError`, `request_sent` is always `Sent` (set by `from_http_response` — an HTTP response requires a sent request). So `definitely_not_sent()` is never true for HTTP errors, making the branch at ~L95 dead code when this runs inside the 503/gone/system-resource-unavailable block.

The subsequent non-idempotent write abort (the `!(is_read_only || is_idempotent)` guard) is a deliberate and defensible spec divergence — safer than the spec for non-idempotent writes. But it should be documented with a comment explaining the rationale and the spec should be updated to match if this behavior is accepted.

Two suggestions:
1. Add a comment like `// Note: For HttpError, request was always sent. This branch is reachable only for TransportError.`
2. Add a comment explaining why non-idempotent writes are aborted here (intentional safety guard beyond spec).

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -355,8 +442,15 @@ impl CosmosDriver {
             .account_metadata_cache()
             .get_or_fetch(account_endpoint, || self.fetch_account_properties(account))
             .await?;
+
+        // Keep the operation routing snapshot in sync with current account metadata.
+        // Uses CAS to preserve unavailable_endpoints marks set by concurrent operations.
+        self.location_state_store.sync_account_properties(
```
FabianMeiswinkel, line 449: 🟡 **Recommendation — Concurrency: CAS contention from `sync_account_properties` on every operation**

Every `execute_operation` call runs `sync_account_properties`, which enters a CAS loop that clones `unavailable_endpoints`, runs expiry `retain`, and builds a new `AccountEndpointState`. Under high concurrency with many concurrent operations, this creates contention on `Atomic<AccountEndpointState>` and repeated CAS retries.

Consider gating this sync with a generation check — only run `sync_account_properties` when the cached `AccountProperties` generation differs from the current `AccountEndpointState.generation`, or use a timestamp-based throttle similar to `refresh_account_properties_if_due`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -169,7 +281,7 @@ fn build_transport_request(
     let resource_ref = operation.resource_reference();
     let request_path = resource_ref.request_path();
     let url = {
-        let mut base = routing.endpoint.clone();
+        let mut base = routing.endpoint.url().clone();
```
FabianMeiswinkel, line 306: 🟡 **Recommendation — Performance: `RoutingDecision`/`CosmosEndpoint` cloned where moves would suffice**

The endpoint is cloned twice here:
1. `routing.endpoint.url().clone()` (L284)
2. `routing.endpoint.clone()` (L325)

The `RoutingDecision` is not used after `build_transport_request`, so passing it by-value and destructuring would avoid cloning `CosmosEndpoint` (which contains a `Url` — a heap-allocated `String`).

(Echoes @analogrelay's existing unresolved comment about pass-by-value.)

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
```diff
@@ -0,0 +1,139 @@
...
+//! Pure routing systems for account endpoint state.
+
+use std::time::{Duration, Instant};
+
+use crate::driver::cache::AccountProperties;
+
+use super::{AccountEndpointState, CosmosEndpoint, UnavailableReason};
+
+/// Builds account endpoint state from account metadata.
+pub(crate) fn build_account_endpoint_state(
```
FabianMeiswinkel, line 20: 🟡 **Recommendation — Completeness: missing `preferred_locations` parameter**

The spec defines:
```rust
fn build_account_endpoint_state(
    properties: &AccountProperties,
    preferred_locations: &[RegionName],  // ← Missing in implementation
    default_endpoint: CosmosEndpoint,
    previous_generation: Option<u64>,
) -> AccountEndpointState;
```

Without `preferred_locations`, endpoint lists use whatever order the server returns, ignoring user configuration. Configuring preferred regions would have no routing effect.

If this is intentionally scoped for a later step, consider adding a `// TODO: Wire preferred_locations from DriverOptions/RuntimeOptions` comment.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/endpoint.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// Creates an initial index at position 0.
+    pub fn initial(generation: u64) -> Self {
+        Self {
+            index: 0,
+            generation,
+        }
+    }
+
+    /// Advances to the next index, wrapping within list bounds.
+    pub fn next(self, list_len: usize) -> Self {
```
FabianMeiswinkel, line 73: 🟢 **Suggestion — `next` with `list_len == 0` returns unchanged index**

If both `preferred_read_endpoints` and `preferred_write_endpoints` are somehow empty (this shouldn't happen due to the fallback in `build_account_endpoint_state`), `advance_location` would return the same index forever, and `resolve_endpoint` would always fall back to `default_endpoint`. This is actually correct fallback behavior.

However, a `debug_assert!(list_len > 0, "endpoint list should never be empty")` would catch unexpected empty lists during development without impacting release builds.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -252,3 +364,176 @@ fn build_cosmos_response(
...
+
+        let partition_key_header = request
+            .headers
+            .get_optional_str(&HeaderName::from_static("x-ms-documentdb-partitionkey"))
+            .expect("partition key header should be set");
+        assert_eq!(partition_key_header, "[\"pk1\"]");
+    }
+
+    #[test]
+    fn resolve_endpoint_uses_write_region_for_single_write_session_retry() {
```
FabianMeiswinkel, line 526: 🟢 **Suggestion — Missing test: `resolve_endpoint` when all preferred endpoints are unavailable**

There's a test for single-write session retry routing, but no test for the fallback path where all endpoints are excluded/unavailable and `default_endpoint` is returned. This is an important edge case for multi-region failover resilience.

Consider adding a test where all preferred endpoints are in `unavailable_endpoints` and verifying that `resolve_endpoint` falls back to `default_endpoint`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/components.rs
```diff
@@ -23,52 +27,115 @@ use crate::{
...
+            session_retry_routing: SessionRetryRouting::PreferredEndpoints,
+            ..self
         }
     }
 
-    /// Whether the retry budget allows another transport retry.
-    pub fn can_retry_transport(&self) -> bool {
-        self.transport_retry_count < self.max_transport_retries
+    /// Returns a new state advanced for session retry.
+    pub fn advance_session_retry(self) -> Self {
```
FabianMeiswinkel, line 112: 🟢 **Suggestion — Missing unit test for `advance_session_retry` routing behavior**

`advance_session_retry` sets `session_retry_routing` to `PreferredEndpoints` for multi-write and `PreferredWriteEndpoints` for single-write. This logic is critical for session consistency failover but is only tested indirectly through `resolve_endpoint_uses_write_region_for_single_write_session_retry`.

A direct unit test of `advance_session_retry` would catch regressions more precisely — e.g. verifying the routing enum, retry count increment, and multi-write vs single-write behavior in isolation.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
```diff
@@ -0,0 +1,361 @@
...
+        }
+    }
+
+    /// Returns the default endpoint.
+    pub fn default_endpoint(&self) -> &CosmosEndpoint {
+        &self.default_endpoint
+    }
+
+    /// Returns a lock-free snapshot of account and partition state.
+    pub fn snapshot(&self) -> LocationSnapshot {
```
FabianMeiswinkel, line 149: 💬 **Observation — `snapshot()` full-clones + `Arc::new` on every call**

Each snapshot pins the epoch guard, dereferences the atomic pointer, clones the full `AccountEndpointState` (including all endpoint vectors and the `unavailable_endpoints` HashMap), and wraps it in `Arc::new(...)`.

The spec suggests `snapshot()` should return cloned `Arc<T>` references. A future optimization would be to store `Arc<AccountEndpointState>` in the `Atomic` pointer directly so snapshots just increment the refcount instead of full cloning. Not blocking for Step 2, but worth considering for the hot path as region counts and unavailable-endpoint maps grow.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
```diff
@@ -0,0 +1,361 @@
...
+    driver::cache::{AccountMetadataCache, AccountProperties},
+    models::AccountEndpoint,
+};
+
+use super::{
+    build_account_endpoint_state, mark_endpoint_unavailable, AccountEndpointState, CosmosEndpoint,
+    LocationEffect,
+};
+
+/// Step 3 adds partition-level state; Step 2 keeps this as a placeholder.
```
FabianMeiswinkel: 💬 **Observation — "Step N" references remaining in code comments**

@analogrelay noted the preference to avoid "Step 1/2/3" framing in code comments. Several new lines still reference "Step 2" or "Step 3" terminology:
- `location_state_store.rs` L27, L98, L145, L157
- `location_effects.rs` L18
- `operation_pipeline.rs` L224
- `retry_evaluation.rs` L125

Consider replacing these with topic-based `TODO` comments (e.g. `// TODO(partition-routing): ...` instead of `// TODO(Step 3): ...`) and descriptive comments instead of step references (e.g. "placeholder for partition-level state" rather than "Step 3 adds partition-level state; Step 2 keeps this as placeholder").

The `TODO(Step 8)` references in the integration tests are fine since they reference future work clearly.

FabianMeiswinkel (commented):
## Deep Review — Step 2: Multi-region failover & endpoint management

**Overall assessment:** Strong implementation. The DOP (data-oriented programming) architecture is faithful to the spec. The CAS-based state management, the pure `evaluate_transport_result` function, and the lock-free snapshot pattern are well-executed. Most critical issues from the prior review round have been addressed.

### Summary of findings

| Severity | Count | Key Items |
|----------|-------|-----------|
| 🔴 Blocking | 1 | Memory leak from missing `Drop` on `LocationStateStore` |
| 🟡 Recommendation | 6 | Diagnostics fidelity, hot-path allocation, dead code branch, CAS contention, cloning, missing `preferred_locations` |
| 🟢 Suggestion | 3 | `debug_assert` for empty lists, test coverage gaps |
| 💬 Observation | 2 | Snapshot clone cost, "Step N" comment style |

**Prior review overlap:** Most existing comments from @analogrelay, Copilot, and @xinlian12 have been addressed. Two remain open (pass-by-value and cloning suggestions from @analogrelay). All findings above are new except Finding 6 which echoes the existing pass-by-value suggestion.

See inline comments for details on each finding.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
FabianMeiswinkel: Won't fix - the change in azure_core to allow exposing a flag in Error that allow distinguishing this will be merged very soon.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
FabianMeiswinkel: Fixed in 0a612744a — Added `Drop` implementation for `LocationStateStore` using `epoch::unprotected()` + `into_owned()`. Safe because `&mut self` guarantees sole access. Then in 7c911c7ee, snapshot caching was added so the hot path avoids full clones entirely.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Fixed in aa969385a — `ExecutionContext` now distinguishes all three cases: `Initial` (first attempt), `Retry` (session retry), `RegionFailover` (failover retry with `failover_retry_count > 0`).

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Fixed in 5414e3be0 — Removed the `expire_unavailable_endpoints` clone from `resolve_endpoint`. Now checks TTL inline during the candidate loop: `account.unavailable_endpoints.get(candidate).is_some_and(|(marked_at, _)| now.saturating_duration_since(*marked_at) < ttl)`. Zero additional allocations on the hot path.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
FabianMeiswinkel: Won't fix — the change in `azure_core` to expose a flag in `Error` for distinguishing this will be merged very soon. The non-idempotent write abort is an intentional safety guard beyond the spec.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
FabianMeiswinkel: Fixed in fe0b55a1a — `sync_account_properties` now skips the CAS loop when the `AccountProperties._etag` matches the last synced value. Added `etag` field to `AccountProperties` (deserialized from `_etag` in the JSON response). On the hot path (cache hit, same etag), the CAS loop is entirely bypassed.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Won't fix — passing `RoutingDecision` by value would require cloning the endpoint beforehand for `evaluate_transport_result` (which needs `&routing.endpoint` after `build_transport_request`). Same rationale as analogrelay's earlier comment.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
FabianMeiswinkel: Updated TODO in 46e569fe4 — references `application_region` / `generate_preferred_region_list` in `azure_data_cosmos` and notes this should be wired when operations migrate to the driver's `execute_operation` API (natural cross-crate change point).

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/endpoint.rs
FabianMeiswinkel: Fixed in 6cd2658de — Added `debug_assert!(list_len > 0, "endpoint list should never be empty")` before the early return.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Fixed in 068287fc6 — Added `resolve_endpoint_falls_back_to_default_when_all_unavailable` test that creates an `AccountEndpointState` with all preferred endpoints in `unavailable_endpoints` and verifies `resolve_endpoint` returns `default_endpoint`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/components.rs
FabianMeiswinkel: Fixed in 962af1a21 — Added `advance_session_retry_single_write_routes_to_write_endpoints` and `advance_session_retry_multi_write_stays_on_preferred_endpoints` unit tests covering both topology variants.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
FabianMeiswinkel: Fixed in 7c911c7ee — Added version-based snapshot caching. A monotonic `account_version` counter is bumped on every CAS success. `snapshot()` compares the version against the cached snapshot — on the fast path (no state changes), it returns `Arc::clone()` (refcount increment only). Full clone only happens when the version has advanced.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
FabianMeiswinkel: Fixed in 2c6729309 — Replaced all "Step N" references with topic-based phrasing:
- `Step 3 adds partition-level state` → `Placeholder for partition-level state (not yet implemented)`
- `Fixed refresh throttling for Step 2` → `TODO(refresh-config): Make refresh interval configurable`
- `Step 2 applies endpoint unavailability...` → `Applies location effects (endpoint unavailability and account refresh)`
- `Step 2 keeps partition-level logic...` → `TODO(partition-routing): Apply partition-level unavailability`
- `TODO(Step 3): Wire the actual partition key range ID` → `TODO(partition-routing): Wire the actual partition key range ID`
- `Step 2 resolves from LocationSnapshot...` → descriptive doc comment

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel, line 1: Blocking: the retry location can become permanently stale after `RefreshAccountProperties` updates the account generation. `resolve_endpoint()` resets stale indices back to `0`, but `advance_location()` preserves the old generation, so later retries in the same operation keep restarting from the first endpoint in the refreshed snapshot instead of continuing to walk it. The problematic interaction is between `advance_location()` in `components.rs`, the `is_current(account.generation)` check here, and generation bumps in `build_account_endpoint_state()`. This can break Step 2 failover traversal after a topology refresh. The retry state needs to rebind to the current generation when staleness is detected, or location advancement needs to become generation-aware.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
FabianMeiswinkel, line 1: Recommendation: the `etag` fast path in `sync_account_properties()` returns before TTL pruning runs, so expired endpoint-unavailability entries may stay in the store indefinitely until account metadata changes again. Routing correctness is partly protected because endpoint selection rechecks timestamps, but the store itself never shrinks on the steady-state path. Consider pruning expired entries before the `etag` equality early return, or moving TTL cleanup into a cheap maintenance path that does not depend on metadata churn.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
FabianMeiswinkel, line 1: Recommendation: `403/3 WriteForbidden` is currently treated as a global endpoint-unavailability signal. Because routing later consults a single `unavailable_endpoints` map for all operation types, one write-forbidden response suppresses reads to that region for the full TTL as well. That looks weaker than the Java/.NET semantics, where write-forbidden typically marks the endpoint unavailable for writes without poisoning read routing. If Step 2 wants parity here, endpoint unavailability should be scoped by operation type rather than shared globally.

FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Fixed

PR #3897, opened 2026-03-10
cosmos: update azure_core reference to 0.33

analogrelay:
This updates our references to `azure_core` and `azure_identity` from 0.32 to 0.33, now that those have been released.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -389,6 +389,7 @@ dependencies = [
  "crossbeam-epoch",
  "futures",
  "h2",
+ "rand 0.10.0",
  "reqwest",
  "serde",
  "serde_json",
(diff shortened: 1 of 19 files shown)


sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -16,7 +16,7 @@ categories = ["api-bindings"]
 [dependencies]
 async-lock.workspace = true
 async-trait.workspace = true
-azure_core = { version = "0.32.0", default-features = false }
+azure_core = { version = "0.33.0", default-features = false }
```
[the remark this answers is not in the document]
FabianMeiswinkel: Noise - won't fix.

eng/scripts/verify-dependencies.rs
```diff
@@ -24,7 +24,9 @@ static EXEMPTIONS: &[(&str, &str)] = &[
     ("azure_core_test", "dotenvy"),
     ("azure_canary", "serde"),
     ("azure_data_cosmos_native", "cbindgen"),
-    // Temporary: Allow azure_data_cosmos to release depending on the latest release of azure_core and azure_identity
+    // Temporary: Allow azure_data_cosmos and friends to release depending on the latest release of azure_core and azure_identity
+    ("azure_data_cosmos_driver", "azure_core"),
+    ("azure_data_cosmos_driver", "azure_identity"),
```
heaths, line 29: This is fine and I'll sign off, but since you're already in your own branch, alternatively you can just remove the `path` from `/Cargo.toml` and keep the version on the latest released version. That's what post-GA will look like anyway. So it should merge nicely once core GA's and no more managing this script.

analogrelay: That's a good point, I'll do that

analogrelay: Ah, actually that causes some problems since I've got to do it across a bunch of packages to keep the build running. This is still the simplest solution. Once the post-GA workspace is in place, I can merge it up to this branch, fix things up, and merge back into main without much issue.

heaths, line 29: Can you open a tracking bug then?

analogrelay: #4014

PR #3910, opened 2026-03-10
RequestInstrumentationPolicy doesn't set newly created span for parenting.

### LINKED ISSUE #3424: RequestInstrumentationPolicy doesn't pass spans to the next policy
This policy starts a new span before invoking the next policy in the pipeline:
https://github.com/Azure/azure-sdk-for-rust/blob/c279b495771d259fc51a7d80f5fb87b1517e908d/sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs#L107-L119

...but it doesn't add this span to the `Context` it passes to that policy, which therefore gets a different span, if any. This probably doesn't matter in most cases today because `RequestInstrumentationPolicy` is the last policy in a default pipeline, however it could enable bugs when an application adds per-try policies (custom policies always run after default policies).

LarryOsterman:
Fixes #3424.

Before this change, a newly created request instrumentation policy span was not added to the context, which means that any spans created later in the pipeline would not correctly be parented to this request.

### DIFF
--- a/sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
+++ b/sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
@@ -117,6 +117,11 @@ impl Policy for RequestInstrumentationPolicy {
             tracer.start_span(method_str.into(), SpanKind::Client, span_attributes)
         };
 
+        // Now add the newly created span to the context. This span will become the parent span
+        // for any downstream spans created by other policies or the transport, thus establishing a correct
+        // distributed trace span hierarchy.
+        let ctx = ctx.clone().with_value(span.clone());
+
         if span.is_recording() {
             if let Some(client_request_id) = request
                 .headers()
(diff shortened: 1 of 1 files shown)


sdk/core/azure_core/src/http/policies/instrumentation/request_instrumentation.rs
```diff
@@ -117,6 +117,11 @@ impl Policy for RequestInstrumentationPolicy {
             tracer.start_span(method_str.into(), SpanKind::Client, span_attributes)
         };
 
+        // Now add the newly created span to the context. This span will become the parent span
+        // for any downstream spans created by other policies or the transport, thus establishing a correct
+        // distributed trace span hierarchy.
+        let ctx = ctx.clone().with_value(span.clone());
```
heaths, line 123: In terms of perf, I wonder if we should be checking if (our) tracing is enabled before cloning the inner `HashMap` and adding `Span`s that are otherwise useless if no one is listening. In certain scenarios you're investigating, I imagine that adds up to _some_ degradation.

LarryOsterman: We do already. See line 65 of this file - if there is no tracer, this policy is a noop.

There's a similar check (somewhat more involved) in the public api instrumentation policy.

We might also be able to check for a logger being attached to save some cycles in the case where the customer enabled distributed tracing but didn't bother to attach a logger (we already do this for some more expensive operations).

LarryOsterman: No, the `is_recording` call doesn't check for a logger :(. It might improve things a bit, but the big one is the attach of a tracer to the call.

PR #3912, opened 2026-03-10
Driver Step 5: HTTP/2 support and adaptive transport (no sharding yet)

FabianMeiswinkel:
the intend of these changes is to implemnt step Step 5: HTTP/2 support & adaptive transport (no sharding yet) of the sdk\cosmos\azure_data_cosmos_driver\docs\TRANSPORT_PIPELINE_SPEC.md spec - note that step 1 and 2 are already merged - step 3 and 4 are worked on in parallel by somebody else - so, do not be surprised that step 3 and 4 are not yet done - there should only be minimla overlap - so doing step 5 now is fine

### DIFF
--- a/sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
+++ b/sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
@@ -13,9 +13,7 @@ use azure_core::http::{AsyncRawResponse, HttpClient, Request};
 use tracing::warn;
 use url::Url;
 
-use crate::{
-    diagnostics::TransportKind, driver::cache::AccountProperties, options::Region,
-};
+use crate::{diagnostics::TransportKind, driver::cache::AccountProperties, options::Region};
 
 use super::http_client_factory::HttpVersionPolicy;
 
--- a/sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
+++ b/sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
@@ -193,9 +193,10 @@ impl CosmosTransport {
             match self.insecure_emulator_metadata_transport.get() {
                 Some(t) => t.clone(),
                 None => {
-                    let config =
-                        HttpClientConfig::metadata(&self.connection_pool).for_emulator();
-                    let client = self.http_client_factory.build(&self.connection_pool, config)?;
+                    let config = HttpClientConfig::metadata(&self.connection_pool).for_emulator();
+                    let client = self
+                        .http_client_factory
+                        .build(&self.connection_pool, config)?;
                     let t = AdaptiveTransport::from_policy(config.version_policy, client);
                     self.insecure_emulator_metadata_transport
                         .get_or_init(|| t)
@@ -255,7 +256,9 @@ impl CosmosTransport {
                 None => {
                     let config =
                         HttpClientConfig::dataplane_gateway(&self.connection_pool).for_emulator();
-                    let client = self.http_client_factory.build(&self.connection_pool, config)?;
+                    let client = self
+                        .http_client_factory
+                        .build(&self.connection_pool, config)?;
                     let t = AdaptiveTransport::from_policy(config.version_policy, client);
                     self.insecure_emulator_dataplane_transport
                         .get_or_init(|| t)
@@ -281,13 +284,12 @@ impl CosmosTransport {
             let transport = match self.dataplane_gateway20_transport.get() {
                 Some(t) => t.clone(),
                 None => {
-                    let config =
-                        HttpClientConfig::dataplane_gateway20(&self.connection_pool);
-                    let client = self.http_client_factory.build(&self.connection_pool, config)?;
+                    let config = HttpClientConfig::dataplane_gateway20(&self.connection_pool);
+                    let client = self
+                        .http_client_factory
+                        .build(&self.connection_pool, config)?;
                     let t = AdaptiveTransport::from_policy(config.version_policy, client);
-                    self.dataplane_gateway20_transport
-                        .get_or_init(|| t)
-                        .clone()
+                    self.dataplane_gateway20_transport.get_or_init(|| t).clone()
                 }
             };
             Ok(TransportContext {
@@ -581,7 +583,10 @@ pub(crate) mod tests {
             AccountEndpoint::try_from("https://myaccount.documents.azure.com:443/").unwrap();
 
         assert!(matches!(
-            transport.get_metadata_transport(&endpoint).unwrap().transport,
+            transport
+                .get_metadata_transport(&endpoint)
+                .unwrap()
+                .transport,
             AdaptiveTransport::Gateway(_)
         ));
     }
@@ -597,7 +602,10 @@ pub(crate) mod tests {
             AccountEndpoint::try_from("https://myaccount.documents.azure.com:443/").unwrap();
 
         assert!(matches!(
-            transport.get_metadata_transport(&endpoint).unwrap().transport,
+            transport
+                .get_metadata_transport(&endpoint)
+                .unwrap()
+                .transport,
             AdaptiveTransport::Gateway(_)
         ));
     }
@@ -612,8 +620,9 @@ pub(crate) mod tests {
         let endpoint =
             AccountEndpoint::try_from("https://myaccount.documents.azure.com:443/").unwrap();
 
-        let ctx =
-            transport.get_dataplane_transport(&endpoint, &account_properties_without_thin_client()).unwrap();
+        let ctx = transport
+            .get_dataplane_transport(&endpoint, &account_properties_without_thin_client())
+            .unwrap();
         assert!(matches!(ctx.transport, AdaptiveTransport::Gateway(_)));
         assert!(ctx.thin_client_overrides.is_none());
     }
@@ -629,7 +638,9 @@ pub(crate) mod tests {
         let endpoint =
             AccountEndpoint::try_from("https://myaccount.documents.azure.com:443/").unwrap();
 
-        let ctx = transport.get_dataplane_transport(&endpoint, &account_properties_with_thin_client()).unwrap();
+        let ctx = transport
+            .get_dataplane_transport(&endpoint, &account_properties_with_thin_client())
+            .unwrap();
         assert!(matches!(ctx.transport, AdaptiveTransport::Gateway20(_)));
         assert!(ctx.thin_client_overrides.is_some());
     }
@@ -645,7 +656,9 @@ pub(crate) mod tests {
         let endpoint =
             AccountEndpoint::try_from("https://myaccount.documents.azure.com:443/").unwrap();
 
-        let ctx = transport.get_dataplane_transport(&endpoint, &account_properties_without_thin_client()).unwrap();
+        let ctx = transport
+            .get_dataplane_transport(&endpoint, &account_properties_without_thin_client())
+            .unwrap();
         assert!(matches!(ctx.transport, AdaptiveTransport::Gateway(_)));
         assert!(ctx.thin_client_overrides.is_none());
     }
@@ -661,7 +674,9 @@ pub(crate) mod tests {
         let endpoint =
             AccountEndpoint::try_from("https://myaccount.documents.azure.com:443/").unwrap();
 
-        let ctx = transport.get_dataplane_transport(&endpoint, &account_properties_with_thin_client()).unwrap();
+        let ctx = transport
+            .get_dataplane_transport(&endpoint, &account_properties_with_thin_client())
+            .unwrap();
         assert!(matches!(ctx.transport, AdaptiveTransport::Gateway(_)));
         assert!(ctx.thin_client_overrides.is_none());
     }
@@ -677,10 +692,12 @@ pub(crate) mod tests {
         let endpoint =
             AccountEndpoint::try_from("https://myaccount.documents.azure.com:443/").unwrap();
 
-        let ctx = transport.get_dataplane_transport(
-            &endpoint,
-            &account_properties_with_only_malformed_thin_client(),
-        ).unwrap();
+        let ctx = transport
+            .get_dataplane_transport(
+                &endpoint,
+                &account_properties_with_only_malformed_thin_client(),
+            )
+            .unwrap();
 
         assert!(matches!(ctx.transport, AdaptiveTransport::Gateway(_)));
         assert!(ctx.thin_client_overrides.is_none());


sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
```diff
@@ -130,115 +158,146 @@ impl CosmosTransport {
...
-            if !pool.is_proxy_allowed() {
-                builder = builder.no_proxy();
-            }
-            // When proxy is allowed, reqwest automatically respects HTTP_PROXY/HTTPS_PROXY env vars
-
-            // Local address binding
-            if let Some(local_addr) = pool.local_address() {
-                builder = builder.local_address(local_addr);
+        } else if self.connection_pool.is_gateway20_allowed()
+            && account_properties.has_thin_client_endpoints()
```
FabianMeiswinkel, line 273: This Gateway20 enablement check looks too broad for the operations we currently route through the dataplane path. `uses_dataplane_pipeline()` still includes stored procedure execute, so once `thinClient*Locations` are present we can send stored procedure execution over Gateway 2.0 as well. That appears broader than the other SDKs' thin-client eligibility, and if Gateway 2.0 does not support stored procedure execute this becomes a rollout-dependent regression. Can we add an explicit Gateway20 eligibility gate here so only supported dataplane operations opt into thin-client routing, plus a test covering stored procedure execute with thin-client metadata present?

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -296,14 +298,23 @@ fn preferred_endpoints_for_attempt<'a>(
...
-        let mut base = routing.endpoint.url().clone();
+        let mut base = thin_client_overrides
+            .and_then(|overrides| {
+                routing
+                    .endpoint
+                    .region()
+                    .and_then(|region| overrides.get(region))
+            })
+            .cloned()
+            .unwrap_or_else(|| routing.endpoint.url().clone());
```
FabianMeiswinkel: This fallback worries me when paired with `Gateway20` selection. `get_dataplane_transport()` turns on HTTP/2 prior-knowledge as soon as the account has any thin-client endpoints, but this code silently falls back to the normal gateway URL whenever the chosen region has no override entry. That can leave us using the Gateway20 transport against the standard gateway endpoint on partial/malformed account metadata. Could we fail closed here, or tighten the selection logic so Gateway20 is only chosen when we have a usable override for the routed region?

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
FabianMeiswinkel: Won't fix - this is intentional since Stored Proc execution is supporte din GW2.0 endpoint now in the service (and Java and .Net are also making changes to use GW2.0 for Stored Proc execution)

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: thinclient read and write endpoints would either both exist or none in the account properties - but I am fine with making this validation more explicit.

sdk/cosmos/azure_data_cosmos_driver/Cargo.toml
```diff
@@ -49,4 +48,15 @@ workspace = true
...
+  "dep:reqwest",
+  "reqwest/http2",
+  "azure_core/reqwest",
+  "azure_core/tokio",
+]
+reqwest_native_tls = [
+  "reqwest",
+  "reqwest/native-tls",
+  "azure_core/reqwest_native_tls",
+]
```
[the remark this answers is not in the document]
FabianMeiswinkel: Won't fix for this thread. We are intentionally keeping 
eqwest and 
eqwest_native_tls as default features to stay aligned with zure_core and because the driver currently requires a known HTTP/2-capable transport rather than an arbitrary custom client. No code change was made for this thread.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
```diff
@@ -142,7 +143,7 @@ pub(crate) fn evaluate_transport_retry(
 /// This is the core transport loop described in §5.2 of the spec.
 pub(crate) async fn execute_transport_pipeline(
     request: TransportRequest,
-    http_client: &dyn azure_core::http::HttpClient,
+    transport: &AdaptiveTransport,
     credential: &Credential,
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed in https://github.com/Azure/azure-sdk-for-rust/commit/3ff979e69c70e6458fd83ecf880f2fb3c44fbdc1 . Updated the transport pipeline doc comment to refer to the selected transport / AdaptiveTransport abstraction instead of a raw HttpClient.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -459,10 +459,23 @@ impl CosmosDriver {
         let operation_type = operation.operation_type();
         let resource_type = operation.resource_type();
         let is_dataplane = uses_dataplane_pipeline(resource_type, operation_type);
-        let http_client = if is_dataplane {
-            transport.get_dataplane_http_client(&endpoint)
+        let adaptive_transport = if is_dataplane {
```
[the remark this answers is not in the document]
FabianMeiswinkel: Fixed in https://github.com/Azure/azure-sdk-for-rust/commit/3ff979e69c70e6458fd83ecf880f2fb3c44fbdc1 . Updated the Step 5 comment to describe selecting the adaptive transport context rather than a raw HttpClient.

FabianMeiswinkel:
Review follow-up summary:

- Cargo default features comment: Won't fix. Keeping 
eqwest and 
eqwest_native_tls enabled by default is intentional so the driver stays aligned with zure_core and with the current built-in HTTP/2 transport model.
- Transport pipeline doc comment: Fixed in https://github.com/Azure/azure-sdk-for-rust/commit/3ff979e69c70e6458fd83ecf880f2fb3c44fbdc1 by updating the wording to reference the selected transport / AdaptiveTransport.
- Cosmos driver Step 5 comment: Fixed in https://github.com/Azure/azure-sdk-for-rust/commit/3ff979e69c70e6458fd83ecf880f2fb3c44fbdc1 by updating the wording to reference the adaptive transport context.
- Gateway20 stored procedure eligibility: Won't fix. Stored procedure execute is intentionally allowed to use Gateway20 because the service supports it now; the temporary client-side restriction was removed.
- Gateway20 thin-client override fallback: Fixed in https://github.com/Azure/azure-sdk-for-rust/commit/3ff979e69c70e6458fd83ecf880f2fb3c44fbdc1 by failing closed when a regional thin-client override is missing and by falling back to the standard gateway when thin-client metadata parses as empty or malformed. Added focused unit coverage for missing regional overrides and malformed thin-client metadata.

FabianMeiswinkel:
Review follow-up summary:

- Cargo default features comment: Won''t fix. Keeping reqwest and reqwest_native_tls enabled by default is intentional so the driver stays aligned with azure_core and with the current built-in HTTP/2 transport model.
- Transport pipeline doc comment: Fixed in https://github.com/Azure/azure-sdk-for-rust/commit/3ff979e69c70e6458fd83ecf880f2fb3c44fbdc1 by updating the wording to reference the selected transport / AdaptiveTransport.
- Cosmos driver Step 5 comment: Fixed in https://github.com/Azure/azure-sdk-for-rust/commit/3ff979e69c70e6458fd83ecf880f2fb3c44fbdc1 by updating the wording to reference the adaptive transport context.
- Gateway20 stored procedure eligibility: Won''t fix. Stored procedure execute is intentionally allowed to use Gateway20 because the service supports it now; the temporary client-side restriction was removed.
- Gateway20 thin-client override fallback: Fixed in https://github.com/Azure/azure-sdk-for-rust/commit/3ff979e69c70e6458fd83ecf880f2fb3c44fbdc1 by failing closed when a regional thin-client override is missing and by falling back to the standard gateway when thin-client metadata parses as empty or malformed. Added focused unit coverage for missing regional overrides and malformed thin-client metadata.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// The adaptive transport to use for HTTP requests.
+    pub transport: AdaptiveTransport,
+    /// Gateway 2.0 (thin client) endpoint overrides, keyed by region.
+    /// Present only when Gateway 2.0 is enabled and the account has
+    /// thin-client endpoints. Contains **both** read and write locations
+    /// merged into a single map so that retries across read/write
+    /// endpoint lists do not miss regions.
+    pub thin_client_overrides: Option<Arc<HashMap<Region, Url>>>,
+    /// Whether Gateway 2.0 was selected for this transport.
+    #[allow(dead_code)] // Used for diagnostics and future routing decisions.
```
FabianMeiswinkel: Why dead code - either use in diagnostics or remove?

FabianMeiswinkel: Fixed

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -296,14 +298,25 @@ fn preferred_endpoints_for_attempt<'a>(
...
     activity_id: &ActivityId,
     execution_context: ExecutionContext,
     deadline: Option<Instant>,
 ) -> azure_core::Result<TransportRequest> {
     let resource_ref = operation.resource_reference();
     let request_path = resource_ref.request_path();
     let url = {
-        let mut base = routing.endpoint.url().clone();
+        let mut base = match (thin_client_overrides, routing.endpoint.region()) {
+            (Some(overrides), Some(region)) => overrides.get(region).cloned().ok_or_else(|| {
```
FabianMeiswinkel: **🔴 Potential issue: Missing thin-client region causes hard operation failure during failover**

When Gateway 2.0 is selected and a retry routes to a region not present in the thin-client overrides map, this returns a non-retryable `ErrorKind::Other` error. During a regional failover (region exists in `readableLocations`/`writableLocations` but not yet in `thinClient*Locations` — possible during gradual Gateway 2.0 rollout), the user gets a hard failure with a cryptic error message instead of graceful degradation.

The operation pipeline treats this as a stage-3 failure and surfaces it directly to the caller — it cannot retry because the transport is already locked to `Gateway20`.

**Suggestion:** Fall back to the standard gateway URL from `routing.endpoint.url()` when the thin-client override is missing for the target region, and log a warning indicating the fallback. This matches how `get_dataplane_transport` already falls back to gateway when all thin-client URLs are malformed.

```rust
(Some(overrides), Some(region)) => overrides
    .get(region)
    .cloned()
    .unwrap_or_else(|| {
        warn!(
            %region,
            "No thin-client endpoint override for region; falling back to standard gateway URL"
        );
        routing.endpoint.url().clone()
    }),
```

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
```diff
@@ -130,115 +158,217 @@ impl CosmosTransport {
...
+                .dataplane_gateway20_transport
+                .get_or_init(|| {
+                    let config = HttpClientConfig::dataplane_gateway20(&self.connection_pool);
+                    AdaptiveTransport::from_policy(
+                        config.version_policy,
+                        self.http_client_factory
+                            .build(&self.connection_pool, config)
+                            .expect("failed to create Gateway 2.0 dataplane client"),
+                    )
+                })
```
FabianMeiswinkel: **🟡 Recommendation: `OnceLock` + `.expect()` can panic at runtime in production**

The lazy transport initialization for Gateway 2.0 (and emulator paths at lines 197-206) uses `.expect()` inside `OnceLock::get_or_init`. If the reqwest client builder fails (e.g., TLS library not available, conflicting settings), this panics the entire process.

Unlike `CosmosTransport::new()` which propagates errors as `Result`, these lazy paths will crash on first use. Gateway 2.0 is only tried when the account has thin-client endpoints, so this panic would surface only in specific deployment configurations — making it hard to reproduce and debug.

**Suggestion:** Change `get_dataplane_transport` to return `azure_core::Result<TransportContext>` so the error propagates to callers, or use `OnceLock<Result<AdaptiveTransport, ...>>` and propagate the stored error on subsequent calls.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
```diff
@@ -130,115 +158,217 @@ impl CosmosTransport {
...
-            builder = builder.timeout(request_timeout);
-
-            // Proxy settings
-            if !pool.is_proxy_allowed() {
-                builder = builder.no_proxy();
+        } else if self.connection_pool.is_gateway20_allowed()
+            && account_properties.has_thin_client_endpoints()
+        {
+            let overrides = thin_client_endpoint_overrides(account_properties);
+            if overrides.is_empty() {
```
FabianMeiswinkel, line 276: **🟡 Recommendation: `thin_client_endpoint_overrides` recomputed on every operation call**

`get_dataplane_transport` calls `thin_client_endpoint_overrides(account_properties)` on every invocation, parsing all thin-client URLs and building a new `HashMap` each time.

For a high-throughput driver, every operation performs URL parsing and allocation. While each individual parse is cheap (~50ns per URL), at thousands of operations/second, the repeated allocation of `HashMap<Region, Url>` adds up. The Gateway 2.0 transport itself is cached in `OnceLock`, but the overrides are not.

**Suggestion:** Cache the overrides map alongside the transport, keyed by the account properties etag or generation, and recompute only when account properties change rather than per-operation.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
```diff
@@ -130,115 +158,217 @@ impl CosmosTransport {
...
+                return TransportContext {
+                    transport: self.dataplane_gateway_transport.clone(),
+                    thin_client_overrides: None,
```
FabianMeiswinkel, line 280: **🟡 Recommendation: No tracing when falling back from Gateway 2.0 to standard gateway due to all-malformed URLs**

When `is_gateway20_allowed` is true and `has_thin_client_endpoints()` is true, but `thin_client_endpoint_overrides()` returns an empty map (all URLs are malformed), the code silently falls back to the standard gateway transport. This is a significant operational event — the driver intended to use Gateway 2.0 but couldn't. Without a log entry, operators debugging latency or routing issues would have no visibility into why Gateway 2.0 wasn't activated.

**Suggestion:** Add a `warn!` before the fallback return:
```rust
if overrides.is_empty() {
    warn!("Gateway 2.0 enabled but all thin-client endpoint URLs are malformed; falling back to standard gateway transport");
    return TransportContext { ... };
}
```

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
```diff
@@ -171,6 +172,10 @@ pub(crate) async fn execute_transport_pipeline(
             execution_context,
             pipeline_type,
             transport_security,
+            match transport {
+                AdaptiveTransport::Gateway(_) => TransportKind::Gateway,
```
FabianMeiswinkel: **🟡 Recommendation: Duplicate `TransportKind` derivation logic**

This manually matches on `AdaptiveTransport` variants to produce `TransportKind`. `AdaptiveTransport` already has a `transport_kind()` method (returns `&'static str`), but there's no direct mapping to `TransportKind`. When Step 6 adds new transport variants (e.g., sharded), every place that manually matches on `AdaptiveTransport` needs updating — a maintenance risk.

**Suggestion:** Add a `fn diagnostics_kind(&self) -> TransportKind` method to `AdaptiveTransport` and use it here and anywhere else that needs the classification, so there's a single authoritative conversion.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
```diff
@@ -0,0 +1,162 @@
...
+/// reroute reads to write endpoints — if only read locations were included,
+/// the write-region URL override would be missing, causing requests to be
+/// sent to the wrong gateway endpoint while using HTTP/2-only transport.
+pub(crate) fn thin_client_endpoint_overrides(
+    properties: &AccountProperties,
+) -> HashMap<Region, Url> {
+    properties
+        .thin_client_readable_locations
+        .iter()
+        .chain(properties.thin_client_writable_locations.iter())
```
FabianMeiswinkel, line 109: **🟡 Recommendation: HashMap silently overwrites when same region appears in both readable and writable thin-client lists**

`collect()` uses last-write-wins semantics for `HashMap`. Since writable locations are chained after readable ones, the writable endpoint silently wins if the same region appears in both lists with different URLs. In a multi-write account or during metadata transitions, readable and writable thin-client locations for the same region could conceivably have different endpoints.

**Suggestion:** Either (a) log a warning when a duplicate region has conflicting URLs, or (b) document the last-write-wins behavior explicitly. Consider also adding a test for the duplicate-region-across-lists case (same region in both `thinClientReadableLocations` and `thinClientWritableLocations`).

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -454,15 +454,18 @@ impl CosmosDriver {
...
         let operation_type = operation.operation_type();
         let resource_type = operation.resource_type();
         let is_dataplane = uses_dataplane_pipeline(resource_type, operation_type);
-        let http_client = if is_dataplane {
-            transport.get_dataplane_http_client(&endpoint)
+        let transport_context = if is_dataplane {
+            transport.get_dataplane_transport(&endpoint, account_properties.as_ref())
         } else {
-            transport.get_metadata_http_client(&endpoint)
+            super::transport::adaptive_transport::TransportContext {
```
FabianMeiswinkel: **🟡 Recommendation: Asymmetric public API — `get_metadata_transport` vs `get_dataplane_transport`**

`get_metadata_transport` returns `AdaptiveTransport` while `get_dataplane_transport` returns `TransportContext`. The caller here manually wraps the result, coupling it to `TransportContext`'s internal structure and the knowledge that metadata never uses thin-client overrides. If `TransportContext` gains new fields (likely in Step 6), every caller constructing it manually must be updated.

**Suggestion:** Either have `get_metadata_transport` return `TransportContext` directly, or add a `TransportContext::metadata(transport)` constructor to encapsulate this.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
```diff
@@ -0,0 +1,162 @@
...
+        .filter_map(
+            |region| match Url::parse(&region.database_account_endpoint) {
+                Ok(url) => Some((region.name.clone(), url)),
```
FabianMeiswinkel: **🟡 Recommendation: No HTTPS scheme validation on thin-client endpoint URLs**

The URL is parsed but the scheme is not validated. A non-HTTPS URL (e.g., `http://...`) would be accepted and used with the HTTP/2-only transport. Gateway 2.0 with `http2_prior_knowledge()` over HTTP sends data in plaintext.

While the service should never return non-HTTPS endpoints, defense-in-depth at this system boundary would prevent accidental plaintext communication if metadata is ever corrupted or spoofed.

**Suggestion:** After parsing the URL, validate `url.scheme() == "https"` and log a warning + skip the entry on mismatch (similar to how malformed URLs are already skipped).

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/http_client_factory.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::options::ConnectionPoolOptionsBuilder;
+
+    #[test]
+    fn metadata_uses_http11_only_when_http2_disabled() {
```
FabianMeiswinkel, line 75: **🟢 Suggestion: `HttpClientConfig` tests only cover the `Http11Only` branch**

These two tests verify `is_http2_allowed(false)` → `Http11Only`. Consider also testing:

- `metadata` with `is_http2_allowed(true)` → `Http2Preferred`
- `dataplane_gateway20` → always `Http2Only` regardless of pool settings
- `for_emulator()` → sets `for_emulator = true`

FabianMeiswinkel (commented):
## Review: Driver Step 5 — HTTP/2 support and adaptive transport

Overall the design is solid and well-tested. The adaptive transport abstraction is clean, the feature gating is well-designed (compile-time capability vs. runtime configuration), and the crossbeam-epoch CAS fix is a critical correctness win.

**Positive highlights:**
- The CAS loop fix correctly retires the old value instead of the newly-installed pointer — this was a use-after-free
- The new Drop implementation (`std::mem::replace` + `try_into_owned()`) is more obviously correct than the old `epoch::unprotected()` approach
- Good edge case test coverage for thin-client routing (regional override, global endpoint, missing override, malformed URLs, all-malformed fallback, gateway20 disabled)
- Clean feature hierarchy: `default` → `reqwest_native_tls` → `reqwest`, with `reqwest/http2` auto-enabled

**Key concern:** The main issue is around failover resilience when Gateway 2.0 is selected but a retry targets a region without thin-client endpoints — this returns a hard error instead of gracefully degrading to standard gateway. See inline comment on `operation_pipeline.rs`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: ACK - makes sense and is a good proposal

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
FabianMeiswinkel: ACK - makes sense and is a good proposal

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
FabianMeiswinkel: ACK - do this

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
FabianMeiswinkel: ACk - do this

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
FabianMeiswinkel: ACK - do this

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
FabianMeiswinkel: This should never happen - would be a bug in the service - so warning log is sufficient

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
FabianMeiswinkel: Add get_metadata_transport and have it return TransportContext

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
FabianMeiswinkel: validate https and log a warning

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/http_client_factory.rs
FabianMeiswinkel: Ok - add those tests

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Fixed in 6d7bfd97d — changed the hard `ErrorKind::Other` error to a graceful fallback: when the thin-client override is missing for the target region, we now `warn!` and fall back to `routing.endpoint.url()`. Updated the test to verify the fallback produces the correct standard gateway URL.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
FabianMeiswinkel: Fixed in 0998b5df9 — replaced all three `.expect()` calls inside `OnceLock::get_or_init` with a get-or-try-init pattern: the fallible `http_client_factory.build()` runs outside the closure and errors propagate via `?`. Both `get_metadata_transport` and `get_dataplane_transport` now return `azure_core::Result<...>`, and callers propagate with `?`. No more panics on HTTP client builder failures.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
FabianMeiswinkel: Fixed in 6e63a8059 — added a `Mutex<Option<CachedOverrides>>` to `CosmosTransport` that caches the parsed thin-client endpoint overrides map keyed by the account properties `etag`. The hot path now only requires a mutex lock + `Arc::clone` when the etag hasn't changed, avoiding URL parsing and HashMap allocation on every operation.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
FabianMeiswinkel: Fixed in 73af9f2f9 — added a `warn!` log before the empty-overrides fallback: `"Gateway 2.0 enabled but all thin-client endpoint URLs are malformed; falling back to standard gateway transport"`.

FabianMeiswinkel (commented):
Stale pending review — submitting to unblock inline replies.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
FabianMeiswinkel: Fixed in d04fc7a5a — added `diagnostics_kind() -> TransportKind` to `AdaptiveTransport` and replaced the manual match in `transport_pipeline.rs`. Now there's a single authoritative mapping for Step 6 to update.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
FabianMeiswinkel: Fixed in 50618cb38 — replaced `.collect()` with an explicit loop that detects duplicate regions. Same URL → deduplicates silently. Conflicting URLs → keeps the first (readable) entry and emits a `warn!`. Added tests for both cases.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
```diff
@@ -142,22 +144,16 @@ impl LocationStateStore {
     }
 
     /// Returns a snapshot of account and partition state.
-    ///
-    /// On the fast path (no state changes since last call), this returns
-    /// `Arc::clone()` of the cached snapshot (refcount increment only).
-    /// A full clone is only performed when the version counter has advanced.
     pub fn snapshot(&self) -> LocationSnapshot {
```
FabianMeiswinkel: ACK - makes sense - add it back

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
```diff
@@ -0,0 +1,162 @@
...
+            .finish_non_exhaustive()
+    }
+}
+
+/// Bundles an [`AdaptiveTransport`] with optional Gateway 2.0 endpoint
+/// overrides, determined once per operation and threaded through the pipeline.
+#[derive(Clone, Debug)]
+pub(crate) struct TransportContext {
+    /// The adaptive transport to use for HTTP requests.
+    pub transport: AdaptiveTransport,
```
FabianMeiswinkel: Won't fix - it is pub(crate) no need to make fields private

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
```diff
@@ -238,9 +232,10 @@ impl LocationStateStore {
                 Ordering::Acquire,
                 &guard,
             ) {
-                Ok(old) => {
-                    // SAFETY: old pointer is detached after successful exchange.
-                    unsafe { guard.defer_destroy(old) };
+                Ok(_new) => {
```
FabianMeiswinkel: Not critical - but ok to fix - so please fix it.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
FabianMeiswinkel: Fixed in e96af37eb — `get_metadata_transport` now returns `azure_core::Result<TransportContext>` matching `get_dataplane_transport`. The manual `TransportContext` wrapping in `cosmos_driver.rs` is eliminated.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
FabianMeiswinkel: Fixed in 91a92db15 — added scheme validation after URL parsing: non-HTTPS thin-client URLs are now skipped with a `warn!` log, preventing accidental plaintext communication via HTTP/2 prior knowledge. Added a test for the rejection case.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
xinlian12: **🔴 Blocking** · Correctness: Misleading Variable Name in Unsafe Code

**CAS `Ok(_new)` binding name is factually wrong and dangerous in unsafe code**

`compare_exchange` returns `Ok(previous_value)` — the old value that was just swapped out. Binding it as `_new` is misleading: it suggests the return value is the newly installed pointer, when in fact it is the old one.

The `defer_destroy(current)` call is correct, but if someone later reads `Ok(_new)` and assumes it is the new value, they might `defer_destroy(_new)` instead — destroying live data.

Use `Ok(_)` since the value is unused.

---

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
xinlian12, line 86: **🟡 Recommendation** · API Design: Encapsulation
`[History Rule R-006: Make struct fields private with public getters]`

**`TransportContext` pub fields break the encapsulation pattern used everywhere else in the driver**

Every other struct in the driver uses private fields with `pub` getters (`RequestDiagnostics`, `LocationSnapshot`, `CosmosEndpoint`, etc.). `TransportContext` exposes `transport` and `thin_client_overrides` as `pub` fields directly. Even though this is `pub(crate)`, the team consistently enforces encapsulation for internal types too.

Consider making fields private with getters, or documenting why this struct is an exception.

> 📏 **Rule R-006** (Confidence: High, Adoption: 79%): *"Should this field be private with a public getter instead of directly exposed?"* — This rule is the 6th most common review pattern, consistently enforced even for `pub(crate)` types. Structs like `RequestDiagnostics` and `CosmosEndpoint` follow this pattern throughout the driver codebase.

---

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
xinlian12, line 153: **🟡 Recommendation** · Maintainability: Documentation Removed
`[History Rule R-009: Add doc comments to public APIs]`

**`snapshot()` lost its performance documentation**

The diff removes doc comments explaining the fast-path (`Arc::clone` when version unchanged) vs slow-path (full clone when state changed). This optimization rationale helps future readers understand *why* the lock+version-check pattern exists.

Consider keeping the fast-path/slow-path explanation.

> 📏 **Rule R-009** (Confidence: High, Adoption: 89%): *"Does this public API have adequate documentation?"* — Doc comment quality is the most consistently enforced review pattern in the codebase at 89% adoption. The removed comments documented a non-obvious performance optimization that would be hard to rediscover.

---

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/http_client_factory.rs
```diff
@@ -0,0 +1,177 @@
...
+            .unwrap();
+
+        assert_eq!(
+            HttpClientConfig::dataplane_gateway(&pool).version_policy,
+            HttpVersionPolicy::Http11Only
+        );
+    }
+}
+
+pub(crate) trait HttpClientFactory: fmt::Debug + Send + Sync {
```
xinlian12, line 159: **🟡 Recommendation** · Maintainability: Missing Documentation
`[History Rule R-009: Add doc comments to public APIs]`

**`HttpClientFactory` trait has no doc comment**

This trait is the key extension point for pluggable HTTP transport creation. A doc comment explaining when/why someone would implement a custom factory (e.g., testing, custom TLS config, non-reqwest HTTP stacks) would help future maintainers.

> 📏 **Rule R-009** (Confidence: High, Adoption: 89%): *"Does this public API have adequate documentation?"* — Doc comments on public APIs have 89% adoption — the most consistently enforced review pattern. This trait is the main abstraction boundary for transport creation.

---

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub(crate) fn from_policy(policy: HttpVersionPolicy, client: Arc<dyn HttpClient>) -> Self {
+        match policy {
+            HttpVersionPolicy::Http11Only | HttpVersionPolicy::Http2Preferred => {
+                Self::Gateway(client)
+            }
+            HttpVersionPolicy::Http2Only => Self::Gateway20(client),
+        }
+    }
+
+    fn transport_kind(&self) -> &'static str {
```
xinlian12: **🟢 Suggestion** · API Design: Type Consistency

**`transport_kind()` returns `&str` but `TransportKind` enum exists**

`AdaptiveTransport::transport_kind()` returns a raw string but the diagnostics module defines a `TransportKind` enum with the same variants. This method could return `TransportKind` instead, since it implements `Display` and `AsRef<str>`.

---

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
FabianMeiswinkel: Won't fix - not needed for pub(crate) structs

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/http_client_factory.rs
FabianMeiswinkel: Fixed in ef819af9b — added 4 new tests covering the missing branches: `metadata` with http2 allowed → `Http2Preferred`, `dataplane_gateway` with http2 allowed → `Http2Preferred`, `dataplane_gateway20` → always `Http2Only` regardless of pool settings, `for_emulator()` → sets `for_emulator = true`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
FabianMeiswinkel: Makes sense - fix

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
FabianMeiswinkel: ACK fix (although not blocking)

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/http_client_factory.rs
FabianMeiswinkel: Won't fix - it is pub(crate)

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
FabianMeiswinkel: ACK fix

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
FabianMeiswinkel: Fixed in 0cfd2a7aa — renamed `Ok(_new)` to `Ok(_)` since the value is unused.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
FabianMeiswinkel: Fixed in 0cfd2a7aa — restored the fast-path/slow-path doc comment on `snapshot()`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
FabianMeiswinkel: Fixed in 0cfd2a7aa — consolidated `transport_kind()` into `diagnostics_kind()` which now returns `TransportKind` everywhere (including in the `Debug` impl via `.as_ref()`). Single authoritative method.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/adaptive_transport.rs
FabianMeiswinkel: Won't fix — `TransportContext` is `pub(crate)` and used as a simple data carrier between `CosmosTransport` and the operation pipeline. Adding getters for a `pub(crate)` struct with only 2 fields would add boilerplate without meaningful encapsulation benefit.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/http_client_factory.rs
FabianMeiswinkel: Won't fix — `HttpClientFactory` is `pub(crate)` and not a publicly extensible API. The module doc comment at the top of the file already explains its purpose. Adding trait-level docs for an internal-only trait would be over-documenting.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -296,14 +299,26 @@ fn preferred_endpoints_for_attempt<'a>(
...
+    thin_client_overrides: Option<&std::collections::HashMap<Region, Url>>,
     activity_id: &ActivityId,
     execution_context: ExecutionContext,
     deadline: Option<Instant>,
 ) -> azure_core::Result<TransportRequest> {
     let resource_ref = operation.resource_reference();
     let request_path = resource_ref.request_path();
     let url = {
-        let mut base = routing.endpoint.url().clone();
+        let mut base = match (thin_client_overrides, routing.endpoint.region()) {
```
analogrelay, line 310: It feels to me like thin client overrides should be done during _routing_, not during execution. I know it's not the same as what we normally consider "routing" but it's still about selecting the specific endpoint URL to send the request to. That shouldn't be a transport concern IMO. The transport certainly needs to _know_ what kind of URL it's sending the request to (so we still need to pass the gateway version down to the transport) but I don't think it should be responsible for selecting the endpoint.

analogrelay: Honestly, I'd expect thin client endpoints to end up as part of location state (each region would have a Gateway v1 endpoint and an optional Gateway v2 endpoint)

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
```diff
@@ -87,17 +87,19 @@ impl std::fmt::Debug for LocationStateStore {
 
 impl Drop for LocationStateStore {
     fn drop(&mut self) {
-        // SAFETY: We have `&mut self`, so no concurrent access is possible.
-        // `epoch::unprotected()` is safe here because we are the sole owner.
+        // crossbeam_epoch::Atomic<T> intentionally has no destructor.
+        // Detach each field and convert to Owned for immediate deallocation.
+        // Safety: &mut self in Drop guarantees exclusive access.
+        let account = std::mem::replace(&mut self.account, Atomic::null());
+        let partitions = std::mem::replace(&mut self.partitions, Atomic::null());
```
analogrelay, line 94: Ok, this is clear to me now. I'm OK with it, since we're basically copying crossbeam-recommended logic.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: As discussed offline - ACK and agreed - but will be done as follow-up PR.

FabianMeiswinkel: See https://github.com/Azure/azure-sdk-for-rust/issues/3926

Issue #3467, opened 2025-12-18
TTL can't be set to negative


jamescarterbell:
### Bug Title

TTL can't be set to negative

### Crate Name

azure_data_cosmos

### Crate Version

.29

### Description

https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/time-to-live#time-to-live-configurations

```
If TTL is set to "n" on a container, then the items in that container expire after n seconds. If there are items in the same container that have their own time to live, set to -1 (indicating they don't expire). If some items override the TTL setting with a different number, these items expire based on their own configured TTL value.

If TTL isn't set on a container, then the time to live on an item in this container has no effect.

If TTL on a container is set to -1, an item in this container that has the time to live set to n expires after n seconds, and remaining items don't expire.
```

According to the docs, to use the ttl field on items you need to set ttl on the container to -1.

However, `ContainerProperties::default_ttl` uses `std::time::Duration`, which doesn't support negative durations.

Further, the [serialization](https://docs.rs/azure_data_cosmos/latest/src/azure_data_cosmos/models/container_properties.rs.html#17) defaults to not writing the expiry, instead of `-1` (I think this is the correct thing to do, but I'm adding this note since at the very least doing this would fix the issue I'm htting :))

### Steps to Reproduce

1. Construct a [ContainerProperties](https://docs.rs/azure_data_cosmos/latest/azure_data_cosmos/models/struct.ContainerProperties.html)
2. Try to set `default_ttl` to a negative value.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/Azure/azure-sdk-for-rust/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

11alieberman:
I think there is also an issue related to this bug in the latest version of the Cosmos SDK 0.31.0, where if we have a container with TTL enabled but no default (so -1 is the default), when it now tries to deserialize the container properties when a request is made, we get a deserialization error because it tries to deserialize a -1 into a u64, which clearly will not work.

invalid value: integer `-1`, expected u64 at line 1 column 241

This completely breaks everything for us since we use TTL quite a lot, but defaulting to -1.

11alieberman:
[...]

PR #3915, opened 2026-03-11
Use TTL enum instead of straight duration

11alieberman:
Right now, it looks like the container TTL is modeled as an optional Duration. This could currently represent two states, where a None would signify there is no TTL and Some Duration would signify that there exists a duration that is non-negative, since Duration cannot represent a negative time.

This does not actually cover all the TTL use cases as defined https://learn.microsoft.com/azure/cosmos-db/time-to-live#time-to-live-configurations, namely when a container has TTL enabled, but there is no default set. This is represented as a ttl value of -1. Currently when the container TTL is deserialized, it was trying to map the TTL value to an Option of u64, which works for the no TTL and Some positive TTL case, but for the no default TTL case of -1, this breaks the deserialization and the code panics. 

This PR refactors the TTL from a Duration into an enum, which seems to be a more Idiomatic way to represent the non-binary state that TTL can be.  Now it can support the TTL off state, the TTL not defined but on state, and the TTL defined and defined state.

PR addresses https://github.com/Azure/azure-sdk-for-rust/issues/3467


sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
```diff
@@ -8,20 +8,77 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
...
+    where
+        S: Serializer,
+    {
+        match self {
+            TimeToLive::Off => serializer.serialize_none(),
+            TimeToLive::NoDefault => serializer.serialize_i64(-1),
+            TimeToLive::Seconds(d) => {
+                let secs = i64::try_from(d.as_secs()).map_err(serde::ser::Error::custom)?;
+                serializer.serialize_i64(secs)
+            }
```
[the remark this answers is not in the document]
11alieberman: The previous serialization/deserialization already disregarded sub-seconds so this behavior is consistent.

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
```diff
@@ -8,20 +8,79 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
...
+
+    /// TTL is enabled, but items have no default expiration.
+    ///
+    /// Individual items can still define their own TTL.
+    NoDefault,
+
+    /// TTL is enabled with a default expiration of the given duration.
+    ///
+    /// Only whole seconds are used; any sub-second component is truncated during serialization.
+    Seconds(Duration),
```
FabianMeiswinkel: I like the direction o fthe change - but my preference would be to use an int here instead of Duration. The whole-second conversion otherwise keeps over-complicating sceanrios where the duration has fractional seconds.

FabianMeiswinkel: Also TTL and analytical storage TTL shoudl probably be added to SystemProperties

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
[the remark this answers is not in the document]
FabianMeiswinkel: Understood - but while cleaning-up the API, the confusion can be fixed once and for all?

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
11alieberman: Makes sense. Let me address those!

11alieberman: I have refactored the duration into just using i64 for the tracking of the seconds.

Question regarding the addition to SystemProperties. Does that make sense given SystemProperties are also used in the DatabaseProperties. But the concept of TTL would not apply to a Database, right?

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
```diff
@@ -1,27 +1,81 @@
...
+    #[default]
+    Off,
+
+    /// TTL is enabled, but items have no default expiration.
+    ///
+    /// Individual items can still define their own TTL.
+    NoDefault,
+
+    /// TTL is enabled with a default expiration of the given number of seconds.
+    Seconds(i64),
```
FabianMeiswinkel:
```suggestion
    Seconds(u32),
```

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
```diff
@@ -1,27 +1,81 @@
...
+impl TimeToLive {
+    /// Returns `true` if TTL is [`Off`](TimeToLive::Off).
+    pub fn is_off(&self) -> bool {
+        matches!(self, TimeToLive::Off)
+    }
+}
+
+impl From<i64> for TimeToLive {
+    fn from(n: i64) -> Self {
+        TimeToLive::Seconds(n)
```
FabianMeiswinkel, line 48: Same validations as in deserialize should apply. My preference would be to use u32 (to clarify that it can only be positive number and 4 billion seconds as TTL seems perfectly sufficient). But I guess u64 would also be an option.

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
```diff
@@ -1,27 +1,81 @@
...
+///   number of seconds unless they override it with their own `ttl` property.
+///
+/// For more information see <https://learn.microsoft.com/azure/cosmos-db/time-to-live#time-to-live-configurations>
+#[derive(Clone, Default, SafeDebug, PartialEq, Eq)]
+#[safe(true)]
+#[non_exhaustive]
+pub enum TimeToLive {
+    /// TTL is disabled; items never expire.
+    #[default]
+    Off,
```
analogrelay: naming nit


```suggestion
    Forever,
```

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
```diff
@@ -1,27 +1,81 @@
...
-fn serialize_ttl<S>(duration: &Option<Duration>, serializer: S) -> Result<S::Ok, S::Error>
-where
-    S: Serializer,
-{
-    match duration {
-        Some(d) => serializer.serialize_some(&d.as_secs()),
-        None => serializer.serialize_none(),
+impl TimeToLive {
+    /// Returns `true` if TTL is [`Off`](TimeToLive::Off).
+    pub fn is_off(&self) -> bool {
```
analogrelay:
```suggestion
    pub fn is_forever(&self) -> bool {
```

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
```diff
@@ -260,52 +313,139 @@ pub enum ConflictResolutionMode {
...
+
+    #[test]
+    fn round_trip_ttl_no_default() {
+        let original = TtlHolder {
+            ttl: TimeToLive::NoDefault,
+        };
+        let json = serde_json::to_string(&original).unwrap();
+        let deserialized: TtlHolder = serde_json::from_str(&json).unwrap();
+        assert_eq!(original.ttl, deserialized.ttl);
+    }
```
analogrelay, line 400: Round-trip tests are false coverage, because they can succeed if symmetric bugs are present in the deserialize and serialize implementations. Let's just have these be serialize tests that assert the resulting JSON :). You already have the deserialize tests anyway :)

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
11alieberman: Yes of course `u32` makes more sense here, sorry. I saw the word int and my mind was just like, use a type that has an i...

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
11alieberman: Tests have been updated!

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
11alieberman: Updated

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
11alieberman: Updated

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
11alieberman: This has been refactored

sdk/cosmos/azure_data_cosmos/src/models/container_properties.rs
```diff
@@ -1,27 +1,81 @@
...
-{
-    match duration {
-        Some(d) => serializer.serialize_some(&d.as_secs()),
-        None => serializer.serialize_none(),
+impl<'de> Deserialize<'de> for TimeToLive {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        match Option::<i32>::deserialize(deserializer)? {
```
11alieberman, line 70: I updated this to deserialize from an `i32` since that's what the ttl property in cosmos is actually modelled as, at least it's what I assume from this page https://learn.microsoft.com/en-us/azure/cosmos-db/time-to-live#examples

I also updated it to fail if the int is 0, because the docs state that it must be nonzero or -1. I also tested this in the cosmos storage explorer trying to set the ttl value to 0 and it failed giving me a similar error. Tests were updated accordingly.

FabianMeiswinkel: Thanks !

PR #3936, opened 2026-03-11
[Storage] Increased queue testing coverage 1 (test consolidation, no options untested, overall cleanup)

vincenttran-msft:
Generating off this: https://github.com/Azure/azure-rest-api-specs/pull/41338
Queues also needs the fixed-width fix from Blobs. 

This PR broadens Azure Storage Queue coverage and improves the surrounding developer experience 😉 

- Expands the queue test suite with more end-to-end scenarios, then cleans up the queue test files so they read more consistently and follow the same “group related states by operation” pattern used in the blob tests.
- Adds new queue examples and documentation, including a true hello-world sample and an access policy sample, and updates the `README/example` index.

- In `sdk/storage/azure_storage_queue/tests/queue_client.rs`, consolidates several tests around single operation families instead of separate variants.
  - `get_properties` now covers missing queue, empty queue, and populated queue checks.
  - `peek_messages` now covers missing queue, empty queue, and populated queue behavior.
  - `receive_messages` now covers missing queue, empty queue, and populated queue behavior.
  - `delete`, `set_metadata`, and access policy tests now include both error and success paths in the same test case.

### DIFF
--- a/sdk/storage/azure_storage_queue/tsp-location.yaml
+++ b/sdk/storage/azure_storage_queue/tsp-location.yaml
@@ -1,4 +1,4 @@
 directory: specification/storage/Microsoft.QueueStorage
-commit: 7ac92a50fb8bae4a08753187813b53b79863b696
+commit: f284c41212d524b109dc4c33eb259f7e05b1980e
 repo: Azure/azure-rest-api-specs
 additionalDirectories: 


sdk/storage/azure_storage_queue/examples/access_policy.rs
```diff
@@ -0,0 +1,101 @@
...
+    queue_client.create(None).await?;
+    println!("Created queue '{queue_name}'");
+
+    // --- Set access policies ---
+    // Define a policy that grants read ('r') and process ('p') access.
+    // start/expiry are optional; omitting them means the policy is active immediately
+    // and never expires. Use a Shared Access Signature (SAS) to attach this policy
+    // to a token with a specific expiry instead.
+    let policy = AccessPolicy {
+        start: None,
```
jalauzon-msft, line 65: Is there still something weird customers need to do with the date here? Maybe fill one of these in as an example (either way)

sdk/storage/azure_storage_queue/examples/queue_client.rs
```diff
@@ -1,281 +1,139 @@
...
     models::{
         QueueClientGetPropertiesResultHeaders, QueueClientPeekMessagesOptions,
-        QueueClientReceiveMessagesOptions, QueueClientUpdateMessageOptions, QueueMessage,
-        SentMessage,
+        QueueClientReceiveMessagesOptions, QueueClientSendMessageOptions, QueueMessage,
     },
     QueueClient,
 };
 
-async fn send_message(
```
jalauzon-msft, line 20: Not going to lie, I think I preferred the previous formatting where specific operations had their own functions all called from a main. Yes, it's a little more code but keeps the examples more targeted and easier to find specific stuff. Also, probably easier for AI to parse/learn from.

sdk/storage/azure_storage_queue/examples/access_policy.rs
vincenttran-msft: Nope just business as usual, but this was a really really good comment. Although this initial PR was just leaving no stones unturned and didn't start nailing the options bag like we are doing in Blobs, this already revelead that the Queues package is missing the fixed-width fix from TypeSpec that we had in Blobs. Fixing now!

sdk/storage/azure_storage_queue/examples/queue_client.rs
vincenttran-msft: Reverted to old style!

PR #3939, opened 2026-03-11
Update cargo package versions to latest...

LarryOsterman:
Zip 7.4 was yanked, so moved to latest Zip.

Eventhubs processor referenced a now deprecated method, so updated to current function.

### DIFF
--- a/sdk/eventhubs/azure_messaging_eventhubs/src/event_processor/processor.rs
+++ b/sdk/eventhubs/azure_messaging_eventhubs/src/event_processor/processor.rs
@@ -425,7 +425,7 @@ impl EventProcessor {
         // Close all partition clients.
         info!("Closing all partition clients.");
         let mut clients = self.next_partition_clients.lock().await;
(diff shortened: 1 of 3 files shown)


sdk/eventhubs/azure_messaging_eventhubs/src/event_processor/processor.rs
heaths, line 1: Is this actually related?

heaths (approved):
Looks okay sans the possibly unrelated file I'm guessing is from other work you were doing today.

sdk/eventhubs/azure_messaging_eventhubs/src/event_processor/processor.rs
LarryOsterman: The update from futures_channel version 3.31 to 3.32 deprecated the `try_next()` function in favor of the `try_recv()` function. Rather than run the risk of a deprecated API, I decided to fix it in this PR.

PR #3942, opened 2026-03-12
Fixes #3926 by simplifying regional routing state and moving Gateway20 handling into a single `CosmosEndpoint` model.

### LINKED ISSUE #3926: Cosmos Driver: Add ThinClient endpoints to LocationState instead of treating it as parallel concept
Also see https://github.com/Azure/azure-sdk-for-rust/pull/3912#discussion_r2920417920 for more details

FabianMeiswinkel:
## Summary

Fixes #3926 by simplifying regional routing state and moving Gateway20 handling into a single `CosmosEndpoint` model.

This change:
- removes `PreferredEndpointGroup`
- represents each logical region as one `CosmosEndpoint` with a standard gateway URL and an optional Gateway20 URL
- keeps account routing state as flat `Vec<CosmosEndpoint>` lists for preferred read and write endpoints
- has routing produce one concrete attempt target via `RoutingDecision.selected_url` and `RoutingDecision.use_gateway20`
- keeps request construction and transport selection driven by the same routing decision
- treats endpoint availability at the region level, so if either the standard gateway path or the Gateway20 path is unavailable, the region is skipped
- keeps transport selection after routing and removes any need for transport-layer endpoint rewriting

## Why

The previous refactor moved Gateway20 selection into routing, but it introduced `PreferredEndpointGroup` as an intermediate state shape. That still made the routing model more complicated than necessary.

This PR simplifies the state flow so that:
- account state is just preferred regional endpoints
- each regional endpoint can optionally expose a Gateway20 URL
- routing decides the concrete target once
- downstream stages consume that decision instead of re-deriving it

That keeps the model closer to the intended architecture: simple regional state, single-source target selection, and no parallel endpoint concept in transport.

## Spec Update

Updated `sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md` to match the implementation:
- flattened `CosmosEndpoint` shape
- `RoutingDecision.selected_url` / `use_gateway20`
- post-routing transport selection
- removal of stale references to grouped endpoints and transport-side endpoint context

### DIFF
--- a/sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md
+++ b/sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md
@@ -141,6 +141,18 @@ focused component types. Each pipeline stage operates on only the components it 
 /// Already exists as `CosmosOperation`.
 // (reuse existing CosmosOperation)
 
+/// Transport mode for a routed attempt.
+///
+/// Determines which HTTP transport (and protocol version) is used
+/// for a given request attempt.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum TransportMode {
+    /// Standard gateway (HTTP/1.1 or HTTP/2 via ALPN).
+    Gateway,
+    /// Gateway 2.0 (HTTP/2 with prior knowledge).
+    Gateway20,
+}
+
 /// COMPONENT: Routing decision for the current attempt.
 /// Produced by endpoint resolution, consumed by transport.
 /// Replaced (not mutated) on retry/failover.


sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/account_endpoint_state.rs
```diff
@@ -29,8 +29,8 @@ impl AccountEndpointState {
     pub fn single(default_endpoint: CosmosEndpoint) -> Self {
         Self {
             generation: 0,
-            preferred_read_endpoints: vec![default_endpoint.clone()],
-            preferred_write_endpoints: vec![default_endpoint.clone()],
+            preferred_read_endpoints: Vec::new(),
+            preferred_write_endpoints: Vec::new(),
```
FabianMeiswinkel: 🔴 **Blocking · Correctness: Invariant Violation — Empty Endpoint Lists**

`AccountEndpointState::single()` now creates empty `preferred_read_endpoints` and `preferred_write_endpoints` (changed from `vec![default_endpoint.clone()]`). Additionally, `build_account_endpoint_state` no longer falls back to `default_endpoint` when the lists are empty.

`LocationIndex::next()` and `next_for_generation()` both assert:
```rust
debug_assert!(list_len > 0, "endpoint list should never be empty");
```

If a retry occurs while endpoint lists are empty (e.g., before metadata refresh, or when all location URLs are malformed), `advance_location(0, ...)` panics in debug builds. In release, it silently returns `self` — no crash, but the index becomes stale.

Tests and CI run in debug mode, so this assertion can fire when the initial metadata fetch is delayed or when `AccountProperties` has empty/malformed location lists. The old code's `if preferred_read_endpoints.is_empty() { push(default_endpoint) }` guaranteed non-empty lists.

**Suggestion**: Restore the default-endpoint fallback in `build_preferred_endpoints` (or `build_account_endpoint_state`) when the result is empty, and/or keep `single()` with `vec![default]` as before.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
```diff
@@ -55,6 +49,86 @@ pub(crate) fn build_account_endpoint_state(
...
+        parse_thin_client_locations(thin_client_locations)
+    } else {
+        HashMap::new()
+    };
+
+    let mut endpoints = Vec::with_capacity(standard_locations.len());
+    for region in standard_locations {
+        let Ok(url) = url::Url::parse(&region.database_account_endpoint) else {
+            continue;
+        };
```
FabianMeiswinkel, line 86: 🟡 **Recommendation · Correctness: Missing Warning for Standard Location URL Validation**

`build_preferred_endpoints` silently skips regions with unparseable standard gateway URLs (`let Ok(url) = url::Url::parse(...) else { continue; }`). If ALL standard locations have malformed URLs, the endpoint list ends up empty — compounding Finding #1 above.

The thin-client path below (`parse_thin_client_locations`) has explicit `warn!()` logging for malformed URLs, but this standard-location path is silent.

**Suggestion**: Add a `warn!()` when a standard location URL fails to parse, matching the logging pattern in `parse_thin_client_locations`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
```diff
@@ -21,29 +23,21 @@ pub(crate) fn build_account_endpoint_state(
...
-    if preferred_write_endpoints.is_empty() {
-        preferred_write_endpoints.push(default_endpoint.clone());
-    }
+    let preferred_write_endpoints = build_preferred_endpoints(
+        &properties.writable_locations,
+        &properties.thin_client_writable_locations,
+        gateway20_enabled,
+    );
 
     AccountEndpointState {
```
FabianMeiswinkel, line 52: 🟡 **Recommendation · Reliability: Thin-Client URL Pairing No Longer Merges Read+Write**

The old `thin_client_endpoint_overrides` explicitly merged read and write thin-client locations into a single map, with a comment explaining: *"Merges both read and write thin-client locations into a single map. This is necessary because session retries on single-write accounts can reroute reads to write endpoints — if only read locations were included, the write-region URL override would be missing."*

The new code pairs `readable_locations` → `thin_client_readable_locations` and `writable_locations` → `thin_client_writable_locations` separately. This is correct as long as the service always advertises `thinClientWritableLocations` for write regions. But if a region appears in `writable_locations` but NOT in `thin_client_writable_locations` (yet IS in `thin_client_readable_locations`), session retry reads routed to write endpoints would lose Gateway 2.0.

This is a behavior change from the previous PR (#3912). The new design is arguably cleaner, but it drops an explicit cross-contamination safety net.

**Suggestion**: Add a comment here explaining the intentional per-side pairing decision, and confirming that session retries using write endpoints will use write-side thin-client URLs (or fall back to standard gateway).

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/endpoint.rs
```diff
@@ -38,21 +47,35 @@ impl CosmosEndpoint {
...
+        prefer_gateway20 && self.gateway20_url.is_some()
+    }
+
+    /// Returns the concrete URL selected for this attempt.
+    pub(crate) fn selected_url(&self, prefer_gateway20: bool) -> &Url {
+        if self.uses_gateway20(prefer_gateway20) {
+            return self
+                .gateway20_url
+                .as_ref()
+                .expect("Gateway20 URL presence checked before selection");
```
FabianMeiswinkel: 🟡 **Recommendation · Robustness: Fragile `.expect()` Pattern**

`selected_url()` calls `uses_gateway20()` then `.expect()` on `gateway20_url`. The expect is currently unreachable because `uses_gateway20()` checks `self.gateway20_url.is_some()`. But this creates a hidden coupling — if someone changes `uses_gateway20()` later (e.g., adding more conditions), the expect could fire at runtime.

**Suggestion**: Replace with a direct pattern match that doesn't require the expect:
```rust
pub(crate) fn selected_url(&self, prefer_gateway20: bool) -> &Url {
    if prefer_gateway20 {
        if let Some(url) = &self.gateway20_url {
            return url;
        }
    }
    &self.gateway_url
}
```

sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md
```diff
@@ -1376,6 +1391,7 @@ The transport pipeline does **NOT** handle:
 Following the same DOP pattern as the operation pipeline, the transport pipeline is a bare
 async function — not a struct with methods. Its dependencies (`AdaptiveTransport`,
 `AuthContext`) are passed as parameters. Header application and request signing are also bare
+`Credential`) are passed as parameters. Header application and request signing are also bare
```
FabianMeiswinkel: 🟡 **Documentation: Duplicated Paragraph with Stale `AuthContext` Reference**

Lines 1393-1394 (in the new file) have a duplicated paragraph:
```
`AuthContext`) are passed as parameters. Header application and request signing are also bare
`Credential`) are passed as parameters. Header application and request signing are also bare
```

The old `AuthContext` line (1393) should have been removed when the `Credential` line (1394) was added. Only the `Credential` version should remain.

sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md
```diff
@@ -657,7 +678,7 @@ async fn execute_hedged(
 
     // Start primary attempt
     let primary_fut = execute_transport_pipeline(
-        primary_req, transport, auth_context, diagnostics,
+        primary_req, transport, credential, diagnostics,
```
FabianMeiswinkel, line 694: 🟡 **Documentation: §4.2 Hedging Uses `credential` Not in Parameter List + §4.1 Still Passes `auth_context`**

Two related spec issues:

1. **§4.2**: The `execute_hedged` function signature had `auth_context: &AuthContext` removed (see line 667 in this diff), but the body here calls `execute_transport_pipeline(primary_req, transport, credential, diagnostics)` — `credential` is not in the parameter list. Either add `credential: &Credential` to the parameters or update the body to match the new design.

2. **§4.1**: The `OperationAction::Hedge` match arm (around line 638-640 in the new file, outside this diff hunk) still passes `auth_context` to `execute_hedged`, which no longer takes that parameter. Should be updated to `credential` or removed.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -823,8 +812,95 @@ mod tests {
...
             &advanced_state,
             &location,
+            false,
             Duration::from_secs(60),
         );
         assert_eq!(second_routing.endpoint, endpoint_b);
     }
+
+    #[test]
+    fn resolve_endpoint_prefers_gateway20_for_dataplane_reads() {
```
FabianMeiswinkel, line 833: 🟡 **Testing Gap: No Test for Empty Preferred Endpoint Lists**

There's no test that exercises the scenario where all standard location URLs are malformed (or `readable_locations`/`writable_locations` is empty in `AccountProperties`), resulting in empty `preferred_*_endpoints`. This would help verify:
1. `resolve_endpoint` correctly falls back to `default_endpoint` 
2. Retries handle the empty-list case without panicking on the `debug_assert!(list_len > 0, ...)` in `LocationIndex::next()`

This gap is directly related to the correctness concern about empty endpoint lists described in the `account_endpoint_state.rs` comment.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
```diff
@@ -122,16 +196,43 @@ mod tests {
...
+            "systemReplicationPolicy": { "minReplicaSetSize": 3, "maxReplicasetSize": 4 },
+            "readPolicy": { "primaryReadCoefficient": 1, "secondaryReadCoefficient": 1 },
+            "queryEngineConfiguration": "{}"
+        }))
+        .unwrap();
+
+        let state = build_account_endpoint_state(&properties, default_endpoint(), None, true);
+
+        assert!(state.preferred_read_endpoints[0].gateway20_url().is_some());
+        assert!(state.preferred_write_endpoints[0].gateway20_url().is_none());
```
FabianMeiswinkel, line 251: 🟡 **Testing Gap: Missing Test for Write-Side Gateway 2.0 URLs**

This test verifies that read endpoints get Gateway 2.0 URLs and write endpoints don't (because `thinClientWritableLocations` is absent in the test fixture). 

There's no test where `thinClientWritableLocations` IS present, verifying that write endpoints also pick up their Gateway 2.0 URLs. This is important given the behavior change in thin-client pairing (read and write thin-client locations are now paired separately rather than merged).

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -117,16 +118,22 @@ pub(crate) async fn execute_operation_pipeline(
...
             &routing,
-            transport_context.thin_client_overrides.as_deref(),
             activity_id,
             execution_context,
             deadline,
         )?;
 
+        let selected_transport = match pipeline_type {
+            PipelineType::DataPlane => {
+                transport.get_dataplane_transport(account_endpoint, routing.use_gateway20)?
```
FabianMeiswinkel: 🟡 **Design: URL Selection and Transport Selection Have Unchecked Consistency Assumption**

`routing.selected_url` and `routing.use_gateway20` are set by routing. Then `transport.get_dataplane_transport(account_endpoint, routing.use_gateway20)` selects the HTTP client. These are determined independently — no compile-time guarantee that the URL and transport match.

If `use_gateway20` is `true` but `is_gateway20_allowed()` returns `false` inside `get_dataplane_transport`, the request would use a Gateway 2.0 URL with the standard gateway HTTP/1.1 transport. In practice both flags come from the same `ConnectionPoolOptions`, so they can't diverge. But the design now has this decoupled assumption that the old `TransportContext` bundled together.

**Suggestion**: Consider adding a `debug_assert` in the pipeline verifying that when `routing.use_gateway20` is `true`, the selected transport is actually the Gateway 2.0 transport.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
```diff
@@ -265,42 +208,25 @@ impl CosmosTransport {
...
-                transport,
-                thin_client_overrides: Some(overrides),
-            })
-        } else {
-            Ok(TransportContext {
-                transport: self.dataplane_gateway_transport.clone(),
-                thin_client_overrides: None,
-            })
+        match use_gateway20 {
+            true if self.connection_pool.is_gateway20_allowed() => {
```
FabianMeiswinkel: 🟡 **Design: Redundant `is_gateway20_allowed()` Guard — Consider Adding Explanatory Comment**

If we reach `use_gateway20: true`, it means `gateway20_enabled` was `true` when building account state (same source: `connection_pool.is_gateway20_allowed()`). So the `true if self.connection_pool.is_gateway20_allowed()` branch condition is effectively always `true` when `use_gateway20` is `true`. The check is defense-in-depth.

This is reasonable as a safety net, but a short comment explaining it's intentional defense-in-depth would help future readers understand it's not dead code.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/endpoint.rs
```diff
@@ -38,21 +47,35 @@ impl CosmosEndpoint {
...
+        }
+
+        &self.gateway_url
+    }
+
```
FabianMeiswinkel: 🟢 **Nit: Extra blank line before closing `}`**

There's an extra blank line between `selected_url` and the closing `}` of `impl CosmosEndpoint`. `rustfmt` typically wouldn't add this.

sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md
```diff
@@ -145,8 +145,15 @@ focused component types. Each pipeline stage operates on only the components it
...
-    /// The resolved Cosmos DB service endpoint.
+    /// The resolved Cosmos DB service endpoint for the chosen region.
     pub endpoint: CosmosEndpoint,
+    /// The concrete URL selected for this attempt.
+    ///
+    /// For dataplane operations this may be the Gateway 2.0 URL when the
+    /// endpoint exposes one and the runtime configuration allows it.
+    pub selected_url: Url,
+    /// Whether the selected URL requires Gateway 2.0 transport.
+    pub use_gateway20: bool,
```
analogrelay: nit: I'd use a `EndpointKind` enum or something, with `Gateway`, `Gateway20`, and potential future values here, just for clarity. But as this is internal, this is non-blocking for sure.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/account_endpoint_state.rs
analogrelay: This is probably fine given we want to block client creation on constructing a complete AccountEndpointState (including all regional endpoints).

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/endpoint.rs
analogrelay: This is valid. I would prefer something like this pattern instead of using `.expect`, even if we know it should be fine because `uses_gateway20` already checked for `is_some`. Alas, the double-nested ifs are necessary since in Rust 2021 you can't pair `if let` with another boolean condition (this is fixed in edition 2024, but we're pinned at 2021 for now).

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
analogrelay: > session retry reads routed to write endpoints would lose Gateway 2.0

That's probably OK though, right? If the server doesn't advertise that a write endpoint is available with Gateway 2.0, we should just fall back to Gateway 1.0. It's up to the server to tell us which endpoints exist.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
```diff
@@ -55,6 +49,86 @@ pub(crate) fn build_account_endpoint_state(
...
+                if entry.get() != &url {
+                    warn!(
+                        region = %region.name,
+                        existing_url = %entry.get(),
+                        new_url = %url,
+                        "Duplicate thin-client region with conflicting URL; keeping first entry"
+                    );
+                }
+            }
+        }
```
analogrelay: Could probably do this with `urls.entry(region.name.clone()).and_modify(... check for duplicate ...).or_insert(url)` to avoid the noisy match. Not blocking though. Just because it's called `and_modify` doesn't mean it _has_ to modify the value :). It just lets you run a closure over an occupied entry's value IF there is one.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
```diff
@@ -265,42 +208,25 @@ impl CosmosTransport {
...
-            Ok(TransportContext {
-                transport,
-                thin_client_overrides: Some(overrides),
-            })
-        } else {
-            Ok(TransportContext {
-                transport: self.dataplane_gateway_transport.clone(),
-                thin_client_overrides: None,
-            })
+        match use_gateway20 {
```
analogrelay: This is just a verbose way to write `if use_gateway20 && self.connection_pool.is_gateway20_allowed()`, isn't it?

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
analogrelay: Yeah, that's a fair point. `is_gateway20_allowed` should only be used in routing anyway. Once the endpoint has been selected, the transport pipeline should just use it with whatever protocol is specified.

analogrelay (changes requested):
Mostly minor stuff

sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md
FabianMeiswinkel: ACK - will fix.

sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md
FabianMeiswinkel: ACK - will fix

sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md
FabianMeiswinkel: ACK - will fix

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: ACK - will fix

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Won't fix - this is overkill - if the service returns just broken Uris we have bigger issues

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/account_endpoint_state.rs
FabianMeiswinkel: True - but I think it still makes sense to avoid it. So, my preference is to keep the 

'''
if preferred_read_endpoints.is_empty() { push(default_endpoint) }
'''

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/endpoint.rs
FabianMeiswinkel: ACK - will fix

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/endpoint.rs
FabianMeiswinkel: ACK will fix

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
FabianMeiswinkel: Won't fix - this is just fine - if the write region only has Gateway (no GW2.0) then we can always use just Gateway - this is not a functional problem.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
FabianMeiswinkel: ACK - will fix

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
FabianMeiswinkel: ACK - will fix.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
FabianMeiswinkel: ACK - will fix

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
FabianMeiswinkel: ACK - will fix

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
FabianMeiswinkel: Won't fix - the less noisy match approach makes it readable enough.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/account_endpoint_state.rs
FabianMeiswinkel: Fixed in d86cc94ef — restored the `default_endpoint` fallback in both `single()` (populates both lists with `vec![default_endpoint.clone()]`) and `build_account_endpoint_state()` (pushes `default_endpoint` when `build_preferred_endpoints()` returns empty). This prevents the `debug_assert!(list_len > 0)` panic in `LocationIndex::next()`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
FabianMeiswinkel: Fixed in dc579dc05 — added `warn!()` logging when a standard location URL fails to parse, matching the pattern in `parse_thin_client_locations()`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/endpoint.rs
FabianMeiswinkel: Fixed in 38c88baeb — replaced `.expect()` with nested `if`/`if let` as suggested. Also removed the trailing blank line. Rust 2021 doesn't support `if let` paired with `&&`, so nested ifs are fine.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/endpoint.rs
FabianMeiswinkel: Fixed in 38c88baeb — removed the extra blank line as part of the `selected_url()` refactor.

sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md
FabianMeiswinkel: Fixed in f481984c3 — removed the duplicated `AuthContext` line, keeping only the `Credential` version.

sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md
FabianMeiswinkel: Fixed in f481984c3 — replaced `auth_context` with `credential` in the §4.1 call site and added `credential: &Credential` to the `execute_hedged` signature in §4.2.

sdk/cosmos/azure_data_cosmos_driver/docs/TRANSPORT_PIPELINE_SPEC.md
FabianMeiswinkel: Fixed in 8c0fbc544 — introduced `TransportMode` enum (`Gateway`, `Gateway20`) in `components.rs`, replacing `use_gateway20: bool` in `RoutingDecision`. Updated `resolve_endpoint`, `get_dataplane_transport`, all tests, and the spec doc.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
FabianMeiswinkel: Fixed in c69e8825f — added `build_state_adds_gateway20_for_write_endpoints_when_present` test that includes `thinClientWritableLocations` and asserts both read and write endpoints pick up their Gateway 2.0 URLs.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Won't fix — the `TransportMode` enum now makes the consistency assumption structurally clear. Since both flags derive from the same `ConnectionPoolOptions`, they can't diverge in practice. The defense-in-depth guard in `get_dataplane_transport` already handles the theoretical mismatch gracefully (falls back to Gateway).

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
FabianMeiswinkel: Fixed in 1c6d6d541 — refactored to `and_modify`/`or_insert` pattern as suggested.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/mod.rs
FabianMeiswinkel: Addressed in 8c0fbc544 + cee09127c — the `match` now reads naturally with `TransportMode::Gateway20 if ...`, making @analogrelay's concern about verbosity moot. Also added a defense-in-depth explanatory comment per the self-review suggestion.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/emulator.rs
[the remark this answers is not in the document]
FabianMeiswinkel: Won't fix — this is a false positive. The existing tests (`ipv6_loopback_short_is_emulator` and `ipv6_loopback_full_is_emulator`) pass, confirming `host_str()` returns bracketed IPv6 in this context. Also, this file is not changed in this PR.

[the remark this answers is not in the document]
FabianMeiswinkel: Won't fix — duplicate of the comment above; same false positive. File not changed in this PR.

sdk/cosmos/azure_data_cosmos_perf/run_perf.sh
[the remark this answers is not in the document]
FabianMeiswinkel: Won't fix — valid concern for a future PR, but this file is not changed in this PR (likely residual from when PR was based against main with a different diff).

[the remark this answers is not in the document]
FabianMeiswinkel: Won't fix — duplicate thread, same as above. File not changed in this PR.

sdk/cosmos/azure_data_cosmos/src/retry_policies/client_retry_policy.rs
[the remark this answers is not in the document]
FabianMeiswinkel: Won't fix — `ErrorKind::Connection` maps to `RequestSentStatus::NotSent` in `tracked_transport.rs`, so it enters the `NotSent` branch (line 686) and never reaches this `Sent/Unknown` path. The `Io` check here specifically handles the case where request was sent but response had an I/O failure. Also, this file is not changed in this PR.

sdk/cosmos/azure_data_cosmos/src/fault_injection/http_client.rs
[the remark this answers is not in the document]
FabianMeiswinkel: Won't fix — file not changed in this PR. The `ErrorKind::Io` vs `ErrorKind::Connection` semantics are correct as-is (see analysis on the retry_policy thread above).

Issue #3934, opened 2026-03-11
Cosmos Driver: Background Task Manager


analogrelay:
The Driver needs a Background Task Manager to handle things like PPAF failback, heath checks, etc.

This issue tracks building such a manager behind a `tokio` feature flag which will be **enabled by default**. In the future, when we need to support other async runtimes, we can add alternative implementations with other feature flags. Tokio should be an optional dependency of the driver, and the `tokio` feature (and by implication, the default feature) should enable that dependency.

PR #3945, opened 2026-03-12
Cosmos: Introduce `BackgroundTaskManager` To Spawn Background Tasks Using Tokio Runtime

kundadebdatta:
## Description

Introduces a `BackgroundTaskManager` in the `azure_data_cosmos_driver` crate to manage the lifecycle of background tasks spawned on the tokio runtime. This is a foundational building block for transport-layer operations that require background work (e.g., connection warmup, health monitoring, periodic refresh).

### Why not `azure_core::async_runtime`?

The manager uses `tokio::spawn` directly instead of the `azure_core` async runtime abstraction because it requires `JoinHandle::abort()` for correct task cancellation on drop. The `AsyncRuntime::spawn` abstraction returns a `SpawnedTask` (a boxed future) which only detaches on drop — it does not cancel the underlying task.

> **Note:** The `tokio` default feature is temporary scaffolding. It should be replaced by the `tokio` feature flag in `azure_core` once available.

### Key design decisions

- **Abort-on-drop semantics** — Unlike raw tokio `JoinHandle`s (which detach on drop), the manager explicitly aborts all stored tasks when dropped, preventing orphaned background work.
- **Panic safety** — Spawned futures are wrapped in `catch_unwind` to log panics without propagating them to the runtime.
- **Handle pruning** — Completed `JoinHandle`s are pruned on each `spawn()` call to prevent unbounded accumulation in long-lived managers.
- **Graceful shutdown** — An async `shutdown()` method aborts and awaits all tasks for deterministic cleanup, complementing the fire-and-forget `Drop` impl.
- **Thread safety** — Uses `std::sync::Mutex<Vec<JoinHandle<()>>>` for interior mutability, allowing `spawn(&self)` to work when the manager is held behind an `Arc`.

### `azure_data_cosmos_driver/Cargo.toml`
- Added `tokio` as an optional workspace dependency with the `rt` feature.
- Added a `tokio` feature flag (`dep:tokio`), included in the `default` feature set.

### `azure_data_cosmos_driver/src/driver/transport/background_task_manager.rs` *(new file)*
- `BackgroundTaskManager` struct with `Debug`, `Default`, and `Drop` implementations.
- `new()` — creates an empty manager.
- `spawn()` — spawns a future on the tokio runtime with panic-catching, stores the handle, and prunes finished handles.
- `shutdown()` — aborts all tasks and awaits their termination for graceful cleanup.
- 8 unit tests covering: construction, drop cleanup, task completion, abort-on-drop, shutdown semantics, handle pruning, and concurrent spawn safety (multi-threaded).

### `azure_data_cosmos_driver/src/driver/transport/mod.rs`
- Registered the `background_task_manager` module behind `#[cfg(feature = "tokio")]`.

## Issue

Fixes Azure/azure-sdk-for-rust#3934

### DIFF
--- a/sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/background_task_manager.rs
+++ b/sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/background_task_manager.rs
@@ -40,19 +40,13 @@ impl std::fmt::Debug for BackgroundTaskManager {
     }
 }
 
-impl Default for BackgroundTaskManager {
-    fn default() -> Self {
-        Self {
-            tasks: Mutex::new(Vec::new()),
-        }
-    }
-}
-
 #[allow(dead_code)]
 impl BackgroundTaskManager {
     /// Creates a new [`BackgroundTaskManager`] with no active tasks.
     pub fn new() -> Self {
-        Self::default()
+        Self {
+            tasks: Mutex::new(Vec::new()),
+        }
     }
 
     /// Spawns a background task on the tokio runtime and stores the handle.
@@ -84,7 +78,7 @@ impl BackgroundTaskManager {
         let mut tasks = self
             .tasks
             .lock()
-            .expect("BackgroundTaskManager mutex poisoned");
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
         tasks.retain(|h| !h.is_finished());
         tasks.push(handle);
     }
@@ -98,7 +92,7 @@ impl BackgroundTaskManager {
         let tasks: Vec<_> = self
             .tasks
             .lock()
-            .expect("BackgroundTaskManager mutex poisoned")
+            .unwrap_or_else(|poisoned| poisoned.into_inner())
             .drain(..)
             .collect();
         let count = tasks.len();
@@ -141,9 +135,10 @@ mod tests {
     }
 
     #[test]
-    fn default_creates_empty_manager() {
-        let manager = BackgroundTaskManager::default();
-        assert_eq!(manager.tasks.lock().unwrap().len(), 0);
+    fn debug_shows_task_count() {
+        let manager = BackgroundTaskManager::new();
+        let debug_str = format!("{:?}", manager);
+        assert!(debug_str.contains("tasks_count"));
     }
 
     #[tokio::test]


sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/background_task_manager.rs
```diff
@@ -0,0 +1,315 @@
...
+/// When the manager is dropped, all handles are aborted, cancelling the
+/// associated tasks (tokio `JoinHandle`s detach on drop rather than cancel,
+/// so explicit abort is required).
+#[allow(dead_code)]
+pub(crate) struct BackgroundTaskManager {
+    /// Stored task handles. Aborting these cancels the tasks.
+    /// Uses a [`Mutex`] for interior mutability so that [`spawn`](Self::spawn)
+    /// can accept `&self`, which is required when the manager lives inside an
+    /// `Arc`.
+    tasks: Mutex<Vec<tokio::task::JoinHandle<()>>>,
```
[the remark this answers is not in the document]
kundadebdatta: This is a foundational infrastructure PR. The next PR will wire `BackgroundTaskManager` into the transport layer for running the PPCB background task to failback partitions. It's standard practice to introduce building blocks in one PR and integrate them in the next for clean PR boundaries and reviewability. The `#[allow(dead_code)]` annotations are explicitly temporary until the integration PR lands. Requiring every new type to have a call site in the same PR would conflate infrastructure with integration, making both harder to review.

sdk/cosmos/azure_data_cosmos_driver/Cargo.toml
```diff
@@ -48,7 +49,8 @@ workspace = true
 
 
 [features]
-default = ["reqwest_native_tls"]
+default = ["tokio", "reqwest_native_tls"]
```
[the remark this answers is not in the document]
kundadebdatta: The BackgroundTaskManager module is gated behind #[cfg(feature = "tokio")], so removing it from defaults would make the module not compile by default. Since the driver's transport layer needs this for connection warmup / health monitoring, having it on by default is pragmatic. Resolving this sunce we are already aware and has a migration plan documented.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/background_task_manager.rs
```diff
@@ -0,0 +1,315 @@
...
+    /// accumulation when the manager is long-lived.
+    pub fn spawn<F>(&self, future: F)
+    where
+        F: Future<Output = ()> + Send + 'static,
+    {
+        // NOTE: We use tokio::spawn directly instead of azure_core::async_runtime
+        // because we need JoinHandle::abort() for correct task cancellation on drop.
+        // The AsyncRuntime::spawn abstraction returns SpawnedTask (a boxed future)
+        // which only detaches on drop — it does not cancel the underlying task.
+        let handle = tokio::spawn(async move {
```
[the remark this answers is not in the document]
kundadebdatta: The concern about runtime-agnostic design is valid. However, there's an important nuance Copilot misses:

`JoinHandle::abort() + .await` gives provides deterministic shutdown — you can abort and then await full termination. This is used in shutdown(). With `AbortHandle`/`Abortable`, you'd still need to await the `SpawnedTask`, but `SpawnedTask` is a boxed future without `is_finished()`.

`JoinHandle::is_finished()` enables handle pruning in `spawn()` (line 89). `SpawnedTask` (a boxed future) has no equivalent — you'd lose pruning or need a more complex polling approach.

The entire module is already behind #[cfg(feature = "tokio")], so tokio-only consumers is already the scoping mechanism.

Bottom line: The AbortHandle/Abortable approach is technically feasible but strictly worse — more code, less capability (no pruning, no deterministic shutdown-await). The driver version's tokio-specific approach is the cleaner design given the requirements.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/background_task_manager.rs
```diff
@@ -0,0 +1,315 @@
...
+        let mut s = f.debug_struct("BackgroundTaskManager");
+        match self.tasks.lock() {
+            Ok(tasks) => s.field("tasks_count", &tasks.len()),
+            Err(_) => s.field("tasks_count", &"<poisoned>"),
+        };
+        s.finish()
+    }
+}
+
+impl Default for BackgroundTaskManager {
```
analogrelay: I don't think we need to implement Default here. The new function is enough.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/background_task_manager.rs
```diff
@@ -0,0 +1,315 @@
...
+                    .copied()
+                    .or_else(|| panic_payload.downcast_ref::<String>().map(|s| s.as_str()))
+                    .unwrap_or("<non-string panic>");
+                error!("Background task panicked: {msg}");
+            }
+        });
+        let mut tasks = self
+            .tasks
+            .lock()
+            .expect("BackgroundTaskManager mutex poisoned");
```
analogrelay: I think we can ignore poisoning errors here (see Simon's earlier PR)

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/background_task_manager.rs
kundadebdatta: Yeah, refactored the code to remove the `Default` implementation.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/background_task_manager.rs
kundadebdatta: Updated.

PR #3950, opened 2026-03-13
[Storage] Managed download perf: runtime task per-chunk & separate download work queues

jaschrep-msft:
Major performance improvements for managed download.
- Uses Core's runtime abstraction to spawn workers for each individual chunk of the download.
  - If download is a one-shot, does not spawn any workers. Runs async on whatever task the caller spawned the download on.
- Separates the work of downloading chunks from resequencing chunks to return in the overall download stream.
  - Active chunk downloads capped by existing `parallel` bound.
  - Buffers waiting to be re-sequenced capped at 2x `parallel` bound.
- Completed chunks stored in ring buffer waiting to be returned in overall stream.
- Download tasks tagged with index for resequencing, allowing them to be placed in the correct position in the ring buffer.

Credit to @nateprewitt's initial implementation I ported over to use the tools available in our dependency chain.

### DIFF
--- a/sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
+++ b/sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
@@ -21,7 +21,7 @@ use azure_core::{
 };
 use bytes::{Bytes, BytesMut};
 use futures::{
-    channel::mpsc::{self, UnboundedSender},
+    channel::mpsc::{self, UnboundedReceiver, UnboundedSender},
     future::Either,
     SinkExt, TryStream,
 };
@@ -146,37 +146,51 @@ where
             // this will not change until either:
             //   1. a task sends a message through this channel
             //   2. a task fails
-            // Race awaiting a message vs checking if tasks have completed successfully,
-            // until either message is received or a task failure is found
-            let mut message_fut = rx.recv();
             let channel_message;
-            loop {
-                match future::select(message_fut, future::select_all(task_bucket)).await {
-                    Either::Left((message, task_select)) => {
-                        channel_message = message;
-                        task_bucket = task_select.into_inner();
-                        break;
-                    }
-                    Either::Right(((completed_task, _, remaining_tasks), m_fut)) => {
-                        completed_task.map_err(|e| Error::with_message(ErrorKind::Other, e.to_string()))?;
-                        task_bucket = remaining_tasks;
-                        message_fut = m_fut;
-                    }
-                }
-            }
-            match channel_message {
-                Ok(Ok((idx, bytes))) => drain.push(idx, bytes)?,
-                Ok(Err(err)) => Err(err)?,
-                // Unexpected channel close
-                Err(_) => {
-                    let received = drain.position();
-                    Err(Error::with_message(ErrorKind::Other, format!("Download incomplete: received {received} of {total_chunks} chunks.")))?;
-                }
-            };
+            (channel_message, task_bucket) = await_message_while_joining_workers(&mut rx, task_bucket).await?;
+            let (idx, bytes) = channel_message?;
+            drain.push(idx, bytes)?;
         }
     };
 
     Ok(Box::pin(stream))
+}
+
+/// Race awaiting a message vs checking if tasks have completed successfully,
+/// until either message is received or a task failure is found.
+///
+/// # Returns
+///
+/// - Returns Ok with received message and remaining un-joined tasks.
+/// - Returns Err if channel closed before a message received.
+/// - Returns Err if joined task closed with an error.
+async fn await_message_while_joining_workers<T>(
+    receiver: &mut UnboundedReceiver<T>,
+    mut task_bucket: Vec<SpawnedTask>,
+) -> AzureResult<(T, Vec<SpawnedTask>)> {
+    let on_recv_err = |_| {
+        Error::with_message(
+            ErrorKind::Other,
+            "Download incomplete. Premature channel close.",
+        )
+    };
+
+    let mut message_fut = receiver.recv();
+    // `task_bucket`` may be empty. `select_all` cannot handle that.
+    while !task_bucket.is_empty() {
+        match future::select(message_fut, future::select_all(task_bucket)).await {
+            Either::Left((message, task_select)) => {
+                return Ok((message.map_err(on_recv_err)?, task_select.into_inner()));
+            }
+            Either::Right(((completed_task, _, remaining_tasks), m_fut)) => {
+                completed_task.map_err(|e| Error::with_message(ErrorKind::Other, e.to_string()))?;
+                task_bucket = remaining_tasks;
+                message_fut = m_fut;
+            }
+        }
+    }
+
+    Ok((message_fut.await.map_err(on_recv_err)?, task_bucket))
 }
 
 async fn download_range_to_bytes(


sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -50,63 +58,121 @@ where
...
 
-    let initial_response = match client
-        .transfer_range(Some(
-            max_download_range.start
-                ..min(
-                    max_download_range.end,
-                    max_download_range.start.saturating_add(partition_size),
-                ),
-        ))
+    let initial_response = download_handle_invalid_range(
```
LarryOsterman: This is a fairly complicated bit of code and there aren't many comments explaining it. 

That can end up being a long-term maintenance issue.

Comments don't need to be extensive, but for instance, explaining what the `download_handle_invalid_range` function does here would help to understand the algorithm.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -115,8 +181,107 @@ async fn download_range_to_bytes(
     client: Arc<impl PartitionedDownloadBehavior>,
     range: Range<usize>,
 ) -> AzureResult<Bytes> {
+    let dst = BytesMut::with_capacity(range.len());
     let response = client.transfer_range(Some(range)).await?;
-    response.into_body().collect().await
+    collect_into(response.into_body(), dst).await
```
LarryOsterman: When the core collect_into PR completes, this can be replaced with `response.into_body().collect_into(dst).await`, I believe.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -307,7 +472,7 @@ mod tests {
             };
         let offset = (rand::random::<u64>() as usize % 500) + 500;
 
-        for parallel in [1, 4] {
+        for parallel in [4, 1] {
```
LarryOsterman: Is the goal of this to test the highly parallel version first? If so, would it make sense for these to be two different tests, one which is highly parallelized and the other which is not?

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
LarryOsterman, line 488: Does it make sense to make these recorded tests? 

If they are live-only, would `#[recorded::test(live)]` make more sense than a raw `tokio` test?

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
heaths: When I took a glance at this initially, this was my thought as well.

jaschrep-msft: I may have just forgotten to flip these back while trying to figure something out in a test failure.

Realistically, we need test parameterization macros that are compatible with our recorded test markers. I struggle to justify making additional tests just to copy over 99% of code, and helper functions aren't enough. I dislike the degraded output of looping multiple tests as well as the fail-fast nature where tests stop after the first failure.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
jaschrep-msft: The function itself is documented. I can find a better name but it seems silly to just copy over the existing comments.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
jaschrep-msft: there's no network here. This is a unit test of the download using a mock of the trait used by this download.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
LarryOsterman: I understood what the code was doing, in this case I was asking "why are you starting this operation by initiating something which is supposed to fail". That seems utterly counterintuitive and strange.

You've explained why in person last week, so I know the answer, but the next person reading the code would not.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
heaths: I don't disagree, but keep in mind this entire test framework for recorded tests is entirely custom. Rust doesn't have a good test infrastructure compared to other Azure SDK languages. It's currently not a priority given staffing, but we're always open to pull requests.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
LarryOsterman: That PR is in, I believe this can be replaced with:

`response.into_body().collect_into(dst).await?`

One minor complication is that the `collect_into` call can fail if the provided buffer isn't sufficient to hold received chunks. It will fill up to the buffer, and return the actual amount of data received (if the stream ends before the buffer is filled).

If it cannot be replaced, let me know how I can fix the `collect_into` function to better meet your needs.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -50,63 +67,114 @@ where
...
+            // max tasks are spawned and sequential ready bytes already returned
+            // this will not change until either:
+            //   1. a task sends a message through this channel
+            //   2. a task fails
+            // Race awaiting a message vs checking if tasks have completed successfully,
+            // until either message is received or a task failure is found
+            let mut message_fut = rx.recv();
+            let channel_message;
+            loop {
+                match future::select(message_fut, future::select_all(task_bucket)).await {
```
jalauzon-msft: We ran this thing by Copilot, and it called out that `future::select_all` will panic if the iterator is empty meaning we could have a possible race condition here. To be honest, I'm not familiar enough with all this to know but worth having a look.

Further questioning Copilot about it, it claims it's probably not an issue if using `tokio` but could be with other runtimes.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -50,63 +67,114 @@ where
...
+                    }
+                    Either::Right(((completed_task, _, remaining_tasks), m_fut)) => {
+                        completed_task.map_err(|e| Error::with_message(ErrorKind::Other, e.to_string()))?;
+                        task_bucket = remaining_tasks;
+                        message_fut = m_fut;
+                    }
                 }
-                None => break,
             }
+            match channel_message {
```
jalauzon-msft: This looks to me like we are only ever processing one "completed" message at a time and then going back to the top of the loop to queue up more. What if multiple messages finished around the same time? Would there be any efficiency gains to processing multiple here and then queueing multiple at the top of the loop?

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
jaschrep-msft: That's so right. We do prevent this case elsewhere and should prevent it here.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
jaschrep-msft: I don't have a good sense of whether that would help perf. I lean "no," but don't really have a reason.

Given we've hit a target GA perf, I'd rather save this for my post-GA investigations.

jalauzon-msft: Sounds good. If we ever need to squeeze more perf out, this _could_ be something to look at.

PR #3951, opened 2026-03-13
Get cosmos driver integration tests running again

analogrelay:
Marks Cosmos Driver package as shipping (since it will be in the next release wave) and fixes up the integration test. Integrates emulator tests into CI.

In addition, I added a bit of tracing because I had to fix some bugs in the pipeline. Specifically:

* We weren't generating the correct URL for creating an item. We included the item ID in the URL, but it should be a POST to the container, not to the future item.
* We weren't properly detecting the emulator and disabling cert verification.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
+++ b/sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
@@ -123,7 +123,7 @@ pub(crate) async fn execute_operation_pipeline(
         } else {
             ExecutionContext::RegionFailover
         };
-        attempt_span.record("context", &tracing::field::debug(&execution_context));
+        attempt_span.record("context", tracing::field::debug(&execution_context));
         tracing::debug!(routing_decision = %routing, "routing decision made");
 
         let transport_request = build_transport_request(


eng/scripts/Test-Semver.ps1
```diff
@@ -64,6 +64,11 @@ foreach ($packageName in $outputPackageNames) {
     continue
   }
 
+  if ($output -match 'not found in registry') {
+    LogWarning "$packageName has not been published yet and will be ignored"
+    continue
+  }
+
```
analogrelay, line 71: @heaths It seems like this is necessary because `cargo semver-checks` fails if it can't find the previous version on crates.io. Since we're bringing a new package into the picture (`azure_data_cosmos_driver`), that's causing an error.

What wasn't entirely clear to me is if this has already been an issue for any of the other packages added in this repo. I know several, like `azure_data_cosmos` did exist prior to the "reboot", but there are also completely new packages like typespec ones. Is there some other way to handle new packages and semver-checks? Or did we just happen to be the first one to be added after semver-checks were added?

Anyway, since new packages will be added more frequently after GA, this seems like a reasonable fix to have in general. I've made it in our release branch but since we'll be merging to main soon, it should be OK.

heaths: Are you sure there's not a yanked `bin` crate in its place? Last I checked, absent packages weren't a problem, but all those placeholder `bin` crates I published were, which I already added a workaround for.

I don't see any problems with this, just have some doubt as to whether this is the right fix or if something else is wrong.

analogrelay: I'm pretty sure there isn't, and it's `azure_data_cosmos_driver` that's failing, which I don't imagine you set up a placeholder for.

sdk/cosmos/azure_data_cosmos_driver/Cargo.toml
```diff
@@ -34,6 +34,7 @@ uuid = { workspace = true, features = ["v4", "fast-rng"] }
 azure_identity.workspace = true
 serde = { workspace = true, features = ["derive"] }
 tokio = { workspace = true, features = ["rt-multi-thread", "macros", "time"] }
+tracing-subscriber = { workspace = true, features = ["env-filter", "fmt"] }
 uuid = { workspace = true, features = ["v4", "fast-rng"] }
```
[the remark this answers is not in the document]
analogrelay: Copilot, you crazy. This IS in the dev-dependencies.

heaths (approved):

PR #3952, opened 2026-03-13
[Storage] Generate `azure_storage_blob` after correcting typespec `client.tsp` -> `routes.tsp`

vincenttran-msft:
This regenerates against this corrected tsp: https://github.com/Azure/azure-rest-api-specs/pull/41387

The one change was something we've wanted but was silently swallowed because we weren't using the rename properly. So this properly renames it

### DIFF
--- a/sdk/storage/azure_storage_blob/tsp-location.yaml
+++ b/sdk/storage/azure_storage_blob/tsp-location.yaml
@@ -1,4 +1,4 @@
 directory: specification/storage/Microsoft.BlobStorage
-commit: 127fa75316efb325b0cce50140871a25f50a1ca1
+commit: b484d68da2ad57a40cb9371c77e56e6bd240810e
 repo: Azure/azure-rest-api-specs
 additionalDirectories: 


sdk/storage/azure_storage_blob/src/clients/block_blob_client.rs
```diff
@@ -196,7 +196,7 @@ impl BlockBlobClient {
             if_none_match: options.if_none_match,
             if_tags: options.if_tags,
             if_unmodified_since: options.if_unmodified_since,
-            immutability_policy_expiry: options.immutability_policy_expiry,
+            expiry: options.immutability_policy_expiry,
             immutability_policy_mode: options.immutability_policy_mode,
```
vincenttran-msft, line 200: So this rename was dormant in `client.tsp` for awhile. Want to get your take here:

- Option A: Also rename `immutability_policy_mode` -> `mode` to align both
- Option B: Leave as-is (with immutability_policy prefix on both)
- Option C: Leave as-is state of the diff (expiry without, mode with)

vincenttran-msft: Discussed offline, take B

PR #3959, opened 2026-03-16
Pre-Push Verification Script

analogrelay:
Honestly, I'm tired of getting random CI failures from things like spell check, cargo fmt, or other things I forget to run before pushing. Especially when I'm building more and more in CLI-based Agent Harnesses. VS Code extensions help, but they aren't really "gates" in the traditional sense.

So, this PR adds two OPTIONAL tools to help users out with this IF they want to use them (most critical, nothing here changes the required flow, it's all opt-in):

`eng/scripts/Invoke-PrePush.ps1` is a PowerShell script that can be invoked with NO arguments, while sitting on a branch, to perform _mostly_ the same checks as the CI in almost exactly the same **way** as the CI. I wish it could be perfectly identical, but with AzDo yaml being the authoritative source for these checks, it's always going to be stuck as a separate script. Still, for now, it's quite functional. It does use existing scripts like `Analyze-Code` and `Test-Packages` though, so it's really just the yaml orchestration that's being cloned here.

`.github/skills/pre-push-verify` is a GitHub Copilot Skill that directs an agent to run the above script and present the results.

I've done some testing of this locally and it's quite effective. It catches most issues I try to introduce. Remembering to run it can be a challenge, but if it's working out, we could put it in AGENTS.md and instruct agents to use it when finishing a task.

### DIFF
--- a/eng/scripts/Analyze-Code.ps1
+++ b/eng/scripts/Analyze-Code.ps1
@@ -33,6 +33,8 @@ if ($cargoAuditCmd) {
   $versionOutput = cargo audit --version 2>$null
   if ($versionOutput -match '(\d+\.\d+\.\d+)') {
     $installedVersion = $Matches[1]
+  } else {
+    LogWarning "Unable to parse cargo-audit version from output: '$versionOutput'. Will reinstall cargo-audit $requiredVersion."
   }
 }
 if ($installedVersion -ne $requiredVersion) {
--- a/eng/scripts/Invoke-PrePush.ps1
+++ b/eng/scripts/Invoke-PrePush.ps1
@@ -13,8 +13,8 @@ that the pull request CI pipeline performs. Use this script before pushing to
 gain high confidence that CI will pass.
 
 The script determines which packages have changed relative to a target branch
-(default: main) and runs checks against those packages. You can also specify
-packages explicitly.
+(default: main) and runs checks against those packages. You can also use the
+-PackageNames parameter to restrict checks to a subset of these detected packages.
 
 All check steps run to completion even when earlier steps fail, so you can
 discover every issue in a single invocation.
@@ -178,10 +178,10 @@ Pre-push verification
 "@
 
 # Try to fetch the target branch ref for diffing. If this fails (e.g., no
-# credentials or network access), warn and continue with the local ref.
+# credentials or network access), warn and continue with the existing local origin ref.
 Invoke-LoggedCommand "git fetch origin $TargetBranch --quiet" -GroupOutput -DoNotExitOnFailedExitCode
 if ($LASTEXITCODE -and $LASTEXITCODE -ne 0) {
-  LogWarning "Could not fetch origin/$TargetBranch (exit code $LASTEXITCODE). Falling back to local ref."
+  LogWarning "Could not fetch origin/$TargetBranch (exit code $LASTEXITCODE). Continuing with the existing local origin/$TargetBranch ref, which may be missing or out of date."
 }
 
 # ---------------------------------------------------------------------------
@@ -260,7 +260,7 @@ else {
   }
 }
 
-# If the caller specified explicit package names, filter the PackageInfo files.
+# If the caller specified explicit package names, filter the detected changed packages to that subset.
 if ($PackageNames) {
   if (Test-Path $PackageInfoDirectory) {
     foreach ($file in $packageInfoFiles) {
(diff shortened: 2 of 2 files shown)


eng/scripts/Analyze-Code.ps1
```diff
@@ -26,7 +26,20 @@ if ($Deny) {
 }
 
 $cargoAuditVersionParams = Get-VersionParamsFromCgManifest cargo-audit
-Invoke-LoggedCommand "cargo install cargo-audit --locked $($cargoAuditVersionParams -join ' ')" -GroupOutput
+$requiredVersion = $cargoAuditVersionParams[1]
```
analogrelay, line 29: Basically, since cargo-audit may be used on dev machines (while running the pre-push script) this just helps avoid a call to `cargo install` if the tool is already installed.

.github/skills/pre-push-verify/SKILL.md
```diff
@@ -0,0 +1,87 @@
...
+- `<step>.log` — merged stdout/stderr for each step
+- `summary.json` — overall result and per-step status (`passed`, `failed`, or `skipped`)
+
+## Parameters
+
+| Parameter | Default | Description |
+|-----------|---------|-------------|
+| `-PackageNames` | _(detected)_ | Explicit list of package names to check instead of auto-detecting from the diff |
+| `-Toolchain` | `stable` | Rust toolchain passed to analysis and test steps |
+| `-TargetBranch` | `main` | Branch to diff against when detecting changed packages |
```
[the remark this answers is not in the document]
analogrelay: It doesn't?

.github/skills/pre-push-verify/SKILL.md
```diff
@@ -0,0 +1,87 @@
+---
+name: pre-push-verify
+description: Run pre-push CI verification checks locally and collect results.
+---
+
+# Pre-push verification
+
+Run `eng/scripts/Invoke-PrePush.ps1` to execute the same analysis, testing, spell check, and link verification steps that the pull request CI pipeline performs. The script detects which packages changed relative to a target branch (default: `main`) and runs checks only for those packages.
```
heaths, line 8: I really don't want to have to maintain the same steps in multiple places. Can't we call this script (probably rename it since it's no longer pre-push) instead of all those steps in the pipeline? /cc @danieljurek @hallipr

.github/skills/pre-push-verify/SKILL.md
```diff
@@ -0,0 +1,87 @@
...
+Run from the repository root:
+
+```pwsh
+./eng/scripts/Invoke-PrePush.ps1
+```
+
+To collect machine-readable results and per-step log files, pass `-OutputDir`:
+
+```pwsh
+./eng/scripts/Invoke-PrePush.ps1 -OutputDir ./target/prepush
```
heaths, line 21: Why not just capture stdout like 1) all our scripts alreayd do, and 2) most instructions seem to capture anyway e.g., `2>&1`.

.github/skills/pre-push-verify/SKILL.md
```diff
@@ -0,0 +1,87 @@
...
+When `-OutputDir` is set the script writes:
+
+- `<step>.log` — merged stdout/stderr for each step
+- `summary.json` — overall result and per-step status (`passed`, `failed`, or `skipped`)
+
+## Parameters
+
+| Parameter | Default | Description |
+|-----------|---------|-------------|
+| `-PackageNames` | _(detected)_ | Explicit list of package names to check instead of auto-detecting from the diff |
```
heaths, line 33: Nit: idiomatically, PowerShell 0..n or 1..n parameters are singular i.e., `PackageName`.

eng/scripts/Invoke-PrePush.ps1
```diff
@@ -0,0 +1,431 @@
...
+    if ($logFile) {
+      & $Action *>&1 | Tee-Object -FilePath $logFile | Out-Host
+    }
+    else {
+      & $Action | Out-Host
+    }
+
+    $ErrorActionPreference = $savedErrorPref
+
+    if ($LASTEXITCODE -and $LASTEXITCODE -ne 0) {
```
heaths, line 143: Not idiomatic powershell. It's truthy, so just `if ($LASTEXITCODE)` is completely sufficient. If it's not undefined, not `0`, not `""`, not `$alse`, and not `$null` it's true.

.github/skills/pre-push-verify/SKILL.md
analogrelay: In my testing this seemed more consistent when agents were running the script but I'm happy to start with stdout/stderr

.github/skills/pre-push-verify/SKILL.md
analogrelay: I absolutely hate that too, I just hate not having confidence my PR will go green even more

analogrelay: I would LOVE to just have a "do all the CI checks" script that's also used by CI. I just don't know how feasible that is and want to find solutions that don't have to wait for us to redesign the pipeline

PR #3960, opened 2026-03-16
[Cosmos] Add additional response headers

### LINKED ISSUE #3905: [Cosmos] Allow access to additional response headers
With the current model we have defined for responses, we have chosen to only share a specific set of response headers to the user, like `session_token`, `request_charge` and `etag`. However, our customers might have applications that seek additional response headers in their responses, some like:
- Index Metrics
- Query Metrics
- LSN
- Request Duration
to name a few - we should re-evaluate the response headers we currently return to see whether we should expand the list of these to accommodate further uses.

tvaron3:
## Summary

Adds type-safe response types for Cosmos DB operations with access to additional response headers for diagnostics and metrics. Fixes #3905.

## Design

Uses dedicated wrapper types for each operation category. A design spec is at `sdk/cosmos/azure_data_cosmos/docs/RESPONSE_METADATA_SPEC.md`.

### Response Types

| Type | Operations | Extra Fields |
|------|-----------|-------------|
| `ItemResponse<T>` | create/read/replace/upsert/delete item | `etag() -> Option<&Etag>` |
| `ResourceResponse<T>` | create/read/delete database/container, throughput | (future-proof placeholder) |
| `BatchResponse` | execute_transactional_batch | `etag() -> Option<&Etag>` |
| `QueryFeedPage<T>` | query_items, query_containers, query_databases | `index_metrics()`, `query_metrics()` |
| `FeedPage<T>` | generic feed page (reusable for future read-many, change feed) | common fields only |
| `CosmosDiagnostics` | all operations (via `diagnostics()`) | `activity_id()`, `server_duration_ms()` |

### Principle

Each operation category gets a dedicated wrapper type. Common accessors (`request_charge()`, `session_token()`, `diagnostics()`) are on all types. Operation-specific fields are only on the relevant type. `QueryFeedPage<T>` composes over `FeedPage<T>`, adding query-specific metadata. `CosmosResponse<T>` is internal (`pub(crate)`).

## Driver Changes (`azure_data_cosmos_driver`)

- Added `index_metrics`, `query_metrics`, `server_duration_ms`, and `lsn` fields to `CosmosResponseHeaders`.
- `index_metrics` is base64-decoded (matching Java/.NET) with `tracing::warn!` on decode failure.
- `server_duration_ms` filters non-finite and negative values.
- Added `server_duration_ms` to `RequestDiagnostics`, populated from response headers in transport pipeline.
- Made `from_headers()` public for cross-crate access.

## SDK Changes (`azure_data_cosmos`)

- `ItemResponse<T>` wraps `CosmosResponse<T>` with `etag()` using `azure_core::http::Etag`.
- `ResourceResponse<T>` wraps `CosmosResponse<T>` for resource management operations.
- `BatchResponse` wraps `CosmosResponse<TransactionalBatchResponse>` with batch-level `etag()`.
- `QueryFeedPage<T>` composes over `FeedPage<T>`, adding `index_metrics()` and `query_metrics()`.
- `FeedPage<T>` is a public generic type with common feed fields, reusable for future read-many and change feed.
- `CosmosDiagnostics` provides `activity_id()` and `server_duration_ms()` on all response types.
- SDK delegates all header parsing to the driver's `CosmosResponseHeaders`.

## Tests

- Driver unit tests: base64 decode, invalid base64, NaN/inf/negative filtering, JSON serialization.
- SDK unit tests: wrapper types, diagnostics accessors, etag with `Etag` type, edge cases.
- Integration tests: `assert_response` helper validates `diagnostics().activity_id()` and `diagnostics().server_duration_ms()` on all point operations; `query_returns_index_and_query_metrics` validates query-specific metadata with opt-in headers.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
+++ b/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
@@ -4,7 +4,6 @@
 
 ### Features Added
 
-- Added type-safe response wrappers (`ItemResponse<T>`, `ResourceResponse<T>`, `BatchResponse`, `QueryFeedPage<T>`) with operation-specific accessors like `etag()`, `index_metrics()`, `query_metrics()`, and universal `diagnostics()` providing `activity_id()` and `server_duration_ms()`. ([#3960](https://github.com/Azure/azure-sdk-for-rust/pull/3960))
 - Added `CustomResponseBuilder` and `FaultInjectionRule::hit_count()` APIs for fault injection, enabling ergonomic construction of synthetic HTTP responses and test verification of rule activation counts. ([#3888](https://github.com/Azure/azure-sdk-for-rust/pull/3888))
 
 ### Breaking Changes
--- a/sdk/cosmos/azure_data_cosmos_driver/CHANGELOG.md
+++ b/sdk/cosmos/azure_data_cosmos_driver/CHANGELOG.md
@@ -6,9 +6,7 @@
 
 - Initial release of `azure_data_cosmos_driver` (core Cosmos DB protocol implementation for cross-language SDK reuse). ([#3772](https://github.com/Azure/azure-sdk-for-rust/pull/3772) and [#3592](https://github.com/Azure/azure-sdk-for-rust/pull/3592))
 - Added cache priming via `CosmosDriver::initialize()` and `CosmosDriver::prime_container()` to avoid cold-start latency on the first data-plane operation. ([#3864](https://github.com/Azure/azure-sdk-for-rust/pull/3864))
-- Added `index_metrics`, `query_metrics`, `server_duration_ms`, and `lsn` fields to `CosmosResponseHeaders` for access to additional response metadata. The `index_metrics` field is base64-decoded from the raw header value.
-- Added `server_duration_ms` to `RequestDiagnostics`, populated from the `x-ms-request-duration-ms` response header.
-- Added fault injection support behind `fault_injection` feature flag for transport-level fault injection per Transport Pipeline Spec §7.
+- Added response metadata fields (`index_metrics`, `query_metrics`, `server_duration_ms`, `lsn`) to `CosmosResponseHeaders` and `RequestDiagnostics`, with base64 decoding for `index_metrics`. ([#3960](https://github.com/Azure/azure-sdk-for-rust/pull/3960))
 
 ### Breaking Changes
 


sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
```diff
@@ -87,6 +87,32 @@ impl<T> CosmosResponse<T> {
...
+    pub fn activity_id(&self) -> Option<&str> {
+        self.get_optional_header_str(&crate::constants::ACTIVITY_ID)
+    }
+
+    /// Returns the index utilization metrics as a JSON string, if available.
+    ///
+    /// Only populated when the request included the `x-ms-cosmos-populateindexmetrics` header.
+    pub fn index_metrics(&self) -> Option<&str> {
+        self.get_optional_header_str(&crate::constants::INDEX_METRICS)
+    }
```
xinlian12: 🟡 **Recommendation: `index_metrics` likely needs base64 decoding**

The Cosmos DB service returns the `x-ms-cosmos-index-utilization` header as a **base64-encoded** JSON string. All three other Azure SDKs decode it before returning:

- **Java**: `Utils.decodeBase64String()` → JSON parse → `IndexUtilizationInfo`
- **.NET**: base64 decode → JSON parse → pretty-printed string
- **Python**: `base64.b64decode()` → `json.loads()` → dict

The doc comment here says "Returns the index utilization metrics as a JSON string" — but callers will receive an opaque base64 string, not valid JSON. Customers would need to manually decode it, which isn't documented.

**Suggested fix**: Either base64-decode the value before returning (matching all other SDKs), or update the doc comment to note the value is base64-encoded and needs decoding. The driver-layer field and test should be updated similarly.

**Also**: the driver-layer test uses raw JSON (`r#"{"UtilizedSingleIndexes":[]}"#`) which doesn't exercise the real wire format. If decoding is added, the test should use a base64-encoded value.

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_headers.rs
```diff
@@ -94,6 +98,23 @@ pub struct CosmosResponseHeaders {
 
     /// Cosmos substatus code (`x-ms-substatus`).
     pub substatus: Option<SubStatusCode>,
+
+    /// Index utilization metrics as a JSON string (`x-ms-cosmos-index-utilization`).
+    ///
+    /// Only populated when the `x-ms-cosmos-populateindexmetrics` request header is set.
+    pub index_metrics: Option<String>,
```
xinlian12: 🟡 **Recommendation: Same base64 decoding concern at the driver layer**

This field stores the raw `x-ms-cosmos-index-utilization` header value, but the Cosmos service base64-encodes this header. See the [inline comment on `cosmos_response.rs`](https://github.com/Azure/azure-sdk-for-rust/pull/3960) for the full cross-SDK comparison — Java, .NET, and Python all decode before surfacing this value.

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+- Added `activity_id()`, `index_metrics()`, `query_metrics()`, and `server_duration_ms()` convenience methods to `CosmosResponse<T>` for access to additional response metadata.
```
xinlian12: ## 🟢 Overall Review

**Summary:** Clean, well-patterned additive change. The code follows every established pattern in the codebase: naming, parsing strategy, helper delegation, documentation, and test structure. No regressions found.

**Design:** The `lsn` driver-only exposure correctly mirrors Java/.NET where LSN is internal. The raw string approach for `query_metrics` matches Python. `activity_id()` fills a gap in the SDK's public surface.

**Tests:** Driver-layer tests are thorough. SDK-layer accessor tests are absent, but this matches the pre-existing pattern (existing comments already cover this).

**One finding:** `index_metrics` likely returns base64-encoded data from the service, not valid JSON — see inline comments. All other SDKs decode it before returning.

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_headers.rs
```diff
@@ -129,6 +150,18 @@ impl CosmosResponseHeaders {
...
                 .and_then(SubStatusCode::from_header_value),
+            index_metrics: headers
+                .get_optional_str(&response_header_names::INDEX_METRICS)
+                .map(|s| s.to_owned()),
+            query_metrics: headers
+                .get_optional_str(&response_header_names::QUERY_METRICS)
+                .map(|s| s.to_owned()),
+            server_duration_ms: headers
+                .get_optional_str(&response_header_names::SERVER_DURATION_MS)
+                .and_then(|s| s.parse::<f64>().ok()),
```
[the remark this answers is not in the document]
tvaron3: Addressed: `server_duration_ms` is now filtered with `.filter(|v| v.is_finite())` in both the driver's `CosmosResponseHeaders::from_headers()` and the SDK. Non-finite values return `None`.

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
```diff
@@ -87,6 +87,32 @@ impl<T> CosmosResponse<T> {
...
+    /// The value is a semicolon-delimited string of key=value pairs.
+    /// Only populated when the request included the `x-ms-documentdb-populatequerymetrics` header.
+    pub fn query_metrics(&self) -> Option<&str> {
+        self.get_optional_header_str(&crate::constants::QUERY_METRICS)
+    }
+
+    /// Returns the server-side request processing duration in milliseconds, if available.
+    pub fn server_duration_ms(&self) -> Option<f64> {
+        self.get_optional_header_str(&crate::constants::REQUEST_DURATION_MS)
+            .and_then(|s| s.parse::<f64>().ok())
```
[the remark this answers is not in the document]
tvaron3: Addressed: `server_duration_ms()` now lives on `CosmosDiagnostics` (accessed via `response.diagnostics().server_duration_ms()`) and the underlying driver parsing filters non-finite values.

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
```diff
@@ -87,6 +87,32 @@ impl<T> CosmosResponse<T> {
...
+    }
```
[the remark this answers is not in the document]
tvaron3: Addressed: Added unit tests for all accessors — `activity_id`, `index_metrics` (base64 decode), `query_metrics`, `server_duration_ms` (including NaN/inf rejection), and missing-headers-return-none. Also added integration tests verifying these on point operations and query pages.

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_headers.rs
tvaron3: Addressed: Same fix at the driver layer — from_headers() now base64-decodes x-ms-cosmos-index-utilization with tracing::warn! on decode failures. Tests updated to use base64-encoded values.

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
tvaron3: updated

sdk/cosmos/azure_data_cosmos/docs/RESPONSE_METADATA_SPEC.md
```diff
@@ -0,0 +1,281 @@
...
+response.index_metrics(); // ERROR: method not found
+
+// This doesn't compile — no etag on query pages:
+let page: FeedPage<MyItem, QueryMetadata> = ...;
+page.etag(); // ERROR: method not found
+```
+
+### Decisions made
+
+1. **Second generic param** (`CosmosResponse<T, M = ()>`) preferred over wrapper types.
```
analogrelay: This is still a little bit open. Tomas and I have discussed it, and I'm kinda 50/50 on it so would love some extra feedback. Basically, different operations may have different response headers. We have three ways we can model that:

1. One single `CosmosResponse<T>` (`T` is the body type) with the union of all headers as `Option`s that are just `None` if they're not applicable. **This is what the driver will return no matter what, because it doesn't have separate APIs per operation.**
2. One single `CosmosResponse<T, M>`, `M` is this "metadata" type that varies by operation and contains operation-specific response headers.
3. SDK wrappers, `ItemResponse<T>` (point read or content-response-on-write), `QueryResponse<T>`, `ChangeFeedResponse`, etc.

In a language like C#, we'd almost certainly use 3, but Rust has no data inheritance (which is good!). If we do option 3, we'd have structs that look like this in the SDK:

```rust
pub struct ItemResponse<T> {
  response: CosmosResponse<T> // From driver
}
```

Then `ItemResponse<T>` would have method accessors to get only the things that are relevant to it (including the raw `CosmosResponse` if we want).

I think in typing it out, maybe I like option 3 better now 😅 . @tvaron3

sdk/cosmos/azure_data_cosmos/src/models/response_metadata.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// Returns the server-side request processing duration in milliseconds, if available.
+    pub fn server_duration_ms(&self) -> Option<f64> {
+        self.server_duration_ms
+    }
+}
+
+/// Metadata specific to point item operations (create, read, replace, upsert, delete).
+#[derive(Debug, Clone, Default)]
+pub struct ItemMetadata {
+    etag: Option<String>,
```
analogrelay: Should be an `azure_core::Etag`

sdk/cosmos/azure_data_cosmos/src/models/response_metadata.rs
```diff
@@ -0,0 +1,133 @@
...
+/// Structurally identical to [`ItemMetadata`] but kept separate because batch-level
+/// ETags have different semantics than single-item ETags, and the types may diverge
+/// in the future.
+///
+/// Note: The batch-level ETag differs from a single-item ETag. It represents
+/// the ETag for the entire batch operation, not an individual item's concurrency token.
+/// Use individual `TransactionalBatchOperationResult` entries for per-item ETags.
+#[derive(Debug, Clone, Default)]
+pub struct BatchMetadata {
+    etag: Option<String>,
```
analogrelay: Another place for `azure_core::Etag`

sdk/cosmos/azure_data_cosmos/docs/RESPONSE_METADATA_SPEC.md
tvaron3: Made the changes for option 3

sdk/cosmos/azure_data_cosmos/src/models/response_metadata.rs
tvaron3: Updated

sdk/cosmos/azure_data_cosmos/src/models/response_metadata.rs
tvaron3: updated

sdk/cosmos/azure_data_cosmos/tests/multi_write_tests/cosmos_multi_write_fault_injection.rs
```diff
@@ -934,136 +933,3 @@ pub async fn fault_injection_connection_error_local_retry_succeeds() -> Result<(
     )
     .await
 }
-
-/// Test that a 403.3 (WriteForbidden) via `CustomResponse` triggers a database
```
tvaron3, line 938: Removed as this test is broken because the get database account call does not add the fault injection headers because it is not part of the cosmos pipeline. This is not the case in the driver so once we pass over the driver this test can be readded.

sdk/cosmos/azure_data_cosmos/src/models/item_response.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    /// Returns a reference to all response headers.
+    pub fn headers(&self) -> &Headers {
+        self.response.headers()
+    }
+
+    /// Gets an optional header value as a string by name.
+    pub fn get_optional_header_str(&self, name: &HeaderName) -> Option<&str> {
+        self.response.get_optional_header_str(name)
+    }
```
analogrelay: Can we get rid of this function from the wrapper types, at least until we need it? The `headers()` accessor should give the user everything they need here, and I don't want to introduce new helpers as public APIs until we know how important they are.

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -4,10 +4,22 @@
 
 ### Features Added
 
+- Added dedicated response wrapper types: `ItemResponse<T>` (point operations), `ResourceResponse<T>` (resource management), `BatchResponse` (transactional batch), and `QueryFeedPage<T>` (query pages).
+- Added `CosmosDiagnostics` type accessible via `diagnostics()` on all response types, providing `activity_id()` and `server_duration_ms()`.
+- `QueryFeedPage<T>` provides typed `index_metrics()` and `query_metrics()` accessors. The `index_metrics()` method base64-decodes the raw header value to return valid JSON.
+- `ItemResponse<T>` and `BatchResponse` expose `etag()` returning `Option<&azure_core::http::Etag>`.
+- Added `request_charge()`, `session_token()`, and `diagnostics()` convenience methods to `FeedPage` and `QueryFeedPage`.
```
simorenoh: this + the ones below seems like too much (noise)? also missing PR link, but ideally we get to summarize or group some together if possible

sdk/cosmos/azure_data_cosmos/src/models/item_response.rs
tvaron3: Removed from all the wrapper types.

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
tvaron3: Agreed. grouped and summarized and added pr link

PR #3961, opened 2026-03-16
[Internal] Add automated perf deployment infrastructure

tvaron3:
## Automated Perf Deployment Infrastructure

Adds infrastructure scripts and AKS deployment for continuous Cosmos DB Rust SDK performance testing.

### Deploy scripts (`deploy/`)

Modular deployment scripts for provisioning all required Azure resources:

- **`deploy-aks.sh`** — Orchestrator that calls subscripts in order
- **`setup-cosmos-db.sh`** — Cosmos DB account, database, container creation (with ARM template for `disableLocalAuth`)
- **`setup-identity-rbac.sh`** — Managed identity + Cosmos DB native RBAC (data plane)
- **`setup-aks.sh`** — AKS cluster + ACR + pull secrets
- **`setup-k8s-identity.sh`** — K8s namespace, service account, federated credential for workload identity
- **`setup-adx.sh`** — Azure Data Explorer cluster, database, tables, ingestion mappings, Cosmos DB change feed connection
- **`deploy-k8s-job.sh`** — K8s Job deployment + optional CronJob for branch polling
- **`run_perf.sh`** — VM-based multi-process launcher (moved from crate root)

### Features

- **One-command deployment**: `./deploy-aks.sh --create-workload-account --create-results-account --location eastus2`
- **Multi-region support**: `--multi-write` and `--regions "East US,West US"` for multi-region Cosmos accounts
- **Branch polling**: `--poll-branch release/azure_data_cosmos-previews` deploys a CronJob that checks for new commits every 12h, rebuilds, and restarts
- **Kusto integration**: Auto-creates ADX cluster with `PerfResults` and `ErrorResults` tables, change feed data connection from Cosmos results container
- **Commit tracking**: `--commit-sha` stamped on every result document for Kusto correlation
- **WorkloadIdentity auth**: Supports both AKS (WorkloadIdentityCredential) and VM (ManagedIdentityCredential)

### Perf tool changes

- `--application-region` is now required (SDK API change: `RoutingStrategy::ProximityTo`)
- `--commit-sha` auto-detected from git or passed explicitly
- `TIMESTAMP` field renamed for Kusto `datetime` compatibility
- Results container uses 10K RU/s throughput with 24h TTL
- Partition key uses random UUID for better distribution

### Note

> **Windows support**: These scripts currently require bash (Linux/macOS or WSL on Windows). A follow-up will address native Windows support through a cross-platform Rust CLI application.


sdk/cosmos/azure_data_cosmos_perf/src/main.rs
```diff
@@ -68,8 +64,14 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             builder.build(account, strategy).await?
         }
         AuthMethod::Aad => {
+            // Try WorkloadIdentityCredential first (AKS), fall back to ManagedIdentityCredential (VMs)
```
tvaron3: **🟡 Recommendation** · Reliability: Duplicated Credential Fallback with Panic

**WorkloadIdentity → ManagedIdentity fallback is copy-pasted twice and panics on failure**

This pattern appears identically at lines 67–75 (workload client) and lines 143–151 (results client):

```rust
azure_identity::WorkloadIdentityCredential::new(None)
    .map(|c| c as Arc<dyn azure_core::credentials::TokenCredential>)
    .unwrap_or_else(|_| {
        azure_identity::ManagedIdentityCredential::new(None)
            .expect("Failed to create ManagedIdentityCredential")
    });
```

The `.expect()` panics if both credential sources fail (e.g., running on a developer laptop without any managed identity). Extract to a helper function and return `Result` to fail gracefully with a descriptive error message instead of a panic.

---

sdk/cosmos/azure_data_cosmos_perf/deploy/updater.sh
```diff
@@ -0,0 +1,128 @@
...
+    --output none
+
+echo "  Image built and pushed."
+
+# Restart the perf Job with the new commit SHA
+echo "  Restarting perf job..."
+kubectl delete job cosmos-perf -n "$PERF_NAMESPACE" 2>/dev/null || true
+
+# Generate and apply the new job manifest
+cat <<JOBEOF | kubectl apply -f -
```
tvaron3: **🟡 Recommendation** · Maintainability: Duplicated Job Manifest

**Job spec is duplicated between `perf-job.yaml` and this heredoc**

The K8s Job spec here (lines 79–126) is a full copy of `perf-job.yaml`. If the job spec changes (new args, resource limits, labels), both need identical updates. A missed sync means the CronJob deploys a stale manifest.

**Fix**: Have `updater.sh` use `envsubst` on the same `perf-job.yaml` template, just like `deploy-k8s-job.sh` does. Mount the template as a ConfigMap or clone the repo (which the updater already does).

---

sdk/cosmos/azure_data_cosmos_perf/deploy/setup-adx.sh
```diff
@@ -0,0 +1,207 @@
...
+ADX_PRINCIPAL_ID=$(az kusto cluster show --name "$CLUSTER_NAME" --resource-group "$RESOURCE_GROUP" --query "identity.principalId" -o tsv 2>/dev/null)
+
+# Enable system-assigned identity on ADX cluster if not already
+if [[ -z "$ADX_PRINCIPAL_ID" || "$ADX_PRINCIPAL_ID" == "null" ]]; then
+    echo "  Enabling system-assigned identity on ADX cluster..." >&2
+    az rest --method patch \
+        --url "https://management.azure.com/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${RESOURCE_GROUP}/providers/Microsoft.Kusto/clusters/${CLUSTER_NAME}?api-version=2023-08-15" \
+        --body '{"identity":{"type":"SystemAssigned"}}' \
+        --output none
+    sleep 30
```
tvaron3: **🟢 Suggestion** · Reliability: Fixed Sleep for Identity Propagation

**`sleep 30` is fragile for waiting on identity propagation**

After enabling system-assigned identity on the ADX cluster, a fixed 30s sleep waits for the principal ID to propagate. In busy subscriptions or slow ARM backends, 30s may not be enough; in fast environments, it wastes time.

Consider polling `az kusto cluster show` for `identity.principalId` until non-empty, with a timeout:
```bash
for i in $(seq 1 30); do
    ADX_PRINCIPAL_ID=$(az kusto cluster show ... --query "identity.principalId" -o tsv)
    [[ -n "$ADX_PRINCIPAL_ID" && "$ADX_PRINCIPAL_ID" != "null" ]] && break
    sleep 5
done
```

---

sdk/cosmos/azure_data_cosmos_perf/deploy/deploy-k8s-job.sh
```diff
@@ -0,0 +1,131 @@
...
+done
+
+if [[ -z "$ACR_NAME" || -z "$COSMOS_ENDPOINT" || -z "$RESULTS_ENDPOINT" || -z "$IDENTITY_CLIENT_ID" || -z "$RESOURCE_GROUP" ]]; then
+    echo "Error: --acr-name, --cosmos-endpoint, --results-endpoint, --identity-client-id, and --resource-group are required." >&2
+    exit 1
+fi
+
+# --- Build and push Docker image ---
+echo "  Creating build context..."
+BUILD_CONTEXT=$(mktemp -d /tmp/cosmos-perf-context-XXXXXX)
```
tvaron3, line 83: **🟢 Suggestion** · Reliability: Build Context Leak on Failure

**Temp directory not cleaned up if `az acr build` fails**

`BUILD_CONTEXT` is created here but only cleaned up after a successful build (line 99). Since `set -e` is active, a build failure exits immediately, leaking the temp directory.

**Fix**: Add a trap after creation:
```bash
BUILD_CONTEXT=$(mktemp -d /tmp/cosmos-perf-context-XXXXXX)
trap 'rm -rf "$BUILD_CONTEXT"' EXIT
```

---

sdk/cosmos/azure_data_cosmos_perf/src/runner.rs
```diff
@@ -250,9 +260,10 @@ async fn upsert_results(
     for s in summaries {
         let result = PerfResult {
             id: Uuid::new_v4().to_string(),
-            partition_key: s.name.clone(),
+            partition_key: Uuid::new_v4().to_string(),
```
tvaron3, line 263: **💬 Observation** · Design: Partition Key Trade-off

**Partition key changed from operation name to random UUID**

Good change for write distribution across physical partitions. Trade-off: queries by operation name now require cross-partition scans. Since perf results are low volume and queried via Kusto (not direct Cosmos queries), this is fine. Just noting for awareness.

---

sdk/cosmos/azure_data_cosmos_perf/src/main.rs
tvaron3: Fixed — extracted credential fallback into `create_aad_credential()` helper that returns `Result` instead of panicking. Both the workload client and results client now use this helper.

sdk/cosmos/azure_data_cosmos_perf/deploy/updater.sh
tvaron3: Fixed — replaced the heredoc with `envsubst < perf-job.yaml | kubectl apply -f -` using the template from the cloned repo, eliminating the duplication.

sdk/cosmos/azure_data_cosmos_perf/deploy/setup-adx.sh
tvaron3: Fixed — replaced `sleep 30` with a polling loop that checks `identity.principalId` every 5 seconds (up to 30 attempts), failing with a descriptive error if it doesn't propagate.

sdk/cosmos/azure_data_cosmos_perf/deploy/deploy-k8s-job.sh
tvaron3: Fixed — added `trap 'rm -rf "$BUILD_CONTEXT"' EXIT` after `mktemp`. The temp directory is now cleaned up on any exit, including build failures.

sdk/cosmos/azure_data_cosmos_perf/src/runner.rs
tvaron3: Acknowledged — correct, perf results are queried via Kusto (change feed), not direct Cosmos queries, so the cross-partition trade-off is acceptable.

.dockerignore
heaths, line 1: Was this supposed to be checked in? As much as I'd like to, we don't support building our SDKs in docker containers in other Azure SDK repos. I also don't see how this relates to deploying perf resources which, have you looked at our existing perf framework we already use across all our other Azure SDK repos? /cc @LarryOsterman @mikeharder

LarryOsterman: I would recommend that if the Cosmos team wants to develop their own perf infrastructure, they host it in their own private repository. That's what the storage team has done for their managed upload and download scenarios. It allows allows the storage team to rapidly iterate on performance prototypes (they have a cargo.toml file which depends on a git repo and branch combination, so they don't need to reference official builds).

heaths: I think some or even all source code could live here - the actual perf tests, just as we have for all other crates (and packages across all languages) - because there's a close relationship. But all the infrastructure to provision and deploy does seem out of place here.

tvaron3:
Closing — infra/deployment changes (Dockerfile, deploy/ scripts) will move to a separate repo. Opening a new PR with only source file changes.

.dockerignore
analogrelay: That's basically what we're going to do. Perf _scenarios_ will live here, but the driving logic will live elsewhere. As I keep harping on, Cosmos has some pretty unique needs here since the client is _heavy_. Of course, that infra will all be inner source and we're happy to integrate where we can!

Issue #3937, opened 2026-03-11
Cosmos Driver: Fault Injection


analogrelay:
The fault injection implementation in the SDK needs to be moved over to the driver.

PR #3902, opened 2026-03-10
Cosmos Driver: Add session token cache

simorenoh:
This PR adds session token management for Session consistency to the Cosmos DB driver. The session container caches per-partition session tokens from response headers and resolves them onto outgoing requests, enabling read-your-writes guarantees within a session.

#### `SessionTokenValue` / `VectorSessionToken` — Token Parsing and Merging

- `VectorSessionToken` (V2): parses the version-vector portion of a session token string (`<version>#<globalLSN>#<regionId>=<regionLSN>#...`) and implements merge semantics (higher version, higher global LSN, per-region maximum)
- `SessionTokenValue` enum: supports both V1 (bare integer LSN, e.g. `0:500`) and V2 (version vector) token formats. Parsing tries V2 first, falls back to V1. Cross-type merge: V2 always supersedes V1
- Implements `is_at_least_as_recent_as()` comparison for future use by partition key range cache
- `pub(crate)` only — users interact with the opaque `SessionToken` string wrapper

#### `SessionContainer` — In-Memory Cache

- `RwLock<HashMap>`-based cache keyed by collection resource ID (RID)
- Each collection maps partition-key-range IDs to their `SessionTokenValue`
- Secondary `name_path → RID` index enables name-based lookups
- **RID mismatch detection**: when a name path maps to a different RID than previously seen, old tokens are cleared (the container was likely recreated)
- Uses `std::sync::RwLock` (not tokio) because the lock is never held across `.await` points

#### `SessionManager` — Pipeline-Facing

- `resolve_session_token()`: returns the user-provided token if present, otherwise looks up from cache by container RID with name-based fallback
- `capture_session_token()`: extracts session token, owner ID (`x-ms-content-path`), and owner full name (`x-ms-alt-content-path`) from response headers and merges into cache. Uses `is_reading_from_master()` to gate capture — skips master/metadata operations but correctly captures container CRUD (Create/Replace/Delete/Read), matching Java's `ReplicatedResourceClientUtils.isReadingFromMaster()`
- `clear_session_token()`: clears cached tokens by collection name path on 404/1002 (`ReadSessionNotAvailable`) during session retry

#### `ConsistencyLevel` — Typed Consistency Model

- New enum with 5 variants: `Strong`, `BoundedStaleness`, `Session`, `ConsistentPrefix`, `Eventual`
- Implements `Serialize`/`Deserialize`, `FromStr`, `Display`, and `is_session()` helper
[...]

PR #3888, opened 2026-03-09
Cosmos: Add CustomResponseBuilder and hit_count API for fault injection

tvaron3:
## Summary

Adds a fluent `CustomResponseBuilder` API and exposes rule hit counts for fault injection testing, enabling test verification of complex error-handling flows like 403.3 region failover with account property transitions.

### `CustomResponseBuilder` (new public API)

Provides a builder pattern for constructing synthetic HTTP responses in fault injection tests. Methods: `new(status_code)`, `with_header(name, value)`, `with_sub_status(code)`, `with_body(body)`, `build()`.

### `FaultInjectionRule::hit_count()` (new public API)

Adds a public `hit_count()` method to `FaultInjectionRule`, allowing tests to verify how many times a rule was matched. The `RuleState` wrapper was removed since hit tracking now lives on the rule itself.

### `mock_database_account_response_for_account()` (new test helper)

Extends `mock_account.rs` with a variant that accepts a real account name, so mock endpoint URLs resolve correctly against a live service.

### Integration test: 403.3 single-write transition

Adds `fault_injection_custom_response_403_3_transitions_to_single_write` to the multi-write fault injection test suite. The test:

1. Injects a 403.3 (WriteForbidden) via `CustomResponseBuilder` for writes targeting the hub region
2. Intercepts the `MetadataReadDatabaseAccount` call with a `CustomResponse` returning a mock account where multi-write is disabled and only the satellite region is writable
3. Verifies the write succeeds via failover to the satellite region
4. Verifies the database account refresh was triggered (via `hit_count()`)

### Files changed

- `src/fault_injection/result.rs` — Add `CustomResponseBuilder`
- `src/fault_injection/mod.rs` — Export `CustomResponseBuilder`
- `src/fault_injection/rule.rs` — Add `hit_count`, `hit_count()`, `increment_hit_count()`; add unit tests
- `src/fault_injection/http_client.rs` — Remove `RuleState` wrapper, use `Vec<Arc<FaultInjectionRule>>` directly
- `tests/framework/mock_account.rs` — Add `mock_database_account_response_for_account()`
- `tests/multi_write_tests/cosmos_multi_write_fault_injection.rs` — Add 403.3 single-write transition test

PR #3972, opened 2026-03-17
[Cosmos] Move fault injection to driver crate

tvaron3:
## Summary

Moves fault injection support into `azure_data_cosmos_driver` per Transport Pipeline Spec §7. Addresses #3937.

## Design (Transport Spec §7)

Fault injection integrates at the `HttpClient` level, *below* `AdaptiveTransport` and (future) `ShardedHttpTransport`. This placement is deliberate — injected faults are visible to shard health tracking and eviction logic, enabling tests that validate shard management under failure conditions.

```text
execute_transport_pipeline()
     |
     v
AdaptiveTransport
  +-- ShardedHttpTransport (future)
       +-- ClientShard[N]
            +-- FaultInjectingHttpClient  <-- evaluates rules
                 +-- real HttpClient      <-- actual HTTP I/O
```

### Injection via `HttpClientFactory` (§7.2)

`FaultInjectingHttpClientFactory` wraps the real `DefaultHttpClientFactory` and produces `FaultInjectingHttpClient` instances. Each shard gets its own wrapper — the sharded transport is completely unaware of interception.

### Driver (`azure_data_cosmos_driver`) — new `fault_injection` feature

6 new files in `src/fault_injection/`:

| File | Types |
|------|-------|
| `mod.rs` | `FaultInjectionErrorType` (10 variants), `FaultOperationType` (13 variants), re-exports |
| `condition.rs` | `FaultInjectionCondition`, `FaultInjectionConditionBuilder` |
| `result.rs` | `FaultInjectionResult`, `FaultInjectionResultBuilder`, `CustomResponse` |
| `rule.rs` | `FaultInjectionRule`, `FaultInjectionRuleBuilder` |
| `fault_injecting_client.rs` | `FaultInjectingHttpClient` — wraps `Arc<dyn HttpClient>`, evaluates rules (21 unit tests) |
| `fault_injecting_factory.rs` | `FaultInjectingHttpClientFactory` — wraps `HttpClientFactory`, produces fault-injecting clients |

### SDK (`azure_data_cosmos`)

- Trimmed redundant tests from `http_client.rs` (SDK retains its own types since the two crates use different `azure_core` versions, per the "No Model Sharing" versioning strategy in `AGENTS.md`)

### Architecture notes

- The SDK retains its own fault injection types rather than re-exporting from the driver. This follows the versioning strategy documented in `AGENTS.md` — `azure_data_cosmos` must maintain multi-year backward compatibility, so it cannot expose `azure_data_cosmos_driver` types directly.
- All driver fault injection code is behind `#[cfg(feature = "fault_injection")]`.
- The `FaultInjectingHttpClient` adapts the existing `FaultClient` logic from the SDK, using the driver's own `OperationType` and `ResourceType` enums for operation matching.


sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/fault_injecting_client.rs
```diff
@@ -0,0 +1,789 @@
...
+        &self,
+        server_error: &FaultInjectionResult,
+    ) -> Option<azure_core::Result<AsyncRawResponse>> {
+        // Check probability
+        if server_error.probability() < 1.0 {
+            // Use a simple time-based pseudo-random check
+            let nanos = std::time::SystemTime::now()
+                .duration_since(std::time::UNIX_EPOCH)
+                .unwrap_or_default()
+                .subsec_nanos();
```
tvaron3: **🟡 Recommendation** · Correctness: Non-Random Probability Check

**`subsec_nanos` is deterministic, not random — defeats probabilistic injection**

```rust
let nanos = std::time::SystemTime::now()
    .duration_since(std::time::UNIX_EPOCH)
    .unwrap_or_default()
    .subsec_nanos();
let random = (nanos % 1000) as f32 / 1000.0;
```

`subsec_nanos` is the nanosecond component of wall-clock time, not a random value. On fast machines, consecutive requests within the same microsecond get identical values. Requests arriving at regular intervals (e.g., every 1ms from a benchmark loop) produce perfectly predictable patterns, defeating probabilistic injection.

The `rand` crate is already a dependency of this crate (`Cargo.toml` line 20). Use `rand::random::<f32>()` instead.

---

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/http_client.rs
```diff
@@ -0,0 +1,789 @@
...
+        // Generate the appropriate error based on error type
+        let error_type = match server_error.error_type {
+            Some(et) => et,
+            None => return None, // No error type set, pass through
+        };
+
+        // Connection-level faults return simple errors with the appropriate ErrorKind.
+        // HTTP-level faults produce synthetic error responses.
+        let (status_code, sub_status, message) = match error_type {
+            FaultInjectionErrorType::ConnectionError => {
```
tvaron3, line 154: **🟡 Recommendation** · Correctness: Wrong ErrorKind for Connection Errors

**`ConnectionError` uses `ErrorKind::Io` — should use `ErrorKind::Connection` for proper retry classification**

```rust
FaultInjectionErrorType::ConnectionError => {
    return Some(Err(azure_core::Error::with_message(
        ErrorKind::Io,
        "Injected fault: connection error",
    )));
}
```

The transport pipeline's `is_connectivity_error()` checks `ErrorKind::Connection | ErrorKind::Io`, and `infer_request_sent_status()` maps `ErrorKind::Connection` to `NotSent` but `ErrorKind::Io` to `Unknown`. A real connection-refused error would be `Connection` + `NotSent`, meaning it's safe to retry even for non-idempotent writes.

Using `Io` here means the injected fault produces `Unknown` sent-status, which is more conservative than a real connection failure — shard-local retry won't fire for non-idempotent writes. Tests using this fault type will miss bugs in the `NotSent` retry path.

**Fix**: Use `ErrorKind::Connection` for `ConnectionError`, keep `ErrorKind::Io` for `ResponseTimeout`.

---

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/fault_injecting_client.rs
```diff
@@ -0,0 +1,789 @@
...
+            .into_iter()
+            .map(|rule| RuleState {
+                rule,
+                hit_count: AtomicU32::new(0),
+            })
+            .collect();
+
+        Self {
+            inner,
+            rules: Arc::new(Mutex::new(rule_states)),
```
tvaron3: **🟢 Suggestion** · Performance: Mutex on Every Request

**`Mutex<Vec<RuleState>>` serializes all requests through the fault client**

The rules vector is only read during `execute_request()` — no rules are ever added or removed at runtime. A `RwLock` would allow concurrent reads. With the current `Mutex`, all requests serialize through the fault client even in the no-match case. For high-concurrency fault injection tests (e.g., testing shard health under load), this becomes a bottleneck.

The `AtomicU32` hit count inside `RuleState` already handles concurrent increments correctly.

---

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/fault_injecting_client.rs
```diff
@@ -0,0 +1,789 @@
...
+            }
+        }
+
+        true
+    }
+
+    /// Checks if the request matches the rule's condition.
+    fn matches_condition(&self, request: &Request, rule: &FaultInjectionRule) -> bool {
+        let condition = &rule.condition;
+        let mut matches = true;
```
tvaron3: **🟢 Suggestion** · Style: Early-Return Instead of Flag Tracking

**`matches_condition` continues checking after first mismatch**

The function tracks a `matches` flag and checks all conditions even after the first mismatch. Return `false` immediately on the first mismatch for clearer control flow:

```rust
fn matches_condition(&self, request: &Request, rule: &FaultInjectionRule) -> bool {
    let condition = &rule.condition;

    if let Some(expected_op) = condition.operation_type {
        let request_op = request.headers()
            .get_optional_str(&FAULT_INJECTION_OPERATION)
            .and_then(|s| s.parse::<FaultOperationType>().ok());
        if request_op != Some(expected_op) { return false; }
    }

    if let Some(region) = &condition.region {
        if !request.url().as_str().contains(region.as_str()) { return false; }
    }

    if let Some(container_id) = &condition.container_id {
        if !request.url().as_str().contains(container_id) { return false; }
    }

    true
}
```

---

tvaron3 (commented):
## PR Review — Fault Injection for Driver Crate

Clean, well-tested implementation with good architecture (injection below AdaptiveTransport via HttpClientFactory). 4 comments (2 recommendations, 2 suggestions).

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/fault_injecting_factory.rs
```diff
@@ -0,0 +1,51 @@
...
+use crate::driver::transport::http_client_factory::{HttpClientConfig, HttpClientFactory};
+use crate::options::ConnectionPoolOptions;
+
+/// An [`HttpClientFactory`] decorator that wraps clients with fault injection.
+///
+/// When `create` is called, this factory delegates to the inner factory to build
+/// a real HTTP client, then wraps it in a [`FaultInjectingHttpClient`] that
+/// evaluates the configured rules on every request.
+#[derive(Debug)]
+#[allow(dead_code)]
```
tvaron3: Why is this dead code? There should be no dead code.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+pub use condition::{FaultInjectionCondition, FaultInjectionConditionBuilder};
+pub use fault_injecting_client::FaultInjectingHttpClient;
+#[allow(unused_imports)]
+pub(crate) use fault_injecting_factory::FaultInjectingHttpClientFactory;
+pub use result::{CustomResponse, FaultInjectionResult, FaultInjectionResultBuilder};
+pub use rule::{FaultInjectionRule, FaultInjectionRuleBuilder};
+
+/// Represents different server error types that can be injected for fault testing.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum FaultInjectionErrorType {
```
tvaron3, line 50: **🟡 Recommendation** · Convention: Missing `#[non_exhaustive]` on Public Enums

**`FaultInjectionErrorType` and `FaultOperationType` lack `#[non_exhaustive]`**

Per AGENTS.md §"Cosmos-Specific Patterns": `#[non_exhaustive] // For forward compatibility`

Both enums are `pub` and will likely grow — adding `RetryWith` (449), `Conflict` (409), or `PatchItem` would be a **breaking change** for any downstream code that exhaustively matches on them. The existing `ResourceType` and `OperationType` enums in the driver both use `#[non_exhaustive]`.

**Fix**: Add `#[non_exhaustive]` to both enums.

---

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/condition.rs
```diff
@@ -0,0 +1,78 @@
...
+//! Defines conditions for when fault injection rules should be applied.
+
+use super::FaultOperationType;
+use crate::options::Region;
+
+/// Defines the condition under which a fault injection rule should be applied.
+#[derive(Clone, Default, Debug)]
+pub struct FaultInjectionCondition {
+    /// The type of operation to which the fault injection applies.
+    pub operation_type: Option<FaultOperationType>,
```
tvaron3: **🟡 Recommendation** · Convention: Public Fields Bypass Builders

**`FaultInjectionCondition`, `FaultInjectionResult`, and `FaultInjectionRule` expose `pub` fields directly**

```rust
pub struct FaultInjectionCondition {
    pub operation_type: Option<FaultOperationType>,
    pub region: Option<Region>,
    pub container_id: Option<String>,
}
```

These types have builders, but the builders are optional — users can construct and mutate the structs directly, bypassing any future validation. The rest of the driver codebase uses `pub(crate)` fields with accessor methods for internal types.

Since these are `pub` API types (exported from the `fault_injection` module), downstream code can depend on field access. Making fields `pub` now means you can’t add validation later (e.g., clamping `probability`, validating `container_id` format) without a breaking change.

**Fix**: Use private fields with getters, and make the builder the only construction path.

---

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+    /// 500 from server.
+    InternalServerError,
```
tvaron3, line 52: **🟢 Suggestion** · Consistency: `FaultInjectionErrorType` Missing `Display`/`FromStr`

`FaultOperationType` implements both `Display` and `FromStr` (per AGENTS.md §"Prefer Standard Traits Over Custom Methods"), but `FaultInjectionErrorType` implements neither. If error types need to appear in diagnostics, log messages, or configuration, they should have the same trait coverage.

Consider adding `Display` (for logging injected faults) and `FromStr` (for config-driven fault injection) to `FaultInjectionErrorType` for parity.

---

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/fault_injecting_factory.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+impl HttpClientFactory for FaultInjectingHttpClientFactory {
+    fn build(
+        &self,
+        connection_pool: &ConnectionPoolOptions,
+        config: HttpClientConfig,
+    ) -> azure_core::Result<Arc<dyn HttpClient>> {
+        let real_client = self.inner.build(connection_pool, config)?;
+        let rules = self.rules.lock().unwrap().clone();
```
tvaron3: **🟢 Suggestion** · Convention: `.unwrap()` on Mutex Instead of Poison Recovery

```rust
let rules = self.rules.lock().unwrap().clone();
```

The codebase convention (established in PR #3902 review for `SessionContainer`) is `.unwrap_or_else(|e| e.into_inner())` for lock poisoning recovery. While this is test-only code behind a feature flag, consistency matters — a poisoned lock in a fault injection factory would crash the test harness with an unhelpful panic instead of recovering.

The same pattern appears in `execute_request()` at `fault_injecting_client.rs:135`.

---

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/fault_injecting_client.rs
```diff
@@ -0,0 +1,789 @@
...
+
+        // Check operation type if specified
+        if let Some(expected_op) = condition.operation_type {
```
tvaron3: The diagnostic context should be updated to include if the operation happened due to fault injection

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/condition.rs
```diff
@@ -0,0 +1,78 @@
...
+
+/// Builder for creating a FaultInjectionCondition.
+#[derive(Default)]
+pub struct FaultInjectionConditionBuilder {
+    operation_type: Option<FaultOperationType>,
+    region: Option<Region>,
+    container_id: Option<String>,
+}
+
+impl FaultInjectionConditionBuilder {
```
tvaron3, line 43: Include changes from https://github.com/Azure/azure-sdk-for-rust/pull/3888/changes

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/fault_injecting_client.rs
tvaron3: Fixed: replaced `subsec_nanos` with `rand::random::<f32>()`. `rand` added to Cargo.toml.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/http_client.rs
tvaron3: Fixed: `ConnectionError` now uses `ErrorKind::Connection` → `NotSent` → retries reads+writes, matching real connection failures.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/fault_injecting_client.rs
tvaron3: Fixed: `RwLock` replaces `Mutex` for concurrent rule evaluation.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/fault_injecting_client.rs
tvaron3: Fixed: `matches_condition` refactored to early-return on first mismatch.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/fault_injecting_factory.rs
tvaron3: Removed `#[allow(dead_code)]`. Added unit test that exercises the factory. The 2 remaining `dead_code` warnings are expected — the factory is `pub(crate)` but not yet wired into `CosmosTransport` (future PR).

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/mod.rs
tvaron3: Fixed: `#[non_exhaustive]` added to both `FaultInjectionErrorType` and `FaultOperationType`.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/condition.rs
tvaron3: Fixed: all fields on `FaultInjectionCondition`, `FaultInjectionResult`, `FaultInjectionRule`, and `CustomResponse` are now private with getter methods. Builders are the only construction path. `#[non_exhaustive]` added to all.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/mod.rs
tvaron3: Fixed: `Display` and `FromStr` added to `FaultInjectionErrorType` for parity with `FaultOperationType`.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/fault_injecting_factory.rs
tvaron3: Fixed: all lock acquisitions use `.unwrap_or_else(|e| e.into_inner())` for poison recovery.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/fault_injecting_client.rs
tvaron3: Added `x-ms-fault-injected: <rule-id>` header to synthetic HTTP responses for per-request diagnostics. Global `fault_injection_enabled` flag on `DiagnosticsContextBuilder` will be wired when the factory integrates into `CosmosTransport`.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/condition.rs
tvaron3: Incorporated PR #3888 changes: moved `hit_count`/`AtomicU32` into `FaultInjectionRule`, removed `RuleState`, added `CustomResponseBuilder`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -878,6 +878,10 @@ impl CosmosDriver {
         );
         diagnostics_builder.set_cpu_monitor(self.runtime.cpu_monitor().clone());
         diagnostics_builder.set_machine_id(Arc::clone(self.runtime.machine_id()));
+        if self.runtime.fault_injection_enabled() {
+            #[cfg(feature = "fault_injection")]
+            diagnostics_builder.set_fault_injection_enabled(true);
```
FabianMeiswinkel, line 886: Hmm - In java we track also which faul inejction rule was applicable (or ignore - and why) - should we add this here as well?

FabianMeiswinkel: But definitely not blocking for this PR.

tvaron3: Good idea — we already track which rule fired via the `x-ms-fault-injected: <rule-id>` response header. Tracking why rules were skipped (timing, probability miss, hit limit, condition mismatch) would be a nice follow-up for richer diagnostics. Will defer to a future PR as noted.

tvaron3: The current x-ms-fault-injected header covers the "which rule fired" case. The "why rules were skipped" evaluation log is a deeper integration requiring operation-level context propagation, so I will do it in follow up pr.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/http_client.rs
```diff
@@ -0,0 +1,786 @@
...
+
+        if let Some(expected_op) = condition.operation_type() {
+            let request_op = request
+                .headers()
+                .get_optional_str(&FAULT_INJECTION_OPERATION)
+                .and_then(|s| s.parse::<FaultOperationType>().ok());
+            if request_op != Some(expected_op) {
+                return false;
+            }
+        }
```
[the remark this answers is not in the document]
tvaron3: Fixed: `build_transport_request()` now injects `x-ms-fault-injection-operation` header from the operation/resource types when `fault_injection` feature is enabled.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/result.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    /// Sets the delay before injecting the error.
+    pub fn with_delay(mut self, delay: Duration) -> Self {
+        self.delay = Some(delay);
+        self
+    }
+
+    /// Sets the probability of injecting the error (0.0 to 1.0).
+    pub fn with_probability(mut self, probability: f32) -> Self {
+        self.probability = probability.clamp(0.0, 1.0);
```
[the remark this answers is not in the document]
tvaron3: Fixed: `with_probability()` now checks `is_finite()` first — NaN/Inf normalized to 0.0. Added test.

sdk/cosmos/azure_data_cosmos_driver/Cargo.toml
```diff
@@ -71,3 +72,4 @@ reqwest_native_tls = [
   "reqwest/native-tls",
   "azure_core/reqwest_native_tls",
 ]
+fault_injection = ["rand"]
```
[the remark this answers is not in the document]
tvaron3: Fixed: changed to `fault_injection = ["dep:rand"]` for consistency.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+//! - Operation-specific error handling
+//!
+//! # Core Components
+//!
+//! - [`FaultInjectionCondition`] — Defines when a fault should be applied, filtering by
+//!   operation type, region, or container ID.
+//! - [`FaultInjectionResult`] — Defines what error to inject, including error type, delay,
+//!   and probability.
+//! - [`FaultInjectionRule`] — Combines a condition with a result and additional controls
+//!   like duration, start delay, and hit limit.
```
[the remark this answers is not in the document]
tvaron3: Fixed: updated module docs to reflect actual API — timing windows via `start_time`/`end_time`, `hit_limit`, and `probability`.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/rule.rs
```diff
@@ -0,0 +1,209 @@
...
+        self.enabled.store(false, Ordering::SeqCst);
+    }
+
+    /// Returns the number of times this rule has been applied.
+    pub fn hit_count(&self) -> u32 {
+        self.hit_count.load(Ordering::SeqCst)
+    }
+
+    /// Increments the hit count by one.
+    pub fn increment_hit_count(&self) {
```
analogrelay: Should this be `pub(crate)` so that only fault injection can actually increment it?

sdk/cosmos/azure_data_cosmos_driver/tests/emulator_tests/driver_fault_injection.rs
```diff
@@ -0,0 +1,267 @@
...
+                .await?;
+
+            let item_json = br#"{"id": "item1", "pk": "pk1", "value": "test"}"#;
+            context.create_item(&container, "pk1", item_json).await?;
+
+            // Execute reads to consume the hit limit.
+            // Due to internal retries, the limit may be exhausted within fewer
+            // top-level calls than the limit value.
+            for _ in 0..5 {
+                let _ = context.read_item(&container, "item1", "pk1").await;
```
analogrelay: nit: We could stop the loop when it succeeds, just to make the test slightly faster. But that's absolutely non-blocking.

simorenoh (approved):

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/rule.rs
tvaron3: I will follow up in a different pr with the diagnostic improvements fabian mentioned.

sdk/cosmos/azure_data_cosmos_driver/tests/emulator_tests/driver_fault_injection.rs
tvaron3: I will follow up in a different pr with the diagnostic improvements fabian mentioned.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/rule.rs
tvaron3: Good catch — changed to `pub(crate)`. Only `FaultClient` should increment this internally.

sdk/cosmos/azure_data_cosmos_driver/tests/emulator_tests/driver_fault_injection.rs
tvaron3: Updated — loop now breaks early when a read succeeds (hit limit exhausted).

Issue #3971, opened 2026-03-17
Make streaming upload support more discoverable and easier to use


demoray:
### Feature Summary

The azure blob upload should support SeekableStreams in addition to bytes

### Feature Description

the Upload functionality currently exposed requires having the whole blob in memory.   The backend functionality supports uploading via partitions, using Body::SeekableStream in partitioned_stream::upload.   Can this be exposed such that consumers of the SDK can upload a blob without having to load all of it in memory?

PR #3976, opened 2026-03-17
Add streaming upload example to azure_storage_blob

demoray:
As described in #3971, it would be useful to expose the already existing SeekableStream upload support to consumers of the SDK such that blobs are not required to be entirely in-memory to upload them.

### DIFF
--- a/sdk/storage/azure_storage_blob/examples/blob_upload_seekable_stream.rs
+++ b/sdk/storage/azure_storage_blob/examples/blob_upload_seekable_stream.rs
@@ -84,11 +84,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
+#[derive(Clone)]
 pub struct FileStream {
     handle: Arc<Mutex<File>>,
     pub stream_size: u64,
     buffer_size: usize,
-    read_state: Mutex<ReadState>,
+    read_state: Arc<Mutex<ReadState>>,
 }
 
 impl std::fmt::Debug for FileStream {
@@ -97,17 +98,6 @@ impl std::fmt::Debug for FileStream {
             .field("stream_size", &self.stream_size)
             .field("buffer_size", &self.buffer_size)
             .finish()
-    }
-}
-
-impl Clone for FileStream {
-    fn clone(&self) -> Self {
-        Self {
-            handle: self.handle.clone(),
-            stream_size: self.stream_size,
-            buffer_size: self.buffer_size,
-            read_state: Mutex::new(ReadState::default()),
-        }
     }
 }
 
@@ -121,7 +111,7 @@ impl FileStream {
             handle,
             stream_size,
             buffer_size,
-            read_state: Mutex::new(ReadState::default()),
+            read_state: Arc::new(Mutex::new(ReadState::default())),
         })
     }
 }


sdk/storage/azure_storage_blob/src/clients/block_blob_client.rs
```diff
@@ -131,6 +133,36 @@ impl BlockBlobClient {
         &self,
         content: RequestContent<Bytes, NoFormat>,
         options: Option<BlockBlobClientUploadOptions<'_>>,
+    ) -> Result<BlockBlobClientUploadResult> {
+        self.upload_body(content.into(), options).await
```
heaths: This shouldn't be necessary. `RequestContent` should already support this, so long as the client is properly supporting `SeekableStream`. See the example in `azure_core` I wrote ages ago: https://github.com/Azure/azure-sdk-for-rust/blob/ccbcd78aa94b8b0e370ed5c72a28f30e7e103d5e/sdk/core/typespec_client_core/examples/core_binary_data_request.rs#L24-L31

We could write a similar example - pretty what you wrote here - to make it more discoverable. In fact, I'd put it in the `README` for Storage because it's a champion scenario along with streaming downloads.

heaths (changes requested):

sdk/storage/azure_storage_blob/examples/blob_upload_seekable_stream.rs
```diff
@@ -0,0 +1,308 @@
...
+                    format!("blob name must be valid UTF-8: {:?}", value),
+                )
+            })?,
+            file_path: file_path.into(),
+        })
+    }
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
```
heaths: Structure this like our other examples: put the "main stuff" up top and supporting types below that. I appreciate that - for now, at least - those supporting types are also important, but the `main` function serves as the "how do you actually do this" entry point.

sdk/storage/azure_storage_blob/examples/blob_upload_seekable_stream.rs
```diff
@@ -0,0 +1,308 @@
...
+impl<T, F> From<&FileStream> for RequestContent<T, F> {
+    fn from(stream: &FileStream) -> Self {
+        Body::from(stream).into()
+    }
+}
+
+impl<T, F> From<FileStream> for RequestContent<T, F> {
+    fn from(stream: FileStream) -> Self {
+        Body::from(stream).into()
+    }
```
heaths: I suspect some of these impls aren't needed for this example. I'd cut it down to keep it minimal / less complicated or confusing.

sdk/storage/azure_storage_blob/examples/blob_upload_seekable_stream.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+struct ExampleArgs {
+    account_name: String,
+    container_name: String,
+    blob_name: String,
+    file_path: PathBuf,
+}
+
+impl ExampleArgs {
```
heaths: When re-ordering things to put the important stuff first, seems all this - specific to the example - could go at the bottom. Maybe even past a comment like I did for `sdk/core/azure_core/examples`. The `FileStream` type and related types could all go above that, but I'd still put `main` first after the imports.

sdk/storage/azure_storage_blob/examples/blob_upload_seekable_stream.rs
demoray: I removed all of these, as well as the Builder.  I left SeekableStream and AsyncRead impls for FileStream.

sdk/storage/azure_storage_blob/examples/blob_upload_seekable_stream.rs
demoray: I moved this to the bottom and instead used `clap` which is used in multiple places in the repo for examples.

sdk/storage/azure_storage_blob/examples/blob_upload_seekable_stream.rs
demoray: Done.

sdk/storage/azure_storage_blob/examples/blob_upload_seekable_stream.rs
```diff
@@ -0,0 +1,192 @@
...
+    let blob_client = container_client.blob_client(&args.blob_name);
+    let expected_content = tokio::fs::read(&args.file_path).await?;
+    let stream = FileStream::open(&args.file_path).await?;
+
+    blob_client.upload_stream(stream, None).await?;
+
+    let response = blob_client.download(None).await?;
+    let (_, _, body) = response.deconstruct();
+    let content = body.collect().await?;
+    assert_eq!(content, expected_content);
```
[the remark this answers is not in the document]
demoray: This has been significantly refactored, however as an example, this behavior should be ok.

sdk/storage/azure_storage_blob/src/clients/block_blob_client.rs
demoray: This was reverted per discussion as well as updating the storage blob example to use methods lifted from your FileStreamBuilder example.

sdk/storage/azure_storage_blob/examples/blob_upload_seekable_stream.rs
```diff
@@ -0,0 +1,163 @@
...
+
+use azure_core::{
+    http::{Body, NoFormat, RequestContent},
+    stream::{SeekableStream, DEFAULT_BUFFER_SIZE},
+    Bytes,
+};
+use azure_identity::AzureCliCredential;
+use azure_storage_blob::{models::BlockBlobClientUploadOptions, BlobContainerClient};
+use clap::Parser;
+use futures::{io::AsyncRead, task::Poll, Future};
```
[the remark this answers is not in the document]
demoray: addressed

sdk/storage/azure_storage_blob/examples/blob_upload_seekable_stream.rs
```diff
@@ -0,0 +1,163 @@
...
+        })
+    }
+
+    async fn read(&mut self, slice: &mut [u8]) -> std::io::Result<usize> {
+        let mut handle = self.handle.clone().lock_owned().await;
+        handle.read(slice).await
+    }
+}
+
+#[async_trait::async_trait]
```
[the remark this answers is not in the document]
demoray: addressed by adding a comment per recommendation.

sdk/storage/azure_storage_blob/examples/blob_upload_seekable_stream.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+impl futures::io::AsyncRead for FileStream {
+    fn poll_read(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<std::io::Result<usize>> {
+        std::pin::pin!(self.read(buf)).poll(cx)
+    }
```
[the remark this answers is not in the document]
demoray: This was addressed.

heaths:
Per offline discussion with @demoray, closing in favor of adding a Storage example in #4001 since I refactored a bunch of things.

PR #3987, opened 2026-03-18
[Cosmos] Feed Range API

simorenoh:
Feed Range API

Adds FeedRange type and container-level APIs for working with physical partition ranges in Azure Cosmos DB.

New Public APIs

 - FeedRange — represents a contiguous range of the container's EPK (effective partition key) space. Supports:
  - FeedRange::full() — the complete EPK space
  - contains() / overlaps() — containment and overlap checks between ranges
  - Display / FromStr — base64-encoded JSON serialization (cross-SDK compatible)
  - Serialize / Deserialize — structured JSON for embedding in documents
  - Uses EffectivePartitionKey newtype internally for type-safe EPK comparisons
 - ContainerClient::read_feed_ranges() — returns the physical partition feed ranges for a container. Accepts optional ReadFeedRangesOptions with force_refresh to 
bypass the routing map cache.
 - ContainerClient::feed_range_from_partition_key() — maps a partition key to the FeedRange of the physical partition that contains it. Includes 
retry-on-stale-routing-map logic for resilience during partition splits.
 - ReadFeedRangesOptions — options type with with_force_refresh() for cache bypass.

Internal Changes

 - Added partition_key_definition(), collection_rid(), and resolve_routing_map() accessor methods on ContainerConnection, using the eagerly-resolved 
ContainerReference from the driver.
 - resolve_routing_map() returns Result<CollectionRoutingMap> directly (not Option), encapsulating the error conversion.
 - Added PartialOrd, Ord, Hash, From<String>, and From<&str> impls to EffectivePartitionKey.
 - Exported EPK boundary constants as pub(crate) for reuse.

Validation

 - FromStr and Deserialize validate inclusivity flags ([min, max) semantics) and reject inverted ranges (min > max).
 - from_range() returns Result instead of using debug_assert for release-build safety.

Tests

 - 10 unit tests covering serialization round-trips, cross-SDK compatibility, containment/overlap logic, and validation rejection paths.
 - 2 emulator integration tests: physical partition enumeration (with non-overlap assertions) and partition key → feed range mapping.

This is the foundation for follow-up work plugging feed ranges into query and change feed APIs.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/Cargo.toml
+++ b/sdk/cosmos/azure_data_cosmos/Cargo.toml
@@ -18,7 +18,6 @@ async-lock.workspace = true
 async-trait.workspace = true
 azure_core = { workspace = true, default-features = false }
 azure_data_cosmos_driver = { workspace = true }
-base64.workspace = true
 futures.workspace = true
 pin-project.workspace = true
 reqwest = { workspace = true, optional = true }


sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -344,6 +344,30 @@ pub struct ReadDatabaseOptions;
...
+}
+
+impl ReadFeedRangesOptions {
+    /// Divides the EPK space into the specified number of evenly-sized artificial ranges
+    /// instead of returning physical partition boundaries.
+    ///
+    /// This is useful when you have more workers than physical partitions and want
+    /// finer-grained parallelism. When not set, one [`FeedRange`](crate::FeedRange)
+    /// is returned per physical partition.
+    pub fn with_num_of_ranges(mut self, n: usize) -> Self {
```
FabianMeiswinkel: What is leading to this API? It can be incredibly inefficient RU-charge wise. NOt sure we should expose this as public API - this is an honest question - but understanding requirements/motivation is important.

sdk/cosmos/azure_data_cosmos/src/feed_range.rs
```diff
@@ -0,0 +1,505 @@
...
+///
+/// # Comparison Methods
+///
+/// Feed ranges support containment and overlap checks:
+/// - [`contains()`](FeedRange::contains) — checks if another feed range is entirely within this one
+/// - [`overlaps()`](FeedRange::overlaps) — checks if two feed ranges share any portion of the EPK space
+#[derive(Clone, SafeDebug, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+pub struct FeedRange {
+    pub(crate) min_inclusive: String,
```
FabianMeiswinkel: EPKs should use newType instead of just String?

sdk/cosmos/azure_data_cosmos/src/feed_range.rs
```diff
@@ -0,0 +1,505 @@
...
+    }
+
+    /// Creates a `FeedRange` from an internal `Range<String>`.
+    #[allow(
+        dead_code,
+        reason = "will be used when query/change-feed gain FeedRange support"
+    )]
+    pub(crate) fn from_range(range: &Range<String>) -> Self {
+        Self {
+            min_inclusive: range.min.clone(),
```
FabianMeiswinkel: This is functionally incorrect - you also have to verify/convert is isMinIncluded/isMaxIncluded and the min/maxEPK

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
FabianMeiswinkel: This should probbaly just contain a vector of FeedRanges?

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -54,6 +54,27 @@ impl ContainerConnection {
             .await;
     }
 
+    /// Resolves the container properties from the cache or service.
+    pub(crate) async fn resolve_container_properties(
```
FabianMeiswinkel: why - if you consequently use ContainerReference in the public API you have the PKDef everywhere without cache resolution?

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -54,6 +54,27 @@ impl ContainerConnection {
...
+        container_id: &str,
+    ) -> Result<ContainerProperties, azure_core::Error> {
+        self.container_cache
+            .resolve_by_id(container_id.to_string(), None, false)
+            .await
+    }
+
+    /// Resolves the routing map for the given collection RID.
+    pub(crate) async fn resolve_routing_map(
+        &self,
```
FabianMeiswinkel, line 61: Dito - remove - use ContainerReference instead

sdk/cosmos/azure_data_cosmos/src/feed_range.rs
FabianMeiswinkel: We also need a factory method to produce FeedRange by PartitionKey (even partial) + ContainerReference - could be here or on container.

Also factory method just returning FeedRanges for physical partitions is needed.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
FabianMeiswinkel, line 40: Ouch - this is still using resourceLink based APIs - IMO we should switch the SDK also to use something like ContainerReference in Driver - it is a much cleaner option that doing async cache lookups for ContainerProperties all the time.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -744,4 +745,147 @@ impl ContainerClient {
...
+            .as_deref()
+            .ok_or_else(|| {
+                azure_core::Error::with_message(
+                    azure_core::error::ErrorKind::DataConversion,
+                    "container properties missing '_rid'",
+                )
+            })?;
+
+        // Hash the partition key to its EPK.
+        let pk_version = pk_def.version.unwrap_or(2) as u8;
```
FabianMeiswinkel: using V2 as default is functionally incorrect (i believe) - please double check!

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -744,4 +745,147 @@ impl ContainerClient {
...
+            .await?
+            .ok_or_else(|| {
+                azure_core::Error::with_message(
+                    azure_core::error::ErrorKind::Other,
+                    "failed to resolve routing map for container",
+                )
+            })?;
+
+        let feed_ranges = routing_map
+            .ordered_partition_key_ranges()
```
FabianMeiswinkel, line 823: This will only ever get feed ranges from cache? Do we also need an option to sidestep the cache? I believe default should go to cache - but we probably have to provide an option that forces a refresh (which should also update the cache)

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
simorenoh: I've been going over Copilot's plan.md and it seems it just hallucinated putting this together based on the use-case of feed ranges - it asked me about it and told me other SDKs had this too. Will just remove it.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
simorenoh: This had been added as an option for the FeedRangeOptions

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
```diff
@@ -41,6 +42,54 @@ impl ContainerConnection {
...
+        let collection_name = self.container_ref.name();
+        let collection_rid = self.container_ref.rid();
+        let routing_map = self
+            .pk_range_cache
+            .try_lookup(collection_name, collection_rid, None)
+            .await?;
+        if force_refresh {
+            if let Some(rm) = routing_map {
+                return self
+                    .pk_range_cache
```
tvaron3, line 77: Should we start using the driver pkrange call for this api

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -780,4 +781,96 @@ impl ContainerClient {
...
+    ///     .feed_range_from_partition_key("my_partition_key")
+    ///     .await?;
+    /// println!("Partition key maps to feed range: {}", feed_range);
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[tracing::instrument(skip_all, fields(id = self.container_id))]
+    pub async fn feed_range_from_partition_key(
+        &self,
+        partition_key: impl Into<PartitionKey>,
```
tvaron3, line 868: Should we add the feed range options here as well to refresh the cache? If so maybe rethink the naming of the read feed range options to `FeedRangeOptions`

sdk/cosmos/azure_data_cosmos/src/handler/container_connection.rs
simorenoh: hooked it up to the driver

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
simorenoh: went ahead and added those options

sdk/cosmos/azure_data_cosmos/tests/emulator_tests/cosmos_feed_ranges.rs
```diff
@@ -0,0 +1,99 @@
...
+use framework::TestClient;
+
+#[tokio::test]
+pub async fn read_feed_ranges_returns_physical_partitions() -> Result<(), Box<dyn Error>> {
+    TestClient::run_with_unique_db(
+        async |run_context, db_client| {
+            let properties = ContainerProperties::new("FeedRangeContainer", "/pk".into());
+
+            let container_client = run_context
+                .create_container(db_client, properties, None)
```
tvaron3: Nit: I think this will create a container with one partition. Maybe to make the test a little stricter it would be better to create container with two physical partitions. (don't block on this)

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -980,6 +990,91 @@ impl CosmosDriver {
...
+                let ranges = parse_pk_ranges_response(response.body()).unwrap_or_default();
+                Some(PkRangeFetchResult {
+                    ranges,
+                    continuation: etag_continuation,
+                    not_modified: false,
+                })
+            }
+            Err(e) => {
+                // HTTP 304 Not Modified is expected for incremental change-feed refreshes.
+                if let azure_core::error::ErrorKind::HttpResponse { status, .. } = e.kind() {
```
tvaron3, line 1064: Hmm maybe we should track changing 304s from being an error to successful response in gh issue. Not an issue here as this is internal but thinking of when we expose change feed customers shouldn't think 304s are errors.

sdk/cosmos/azure_data_cosmos/src/feed_range.rs
```diff
@@ -0,0 +1,477 @@
...
+    /// Creates a `FeedRange` from an internal `Range<String>`.
+    ///
+    /// The source range must have `[min, max)` semantics (min inclusive, max exclusive),
+    /// which is the invariant for all partition key ranges from the service.
+    #[allow(
+        dead_code,
+        reason = "will be used when query/change-feed gain FeedRange support"
+    )]
+    pub(crate) fn from_range(range: &Range<String>) -> azure_core::Result<Self> {
+        if !range.is_min_inclusive || range.is_max_inclusive {
```
analogrelay, line 148: We're sure the backend will always return them in this way?

sdk/cosmos/azure_data_cosmos/src/feed_range.rs
```diff
@@ -0,0 +1,477 @@
...
+    #[test]
+    fn display_and_from_str_round_trip() {
+        let range = FeedRange {
+            min_inclusive: EffectivePartitionKey::from(""),
+            max_exclusive: EffectivePartitionKey::from("FF"),
+        };
+        let serialized = range.to_string();
+        let restored: FeedRange = serialized.parse().unwrap();
+        assert_eq!(range, restored);
+    }
```
analogrelay, line 380: The sign! Remember the sign!

Image

sdk/cosmos/azure_data_cosmos/src/feed_range.rs
```diff
@@ -0,0 +1,477 @@
...
+            min_inclusive: EffectivePartitionKey::from("3FFFFFFFFFFF"),
+            max_exclusive: EffectivePartitionKey::from("7FFFFFFFFFFF"),
+        };
+        let serialized = range.to_string();
+        let restored: FeedRange = serialized.parse().unwrap();
+        assert_eq!(range, restored);
+    }
+
+    #[test]
+    fn serde_json_round_trip() {
```
analogrelay: Tap tap ;P

sdk/cosmos/azure_data_cosmos/src/feed_range.rs
simorenoh: 😨

sdk/cosmos/azure_data_cosmos/src/feed_range.rs
simorenoh: such has been the case but went ahead and added a validation

sdk/cosmos/azure_data_cosmos/src/feed_range.rs
simorenoh: 😭 changed this to the agents.md, seems the skill is not invoked properly

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
simorenoh: this was resolved since we now just use the version from the ContainerReference's resolved metadata

sdk/cosmos/azure_data_cosmos/src/feed_range.rs
simorenoh: migrated to just use driver types

sdk/cosmos/azure_data_cosmos/src/feed_range.rs
simorenoh: added this, except for partial since HPK work is currently in progress

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
simorenoh: ContainerReference is now being used everywhere

sdk/cosmos/azure_data_cosmos/src/feed_range.rs
analogrelay: Yeah, that's the thing. Teaching agents is proving to be very difficult :(

sdk/cosmos/azure_data_cosmos/src/feed_range.rs
```diff
@@ -331,59 +352,82 @@ mod tests {
...
     }
 
     #[test]
-    fn display_and_from_str_round_trip() {
+    fn display_produces_expected_base64_full_range() {
         let range = FeedRange {
-            min_inclusive: EffectivePartitionKey::from(""),
-            max_exclusive: EffectivePartitionKey::from("FF"),
+            min_inclusive: String::from(""),
+            max_exclusive: String::from("FF"),
```
analogrelay: Why aren't we using the `EffectivePartitionKey` type here anymore?

sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "azure_data_cosmos"
-version = "0.32.0"
+version = "0.33.0"
```
analogrelay: This shouldn't be in this PR.

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -1,27 +1,47 @@
...
+## 0.33.0 (Unreleased)
+
+### Features Added
+
+### Breaking Changes
+
+### Bugs Fixed
+
+### Other Changes
+
```
analogrelay, line 13: Same here, this shouldn't be in this PR.

simorenoh:
*sigh* copilot was merging, I was merging, and together we made a mess. Making a new branch for this now.

PR #3996, opened 2026-03-19
[Cosmos] Add fault injection rule evaluation diagnostics

tvaron3:
## Summary

Adds per-request visibility into fault injection rule evaluation, aligning with the Java SDK's `FaultInjectionRequestContext` pattern. When fault injection is enabled, every request evaluates all configured rules and records **why** each rule was applied, skipped, or missed.

Addresses feedback from @FabianMeiswinkel on PR #3972.

## Motivation

Previously, when a fault injection rule didn't fire, there was no way to know why — was the condition wrong? Was the hit limit exhausted? Was the timing window off? Users had to guess. The Java SDK solves this with `FaultInjectionRequestContext.recordFaultInjectionRuleEvaluation()`, which records structured evaluation results per transport request.

### `FaultInjectionEvaluation` enum

A new public enum with 9 variants covering every possible evaluation outcome:

| Variant | When |
|---------|------|
| `Applied { rule_id }` | Rule matched and fault was injected |
| `ProbabilityMiss { rule_id, probability }` | Rule matched but random check failed |
| `Disabled { rule_id }` | Rule is disabled via `rule.disable()` |
| `BeforeStartTime { rule_id }` | Current time is before the rule's start window |
| `AfterEndTime { rule_id }` | Current time is at/after the rule's end window |
| `HitLimitExhausted { rule_id, hit_count, hit_limit }` | Rule's hit limit has been reached |
| `OperationMismatch { rule_id }` | Request operation type doesn't match |
| `RegionMismatch { rule_id }` | Request region doesn't match |
| `ContainerMismatch { rule_id }` | Request container ID doesn't match |

### Refactored evaluation flow

`FaultClient` now evaluates **all** rules per request (not just until the first match), collecting structured `FaultInjectionEvaluation` results:

```
execute_request()
  for each rule:
    evaluate_applicability() → Option<FaultInjectionEvaluation>  // None = applicable
    evaluate_condition()     → Option<FaultInjectionEvaluation>  // None = matches
  if matched rule:
    apply_fault() → Applied or ProbabilityMiss
  log all evaluations via tracing::trace!
```

Previously, `is_rule_applicable` and `matches_condition` returned `bool`. Now `evaluate_applicability` and `evaluate_condition` return `Option<FaultInjectionEvaluation>` — `None` means the check passed, `Some(evaluation)` carries the specific skip reason.

### Current scope

- ✅ Collect evaluations for all rules per request
- ✅ Log evaluations via `tracing::trace!`
- ✅ Public `FaultInjectionEvaluation` enum with `rule_id()` and `was_applied()` accessors

## Changes

| File | Change |
|------|--------|
| `fault_injection/mod.rs` | Export `FaultInjectionEvaluation` |
| `fault_injection/http_client.rs` | Add enum, refactor evaluation methods, collect per-request evaluations, add test |

## Testing

30 fault injection tests pass (29 existing + 1 new `evaluation_records_disabled_rule`). Clippy clean with `-D warnings`.


sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/evaluation.rs
```diff
@@ -0,0 +1,147 @@
...
+        rule_id: String,
+    },
+}
+
+// Manual `Eq` because `ProbabilityMiss` contains `f32` which doesn't
+// implement `Eq`. This is safe because the enum is a diagnostic value
+// not used for hashing or ordering — the `f32` probability is always
+// a finite value in [0.0, 1.0] (validated by the builder).
+impl Eq for FaultInjectionEvaluation {}
+
```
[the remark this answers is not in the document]
tvaron3: As mentioned in the comment above this is validated by the builder so should not be an issue.

sdk/cosmos/azure_data_cosmos_driver/src/diagnostics/diagnostics_context.rs
```diff
@@ -444,6 +444,15 @@ pub struct RequestDiagnostics {
...
     /// Error message if the request failed.
     error: Option<String>,
+
+    /// Fault injection rule evaluations for this request.
+    ///
+    /// Populated only when the `fault_injection` feature is enabled and
+    /// evaluations are propagated from the [`FaultClient`](crate::fault_injection::FaultClient)
+    /// via a request-scoped concurrent map keyed by evaluation request ID.
+    #[cfg(feature = "fault_injection")]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
```
FabianMeiswinkel: I think we said we do NOT want to skip serialization in diagnostics and leave empty/null es explicit signals.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/http_client.rs
```diff
@@ -228,20 +264,70 @@ impl FaultClient {
...
         let fault_response = if let Some(ref rule) = matched_rule {
-            self.apply_fault(rule.result(), rule).await
+            let result = self
+                .apply_fault(rule.result(), rule, &mut evaluations)
+                .await;
+            if result.is_none() {
+                evaluations.push(FaultInjectionEvaluation::ProbabilityMiss {
+                    rule_id: rule.id().to_owned(),
+                    probability: rule.result().probability(),
+                });
```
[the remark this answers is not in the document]
tvaron3: Fixed: added `ApplyResult` enum (`Injected`/`ProbabilityMiss`/`NoEffect`) so delay-only and empty results are distinguished from actual probability misses.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/http_client.rs
```diff
@@ -770,17 +855,154 @@ mod tests {
...
+            .build();
+        let rule = Arc::new(FaultInjectionRuleBuilder::new("disabled-rule", error).build());
+        rule.disable();
+
+        let fault_client = FaultClient::new(mock_client, vec![rule]);
+        let (request, _eval_id) = create_test_request();
+        let result = fault_client.execute_request(&request).await;
+        // The evaluation is logged via tracing - verified by the trace output
+        // For now, just verify the request still succeeds (rule is disabled)
+        assert!(result.is_ok());
```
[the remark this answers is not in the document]
tvaron3: Fixed: test now calls `take_evaluations(eval_id)` and asserts `Disabled { rule_id: "disabled-rule" }`.

sdk/cosmos/azure_data_cosmos_driver/src/diagnostics/diagnostics_context.rs
tvaron3: Fixed: removed `skip_serializing_if` — empty array is now always serialized.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/mod.rs
```diff
@@ -26,24 +26,67 @@
...
+/// Storage for fault injection evaluations keyed by request ID.
+///
+/// Written by [`FaultClient`] during `execute_request()`, read by the transport
+/// pipeline in `finalize_http_attempt()`.
+///
+/// **Contract**: Every `store_evaluations()` call must be paired with exactly one
+/// `take_evaluations()` call using the same request ID. Failure to call
+/// `take_evaluations()` (e.g., due to a panic) will leak the entry.
+static EVALUATION_STORE: std::sync::LazyLock<Mutex<HashMap<u64, Vec<FaultInjectionEvaluation>>>> =
+    std::sync::LazyLock::new(|| Mutex::new(HashMap::new()));
```
analogrelay, line 67: This is pretty clunky. Is there not a way we could encode this into the operation metadata itself? Can the fault injection handling get access to the diagnostic_context directly?

Since we're taking control over the pipeline, we can flow whatever we need down to the HTTP Client. Given that, I'd be OK with keeping this here as long as we file a tracking issue to replace this as soon as we have our own HTTP pipeline and can pass our own context object that has access to the necessary diagnostics stuff.

tvaron3: Agreed — filed #4022 to track replacing this with pipeline-level context when the driver has its own HTTP pipeline with a custom context object. Added a TODO comment referencing the issue.

PR #4001, opened 2026-03-20
Refactor HTTP client request handling

### LINKED ISSUE #3971: Make streaming upload support more discoverable and easier to use

heaths:
Add `FileStream` API for easier streaming of request bodies e.g., from a file including a `FileReader` for `tokio::fs::File`. Makes some methods like `SeekableStream::len()` async since underlying APIs for file access are async.

Fixes #3971

### DIFF
--- a/sdk/core/typespec_client_core/src/stream/bytes_stream.rs
+++ b/sdk/core/typespec_client_core/src/stream/bytes_stream.rs
@@ -3,11 +3,8 @@
 
 use super::{Bytes, SeekableStream};
 use crate::http::Body;
-use futures::{
-    io::{AsyncRead, AsyncSeek},
-    stream::Stream,
-};
-use std::{fmt, io, pin::Pin, task::Poll};
+use futures::{io::AsyncRead, stream::Stream};
+use std::{fmt, pin::Pin, task::Poll};
 
 /// Convenience struct that maps a `bytes::Bytes` buffer into a stream.
 ///
@@ -86,8 +83,8 @@ impl SeekableStream for BytesStream {
         Ok(())
     }
 
-    async fn len(&self) -> usize {
-        self.bytes.len()
+    async fn len(&self) -> crate::Result<usize> {
+        Ok(self.bytes.len())
     }
 }
 
@@ -122,37 +119,11 @@ impl AsyncRead for BytesStream {
     }
 }
 
-impl AsyncSeek for BytesStream {
-    fn poll_seek(
-        self: Pin<&mut Self>,
-        _cx: &mut std::task::Context<'_>,
-        pos: io::SeekFrom,
-    ) -> Poll<io::Result<u64>> {
-        let self_mut = self.get_mut();
-        let len = self_mut.bytes.len() as i64;
-        let new_pos = match pos {
-            io::SeekFrom::Start(offset) => offset as i64,
-            io::SeekFrom::End(offset) => len + offset,
-            io::SeekFrom::Current(offset) => self_mut.bytes_read as i64 + offset,
-        };
-
-        if new_pos < 0 {
-            return Poll::Ready(Err(io::Error::new(
-                io::ErrorKind::InvalidInput,
-                "seek to a negative position",
-            )));
-        }
-
-        self_mut.bytes_read = std::cmp::min(new_pos as usize, self_mut.bytes.len());
-        Poll::Ready(Ok(self_mut.bytes_read as u64))
-    }
-}
-
 // Unit tests
 #[cfg(test)]
 mod tests {
     use super::*;
-    use futures::io::{AsyncReadExt, AsyncSeekExt};
+    use futures::io::AsyncReadExt;
     use futures::stream::StreamExt;
 
     // Test BytesStream Stream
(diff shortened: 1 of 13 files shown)


sdk/storage/azure_storage_blob/src/clients/block_blob_client.rs
```diff
@@ -287,7 +287,7 @@ impl PartitionedUploadBehavior for BlockBlobClientUploadBehavior<'_, '_> {
 
     async fn transfer_partition(&self, offset: usize, content: Body) -> Result<()> {
         let block_id = Uuid::new_v4();
-        let content_len = content.len().try_into().unwrap();
+        let content_len = content.len().await.try_into().unwrap();
```
jaschrep-msft: While you're here, can I get you to change this into `content.len().await as u64`, as it is in `transfer_oneshot()`? It's functionally equivalent, but I saw the unwrap and did a double-take.

.github/lsp.json
LarryOsterman, line 1: Is this change still necessary?

sdk/core/azure_core/src/http/pager.rs
```diff
@@ -380,7 +380,7 @@ where
     ///     // The callback must be 'static, so you have to clone and move any values you want to use.
     ///     let pipeline = pipeline.clone();
     ///     let api_version = api_version.clone();
-    ///     let mut req = base_req.clone();
+    ///     let mut req = base_req.try_clone().unwrap();
```
LarryOsterman: Might consider changing this from `.unwrap()` to `.expect("Pageable bodies should never be SeekableStreams.")` or something similar.

sdk/core/typespec_client_core/src/stream/tokio.rs
```diff
@@ -0,0 +1,140 @@
...
+#[async_trait::async_trait]
+impl SeekableStream for FileStream {
+    async fn reset(&mut self) -> crate::Result<()> {
+        self.seek(io::SeekFrom::Start(0)).await?;
+        Ok(())
+    }
+
+    async fn len(&self) -> usize {
+        // We can't seek on &self, so use the file metadata instead.
+        self.file
```
LarryOsterman: It might be helpful to mention that this is unreliable if the file is writable, because the metadata on the file is only updated when the file is closed.

For read-only files, obviously this isn't a problem.

.github/lsp.json
heaths: Once I rebase it'll go away.

sdk/core/typespec_client_core/src/http/clients/reqwest.rs
```diff
@@ -36,16 +36,15 @@ pub fn new_reqwest_client() -> Arc<dyn HttpClient> {
...
         let url = request.url().clone();
         let method = request.method();
         let mut req = self.request(from_method(method), url.clone());
         for (name, value) in request.headers().iter() {
             req = req.header(name.as_str(), value.as_str());
         }
-        let body = request.body().clone();
 
-        let reqwest_request = match body {
+        let reqwest_request = match request.body_mut().take() {
```
heaths: This will break retries because we're emptying the body from the original `Request`.

sdk/core/typespec_client_core/src/http/policies/retry/mod.rs
```diff
@@ -258,7 +258,8 @@ mod test {
         RetryOptions, Url,
     };
     use ::time::macros::datetime;
-    use std::sync::{Arc, Mutex};
+    use std::sync::Arc;
+    use tokio::sync::Mutex;
```
LarryOsterman, line 262: trigger on the `tokio` feature?

LarryOsterman: or use async_lock::Mutex?

sdk/core/typespec_client_core/src/stream/tokio.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    /// Get the length of the file.
+    ///
+    /// # Notes
+    ///
+    /// This may be inaccurate if the file is writable since it may be updated after getting the length.
+    /// This is best used on files opened read-only.
+    async fn len(&self) -> usize {
+        let file = self.file.lock().await;
+        file.metadata().await.map(|m| m.len() as usize).unwrap_or(0)
```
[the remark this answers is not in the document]
heaths: @jaschrep-msft any concerns if I make `len()` and `is_empty()` return a `Result` instead? That it can return an `Err` should be extremely rare, but I'd feel much better than just hiding the fact we erred.

sdk/storage/azure_storage_blob/src/streams/partitioned_stream.rs
```diff
@@ -13,17 +13,20 @@ pub(crate) struct PartitionedStream {
...
+        let total_len = inner.len().await;
         Self {
-            buf: BytesMut::with_capacity(std::cmp::min(partition_len, inner.len())),
+            buf: BytesMut::with_capacity(std::cmp::min(partition_len, total_len)),
             inner,
             partition_len,
+            total_len,
             total_read: 0,
             inner_complete: false,
         }
```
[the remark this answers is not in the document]
LarryOsterman: Unfortunately I think this one's accurate :(. If `inner.len()` ever returns 0, this is going to do bad things

sdk/storage/azure_storage_blob/src/partitioned_transfer/upload.rs
```diff
@@ -26,12 +26,12 @@ pub(crate) async fn upload(
...
     client: &impl PartitionedUploadBehavior,
 ) -> AzureResult<()> {
-    if content.len() <= partition_size.get() {
+    if content.len().await <= partition_size.get() {
         client.transfer_oneshot(content).await?;
         return Ok(());
     }
 
-    client.initialize(content.len()).await?;
+    client.initialize(content.len().await).await?;
```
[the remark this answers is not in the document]
LarryOsterman: IMHO, a reasonable comment.

sdk/core/typespec_client_core/src/stream/tokio.rs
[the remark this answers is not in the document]
heaths: Talking this over with @LarryOsterman w.r.t. some of the comments above, I think going back to a sync `len()` and `is_empty()` is better and have construction get the length asynchronously, which could fail. Do that in one place vs. all over, which really changes the API this late in the game.

sdk/storage/azure_storage_blob/src/streams/partitioned_stream.rs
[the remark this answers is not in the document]
jaschrep-msft: Agreed with Larry. Although I would argue from the storage perspective that we shouldn't be supporting streaming files as they are actively edited, the fact that it can just return 0 on error suggests we should guard against this.

sdk/core/typespec_client_core/src/stream/tokio.rs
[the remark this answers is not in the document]
jaschrep-msft: I think the sync, cached `len()` would be far more useable, yes. The moment we introduce a cache though, I tend to think we need an async `refresh()` of some sort to be available in case it can help recover from an error or otherwise unexpected state.

sdk/storage/azure_storage_blob/examples/blob_storage_upload_file.rs
```diff
@@ -0,0 +1,119 @@
...
+        &endpoint,
+        &args.container_name,
+        &args.blob_name,
+        Some(credential),
+        None,
+    )?;
+
+    let file_size = std::fs::metadata(&args.file_path)?.len();
+
+    let content: Body = if file_size <= DEFAULT_BUFFER_SIZE as u64 {
```
demoray, line 54: 👍

sdk/core/typespec_client_core/src/stream/tokio.rs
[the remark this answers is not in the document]
heaths: Unfortunately, we can't prevent a file from being modified. Someone could open a file with a shared handle and the `File` we get could effectively be modified; however, that's true with any app. We should mitigate it as best we could.

Per internal discussions, I'm going to start over. I'll resolve these comments, but will open a new PR (I think; might just revert and update this PR to show history of decisions made) that basically starts over, keeps `len()` sync and essentially infallible, and go with the approach we've been talking about internally.

sdk/core/typespec_client_core/src/stream/bytes_stream.rs
```diff
@@ -115,11 +122,37 @@ impl AsyncRead for BytesStream {
...
+        if new_pos < 0 {
+            return Poll::Ready(Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "seek to a negative position",
+            )));
+        }
+
+        self_mut.bytes_read = std::cmp::min(new_pos as usize, self_mut.bytes.len());
+        Poll::Ready(Ok(self_mut.bytes_read as u64))
+    }
```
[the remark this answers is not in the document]
heaths: Removed. Made irrelevant in a previous commit.

PR #4005, opened 2026-03-20
[Cosmos] Use ContainerReference in Cosmos Client SDK, initial driver dependency for caches

simorenoh:
# Summary

Adds `azure_data_cosmos_driver` as a dependency to the SDK and introduces the `ContainerReference` pattern for eager container metadata resolution. This is the first step toward using the driver as the SDK's internal transport/routing layer.

This PR is *not* the full cutover into the driver under the hood - its main purpose is to start that process with the bare minimum (current caches) without fully replacing the transport pipeline. That will be done in an entirely separate issue/PR. The cutover from the SDK's fault injection into the Driver's fault injection for end-to-end testing against the driver will also be in a separate issue/PR. This work is meant to unblock other work as well, like https://github.com/Azure/azure-sdk-for-rust/pull/3987/.

## Design

`ContainerClient` construction now eagerly resolves immutable container metadata (RID, partition key definition) via the driver's `resolve_container()`, rather than doing per-operation cache lookups in `send()`. This mirrors how the driver's own `ContainerReference` works.

```
CosmosClient::build()
      |
      v
CosmosDriverRuntime → CosmosDriver (per-account singleton)
      |
      v
DatabaseClient::container_client("name").await?
      |
      v
driver.resolve_container(db, name) → ContainerReference
      |
      v
ContainerConnection stores ContainerReference
      → send() uses stored RID + PK def (no per-op cache lookup)
```

### SDK `ContainerReference` (No Model Sharing)

The SDK defines its own `pub(crate) ContainerReference` adapted from the driver's type via `from_driver_ref()`. This follows the versioning strategy in `AGENTS.md` — `azure_data_cosmos` cannot expose `azure_data_cosmos_driver` types directly.

### SDK (`azure_data_cosmos`)

| File | Change |
|------|--------|
| `models/container_reference.rs` | New — `ContainerReference` with `from_driver_ref()`, `from_parts()`, accessors |
| `clients/cosmos_client.rs` | Added `driver: Arc<CosmosDriver>` field, passes to `DatabaseClient` |
| `clients/cosmos_client_builder.rs` | Creates `CosmosDriverRuntime` + `CosmosDriver` in `build()`, commented out 5 builder unit tests (need fault injection linked from SDK to driver) |
| `clients/database_client.rs` | Added `driver` field, `container_client()` now returns `azure_core::Result<ContainerClient>` (breaking) |
| `clients/container_client.rs` | `new()` calls `driver.resolve_container()`, builds `ContainerReference`, returns `Result` |
| `handler/container_connection.rs` | Stores `ContainerReference`, `send()` uses stored metadata, fixed dual-cache-key bug |

### Dependency alignment

| File | Change |
|------|--------|
| Root `Cargo.toml` | Added `azure_data_cosmos_driver` workspace dependency |
| SDK `Cargo.toml` | Added driver dep, `azure_core` → workspace, `reqwest` feature forwards `driver/reqwest_native_tls` |
| Native `Cargo.toml` | `azure_core` → workspace |
| Perf `Cargo.toml` | `azure_core` + `azure_identity` → workspace |

### Call site updates (~59 sites across 19 files)

All `.container_client()` calls updated to `.container_client().await?` across tests, examples, native crate, and perf crate.

### Bug fix

`send()` previously used the container name (e.g., `"MyContainer"`) as the `pk_range_cache` key, while the cache parameter is named `collection_rid` and expects a RID. All lookups now consistently use `ContainerReference::collection_rid()`.

## Architecture notes

- The driver is only used for `resolve_container()` in this PR. The SDK's `GatewayPipeline` still handles all data plane operations. Full transport cutover is planned for a future PR.
- Both the SDK and driver maintain independent HTTP transports — acceptable overhead for this phase.
- Delete+recreate of a same-name container will cause existing `ContainerClient` instances to fail — this will be addressed in a follow up taking care of container re-creation scenarios.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -101,9 +101,9 @@ checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
 name = "arc-swap"
-version = "1.8.2"
+version = "1.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f9f3647c145568cec02c42054e07bdf9a5a698e15b466fb2341bfc393cd24aa5"
+checksum = "a07d1f37ff60921c83bdfc7407723bdefe89b44b98a9b772f225c8f9d67141a6"
 dependencies = [
  "rustversion",
 ]
@@ -349,6 +349,7 @@ dependencies = [
  "async-lock",
  "async-trait",
  "azure_core",
+ "azure_data_cosmos_macros",
  "azure_identity",
  "base64 0.22.1",
  "crossbeam-epoch",
@@ -1123,9 +1124,9 @@ dependencies = [
 
 [[package]]
 name = "der"
-version = "0.7.10"
+version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e7c1832837b905bbfb5101e07cc24c8deddf52f93225eee6ead5f4d63d53ddcb"
+checksum = "71fd89660b2dc699704064e59e9dba0147b903e85319429e131620d022be411b"
 dependencies = [
  "pem-rfc7468",
  "zeroize",
@@ -1869,9 +1870,9 @@ checksum = "d98f6fed1fde3f8c21bc40a1abb88dd75e67924f9cffc3ef95607bad8017f8e2"
 
 [[package]]
 name = "iri-string"
-version = "0.7.10"
+version = "0.7.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c91338f0783edbd6195decb37bae672fd3b165faffb89bf7b9e6942f8b1a731a"
+checksum = "d8e7418f59cc01c88316161279a7f665217ae316b388e58a0d10e29f54f1e5eb"
 dependencies = [
  "memchr",
  "serde",
@@ -1894,9 +1895,9 @@ dependencies = [
 
 [[package]]
 name = "itoa"
-version = "1.0.17"
+version = "1.0.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "92ecc6618181def0457392ccd0ee51198e065e016d1d527a7ac1b6dc7c1f09d2"
+checksum = "8f42a60cbdf9a97f5d2305f08a87dc4e09308d1276d28c869c684d7777685682"
 
 [[package]]
 name = "js-sys"
@@ -2257,9 +2258,9 @@ dependencies = [
 
 [[package]]
 name = "pem-rfc7468"
-version = "0.7.0"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "88b39c9bfcfc231068454382784bb460aae594343fb030d46e9f50a645418412"
+checksum = "a6305423e0e7738146434843d1694d621cce767262b2a86910beab705e4493d9"
 dependencies = [
  "base64ct",
 ]
@@ -2946,9 +2947,9 @@ dependencies = [
 
 [[package]]
 name = "serde_spanned"
-version = "1.0.4"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f8bbf91e5a4d6315eee45e704372590b30e260ee83af6639d64557f51b067776"
+checksum = "876ac351060d4f882bb1032b6369eb0aef79ad9df1ea8bc404874d8cc3d0cd98"
 dependencies = [
  "serde_core",
 ]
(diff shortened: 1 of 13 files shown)


sdk/cosmos/azure_data_cosmos/src/models/container_reference.rs
```diff
@@ -0,0 +1,112 @@
...
+/// Created once when a [`ContainerClient`](crate::clients::ContainerClient) is constructed
+/// by calling [`CosmosDriver::resolve_container()`](azure_data_cosmos_driver::CosmosDriver::resolve_container).
+/// Holds the container's name, resource ID (RID), and partition key definition —
+/// all of which are immutable after container creation.
+///
+/// This is the SDK's own type, adapted from the driver's
+/// [`ContainerReference`](azure_data_cosmos_driver::models::ContainerReference)
+/// to maintain independent versioning between the two crates.
+#[derive(Clone, Debug)]
+pub(crate) struct ContainerReference {
```
analogrelay: I actually _don't_ think we need the SDK to mirror every type from the driver. Re-exporting the driver's types is OK as long as we're happy with the public surface of any given type. Because Rust is a source-dependency language, it's OK if the formal type identity (whether a type is `azure_data_cosmos::ContainerReference` or `azure_data_cosmos_driver::ContainerReference`) changes between minor versions. Moving from a re-exported driver type to a new type in the SDK itself is **not inherently breaking** so long as the new type in the SDK preserves the API surface of the old re-exported type. Given that, I think we should avoid the double types as much as possible. We can aim for making non-breaking changes to driver types, given that we know they're used in the SDK, but when breaking changes are necessary, we have ways to avoid propagating that to SDKs.

The main place type identity could matter is trait impls. If some other crate is implementing a trait on `azure_data_cosmos_driver::ContainerReference`, that wouldn't automatically transfer over to `azure_data_cosmos::ContainerReference` when we migrate to an SDK-specific type. **However**, if that other create implements the trait on `azure_data_cosmos::ContainerReference`, they'll get the migration just fine. Basically, a re-export is "transparent". The underlying identity of the type is **always** the identity of the original type in it's defining crate. The key difference between Rust and a language like .NET here, is that because Rust packages are **always** distributed as source (essentially), a number of migration techniques that could never work in .NET are easy in Rust.

analogrelay: **If** we want to avoid the trait issue, I would prefer we use wrapping newtypes (where internal code can just read the inner driver type but the SDK type defines public accessors that external users are expected to use). But again, I **don't think that's needed here, yet**.

sdk/cosmos/azure_data_cosmos/src/models/container_reference.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// `Option<i32>` version).
+    pub(crate) fn from_driver_ref(
+        driver_ref: &azure_data_cosmos_driver::models::ContainerReference,
+    ) -> Self {
+        let driver_pk = driver_ref.partition_key_definition();
+        let paths: Vec<String> = driver_pk.paths().iter().map(|p| p.to_string()).collect();
+
+        // The driver distinguishes Hash vs Range (legacy). The SDK distinguishes
+        // Hash vs MultiHash (hierarchical PK). MultiHash is determined by path
+        // count, matching how PartitionKeyDefinition::new() works in the SDK.
```
analogrelay: We should fix inconsistencies like this. The Driver is brand new code and the SDK is still pre-1.0. There's no need to live with legacy in either.

simorenoh: Fixed to handle Range in the SDK as well, since that's what we want. Legacy here is referring to containers that were made back when we supported these older partition key types, which may still be in use by users who onboarded that long ago and is likely the reason we have it in the Driver as well - unfortunately I think we have to support it. @FabianMeiswinkel to correct me in case I'm wrong.

analogrelay: Yeah, we absolutely need that kind of legacy, I was mostly talking about naming and internal inconsistencies that might have crept in. Sounds like a plan!

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -305,255 +309,92 @@ impl CosmosClientBuilder {
...
             fault_injection_enabled,
         ));
 
+        // Create the CosmosDriver for eager container metadata resolution.
+        let driver_account = build_driver_account(endpoint, driver_credential);
+        let driver_runtime = CosmosDriverRuntimeBuilder::new().build().await?;
+        let driver = driver_runtime
+            .get_or_create_driver(driver_account, None)
+            .await?;
+
```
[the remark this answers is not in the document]
analogrelay: That's exactly what we want to avoid Copilot!

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
```diff
@@ -305,255 +309,92 @@ impl CosmosClientBuilder {
...
-            strings,
-            vec!["https://test.documents.azure.com/"],
-            "should fall back to default endpoint"
-        );
-    }
+#[cfg(test)]
+mod tests {
+    // TODO: These tests were disabled because CosmosClient::builder().build() now eagerly
+    // creates a CosmosDriver, which requires a real endpoint. Re-add once fault injection
+    // is linked from the SDK to the driver.
```
analogrelay: We should either remove these or restore them, not leave them commented out.

sdk/cosmos/azure_data_cosmos/src/models/container_reference.rs
FabianMeiswinkel: Service still supports V1 Hash even for brand new containers  - so, absolutely needs to be supported. MultiHash - I don't care too much - my preference would be to be explicit - but nothing I would block on as long as it is consistent.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client_builder.rs
simorenoh: will go ahead and remove them - didn't want to lose the context of tests we might be missing once we do the cutover but I can keep track locally

PR #4008, opened 2026-03-20
Propose ReadMany API design spec for driver crate

TheovanKraay:
Proposes a ReadMany API for azure_data_cosmos_driver that groups items by physical partition using EPK hashing, dispatches single-item groups as point reads and multi-item groups as batched parameterized queries, and executes concurrently.

Covers driver, SDK, and C FFI layer APIs with query shape selection, error handling, and test plan.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos_driver/docs/ReadMany.md
+++ b/sdk/cosmos/azure_data_cosmos_driver/docs/ReadMany.md
@@ -84,10 +84,6 @@ pub enum ReadManyInput {
 pub struct ReadManyOptions {
     /// Maximum concurrent sub-operations. Default: `min(num_ranges, 32)`.
     pub max_concurrency: Option<usize>,
-    /// Optional projection — field paths to return (e.g., `["/id", "/name"]`).
-    /// When set, queries use `SELECT c.id, c.name` instead of `SELECT *`.
-    /// Ignored for point reads (point reads always return full documents).
-    pub projections: Option<Vec<String>>,
 }
 ```
 
@@ -179,9 +175,6 @@ read_many(container, input, read_many_options, operation_options)
 
 All queries use parameterized values. Shape selection depends on the mode, PK
 definition, and item distribution within each chunk.
-
-When `projections` is set, `SELECT *` becomes `SELECT c.id, c.name, ...` with the
-requested fields. Point reads always return full documents (projections are ignored).
 
 #### (id, pk) Mode Shapes
 
@@ -383,8 +376,6 @@ pub struct ReadManyOptions {
     pub custom_headers: HashMap<HeaderName, HeaderValue>,
     /// Maximum concurrent sub-operations.
     pub max_concurrency: Option<usize>,
-    /// Optional field-path projections (e.g., `["/id", "/name"]`).
-    pub projections: Option<Vec<String>>,
 }
 
 pub struct ReadManyResponse<T> {
@@ -479,7 +470,7 @@ void cosmos_read_many_response_free(cosmos_read_many_response* response);
 |------|--------|-------------|
 | `src/models/item_identity.rs` | Create | `ItemIdentity`, `ReadManyInput` types |
 | `src/models/read_many_response.rs` | Create | `ReadManyResponse` type |
-| `src/models/read_many_options.rs` | Create | `ReadManyOptions` (max_concurrency, projections) |
+| `src/models/read_many_options.rs` | Create | `ReadManyOptions` (max_concurrency) |
 | `src/models/partition_key_range.rs` | Create | `PartitionKeyRange` boundaries |
 | `src/models/mod.rs` | Modify | Export new types |
 | `src/driver/read_many.rs` | Create | Core algorithm |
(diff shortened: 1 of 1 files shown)


sdk/cosmos/azure_data_cosmos_driver/docs/ReadMany.md
```diff
@@ -0,0 +1,418 @@
...
+}
+
+/// Response from a ReadMany operation.
+#[derive(Debug)]
+pub struct ReadManyResponse {
+    /// Raw JSON bodies. Items not found are silently omitted.
+    /// Order is **unspecified**.
+    items: Vec<Vec<u8>>,
+    request_charge: RequestCharge,
+    diagnostics: Vec<OperationDiagnostics>,
```
FabianMeiswinkel: should be 1:1 cardinality between diagnostics ang response

sdk/cosmos/azure_data_cosmos_driver/docs/ReadMany.md
```diff
@@ -0,0 +1,418 @@
...
+impl CosmosDriver {
+    /// Reads multiple items by (id, partition_key).
+    ///
+    /// Single-item partition groups use point reads (1 RU, lowest latency).
+    /// Multi-item groups use batched parameterized queries. All execute
+    /// concurrently. Items not found are silently omitted.
+    pub async fn read_many(
+        &self,
+        container: &ContainerReference,
+        items: Vec<ItemIdentity>,
```
FabianMeiswinkel: I would like to design this API to scale to item identities as well as list of just (partial) partition keys - this is a requirement we will need to meet very soon for LSEG in Java as well - not super urgent - but the design should include that as a required scenario and ideallly allow this without requiring yet another API

Also there is an ask on Java right now to allow apps to control concurrency - seems reasonable ask - but can be addressed via OperationOptions I guess

sdk/cosmos/azure_data_cosmos_driver/docs/ReadMany.md
```diff
@@ -0,0 +1,418 @@
...
+  ├─ 2. Get container metadata (PK definition, container RID)
+  │
+  ├─ 3. Fetch partition key ranges
+  │     └─ GET /dbs/{db}/colls/{coll}/pkranges
+  │
+  ├─ 4. Group items by physical partition range
+  │     ├─ Compute EPK hash per item (MurmurHash3 V1/V2)
+  │     ├─ Binary search sorted ranges → owning range
+  │     └─ HashMap<RangeId, Vec<ItemIdentity>>
+  │
```
FabianMeiswinkel, line 152: 410 / Partition split/merge retry policy handling should be described here

sdk/cosmos/azure_data_cosmos_driver/docs/ReadMany.md
```diff
@@ -0,0 +1,418 @@
...
+  │     └─ Query: one of three shapes (see below)
+  │
+  ├─ 7. Execute all work units concurrently
+  │     ├─ Paginate queries until continuation exhausted
+  │     └─ On error (except 404): cancel remaining, propagate
+  │
+  └─ 8. Aggregate: concatenate items, sum RU, merge diagnostics
+```
+
+### Query Shapes
```
FabianMeiswinkel, line 174: should we allow projections? That could be really efficient - but not sure how common it is required?

sdk/cosmos/azure_data_cosmos_driver/docs/ReadMany.md
```diff
@@ -0,0 +1,558 @@
...
+    /// Reads multiple items identified by (id, pk) pairs, or all items
+    /// matching a set of partition key values.
+    ///
+    /// In (id,pk) mode, single-item partition groups use point reads
+    /// (1 RU, lowest latency). Multi-item groups use batched parameterized
+    /// queries. In PK-only mode, all groups use queries.
+    ///
+    /// All work units execute concurrently. Items not found are silently
+    /// omitted.
+    pub async fn read_many(
```
analogrelay, line 122: Here's a thought. What if this is `execute_query_operation` instead, and we define a new `CosmosQueryOperation` type, much like `CosmosOperation`. This type would be used to represent:

* A Query (single or multi-partition).
* A ReadMany operation using item identities.
* Any other future composite operations we need to add in the future.

For _most_ composite operations, there is a similar workflow: Using the user's input, internal logic, and the assistance of the backend, compute a query plan (issue queries `[Q1, Q2, Q3, ...]` to PK ranges `[P1, P2, P3, ...]` and merge with algorithm `A`).

This was the observation that led us to implementing ReadMany in the Rust Query Engine initially. A ReadMany is effectively a client-driven query plan. We don't need the backend's support to generate it, but once we've generated it, we basically have the same kind of query plan we'd get from the backend. If we create our own internal `QueryPlan` that can be generated from the backend response, a future library (like ServiceInterop), or synthesized directly in the driver, then we can effectively have a single query pipeline. I think the pipeline node model we've used in other SDKs could really be adapted here.

That also opens the door to more advanced composite operations in the future, like ReadMany w/Projection.

analogrelay: cc @FabianMeiswinkel to get his attention on this comment.

FabianMeiswinkel: @analogrelay - Yeah - I would not call it execute_query_operation - but execute_feed_operation - but that is irrelevant. I think it would make sense to have one entry point for all the (possibly) cross partition oepration that require continuations - ReadFeed, ChangeFeed, ReadMany, ReadAll (PK-specific ReadFeed) and Query

TheovanKraay:
Close this in favour of new new pipeline/planning API pattern

PR #4010, opened 2026-03-23
Add live test matrix configs to identity

danieljurek:
A change to live testing matrix config for Cosmos resulted in a regression to identity's weekly pipeline where the default live test matrix was used instead of the intended matrix. This happened when we split `MatrixConfigs` into `MatrixConfigs` and `LiveTestMatrixConfigs`. 

This PR fixes identity's configuration by migrating the matrix configuration `LiveTestMatrixConfigs` and leaving the default `MatrixConfigs` for non-live testing (as discussed with @chlowell)

Produces expected build matrix: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6049029&view=results

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2642,9 +2642,9 @@ dependencies = [
 
 [[package]]
 name = "rustls-webpki"
-version = "0.103.9"
+version = "0.103.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d7df23109aa6c1567d1c575b9952556388da57401e4ace1d15f79eedad0d8f53"
+checksum = "df33b2b81ac578cabaf06b89b0631153a3f416b0a886e8a7a1707fb51abbd1ef"
 dependencies = [
  "ring",
  "rustls-pki-types",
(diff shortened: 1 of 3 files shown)


sdk/identity/ci.yml
```diff
@@ -28,7 +28,7 @@ extends:
       Location: uksouth
       RunLiveTests: true
       PersistOidcToken: true
-      MatrixConfigs:
+      LiveTestMatrixConfigs:
         - Name: managed_identity_matrix
           GenerateVMJobs: true
           Path: sdk/identity/azure_identity/managed-identity-matrix.json
```
RickWinter, line 34: I'm not opposed to having tow of these.

Can you give background for why this isn't an entry in the config "LIVE = true" or something like that.

danieljurek: Cosmos wanted to keep default test coverage in the weekly test pipeline for the unit tests and have special coverage for live testing. Charles and I spoke and a similar setup is useful for identity.

chlowell (approved):

PR #4053, opened 2026-03-26
[Cosmos] Port `read-item` to use Driver

simorenoh:
This PR will contain the work needed in order to make the changes to port over an initial method (`read_item`) to use the underlying driver as the connection. For now, I am sharing a spec of the proposed changes, in the hopes that this same spec will work to migrate all other remaining methods after we verify this one works.

The spec file can be found in the PR to facilitate review, but will also be the description below. Actual code implementation to follow.

# SDK-to-Driver Cutover: Design Specification

## Overview

This document describes the design for routing `azure_data_cosmos` SDK operations through the `azure_data_cosmos_driver` execution engine, replacing the legacy gateway pipeline path. The first operation cut over is `ContainerClient::read_item`, which serves as the **reference pattern** for all subsequent operations.

### Context

Prior to this work, the Cosmos SDK had two separate execution paths:

- **Gateway pipeline** (`azure_data_cosmos`): The SDK handled auth, routing, retries, and request construction via `CosmosRequest` → `GatewayPipeline` → HTTP.
- **Driver** (`azure_data_cosmos_driver`): A newer execution engine with its own transport, routing, and operation model (`CosmosOperation` + `OperationOptions`). Previously used only in driver-level tests.

[PR #4005](https://github.com/Azure/azure-sdk-for-rust/pull/4005) bridged the two worlds by having `ContainerClient::new()` call `driver.resolve_container()` for eager metadata resolution. This PR takes the next step: routing the first data operation through the driver.

### Goal

Make the SDK client a **thin wrapper** over the driver. The SDK translates public-facing types into driver concepts, delegates execution, and translates the response back. All real work (auth, routing, retries, transport) happens inside `driver.execute_operation()`.

### Data Flow

```text
User calls:     container_client.read_item(pk, id, options)
                              │
                    ┌─────────▼────────────┐
                    │  SDK ContainerClient │
                    └─────────┬────────────┘
                              │
          ┌───────────────────┼───────────────────┐
          │                   │                   │
  PartitionKey           ItemOptions        ContainerRef
  (SDK type)             (SDK type)        (driver type,
          │                   │           stored on client)
          │                   │                   │
          ▼                   ▼                   ▼
  into_driver_pk()   item_options_to_       ItemReference::
          │           operation_options()    from_name()
          │                   │                   │
          └───────────────────┼───────────────────┘
                              │
                    ┌─────────▼──────────┐
                    │  CosmosOperation:: │
                    │    read_item()     │
                    └─────────┬──────────┘
                              │
                    ┌─────────▼───────────┐
                    │  driver.execute_    │
                    │  operation(op, opts)│
                    │                     │
                    │  (auth, routing,    │
                    │   retries, HTTP)    │
                    └─────────┬───────────┘
                              │
                    ┌─────────▼───────────┐
                    │  driver_response_   │
                    │  to_cosmos_response │
                    └─────────┬───────────┘
                              │
                    ┌─────────▼───────────┐
                    │  CosmosResponse<T>  │
                    │  (SDK public type)  │
                    └─────────────────────┘
```

### Key Principle

The SDK's public API does not change. `read_item` retains the same signature, return type, and observable behavior. This is a pure internal refactor.

## Design Decision: Driver as Required Infrastructure

An alternative approach was explored where the driver is **optional** — stored as `Option<Arc<CosmosDriver>>` on `CosmosClient`, `DatabaseClient`, and `ContainerClient`. In that model, each operation checks at runtime whether a driver is available: if so, it takes the driver path; otherwise, it falls back to the legacy gateway pipeline. Container metadata resolution is also optional and failure is silently ignored.

We chose **not** to take that approach, since we want to verify the behavior of the driver being used only and this single method will serve as the test. In this design, the driver is **required**:

- `CosmosClient` stores `Arc<CosmosDriver>` (not `Option`).
- `ContainerClient::new()` eagerly resolves container metadata via the driver and returns `Result` — if resolution fails, the client cannot be created.
- Operations have a **single codepath** through the driver, with no gateway fallback.

### Rationale

The purpose of this cutover is to validate that the driver can fully replace the gateway pipeline for each operation. A fallback path undermines that goal:

- **Testability:** If the driver path can silently fall back to the gateway, we can't be 100% sure that the driver path is exercised in production or tests. Failures would be hidden rather than surfaced.
- **Correctness:** A dual-codepath design requires maintaining behavioral parity between two implementations indefinitely. A single path is easier to reason about, test, and debug.
- **Options fidelity:** A fallback path tempts skipping the options translation (e.g., passing empty `OperationOptions` on the driver path), which silently drops user-configured session tokens, etags, and excluded regions.
- **Response fidelity:** A minimal fallback implementation may skip reconstructing response headers from the driver's typed response, causing callers to get `None` for `request_charge()`, `session_token()`, and `etag()`.

The cutover is intentionally incremental — one operation at a time. Operations that haven't been cut over yet continue using the gateway pipeline naturally (they don't call the driver). This gives us the gradual rollout benefit without the complexity of runtime branching within a single operation.

### PartitionKey (SDK → Driver)

The SDK and driver define **separate `PartitionKey` types** with identical structure but in different crates. Both represent a JSON array of typed values (string, number, bool, null).

**Approach:** Added `into_driver_partition_key()` on the SDK's `PartitionKey` that maps each `InnerPartitionKeyValue` variant to the driver's `PartitionKeyValue`.

**Driver change required:** Made `PartitionKeyValue` `pub` (was `pub(crate)`) so the SDK crate can construct `Vec<PartitionKeyValue>` for the conversion.

**Future consideration:** Once Ashley's options alignment work unifies these types, this conversion can be eliminated, and we can just use the Driver's definitions the way we did with the ContainerReference.

```rust
// SDK partition_key.rs
pub(crate) fn into_driver_partition_key(self) -> driver::PartitionKey {
    let driver_values: Vec<DriverPKV> = self.0.into_iter()
        .map(|v| match v.0 {
            InnerPartitionKeyValue::String(s) => DriverPKV::from(s),
            InnerPartitionKeyValue::Number(n) => DriverPKV::from(n),
            InnerPartitionKeyValue::Bool(b) => DriverPKV::from(b),
            InnerPartitionKeyValue::Null => DriverPKV::from(Option::<String>::None),
            // ...
        })
        .collect();
    DriverPK::from(driver_values)
}
```

### ItemOptions → OperationOptions

The SDK's `ItemOptions` (item-scoped request options) maps to the driver's `OperationOptions` field-by-field. The types in each field differ between crates, so values are bridged via their string representations.

| SDK `ItemOptions` field | Driver `OperationOptions` | Conversion |
| --- | --- | --- |
| `session_token: Option<SessionToken>` | `.with_session_token()` | `DriverSessionToken::new(token.to_string())` |
| `if_match_etag: Option<Etag>` | `.with_etag_condition()` | `Precondition::if_match(ETag::new(etag.to_string()))` |
| `custom_headers: HashMap<...>` | `.with_custom_headers()` | Passed through directly (types are the same) |
| `excluded_regions: Option<Vec<RegionName>>` | `.with_excluded_regions()` | `Region::new(name.to_string())` for each |
| `content_response_on_write_enabled: bool` | *Ignored for reads* | Driver always returns body for point reads |

**Driver change required:** Added `custom_headers` support to `OperationOptions` (new field, setter, getter) and wired it into `build_transport_request` in `operation_pipeline.rs`. Custom headers may be removed in the future as we analyze which options are truly needed.

### Response Bridge (Driver → SDK)

The driver returns an untyped `CosmosResponse { body: Vec<u8>, headers: CosmosResponseHeaders, status: CosmosStatus }`. The SDK returns a typed `CosmosResponse<T>` wrapping `azure_core::Response<T>`.

**Approach:** Reconstruct the SDK response from driver parts:

```rust
pub(crate) fn driver_response_to_cosmos_response<T>(
    driver_response: DriverResponse,
) -> CosmosResponse<T> {
    let status_code = driver_response.status().status_code();
    let headers = cosmos_response_headers_to_headers(driver_response.headers());
    let body = driver_response.into_body();

    let raw = RawResponse::from_bytes(status_code, headers, Bytes::from(body));
    let typed: Response<T> = raw.into();
    CosmosResponse::new(typed, None)
}
```

The header conversion maps each typed `CosmosResponseHeaders` field back to its raw header name/value pair (reverse of the driver's `from_headers()` parser).

**Caveat:** Only headers that the driver explicitly parses are preserved (activity ID, request charge, session token, etag, continuation, item count, substatus). Any other server headers are lost. This covers all standard Cosmos response metadata. We will probably come back to this when we do the work on verifying the headers we want.

### CosmosRequest → Optional

The SDK's `CosmosResponse<T>` previously held the original `CosmosRequest` — a gateway pipeline concept with no driver equivalent. The driver uses `CosmosOperation` + `OperationOptions` instead, which are consumed during execution.

**Decision:** Made the `request` field `Option<CosmosRequest>`:

- Gateway-routed operations (all methods not yet cut over) continue setting `Some(request)`.
- Driver-routed operations set `None`.
- The field is only accessed behind `#[cfg(feature = "fault_injection")]` and marked `#[allow(dead_code)]`.
- A TODO comment marks it for removal once all operations are on the driver.

### ContainerClient

Added two fields to `ContainerClient` so `read_item` can reach the driver at execution time:

```rust
pub struct ContainerClient {
    // ... existing fields ...
    driver: Arc<CosmosDriver>,         // retained from new()
    container_ref: ContainerReference,  // cloned before passing to ContainerConnection
}
```

Previously, the driver was discarded after `new()` and `ContainerReference` was buried inside `ContainerConnection`.

### driver_bridge Module

New private module at `src/driver_bridge.rs` containing:

- `driver_response_to_cosmos_response<T>()` — response conversion
- `item_options_to_operation_options()` — options translation
- `driver_response_headers_to_headers()` — converts the driver's typed response headers (e.g., `activity_id: Option<ActivityId>`, `request_charge: Option<RequestCharge>`) into raw `azure_core::Headers` key-value pairs for the SDK response

This module is the shared foundation for all future operation cutover. When cutting over `create_item`, `delete_item`, etc., they reuse the same bridge functions.

## Applying This Pattern to Other Operations

To cut over another item operation (e.g., `create_item`), follow this template:

1. **Build the operation:** Use the appropriate `CosmosOperation::*` factory method (e.g., `CosmosOperation::create_item(container_ref, pk)`).
2. **Attach the body:** For write operations, serialize the item to bytes and call `.with_body(bytes)` on the operation.
3. **Translate options:** Reuse `item_options_to_operation_options()` from `driver_bridge.rs`. For write-specific options (e.g., `content_response_on_write_enabled`), extend the bridge function.
4. **Execute:** Call `self.driver.execute_operation(operation, driver_options).await?`.
5. **Bridge response:** Reuse `driver_response_to_cosmos_response(driver_response)`.

The public method signature should not change.

## Files Changed

| File | Change |
| --- | --- |
| `azure_data_cosmos_driver/src/options/operation_options.rs` | Added `custom_headers` field + setter/getter |
| `azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs` | Wired custom headers into request construction |
| `azure_data_cosmos_driver/src/models/partition_key.rs` | Made `PartitionKeyValue` `pub` |
| `azure_data_cosmos_driver/src/models/mod.rs` | Re-exported `PartitionKeyValue` |
| `azure_data_cosmos/src/driver_bridge.rs` | **New** — shared conversion module |
| `azure_data_cosmos/src/clients/container_client.rs` | Added `driver`/`container_ref` fields; rewrote `read_item` |
| `azure_data_cosmos/src/models/cosmos_response.rs` | Made `request` field optional |
| `azure_data_cosmos/src/partition_key.rs` | Added `into_driver_partition_key()` |
| `azure_data_cosmos/src/options/mod.rs` | Added `pub(crate)` accessors for bridge |
| `azure_data_cosmos/src/pipeline/mod.rs` | Updated `CosmosResponse::new` call site |
| `azure_data_cosmos/src/lib.rs` | Registered `mod driver_bridge` |

## Open Items and Future Work

- **Options alignment:** Ashley is working on aligning SDK options with the driver's options model. Once complete, the `ItemOptions` → `OperationOptions` translation may simplify or become unnecessary.
- **PartitionKey unification:** The dual `PartitionKey` types and `into_driver_partition_key()` conversion should be eliminated once the types are unified.
- **`CosmosRequest` removal:** Once all operations are routed through the driver, the `Option<CosmosRequest>` field on `CosmosResponse<T>` can be removed entirely.
- **`custom_headers` review:** The `custom_headers` field on `OperationOptions` was added for feature parity. It may be removed as we analyze which options are truly needed at the driver level.
- **Remaining operations:** `create_item`, `delete_item`, `replace_item`, `upsert_item`, `patch_item`, and query operations should follow the same pattern established here.

## Fault Injection Wiring

When cutting `read_item` over to the driver, the SDK's fault injection tests initially failed because the two execution paths (gateway and driver) have **independent fault injection systems**. This section documents how they were connected.

### Problem

The SDK and driver each have their own fault injection module (`azure_data_cosmos::fault_injection` and `azure_data_cosmos_driver::fault_injection`). They define parallel but separate types (`FaultInjectionRule`, `FaultInjectionCondition`, `FaultInjectionResult`, etc.) with identical variants but different Rust types. Prior to this work, only the gateway pipeline received fault injection rules — the driver was built without them.

### Solution: Rule Translation with Shared State

The bridge module (`driver_bridge.rs`) includes `sdk_fi_rules_to_driver_fi_rules()`, which translates SDK fault injection rules into driver fault injection rules. The translation covers:

- `FaultOperationType` — variant-by-variant match (identical variant names)
- `FaultInjectionErrorType` — variant-by-variant match
- `FaultInjectionCondition` — `RegionName` → `Region`, operation type and container ID mapped directly
- `FaultInjectionResult` — `Duration` → `Option<Duration>`, probability copied
- Timing fields — `start_time: Instant` → `Option<Instant>`, `end_time` and `hit_limit` copied

### Shared Mutable State

SDK `FaultInjectionRule` has `enabled: Arc<AtomicBool>` and `hit_count: Arc<AtomicU32>` that tests mutate at runtime (`.disable()`, `.enable()`, `.hit_count()`). The driver's `FaultInjectionRuleBuilder` accepts external `Arc`s via `with_shared_state()`, so both the SDK gateway path and the driver path reference the **same atomic state**. This means:

- Calling `.disable()` on the SDK rule also disables it in the driver
- Hit counts are shared — both paths increment the same counter
- Tests that toggle rules or assert hit counts work correctly across both paths

### Wiring in `CosmosClientBuilder`

In `CosmosClientBuilder::build()`:

1. Before the `FaultInjectionClientBuilder` is consumed for the gateway transport, `rules()` extracts a reference to the SDK rules
2. `sdk_fi_rules_to_driver_fi_rules()` translates them to driver rules with shared state
3. The translated rules are passed to `CosmosDriverRuntimeBuilder::with_fault_injection_rules()`
4. The SDK's `fault_injection` Cargo feature now forwards to the driver's `fault_injection` feature

### Test Patterns for Future Cutover

When cutting over additional operations, **no additional fault injection wiring is needed** — it's handled once at the `CosmosClientBuilder` level. However, tests that assert `request_url()` need to handle `None` for driver-routed operations:

```rust
// Gateway-routed operations return Some(url)
// Driver-routed operations return None
if let Some(url) = response.request_url() {
    assert_eq!(url.host_str().unwrap(), expected_endpoint);
}
```

### `custom_response` Translation

Translation of `CustomResponse` (synthetic HTTP responses) is not yet implemented. None of the current tests use custom responses for `ReadItem` operations. When needed, the bridge function should be extended to translate `CustomResponse` fields (`status_code`, `headers`, `body`).

### Consolidating to Driver Fault Injection After Cutover

The current dual-system architecture (SDK fault injection + driver fault injection + translation bridge) exists only because the cutover is incremental — some operations still go through the gateway while others go through the driver. Once **all** operations are routed through the driver:

1. **Drop `azure_data_cosmos::fault_injection`** — the SDK's HTTP-client-level fault interception module becomes unreachable. Delete the entire `src/fault_injection/` directory.
2. **Re-export driver types** — the SDK re-exports the driver's fault injection types directly:

   ```rust
   #[cfg(feature = "fault_injection")]
   pub use azure_data_cosmos_driver::fault_injection;
   ```

3. **Remove the translation layer** — `sdk_fi_rules_to_driver_fi_rules()` in `driver_bridge.rs` and the `shared_enabled()`/`shared_hit_count()` accessors on the SDK rule are no longer needed.
4. **Simplify `CosmosClientBuilder`** — `with_fault_injection()` accepts `Vec<Arc<driver::FaultInjectionRule>>` directly and passes them to `CosmosDriverRuntimeBuilder::with_fault_injection_rules()`. No translation, no cloning, no intermediary builder.
5. **Update tests** — tests construct driver `FaultInjectionRule` directly (same builders, same API) instead of SDK rules.

At that point the SDK has **no fault injection logic of its own** — it's a pass-through to the driver, matching the overall "SDK as thin wrapper" goal. The driver is the single source of truth for all transport-related concerns including fault injection.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/docs/sdk-to-driver-cutover.md
+++ b/sdk/cosmos/azure_data_cosmos/docs/sdk-to-driver-cutover.md
@@ -274,6 +274,21 @@ The gateway pipeline tracked this via `CosmosRequest` (which held the final URL)
 2. Providing users with routing transparency for observability
 3. Removing the `request_url()` → `Option` workaround once all operations are driver-routed
 
+**Tests with skipped endpoint assertions** (these should be restored once the driver exposes the effective endpoint):
+
+| Test File | Test Name | What it verifies |
+| --- | --- | --- |
+| `cosmos_items.rs` | `assert_response` helper (all item tests) | Endpoint matches expected host |
+| `cosmos_fault_injection.rs` | `fault_injection_429_retry_with_hit_limit` | Endpoint matches hub |
+| `cosmos_multi_write_retry_policies.rs` | `read_cross_region_retry_on_408` | Failover to satellite region |
+| `cosmos_multi_write_retry_policies.rs` | `read_cross_region_retry_on_500` | Failover to satellite region |
+| `cosmos_multi_write_fault_injection.rs` | `fault_injection_read_unaffected_by_create_rule` | Endpoint matches hub |
+| `cosmos_multi_write_fault_injection.rs` | `fault_injection_read_region_retry_503` | Failover to satellite region |
+| `cosmos_multi_write_fault_injection.rs` | `fault_injection_read_session_retry_404_1002` | Failover to hub region |
+| `cosmos_multi_write_fault_injection.rs` | `fault_injection_read_connection_error_failover` | Failover to satellite region |
+| `cosmos_multi_write_fault_injection.rs` | `fault_injection_read_response_timeout_retries_to_satellite` | Failover to satellite region |
+| `cosmos_multi_write_fault_injection.rs` | `fault_injection_connection_error_local_retry_succeeds` | Stays on hub (no failover) |
+
 ## Files Changed
 
 | File | Change |
--- a/sdk/cosmos/azure_data_cosmos/tests/multi_write_tests/cosmos_multi_write_fault_injection.rs
+++ b/sdk/cosmos/azure_data_cosmos/tests/multi_write_tests/cosmos_multi_write_fault_injection.rs
@@ -300,16 +300,14 @@ pub async fn fault_injection_read_region_retry_503() -> Result<(), Box<dyn Error
                 .await;
 
             let response = result.unwrap();
-            let request_url = response
-                .request_url()
-                .map(|u| u.to_string())
-                .unwrap_or_default();
-            println!("Request succeeded via failover, final URL: {}", request_url);
-            // Verify the request went to a different endpoint than the faulted one
-            assert!(
-                request_url.contains(&SATELLITE_REGION.as_str()),
-                "request should have failed over to secondary region"
-            );
+            if let Some(request_url) = response.request_url().map(|u| u.to_string()) {
+                println!("Request succeeded via failover, final URL: {}", request_url);
+                // Verify the request went to a different endpoint than the faulted one
+                assert!(
+                    request_url.contains(&SATELLITE_REGION.as_str()),
+                    "request should have failed over to secondary region"
+                );
+            }
 
             Ok(())
         },
@@ -380,15 +378,13 @@ pub async fn fault_injection_write_region_retry_503() -> Result<(), Box<dyn Erro
             );
 
             let response = result.unwrap();
-            let request_url = response
-                .request_url()
-                .map(|u| u.to_string())
-                .unwrap_or_default();
-            // Verify the request went to a different endpoint than the faulted one
-            assert!(
-                request_url.contains(&SATELLITE_REGION.as_str()),
-                "request should have failed over to secondary region"
-            );
+            if let Some(request_url) = response.request_url().map(|u| u.to_string()) {
+                // Verify the request went to a different endpoint than the faulted one
+                assert!(
+                    request_url.contains(&SATELLITE_REGION.as_str()),
+                    "request should have failed over to secondary region"
+                );
+            }
 
             Ok(())
         },
@@ -472,16 +468,14 @@ pub async fn fault_injection_read_region_retry_404_1002() -> Result<(), Box<dyn 
                 .await;
 
             let response = result.unwrap();
-            let request_url = response
-                .request_url()
-                .map(|u| u.to_string())
-                .unwrap_or_default();
-            println!("Request succeeded via failover, final URL: {}", request_url);
-            // Verify the request was retried on the hub region
-            assert!(
-                request_url.contains(&HUB_REGION.as_str()),
-                "request should have failed over to hub region"
-            );
+            if let Some(request_url) = response.request_url().map(|u| u.to_string()) {
+                println!("Request succeeded via failover, final URL: {}", request_url);
+                // Verify the request was retried on the hub region
+                assert!(
+                    request_url.contains(&HUB_REGION.as_str()),
+                    "request should have failed over to hub region"
+                );
+            }
 
             Ok(())
         },
@@ -549,14 +543,12 @@ pub async fn fault_injection_write_connection_error_failover() -> Result<(), Box
                 .await
                 .expect("write should succeed via failover to satellite");
 
-            let request_url = response
-                .request_url()
-                .map(|u| u.to_string())
-                .unwrap_or_default();
-            assert!(
-                request_url.contains(SATELLITE_REGION.as_str()),
-                "request should have failed over to satellite region, got: {request_url}"
-            );
+            if let Some(request_url) = response.request_url().map(|u| u.to_string()) {
+                assert!(
+                    request_url.contains(SATELLITE_REGION.as_str()),
+                    "request should have failed over to satellite region, got: {request_url}"
+                );
+            }
 
             Ok(())
         },
@@ -633,14 +625,12 @@ pub async fn fault_injection_read_connection_error_failover() -> Result<(), Box<
                 .await
                 .expect("read should succeed via failover to satellite");
 
-            let request_url = response
-                .request_url()
-                .map(|u| u.to_string())
-                .unwrap_or_default();
-            assert!(
-                request_url.contains(SATELLITE_REGION.as_str()),
-                "request should have failed over to satellite region, got: {request_url}"
-            );
+            if let Some(request_url) = response.request_url().map(|u| u.to_string()) {
+                assert!(
+                    request_url.contains(SATELLITE_REGION.as_str()),
+                    "request should have failed over to satellite region, got: {request_url}"
+                );
+            }
 
             Ok(())
         },
@@ -784,14 +774,12 @@ pub async fn fault_injection_read_response_timeout_retries_to_satellite(
                 .await
                 .expect("read should succeed via failover after response timeout on hub");
 
-            let request_url = response
-                .request_url()
-                .map(|u| u.to_string())
-                .unwrap_or_default();
-            assert!(
-                request_url.contains(SATELLITE_REGION.as_str()),
-                "request should have failed over to satellite region, got: {request_url}"
-            );
+            if let Some(request_url) = response.request_url().map(|u| u.to_string()) {
+                assert!(
+                    request_url.contains(SATELLITE_REGION.as_str()),
+                    "request should have failed over to satellite region, got: {request_url}"
+                );
+            }
 
             Ok(())
         },
@@ -858,14 +846,12 @@ pub async fn fault_injection_connection_error_reverse_failover() -> Result<(), B
                 .await
                 .expect("write should succeed via reverse failover to hub");
 
-            let request_url = response
-                .request_url()
-                .map(|u| u.to_string())
-                .unwrap_or_default();
-            assert!(
-                request_url.contains(HUB_REGION.as_str()),
-                "request should have failed over to hub region, got: {request_url}"
-            );
+            if let Some(request_url) = response.request_url().map(|u| u.to_string()) {
+                assert!(
+                    request_url.contains(HUB_REGION.as_str()),
+                    "request should have failed over to hub region, got: {request_url}"
+                );
+            }
 
             Ok(())
         },
@@ -936,15 +922,13 @@ pub async fn fault_injection_connection_error_local_retry_succeeds() -> Result<(
                 .await
                 .expect("read should succeed on hub after transient fault clears");
 
-            let request_url = response
-                .request_url()
-                .map(|u| u.to_string())
-                .unwrap_or_default();
-            // The fault cleared before MAX_RETRY_COUNT, so no failover — still on hub.
-            assert!(
-                request_url.contains(HUB_REGION.as_str()),
-                "request should have succeeded on hub without failover, got: {request_url}"
-            );
+            if let Some(request_url) = response.request_url().map(|u| u.to_string()) {
+                // The fault cleared before MAX_RETRY_COUNT, so no failover — still on hub.
+                assert!(
+                    request_url.contains(HUB_REGION.as_str()),
+                    "request should have succeeded on hub without failover, got: {request_url}"
+                );
+            }
 
             Ok(())
         },


sdk/cosmos/azure_data_cosmos/docs/sdk-to-driver-cutover.md
```diff
@@ -0,0 +1,229 @@
...
+            InnerPartitionKeyValue::Bool(b) => DriverPKV::from(b),
+            InnerPartitionKeyValue::Null => DriverPKV::from(Option::<String>::None),
+            // ...
+        })
+        .collect();
+    DriverPK::from(driver_values)
+}
+```
+
+### ItemOptions → OperationOptions
```
analogrelay, line 120: _Some_ of this is what I'm working on, I'll have a PR out pretty soon. I have two PRs coming:

1. A refactoring of the Driver's Options to align with the spec I posted earlier
2. A restructuring of the SDK to use the driver's options types (even for operations that still use it's own pipeline)

I don't think that will dramatically affect your ability to start off with this though.

sdk/cosmos/azure_data_cosmos/src/partition_key.rs
```diff
@@ -113,6 +113,33 @@ impl PartitionKey {
...
+            .into_iter()
+            .map(|v| match v.0 {
+                InnerPartitionKeyValue::String(s) => DriverPKV::from(s),
+                InnerPartitionKeyValue::Number(n) => DriverPKV::from(n),
+                InnerPartitionKeyValue::Bool(b) => DriverPKV::from(b),
+                InnerPartitionKeyValue::Null => DriverPKV::from(Option::<String>::None),
+                InnerPartitionKeyValue::Undefined | InnerPartitionKeyValue::Infinity => {
+                    // These are edge cases not supported by the driver's PK model.
+                    // Undefined/Infinity PKs are rare in practice.
+                    DriverPKV::from(Option::<String>::None)
```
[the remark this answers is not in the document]
analogrelay: I think we do need an `undefined` value in the driver and shouldn't use `null`. They're not the same. Undefined is represented in headers as `{}` whereas null is `null`. Now, how likely or even supported is it for a user to use these values? Probably not very, but it's fairly easy to add for completeness

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -474,6 +476,14 @@ fn build_transport_request(
         );
     }
 
+    // Add custom headers from operation options (inserted last so they don't
+    // override SDK-set headers above).
+    if let Some(custom) = options.custom_headers_ref() {
+        for (name, value) in custom {
+            headers.insert(name.clone(), value.clone());
```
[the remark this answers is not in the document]
analogrelay: Yep, Copilot is right, I believe

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -533,22 +538,31 @@ impl ContainerClient {
...
-            .build()?;
-        options.apply_headers(&mut cosmos_request.headers);
+        // Create the driver operation.
+        let operation = CosmosOperation::read_item(item_ref);
 
-        self.container_connection
-            .send(cosmos_request, Context::default())
-            .await
+        // Translate SDK options to driver options.
+        let driver_options = crate::driver_bridge::item_options_to_operation_options(&options);
```
analogrelay, line 577: Love that this is in it's own function. That will make it a lot easier to align with the work I'm doing :)

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
```diff
@@ -49,20 +53,24 @@ impl<T> CosmosResponse<T> {
...
     /// Returns the final request URL used to fulfill the operation.
     /// This api is subject to change without a major version bump.
     #[cfg(feature = "fault_injection")]
-    pub fn request_url(&self) -> azure_core::http::Url {
-        self.request.clone().into_raw_request().url().clone()
+    pub fn request_url(&self) -> Option<azure_core::http::Url> {
+        self.request
+            .clone()
+            .map(|r| r.into_raw_request().url().clone())
     }
```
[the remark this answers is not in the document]
analogrelay: [Non-blocking]

Hmm, I know it's hard for the driver to give us a specific url, but we may want to consider returning _something_ here. Anyway, I think making this clean break to return `Option<Url>` makes sense for now. We might even want to remove it entirely and replace it with metadata like the Endpoint, OperationType, ResourceType, ResourceId, etc. that can be used to describe the endpoint used to serve the request but aren't specific URLs.

sdk/cosmos/azure_data_cosmos/src/driver_bridge.rs
```diff
@@ -0,0 +1,120 @@
...
+/// status code, and headers, then wraps it in the SDK's response type.
+pub(crate) fn driver_response_to_cosmos_response<T>(
+    driver_response: DriverResponse,
+) -> CosmosResponse<T> {
+    let status_code: StatusCode = driver_response.status().status_code();
+    let headers = driver_response_headers_to_headers(driver_response.headers());
+    let body = driver_response.into_body();
+
+    let raw_response = RawResponse::from_bytes(status_code, headers, Bytes::from(body));
+    let typed_response: Response<T> = raw_response.into();
```
analogrelay, line 40: We'll stop using `azure_core::http::Response` eventually but this makes sense for now!

sdk/cosmos/azure_data_cosmos/src/driver_bridge.rs
analogrelay, line 1: This is great. We should be able to get rid of it over time, but it gives us a simple way to bridge the gaps while we get everything using the driver.

sdk/cosmos/azure_data_cosmos/src/models/cosmos_response.rs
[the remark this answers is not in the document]
tvaron3: Once the diagnostics are fully passed over we can get the necessary info for validating the tests from that and remove this api. The diagnostics should have the information for all requests and retries.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -112,8 +113,10 @@ pub(crate) async fn execute_operation_pipeline(
             "attempt",
             attempt = attempt,
             context = tracing::field::Empty
-        )
-        .entered();
+        );
+        // Enter the span for sync work but drop the guard before the first
+        // `.await` so the future remains `Send` (`EnteredSpan` is `!Send`).
+        let attempt_guard = attempt_span.enter();
```
analogrelay, line 119: Filed https://github.com/Azure/azure-sdk-for-rust/issues/4065 to review this, because we _should_ be able to handle spans across `.await`s, but I believe we have to exit the span and hand it to the future in a different way (through the [`.instrument`](https://docs.rs/tracing/latest/tracing/trait.Instrument.html#method.instrument) extension method)

PR #4056, opened 2026-03-27
[storage] download_into()

jaschrep-msft:
Implement managed download to write data directly into a provided buffer.

### DIFF
--- a/sdk/storage/azure_storage_blob/src/models/drains.rs
+++ b/sdk/storage/azure_storage_blob/src/models/drains.rs
@@ -85,17 +85,15 @@ pub(crate) struct UnorderedFuturesDrain<F> {
 impl<F: Future + Unpin> UnorderedFuturesDrain<F> {
     // Constructs a new drain.
     pub fn new() -> Self {
-        Self {
-            futures: Vec::new(),
-            ..Default::default()
-        }
+        Self::default()
     }
 
-    // Constructs a new drain. The drain will automatically resize its capacity.
+    // Constructs a new drain with a given starting capacity.
+    // The drain will still automatically resize its capacity as needed.
     pub fn with_capacity(capacity: usize) -> Self {
         Self {
             futures: Vec::with_capacity(capacity),
-            ..Default::default()
+            ..Self::default()
         }
     }
 


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -159,6 +159,59 @@ impl BlobClient {
...
+            version_id: options.version_id,
+            ..Default::default()
+        };
+
+        let client = GeneratedBlobClient {
+            endpoint: self.endpoint.clone(),
+            pipeline: self.pipeline.clone(),
+            version: self.version.clone(),
+            tracer: self.tracer.clone(),
+        };
```
heaths, line 202: The `..` operator can do this without all the boilerplate:

```suggestion
        let get_range_options = BlobClientDownloadOptions {
            range: None,
            ..options
        };

        // You could also just impl `Clone` if the `GeneratedBlobClient` itself doesn't already - I don't recall.
        let client = GeneratedBlobClient {
            endpoint: self.endpoint.clone(),
            pipeline: self.pipeline.clone(),
            version: self.version.clone(),
            tracer: self.tracer.clone(),
        };
```

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -159,6 +159,59 @@ impl BlobClient {
             .await
     }
 
+    /// The managed download operation retrieves the content of an existing blob.
+    ///
+    /// # Arguments
+    ///
+    /// * `options` - Optional parameters for the request.
+    pub async fn managed_download_into(
```
heaths, line 167: I think this name can work. I can't find any specific examples in `std` or a few other popular crates but is similar to `clone_into`.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -156,6 +142,197 @@ where
...
+                stream::iter(stats.remaining_download_ranges).then(|range| {
+                    let client = client.clone();
+                    async move { client.transfer_range(Some(range)).await }
+                }),
+            ));
+
+            let mut total_written = 0;
+            while let Some(response) = all_responses.try_next().await? {
+                let written = response.into_body().collect_into(buffer).await?;
+                buffer = &mut buffer[written..];
```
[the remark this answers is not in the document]
jaschrep-msft: This clearly compiles.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
jaschrep-msft: Re: `..`
No we can't. It doesn't compile. They're different types.

Re: impl `Clone`
It's not already implemented. If I implement `Clone` on the client, will that be public? I assume since it's not already implemented on generated code that we don't want it to be.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
jalauzon-msft: We should call this just `download_into` and drop the `managed` prefix. Fine to do this now into main, we will soon be renaming `managed_download` -> `download` (replacing the existing one). We will also be renaming `BlobClientManagedDownloadOptions` to `BlobClientDownloadOptions`. @vincenttran-msft should be doing this soon.

sdk/storage/azure_storage_blob/src/models/drains.rs
```diff
@@ -73,3 +74,84 @@ impl<T> SequentialBoundedDrain<T> {
...
+}
+
+impl<F: Future + Unpin> UnorderedFuturesDrain<F> {
+    // Constructs a new drain.
+    pub fn new() -> Self {
+        Self {
+            futures: Vec::new(),
+            ..Default::default()
+        }
+    }
```
vincenttran-msft, line 89: nit: 
```rust
pub fn new() -> Self {
    Self::default()
}
```

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -156,6 +142,210 @@ where
...
+            ))
+        } else {
+            Ok(bytes_copied)
+        }
+    }
+    .await;
+
+    // SAFETY: This is where we prevent any tasks from living beyond this method return,
+    // guaranteeing no references to `buffer` exist.
+    // TODO: Implement aborting spawned runtime workers.
```
vincenttran-msft, line 306: 👀

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
jalauzon-msft: Why do we need to clone at all? We are a `GeneratedBlobClient`, could we not just use `self` or is something Rusty that doesn't allow that?

jaschrep-msft: Lifetimes. Sending a reference to another thread already requires unsafe code. This is even worse in streaming `download()` where the returned stream is now tied to the lifetime of the input client reference. An `Arc` eliminates all this using safe code in exchange for needing ownership.

We could have download take `self: Arc<Self>` instead of `&self`, but I don't know enough to comment on the cx of that. Cloning the client and putting our clone in an `Arc` is the least-intrusive option. Endpoint and version are just small strings, pipeline is a list of `Arc`s, idk how cheap a tracer clone is but no one seems to be yelling at me.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
jaschrep-msft: It looks like #4060 just merged yesterday, which is what we need to complete this! Sean and I had discussed merging without that and adding in later, but I can do that now instead.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
jalauzon-msft: Sounds good, I figured it was something like that. This is probably fine as is then.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
jaschrep-msft: hmm this is taking some intensive investigation to validate.

What we have now is safe, it just doesn't fail fast. I feel comfortable checking this in and optimizing after.

vincenttran-msft (approved):
Fine to push fast-fails to a later PR, thanks for this!

PR #4059, opened 2026-03-27
Cosmos: Options alignment Step 2 of 2 - Align SDK options with driver types

analogrelay:
﻿Replaces SDK-local options definitions with types from azure_data_cosmos_driver, so that when individual
operations are ported to the driver pipeline, options can be passed through as-is.

Key changes:

 - Split ItemOptions into ItemReadOptions and ItemWriteOptions, each embedding OperationOptions for cross-cutting
 settings (custom headers, excluded regions, content response, consistency strategy).
 - Rewrote QueryOptions and BatchOptions to also embed OperationOptions.
 - Replaced RegionName with Region, SessionToken with the driver's SessionToken, if_match_etag with Precondition,
 and content_response_on_write_enabled: bool with ContentResponseOnWrite enum.
 - Switched batch per-item options (BatchUpsertOptions, etc.) from if_match/if_none_match: Option<String> to
precondition: Option<Precondition>.
 - Added custom_headers to the driver's OperationOptions, eliminating the last SDK-only temporary field.
 - Removed convenience with_ methods that delegated to OperationOptions fields; users set those via
with_operation_options().

All re-exports use #[doc(inline)] so types appear as native SDK types in documentation and don't redirect to `azure_data_cosmos_driver` documentation. That will help reinforce to users that these types are only public API **when used through the SDK**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -352,6 +352,7 @@ dependencies = [
  "azure_data_cosmos_macros",
  "azure_identity",
  "base64 0.22.1",
+ "bytes",
  "crossbeam-epoch",
  "futures",
  "h2",
--- a/sdk/cosmos/azure_data_cosmos_driver/Cargo.toml
+++ b/sdk/cosmos/azure_data_cosmos_driver/Cargo.toml
@@ -23,6 +23,7 @@ azure_core = { workspace = true, default-features = false, features = [
 ] }
 azure_data_cosmos_macros.workspace = true
 base64.workspace = true
+bytes.workspace = true
 crossbeam-epoch = { workspace = true, features = ["std"] }
 futures.workspace = true
 h2 = { workspace = true, optional = true }
--- a/sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
+++ b/sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
@@ -18,7 +18,6 @@ use crate::{
     },
 };
 use arc_swap::ArcSwap;
-use azure_core::http::Request;
 use futures::future::BoxFuture;
 use std::error::Error as _;
 use std::sync::atomic::{AtomicBool, Ordering};
@@ -28,8 +27,8 @@ use std::time::Duration;
 use super::{
     cache::AccountRegion,
     transport::{
-        cosmos_headers, is_emulator_host, request_signing, uses_dataplane_pipeline,
-        AuthorizationContext, CosmosTransport,
+        cosmos_headers, cosmos_transport_client::HttpRequest, is_emulator_host, request_signing,
+        uses_dataplane_pipeline, AuthorizationContext, CosmosTransport,
     },
     CosmosDriverRuntime,
 };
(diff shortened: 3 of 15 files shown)


sdk/cosmos/azure_data_cosmos_driver/src/options/region.rs
```diff
@@ -414,6 +452,44 @@ const DISPLAY_NAME_MAPPING: &[(&str, &str)] = &[
     ("swedensouth", "Sweden South"),
     ("koreasouth2", "Korea South 2"),
     ("ussecwestcentral", "USSec West Central"),
+    ("austriaeast", "Austria East"),
```
analogrelay, line 457: These were present in the SDK's list but not the driver's

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -128,201 +157,234 @@ impl Display for ConsistencyLevel {
...
 
-    pub fn with_custom_headers(mut self, custom_headers: HashMap<HeaderName, HeaderValue>) -> Self {
-        self.custom_headers = custom_headers;
+impl ItemReadOptions {
+    // Temporary: applies option values as HTTP headers for the SDK pipeline.
+    // Will be removed when read_item uses the internal pipeline directly.
+    pub(crate) fn apply_headers(&self, headers: &mut Headers) {
+        if let Some(custom_headers) = &self.operation.custom_headers {
+            for (name, value) in custom_headers {
+                headers.insert(name.clone(), value.clone());
```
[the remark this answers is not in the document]
tvaron3: would be good to document the behavior chosen

sdk/cosmos/azure_data_cosmos/CHANGELOG.md
```diff
@@ -12,6 +12,16 @@
...
 - `DatabaseClient::container_client()` now returns `azure_core::Result<ContainerClient>`, eagerly resolving container metadata (RID, partition key definition) at construction time. ([#4005](https://github.com/Azure/azure-sdk-for-rust/pull/4005))
 - `PartitionKeyDefinition` fields (`paths`, `kind`, `version`) are now private; use accessor methods `paths()`, `kind()`, and `version()` instead. `PartitionKeyKind` changed from a string newtype to an enum with variants `Hash`, `MultiHash`, and `Range`. `PartitionKeyVersion` is now an enum (`V1`, `V2`) instead of `Option<i32>`. ([#4005](https://github.com/Azure/azure-sdk-for-rust/pull/4005))
+- Replaced `ItemOptions` with `ItemReadOptions` (for `read_item`) and `ItemWriteOptions` (for `create_item`, `replace_item`, `upsert_item`, `delete_item`). Both types embed an `OperationOptions` field for general-purpose settings like custom headers, excluded regions, and content response behavior.
```
tvaron3: Seems very noisy and doesn't include pr link. Although I do think it's important to call out the breaking changes to make it easier for any customer upgrading...

analogrelay: I slimmed it down a bit. I do want to be fairly complete here, but I think we could do some compacting.

PR #4060, opened 2026-03-27
Implemented `abort` function on spawned tasks

LarryOsterman:
Updates the `async_runtime` trait to add the ability to abort running tasks.

This changes the return type of `spawn` from a future to the implementation of a trait which exposes a synchronous `abort` function that can be used to abandon execution of the future.

For the `tokio` runtime, the `abort` function simply forwards to the tokio `JoinHandle`, for the `std` runtime, this updates the standard runtime join state to mark the future as finished and drop the standard runtime join handle.

### DIFF
--- a/sdk/core/azure_core/CHANGELOG.md
+++ b/sdk/core/azure_core/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### Breaking Changes
 
 - Added `tokio` feature to `default` features.
-- Changed the `SpawnedTask` return value of `async_runtime::spawn` from a raw future to a trait which enables an `abort()` function on the returned `SpawnedTask`
+- Changed `async_runtime::spawn`/`SpawnedTask` to return a trait object that supports `abort()`.
 
 ### Bugs Fixed
 
--- a/sdk/core/typespec_client_core/src/async_runtime/tokio_runtime.rs
+++ b/sdk/core/typespec_client_core/src/async_runtime/tokio_runtime.rs
@@ -50,9 +50,13 @@ impl std::future::Future for JoinHandle {
         let mut this = self.project();
         if let Some(handle) = this.handle.as_mut().as_pin_mut() {
             match handle.poll(cx) {
-                Poll::Ready(_) => {
+                Poll::Ready(Ok(())) => {
                     this.handle.set(None);
                     Poll::Ready(Ok(()))
+                }
+                Poll::Ready(Err(join_err)) => {
+                    this.handle.set(None);
+                    Poll::Ready(Err(Box::new(join_err) as Box<dyn Error + Send>))
                 }
                 Poll::Pending => Poll::Pending,
             }
--- a/sdk/core/typespec_client_core/CHANGELOG.md
+++ b/sdk/core/typespec_client_core/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### Breaking Changes
 
 - Added `tokio` feature to `default` features.
-- Changed the `SpawnedTask` return value of `async_runtime::spawn` from a raw future to a trait which enables an `abort()` function on the returned `SpawnedTask`
+- Changed `async_runtime::spawn` to return a `SpawnedTask` trait (supports `abort()`) instead of a raw future.
 
 ### Bugs Fixed
 
--- a/sdk/core/typespec_client_core/src/async_runtime/tests.rs
+++ b/sdk/core/typespec_client_core/src/async_runtime/tests.rs
@@ -268,7 +268,14 @@ async fn tokio_abort_cancels_task() {
 
     handle.abort();
 
-    // The task should not have completed
+    // Awaiting the aborted task should yield an error (cancellation), not a successful completion.
+    let result = handle.await;
+    assert!(
+        result.is_err(),
+        "aborted task should return an error when awaited"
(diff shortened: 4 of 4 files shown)


sdk/core/azure_core/CHANGELOG.md
```diff
@@ -4,11 +4,10 @@
 
 ### Features Added
 
-- Derived `Clone` on `ErrorResponse`, `ErrorDetail`, and `InnerError`.
-
 ### Breaking Changes
 
 - Added `tokio` feature to `default` features.
+- Changed te `SpawnedTask` return value of `async_runtime::spawn` from a raw future to a trait which enables an `abort()` function on the returned `SpawnedTask`
```
heaths:
```suggestion
- Changed the `SpawnedTask` return value of `async_runtime::spawn` from a raw future to a trait which enables an `abort()` function on the returned `SpawnedTask`
```

sdk/core/typespec_client_core/src/async_runtime/mod.rs
```diff
@@ -42,15 +42,29 @@ mod tests;
...
+// /// A `SpawnedTask` is a future that represents a running task.
+// /// It can be awaited to block until the task has completed.
+// pub type SpawnedTask = Pin<
+//     Box<
+//         dyn Future<Output = std::result::Result<(), Box<dyn std::error::Error + Send>>>
+//             + Send
+//             + 'static,
+//     >,
+// >;
+
```
heaths: Before you undraft it, probably just remove this.

sdk/core/typespec_client_core/src/async_runtime/mod.rs
```diff
@@ -42,15 +42,29 @@ mod tests;
...
-    Box<
-        dyn Future<Output = std::result::Result<(), Box<dyn std::error::Error + Send>>>
-            + Send
-            + 'static,
-    >,
->;
+/// It can be awaited to block until the task has completed,
+/// and it also provides an `abort` method to cancel the task.
+pub trait SpawnedTask:
+    Future<Output = std::result::Result<(), Box<dyn std::error::Error + Send>>> + Send
```
heaths: What if we just make this an `IntoFuture`? `await` still desugars, but might - _might_? - be easier to implement everywhere. That said, I don't dislike this. Just a thought.

heaths: Oh, but now I also better understand how you're basically just adding a method. Still, worth considering but given the timeframe - even regardless of the timeframe - I'm not suggesting holding this up at all.

sdk/core/typespec_client_core/src/async_runtime/tokio_runtime.rs
```diff
@@ -32,3 +35,33 @@ impl AsyncRuntime for TokioRuntime {
...
+                Poll::Ready(_) => {
+                    self.handle = None;
+                    Poll::Ready(Ok(()))
+                }
+                Poll::Pending => Poll::Pending,
+            }
+        } else {
+            Poll::Ready(Ok(()))
+        }
+    }
```
heaths: Might look at using `pin_project` here like we do elsewhere. That said, probably wouldn't clean this up much so maybe not worth it. Again, just a thought.

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -242,7 +242,7 @@ fn start_download_task<Behavior: PartitionedDownloadBehavior + Send + Sync + 'st
     mut sender: UnboundedSender<Result<(usize, Bytes), Error>>,
     active_tasks_counter: Arc<AtomicUsize>,
     chunk_idx: usize,
-) -> SpawnedTask {
+) -> Pin<Box<dyn SpawnedTask>> {
```
heaths: As many changes as there are and given the syntax required, would it make sense to maybe keep `SpawnedTask` as the typedef, but then make a new trait in lieu of making `SpawnedTask` that trait e.g., `TaskFuture` or something?

LarryOsterman: Oh, I *like* that one. Let me try it.

sdk/core/typespec_client_core/src/async_runtime/mod.rs
```diff
@@ -151,17 +171,16 @@ pub fn get_async_runtime() -> Arc<dyn AsyncRuntime> {
...
-///     set_async_runtime, AsyncRuntime, TaskFuture, SpawnedTask};
+///     set_async_runtime, AbortableTask, AsyncRuntime, TaskFuture, SpawnedTask};
 /// use std::sync::Arc;
-/// use futures::FutureExt;
 ///
 /// struct CustomRuntime;
 ///
 /// impl AsyncRuntime for CustomRuntime {
-///    fn spawn(&self, f: TaskFuture) -> SpawnedTask {
+///    fn spawn(&self, _f: TaskFuture) -> SpawnedTask {
```
heaths, line 180: Odd. Did clippy complain about these? Given they are doc tests, I'm surprised you'd get an "unused variable" warning (I imagine). No biggie - just unexpected.

PR #4067, opened 2026-03-30
Add ReadStream for AsyncRead implementations

heaths:
Also adds helpers to `RequestContent` to make it more discoverable to from where request content can come.


heaths:
I'll need to `git rebase --onto` after #4057 merges. At the time of this comment, only the last commit here is substantial.

heaths:
So much has changed, I manually ported most of this over to #4491 along with some other changes.

PR #4068, opened 2026-03-31
[Storage] Finalized `managed_download` and `managed_download_to` drop-in replacements

vincenttran-msft:
Replaces `BlobClient::download()` with `managed_download` and the same exercise for `download_into`.

## Summary

Promotes the managed (multi-part) download implementation to be the sole `download()` API on `BlobClient`, and adds `download_into()` which writes directly into a caller-supplied buffer. Both APIs use parallel range requests under the hood and return structured result types with all service headers parsed out directly.

## Breaking Changes

- **`BlobClient::download()` return type changed**
  - Before: `Result<AsyncResponse<BlobClientDownloadResult>>`
  - After: `Result<BlobClientDownloadResult>` - all headers are fields on the result struct directly; no `.await` on the response wrapper needed
- **`BlobClientDownloadResult` is a new hand-written struct** with a `body: AsyncResponseBody` field plus a `properties: BlobDownloadProperties` field covering every header the Blob service returns (ETag, Content-Type, lease state, copy status, blob type, metadata, object-replication rules, etc.)
- **`BlobClientDownloadResultHeaders` trait removed** — superseded by direct field access on `BlobClientDownloadResult`
- **`range` field type on `BlobClientDownloadOptions` changed**
  - Before: `Option<String>`
  - After: `Option<Range<usize>>`
- **`BlobClientManagedDownloadOptions` renamed** to `BlobClientDownloadOptions`
- **`BlockBlobClient::download()` and `BlockBlobClientDownloadBehavior` removed** - download belongs on `BlobClient`

### `sdk/storage/azure_storage_blob`

- **`src/clients/blob_client.rs`**
  - `BlobClient::download()` now owns the full managed (multi-part) download implementation via `BlobClientDownloadBehavior` and the `DEFAULT_DOWNLOAD_PARALLEL` / `DEFAULT_DOWNLOAD_PARTITION_SIZE` constants
  - `BlobClient::download_into()` added - writes blob content directly into a caller-provided `&mut [u8]` buffer using the same partitioned transfer engine
- **`src/clients/block_blob_client.rs`**
  - Removed `BlockBlobClient::download()`, `BlockBlobClientDownloadBehavior`, and all download-related imports
- **`src/models/download_result.rs`** *(new file)*
  - Hand-written `BlobClientDownloadResult` with `body`, `properties`, and `headers` fields
  - `BlobDownloadProperties` covers every header returned by the Blob service
  - `BlobClientDownloadResult::from_headers(AsyncRawResponse)` handles all header parsing in one place
- **`src/models/method_options.rs`**
  - Renamed `BlobClientManagedDownloadOptions` --> `BlobClientDownloadOptions`
- **`src/models/mod.rs`**
  - Exports `BlobClientDownloadResult`, `BlobDownloadProperties`, and `BlobClientDownloadOptions`
- **`src/parsers.rs`**
  - Added `pub(crate) fn parse_metadata_and_replication_headers(headers)` - extracts `x-ms-meta-*` and `x-ms-or-*` headers into reusable maps
- **`src/partitioned_transfer/download.rs`**
  - Return type of internal `download()` changed to `AzureResult<AsyncRawResponse>` - surfaces the initial range-response headers without wrapping them in a misleading empty-body `RawResponse`
  - Internal unit tests updated to match

### Tests

Updated the following integration test files to use direct field access on `BlobClientDownloadResult`:

- `tests/blob_client.rs`
- `tests/blob_client_options.rs`
- `tests/blob_client_snapshot.rs`
- `tests/blob_client_versioning.rs`
- `tests/blob_cpk.rs`
- `tests/block_blob_client.rs`
- `tests/append_blob_client.rs`
- `tests/page_blob_client.rs`
- `tests/streaming.rs`

### DIFF
--- a/sdk/storage/azure_storage_blob/src/models/download_result.rs
+++ b/sdk/storage/azure_storage_blob/src/models/download_result.rs
@@ -34,7 +34,7 @@ pub struct BlobClientDownloadResult {
     /// All headers from the initial response.
     ///
     /// Use this to access headers that are not surfaced as named fields, such as
-    /// `x-ms-encryption-key-sha256`, `x-ms-request-id`, and more.
+    /// `x-ms-request-id`, `x-ms-client-request-id`, and more.
     pub headers: Headers,
 }
 
@@ -157,6 +157,13 @@ pub struct BlobDownloadProperties {
 
     /// Number of tags on the blob (`x-ms-tag-count` header).
     pub tag_count: Option<i64>,
+
+    /// The name of the encryption scope used to encrypt the blob (`x-ms-encryption-scope` header).
+    pub encryption_scope: Option<String>,
+
+    /// Base64-encoded SHA-256 hash of the customer-provided encryption key
+    /// used to encrypt the blob (`x-ms-encryption-key-sha256` header).
+    pub encryption_key_sha256: Option<String>,
 }
 
 impl BlobClientDownloadResult {
@@ -243,6 +250,10 @@ impl BlobDownloadProperties {
             object_replication_policy_id: headers
                 .get_optional_as(&HeaderName::from_static("x-ms-or-policy-id"))?,
             tag_count: headers.get_optional_as(&HeaderName::from_static("x-ms-tag-count"))?,
+            encryption_scope: headers
+                .get_optional_as(&HeaderName::from_static("x-ms-encryption-scope"))?,
+            encryption_key_sha256: headers
+                .get_optional_as(&HeaderName::from_static("x-ms-encryption-key-sha256"))?,
             metadata,
             object_replication_rules,
         })


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -236,14 +184,19 @@ impl BlobClient {
...
+    /// # Arguments
     ///
     /// * `options` - Optional configuration for the request.
     pub async fn download(
         &self,
         options: Option<BlobClientDownloadOptions<'_>>,
-    ) -> Result<AsyncResponse<BlobClientDownloadResult>> {
-        self.download_internal(options).await
+    ) -> Result<BlobClientDownloadResult> {
+        self.block_blob_client().download(options).await
```
jaschrep-msft: Download is implemented directly on block blob now? Why?

sdk/storage/azure_storage_blob/src/clients/block_blob_client.rs
```diff
@@ -221,6 +230,201 @@ impl BlockBlobClient {
...
+                && name.starts_with(OR_PREFIX)
+                && name != "x-ms-or-policy-id"
+            {
+                object_replication_rules.insert(
+                    name[OR_PREFIX.len()..].to_owned(),
+                    value.as_str().to_owned(),
+                );
+            }
+        }
+        Ok(BlockBlobClientDownloadResult {
```
jaschrep-msft: Lotta oneliners above this just have a lotta lines below to put them in the struct. Maybe we condense these? Or write a separate converter?

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
```diff
@@ -50,7 +50,7 @@ pub(crate) async fn download<Behavior>(
     parallel: NonZero<usize>,
     partition_size: NonZero<usize>,
     client: Arc<Behavior>,
-) -> AzureResult<PinnedStream>
+) -> AzureResult<(RawResponse, PinnedStream)>
```
jaschrep-msft: Returning `RawResponse` is misleading imo. We don't return an actual response, and what we do return contains an empty body which can be very misleading.

I'd rather just return raw headers. Probably even a status code is too much to return (if someone wants a whole blob, they'll be confused at the Partial Content code.

sdk/storage/azure_storage_blob/src/clients/block_blob_client.rs
```diff
@@ -221,6 +230,201 @@ impl BlockBlobClient {
...
+            copy_completed_on,
+            copy_id,
+            copy_progress,
+            copy_source,
+            copy_status,
+            copy_status_description,
+            object_replication_policy_id,
+            object_replication_rules,
+            tag_count,
+            raw_response,
```
jaschrep-msft: Putting a surrogate raw response in here is a little worrying.

sdk/storage/azure_storage_blob/src/clients/block_blob_client.rs
```diff
@@ -221,6 +230,201 @@ impl BlockBlobClient {
...
+            } else if name.len() > OR_PREFIX.len()
+                && name.starts_with(OR_PREFIX)
+                && name != "x-ms-or-policy-id"
+            {
+                object_replication_rules.insert(
+                    name[OR_PREFIX.len()..].to_owned(),
+                    value.as_str().to_owned(),
+                );
+            }
+        }
```
jaschrep-msft: Can you make this a utility actually? We'll definitely be reusing this in the future.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
vincenttran-msft: Woops... Removed

sdk/storage/azure_storage_blob/src/clients/block_blob_client.rs
vincenttran-msft: Added `BlobClientDownloadResult::from_headers()` helper to resolve this

sdk/storage/azure_storage_blob/src/partitioned_transfer/download.rs
vincenttran-msft: Now returning `AzureResult<(Headers, PinnedStream)>`

sdk/storage/azure_storage_blob/src/clients/block_blob_client.rs
vincenttran-msft: Good point. And the subtlety that the status code is probably not what they expect anyways (partial content on a successfully completed multi-part upload). Took your feedback here and removed it entirely.

sdk/storage/azure_storage_blob/src/clients/block_blob_client.rs
vincenttran-msft: Made into `parsers::parse_metadata_and_replication_headers`

sdk/core/typespec_client_core/src/http/response.rs
```diff
@@ -302,7 +302,7 @@ enum Body {
 
 impl AsyncResponseBody {
     /// Create a new [`AsyncResponseBody`] from an async stream of bytes.
-    fn new(stream: PinnedStream) -> Self {
+    pub fn new(stream: PinnedStream) -> Self {
```
heaths: These aren't supposed to be directly creatable. That's a design decision. Why can't you get it from the `AsyncResponse` just like we do with `Response` and `ResponseBody`?

heaths (changes requested):
`AsyncResponse` should not be publicly creatable. We can explore other options.

sdk/storage/azure_storage_blob/src/models/download_result.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// The blob content stream.
+    pub body: AsyncResponseBody,
+
+    /// Blob properties parsed from the initial response.
+    pub properties: BlobDownloadProperties,
+
+    /// All headers from the initial response.
+    ///
+    /// Use this to access headers that are not surfaced as named fields, such as
+    /// `x-ms-encryption-key-sha256`, `x-ms-request-id`, and more.
```
jalauzon-msft: I think the encryption-key ones are fine to parse if you want to bring those back. Just not `x-ms-server-encrypted` since its always just true. Don't forget to update this comment. If you want two examples, request id and client-reqesut-id

vincenttran-msft: Updated comment, brought back `encryption_scope` and `encryption_key_sha256`

PR #4072, opened 2026-03-31
Customize Target Proxy URL

scbedd:
@LarryOsterman here is an example.

@heaths I made a change in `existing()` to honor an optional value. That might get me yelled at but we'll see 😆


heaths (approved):
Yeah, this could work. I was debating on whether we should take an `Option<&str>` or something instead of reading env vars here, but we read env vars above for Kestrel.

I still wonder if we want to take this, though, instead of figuring out how/why the port is already in use. If we (almost definitely) rightly assume no other agents are running on those machines, when we fail to bind should we actually query and report on what's running off port 5000? This feels more like papering over the problem than trying to understand the root cause and fix it. Then again, maybe it's rare enough it's not worth the investigative effort.

PR #4075, opened 2026-03-31
[Storage] `upload` roundtrip concerns with `download` API

vincenttran-msft:
This is applicable to its current state, but will also still be applicable even after `managed_download` replaces `download`.

Against our current authoring, the only way to upload a downloaded blob is to buffer in-memory to be able to get it into a `RequestContent`:
```rust
 let download_response = source_blob_client.download(None).await?;
 let buffered: Bytes = download_response.into_body().collect().await?;
 // Only after full buffering can the data be passed to `upload`.
    dest_blob_client
        .upload(RequestContent::from(buffered.to_vec()), None)
        .await?;
```

Applying a similar workaround as @benbp :
```rust
    // content_length must be read from headers *before* deconstruct() consumes the response,
    // because the body stream itself has no len().
    let download_response = source_blob_client.download(None).await?;
    let content_length = download_response.content_length()?;
    let (_, _, body) = download_response.deconstruct();

 // use our custom SeekableStream adapter impl, but this makes retries impossible due to the always-error reset() impl
    let stream = DownloadBodySeekableStream {
        inner: Arc::new(Mutex::new(DownloadBodyInner {
            stream: Some(Box::pin(body)),
            remainder: Bytes::new(),
        })),
        len: content_length.unwrap_or(0),
    };

// But allows this to flow the bytes directly through instead of buffering in-memory
    dest_blob_client
        .upload(Body::SeekableStream(Box::new(stream)).into(), None)
        .await?;
```

### DIFF
--- a/sdk/storage/azure_storage_blob/tests/blob_client.rs
+++ b/sdk/storage/azure_storage_blob/tests/blob_client.rs
@@ -3,7 +3,11 @@
 
 use azure_core::{
     error::ErrorKind,
-    http::{headers::CONTENT_TYPE, ClientOptions, RequestContent, StatusCode, Url},
+    http::{
+        headers::CONTENT_TYPE, response::PinnedStream, Body, ClientOptions, RequestContent,
+        StatusCode, Url,
+    },
+    stream::SeekableStream,
     time::{parse_rfc3339, to_rfc3339, OffsetDateTime},
     Bytes,
 };
@@ -26,16 +30,18 @@ use azure_storage_blob_test::{
     StorageAccount, TestPolicy,
 };
 use bytes::{BufMut, BytesMut};
-use futures::TryStreamExt;
+use futures::{io::AsyncRead, StreamExt, TryStreamExt};
 use std::{
     cmp::min,
     collections::HashMap,
     error::Error,
     num::NonZero,
+    pin::Pin,
     sync::{
         atomic::{AtomicUsize, Ordering},
-        Arc,
+        Arc, Mutex,
     },
+    task::{Context, Poll},
     time::Duration,
 };
 use tokio::time;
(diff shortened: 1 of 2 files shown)


sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -1151,3 +1151,59 @@ async fn test_set_blob_properties_content_headers(ctx: TestContext) -> Result<()
...
+
+    // The content-length header is available, but the body stream itself has no `len()` method
+    // and no `reset()`. There is no conversion from AsyncResponseBody into RequestContent or
+    // Body::SeekableStream. The only path forward is to collect the entire stream into memory.
+    let buffered: Bytes = download_response.into_body().collect().await?;
+
+    // Only after full buffering can the data be passed to `upload`.
+    dest_blob_client
+        .upload(RequestContent::from(buffered.to_vec()), None)
+        .await?;
```
vincenttran-msft, line 1202: This is the crux of the issue.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -1151,3 +1157,208 @@ async fn test_set_blob_properties_content_headers(ctx: TestContext) -> Result<()
...
+    // Wrap the response body in the SeekableStream adapter.
+    // Box::pin(body) converts AsyncResponseBody (!Unpin internally) into a PinnedStream that
+    // can be polled via poll_next_unpin() from behind a Mutex in poll_read.
+    let stream = DownloadBodySeekableStream {
+        inner: Arc::new(Mutex::new(DownloadBodyInner {
+            stream: Some(Box::pin(body)),
+            remainder: Bytes::new(),
+        })),
+        len: content_length.unwrap_or(0),
+    };
```
vincenttran-msft, line 1346: This is the crux of the workaround: a custom `DownloadBodySeekableStream` adapter so that we can get it into a `RequestContent` without collecting in-memory.

PR #4078, opened 2026-04-01
[Cosmos] ThroughputControl API

### LINKED ISSUE #3904: [Cosmos] ThroughputControl API
We should provide our customers with an API that allows them to interact with different throughput configurations that are exposed in the other SDKs, as well as the concept of throughput control groups that exist in the Java SDK.

In particular, the API we expose should allow for configuring:
- Priority Based Execution server feature
- Throughput Buckets server feature
- Throughput Control Groups client-side feature: https://learn.microsoft.com/en-us/java/api/com.azure.cosmos.ThroughputControlGroupConfigBuilder?view=azure-java-stable

tvaron3:
## Summary
 
 Add server-side throughput control API enabling priority-based execution and throughput buckets.
 
 ## Flow
 
 ### Registration (build time)
 ```mermaid
graph LR
     A[CosmosClientBuilder] -->|with_throughput_control_group| B[stores Vec of groups]
     B -->|build| C[CosmosDriverRuntimeBuilder]
     C -->|register_throughput_control_group| D[ThroughputControlGroupRegistry]
     D --> E[CosmosDriverRuntime]
     E --> F[CosmosDriver]
     F --> G[CosmosClient]
```
Per-request resolution
```mermaid
 graph TD
     A[OperationOptions::throughput_control_group] -->|resolve| B{Name provided?}
     B -->|Yes| C[Look up name in registry]
     C --> D{Found?}
     D -->|Yes| E[Snapshot group state]
     D -->|No| F[Return error]
     B -->|No| G[Look up default group for container]
     G -->|Found| E
     G -->|Not found| H[Return None - no headers set]
     E --> I[build_transport_request]
     I -->|priority_level| J[x-ms-cosmos-priority-level header]
     I -->|throughput_bucket| K[x-ms-cosmos-throughput-bucket header]
```
### Setting a group per request

 let mut operation = OperationOptions::default();
 operation.throughput_control_group = Some(ThroughputControlGroupName::new("my-group"));

A single group can carry both a priority level and a throughput bucket:

```rust
 let group = ThroughputControlGroupOptions::new("my-group", container, true)
     .with_throughput_bucket(42)
     .with_priority_level(PriorityLevel::High);
```

### Key Design Changes

ThroughputControlGroupOptions: enum → struct

The old enum (ServerSideThroughputBucket | ServerSidePriorityBasedThrottling) forced a group to carry either a throughput bucket or a priority level, but not both. The
new struct holds both as optional fields behind a single RwLock<MutableValues>.

### Single group per request

OperationOptions::throughput_control_group is Option<ThroughputControlGroupName> (was Option<Vec<...>>). No multi-group merging needed since a single group carries
both values.

### Minimal public API

Only ThroughputControlGroupOptions is public. All internal types are pub(crate):

 - ThroughputControlGroupSnapshot — internal point-in-time copy for the pipeline
 - ThroughputControlGroupKey — internal composite key for the registry
 - ThroughputControlGroupRegistry — internal group storage
 - ThroughputControlGroupRegistrationError — wrapped into azure_core::Error at the builder boundary

#### ThroughputControlGroupOptions:

 - new(name, container, is_default) + .with_throughput_bucket(u32) / .with_priority_level(PriorityLevel)
 - Getters: name(), container(), is_default(), throughput_bucket(), priority_level()
 - Setters: set_throughput_bucket(u32), set_priority_level(PriorityLevel)

CosmosDriverRuntimeBuilder::register_throughput_control_group() — returns azure_core::Result<Self>

#### Driver changes (azure_data_cosmos_driver)

 - Removed ClientSide variant (deferred for GA; tracked in #4089)
 - Removed dead ThroughputTarget enum
 - Pipeline header application: effective_throughput_control_group() passes resolved snapshot through execute_operation_pipeline() to build_transport_request()
 - Added PRIORITY_LEVEL and THROUGHPUT_BUCKET constants to request_header_names
 - Extracted TransportRequestContext struct to reduce build_transport_request parameter count

#### SDK changes (azure_data_cosmos)

 - CosmosClientBuilder::with_throughput_control_group() registers groups, passes to CosmosDriverRuntimeBuilder at build time
 - ClientContext refactor: extracted struct bundling shared infrastructure threaded through CosmosClient → DatabaseClient → ContainerClient

Depends on #4059

Fixes #3904


sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -9,13 +9,151 @@ use std::fmt::Display;
...
+/// let group = ThroughputControlGroup::priority_based(
+///     "analytics",
+///     container,
+///     PriorityLevel::Low,
+///     false, // not the default group
+/// );
+/// ```
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub enum ThroughputControlGroup {
```
analogrelay: Java's ThroughputControlGroupBuilder seems to allow setting _both_ a throughput bucket AND a priority level. I kinda viewed TCGs as a struct holding optional Priority Level, and Throughput Bucket values (for server-side throughput control) as well as (in the future) Target Throughput (for client-side throughput control).

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -575,6 +582,7 @@ impl ContainerClient {
 
         // Execute through the driver.
         let driver_response = self
+            .context
             .driver
             .execute_operation(operation, options.operation)
             .await?;
```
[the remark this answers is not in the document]
tvaron3: Fixed — throughput control headers are now applied in the driver pipeline (`build_transport_request()`) which handles ALL operations including `read_item`. The SDK no longer has separate `apply_throughput_control()` calls. Resolution happens in `CosmosDriver::effective_throughput_control_group()` before the pipeline executes.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -9,13 +9,151 @@ use std::fmt::Display;
...
+    ThroughputControlGroupKey, ThroughputControlGroupRegistry,
 };
 
+/// Configuration for a throughput control group.
+///
+/// Throughput control groups associate a named policy with a container. When a
+/// request references a group name via
+/// [`OperationOptions::throughput_control_group_name`], the SDK sets the
+/// corresponding HTTP headers so the Cosmos DB service enforces the policy.
+///
```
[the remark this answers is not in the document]
tvaron3: Resolved — the SDK-level `ThroughputControlGroup` wrapper and `apply_throughput_control()` methods have been removed. Throughput control headers are now applied uniformly in the driver pipeline for all operations. The doc inaccuracy no longer exists since driver types are re-exported directly.

sdk/cosmos/azure_data_cosmos/src/clients/mod.rs
```diff
@@ -9,6 +9,33 @@ mod cosmos_client_builder;
...
+///
+/// Bundling these fields avoids passing them individually through every
+/// constructor in the client hierarchy.
+#[derive(Clone, Debug)]
+pub(crate) struct ClientContext {
+    pub pipeline: Arc<GatewayPipeline>,
+    pub driver: Arc<CosmosDriver>,
+    pub global_endpoint_manager: Arc<GlobalEndpointManager>,
+    pub global_partition_endpoint_manager: Arc<GlobalPartitionEndpointManager>,
+    pub throughput_control_groups: ThroughputControlGroupRegistry,
```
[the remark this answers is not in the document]
tvaron3: Addressed differently — `ClientContext` no longer stores a `ThroughputControlGroupRegistry`. The registry now lives exclusively in the driver's `CosmosDriverRuntime`. Groups are registered via `CosmosClientBuilder::with_throughput_control_group()` which passes them to `CosmosDriverRuntimeBuilder` at build time. The SDK accesses the registry read-only through `CosmosClient::throughput_control_groups()` → `driver.runtime().throughput_control_groups()`. No cloning of the registry occurs.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -58,6 +224,12 @@ pub struct CosmosClientOptions {
     pub(crate) operation: OperationOptions,
     pub(crate) user_agent_suffix: Option<String>,
     pub(crate) application_region: Option<Region>,
+
+    /// Registry of throughput control groups for rate limiting.
+    ///
+    /// Groups are registered at client creation time and are immutable after
+    /// that (except for mutable runtime values like priority level).
+    pub(crate) throughput_control_groups: ThroughputControlGroupRegistry,
```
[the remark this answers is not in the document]
tvaron3: Fixed — the registry field and its doc comment have been removed from `CosmosClientOptions`. The registry now lives in the driver runtime. No misleading wording remains on the SDK side.

sdk/cosmos/azure_data_cosmos_driver/src/models/resource_reference.rs
```diff
@@ -217,6 +217,30 @@ impl ContainerReference {
...
+    /// key definition. It should only be used in unit tests that need a
+    /// `ContainerReference` value without connecting to a real service.
+    #[doc(hidden)]
+    pub fn stub(account: AccountReference, db_name: &str, container_name: &str) -> Self {
+        let pk: crate::models::PartitionKeyDefinition =
+            serde_json::from_str(r#"{"paths":["/id"]}"#).expect("valid partition key definition");
+        let props = crate::models::ContainerProperties {
+            id: container_name.to_string().into(),
+            partition_key: pk,
+            system_properties: Default::default(),
```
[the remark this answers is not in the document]
tvaron3: Fixed — `ContainerReference::stub()` has been removed entirely. SDK unit tests that needed a `ContainerReference` were removed; throughput control behavior is covered by driver-level tests and will be validated end-to-end via emulator integration tests.

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
tvaron3: Addressed — instead of merging into a single struct, we kept the enum (each variant represents a distinct server-side feature) and changed `OperationOptions::throughput_control_group_name` to `throughput_control_group_names: Option<Vec<ThroughputControlGroupName>>`. This lets a single request reference multiple groups (e.g., one priority group + one bucket group) and the driver merges their snapshots, applying both `x-ms-cosmos-priority-level` and `x-ms-cosmos-throughput-bucket` headers. The `ClientSide` variant was removed for GA (`#[non_exhaustive]` allows re-adding it later with `target_throughput`).

tvaron3: Actually used a struct instead as recommended to enforce that only one throughput bucket and priority level can be set through the `ThroughputControlGroupAssignment`.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
```diff
@@ -105,18 +96,20 @@ impl CosmosClient {
...
+        &self.context.pipeline.endpoint
+    }
+
+    /// Returns the throughput control group registry.
+    ///
+    /// Groups are registered at client creation time via
+    /// [`CosmosClientBuilder::with_throughput_control_group()`].
+    pub fn throughput_control_groups(&self) -> &ThroughputControlGroupRegistry {
+        self.context.driver.runtime().throughput_control_groups()
     }
```
analogrelay: Does the user need access to this after client creation?

sdk/cosmos/azure_data_cosmos_driver/src/options/throughput_control.rs
```diff
@@ -351,56 +265,88 @@ impl ThroughputControlGroupKey {
...
+/// - `Some({ bucket_group: Some(..), .. })` — apply the named bucket group.
+/// - Both fields set — apply both groups to the request.
+#[non_exhaustive]
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct ThroughputControlGroupAssignment {
+    /// Name of the priority-based throttling group to apply.
+    pub priority_group: Option<ThroughputControlGroupName>,
+    /// Name of the throughput bucket group to apply.
+    pub bucket_group: Option<ThroughputControlGroupName>,
+}
```
analogrelay: This feels a little weird. My understanding is that a client will have a number of named throughput control groups registered. Each group contains 0 or more throughput settings (Priority Level, Throughput Buckets, and in the future Client-side throttling). A request specifies **at most one** throughput group by name. If a request specifies no throughput group, it gets a default group name assigned. Then, IF we find that throughput group in the registry, we apply **all** of that groups settings to the request.

This seems to be allowing for multiple groups, one for priority and one for bucketing? Seems messy to me. I think it's much simpler to just have one group per request. If a user wants to be able to mix and batch priority levels and buckets, they can just do that by creating all the combinations they need.

sdk/cosmos/azure_data_cosmos/src/clients/cosmos_client.rs
tvaron3: Good call — users don't need this after client creation. Will make it `pub(crate)`.

sdk/cosmos/azure_data_cosmos_driver/src/options/throughput_control.rs
tvaron3: Agreed — one group per request is simpler. Users who need both priority and bucket can create groups with the combinations they need. Will simplify to a single `ThroughputControlGroupName` per operation.

sdk/cosmos/azure_data_cosmos/src/clients/mod.rs
```diff
@@ -9,6 +9,31 @@ mod cosmos_client_builder;
...
+/// through [`DatabaseClient`](super::DatabaseClient) to
+/// [`ContainerClient`](super::ContainerClient).
+///
+/// Bundling these fields avoids passing them individually through every
+/// constructor in the client hierarchy.
+#[derive(Clone, Debug)]
+pub(crate) struct ClientContext {
+    pub(crate) pipeline: Arc<GatewayPipeline>,
+    pub(crate) driver: Arc<CosmosDriver>,
+    pub(crate) global_endpoint_manager: Arc<GlobalEndpointManager>,
```
FabianMeiswinkel, line 34: Why do pipeline, global_endpointmanager etc. exist as siblings to driver - shouldn't they either be part of DriverRuntime or Driver? Or is this a left-over because not all operations are cut over to the driver yet?

tvaron3: This is leftover because not all operations are cutover to driver. Post cutover this struct can be removed as only the driver would exist for this struct.

tvaron3: For reference, I made this change to get rid of the clippy error of too many args, but I can move it back to how it was and just disable that error here.

sdk/cosmos/azure_data_cosmos_driver/src/options/throughput_control.rs
```diff
@@ -19,312 +19,101 @@ use std::{
...
-///
-/// These values can be modified at runtime via the registry.
-#[derive(Clone, Debug)]
-pub(crate) struct ClientSideMutableValues {
-    /// Target throughput limit.
-    target_throughput: ThroughputTarget,
-    /// Optional priority level for throttling decisions.
+/// Mutable runtime values for a throughput control group.
+#[derive(Clone, Debug, Default)]
+struct MutableValues {
```
analogrelay, line 24: nit: I'd call this something like `ThroughputControlSettings` or something. `MutableValues` is a bit odd (even if it is module-local).

sdk/cosmos/azure_data_cosmos_driver/src/options/throughput_control.rs
```diff
@@ -19,312 +19,101 @@ use std::{
...
-        Self::ServerSideThroughputBucket {
-            name: name.into(),
-            container,
-            is_default,
-            mutable: Arc::new(RwLock::new(ServerSideThroughputBucketMutableValues::new(
-                throughput_bucket,
-            ))),
-        }
+    /// Sets the initial throughput bucket value.
+    pub fn with_throughput_bucket(self, bucket: u32) -> Self {
```
analogrelay, line 65: Is there a specific reason for the throughput bucket to be a `u32`? If it's just "any unsigned integer", using `usize` is a little more ergonomic. If it has range constraints, maybe we should have a `ThroughputBucket` newtype?

sdk/cosmos/azure_data_cosmos_driver/src/options/throughput_control.rs
tvaron3: Will create a followup pr to address.

sdk/cosmos/azure_data_cosmos_driver/src/options/throughput_control.rs
tvaron3: Will create a followup pr to address.

PR #4085, opened 2026-04-03
[Cosmos] Replace global evaluation store with pipeline-level context

### LINKED ISSUE #4022: [Cosmos] Replace global evaluation store with pipeline-level context for fault injection diagnostics
The fault injection evaluation store (`EVALUATION_STORE` in `fault_injection/mod.rs`) uses a global `LazyLock<Mutex<HashMap<u64, Vec<FaultInjectionEvaluation>>>>` to pass evaluation results from `FaultClient` (at the `HttpClient` level) to the transport pipeline (which has access to `DiagnosticsContextBuilder`).

This works but is clunky — it requires a store/take contract, uses a global static, and passes correlation IDs via request headers.

When the driver has its own HTTP pipeline with a custom context object (rather than using `azure_core::HttpClient` trait), the evaluations should be encoded directly in the operation/transport metadata and flow through the pipeline without a side channel.

**Current approach:**
1. Transport pipeline generates a unique request ID, adds as `x-ms-fault-injection-request-id` header
2. `FaultClient` reads the ID, stores evaluations in a global `HashMap` by that ID
3. Transport pipeline retrieves evaluations by ID after the await

**Target approach:**
Pass an evaluation collector (e.g., `Arc<Mutex<Vec<FaultInjectionEvaluation>>>`) through the pipeline context object directly, eliminating the global store and header-based correlation.

**Blocked on:** Custom HTTP pipeline with context passing (Transport Pipeline Spec future step).

Ref: PR #3996 review comment by @analogrelay

tvaron3:
Replace the global `EVALUATION_STORE` (`LazyLock<Mutex<HashMap>>`) and header-based correlation with a per-request `EvaluationCollector` newtype on `HttpRequest`.

## Changes

- **`fault_injection/mod.rs`** — Removed `EVALUATION_STORE`, `NEXT_EVALUATION_ID`, `store_evaluations()`, `take_evaluations()`, `next_evaluation_id()`. Added `EvaluationCollector` newtype with `push_all()` and `take()` methods that encapsulate poison recovery.
- **`cosmos_transport_client.rs`** — Added feature-gated `evaluation_collector: Option<EvaluationCollector>` field to `HttpRequest`.
- **`http_client.rs`** —
- **`transport_pipeline.rs`** — Creates collector, attaches via `.clone()`, reads via `.take()` after `send()`.
- **`cosmos_headers.rs`** — Removed `FAULT_INJECTION_REQUEST_ID` header constant.
- **`cosmos_driver.rs`** — Added `evaluation_collector: None` to `HttpRequest` construction.
- **`diagnostics_context.rs`** — Updated doc comment to reference new `EvaluationCollector` mechanism.

## Motivation

The driver now has its own `TransportClient` trait and `HttpRequest` struct, so the global side-channel is no longer needed. The new approach:
- Eliminates the global static and Mutex contention
- Removes the `x-ms-fault-injection-request-id` header and ID counter
- Closes the documented memory leak risk (entries leaked on panics)
- Scopes evaluation lifetime to the request via `Arc` drop semantics

Fixes #4022

### DIFF
--- a/sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
+++ b/sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
@@ -283,7 +283,7 @@ pub(crate) async fn execute_transport_pipeline(
         );
 
         #[cfg(feature = "fault_injection")]
-        let evaluation_collector = if diagnostics.fault_injection_enabled() {
+        let mut evaluation_collector = if diagnostics.fault_injection_enabled() {
             let collector = crate::fault_injection::EvaluationCollector::default();
             http_request.evaluation_collector = Some(collector.clone());
             Some(collector)
@@ -303,7 +303,7 @@ pub(crate) async fn execute_transport_pipeline(
         .await;
 
         #[cfg(feature = "fault_injection")]
-        if let Some(ref collector) = evaluation_collector {
+        if let Some(collector) = evaluation_collector.take() {
             let evals = collector.take();
             if !evals.is_empty() {
                 diagnostics.set_fault_injection_evaluations(request_handle, evals);
--- a/sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/http_client.rs
+++ b/sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/http_client.rs
@@ -345,11 +345,6 @@ impl TransportClient for FaultClient {
 
             if let Some(ref collector) = request.evaluation_collector {
                 collector.push_all(&mut evaluations);
-            } else {
-                tracing::debug!(
-                    count = evaluations.len(),
-                    "fault injection evaluations dropped: no collector on request"
-                );
             }
         }
 
--- a/sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/mod.rs
+++ b/sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/mod.rs
@@ -65,9 +65,14 @@ impl EvaluationCollector {
     }
 
     /// Takes all collected evaluations, leaving the collector empty.
-    pub fn take(&self) -> Vec<FaultInjectionEvaluation> {
-        let mut evaluations = self.0.lock().unwrap_or_else(|e| e.into_inner());
-        std::mem::take(&mut *evaluations)
+    pub fn take(self) -> Vec<FaultInjectionEvaluation> {
+        match Arc::try_unwrap(self.0) {
+            Ok(mutex) => mutex.into_inner().unwrap_or_else(|e| e.into_inner()),
+            Err(arc) => {
+                let mut evaluations = arc.lock().unwrap_or_else(|e| e.into_inner());
+                std::mem::take(&mut *evaluations)
+            }
+        }
     }
 }
 


sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
```diff
@@ -281,15 +283,10 @@ pub(crate) async fn execute_transport_pipeline(
...
-            http_request.headers.insert(
-                FAULT_INJECTION_REQUEST_ID.clone(),
-                azure_core::http::headers::HeaderValue::from(id.to_string()),
-            );
-            id
+        let evaluation_collector = {
+            let collector = crate::fault_injection::EvaluationCollector::default();
+            http_request.evaluation_collector = Some(collector.clone());
+            collector
         };
```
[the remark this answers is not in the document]
tvaron3: Fixed in a79ab96. Collector creation is now guarded by `diagnostics.fault_injection_enabled()` — only allocates when fault injection rules are actually configured at runtime. Added a getter to `DiagnosticsContextBuilder` (field + setter already existed).

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/mod.rs
```diff
@@ -49,45 +47,31 @@ pub use result::{
...
-        .unwrap_or_else(|e| e.into_inner())
-        .remove(&request_id)
-        .unwrap_or_default()
+    /// Takes all collected evaluations, leaving the collector empty.
+    pub fn take(&self) -> Vec<FaultInjectionEvaluation> {
+        self.0
+            .lock()
+            .unwrap_or_else(|e| e.into_inner())
+            .drain(..)
+            .collect()
```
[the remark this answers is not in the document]
tvaron3: Fixed in a79ab96. Replaced `drain(..).collect()` with `std::mem::take(&mut *evaluations)` for an O(1) Vec swap.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/http_client.rs
```diff
@@ -338,40 +336,39 @@ impl TransportClient for FaultClient {
...
-                .headers
-                .get_optional_str(&FAULT_INJECTION_REQUEST_ID)
-            {
-                if let Ok(id) = id_str.parse::<u64>() {
-                    super::store_evaluations(id, evaluations);
-                }
+            if let Some(ref collector) = request.evaluation_collector {
+                collector.push_all(&mut evaluations);
+            } else {
+                tracing::debug!(
```
analogrelay: I'd make this `trace` even. Or tbh, it could probably be removed entirely.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/mod.rs
```diff
@@ -49,45 +47,28 @@ pub use result::{
...
 
-/// Takes (removes and returns) evaluations for a request ID.
-pub(crate) fn take_evaluations(request_id: u64) -> Vec<FaultInjectionEvaluation> {
-    EVALUATION_STORE
-        .lock()
-        .unwrap_or_else(|e| e.into_inner())
-        .remove(&request_id)
-        .unwrap_or_default()
+    /// Takes all collected evaluations, leaving the collector empty.
+    pub fn take(&self) -> Vec<FaultInjectionEvaluation> {
```
analogrelay: Could this take `self` by-value and consume it? Later on, we could even do some optimizations here like avoiding the locking if we end up as the sole holder of the `Arc`, but for now just the semantic clarity of taking ownership of the collector makes sense.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/http_client.rs
tvaron3: Removed in a5b2b38.

sdk/cosmos/azure_data_cosmos_driver/src/fault_injection/mod.rs
tvaron3: Done in a5b2b38. Changed `take()` to consume `self` by value. Also added `Arc::try_unwrap` to skip locking when the caller is the sole holder.

PR #4087, opened 2026-04-03
Cosmos Driver: Hierarchical Partition Key Support

simorenoh:
Adds MultiHash EPK (Effective Partition Key) computation and prefix partition key routing infrastructure to the driver crate, enabling hierarchical partition key (HPK) support for containers with multiple partition key paths.

### MultiHash EPK computation (`effective_partition_key.rs`)

Previously, `EffectivePartitionKey::compute()` fell through to single-hash V2 for `PartitionKeyKind::MultiHash`, producing incorrect EPKs. MultiHash requires each component to be hashed independently — this PR adds that per-component hashing.

- Added `PartitionKeyKind::MultiHash` arm to `compute()` routing to new `effective_partition_key_multi_hash_v2()`
- Each component is independently V2-encoded → MurmurHash3-128 → byte-reversed → top-2-bit masked → hex-encoded, then concatenated (N×32 hex chars)
- Extracted shared `hash_v2_to_epk()` helper used by both single-hash and multi-hash paths
- Algorithm verified against cross-SDK baselines (.NET, Go, Java) via existing `testdata/*.xml` fixtures

### Prefix EPK range computation (`effective_partition_key.rs`)

- Added `compute_range()` for partial/prefix partition keys (fewer components than the container definition)
- Full key → point range (start == end); prefix key on MultiHash → `[prefix_epk, prefix_epk + "FF")` range covering all possible suffix completions

### Prefix routing in PK range cache (`partition_key_range_cache.rs`)

- Added `resolve_partition_key_range_ids()` that handles both full and prefix partition keys
- Full key: point lookup (single range ID); prefix key: EPK range → `resolve_overlapping_ranges()` → multiple range IDs for fan-out

### Tests

9 new unit tests covering:

- Single/two/three-component MultiHash EPK computation with expected value verification
- MultiHash with `Undefined` component (partial HPK)
- MultiHash vs single-hash divergence for multi-component keys
- `compute_range()` for full keys, prefix keys (1-of-3, 2-of-3), and single-hash (always point)

## Follow-up: FeedRange API (PR #3987)

PR [#3987](https://github.com/Azure/azure-sdk-for-rust/pull/3987) introduces `feed_range_from_partition_key()`, which currently uses the SDK's `hash.rs` to compute EPKs. For MultiHash containers, this hits the existing stub and returns incorrect results. Once that method is updated to route through the driver's EPK computation (as part of the SDK-to-driver cutover), it will get correct MultiHash support for free from this PR. Prefix HPK support for the FeedRange API (returning multiple feed ranges for partial keys) will additionally need the `compute_range()` infrastructure added here.

## Follow-up: Query tests and thorough testing

This is the first step of the end-to-end implementation of this feature. The remaining work, piecing together the operations to this logic and ensuring that queries can also use it, rely on the migration to the driver. We need for the migration of requests to the driver to be finalized before we can add these tests.


sdk/cosmos/azure_data_cosmos_driver/src/models/effective_partition_key.rs
tvaron3: I think this is stale now and should be updated.

simorenoh: updated

sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/container_routing_map.rs
```diff
@@ -9,9 +9,49 @@
...
+                Ordering::Equal
+            } else if a.len() > b.len() {
+                Ordering::Greater
+            } else {
+                Ordering::Less
+            }
+        }
+        other => other,
+    }
+}
```
analogrelay: Is there ever a case where two EPKs should _not_ be compared this way? If not, this should just be made into the implementation of `PartialCmp`/`Cmp` for `EffectivePartitionKey`

sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/partition_key_range_cache.rs
```diff
@@ -470,4 +521,143 @@ mod tests {
...
+            .await;
+
+        assert!(result.is_some());
+        let ids = result.unwrap();
+        // The prefix EPK for "tenant1" is a 32-char hex string starting with a digit 0-3
+        // (due to 0x3F mask). With ranges split at "80", the prefix range [epk, epk+"FF")
+        // falls entirely within range "0" (["", "80")). So we expect 1 ID.
+        // This validates the prefix path is exercised (via resolve_overlapping_ranges).
+        assert!(!ids.is_empty());
+        assert!(ids.contains(&"0".to_string()));
```
analogrelay: This test doesn't actually assert that we expect _1 ID_ though. Just that we got some IDs, including `0`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/partition_key_range_cache.rs
```diff
@@ -470,4 +521,143 @@ mod tests {
...
+        let pk = PartitionKey::from("hello");
+
+        let result = cache
+            .resolve_partition_key_range_ids(&container, &pk, false, test_fetch)
+            .await;
+
+        assert!(result.is_some());
+        let ids = result.unwrap();
+        assert_eq!(ids.len(), 1);
+        assert_eq!(ids[0], "0");
```
analogrelay: This could just be one `assert_eq!()` instead of asserting len and first item.

sdk/cosmos/azure_data_cosmos_driver/src/models/effective_partition_key.rs
```diff
@@ -62,11 +62,75 @@ impl EffectivePartitionKey {
...
+            PartitionKeyKind::MultiHash => {
+                // MultiHash is only supported with V2. All MultiHash container definitions
+                // are created with version=2; V1 MultiHash does not exist in Cosmos DB.
+                debug_assert!(
+                    version == PartitionKeyVersion::V2,
+                    "MultiHash requires V2, got {:?}",
+                    version
+                );
+                effective_partition_key_multi_hash_v2(pk_values)
+            }
```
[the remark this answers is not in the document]
analogrelay: Yeah, I'd go ahead and make this a runtime assert. If this slipped through, a user would have a Very Bad Time.

sdk/cosmos/azure_data_cosmos_driver/src/models/effective_partition_key.rs
```diff
@@ -62,11 +62,75 @@ impl EffectivePartitionKey {
...
+        debug_assert!(
+            !pk_values.is_empty(),
+            "compute_range called with empty pk_values"
+        );
+        debug_assert!(
+            pk_values.len() <= pk_definition.paths().len(),
+            "More partition key components ({}) than definition paths ({})",
+            pk_values.len(),
+            pk_definition.paths().len()
+        );
```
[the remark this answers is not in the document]
analogrelay: This should actually be a `Result::Err` I think. Is it possible these values could be coming from user code prior to any validation? Unless we're positive we should already have checked for too many PK components, we shouldn't panic here.

sdk/cosmos/azure_data_cosmos_driver/src/models/effective_partition_key.rs
```diff
@@ -365,21 +458,202 @@ mod tests {
...
+        let pk = vec![PartitionKeyValue::from("redmond".to_string())];
+        let multi = EffectivePartitionKey::compute(
+            &pk,
+            PartitionKeyKind::MultiHash,
+            PartitionKeyVersion::V2,
+        );
+        let single =
+            EffectivePartitionKey::compute(&pk, PartitionKeyKind::Hash, PartitionKeyVersion::V2);
+        assert_eq!(multi.as_str(), single.as_str());
+        assert_eq!(multi.as_str().len(), 32);
```
analogrelay: Not really necessary when string comparison already requires that lengths be the same

sdk/cosmos/azure_data_cosmos_driver/src/models/effective_partition_key.rs
analogrelay, line 26: I wonder if we should enforce that EPKs are always a multiple of 32 characters long, and use padding when necessary. If we do that normalization, then we could have first-class support for fetching any given segment of a multi-hash. Anyway, that's mostly just some data modelling stuff. I don't think the struct shape needs to change, but I think this could be good opportunity to leverage Rust to fully normalize the data representation in-memory.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/container_routing_map.rs
simorenoh: that's right, updated

sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/partition_key_range_cache.rs
simorenoh: updated

sdk/cosmos/azure_data_cosmos_driver/src/driver/cache/container_routing_map.rs
```diff
@@ -617,4 +618,222 @@ mod tests {
         let result = map.try_combine(new_ranges, Some("etag".into()));
         assert!(matches!(result, Err(RoutingMapError::OverlappingRanges)));
     }
+
+    // -- Length-aware EPK ordering tests --
+
+    #[test]
+    fn epk_cmp_equal_strings() {
+        assert_eq!(epk("06AB34CF").cmp(&epk("06AB34CF")), Ordering::Equal);
```
FabianMeiswinkel, line 626: Should there also be a test validating that 06AB34CF and 06AB34CF00000000 are equal?

simorenoh: done

PR #4096, opened 2026-04-05
Expose async state of replace throughput API

### LINKED ISSUE #3792: Cosmos: make replaceThroughput clearly expose the async state/nature of the API
### Feature Summary

Cosmos: make replaceThroughput clearly expose the asny state/nature of the API

### Feature Description

Change replaceThroughput API to provide clear indication whether operation finsihed synchronously or is happening async - and provide a way to check status on the async operation

### Use Case

_No response_

### Alternatives

_No response_

### Additional Context

_No response_

tvaron3:
## What

Resolves #3792 — makes `replace_throughput` clearly expose the async nature of the Cosmos DB throughput API.

The Cosmos DB service may process throughput changes asynchronously (HTTP 202 with `x-ms-offer-replace-pending: true`), but the previous API returned a simple `ResourceResponse<ThroughputProperties>` with no indication of whether the change was still in-flight.

## Changes

- **Renamed** `replace_throughput` → `begin_replace_throughput` on `ContainerClient` and `DatabaseClient` per the [Azure SDK LRO naming guideline](https://azure.github.io/azure-sdk/rust_introduction.html#rust-lro) (`begin_` prefix for LROs).

- **Added `ThroughputPoller`** (in `clients/throughput_poller.rs`) — a custom type that implements both `IntoFuture` (simple `.await`) and `Stream` (manual polling) to handle sync vs async throughput operations:
  - **HTTP 200**: Completes immediately — single-element stream, `.await` resolves instantly.
  - **HTTP 202 / `x-ms-offer-replace-pending: true`**: Polls the offer resource at 5-second intervals until the operation completes.
  - Caches the offer ID from the initial response for efficient single-GET polling (no query per tick).
  - Uses `BoxStream` internally to type-erase the sync vs async stream variants into a single struct field.

- **Migrated offer operations to the driver** — all offer read, query, and replace operations now go through `CosmosDriver::execute_operation()` instead of the gateway pipeline:
  - Eliminated the `OffersClient` struct in favor of free functions in the `offers_client` module (`find_offer`, `read_offer_by_id`, `begin_replace`).
  - Added `CosmosOperation::query_offers()`, `read_offer()`, and `replace_offer()` factory methods to the driver.
  - Added `offer_replace_pending` field to `CosmosResponseHeaders` in the driver and wired it through the driver bridge.
  - The shared `begin_replace()` helper eliminates duplication between `ContainerClient` and `DatabaseClient`.

- **Why not `azure_core::Poller`?** Cosmos throughput LROs use a header-based detection pattern (`x-ms-offer-replace-pending`), not the `Operation-Location` URL pattern that `azure_core::Poller` and `StatusMonitor` expect. A custom type follows the Cosmos SDK convention of wrapping azure_core internals (like `ResourceResponse` wraps `CosmosResponse`, `FeedItemIterator` wraps `BoxStream`).

## Migration

```rust
// Before
let throughput = client
    .replace_throughput(ThroughputProperties::manual(500), None)
    .await?
    .into_model()?;

// After (simple — just await the final result)
let throughput = client
    .begin_replace_throughput(ThroughputProperties::manual(500), None)
    .await?   // start the replace operation
    .await?   // wait for completion (polls if async)
    .into_model()?;

// After (advanced — poll for progress via Stream)
use futures::TryStreamExt;
let mut poller = client
    .begin_replace_throughput(ThroughputProperties::manual(500), None)
    .await?;
while let Some(status) = poller.try_next().await? {
    if let Some(charge) = status.request_charge() {
        println!("Request charge: {charge}");
    }
}
```

## Testing

- Updated all existing throughput test callsites.
- Added `container_throughput_stream_polling` (multi-write test) — exercises the `Stream` interface with 11000 RU/s to trigger async throughput processing against a live Cosmos DB account.
- Added unit tests for `ThroughputPoller` completed path and `is_offer_replace_pending` header detection.
- Added driver bridge tests for `offer_replace_pending` header round-trip.

### DIFF
--- a/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
+++ b/sdk/cosmos/azure_data_cosmos/CHANGELOG.md
@@ -4,7 +4,11 @@
 
 ### Features Added
 
+- Added `ThroughputPoller` type that implements `IntoFuture` and `Stream` for tracking asynchronous throughput replacement operations.
+
 ### Breaking Changes
+
+- Renamed `replace_throughput` to `begin_replace_throughput` on `ContainerClient` and `DatabaseClient`. The return type changed from `ResourceResponse<ThroughputProperties>` to `ThroughputPoller`.
 
 ### Bugs Fixed
 
@@ -16,7 +20,6 @@
 
 - Added `CosmosClientBuilder::with_proxy_allowed(bool)` for explicit opt-in to HTTP proxy usage with documented support limitations. ([#4062](https://github.com/Azure/azure-sdk-for-rust/pull/4062))
 - Added `CustomResponseBuilder` and `FaultInjectionRule::hit_count()` APIs for fault injection, enabling ergonomic construction of synthetic HTTP responses and test verification of rule activation counts. ([#3888](https://github.com/Azure/azure-sdk-for-rust/pull/3888))
-- Added `ThroughputPoller` type that implements `IntoFuture` and `Stream` for tracking asynchronous throughput replacement operations.
 
 ### Breaking Changes
 
@@ -24,7 +27,6 @@
 - Client methods now return dedicated response types instead of `CosmosResponse<T>`: `ItemResponse<T>` for point operations, `ResourceResponse<T>` for resource management, `BatchResponse` for transactional batch, and `QueryFeedPage<T>` for query pages. `etag()` returns `Option<&Etag>` instead of `Option<&str>`, and `activity_id()` / `server_duration_ms()` are accessed via `response.diagnostics()`. ([#3960](https://github.com/Azure/azure-sdk-for-rust/pull/3960))
 - `FeedPage::deconstruct()` has been removed. Use `into_items()`, `continuation()`, `headers()`, and `diagnostics()` instead. ([#3960](https://github.com/Azure/azure-sdk-for-rust/pull/3960))
 - Replaced `CosmosClientBuilder::with_application_region()` with a mandatory `RoutingStrategy` parameter on `build()`. Use `RoutingStrategy::ProximityTo(region)` to specify the application region. Also removed `CosmosClientOptions::with_application_region()`. ([#3889](https://github.com/Azure/azure-sdk-for-rust/pull/3889))
-- Renamed `replace_throughput` to `begin_replace_throughput` on `ContainerClient` and `DatabaseClient`. The return type changed from `ResourceResponse<ThroughputProperties>` to `ThroughputPoller`.
 - Changed `default_ttl` and `analytical_storage_ttl` fields on `ContainerProperties` from `Option<Duration>` to `TimeToLive`, a new enum with variants `Forever`, `NoDefault`, and `Seconds(u32)`, to correctly handle the `-1` wire value (TTL enabled with no default expiration).
 - `DatabaseClient::container_client()` now returns `azure_core::Result<ContainerClient>`, eagerly resolving container metadata (RID, partition key definition) at construction time. ([#4005](https://github.com/Azure/azure-sdk-for-rust/pull/4005))
 - `PartitionKeyDefinition` fields (`paths`, `kind`, `version`) are now private; use accessor methods `paths()`, `kind()`, and `version()` instead. `PartitionKeyKind` changed from a string newtype to an enum with variants `Hash`, `MultiHash`, and `Range`. `PartitionKeyVersion` is now an enum (`V1`, `V2`) instead of `Option<i32>`. ([#4005](https://github.com/Azure/azure-sdk-for-rust/pull/4005))


sdk/cosmos/azure_data_cosmos/src/clients/offers_client.rs
```diff
@@ -65,28 +65,74 @@ impl OffersClient {
...
+        throughput: ThroughputProperties,
+    ) -> azure_core::Result<ThroughputPoller> {
+        // Read the current offer to get the offer RID before replacing.
+        let current = self.read(context.clone()).await?;
+        let mut current_throughput = current.unwrap_or_default();
+        let offer_id = if current_throughput.offer_id.is_empty() {
+            None
+        } else {
+            Some(current_throughput.offer_id.clone())
+        };
```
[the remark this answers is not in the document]
tvaron3: Applied — replaced `unwrap_or_default()` with an explicit error when no offer is found, and added a guard for empty offer RID.

sdk/cosmos/azure_data_cosmos/src/clients/throughput_poller.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// Creates a poller for an operation that is still pending.
+    fn pending(
+        initial_response: CosmosResponse<ThroughputProperties>,
+        offers_client: OffersClient,
+        context: Context<'static>,
+        offer_id: Option<String>,
+    ) -> Self {
+        let polling_interval = DEFAULT_POLLING_INTERVAL;
+
+        let stream = futures::stream::unfold(
```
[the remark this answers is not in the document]
tvaron3: Skipping for now — the emulator integration test (`container_throughput_stream_polling` with 11000 RU) exercises the pending-path polling loop end-to-end. Adding a unit test with an `OffersClient` test double would require a trait refactor; happy to do that in a follow-up if needed.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -181,49 +182,64 @@ impl ContainerClient {
...
             reason = "The 'options' variable may be used in the future"
         )]
         let options = options.unwrap_or_default();
 
-        // We need to get the RID for the database.
-        let db = self.read(None).await?.into_model()?;
-        let resource_id = db
+        // We need to get the RID for the container.
+        let container = self.read(None).await?.into_model()?;
+        let resource_id = container
```
analogrelay: Can we get this from the driver? We should have eagerly resolved it now, right?

sdk/cosmos/azure_data_cosmos/src/clients/throughput_poller.rs
```diff
@@ -0,0 +1,279 @@
...
+//! [`ThroughputPoller`] is a custom type rather than using [`azure_core::http::Poller`], which
+//! expects URL-based [`StatusMonitor`](azure_core::http::poller::StatusMonitor) continuation.
+
+use crate::{
+    clients::offers_client,
+    constants,
+    models::{CosmosResponse, ResourceResponse, ThroughputProperties},
+};
+use azure_core::http::StatusCode;
+use azure_core::time::Duration as AzDuration;
```
analogrelay: Little weird to alias this when I don't see a conflicting symbol in the `use`s

sdk/cosmos/azure_data_cosmos/src/clients/throughput_poller.rs
```diff
@@ -0,0 +1,279 @@
...
+        offer_id: String,
+    ) -> Self {
+        let polling_interval = DEFAULT_POLLING_INTERVAL;
+
+        let stream = futures::stream::unfold(
+            Some(PollState::Initial(Box::new(initial_response))),
+            move |state| {
+                let driver = driver.clone();
+                let account = account.clone();
+                let offer_id = offer_id.clone();
```
analogrelay, line 113: I don't think these clones should be needed. We own the `AccountReference` and the `String` offer ID. Maybe I'm wrong though. Pretty harmless if we do need it, just raising it.

sdk/cosmos/azure_data_cosmos/tests/split_tests/cosmos_split_offers.rs
```diff
@@ -0,0 +1,60 @@
...
+
+            let container_client = run_context
+                .create_container(
+                    db_client,
+                    properties.clone(),
+                    Some(CreateContainerOptions::default().with_throughput(throughput)),
+                )
+                .await?;
+
+            // Use a high RU value (11000) to trigger async processing, which may take ~5 minutes
```
analogrelay, line 36: Oof, 5mins. Can we make sure this isn't going to block the rest of the test run? Maybe put this in it's own test category or at least ensure it's running in way that allows other tests to run in parallel? I'd like to make sure we're not making the full pipeline run 5mins longer, if possible.

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
tvaron3: Good call — `container_ref` is eagerly resolved at construction time, so we already have the RID. Removed the extra `self.read()` round-trip from both `read_throughput` and `begin_replace_throughput`.

sdk/cosmos/azure_data_cosmos/src/clients/throughput_poller.rs
tvaron3: Fixed — removed the alias, using `azure_core::time::Duration` directly.

sdk/cosmos/azure_data_cosmos/src/clients/throughput_poller.rs
tvaron3: The clones are needed because `futures::stream::unfold` calls the closure on each poll iteration — each invocation needs its own copies. `driver` is `Arc` so the clone is just a refcount bump. `account` and `offer_id` are cloned once per 5-second poll tick

sdk/cosmos/azure_data_cosmos/tests/split_tests/cosmos_split_offers.rs
tvaron3: Moved to a new `split` test category so it runs independently and won't block other tests. Created `tests/split_tests/` with `test_category = "split"`, intended to run on a single-write account. In other sdks, these are one of the common long running pipelines for split scenarios.

analogrelay: Sounds great. I definitely want us to test these scenarios, just want to make sure we're not expanding the test run time :).

sdk/cosmos/azure_data_cosmos/src/clients/throughput_poller.rs
analogrelay: I believe you could eliminate that by flowing them through the `PollState`. Make them fields on the state and make sure to _move_ (don't clone them) them to the next state when you return it. That means they're owned by the stream `unfold` creates (which lives for the whole stream), rather than by the closure context (which does indeed get created and destroyed for each call).

Having said that, you're right that the clones are cheap enough that I'm not worried if you can't get a clean solution.

PR #4132, opened 2026-04-09
[Storage] Replace body input type for managed upload

jaschrep-msft:
Replaces the Azure-standard `Body` type with `StorageUploadBody`, which accepts `Bytes` or an `AsyncReadWithLenHint` object (implements `AsyncRead + Send + ...` plus a `len()` method). 

## Usage
Because `AsyncReadWithLenHint` is blanket-implemented for `SeekableStream`, all existing implementations of `SeekableStream` work with `StorageUploadBody` just as smoothly as they do `Body`.

```rust
fn seekable_stream_convert<T>(stream: T) -> StorageUploadBody
where
    T: SeekableStream + 'static,
{
    StorageUploadBody::AsyncRead(Box::new(stream))
}
```

More to come

## To Do
- Convert `partitioned_transfer::upload()` to accept new type (waiting on #4107 to merge to avoid conflict)
- Write samples using arbitrary `AsyncRead` objects.

### DIFF
--- a/sdk/storage/azure_storage_blob/src/models/content_adapters.rs
+++ b/sdk/storage/azure_storage_blob/src/models/content_adapters.rs
@@ -1,9 +1,11 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+use std::future::Future;
+
 use azure_core::stream::SeekableStream;
 use bytes::Bytes;
-use futures::AsyncRead;
+use futures::{AsyncRead, AsyncReadExt};
 
 pub enum StorageUploadBody {
     Bytes(Bytes),
@@ -63,13 +65,15 @@ struct AsyncReadLenHintWrapper<T> {
     pub async_read: T,
     pub len_hint: Option<u64>,
 }
-impl<T: AsyncRead> AsyncRead for AsyncReadLenHintWrapper<T> {
+impl<T: AsyncRead + Unpin> AsyncRead for AsyncReadLenHintWrapper<T> {
     fn poll_read(
-        self: std::pin::Pin<&mut Self>,
-        _cx: &mut std::task::Context<'_>,
-        _buf: &mut [u8],
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+        buf: &mut [u8],
     ) -> std::task::Poll<std::io::Result<usize>> {
-        todo!()
+        let fut = self.async_read.read(buf);
+        futures::pin_mut!(fut);
+        fut.poll(cx)
     }
 }
 impl<T: AsyncRead + Send + Sync + Unpin> AsyncReadWithLenHint for AsyncReadLenHintWrapper<T> {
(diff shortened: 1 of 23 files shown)


sdk/storage/azure_storage_blob/src/models/content_adapters.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+impl From<Box<dyn AsyncReadWithLenHint>> for StorageUploadBody {
+    fn from(async_read: Box<dyn AsyncReadWithLenHint>) -> Self {
+        Self::AsyncRead(async_read)
+    }
+}
+
+pub trait AsyncReadWithLenHint: AsyncRead + Send + Sync + Unpin {
+    fn len(&self) -> Option<u64>;
```
vincenttran-msft, line 52: nit: Can't this be just -> `<u64>`? Cause this seems like the `WithLenHint` variant, whereas if it is not known, cx would use `AsyncRead`?

sdk/storage/azure_storage_blob/src/models/content_adapters.rs
```diff
@@ -0,0 +1,117 @@
...
+{
+    fn from(bytes: B) -> Self {
+        Self::Bytes(bytes.into())
+    }
+}
+
+impl From<&StorageUploadBody> for Bytes {
+    fn from(value: &StorageUploadBody) -> Self {
+        match value {
+            StorageUploadBody::Bytes(bytes) => bytes.clone(),
```
jalauzon-msft: Does `bytes.clone()` copy?

sdk/storage/azure_storage_blob/src/models/content_adapters.rs
jaschrep-msft: We don't accept a plain `AsyncRead` in this design. We accept this trait and leave it to implementors if they return Some or None. We then also implement a wrapper type that stores a normal `AsyncRead` along with a manually set `Option<u64>`, so that a customer with just some `AsyncRead` doesn't need a whole new impl block.

I think this will become clearer as I develop more e2e samples.

heaths (changes requested):
So you redefined `Body`. `Body` is an enum that already takes a `Bytes` or `SeekableStream` in which case I _already_ made `len()` return an `Option<u64>`. This is redundant, and unlike C# you can't simply try-cast traits in Rust, which puts a burden on devs using the API.

And this is also redundant with #4067 which we agreed to wait until after GA since the request came in so late.

sdk/storage/azure_storage_blob/src/models/content_adapters.rs
heaths: This isn't C#. You can't try-cast to a trait in Rust. Nothing close without boxing, even. We already have #4067 ready post-GA, which we discussed over teams. By adding another trait, now you're forcing yet another trait to implement on devs which, because of the orphan rule, makes it even harder to use and less idiomatic to the Azure SDK for Rust. We already made `len` optional. The rest will have to wait post-GA when we merge #4067 for all crates to use with a wrapper around `futures::io::AsyncRead`.

PR #4135, opened 2026-04-10
Allow azure_storage_blob to disable auto-decompression

heaths:
Congruent with other Azure SDK languages. We still leave it enabled for most services because, when payloads are large, it can help. Most languages do send `accept-encoding: gzip,deflate(,br)` already and auto-decompress on `content-encoding` having any one of those. .NET seems to be the exception currently.

### DIFF
--- a/sdk/core/typespec_client_core/src/http/clients/reqwest.rs
+++ b/sdk/core/typespec_client_core/src/http/clients/reqwest.rs
@@ -14,6 +14,13 @@ use tracing::{debug, warn};
 use typespec::error::{Error, ErrorKind, Result, ResultExt};
 
 /// Create a new [`HttpClient`] with the `reqwest` backend.
+///
+/// # Arguments
+///
+/// * `options` - Optional configuration for the [`Client`](::reqwest::Client).
+///   Automatic decompression is enabled if `reqwest_gzip` or `reqwest_deflate` are enabled.
+///   Client libraries can disable this without impacting other client libraries by disabling it
+///   when calling this function.
 #[cfg_attr(
     not(any(feature = "reqwest_gzip", feature = "reqwest_deflate")),
     allow(unused_variable)
--- a/sdk/core/typespec_client_core/src/http/clients/reqwest.rs
+++ b/sdk/core/typespec_client_core/src/http/clients/reqwest.rs
@@ -14,8 +14,9 @@ use tracing::{debug, warn};
 use typespec::error::{Error, ErrorKind, Result, ResultExt};
 
 /// Create a new [`HttpClient`] with the `reqwest` backend.
-pub fn new_reqwest_client() -> Arc<dyn HttpClient> {
+pub fn new_reqwest_client(options: Option<super::HttpClientOptions>) -> Arc<dyn HttpClient> {
     debug!("creating an http client using `reqwest`");
+    let options = options.unwrap_or_default();
 
     // Some customers in the past have reported challenges associated with enabling
     // connection pooling in reqwest. See <https://github.com/hyperium/hyper/issues/2312>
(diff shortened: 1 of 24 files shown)


sdk/storage/azure_storage_blob/src/clients/mod.rs
```diff
@@ -16,3 +18,13 @@ pub use blob_container_client::{BlobContainerClient, BlobContainerClientOptions}
 pub use blob_service_client::{BlobServiceClient, BlobServiceClientOptions};
 pub use block_blob_client::{BlockBlobClient, BlockBlobClientOptions};
 pub use page_blob_client::{PageBlobClient, PageBlobClientOptions};
+
+#[allow(clippy::needless_update)]
+fn apply_client_defaults(options: &mut ClientOptions) {
+    if options.transport.is_none() {
+        options.transport = Some(Transport::new(new_http_client(Some(HttpClientOptions {
```
jalauzon-msft, line 25: Do you ever anticipate the default impl for `Transport` getting anything interesting? We are now missing out on that default impl, always. So, if Core makes a change there, we'll need to remember to update this as well.

heaths: `new_http_client` we'll always keep up to date with anything "interesting", but this is the only way to avoid having you take an explicit dependency on `reqwest` and mimicking `reqwest*` features like we do in `azure_core` which not only makes it harder to consumer (and more confusing), but means you might one day miss out on some core changes without an update - unlikely, but possible.

heaths: To counter this, you should put something not only in your client docs like so:

```rust
/// <existing summary>
///
/// <existing description, if any>
///
/// # Notes
///
/// Clients will use the HTTP transport returned from [`azure_core::http::new_http_client()`]
/// but without automatic decompression by default. If you set a custom transport in [`<your ClientOptions>`]
/// without disabling automatic decompression, partitioned downloads may not succeed.
pub struct BlockBlobClient { /* .. */ }
```

And toward the bottom of your README, much like we have in `azure_core`'s README, something similar. Feel free to workshop the language, of course.

The main problem is that we don't know how to turn off decompression in general. Adding a setter to the `HttpClient` trait wouldn't work for `reqwest::Client` because it has to be set on a `reqwest::ClientBuilder`. Once a `Client` is built from the builder, it cannot be changed. And that could be true for any number of custom HTTP clients.

We would need to revamp the entire HTTP stack and, at this point, that's not feasible.

sdk/core/typespec_client_core/src/http/clients/reqwest.rs
```diff
@@ -14,22 +14,37 @@ use tracing::{debug, warn};
...
+    let mut builder = ::reqwest::ClientBuilder::new()
         // By default, reqwest will chase 3xx redirects up to 10 links. REST API guidelines
         // discourage services from using 3xx redirects, so disabling the reqwest redirect logic
         // simplifies the client logic.
-        .redirect(::reqwest::redirect::Policy::none())
-        .build()
-        .expect("failed to build `reqwest` client");
+        .redirect(::reqwest::redirect::Policy::none());
+
+    #[cfg(feature = "reqwest_gzip")]
```
LarryOsterman, line 44: This is a subtle breaking change from the older `reqwest_gzip` and `reqwest_deflate` features. Previously they unilaterally enabled those features, now the use of those features is controlled by the `.automatic_decompression` option flag.
 
Do we want to keep the original behavior? I'm ok either way, but if we want to keep the semantic change (which *is* reasonable) then updating the readme and changelog to express this change is probably a good idea.

heaths: In practice, this shouldn't change. Customer shouldn't be calling this, only partner crates. But I definitely should document the API and may as well update the `CHANGELOG` change description I used to make that more clear.

LarryOsterman (approved):

LarryOsterman (approved):

LarryOsterman (approved):

analogrelay (approved):
Cosmos changes LGTM

chlowell (approved):

LarryOsterman (approved):

vincenttran-msft (approved):
Thanks for this, I will re-record those other tests once this lands.

jalauzon-msft (approved):

PR #4145, opened 2026-04-10
[Storage] `HttpRange` handwritten model

vincenttran-msft:
Adds `HttpRange` in place of our `format_page_range` helper parser, we are hoping to get typespec `@@alternateType` support pre-GA to remove the necessity of the `into()`.
`.tsp:` https://github.com/Azure/azure-rest-api-specs/pull/42277

### DIFF
--- a/sdk/storage/azure_storage_blob/CHANGELOG.md
+++ b/sdk/storage/azure_storage_blob/CHANGELOG.md
@@ -5,8 +5,11 @@
 ### Features Added
 
 - Added `stream::tokio` module (gated on the `tokio` feature) with `FileStream` and `FileStreamBuilder` for streaming file uploads.
+- Added `models::HttpRange` for specifying byte ranges in blob operations, replacing the removed `format_page_range()` helper.
 
 ### Breaking Changes
+
+- Removed `format_page_range()`. Use `HttpRange::new(offset, length)` or `HttpRange::from_offset(offset)` instead.
 
 - Revised `download()` on `BlobClient` with the following breaking changes:
   - Now uses managed (multi-part) download logic for optimal performance on single-shot and parallel range transfers.
--- a/sdk/storage/azure_storage_blob/src/models/http_ranges.rs
+++ b/sdk/storage/azure_storage_blob/src/models/http_ranges.rs
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 use azure_core::error::{Error, ErrorKind, ResultExt};
-use azure_core::http::headers::{Header, HeaderName};
+use azure_core::http::headers::{Header, HeaderName, HeaderValue};
 use std::fmt;
 use std::ops::{Range, RangeFrom};
 use std::str::FromStr;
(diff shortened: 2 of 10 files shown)


sdk/storage/azure_storage_blob/CHANGELOG.md
```diff
@@ -5,9 +5,12 @@
 ### Features Added
 
 - Added `stream::tokio` module (gated on the `tokio` feature) with `FileStream` and `FileStreamBuilder` for streaming file uploads.
+- Added `models::HttpRange` for specifying byte ranges in blob operations, replacing the removed `format_page_range()` helper.
 
 ### Breaking Changes
 
+- Removed `format_page_range()`. Use `HttpRange::new(offset, length)` or `HttpRange::from_offset(offset)` instead.
```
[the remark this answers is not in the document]
vincenttran-msft: Nah

sdk/storage/azure_storage_blob/src/models/http_ranges.rs
```diff
@@ -188,3 +188,145 @@ mod tests {
...
+    /// # Arguments
+    ///
+    /// * `offset` - The starting byte offset.
+    /// * `length` - The number of bytes in the range.
+    pub fn new(offset: u64, length: u64) -> Self {
+        Self {
+            offset,
+            length: Some(length),
+        }
+    }
```
heaths, line 233: Why a constructor and public fields? I'd do one or the other. Probably just fields, which makes it more like our models.

sdk/storage/azure_storage_blob/src/models/http_ranges.rs
```diff
@@ -188,3 +188,145 @@ mod tests {
...
+            None => write!(f, "bytes={}-", self.offset),
+        }
+    }
+}
+
+impl From<HttpRange> for String {
+    fn from(range: HttpRange) -> Self {
+        range.to_string()
+    }
+}
```
heaths: Is this necessary? It's not common in Rust, especially since - and you even use it - implementing `Display` gives you a free `to_string()` (blanket `ToString` impl).

vincenttran-msft: I think `into()` is better rather than `to_string()` in the unlikely case that again these ergonomics stick around (we are going to alternate type the model in after typespec support drops). And in fact this allows them to choose either!

sdk/storage/azure_storage_blob/src/models/http_ranges.rs
vincenttran-msft: Went with private fields, I like the ergonomics of it better than struct-initialization (in the case this doesn't eventually all go away with the hanging typespec support)

jalauzon-msft: Agreed functions are better here, it lets the customer know what their options are (set both, only set offset, but can't just set length)

sdk/storage/azure_storage_blob/src/models/http_ranges.rs
jalauzon-msft: I think I agree with Heath here and `ToString` via `Display` is good enough. But if you feel strongly, not the end of the world to keep it.

jalauzon-msft (approved):

sdk/storage/azure_storage_blob/src/models/http_ranges.rs
vincenttran-msft: Offline discussions led to this being removed as well.

jalauzon-msft (approved):

PR #4146, opened 2026-04-10
cosmos: fix #4140 by providing features for rustls

### LINKED ISSUE #4140: Cosmos: Allow disabling native TLS in reqwest
When we used the `azure_core` stack, it was possible to disable the `native-tls` stack and `reqwest`'s integration with that stack. This would allow the user to plug in a different TLS stack. We should add the same feature-flags to the Cosmos SDK and Driver so that a user can disable default features and avoid bringing a specific TLS stack in.

analogrelay:
Closes #4140 

This PR gives users a little more control over the TLS stack used by the Cosmos SDK. When we switched to using our own pipeline, we enabled `azure_data_cosmos_driver/reqwest_native_tls` in the dependency from the SDK to the driver, which means users were _forced_ to get it. This prevents users from using other TLS stacks like `rustls`. To fix this, `azure_data_cosmos` no longer sets _any_ feature flags on `azure_data_cosmos_driver` in its dependency (`default-features = false`) and uses it's own default and opt-in features to allow the user to specify the TLS stack they're using. The `reqwest_native_tls`, `reqwest_rustls` and `reqwest_default_tls` features on `azure_data_cosmos` allow the user to do this.

### Why add features for `rustls`?

We could have achieved this by just putting the use of `native-tls` behind a feature flag, and enabled that flag on the default feature. Then, a user who wanted to use `rustls` could just disable our default feature and reference `reqwest = { version = ..., features = ["rustls"]`. Cargo unions features across all dependencies, so this would have the effect of switching `reqwest` to use `rustls` out from under us. However, we have the `allow_invalid_certificates` feature used to allow the emulator's self-signed certificate. That logic requires calling [`ClientBuilder::tls_danger_accept_invalid_certs`](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.tls_danger_accept_invalid_certs) which can only be done if we KNOW one of reqwest's TLS-related features is set. So, we need our own feature that we can use to protect our code that allows enabling that.

There **is** an alternative to this approach. We could allow customizing the `reqwest::ClientBuilder` directly. Since the _user's app_ has enabled `rustls` or some other TLS backend, they can safely call `tls_danger_accept_invalid_certs` on it. But this means exposing the `reqwest` pipeline directly, which we've been very hesitant to do given the major impact that can have on SLAs.

### What about `default-tls`?

The `default-tls` feature of `reqwest` just forwards to `rustls`. We actually want our default to be `native-tls`, so it doesn't make sense to express reqwests `default-tls` as a proxy feature. Instead, we might as well just have `native-tls` and `rustls` as separate features.


deny.toml
```diff
@@ -11,10 +11,11 @@ allow = [
   "Apache-2.0",
   "BSD-3-Clause",
   "MIT",
-  "MPL-2.0",
   # "OpenSSL",
   "Unicode-3.0",
   "Zlib",
+  "CDLA-Permissive-2.0",
+  "ISC",
```
analogrelay, line 18: This is necessary because I'm adding optional dependencies on rustls. cc @heaths .

sdk/cosmos/azure_data_cosmos/Cargo.toml
```diff
@@ -40,14 +40,20 @@ tracing-subscriber = { workspace = true, features = ["env-filter", "fmt"] }
...
-default = ["hmac_rust", "reqwest"]
-reqwest = ["dep:reqwest", "azure_core/reqwest", "azure_core/tokio", "azure_data_cosmos_driver/reqwest_native_tls"]
+default = ["hmac_rust", "reqwest", "reqwest_native_tls"]
+reqwest = ["dep:reqwest", "azure_core/reqwest", "azure_core/tokio", "azure_data_cosmos_driver/reqwest", "azure_data_cosmos_driver/tokio"]
+reqwest_native_tls = [
+  "reqwest",
+  "azure_core/reqwest_native_tls",
+  "azure_data_cosmos_driver/reqwest_native_tls",
+]
+reqwest_rustls = ["reqwest", "azure_data_cosmos_driver/reqwest_rustls"]
```
heaths: 1. We don't want parity features. We intentionaly removed this from `azure_core` a while ago and document how to enable it; _however_
2. We're going to flip this to the default it seems.

That said, we don't want a `reqwest_native_tls` parity feature. They can get that by simply adding `reqwest` themselves with whatever features they want.

heaths (changes requested):

sdk/cosmos/azure_data_cosmos/Cargo.toml
analogrelay: As mentioned offline, we _do_ need to know the TLS backend in use, it's more than just a parity feature here. I'll rename it to make that clearer though. Plus, I'll swap our default to align with the core's.

analogrelay:
Apologies for the over-ping, I think I had a bad merge. I'll close this and open a new one anyway since I've changed some logic here a bit.

PR #4154, opened 2026-04-13
[Internal] Update rand 0.10.0 to 0.10.1 for security advisory

tvaron3:
rand v0.10.0 has a RUSTSEC security advisory causing `cargo deny --all-features check` to fail with `advisories FAILED` in CI for all PRs targeting `release/azure_data_cosmos-previews`.

Updates `Cargo.lock` to use rand v0.10.1 which resolves the advisory.

Single lockfile change — no code changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -340,7 +340,7 @@ dependencies = [
  "dotenvy",
  "flate2",
  "futures",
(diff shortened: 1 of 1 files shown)


FabianMeiswinkel (approved):
LGTM

PR #4156, opened 2026-04-13
Cosmos: Implement PPAF and PPCB in Driver Crate

### LINKED ISSUE #3742: Cosmos: Partition-Level Failover
### Feature Summary

Add Per-Partition Automatic Failover and Circuit Breakers to Driver

### Feature Description

- Move the implementation of PPAF + PPCB into the driver
- Consider refactoring the code to better align with rust idioms
- Allow the SDK via feature-flag opt-in to use the driver instead of its own implementation for testing purposes

Out-of-scope:
- Original code in SDK still exists - and is used by default (unless opted for driver via feature flag)

### Use Case

_No response_

### Alternatives

_No response_

### Additional Context

_No response_

kundadebdatta:
### Summary

Implements partition-level failover in the `azure_data_cosmos_driver` crate, replacing the account-level-only failover with fine-grained per-partition routing. This provides two complementary mechanisms:

- **PPAF (Per-Partition Automatic Failover)** — for writes on single-master accounts. On 403/3, 503, 429/3092, or 410/1022 from a region, the affected partition is unconditionally failed over to the next write region. Subsequent writes to the same partition skip the failed region entirely.
- **PPCB (Per-Partition Circuit Breaker)** — for reads on any account and writes on multi-master accounts. Tracks failure counts per `(PartitionKeyRangeId, Region)` and, once a configurable threshold (default 10 for reads, 5 for writes) is exceeded, routes requests directly to an alternate region. A background failback loop gradually restores routing via probe-based recovery.

The implementation spans the driver's routing, retry evaluation, operation pipeline, caching, background task, and configuration subsystems. Integration tests using the fault injection framework validate all scenarios across both the driver and SDK crates. A Cargo.toml bugfix is also included (the `reqwest` feature was not enabling the crate's `tokio` feature, silently compiling out the failback loop).

### Motivation

Cosmos DB accounts span multiple regions. When a single partition becomes unhealthy, marking the **entire region** unavailable forces all requests — including those targeting healthy partitions — to cross-region, adding unnecessary latency. Partition-level failover provides:

1. **Finer-grained fault isolation** — only the affected partition is rerouted; healthy partitions in the same region continue to be served locally.
2. **Faster recovery** — each partition is tracked independently, and the background failback loop restores routing as soon as the original region recovers.
3. **Multi-master write support** — PPCB enables partition-level failover for both reads and writes on multi-write accounts.

#### Core Routing Subsystem (`driver/routing/`)

| File | Purpose |
|------|---------|
| `partition_endpoint_state.rs` | Immutable CAS-swapped state: two maps (`failover_overrides` for PPAF, `circuit_breaker_overrides` for PPCB) keyed by `PartitionKeyRangeId`, each holding a `PartitionFailoverEntry` with failure counters, timestamps, failed endpoints, current override endpoint, and `HealthStatus` |
| `partition_key_range_id.rs` | Newtype wrapper `PartitionKeyRangeId` around `String` with `FromStr`, `Display`, `Borrow<str>` |
| `routing_systems.rs` | Pure functions: `is_eligible_for_ppaf()`, `is_eligible_for_ppcb()`, `can_circuit_breaker_trigger_failover()`, `mark_partition_unavailable()`, `expire_partition_overrides()`, `remove_probe_succeeded_entry()` |
| `location_effects.rs` | New `LocationEffect::MarkPartitionUnavailable` variant carrying PK range ID, region, read/write flag |
| `location_state_store.rs` | Holds `Atomic<PartitionEndpointState>` alongside existing `Atomic<AccountEndpointState>`. New methods: `apply_partition()` (CAS loop), `sync_account_properties()` (updates PPAF/PPCB flags from `AccountProperties`), `start_failback_loop()` |
| `account_endpoint_state.rs` | Extended with `multiple_write_locations_enabled` flag |
| `endpoint.rs` | `CosmosEndpoint` with generation-tagged `LocationIndex` for stale-detection |

#### Retry/Pipeline (`driver/pipeline/`)

| File | Purpose |
|------|---------|
| `retry_evaluation.rs` | `evaluate_transport_result()` — pure function emitting `MarkPartitionUnavailable` effects for 403/3, 503, 429/3092, 410/1022. Suppresses `MarkEndpointUnavailable` when `is_ppcb_managed()` (failure tracked at partition level instead). Sets `ppaf_write_retry_allowed` for non-idempotent writes on PPAF-enabled accounts |
| `operation_pipeline.rs` | 7-stage loop: (1) snapshot → (2) `resolve_endpoint()` with partition override → (3) build request → (4) transport → (5) evaluate → (6) apply effects → (7) act on action. New: partition key range ID pre-resolution, PPAF/PPCB override in `resolve_endpoint()`, probe routing for `ProbeCandidate` entries |
| `components.rs` | `OperationRetryState` extended with `partition_key_range_id`, `ppaf_write_retry_allowed`, `ppcb_active` |

#### Background Failback

| File | Purpose |
|------|---------|
| `transport/background_task_manager.rs` | Manages spawned `JoinHandle`s with panic isolation (`catch_unwind`), prunes finished tasks, abort-on-drop |
| Failback loop in `location_state_store.rs` | Periodic sweep (configurable interval): PPCB entries whose `first_failure_time + unavailability_duration` has elapsed transition from `Unhealthy → ProbeCandidate`. PPAF entries are intentionally NOT swept. Self-terminates when `LocationStateStore` drops (Weak ref) |

#### Caching (`driver/cache/`)

| File | Purpose |
|------|---------|
| `partition_key_range_cache.rs` | `AsyncCache<ContainerReference, ContainerRoutingMap>`. `resolve_partition_key_range_id()` computes effective partition key → binary search → `PartitionKeyRangeId`. Supports MultiHash prefix resolution |
| `container_routing_map.rs` | Sorted `PartitionKeyRange` entries with O(log n) binary search, overlap/gap validation |
| `container_cache.rs` | Dual-index cache (by-name and by-RID) for `ContainerReference` |

#### Driver Wiring

| File | Purpose |
|------|---------|
| `cosmos_driver.rs` | Builds `PartitionFailoverConfig::from_options()`, creates `LocationStateStore`, calls `start_failback_loop()`. New `pre_resolve_partition_key_range_id()` method before pipeline entry |
| `runtime.rs` | Per-runtime singleton holding caches and bootstrap transport |

#### Configuration (`options/`)

| File | Purpose |
|------|---------|
| `runtime_options.rs` | `PartitionFailoverConfig` — all thresholds/windows env-var overridable: `circuit_breaker_failure_count_for_reads` (default 10), `circuit_breaker_failure_count_for_writes` (default 5), `circuit_breaker_timeout_counter_reset_window_in_minutes` (default 5), `allowed_partition_unavailability_duration_in_seconds` (default 5), `per_partition_circuit_breaker_enabled` (default true) |

#### Fault Injection Framework

| File | Layer | Purpose |
|------|-------|---------|
| `driver/fault_injection/*.rs` | Driver | `FaultInjectionRule`, `FaultInjectionCondition` (operation type + region + container filter), `FaultInjectionResult` (error type + delay + probability), `FaultClient` (HttpClient wrapper), `FaultInjectionEvaluation` (rich enum for why each rule applied/skipped), `FaultInjectingHttpClientFactory` |
| `sdk/fault_injection/*.rs` | SDK | `FaultInjectionClientBuilder`, SDK-side condition/result/rule types. Bridge in `driver_bridge.rs` translates SDK rules to driver rules with shared `enabled`/`hit_count` state via `Arc` |

#### SDK Integration

| File | Purpose |
|------|---------|
| `routing/global_partition_endpoint_manager.rs` | SDK's existing (older) PPAF/PPCB implementation using `RwLock<HashMap>` — will be retired when SDK fully cuts over to `driver.execute_operation()` |
| `retry_policies/client_retry_policy.rs` | SDK's pipeline policy referencing `GlobalPartitionEndpointManager` — also part of the older architecture |
| `driver_bridge.rs` | SDK↔driver translation: `sdk_fi_rules_to_driver_fi_rules()`, `driver_response_to_cosmos_response()` |

#### Bug Fix

| File | Change |
|------|--------|
| `azure_data_cosmos_driver/Cargo.toml` | Added `"tokio"` to the `reqwest` feature dependency list so that `#[cfg(feature = "tokio")]`-gated code (BackgroundTaskManager, failback loop) is compiled when using the reqwest transport |

#### Integration Tests

| File | Crate | Tests |
|------|-------|-------|
| `azure_data_cosmos_driver/tests/emulator_tests/driver_partition_failover.rs` | Driver | 8 tests |
| `azure_data_cosmos/tests/emulator_tests/cosmos_partition_failover.rs` | SDK | 8 tests |

#### Spec & Docs

| File | Description |
|------|-------------|
| `docs/PARTITION_LEVEL_FAILOVER_SPEC.md` | Full design spec covering goals, architecture, eligibility rules, component design, failover/circuit-breaker flows, pipeline integration, background failback, status code handling matrix, configuration surface |
| `docs/TRANSPORT_PIPELINE_SPEC.md` | Transport pipeline spec (inner retry loop, 429 handling, deadline enforcement) |

### Test Scenarios

All 4 scenarios tested with both 503 (ServiceUnavailable) and 403/3 (WriteForbidden) error variants, in both driver and SDK crates (16 tests total):

| # | Scenario | Error | Expected Behavior |
|---|----------|-------|-------------------|
| 1a | PPAF enabled + CreateItem | 503 | Partition fails over to next write region; write succeeds |
| 1b | PPAF enabled + CreateItem | 403/3 | Partition fails over to next write region; write succeeds |
| 2a | PPAF disabled + CreateItem | 503 | No partition failover; non-idempotent write fails |
| 2b | PPAF disabled + CreateItem | 403/3 | Cross-regional retry succeeds but no sticky partition override |
| 3a | PPCB enabled + ReadItem | 503 | After threshold (default 10), reads routed directly to alternate region |
| 3b | PPCB enabled + ReadItem | 403/3 | After threshold (default 10), reads routed directly to alternate region |
| 4a | PPCB disabled + ReadItem | 503 | No circuit breaker; hub region contacted every time, cross-regional retry each read |
| 4b | PPCB disabled + ReadItem | 403/3 | No circuit breaker; hub region contacted every time, cross-regional retry each read |

#### High-Level: Operation Pipeline with Partition-Level Failover

```mermaid
flowchart TB
    subgraph PrePipeline["Pre-Pipeline (CosmosDriver)"]
        A[execute_operation] --> B[sync_account_properties]
        B --> C{PPAF or PPCB enabled<br/>+ partitioned resource<br/>+ has PK?}
        C -->|Yes| D[pre_resolve_partition_key_range_id<br/>via PartitionKeyRangeCache]
        C -->|No| E[Skip pre-resolution]
        D --> F[execute_operation_pipeline]
        E --> F
    end

    subgraph Pipeline["7-Stage Operation Pipeline"]
        F --> S1["Stage 1: Acquire LocationSnapshot<br/>(account + partition state)"]
        S1 --> S2["Stage 2: resolve_endpoint()<br/>Account-level selection<br/>→ Partition-level override check"]
        S2 --> S3["Stage 3: Build TransportRequest"]
        S3 --> S4["Stage 4: Execute transport pipeline<br/>(429 retry, signing, deadline)"]
        S4 --> S5["Stage 5: evaluate_transport_result()<br/>→ (OperationAction, Vec of LocationEffect)"]
        S5 --> S6["Stage 6: location_state_store.apply(effects)<br/>CAS on AccountEndpointState<br/>CAS on PartitionEndpointState"]
        S6 --> S7["Stage 7: Act on action"]
        S7 -->|FailoverRetry| S1
        S7 -->|SessionRetry| S1
        S7 -->|Complete| RES[Return response]
        S7 -->|Abort| ERR[Return error]
    end

    subgraph State["LocationStateStore (lock-free CAS)"]
        AS["AccountEndpointState<br/>• preferred endpoints<br/>• unavailable endpoints<br/>• multi-write flag"]
        PS["PartitionEndpointState<br/>• failover_overrides (PPAF)<br/>• circuit_breaker_overrides (PPCB)<br/>• PPAF/PPCB enabled flags<br/>• PartitionFailoverConfig"]
    end

    subgraph Background["Background Failback"]
        BG["Periodic sweep task<br/>(Weak ref, self-terminating)"]
        BG -->|"PPCB: Unhealthy → ProbeCandidate<br/>after unavailability duration"| PS
    end

    S1 -.->|snapshot| AS
    S1 -.->|snapshot| PS
    S6 -.->|CAS update| AS
    S6 -.->|CAS update| PS
```

#### PPAF vs PPCB: Eligibility and Mutual Exclusivity

```mermaid
flowchart LR
    REQ[Incoming Request] --> CHK{Operation type?}

    CHK -->|Write| WR{Account type?}
    CHK -->|Read| RD[PPCB path]

    WR -->|Single-master| SM{PPAF enabled?}
    WR -->|Multi-master| MM[PPCB path]

    SM -->|Yes| PPAF["PPAF path<br/>• Unconditional failover<br/>• No threshold<br/>• No failback sweep<br/>• Sticky until backend changes"]
    SM -->|No| NOFO["No partition failover<br/>Account-level retry only"]

    RD --> PPCB_CHK{PPCB enabled?}
    MM --> PPCB_CHK

    PPCB_CHK -->|Yes| PPCB["PPCB path<br/>• Threshold-gated<br/>• Counter tracking<br/>• Background failback<br/>• Probe-based recovery"]
    PPCB_CHK -->|No| NOFO2["No partition failover<br/>Cross-regional retry only"]
```

#### PPCB Circuit Breaker State Machine

```mermaid
stateDiagram-v2
    [*] --> Healthy: No failures tracked

    Healthy --> CountingFailures: Request fails (503/403.3/429.3092/410)
    CountingFailures --> CountingFailures: failure_count < threshold
    CountingFailures --> Unhealthy: failure_count >= threshold

    Unhealthy --> Unhealthy: All requests bypass hub,<br/>routed to alternate region

    Unhealthy --> ProbeCandidate: Background sweep:<br/>first_failure_time + unavailability_duration elapsed

    ProbeCandidate --> Healthy: Probe request succeeds<br/>(remove_probe_succeeded_entry)
    ProbeCandidate --> Unhealthy: Probe request fails<br/>(reset counters, restart clock)

    note right of Unhealthy
        Requests skip the failed region entirely.
        No retry latency penalty.
    end note

    note right of ProbeCandidate
        Single user request tentatively routed
        back to original region as a probe.
        TOCTOU-safe via CAS re-check.
    end note
```

#### PPAF Write Failover Flow (Scenarios 1 & 2)

```mermaid
sequenceDiagram
    participant App as Application
    participant SDK as SDK / Driver
    participant Hub as Hub Region (East US 2)
    participant AltR as Alternate Region (West US 3)

    Note over App,AltR: Scenario 1: PPAF Enabled — 503/403.3 on CreateItem

    App->>SDK: CreateItem(partition_key, item)
    SDK->>Hub: POST /docs (CreateItem)
    Hub-->>SDK: 503 ServiceUnavailable / 403.3 WriteForbidden

    Note over SDK: PPAF enabled → partition failover triggered
    SDK->>AltR: POST /docs (CreateItem) [retry to alternate region]
    AltR-->>SDK: 201 Created
    SDK->>SDK: Record partition-level override for this PK range
    SDK-->>App: Success (201 Created)

    Note over App,AltR: Subsequent writes to same partition skip hub
    App->>SDK: CreateItem(same partition)
    SDK->>AltR: POST /docs [routed directly via override]
    AltR-->>SDK: 201 Created
    SDK-->>App: Success
```

```mermaid
sequenceDiagram
    participant App as Application
    participant SDK as SDK / Driver
    participant Hub as Hub Region (East US 2)
    participant AltR as Alternate Region (West US 3)

    Note over App,AltR: Scenario 2a: PPAF Disabled — 503 on CreateItem

    App->>SDK: CreateItem(partition_key, item)
    SDK->>Hub: POST /docs (CreateItem)
    Hub-->>SDK: 503 ServiceUnavailable

    Note over SDK: PPAF disabled + non-idempotent → cannot safely retry
    SDK-->>App: Error (503 ServiceUnavailable)
```

```mermaid
sequenceDiagram
    participant App as Application
    participant SDK as SDK / Driver
    participant Hub as Hub Region (East US 2)
    participant AltR as Alternate Region (West US 3)

    Note over App,AltR: Scenario 2b: PPAF Disabled — 403/3 on CreateItem

    App->>SDK: CreateItem(partition_key, item)
    SDK->>Hub: POST /docs (CreateItem)
    Hub-->>SDK: 403.3 WriteForbidden

    Note over SDK: 403/3 always triggers cross-region retry (regardless of PPAF)
    SDK->>AltR: POST /docs (CreateItem) [failover retry]
    AltR-->>SDK: 201 Created
    SDK-->>App: Success (201 Created)

    Note over SDK: No partition-level override stored (PPAF off)

    Note over App,AltR: Next write to same partition still goes to hub first
    App->>SDK: CreateItem(same partition)
    SDK->>Hub: POST /docs [hub attempted first — no override]
    Hub-->>SDK: 201 Created (fault rule disabled)
    SDK-->>App: Success
```

#### PPCB Read Circuit Breaker Flow (Scenarios 3 & 4)

```mermaid
sequenceDiagram
    participant App as Application
    participant SDK as SDK / Driver
    participant Hub as Hub Region (East US 2)
    participant AltR as Alternate Region (West US 3)
    participant CB as Circuit Breaker

    Note over App,AltR: Scenario 3: PPCB Enabled — 503/403.3 on ReadItem

    loop Reads 1–10 (below threshold)
        App->>SDK: ReadItem(pk, id)
        SDK->>Hub: GET /docs/id
        Hub-->>SDK: 503 / 403.3
        SDK->>CB: Increment failure counter for PK range
        SDK->>AltR: GET /docs/id [cross-regional retry]
        AltR-->>SDK: 200 OK
        SDK-->>App: Success (with retry latency)
    end

    Note over CB: Failure threshold (10) exceeded → circuit OPEN

    loop Reads 11+ (above threshold)
        App->>SDK: ReadItem(pk, id)
        SDK->>CB: Check PK range status
        CB-->>SDK: Circuit OPEN → bypass hub
        SDK->>AltR: GET /docs/id [routed directly]
        AltR-->>SDK: 200 OK
        SDK-->>App: Success (no retry latency)
    end
```

```mermaid
sequenceDiagram
    participant App as Application
    participant SDK as SDK / Driver
    participant Hub as Hub Region (East US 2)
    participant AltR as Alternate Region (West US 3)

    Note over App,AltR: Scenario 4: PPCB Disabled — 503/403.3 on ReadItem

    loop Every read (no circuit breaker)
        App->>SDK: ReadItem(pk, id)
        SDK->>Hub: GET /docs/id [always attempted first]
        Hub-->>SDK: 503 / 403.3
        SDK->>AltR: GET /docs/id [cross-regional retry]
        AltR-->>SDK: 200 OK
        SDK-->>App: Success (always incurs retry latency)
    end

    Note over App,AltR: Hub is contacted on EVERY read — no bypass
```

### Decision Flow: How the Driver Handles a Failed Request

```mermaid
flowchart TD
    A[Request fails with error] --> B{Error type?}

    B -->|503 ServiceUnavailable| C{Is PPAF enabled?}
    B -->|403/3 WriteForbidden| D[Always trigger cross-region retry]

    C -->|Yes| E{Is operation idempotent or write?}
    C -->|No| F{Is operation a write?}

    E -->|Write| G[Partition failover → retry on alternate region]
    E -->|Read| H{Is PPCB enabled?}

    F -->|Write, non-idempotent| I[Abort — cannot safely retry]
    F -->|Read| J{Is PPCB enabled?}

    H -->|Yes| K[Increment failure counter]
    H -->|No| L[Cross-regional retry, no state stored]

    J -->|Yes| K
    J -->|No| L

    K --> M{Threshold exceeded?}
    M -->|Yes| N[Open circuit breaker for PK range]
    M -->|No| L

    N --> O[Future reads bypass hub → route directly to alternate]

    G --> P[Store partition-level override]
    P --> Q[Future ops for this PK range skip hub]

    D --> R[Retry on alternate region]
    R --> S{Is PPAF enabled?}
    S -->|Yes| P
    S -->|No| T[No partition override stored]
```

### Key Design Patterns

1. **Lock-free immutable snapshots** — All routing state uses `crossbeam_epoch::Atomic<T>` with CAS loops. Reads are lock-free (epoch pin + deref). Writes clone-mutate-swap. Zero reader/writer contention on the hot path.

2. **Pure function effects** — `evaluate_transport_result()` and all `routing_systems` functions are pure: state in → new state + effects out. No side effects. Trivially unit-testable.

3. **Circuit breaker state machine** — `Unhealthy → (background sweep) → ProbeCandidate → (single probe request) → Healthy` or `→ Unhealthy (reset counters)`. The probe is a real user request tentatively routed back to the original region, with TOCTOU protection via CAS re-check.

4. **PPAF vs PPCB separation** — Two distinct maps prevent cross-contamination between single-master write failover (service-directed, permanent until backend changes) and threshold-gated circuit breaking (probe-based gradual recovery).

5. **Pre-resolution of PK range ID** — Before the first attempt, `pre_resolve_partition_key_range_id()` uses the `PartitionKeyRangeCache` so partition-level overrides take effect from the very first request, not just after the first retry captures the ID from response headers.

6. **Weak-reference background tasks** — The failback loop holds `Weak<LocationStateStore>`, self-terminating when the store drops. `BackgroundTaskManager::drop()` aborts tasks as a safety net.

### Prerequisites for Running Integration Tests

- 3-region single-master Cosmos DB account with session consistency
- PPAF enabled on the account (for scenarios 1 & 3)
- Default PPCB thresholds (read failure threshold = 10, write failure threshold = 5)
- Hub region: **East US 2**, Satellite region: **West US 3**
- Tests gated by `#[cfg_attr(not(test_category = "emulator"), ignore)]`
- To run: `RUSTFLAGS='--cfg test_category="emulator"' cargo test -p <crate> --all-features`

### Assertion Strategy

| Layer | Assertions Available |
|-------|---------------------|
| **Driver** (`azure_data_cosmos_driver`) | `DiagnosticsContext::request_count()`, `DiagnosticsContext::regions_contacted()`, `FaultInjectionRule::hit_count()`, success/failure |
| **SDK** (`azure_data_cosmos`) | `ItemResponse::status()`, `FaultInjectionRule::hit_count()`, `rule.disable()` for multi-phase tests, success/failure |

The SDK's `CosmosDiagnostics` currently exposes only `activity_id` and `server_duration_ms` (no region info), so SDK tests rely on outcome-based assertions (status codes, hit counts) rather than region routing details.

### Validation

- `cargo fmt` — clean
- `cargo build` — clean for both crates
- `cargo clippy --all-features --all-targets` — clean (no new warnings)
- `cargo doc --no-deps` — clean
- `cargo test` — all 16 new tests compile and are properly recognized as ignored (requires emulator)
- All 724 driver unit tests pass
- All 462 SDK unit tests pass
- All 39 existing driver integration tests pass


sdk/cosmos/azure_data_cosmos_driver/Cargo.toml
```diff
@@ -69,6 +69,7 @@ reqwest = [
   "reqwest/http2",
   "azure_core/reqwest",
   "azure_core/tokio",
+  "tokio",
```
kundadebdatta, line 72: **Note:** The `reqwest` feature should also enable this crate's `tokio` feature, since reqwest itself depends on the tokio runtime and the `BackgroundTaskManager` uses `tokio::spawn` directly.

sdk/cosmos/azure_data_cosmos_driver/tests/multi_region_tests/driver_partition_failover.rs
```diff
@@ -0,0 +1,609 @@
...
+            diagnostics.request_count()
+        );
+
+        // Disable the rule so the next write can succeed without faults.
+        rule.disable();
+
+        // Without PPAF, there is no partition-level override, so the next write
+        // should go back to the hub region (because the endpoint was only marked
+        // unavailable for WriteForbidden, which is ignored for reads and expires
+        // per the TTL).
```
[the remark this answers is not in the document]
kundadebdatta: Updated the code comment.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -103,6 +111,27 @@ pub(crate) async fn execute_operation_pipeline(
...
+    retry_state.partition_key_range_id = pre_resolved_pk_range_id;
+
+    // PPAF write-retry: on single-master accounts with per-partition automatic
+    // failover enabled, non-idempotent writes may be retried to a different
+    // region for write region discovery (the service has already failed the
+    // partition over, so the original region will reject the write).
+    retry_state.ppaf_write_retry_allowed = location_snapshot
+        .partitions
+        .per_partition_automatic_failover_enabled
+        && !location_snapshot.account.multiple_write_locations_enabled;
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/partition_endpoint_state.rs
```diff
@@ -0,0 +1,208 @@
...
+            .copied()
+            .unwrap_or(false);
+
+        Self {
+            read_failure_threshold,
+            write_failure_threshold,
+            counter_reset_window: Duration::from_secs(counter_reset_window_minutes * 60),
+            partition_unavailability_duration: Duration::from_secs(partition_unavailability_secs),
+            failback_sweep_interval: Duration::from_secs(failback_sweep_secs),
+            circuit_breaker_option_enabled,
```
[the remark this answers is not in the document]
kundadebdatta: Used `.max()` calls in `from_options`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/partition_endpoint_state.rs
```diff
@@ -0,0 +1,208 @@
...
+impl Default for PartitionFailoverConfig {
+    fn default() -> Self {
+        Self {
+            read_failure_threshold: 10,
+            write_failure_threshold: 5,
+            counter_reset_window: Duration::from_secs(5 * 60),
+            partition_unavailability_duration: Duration::from_secs(5),
+            failback_sweep_interval: Duration::from_secs(300),
+            circuit_breaker_option_enabled: false,
+        }
```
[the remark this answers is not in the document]
kundadebdatta: PR description is now updated to match with the thresholds mentioned in the spec and implementation.

sdk/cosmos/azure_data_cosmos_driver/tests/multi_region_tests/driver_partition_failover.rs
```diff
@@ -0,0 +1,609 @@
...
+            .await?;
+
+        // Seed an item first (no fault on writes — rule only targets reads in hub region).
+        let item_json = br#"{"id": "ppcb-item-503", "pk": "pk1", "value": "test"}"#;
+        context.create_item(&container, "pk1", item_json).await?;
+
+        // Issue reads to accumulate failures and trigger the circuit breaker threshold.
+        // Default read failure threshold is 5. Each read that hits the hub region
+        // will increment the counter. Reads are retried across regions, so they
+        // should eventually succeed via a non-hub region.
```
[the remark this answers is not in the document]
kundadebdatta: Yes, thanks. The threshold expectations are now updated.

sdk/cosmos/azure_data_cosmos_driver/tests/multi_region_tests/driver_partition_failover.rs
```diff
@@ -0,0 +1,609 @@
...
+#[tokio::test]
+#[cfg_attr(
+    not(test_category = "multi_region"),
+    ignore = "requires test_category 'multi_region'"
+)]
+pub async fn ppaf_disabled_503_on_create_does_not_failover() -> Result<(), Box<dyn Error>> {
+    // This test requires a single-master account with PPAF **disabled**.
+    // When PPAF is off and the request is already sent, a non-idempotent write
+    // (CreateItem) that gets a transient error (503) is not safe to retry, so
+    // the driver should abort.
```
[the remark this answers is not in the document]
kundadebdatta: The PPAF disabled tests are removed from this file already. This comment is now stale.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -370,16 +482,66 @@ fn resolve_endpoint(
...
+                if entry.health_status == HealthStatus::ProbeCandidate {
+                    // Route probe request to the original (first failed) endpoint.
+                    return make_partition_routing(entry.first_failed_endpoint.clone());
+                }
+                if can_circuit_breaker_trigger_failover(entry, is_read, &partitions.config) {
+                    return make_partition_routing(entry.current_endpoint.clone());
+                }
+            }
+        } else if is_eligible_for_ppaf(partitions, account, is_read, is_partitioned) {
+            if let Some(entry) = partitions.failover_overrides.get(pk_range_id) {
```
tvaron3, line 560: I might be missing it but I don't see where the failover overrides gets cleared when there is a 403.3 signal from service. This seems to me once the ppaf failover happens we would never fail back.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
```diff
@@ -172,10 +178,272 @@ pub(crate) fn expire_unavailable_endpoints(
...
+    now: Instant,
+    unavailability_duration: Duration,
+) -> PartitionEndpointState {
+    let mut new_state = state.clone();
+
+    for entry in new_state.circuit_breaker_overrides.values_mut() {
+        if entry.health_status == HealthStatus::Unhealthy
+            && now.saturating_duration_since(entry.first_failure_time) >= unavailability_duration
+        {
+            entry.health_status = HealthStatus::ProbeCandidate;
```
tvaron3, line 512: We should add a protection against thundering herd effect to previously bad region.

sdk/cosmos/azure_data_cosmos_driver/tests/multi_region_tests/driver_partition_failover.rs
```diff
@@ -0,0 +1,263 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
tvaron3, line 1: As this pr is integrating driver pkrange cache into request path. Integration split and pkrange failure tests should be added.

Integration tests for ppcb should be added as well for mm and sm without ppaf

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
```diff
@@ -7,11 +7,13 @@
 //! - Success → Complete
 //! - Transport error (NotSent) → TransportRetry if budget allows
 //! - Transport error (Sent/Unknown, idempotent) → TransportRetry if budget allows
-//! - Transport error (Sent/Unknown, non-idempotent) → Abort
+//! - Transport error (Sent/Unknown, non-idempotent, no PPAF) → Abort
 //! - 403/3 WriteForbidden → FailoverRetry + refresh + mark unavailable
 //! - 404/1002 ReadSessionNotAvailable → SessionRetry (advances region)
+//! - 408 RequestTimeout → FailoverRetry + mark partition/endpoint unavailable
 //! - 503, 429/3092, 410 → FailoverRetry + mark partition/endpoint unavailable
```
tvaron3, line 14: ppcb needs to track all 5xx errors

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -237,9 +281,47 @@ pub(crate) async fn execute_operation_pipeline(
...
+                            } else {
+                                current.clone()
+                            }
+                        });
+                    }
+                }
                 return build_cosmos_response(result, diagnostics);
             }
             OperationAction::FailoverRetry { new_state, delay } => {
+                tracing::warn!(
```
tvaron3: downgrade to debug or remove. Seems noisy

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -370,16 +482,66 @@ fn resolve_endpoint(
             selected = Some(candidate.clone());
             break;
         }
+
+        // Track the first unavailable-but-not-excluded endpoint so we can
+        // de-prioritize it rather than skip it entirely.  This matches the
+        // SDK behavior: [available] → [unavailable regional] → global.
+        if first_unavailable.is_none() {
```
tvaron3: This is great ! please update changelogs and include this behavioral change as well.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -906,6 +931,146 @@ impl CosmosDriver {
...
+                            ranges: vec![],
+                            continuation,
+                            not_modified: true,
+                        });
+                    }
+
+                    // Permanent errors (auth/config issues) are logged at error
+                    // level so operators can distinguish misconfiguration from
+                    // transient blips.
+                    // TODO: Consider adding a negative-cache TTL to suppress
```
tvaron3, line 1020: gh issue might be better for tracking and at least on boot up will be better to fail fast on the pkrange cache not being filled.

sdk/cosmos/azure_data_cosmos_driver/tests/multi_region_tests/driver_partition_failover.rs
```diff
@@ -0,0 +1,263 @@
...
+// Licensed under the MIT License.
```
tvaron3: Integration tests for ppcb should be added as well for mm

sdk/cosmos/azure_data_cosmos_driver/tests/multi_region_tests/driver_partition_failover.rs
```diff
@@ -0,0 +1,263 @@
...
+/// When PPCB is enabled (default when PPAF is enabled) and region 1 returns
+/// 503 for reads, after the default read failure threshold (10) is exceeded,
+/// subsequent reads should be routed directly to the next region via a
+/// partition-level circuit breaker override.
+#[tokio::test]
+#[cfg_attr(
+    not(test_category = "multi_region"),
+    ignore = "requires test_category 'multi_region'"
+)]
+pub async fn ppcb_enabled_503_on_read_fails_over_after_threshold() -> Result<(), Box<dyn Error>> {
```
tvaron3, line 182: Could we add a test as well for failing back to the original region

sdk/cosmos/azure_data_cosmos_driver/src/models/mod.rs
```diff
@@ -368,6 +368,20 @@ impl ResourceType {
...
 
+    /// Returns true if this resource type supports partition-level failover.
+    ///
+    /// Documents are always partitioned. Stored procedures are only
+    /// partitioned when the operation is `Execute` (i.e. executing the
+    /// sproc against a specific partition). CRUD operations on stored
+    /// procedure metadata are not partition-scoped.
+    pub fn is_partitioned(self, operation_type: OperationType) -> bool {
+        match self {
+            ResourceType::Document => true,
```
tvaron3: We should exclude query plan as well proactively.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
```diff
@@ -356,7 +408,20 @@ impl LocationStateStore {
...
+        // Update partition-level PPAF/PPCB flags from account properties.
+        let per_partition_automatic_failover_enabled =
+            properties.enable_per_partition_failover_behavior;
+
         *self.last_synced_properties.lock().unwrap() = Some(properties);
+        self.apply_partition(|current| {
+            let mut next = current.clone();
+            next.per_partition_automatic_failover_enabled =
+                per_partition_automatic_failover_enabled;
+            next.per_partition_circuit_breaker_enabled = per_partition_automatic_failover_enabled
```
tvaron3, line 437: I think this will prevent ppcb from being disabled if ppaf is on. I am not against it but making sure this is a conscious decision

tvaron3 (changes requested):

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
tvaron3: nvm I see it in routing system file. Makes sense.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
kundadebdatta: Yes - that is the intent and it is expected by design. When PPAF is enabled, we do want to enable PPCB as well, to failover the reads per partition.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -1235,28 +1270,42 @@ better trade-off than introducing locks.
...
-- If a partition has been previously failed over and a new request arrives, the
-  override is only applied if the partition key range ID was already learned
-  from a prior operation for the same partition.
-
-**Future improvement**: If the driver implements direct-mode partition key range
-resolution (bypassing the gateway for physical partition discovery), the partition
-key range ID could be known before the first attempt.
+### 13.4 First Attempt Has No Partition Override (Partially Addressed)
+
+~~The partition key range ID is not known until the first response is received
```
FabianMeiswinkel: Why? You know the PkRangeId form calculating Hash/EPK and looking up the overlapping entry in PKRanges?

FabianMeiswinkel: Looks like that is applied below already.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
```diff
@@ -1235,28 +1270,42 @@ better trade-off than introducing locks.
...
+- On the first operation targeting a given container, the cache is empty. The driver
+  fetches partition key ranges from the service (`/pkranges` changefeed endpoint),
+  populates the routing map, and resolves the range ID — all before the first
+  pipeline attempt.
+- Subsequent operations for the same container use the cached routing map (O(log n)
+  binary search on range boundaries).
+- The cache uses incremental change feed (`A-IM: Incremental feed` + `If-None-Match`
+  etag) for efficient refresh.
+
+**Remaining limitation**: If the pre-resolution fetch fails (network error, service
```
FabianMeiswinkel: Why - fail fast instead (and add appropriate retry policy for really transient errors). PKRanges is critical - without it the SDK can't work properly anyway.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -237,9 +281,47 @@ pub(crate) async fn execute_operation_pipeline(
...
+                            // Re-check ProbeCandidate status inside the CAS closure to
+                            // avoid a TOCTOU race: if another thread transitions the
+                            // entry to Unhealthy between the snapshot check and the CAS,
+                            // we must not remove the Unhealthy entry.
+                            let is_probe = current
+                                .circuit_breaker_overrides
+                                .get(pk_range_id.as_str())
+                                .is_some_and(|e| e.health_status == HealthStatus::ProbeCandidate);
+                            if is_probe {
+                                remove_probe_succeeded_entry(current, pk_range_id)
```
FabianMeiswinkel: Why remove - wouldn't we just update state to either healthy or unhealthy?

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -316,6 +398,35 @@ pub(crate) async fn execute_operation_pipeline(
                 }
             }
             OperationAction::Abort { error, status } => {
+                // 304 Not Modified is an expected response for changefeed-based
```
FabianMeiswinkel: Why would 304 Not Modified even hit OperationAction::Abort (and not ::Complete?) - isn't that the real problemn - not the logging level.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
FabianMeiswinkel: And PPAF only 403/3 and 503 from not getting a response or 410/1022? Like if service actually sends a 503/0 I do not think it qualifies for PPAF evaluation - only endpoint unreachable, 404.3 and maybe Lease Lost because in that case the backend might simply have stale memory about topolgy?

FabianMeiswinkel: Never mind - in Gateway mode we can't distinguish this - so, we have to rely on all 503s at leats to trigger PPAF evaluation

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -59,6 +66,7 @@ pub(crate) async fn execute_operation_pipeline(
     session_manager: &SessionManager,
     account_default_consistency: DefaultConsistencyLevel,
     throughput_control: Option<&ThroughputControlGroupSnapshot>,
+    pre_resolved_pk_range_id: Option<PartitionKeyRangeId>,
 ) -> azure_core::Result<CosmosResponse> {
```
FabianMeiswinkel, line 74: with nearly 300 lines of code this method gets unmaintainable - worth refactoring individual stages into separate methods (and probably also tighten unit test coverage aroudn them?)

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
FabianMeiswinkel, line 181: Same here - method should be split-up IMO

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
```diff
@@ -231,8 +239,22 @@ impl LocationStateStore {
...
-                LocationEffect::MarkPartitionUnavailable(_) => {
-                    // TODO(partition-routing): Apply partition-level unavailability.
+                LocationEffect::MarkPartitionUnavailable(partition) => {
+                    if partition.partition_key_range_id.is_none() {
+                        // No partition key range ID available (first attempt);
+                        // skip partition-level marking.
+                        continue;
+                    }
+                    let is_partitioned = partition.is_partitioned_resource;
+                    self.apply_partition(|current_partitions| {
```
FabianMeiswinkel, line 259: I do not see how this can be thread-safe. mark_partition_unavailable creates a snapshot and also uses increments of counters.  then apply-partition will basically use a pre-conditioned CompareExchange to update - but since the mark_partition_unavailable might have created a snapshot sooner this would possibly override concurrent updates? I might have missed something - haven't checked-out your brnach to dig thorugh it line-by-line - but please double check these race-conditions here.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
```diff
@@ -907,6 +932,146 @@ impl CosmosDriver {
...
+    ///
+    /// Builds a GET request to `/dbs/{db_rid}/colls/{container_rid}/pkranges`
+    /// using the `A-IM: Incremental feed` header for changefeed semantics.
+    /// When `continuation` is provided, it is sent as the `If-None-Match` header
+    /// for incremental fetches. The server may return 304 Not Modified if no
+    /// new ranges exist since the last fetch.
+    ///
+    /// Returns `None` if the request fails (transient error, network issue, etc.).
+    /// The caller (the PK range cache) falls back gracefully on `None`.
+    async fn fetch_pk_ranges_from_service(
```
FabianMeiswinkel, line 959: Retry policy? This should be retried (even cross regions) - and we recently discussed even adding hedging in addition to "just" cross region retry?

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_operation.rs
```diff
@@ -539,6 +539,17 @@ impl CosmosOperation {
         Self::query_items(container, PartitionKey::EMPTY)
     }
 
+    /// Reads (lists) all partition key ranges for a container.
+    ///
+    /// Returns a feed of partition key range resources.
+    /// Used internally by the partition key range cache to build routing maps.
+    pub fn read_all_partition_key_ranges(container: ContainerReference) -> Self {
```
FabianMeiswinkel: why public? This makes it public surface area? Any pkranges read (or feedranges) in public surface area shoudl defautl to read form pkrangecache (and only allow forced refresh as opt-in)

sdk/cosmos/azure_data_cosmos_driver/src/models/mod.rs
FabianMeiswinkel: +1

sdk/cosmos/azure_data_cosmos_driver/src/options/operation_options.rs
```diff
@@ -90,6 +90,30 @@ pub struct OperationOptions {
...
+    /// Window (in minutes) after which failure counters reset.
+    #[option(env = "AZURE_COSMOS_CIRCUIT_BREAKER_TIMEOUT_COUNTER_RESET_WINDOW_IN_MINUTES")]
+    pub circuit_breaker_timeout_counter_reset_window_in_minutes: Option<u32>,
+
+    /// Duration (in seconds) a partition must remain unavailable before failback.
+    #[option(env = "AZURE_COSMOS_ALLOWED_PARTITION_UNAVAILABILITY_DURATION_IN_SECONDS")]
+    pub allowed_partition_unavailability_duration_in_seconds: Option<u32>,
+
+    /// Interval (in seconds) between background failback sweep iterations.
+    #[option(env = "AZURE_COSMOS_PPCB_STALE_PARTITION_UNAVAILABILITY_REFRESH_INTERVAL_IN_SECONDS")]
```
FabianMeiswinkel, line 158: IMO it woud be good to document default settings here as well. also maybe a bit of info what happens when changing away from default?

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
kundadebdatta: Yes correct and agreed! The ideal fix is to refactor the `is_success()` to consider the 304 not modified status as well:

image
 
 Updated the code to reflect this.

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_operation.rs
kundadebdatta: Yea, this should be treated as `pub (crate)`. Refactored the code to reflect the change.

sdk/cosmos/azure_data_cosmos_driver/src/options/operation_options.rs
kundadebdatta: Yup, agreed. I have added the code comment to document the default settings.

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/location_state_store.rs
kundadebdatta: Thanks for flagging this — let me walk through why this is actually safe.

**The CAS loop re-executes the closure on every retry**

`apply_partition` doesn't just do a one-shot compare-exchange against a pre-built next-state. The closure is invoked *inside* the loop:

```rust
loop {
    let current = self.partitions.load(Ordering::Acquire, &guard);
    let current_ref = unsafe { current.deref() };
    let next_state = f(current_ref);            // ← re-executed every iteration

    match self.partitions.compare_exchange(current, Owned::new(next_state), ...) {
        Ok(_)  => return,
        Err(_) => continue,                      // ← loops back, re-reads `current`
    }
}
```

So if a concurrent writer wins the CAS, we re-load the latest `partitions` pointer and re-run `mark_partition_unavailable` against that fresher baseline. The counter increment inside `mark_partition_unavailable` is therefore applied to whatever the latest committed state is — concurrent increments compose correctly and no updates are lost. This is the standard "read-modify-write inside the CAS loop" pattern, not "build new-state once then CAS".

**The one cross-state observation that ISN'T CAS-serialised — and why it's intentional**

The closure also calls `self.account_snapshot()`, which reads a *different* `Atomic` (the account-endpoint state). That read is not serialised against the partition CAS. This is intentional:

- `account_snapshot()` is read-only here — used only to compute *which alternate endpoint* to route to next. We never write to account state from this code path.
- Account-state churn (e.g., a concurrent `RefreshAccountProperties`) only changes the *set* of preferred endpoints; it doesn't invalidate the per-partition failure counter we're about to bump.
- If we observe an account snapshot one CAS generation behind the partition snapshot, the worst case is that the next attempt is routed using a slightly stale endpoint list — the subsequent attempt then re-resolves against the latest account snapshot through the normal pipeline path.

**Net guarantee**

- Per-partition failure tracking is **linearizable** (CAS-serialised on `partitions`).
- Cross-state coupling between `account` and `partitions` is **intentionally eventually-consistent** — we accept a one-generation skew in exchange for not having to take a coarser lock across both atomics.

I'll add an in-code comment capturing this so future readers don't have to re-derive it.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
kundadebdatta: Logging level is now downgraded to `debug`.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
kundadebdatta: I have refactored this method to create individual stages and moved them into separate methods.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
kundadebdatta: Thanks, I have refactored the code to split it up.

sdk/cosmos/azure_data_cosmos_driver/tests/multi_region_tests/driver_partition_failover.rs
kundadebdatta: An integration test to cover the PPCB failback is now added.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
kundadebdatta: Removal-on-success should have the same outcome as updating the state as healthy. (no entry = default routing = same as Healthy). Additionally it clears the PPCB HashMap entry, which I feel is the right choice. It also helps to bound memory, since the `circuit_breaker_overrides` is keyed by`PartitionKeyRangeId`, which can grow with the partition count of every container the client has ever touched.

I understand for PPAF, we do not clean up the hashmap, since the hint is provided explicitly by the backend, however, if you would like I can create a follow up tracking item to see how we can safely remove the entries when the PPAF write fails back to the primary write region. Can be taken as a follow up.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
kundadebdatta: Cross-region retry is already in place — each call to `execute_operation` for pkranges flows through the full operation pipeline and retry_evaluation, which rotates through preferred read regions on FailoverRetry (503/408/429/410). 

The hedging part will be covered when I introduce the cross region hedging in the driver.

sdk/cosmos/azure_data_cosmos_driver/src/models/mod.rs
kundadebdatta: This is now taken care with the latest commit. Thanks for bringing this up.

sdk/cosmos/azure_data_cosmos_driver/docs/PARTITION_LEVEL_FAILOVER_SPEC.md
kundadebdatta: I have updated the spec to reflect this behavior. Today, the pk range fetch is already behind the `operation_pipeline` which does the cross regional routing. There are some of the known gaps, which I highlighted [here in this task](https://github.com/Azure/azure-sdk-for-rust/issues/4301). I will address these gaps as a follow up task.

sdk/cosmos/azure_data_cosmos_driver/src/driver/cosmos_driver.rs
kundadebdatta: Yeah - this will be tracked as a [separate gh task](https://github.com/Azure/azure-sdk-for-rust/issues/4301).

sdk/cosmos/azure_data_cosmos_driver/src/driver/routing/routing_systems.rs
kundadebdatta: To prevent this, I have added a jitter which should prevent the thundering herd effect.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
kundadebdatta: This is reflected in the change log.

sdk/cosmos/azure_data_cosmos_driver/tests/multi_region_tests/driver_partition_failover.rs
kundadebdatta: Added the required tests to cover the cases.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/retry_evaluation.rs
kundadebdatta: Yea - to track this, and to reduce more scope for this particular PR, it's better to follow up with a separate PR. We do have [this follow up item](https://github.com/Azure/azure-sdk-for-rust/issues/4301) to cover this, and some of the other areas.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
FabianMeiswinkel: Yeah - ok for follow-up. Perf fro remove vs. update to Healthy State could be relevant. Usually Update would be faster (reason for exsitence on Tombstones in many databases)

FabianMeiswinkel: Memory - I think i mentioned before - predictabiliyt is important we (and any customer) need to support every partition triggering PPCB fail-over - so removal for memory reasons IMO is not a good motivation.

FabianMeiswinkel (approved):
LGTM

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
Pilchie: Big design question, but should we consider a threshold over which we just fail all partitions over? If so, what should that threshold be?

tvaron3 (approved):
lgtm

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
```diff
@@ -205,6 +208,18 @@ pub(crate) async fn execute_operation_pipeline(
         )
         .await;
 
+        // Capture partition key range ID from response headers (first time only).
+        if retry_state.partition_key_range_id.is_none() {
+            if let Some(headers) = result.response_headers() {
```
tvaron3: response headers aren't populated on successes it seems based on the method doc. Should use `cosmos_headers()` instead

kundadebdatta: Thanks Tomas! Good catch. Refactored the code to use `cosmos_headers()`.

sdk/cosmos/azure_data_cosmos_driver/src/options/operation_options.rs
```diff
@@ -90,6 +90,88 @@ pub struct OperationOptions {
...
+    /// alternate region once the threshold is exceeded, then probes the
+    /// original region for recovery via the failback sweep.
+    ///
+    /// **Tuning**: Enable to opt into partition-level circuit breaking on
+    /// reads (any account) and writes (multi-master accounts). When disabled,
+    /// the driver falls back to account-level endpoint marking, which is
+    /// coarser-grained.
+    #[option(env = "AZURE_COSMOS_PER_PARTITION_CIRCUIT_BREAKER_ENABLED")]
+    pub per_partition_circuit_breaker_enabled: Option<bool>,
+
```
analogrelay, line 174: Feels like we should group these into a `PartitionFailoverOptions` type of their own.

analogrelay (approved):
Nothing blocking

sdk/cosmos/azure_data_cosmos_driver/src/options/operation_options.rs
kundadebdatta: Thanks @analogrelay. I will take this as a follow up, and have a task already. Will include this as well.

sdk/cosmos/azure_data_cosmos_driver/src/driver/pipeline/operation_pipeline.rs
kundadebdatta: Yes - we have the [thresholds defined already](https://github.com/Azure/azure-sdk-for-rust/blob/users/kundadebdatta/3742_implement_ppaf_ppcb_in_driver_crate/sdk/cosmos/azure_data_cosmos_driver/src/options/operation_options.rs#L104-L174), and based on these numbers (per partition) we start failing over the partitions. On a high level, in PPCB, for reads, the default value is `10` within `1` min time span and for multi master writes it's `5` failures in a min.

However, these numbers are overridable by setting the respective environment variable.

Pilchie: Those are thresholds for what should trigger an individual partition to failover. I was talking about whether there should be a threshold where if something like 20% or 50% or 80% of the partitions in an account failover, we just proactively fail over the rest. If we did that, we would only ever need to hold state for up to that percentage of partitions.

kundadebdatta: I think this requires some brainstorming. If we choose the threshold as `80%`, and that basically triggers the rest of the partitions (`20%`) to fail over, then from memory perspective we can save the spot for the rest of the `20%` but one cons I see is what if the original region was still good to serve the requests for those `20%` of the partitions, and if we choose to do so, then we may end up increasing the latency for those `20%`, provided the failed over region is further from the original region compared to where the application is hosted.
 
But, I think higher the threshold, higher the chances to being accurate. So definitely a good discussion point to consider and I would like to take this as a follow up.

Thoughts on this @FabianMeiswinkel @analogrelay @tvaron3 ?

FabianMeiswinkel: IMO - keep these two orthogonal. We should not be required to have to enforce a limit for memory consumption. Let's apply an efficient data-structure - then re-evaluate based on info how much memory per partiton really is needed. @Pilchie's proposal could be a good choice also from a consistency perspective etc. - but would like to keep it separate - and probably would need ot be configurable. So, late-binding that feature would be my preference.

sdk/cosmos/azure_data_cosmos_driver/src/options/operation_options.rs
kundadebdatta: Tracked it as a [part of this](https://github.com/Azure/azure-sdk-for-rust/issues/4301).

PR #4159, opened 2026-04-13
Cosmos: Microbenchmark and some optimizations

analogrelay:
I wrote up a quick microbenchmark of the driver's `execute_operation` code path, which excludes the I/O entirely by using a fixed Mock Transport (gated behind an internal feature flag). Using that, I was able to get a 40+% improvement in performance (again, these times **exclude** I/O, but should reflect CPU usage reductions).

Before:

```
point_read              time:   [16.691 µs 16.734 µs 16.783 µs]
```

After

```
point_read              time:   [9.7474 µs 9.7864 µs 9.8319 µs]
```

Obviously, these are tiny numbers, but I suspect they add up under high concurrency. We won't know for sure until this is run in the benchmarks. They should be safe, so I'm game to merge and see, but we could also try running the tip of this branch in the perf infra (cc @tvaron3 ) to get real results.

The optimizations I worked on were predominantly to reduce duplicate parsing and `HashMap::insert`/`HashMap::get` calls. Rather than calling `HashMap::get_optional_str(...)` for each header when materializing a `CosmosResponseHeaders`, we now iterate the map. Rather than re-parsing `url::Url`, the `AccountRegion` type now directly deserializes the gateway response into an `AccountEndpoint`, allowing us to reuse the same `url::Url` for each request. Rather than constantly `format!`-ing the `CosmosResourceReference`, we now produce a single `ResourcePaths` from it and allow getting slices from it for the various purposes.

The benchmarks include instructions for running both profiled and unprofiled runs, and [criterion](https://criterion-rs.github.io/) does a great job of tracking results and showing you improvements over time. Having benchmark infra like this in place will be useful as we keep going.


sdk/cosmos/azure_data_cosmos_benchmarks/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "azure_data_cosmos_benchmarks"
```
heaths, line 2: Does this really need a separate crate? You can't use benches within your existing crates? You can require features in your benches. Have to specified on the command line, but not too annoying, IMO.

sdk/cosmos/azure_data_cosmos_benchmarks/examples/heap_profile.rs
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+//! Heap allocation trace for a single warm `execute_operation` call.
+//!
+//! Uses the `dhat` heap profiler to capture every allocation that occurs
```
heaths: And here I wrote my own. That said, it wasn't hard. Do we need a new dependency? Is this doing sufficiently more than https://github.com/heaths/ordinal-rs/blob/e946015d49e233002212c4ae12022b01642ab6fc/benches/ordinal.rs#L67-L111?

sdk/cosmos/azure_data_cosmos_benchmarks/benches/point_read.rs
```diff
@@ -0,0 +1,113 @@
...
+use tokio::runtime::Builder;
+
+// ---------------------------------------------------------------------------
+// pprof adapter for criterion 0.8
+//
+// pprof's built-in `criterion` feature targets criterion 0.5. We implement
+// the criterion 0.8 `Profiler` trait ourselves using pprof's core API.
+// ---------------------------------------------------------------------------
+
+struct PprofAdapter {
```
heaths: Same here. If we're getting sufficient value - mine is admittedly naive, but with what little we can do to profile memory with criterion anyway, it's probably good enough - sure. pprof, at least, I'm familiar with for rust.

heaths (approved):
Approved for files outside sdk/cosmos. I had some questions/(slight) concerns under sdk/cosmos.

sdk/cosmos/azure_data_cosmos_benchmarks/Cargo.toml
analogrelay: It's that latter thing I was trying to avoid but I'm not entirely opposed to putting them in the driver package itself.

heaths: FWIW, I don't hate it. Just seems like the explosion of bespoke crates might get a little confusing when onboarding new devs as well. At least when you have to pass a required feature, the error tells you exactly what to do. Up to you.

sdk/cosmos/azure_data_cosmos_benchmarks/examples/heap_profile.rs
analogrelay: Probably not, but I don't see a big problem with adding industry-standard tools like dhat and pprof. Having said that, it looks like this implementation of dhat is a bit rough, so I may just replace it with running valgrind myself.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/cosmos_transport_client.rs
```diff
@@ -34,7 +34,7 @@ use crate::fault_injection::EvaluationCollector;
 /// field so the transport layer can enforce a deadline that differs from the
 /// client-level default.
 #[derive(Clone, Debug)]
-pub(crate) struct HttpRequest {
+pub struct HttpRequest {
```
FabianMeiswinkel, line 39: These changes are effectively no-ops and just incraese churn on the PR - remove them?

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/http_client_factory.rs
```diff
@@ -21,7 +21,7 @@ pub(crate) enum HttpVersionPolicy {
 }
 
 #[derive(Clone, Copy, Debug)]
-pub(crate) struct HttpClientConfig {
+pub struct HttpClientConfig {
```
FabianMeiswinkel, line 24: dito

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_headers.rs
```diff
@@ -9,48 +9,34 @@ use base64::{engine::general_purpose::STANDARD, Engine as _};
...
-    pub static SERVER_DURATION_MS: HeaderName = HeaderName::from_static("x-ms-request-duration-ms");
-    pub static LSN: HeaderName = HeaderName::from_static("lsn");
-    pub static OWNER_FULL_NAME: HeaderName = HeaderName::from_static("x-ms-alt-content-path");
-    pub static OWNER_ID: HeaderName = HeaderName::from_static("x-ms-content-path");
-    pub static OFFER_REPLACE_PENDING: HeaderName =
-        HeaderName::from_static("x-ms-offer-replace-pending");
+/// All names are lowercase as required by HTTP/1.1. The azure_core [`Headers`]
+/// type normalizes header names to lowercase on insertion, so lookups are
+/// case-sensitive but will always match since both sides are lowercase.
+pub(crate) mod header_names {
```
FabianMeiswinkel: keepin request and resposne header names separates makes more sense - because that is what RNTBD will force us to do anyways. And all but 2 or 3 headers will move into RNTBD payload with GW2.0

FabianMeiswinkel (approved):
LGTM

sdk/cosmos/azure_data_cosmos_driver/src/models/cosmos_headers.rs
analogrelay: Fair enough, there was just enough similarity there I had merged it.

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/cosmos_transport_client.rs
analogrelay: Weird, I guess Copilot really wanted to simplify these. I'll clean them up

sdk/cosmos/azure_data_cosmos_benchmarks/Cargo.toml
analogrelay: In some ways, I think this separation helps, and I'd almost prefer to do the same for some of our more esoteric integration tests. Most external contributors don't need to worry about this, nor do they need to worry about many of the integration tests. This makes it easier for the simple-fix contributor to come in and make changes without getting bogged down with a lot of infrastructure.

sdk/cosmos/azure_data_cosmos_benchmarks/examples/heap_profile.rs
analogrelay: The main thing about `dhat` is that it produces a standard-format output that is readable using existing tools. The author of dhat-rs is also the author of the dhat tool built in to valgrind, so I'm not too worried about having it available for us when we want to use it.

I _am_ planning on removing `pprof` and shifting to a model of using an external profiler like `perf` on Linux though.

sdk/cosmos/azure_data_cosmos_benchmarks/benches/point_read.rs
analogrelay: Yeah, I'll revisit this a bit. I think `perf` works well enough here and I'm doing most benchmarking on Linux (WSL) right now.

FabianMeiswinkel (approved):
LGTM

tvaron3 (approved):
LGTM

simorenoh (approved):

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+// cSpell:ignore evals
```
FabianMeiswinkel, line 4: Why ignore all cspell in this file? Either we enable cspell checks or we disable it - disabling it for entire files defeats the purpose. I have no strong feeling - can live with either - but think this is the worst trade-off

FabianMeiswinkel (changes requested):
cspell - either we disable it everywhere (except maybe md files) or we enforce cspell and whitelist exceptions - but disabling cspell for entire files defeats the purpose of having it (silently accidental regressions in those files will happen)

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
FabianMeiswinkel: Ok - misunderstanding - cspell_ignore xxx just ignores the word xxx

FabianMeiswinkel (approved):
LGTM

sdk/cosmos/azure_data_cosmos_driver/src/driver/transport/transport_pipeline.rs
analogrelay: As clarified offline, this is just ignoring the _word_ `evals` for this file. It's understandably confusing! 😅

PR #4164, opened 2026-04-14
[Storage] `range` parameter type `HttpRange` generated, fortify failing queue test in windows nightly

vincenttran-msft:
`.tsp`: https://github.com/Azure/azure-rest-api-specs/pull/42277

- Changes `range` parameter from `String` to `HttpRange`
- Fortifies queue failing test in pipeline from 15s to 30s wait time


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -369,7 +369,7 @@ impl<'a> BlobClientDownloadBehavior<'a> {
 impl PartitionedDownloadBehavior for BlobClientDownloadBehavior<'_> {
     async fn transfer_range(&self, range: Option<Range<usize>>) -> Result<AsyncRawResponse> {
         let mut opt = self.options.clone();
-        opt.range = range.map(|r| r.as_range_header());
+        opt.range = range.map(|r| HttpRange::new(r.start as u64, (r.end - r.start) as u64));
```
heaths, line 372: IIRC, isn't there a `From<Range> for HttpRange` or something so you can just `into()`? Not blocking, but better. If not, I'd add one for GA.

sdk/storage/azure_storage_blob/tsp-location.yaml
```diff
@@ -1,4 +1,4 @@
 directory: specification/storage/Microsoft.BlobStorage
-commit: 3928c0a2de8aad3536d42578eefdd1dd7126072b
+commit: 8a8972d895b592eb354e1b75d00b1b01f93b40c1
 repo: Azure/azure-rest-api-specs
-additionalDirectories:
+additionalDirectories:
```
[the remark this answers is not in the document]
heaths: I'd even just remove this line.

sdk/storage/azure_storage_queue/tests/queue_service_client.rs
```diff
@@ -506,16 +506,35 @@ async fn test_set_service_properties(ctx: TestContext) -> Result<()> {
...
+                let propagated = current
+                    .logging
+                    .as_ref()
+                    .and_then(|l| l.delete)
+                    .unwrap_or(false);
+                if propagated || std::time::Instant::now() >= deadline {
+                    break current;
+                }
+                time::sleep(Duration::from_secs(5)).await;
+            }
```
heaths, line 530: Tip: this works, but if you think you'll need code like this in other places - not surprising for some scenarios that need to wait - you might consider a helper in your `azure_storage_test` crate like I did for KV: `sdk/keyvault/azure_security_keyvault_test/src/lib.rs`. It's not exactly what you need, but a start.

The key thing here, though, is to make sure your "setup" code - if that's all it is - isn't recorded, unless it's repeatable. You can use `let _guard = recording.skip()` (or something like that) if you want to disable recording for sections of code. Just drop the guard explicitly or do that in a separate scope i.e., `{ /* .. */ }` to have it drop automatically.

heaths (approved):
Nothing blocking, but a few suggestions either now or after release.

jalauzon-msft (approved):

vincenttran-msft:
Pausing on getting this change in since we are still deciding on how `range` should be surfaced.

jaschrep-msft:
~~It's not clear to me what this type buys us in the public API specifically that `Range<u64>` doesn't. I don't think there's anything the HTTP Range header can do that isn't representable by std's `Range<T>`, and using features of the standard library is going to be much more intuitive.~~

~~While we have a `ContentRange` type, that is an internal type used for parsing response headers. It's not exposed anywhere. And we explicitly convert it at parse time from an inclusive to an exclusive end so that we can more easily write code using `Range<T>`.~~

It is now clear.

sdk/storage/azure_storage_blob/src/clients/blob_client.rs
jaschrep-msft: @heaths do we expect customers to write `opt.range = my_std_range.into()`? Why not just accept the range itself? I suppose a single `.into()` isn't that bad, but I'm not sure what it buys us.

jaschrep-msft: Poking around at a possible implementation, I understand exactly why you want it this way 😅

vincenttran-msft:
Closing in favor of the new version.

PR #4174, opened 2026-04-14
Cosmos: Cut over `query` methods to the driver

### LINKED ISSUE #4139: Cosmos: Cut over `query` methods to driver
Issue to track the moving of the current SDK-side executor to linking the Driver's query executor to further the migration process. Right now the SDK by default only supports streamable cross-partition queries and single-pk queries, which is the same logic currently present in the Driver. As such, this issue only covers the cutover of SDK to driver for the same default behavior.

Follow-ups to this are:
https://github.com/Azure/azure-sdk-for-rust/issues/4155 - HPK testing
https://github.com/Azure/azure-sdk-for-rust/issues/3507 - Integrating query engine

simorenoh:
Migrates query_items, query_databases, and query_containers from the gateway HTTP pipeline to the Cosmos driver pipeline, aligning query
operations with the item CRUD operations that were already cut over.

 - QueryExecutor now takes an Arc<CosmosDriver> and an operation factory closure instead of Arc<GatewayPipeline> and a ResourceLink, 
executing queries via driver.execute_operation().
 - Query-specific headers (content type, continuation, custom headers) are passed through DriverOperationOptions instead of being set 
directly on the HTTP request.
 - QueryOptions::apply_headers is now #[cfg(test)]-only since production code no longer uses it.
 - query_items_cross_partition simplified to delegate to query_items with PartitionKey::EMPTY.
 - Adds emulator tests for single-partition and cross-partition query pagination.


simorenoh:
@copilot resolve the merge conflicts in this pull request

sdk/cosmos/azure_data_cosmos/src/options/mod.rs
```diff
@@ -361,7 +361,9 @@ impl QueryOptions {
 
 impl QueryOptions {
     // Temporary: applies option values as HTTP headers for the SDK pipeline.
-    // Will be removed when query operations use the internal pipeline directly.
+    // Retained for test coverage; no longer called from production code since
```
tvaron3: Should this be removed if it is no longer used ?

sdk/cosmos/azure_data_cosmos/src/clients/container_client.rs
```diff
@@ -21,7 +21,6 @@ use crate::cosmos_request::CosmosRequest;
 use crate::handler::container_connection::ContainerConnection;
```
tvaron3, line 21: Changelog entry ?

tvaron3 (approved):
LGTM

PR #4191, opened 2026-04-16
[Storage] Mitigate live tests failures in Storage SDKs

vincenttran-msft:
Live Test Pipeline: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6170408&view=results


vincenttran-msft:
As discussed offline, failure in `post-job` steps is known issue aside from this.

jalauzon-msft (approved):

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -930,6 +938,7 @@ fn test_managed_download_args() -> impl IntoIterator<Item = TestManagedDownloadA
 }
 
 #[recorded::test]
+#[ignore = "Temporarily ignoring until we can figure out how to get this to not take down the whole test pipeline."]
 async fn test_managed_download(ctx: TestContext) -> Result<(), Box<dyn Error>> {
```
vincenttran-msft, line 942: So looking forward, we either need to give this a custom set of timeouts and stuff, or mark as playback only.

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -1409,6 +1414,7 @@ async fn test_gzip_blob_no_metadata_roundtrip(ctx: TestContext) -> Result<(), Bo
 }
 
 #[recorded::test]
+#[ignore = "Temporarily ignoring until we can figure out how to get this to not take down the whole test pipeline."]
 async fn test_gzip_blob_with_metadata_roundtrip(ctx: TestContext) -> Result<(), Box<dyn Error>> {
```
vincenttran-msft, line 1418: This one can probably be easier to bring back, but not going to push my luck. I think my issue was conflicting with my custom HTTP client settings I was using for logging.

sdk/storage/azure_storage_blob/tests/block_blob_client.rs
```diff
@@ -431,7 +431,8 @@ async fn upload_empty(ctx: TestContext) -> Result<(), Box<dyn Error>> {
     Ok(())
 }
 
-#[recorded::test(live)]
+#[recorded::test]
+#[ignore = "Temporarily ignoring until we can figure out how to get this to not take down the whole test pipeline."]
 async fn upload_large(ctx: TestContext) -> Result<(), Box<dyn Error>> {
```
vincenttran-msft, line 436: Same thing - looking forward maybe playback only or tune the knobs to allow for this not to fail.

sdk/storage/azure_storage_blob/tests/streaming.rs
```diff
@@ -15,7 +15,8 @@ use azure_storage_blob_test::{get_blob_name, get_container_client, StorageAccoun
 use futures::TryStreamExt as _;
 use std::error::Error;
 
-#[recorded::test(live)]
+#[recorded::test]
+#[ignore = "Temporarily ignoring until we can figure out how to get this to not take down the whole test pipeline."]
 async fn stream(ctx: TestContext) -> Result<(), Box<dyn Error>> {
```
vincenttran-msft, line 20: Same thing, maybe cut down on payload to make this more CI-friendly.

vincenttran-msft:
image

Just hitting some errors in `Build Pack Crate` now.

sdk/storage/azure_storage_blob/tests/blob_client.rs
heaths: What exactly is timing out? As discussed internally, we'll either hardcode sensible defaults for connection and read timeouts (hopefully the latter is merely an idle timeout - will test) or provide options to `new_http_client()` for now until we get user feedback to help guide full public API surface. But if you have a connection or read that is frequently timing out, that is cause for concern - either a service problem, or some underlying connection issue. Is this limited to a specific platform? Did you pick up the `rustls` change to use `aws-lc-rs` by default? What TLS version is this using?

sdk/storage/azure_storage_blob/tests/block_blob_client.rs
heaths: Again, what is timing out here? I'm not consider a total timeout - i.e. severing the connection after a certain wall clock time has elapsed. I'm sure you'd agree: that's bad. If this frequently times out, something else is going wrong.

FWIW, we never noticed any timeouts for other crates. Sure, they are smaller responses, but this still indicates a potential problem elsewhere. If we wouldn't add a total timeout and either a connection or read timeout would fire in these couple cases you called out - but, more importantly, not in others - that sounds more like an underlying service issue or client transport issue.

sdk/storage/azure_storage_queue/tests/queue_service_client.rs
```diff
@@ -410,7 +409,7 @@ pub async fn test_list_queues_include_metadata(ctx: TestContext) -> Result<()> {
 }
 
 /// Gets Queue service statistics.
-#[recorded::test]
+#[recorded::test(playback)]
```
heaths, line 412: Why can't this be recorded? Your checks seem robust toward changes. This is a problem for people trying to re-record.

jalauzon-msft: Secondary endpoints are not spun up immediately in Storage and this API will fail for a while after a storage account is created, since the Live test pipeline creates new accounts, the `get_statistics` will always fail there.

jalauzon-msft (approved):
Approving to unblock pipelines. Need to investigate failures further ASAP.

sdk/storage/azure_storage_queue/tests/queue_service_client.rs
heaths: How long? A test resources post-script could add a delay. We do this for KV in .NET when provisioning a Managed HSM because it takes a little while to provision before we can upload the required 3+ certs to activate it.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: I haven't been able to fully root cause, I've only recently been able to identify the "problematic" tests.

It seems like for the most part the failures were happening more frequently on Windows/Ubuntu, but I have been failures in MacOS just at a lower occurence rate.

For example, `test_managed_download` consumed my http client-imposed 300s request timeout limit (that has since been removed). So remaining questions there:

- Is 300s generous enough for the standard case (I think yes, majority platforms can psas but only inconsistently flaky)
- When it does hit a stall, what is stalling and why doesn't it try to make any progress? Or is it making progress the whole time and hitting network issues? Or only when hitting network issues, since there aren't reasonable timeouts _somewhere_, it hangs instead of failing gracefully or at least trying to be productive

Another example is `stream` in `streaming.rs`. 40MB upload took >1s, but it also dies to my 300s request timeout (that has since been removed):
time. Busy=2.48ms time.idle=298s

So in this case I can say confidently that `stream` is one of the problem tests. It hangs and only got cut because of my custom HTTP configurations I was toying with.

sdk/storage/azure_storage_blob/tests/block_blob_client.rs
vincenttran-msft: Yeah this one failed in one of the many live test runs I've done, I will revisit this with better logging in place but right now I'm in the interest of being ready for release next week.

sdk/storage/azure_storage_blob/tests/blob_client.rs
heaths: We certainly don't want a total timeout, but an idle timeout. Makes me wonder if either the connection isn't terminating or if you're hitting some TLS issue. I can't easily tell what commit you're targeting, but are you sure you picked up the TLS change from yesterday? You might rebase and try again. Granted, if the blob was uploaded it's probably not that, but, still, maybe the connection isn't being terminated correctly.

sdk/storage/azure_storage_queue/tests/queue_service_client.rs
jalauzon-msft: I'm not entirely sure and can't find any concrete docs but Copilot seems to indicate 15 minutes+. We actually need to wait for the Geo Replication service to pick up the account. Honestly, I don't think it's worth the delay to our pipeline runtime for this as we'd need to probably set to 30 mins+ to avoid it being flakey.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: The other set of "problematic" tests I can take full credit for- some of them were just setting tags and the Service can be a little more unreliable than only a `5s` wait time. I have since bumped those to 15. But the class of failures outlined above are not the same.

sdk/storage/azure_storage_blob/tests/streaming.rs
heaths: Or keep it live. If it's exposing problems - wherever they might be - it's probably worth keeping.

jalauzon-msft (approved):

PR #4200, opened 2026-04-16
[Storage] Re-record ignored marked tests, removing extraneous Storage policy, improve docs for auto-decompression behavior

vincenttran-msft:
Title states it all!


sdk/storage/azure_storage_blob/src/clients/blob_client.rs
```diff
@@ -122,6 +115,13 @@ impl BlobClient {
     ///
     /// * `buffer` - The buffer to write the blob content into. Must be large enough to hold the requested range or the entire blob.
     /// * `options` - Optional parameters for the request.
+    ///
+    /// # Notes
+    ///
+    /// Clients will use the HTTP transport returned from [`azure_core::http::new_http_client()`]
+    /// but without automatic decompression by default. If you set a custom transport in
+    /// [`BlobClientOptions`] without disabling automatic decompression, partitioned downloads
+    /// may not succeed.
```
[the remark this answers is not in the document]
vincenttran-msft: Fixed.

sdk/storage/azure_storage_blob/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### Breaking Changes
 
+- Responses are no longer automatically decompressed.
+
```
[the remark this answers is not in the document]
vincenttran-msft: This was in the wrong spot previously. I am correcting it properly here,

sdk/storage/azure_storage_blob/README.md
```diff
@@ -124,6 +124,38 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+```rust no_run
+use std::sync::Arc;
+use azure_core::http::{ClientOptions, Transport};
+use azure_storage_blob::BlobClientOptions;
+
+let client = Arc::new(
+    ::reqwest::ClientBuilder::new()
+        .gzip(false)
+        .build()?,
+);
```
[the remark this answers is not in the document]
vincenttran-msft: Good idea.

sdk/storage/azure_storage_blob/README.md
```diff
@@ -124,6 +124,36 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
 
+## Known issues
+
+### Automatic decompression with custom HTTP transports
+
+By default, `BlobClient` creates an HTTP transport with automatic decompression disabled,
+which is required for partitioned (multi-part) downloads to work correctly. If you set a custom transport
+in `BlobClientOptions` (e.g., a `reqwest::Client` with gzip enabled) without disabling automatic
+decompression, partitioned downloads via [`BlobClient::download`](https://docs.rs/azure_storage_blob/latest/azure_storage_blob/clients/struct.BlobClient.html#method.download) and
+[`BlobClient::download_into`](https://docs.rs/azure_storage_blob/latest/azure_storage_blob/clients/struct.BlobClient.html#method.download_into) may not succeed.
```
heaths: Nit: I generally try to document functions with `()`. rustdoc works with references with or without so what's here is fine, but then it's obvious they are functions. No biggie. Be consistent in your docs, though, so if they are all like this, stick with it (or change them all if you really want).

sdk/storage/azure_storage_blob/tests/blob_client.rs
```diff
@@ -689,9 +689,9 @@ async fn test_set_legal_hold(ctx: TestContext) -> Result<(), Box<dyn Error>> {
     Ok(())
 }
 
-#[recorded::test]
-#[ignore = "needs to be re-recorded by service team"]
+#[recorded::test(playback)]
 async fn test_immutability_policy(ctx: TestContext) -> Result<(), Box<dyn Error>> {
+    // This test requires ImmutableStorageWithVersioning feature enabled.
```
heaths, line 694: Can't `test-resources.bicep` enable that so we have one less playback-only test?

sdk/storage/azure_storage_blob/tests/blob_client_versioning.rs
```diff
@@ -349,9 +349,9 @@ async fn test_blob_version_feature_interactions(ctx: TestContext) -> Result<(),
     Ok(())
 }
 
-#[recorded::test]
-#[ignore = "needs to be re-recorded by service team"]
+#[recorded::test(playback)]
 async fn test_blob_version_immutability_operations(ctx: TestContext) -> Result<(), Box<dyn Error>> {
+    // This test requires ImmutableStorageWithVersioning feature enabled.
```
heaths, line 354: Ditto. I'll not repeat.

sdk/storage/azure_storage_blob/tests/blob_service_client.rs
```diff
@@ -265,9 +265,9 @@ async fn test_find_blobs_by_tags_service(ctx: TestContext) -> Result<(), Box<dyn
     Ok(())
 }
 
-#[recorded::test]
-#[ignore = "needs to be re-recorded by service team"]
+#[recorded::test(playback)]
 async fn test_get_service_stats(ctx: TestContext) -> Result<(), Box<dyn Error>> {
+    // This test requires a Storage account with secondary endpoint enabled.
```
heaths, line 270: Can't we do that in `test-resources.bicep`? If it's generally not needed and/or expensive normally, you could make it optional as we do with Managed HSMs for Key Vault testing in .NET e.g., <https://github.com/Azure/azure-sdk-for-net/blob/ded185d2b2695dd7b960b8ede052463e73d6808a/sdk/keyvault/test-resources.json#L179> and then call `New-TestResources.ps1 storage -AdditionalParameters @{enableSecondaryEndpoint=$true}`. It's actually a little different in bicep, though, e.g. <https://github.com/heaths/azcrypto/blob/f2c2399950a8a3ecc52f52c2fe9b595a8ce9b22e/infra/resources.bicep#L56-L72>.

heaths (approved):
Couple of nits but nothing blocking. It'd be great if we could make more tests recordable. See ideas, though I don't know enough about the scenarios to know if that's even feasible. Just something to think about. I wrote the initial test resource scripts years back and can help brainstorm ideas if you'd rather.

sdk/storage/azure_storage_blob/tests/blob_client.rs
vincenttran-msft: The root cause of immutability policy being a scenario where we live only record and then make it playback, is because of serious flakiness with consistency issues + these are extremely difficult to get the test to clean up the resources properly because of that consistency issue.

sdk/storage/azure_storage_blob/tests/blob_client_versioning.rs
vincenttran-msft: Ditto

sdk/storage/azure_storage_blob/tests/blob_service_client.rs
vincenttran-msft: This is also another case where we want to use sort of a contrived example-- mostly because the service stats don't become consistent on the Service until a little while -- so we just use an old account that already has stats generated.

sdk/storage/azure_storage_blob/README.md
vincenttran-msft: Noted, if CI passes I won't do anything with this but I can fold it in if relevant.

jalauzon-msft (approved):

PR #4204, opened 2026-04-16
Fix lint issue in main

analogrelay:
The clippy build is failing in main because of these two lints:

* `unnecessary_sort_by` in `azure_storage_blobs`
* `useless_conversion` in `azure_core_test`

These were all trivially fixable with `cargo clippy --fix`, so I did so ;).

### DIFF
--- a/sdk/core/azure_core_test/src/recording.rs
+++ b/sdk/core/azure_core_test/src/recording.rs
@@ -542,7 +542,7 @@ impl Recording {
             TestMode::Playback => {
                 let result = client.playback_start(payload.try_into()?, None).await?;
                 let mut variables = self.variables.write().map_err(write_lock_error)?;
-                variables.extend(result.variables.into_iter());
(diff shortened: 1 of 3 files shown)


kundadebdatta (approved):
LGTM.

heaths:
Not seeing these in 1.94 even with `cargo clippy --all-targets --all-features`, so it appears these are newly elevated and exactly the sort of stylistic lints we have to fix every rust toolchain release. It's exhausting and a useless waste of time. I've talked to them in zulip about stopping these sorts of stylistic warnings (since everyone builds with `RUSTFLAGS=-Dwarnings` but only got disagreement. One idea that came up was to pin lints but, IIRC, was not a straight-forward approach. Something I can look into again, though

heaths (approved):

heaths (approved):

heaths (approved):

PR #4205, opened 2026-04-16
don't start the cosmos emulator on Linux builds

analogrelay:
We made a similar change in `release/azure_data_cosmos-previews`, but it needs to be included here to avoid launching the emulator on Linux builds. We were already marking them as skipped, but still starting/stopping the emulator. This change causes us to skip the entire emulator start process.


tvaron3 (approved):

FabianMeiswinkel (approved):
LGTM

simorenoh (approved):

